PR #12, opened 2015-01-17
Refactoring Dimensioned objects for uniform API

philippjfr:
This is a major refactor of HoloViews to provide a common interface for multi-dimensional objects in HoloViews. For that reason it touches almost every aspect of HoloViews. While all the notebooks now work and the unit tests have been updated a number of features have not been fully tested or still need to be updated, this includes:

* Titles
* Legends
* Reduce, sample, collate
* Support for value dimensions in the Pandas interface


jlstevens:
Ok. I'm happy with all this. Lots of changes but I do think it will be worth it!

PR #14, opened 2015-02-25
 small fix to doc

meduz:


jlstevens:
Thanks!

meduz:
this was more to test if I could easily do a PR on the repo :-)

jlstevens:
Of course you can! We welcome PRs!

I would suggest that you first file your desired functionality in GitHub Issues. Then we can discuss how we want to do things before you start working on a PR.

This is probably good advice for anyone who wants to contribute, so we may want to put it up on the webpage...

PR #23, opened 2015-03-19
Fix link on features page

takluyver:
I'm not quite sure if this link is pointing to what you want, but this makes it point there correctly.


jbednar:
Perfect,  thanks!  I'll merge that as soon as Github lets me!

PR #27, opened 2015-04-04
Unifying live widgets and adding nbagg support

philippjfr:
This pull request introduces three major changes to widgets:

* IPySelectionWidget is deprecated for a number of reasons:
   * It is based on the IPython widget system, which is still subject to change.
   * It made selection of the appropriate widget confusing.
   * It always requires a live notebook.
   * Its layout didn't look any good.

* The SelectionWidget can now render plots dynamically
   * This can be activated via %output widgets='live'
   * It supports live rendering of all current figure backends (i.e. png, svg, mpld3 and nbagg)
   * It supports dynamic caching, the size of the cache can be controlled via the SelectionWidget.cache_size parameter

* Support for nbagg
   * Can be used for static frames and widgets
   * The nbagg matplotlib backend provides widgets for zooming, panning and most interestingly resizing.
   * To use widgets with nbagg, widgets='live' must be selected.

### DIFF
--- a/holoviews/ipython/display_hooks.py
+++ b/holoviews/ipython/display_hooks.py
@@ -16,6 +16,11 @@ try:
 except:
     mpld3 = None
 
+try:
+    from matplotlib.backends.backend_nbagg import new_figure_manager_given_figure
+except:
+    new_figure_manager_given_figure = None
+
 import param
 
 from ..core.options import Store
@@ -143,10 +148,8 @@ def display_figure(fig, message=None, max_width='100%'):
     dpi = OutputMagic.options['dpi']
     backend = OutputMagic.options['backend']
 
-    if backend == 'nbagg':
-        import numpy as np
-        from matplotlib.backends.backend_nbagg import FigureManagerNbAgg, new_figure_manager_given_figure
         manager = new_figure_manager_given_figure(np.random.randint(10**8), fig)
+    if backend == 'nbagg' and new_figure_manager_given_figure is not None:
         manager.show()
         return ''
     elif backend == 'd3' and mpld3:
--- a/holoviews/ipython/display_hooks.py
+++ b/holoviews/ipython/display_hooks.py
@@ -135,6 +135,8 @@ def display_widgets(plot):
         return ScrubberWidget(plot)()
     if widget_mode == 'embed':
         return SelectionWidget(plot)()
+    elif widget_mode == 'live':
+        return SelectionWidget(plot, embed=False)()
 
 
 def display_figure(fig, message=None, max_width='100%'):
--- a/holoviews/ipython/display_hooks.py
+++ b/holoviews/ipython/display_hooks.py
@@ -148,8 +148,9 @@ def display_figure(fig, message=None, max_width='100%'):
     dpi = OutputMagic.options['dpi']
     backend = OutputMagic.options['backend']
 
-        manager = new_figure_manager_given_figure(np.random.randint(10**8), fig)
     if backend == 'nbagg' and new_figure_manager_given_figure is not None:
+        manager = new_figure_manager_given_figure(OutputMagic.nbagg_counter, fig)
+        OutputMagic.nbagg_counter += 1
         manager.show()
         return ''
     elif backend == 'd3' and mpld3:
--- a/holoviews/ipython/display_hooks.py
+++ b/holoviews/ipython/display_hooks.py
@@ -42,10 +42,6 @@ ENABLE_TRACEBACKS=True
 
 
 def animate(anim, dpi, writer, fmt, anim_kwargs, extra_args):
-    if OutputMagic.options['holomap'] == "nbagg":
-        plt.show()
-        return ''
-
     if extra_args != []:
         anim_kwargs = dict(anim_kwargs, extra_args=extra_args)
 
@@ -64,7 +60,7 @@ def HTML_video(plot):
     writers = animation.writers.avail
     current_format = OutputMagic.options['holomap']
     for fmt in [current_format] + list(OutputMagic.ANIMATION_OPTS.keys()):
-        if fmt == 'nbagg' or OutputMagic.ANIMATION_OPTS[fmt][0] in writers:
+        if OutputMagic.ANIMATION_OPTS[fmt][0] in writers:
             try:
                 return animate(anim, dpi, *OutputMagic.ANIMATION_OPTS[fmt])
             except: pass
--- a/holoviews/ipython/display_hooks.py
+++ b/holoviews/ipython/display_hooks.py
@@ -42,6 +42,10 @@ ENABLE_TRACEBACKS=True
 
 
 def animate(anim, dpi, writer, fmt, anim_kwargs, extra_args):
+    if OutputMagic.options['holomap'] == "nbagg":
+        plt.show()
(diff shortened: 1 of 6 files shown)


holoviews/ipython/widgets.py
```diff
@@ -506,20 +340,56 @@ def __call__(self):
...
     to json and dynamically loaded from a server.
     """
 
+    embed = param.Boolean(default=True, doc="""
+        Whether to embed all plots in the Javascript, generating
+        a static widget not dependent on the IPython server.""")
+
+    cache_size = param.Integer(default=100, doc="""
+        Size of dynamic cache if frames are not embedded.""")
+
```
jlstevens: Is ``cache_size`` something to control via the `%output` magic or not?

holoviews/ipython/widgets.py
```diff
@@ -506,20 +340,56 @@ def __call__(self):
 
 
 
-class SelectionWidget(ScrubberWidget):
+class CustomCommSocket(CommSocket):
+    """
+    A CustomCommSocket is required to suspend displaying the nbagg
+    output until the HTML has been rendered by the widget.
+    """
```
jlstevens: I still have no idea what this does. Either needs more documentation or some link to what this is about.

holoviews/ipython/widgets.py
```diff
@@ -453,47 +249,85 @@ class ScrubberWidget(NdWidget):
...
 
 
-    def get_frames(self, id):
-        use_mpld3 = OutputMagic.options['backend'] == 'd3'
-        frames = {idx: frame if use_mpld3 or self.export_json else
-                  str(frame) for idx, frame in enumerate(self.frames.values())}
+    def render_html(self, data):
+        # Set up jinja2 templating
+        import jinja2
+        path, _ = os.path.split(os.path.abspath(__file__))
```
jlstevens: Is using ``__file__`` here the safest options?

holoviews/ipython/widgets.py
```diff
@@ -257,176 +218,11 @@ def get_plot_size():
...
-        else:
-            self.image_widget.value = self._plot_figure(self.keys.index(checked))
-
-
-
-class ScrubberWidget(NdWidget):
+    NdWidget is an abstract base class implementing a method to find
+    the dimensions and keys of any ViewableElement, GridSpace or
+    UniformNdMapping type.  In the process it creates a mock_obj to
+    hold the dimensions and keys.
```
jlstevens: Why is this class ``Parameterized`` if you aren't declaring any parameters? Maybe some of you ``self`` attribute should be parameters instead?

holoviews/ipython/widgets.py
```diff
@@ -3,22 +3,16 @@
 
 import numpy as np
 
+from matplotlib import pyplot as plt
+from matplotlib.backends.backend_nbagg import CommSocket, new_figure_manager_given_figure
+
 try:
     import IPython
     from IPython.core.display import clear_output
+    from IPython.kernel.comm import Comm
```
jlstevens: Have you tested this with IPython 3? As we sure this will be stable?

holoviews/ipython/jsslider.jinja
```diff
@@ -32,31 +62,29 @@
...
+			data = JSON.parse(data);
+		    } else {
+			data = {0: data};
+		    }
+		    this.frames = data;
+		    this.update()
+		}
+	    }
+	    callbacks = {iopub: {output: $.proxy(callback, this)}};
+	    var cmd = "holoviews.ipython.widgets.SelectionWidget.widgets['" + this.id + "'].update(" + this.keyMap[key] + ")";
```
jlstevens: This is a fairly crazy callback but at least I think it makes sense...

holoviews/ipython/display_hooks.py
```diff
@@ -146,7 +143,13 @@ def display_figure(fig, message=None, max_width='100%'):
     dpi = OutputMagic.options['dpi']
     backend = OutputMagic.options['backend']
 
-    if backend == 'd3' and mpld3:
+    if backend == 'nbagg':
+        import numpy as np
+        from matplotlib.backends.backend_nbagg import FigureManagerNbAgg, new_figure_manager_given_figure
```
jlstevens: I would move this to a conditional import at the top of the file and then check for None here. E.g if ``FigureManagerNbAgg` is None`` (because the import failed) you could then generate an appropriate message here.

holoviews/ipython/display_hooks.py
```diff
@@ -146,7 +143,13 @@ def display_figure(fig, message=None, max_width='100%'):
...
+        manager = new_figure_manager_given_figure(np.random.randint(10**8), fig)
```
jlstevens: You could create a numpy ``RandomState`` instance here with a fixed seed to get deterministic figure ids. Or use a counter...Or maybe keep track of the figure ids already used to avoid clashes.

holoviews/ipython/jsslider.jinja
```diff
@@ -32,31 +62,29 @@
...
-                mpld3.draw_figure(this.img_id, this.frames[this.current_frame]);
-            }else {
-                $("#" + this.img_id).html(this.frames[this.current_frame]);
-            }
-        }
+	if(this.cached) {
+	    this.current_frame = this.keyMap[key];
+            this.update()
+	} else {
+	    var kernel = IPython.notebook.kernel;
```
jlstevens: You didn't see to need ``IPython.notebook.kernel`` before (you used methods on something called ``mpld3``). Does this mean the mechanism by which this works has fundamentally changed?

holoviews/ipython/jsslider.jinja
```diff
@@ -32,31 +62,29 @@
...
+	    if(this.nbagg) {
+		function callback(msg) { }
+	    } else {
+		function callback(msg){
+		    var data = msg.content.data['text/plain'].slice(1, -1);
```
jlstevens: Maybe add a comment here to explain what the slice is for...

holoviews/ipython/widgets.py
```diff
@@ -528,11 +398,22 @@ class SelectionWidget(ScrubberWidget):
...
-                                  for idx, k in enumerate(self.keys))
+        self.nbagg = OutputMagic.options['backend'] == 'nbagg'
+        self.frames = {}
+        if self.embed:
+            frames = {idx: self._plot_figure(idx)
+                      for idx in range(len(self.keys))}
+            self.frames = self.encode_frames(frames)
+        elif self.nbagg:
+            fig = self.plot[0]
+            self.manager = new_figure_manager_given_figure(np.random.randint(10**8), fig)
```
jlstevens: Another use of random numbers where some deterministic method may be better.

jlstevens:
Ok. Mostly looks good and it is nice to see the old IPython widgets go away. My main concerns are the new imports from IPython and the concern that maybe the nbagg API isn't completely stable yet.

holoviews/ipython/widgets.py
philippjfr: Yes, we can add this later.

jlstevens:
And we also need to make sure export with widgets work before the merge...

jlstevens:
Ok...all looks good. Now merged.

PR #64, opened 2015-05-12
Removed np.array(list(...)), not needed.

basnijholt:
Could be confusing, the important thing is that a list is used in ``Contours``.


philippjfr:
As you can tell now it is needed for Python 3 compatibility. I think the np.array part could be removed but the zip object does need to be cast to a list in Python 3.

PR #106, opened 2015-05-24
Fixing some more typo's in tutorials.

basnijholt:


basnijholt:
There are lots of things that are discutabel such as:
*  sublabels
*  errorprone
*  filenames
*  datastructure
*  dataframes
*  pairgrid
*  inconsequent use of HoloViews vs Holoviews vs holoviews

doc/Tutorials/Exploring_Data.ipynb
```diff
@@ -432,7 +432,7 @@
      "cell_type": "markdown",
      "metadata": {},
      "source": [
-      "Now that we have removed the redundant Dimension we can create an NdLayout indexed just by the date:"
```
jbednar: I think "an NdLayout" is correct, not "a NdLayout", because we pronounce it en-dee-layout.  But that can be fixed afterwards.

jbednar:
I think there are clear answers to all but one of the ones in your list:

*  sublabels: I think this is correct; it's a new word we made up, but it's well defined as a shorthand for "subfigure label", which would be too awkward to keep saying
*  errorprone: this is incorrect, and should be "error-prone"
*  filenames: this is correct; at least in computer science that concept can be a single word, and is usually clearer that way.
*  datastructure: this is incorrect; it should always be "data structure".
*  dataframes: not sure about this one; it's a new word created to refer to a data structure in Pandas.  It should probably usually be DataFrame (when it's talking about a specific Pandas object), or else reworded.
*  pairgrid: this should only ever appear in fixed-width font, because it's a keyword from the software, so if it ever appears in regular text the font should be changed or it should be reworded
*  inconsistent use of HoloViews vs Holoviews vs holoviews: it should consistently be "HoloViews", as the name of the software, or "holoviews", if referring to the Python package (and if so always typeset in fixed-width font), never Holoviews.

Issue #232, opened 2015-08-19
Conversions should accept Dimension objects not just strings


vascotenner:
In the conversion from a table to a curve, one can select which vdim has to be used. However, this can only be a string, not a dimension. This will not work:
```
hv.Layout([params.to.curve(kdims='Power',vdims=vdim)  for vdim in params.vdims])
```
But this works:
```
hv.Layout([params.to.curve(kdims='Power',vdims=vdim.name)  for vdim in params.vdims])
```

philippjfr:
True, this is the case almost everywhere. This would be a fairly straightforward project for someone new to HoloViews development to tackle. It's not a priority for me so if you or anyone else wants to make this work please submit a PR.

philippjfr:
Addressed in the commit referenced above. Closing.

PR #233, opened 2015-08-20
Selecting dimensions

vascotenner:
To select a dimension, it is now possible to supply a Dimension instead of a name. Furthermore it is allowed to create empty units (which is the default)

This fixes issue #232

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -76,7 +76,7 @@ class Dimension(param.Parameterized):
         may be an inbuilt constructor (such as int, str, float) or a
         custom class object.""")
 
-    unit = param.String(default=None, doc="""
+    unit = param.String(default=None, allow_None=True, doc="""
         Optional unit string associated with the Dimension. For
         instance, the string 'm' may be used represent units of meters
         and 's' to represent units of seconds.""")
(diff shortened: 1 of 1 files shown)


philippjfr:
Hadn't even thought about this but yes in most cases this should work. I think that test failure is random, I'll restart the build and hopefully it'll pass. Otherwise I'd be happy to merge this.

vascotenner:
What does coverage mean?

philippjfr:
I'm not entirely sure why coverage is failing, a small decline in testing coverage shouldn't cause a failure in itself. I'm okay merging this, next time please split the changes into separate commits though. The change to units is a valid one but shouldn't have been mixed in with the other change.

philippjfr:
I assume you're splitting it into separate commits? I'd have been happy to merge it in this case. Either way thanks for the contribution!

vascotenner:
Hmm, I thought that github could remember to which commits I was referencing. I restored the branch again. For the next PR I will split things properly.

philippjfr:
Hmm, I just thought about it again and realized this PR won't actually do what you want it to do. The ``get_dimension`` method simply returns the ``Dimension`` object given the dimension name or index. If you already have the Dimension object there's no point looking it up again using ``get_dimension``. To allow different methods to accept Dimension objects and the names you'll probably have to look at the particular method you're want this to work with. However we should probably consistently support this everywhere, so I'm still happy to merge this PR as it's a first step towards this.

PR #235, opened 2015-08-21
Replaced mailinglist with gitter channel in README

vascotenner:
The mailinglist is not in use anymore, while the gitter channel is.


philippjfr:
Thanks for catching that, merging now.

jlstevens:
Thanks! I agree with this change and the mailing list has been updated to suggest using gitter instead.

Issue #238, opened 2015-08-21
operation.gradient and cyclic ranges


vascotenner:
The operation.gradient is not working on Images with cyclic ranges, and not working properly on images with no cyclic ranges.

In element.py:319 to 329 the logic is wrong. If the input is not a cyclic range, then it should not try to wrap it.

I did not supply a PR, because also when the image has a cyclic range, but an asymnetric range, it requires some thought what to do.

jlstevens:
The implementation of the gradient operation inherited some rather old code that pre-dated HoloViews. It was probably never designed to handle cyclic quantities so I would be very happy to merge any PR that helps generalizes this operation properly.

I am also happy to fix the lines you mention although I would prefer to fix generalize this code properly. Even a PR that raises an exception when an asymmetric cyclic range is detected would be better than what we have now! If you are willing to submit such a PR, the issue of asymmetric cyclic ranges could be addressed at a later date,

vascotenner:
Furthermore, it would be useful to to extend this function also to different dimensions (1D curves for example).

jbednar:
The gradient code *was* originally designed for cyclic quantities, because it was first implemented for 2D maps of orientation (a pi-periodic quantity).  So it's the non-cyclic case that would need the most attention.  Can you explain more specifically what you mean by an asymmetric cyclic range, perhaps with  some code and output and what the problem is in that case?

vascotenner:
For the non_cyclic range, there is no wrapping needed. In that case the conditional codeblock should not run.

One of the problems is that Dimension.range is a tuple, not an float.
For range=(-5,100), the code should wrap between -5 and 100, not between 0 and 100. This can be obtained by first shifting the values to the range 0,105, and then wrapping, and then back again.

jlstevens:
> The gradient code was originally designed for cyclic quantities

Yes sorry! Wasn't thinking properly when I typed that.

> For the non_cyclic range, there is no wrapping needed. In that case the conditional codeblock should not run.

That sounds correct. As for the handling of the tuple, it sounds plausible but I would need to think about it a bit more carefully first...

vascotenner:
I submitted a PR with my prosition.

jbednar:
Yes, what you suggest for -5..100 sounds correct to me.

vascotenner:
Now there is taken into account for the wrapping. Small differences are preferred above large ones.

philippjfr:
The PR has merged a while back, so I'll close the issue.

PR #239, opened 2015-08-21
Fixing the gradient for cyclic ranges

vascotenner:
The gradient operation wraps the gradient when the input has a cyclic range see #238

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -316,9 +316,12 @@ class gradient(ElementOperation):
         dx = np.diff(data, 1, axis=1)[0:r-1, 0:c-1]
         dy = np.diff(data, 1, axis=0)[0:r-1, 0:c-1]
 
-        cyclic_range = None if not matrix_dim.cyclic else matrix_dim.range
+        cyclic_range = None if not matrix_dim.cyclic else np.diff(matrix_dim.range)
         if cyclic_range is not None: # Wrap into the specified range
-            raise NotImplementedError('Cyclic ranges are currently not supported')
+            # shift values such that wrapping works ok
+            dx += matrix_dim.range[0]
+            dy += matrix_dim.range[0]
+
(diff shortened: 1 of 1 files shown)


holoviews/operation/element.py
```diff
@@ -316,8 +316,12 @@ def _process(self, matrix, key=None):
         dx = np.diff(data, 1, axis=1)[0:r-1, 0:c-1]
         dy = np.diff(data, 1, axis=0)[0:r-1, 0:c-1]
 
-        cyclic_range = 1.0 if not matrix_dim.cyclic else matrix_dim.range
+        cyclic_range = None if not matrix_dim.cyclic else np.diff(matrix_dim.range)
```
jlstevens: Using ``np.diff(matrix_dim.range)`` if fine assuming that both values are numeric. This assumption may be violated as the range tuple is allowed to use a ``None`` value to indicate lower/upper bounds that are not set. For this reason, I would have an extra line with something like:

```if  matrix_dim.cyclic and (None in matrix_dim.range):
   raise Exception('Cyclic range must be specified to compute the gradient over cyclic quantities')

jlstevens:
I've made one comment but other than that, this PR looks fine to me.

jbednar:
Jean-Luc's suggestion sounds good.  The code looks good to me, but it sure would be nice to see test cases. :-)

vascotenner:
I just did some test, there are some fundamental problems. I prefer that if a gradient is zero, it should stay zero, no matter what wrapping is done!

vascotenner:
Zero gradient stays zero, and small steps are preferred above large ones.

jlstevens:
Do we think this is now ready to merge?

vascotenner:
This PR makes that the gradient operator works again, and is ready to merge. Extra tweaks should be proposed in additional PRs.

philippjfr:
Looks fine to me. Happy to merge.

Issue #254, opened 2015-08-28
Overview of holomap


vascotenner:
Quite often, I just want to have an overview of my holomap. For that I use the layout or overlay or grid method. Most of the time, I want to show all dimensions (often 1). Typing in that dimension is much work. I propose that the default behaviour of these functions is to show all elements of the holomap.

jlstevens:
Yes, I completely agree! I've checked the signature of both these methods and this should be trivial to implement.

philippjfr:
Just merged your PR. Will close this issue.

PR #255, opened 2015-08-28
Layout, overlay and grid of a HoloMap now reduces all key dimensions by default.

vascotenner:
This makes it possible to easily get an overview of the contents of a holomap, as discussed in #254

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -517,9 +517,11 @@ class Dimensioned(LabelledData):
 
 
     def _valid_dimensions(self, dimensions):
-        "Validates key dimension input"
-        if not dimensions:
-            return dimensions
+        """Validates key dimension input
+        
+        Returns kdims if no dimensions are specified"""
+        if dimensions is None:
+            dimensions = self.kdims
         elif not isinstance(dimensions, list):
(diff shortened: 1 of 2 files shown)


vascotenner:
Should this be mentioned in the docstrings?

jlstevens:
> Should this be mentioned in the docstrings?

Yes I think a docstring is needed and as I mentioned in the issue, I am in favour of this change.

philippjfr:
If you look at Line 519 in core/dimension.py, there is a method called ``_valid_dimensions``, instead of duplicating this three times in each of these method you could replace these lines:

```
        if not dimensions:
            return dimensions
```

with your changes:

```
if dimensions is None:
    dimensions = self.kdims
```

While you're at it you could also rename the method to ``_validate_kdims``. If you want I can make these changes quickly and push them but I'm of course also happy for you to amend this PR, in which case I'll merge it immediately.

vascotenner:
I implemented the philippjfr's suggestion.

In core/element.py line 126, there is some code that uses the _valid_kdims metho, but I do not understand how it works, but I guess it is broken by this commit

philippjfr:
Seems fine, merging now.

PR #275, opened 2015-09-27
Bokeh Callbacks

philippjfr:
This PR adds support for Bokeh callbacks to HoloViews, allowing both pure javascript callbacks and callbacks to the IPython kernel. The design allows defining various attributes on the plot, data source and callback object to listen to, which are sent to Python allowing the user to modify any the plot object or data source. It also automatically chains multiple callbacks defined on one plot object. This means the design should easily extend to support what we call ``Streams``, allowing the user to link any callback to a DynamicMap, even across plots.

The two baseclasses that are added are:

* ``Callback``: A ``ParameterizedFunction`` which defines the attributes to subscribe to, any custom js code and ``__call__`` method, which defines what operation to apply to the data, returning the bokeh plot objects that are to be updated.
* ``Callbacks``: A Parameterized object allowing the user to define callbacks for various plot attributes as parameters. It initializes the Callback functions, setting up any javascript code required to listen to the requested bokehjs attributes, sets the callback on the corresponding plot object and chains callbacks if multiple are defined on one plot object.

Currently only callbacks on ranges are implemented and two examples are provided. One downsamples any Image depending on the current zoom level, the other applies downsampling to any Points or Scatter object. I've put together a little video demo of the callback to demonstrate what the downsampling looks like in action, you can view it [here](https://sendvid.com/4l7u6ipi). As you can see the downsampling callbacks allow me to display and interactively explore a 21600x10800 pixel image and 1M scatter points with great responsiveness. It even works for axes shared across plots and multiple objects being downsampled at once.

[HoloViews Callback Demo](https://sendvid.com/4l7u6ipi)


jlstevens:
The functionality here is great and I see you've implemented some of the API we discussed together earlier on. I'm not certain it should be merged with master (or at least not just yet) as I think we should think this through this system carefully.

Given that this functionality is tied to Bokeh, there are other systems that will make use of the callbacks (e.g the proposed ``Stream`` dimensions for ``DynamicMap``) how about we update the existing bokeh branch and merge with that? The bokeh branch will then be for testing these more advanced plotting features now that all the basic bokeh plots have been merged with master.

philippjfr:
This PR is ready to merge, we might revisit the API after the release, but it works and we can improvements and finalize the API after it's been in usage for a while.

PR #276, opened 2015-09-28
Fixed Timer reporting for long durations

jbednar:
The %timer magic output is mixed up for long durations (anything over an hour).  E.g. for a 9-hour computation:

```
In [1]: %timer
Out [1]:
Timer elapsed: 08:533:41
```

Someone should check the logic in my fix, but I think it will now say ``08:53:41`` (which is how long I *think* the computation took in this case).

### DIFF
--- a/holoviews/ipython/magics.py
+++ b/holoviews/ipython/magics.py
@@ -696,12 +696,11 @@ class TimerMagic(Magics):
 
     @staticmethod
     def elapsed_time():
-        elapsed = time.time() -  TimerMagic.start_time
-        minutes = elapsed // 60
+        seconds = time.time() -  TimerMagic.start_time
+        minutes = seconds // 60
         hours = minutes // 60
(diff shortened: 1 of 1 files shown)


jbednar:
I tried following up on that Travis build failure, but didn't see much to go on.  It seems to indicate that testing the notebooks fails on Python 3.4, which I don't *think* should be related to this pull request, since I believe that the timer output is explicitly skipped over during notebook testing...

jbednar:
Looks like the checks are passing now; no idea what that was about.

jlstevens:
Oops! I guess the thought of waiting more than an hour for something to finish in a notebook was simply too horrible for me to contemplate! :-)

And yes, I re-ran the Travis build because it was one of those annoying transient errors that sometimes crop up. As you said, that failure was nothing to do with your code so I'll go and merge this now.

jbednar:
Thanks!

PR #278, opened 2015-10-01
DynamicMap

philippjfr:
This PR adds the new ``DynamicMap`` class, which allows generating frames on the fly when working with widgets. The ``DynamicMap`` has two operating modes or intervals:

* 'open': In open mode the ScrubberWidget allows the user to request a new frame via the play and next frame buttons. When creating the DynamicMap in open mode the user should supply an iterator that returns a new key and Element when ``next`` is called. This mode allows for streaming data for example from a simulator.

* 'closed': In closed mode the user should define the valid Dimension ranges or values. Using the ``SelectionWidget`` the user may then select any value in the space defined by the Dimensions, allowing much larger parameter spaces with near continuous sampling to be explored.

Both modes already work but there are a few to-do items left:

1. Currently the cache is infinite which means that all frames are buffered both server-side and client-side, there should be options to limit the buffer size and only buffering every n-th frame.
2. ~~Currently nothing communicates a StopIteration to javascript so the scrubber will continue running and expand the slider even when the DynamicMap has run out of new frames.~~


jbednar:
This sounds great!  But it took changes to 17 files to do this?  Ack!

jlstevens:
Ok, I've had a look at this and I have a few notes to make. Firstly, it seems that the updated widgets don't seem to work in IPython 2 which is rather unfortunate.

Secondly, I feel the API could probably be improved further and I've discussed the following improvements together with Philipp:

*  You shouldn't need to specify the interval mode as it can be inferred. Generators always imply open mode.
* We don't see much use for 'both' mode at this time where you have a class that is both a iterator/generator (has a ``next()`` method) and a callabled.
* For a ``Layout`` of ``DynamicMaps`` reflecting the state of an underlying model (e.g a topographica simulation), you don't want each map to call ``next()`` advancing the simulator. In this case, what you want is a callable so that you can specify the interaction number as the argument  but you also want to be in 'open' mode.
* To specify this, any callable where *any* dimension range is the default of ``(None,None)`` will be treated as open mode where the callable is expected to take a single value as input (the iteration number). Then a simulator will know whether to advance the simulation time or not. For closed mode, *all* dimensions need bounded ranges.
*  In this mode, the callable can still return the key together with the element. The key dimension is with reference to this key value and not the implicit counter on the input side.

We also discussed the meaning of dimension ranges in HoloMaps which are currently ignored. We came to the conclusion that the ``soft_range`` and ``range`` together could be used for validation: if you try to add a key outside the bounds specified by the intersection of the ``soft_range`` and ``range``, an exception (``StopIteration``?) is raised. Note that this semantics allows you to use dynamic maps with callables and range ``(None, None)`` by specifying the ``soft_range`` only. This is behavior we need to implement for ``HoloMap`` as well as ``DynamicMap``.

I think that summarizes what we discussed for now!

jlstevens:
I've been working through the proposal above and making good progress. One quick bug report for Dimensions though: 

```python
kdims=[hv.Dimension('size', range=(1,5), type=float)
```

This results in a slider that snaps to integers despite the explicit type declaration saying otherwise.

jlstevens:
Ok, I've gone and rewritten ``DynamicMap`` according to the proposal above.

I am quite happy with it! Not having the specify the mode explicitly is very nice, making ``DynamicMap`` easier to use. It is currently working with all the tests I have for all the different modes.

As it is working with an improved API, I am happy to see it merged whenever you like. There are a few outstanding issues, most of them are minor:

**Bug fixes**

* The dimension type issue above.
* An odd bug where the first frame is the same as the last frame.

**To implement**

* When using ``__getitem__`` with a key set, ``DynamicMap`` should regenerate the data (closed mode only).

* Caching needs to be improved on the client-side (i.e in the Javascript code).

These are all small jobs that I am happy to see implemented on master.

philippjfr:
Okay I've fixed all the bugs you mentioned and have got the tests passing. The other two items are features and can be implemented on master (or as separate PRs). I'll go ahead and merge.

jlstevens:
Ok great! We may still want to use this PR to decide how we want caching on the client-side to work...

Issue #269, opened 2015-09-14
Data API


philippjfr:
Now that we've made HoloViews plotting backend independent it's time to talk about generalizing the data backends. Reviewing the current situation I have to say we haven't done quite as good a job as we could have done but all the foundations are there to unify things very nicely and iron out kinks like the lack of support for categorical or date types in various charts. As part of the dataframe refactor we've already begun looking at this but it's been more of an ad-hoc way of patching in support for dataframes.

Today it occurred to me that we already have three different storage formats for table-like data and a lot of code built up to convert and operate on the different data types. They include:

1. ``NdElement`` implements a table as an NdMapping, i.e. a fancy dictionary. For large tables this is very slow. 

2. ``Chart`` objects while never displayed as such are effectively multiple columns stored in an NxD numpy array, where N is the number of samples and D the number of dimensions/columns. 

3. ``DFrame`` as it's name suggests simply uses a pandas dataframe as the underlying datasource.

It would be nice if all subclasses of these Element types could be unified to use any of these datasources. Currently different Element types use different data formats and it's all a bit of a mess and not very flexible. If we are successful with this the following Elements will interchangeably support all three data formats: Curve, Points, VectorField, Bars, Table, ErrorBars, Spread, Scatter, Scatter3D, Distribution, TimeSeries and with some minor adjustments Histogram.

I'm not entirely certain about the mechanism yet but I envision that a we introduce a new wrapper class which dispatches any method calls to one of the three baseclasses depending on the type of the data. We would provide a common API for each of the backends so custom Elements, plotting code and the user can access the data in very well defined ways independent of the data backend. Luckily we've generally made sure that such a common API exists even if we haven't used it consistently.

The current API for Element types is this:

   * ``dimension_values``: Returns an array, list or pandas series of one column
   * ``range``: Returns the minimum and maximum values along a dimension excluding NaNs.
   * ``select``: Same as ``__getitem__`` but works for one or multiple dimensions.
   * ``dframe``: Converts data to a dataframe
   * ``collapse_data``: Applies a function across a list of Element.data attributes
   * ``reduce``: Applies a reduce function across one or multiple axes
   * ``sample``: Returns a Table with of the samples specified as lists
   * ``groupby``: Groups by the values along one or multiple dimensions returning an NdMapping type indexed by the grouped dimensions with Elements of the same type as values. (not implemented for Charts)
   * ``reindex``: Reorder specified dimensions dropping unspecified dimensions. (not implemented for Charts)
 
[...]

Issue #282, opened 2015-10-08
API for static HTML export


jlstevens:
Philipp's has recently pulled out some more functionality out of the ipython subpackage involved in  HTML generation. As this functionality is useful and something we want to support in general (so HoloViews is not necessarily tied to IPython/Jupyter) we should consider implementing a useful API for our users. Here is what I suggest:

```python
from holoviews.plotting import Renderer
data =  Renderer.static_html(obj)  # Dictionary  e.g data['css'], data['js']
html =  Renderer.static_html(obj, html_only=True)  # Standalone HTML
```

One other suggestion is a ``template`` argument which defaults to ``'default'`` (or ``'basic'``?) which defines the overall HTML skeleton. This would allow you to specify a custom template as a formatter with the appropriate fields (i.e ``'css'`` and ``'js'`` in the current example).

Lastly, I might consider splitting the HTML up a little bit more. Perhaps a ``'header'`` field could correspond to the HTML header used to load the appropriate JavaScript libraries?

**Fix:** I forgot to add the ``obj`` argument to define what you want to export!

philippjfr:
For reference [here](http://philippjfr.com/work/work-in-progress/html-export/) is an example of the current API for exporting a standalone HTML file.

jlstevens:
Note that above the methods are on ``Renderer`` (i.e the class) but this should also work for renderer instances. In other words, this will have to be implemented as a ``@bothmethod``

nataliabielczyk:
hi! I am on the user end, and I would like just to ask if this API is still under construction. I tried to run the commands given in the first post of this thread, and I got an error
'type object 'Renderer' has no attribute 'static_html''

philippjfr:
Yes, this has not been implemented yet. Hopefully we'll get around to it soon.

philippjfr:
This has been implemented as part of #311, closing.

Issue #280, opened 2015-10-08
Export archive


vascotenner:
I cannot get export an archive. No files are exported:
```python
import holoviews as hv
%load_ext holoviews.ipython
import numpy as np
from holoviews.ipython import display_hooks
display_hooks.ABBREVIATE_TRACEBACKS=False
hv.archive.auto()
```
```Automatic capture is now enabled. [2015-10-08 09:47:07]```
```python
%%opts ErrorBars [yrotation=15, xrotation=45]
hv.ErrorBars((0,1,.5,3))
```
```
WARNING:root:ErrorBars01244: Could not pickle custom style information.
```
+image
```python
hv.archive.export()
```
```
Export name: '{notebook}'
Directory    '/media/data2/tenner/data/ipython-notebooks'

If no output appears, please check holoviews.archive.last_export_status()
```
```python
hv.archive.last_export_status()
```
```
Status of the last call to holoviews.archive.export is unknown.
(Re-execute this method once kernel status is idle.)
```
```python
hv.archive.contents()
```
```
ErrorBars.svg : image/svg+xml
ErrorBars.hvz : application/zip
```

jlstevens:
Thanks for reporting this! I'll have to look into it...

philippjfr:
Yes, thanks I meant to file an issue about problems with archiving myself but forgot about it. Here's a related issue: https://github.com/ioam/topographica/issues/631

vascotenner:
The export is not totally fixed in #284. When I run the Exporting tutorial, it does not export in cell 13. Also ```hv.archive.last_export_status()``` does not give usefull output. However, when I add the lines:
```
holoviews = hv
holoviews.archive.export()
```

It exports the data, but in the directory {notebook}, instead of Exporting.

```hv.archive.last_export_status()``` now gives the right output.

**Comment by philippjfr** [comment]
Thanks, for looking into this. We're trying to get a release candidate out by the end of this week, so hopefully @jlstevens will get to this issue soon.

**Comment by jlstevens** [comment]
I've finally had a chance to have a look...

I suspect the issue is that IPython/Jupyter have changed their JavaScript API needed for the fancier parts of the export (trying to get the notebook name and capturing the active notebook). If in the tutorial, you set the options like this instead:

```python
hv.archive.auto(skip_notebook_export=True, export_name='Exporting')
```

Then it works again. I feel we ought to document this in the tutorial - I don't like the Javascript code we are using and I don't trust it not to break between Jupyter releases.

[...]

Issue #266, opened 2015-09-11
Document renderers and display function


philippjfr:
Looking through the traffic for this repository I saw that using the issue interact (https://github.com/ioam/holoviews/issues/62) is one of the most commonly visited ones. We should add more documentation about rendering manually and the display function.

wmayner:
Bumping this issue, since it's very old :)

philippjfr:
Very old indeed, the main problem has always been finding a good place to put this material. Do you have any suggestions?

wmayner:
I'd say the most intuitive place to put it would be as a section on the [Creating interactive dashboards](http://holoviews.org/user_guide/Dashboards.html) page.

PR #284, opened 2015-10-18
Data API

philippjfr:
While not yet ready to be merged this PR will implement the proposals laid out in #269. I'll reproduce the main points of discussion here and outline a to-do list.

The core of the proposal is to add a core ``Columns`` Element class, which subsumes the functionality of the ``Chart``, ``DFrame`` and ``NdElement`` classes. This will allow a wide range of Element classes to interchangeably use one of three data formats:

* An array based format of NxD numpy array, where N is the number of samples and D the number of dimensions/columns.

* A dataframe based format with N samples and D columns, which should work with pandas, dask and blaze dataframes.

* A pure-Python NdElement based data format storing the data as keys and values of an OrderedDict.

This would mean the Curve, Points, VectorField, Bars, Table, ErrorBars, Spread, Scatter, Scatter3D, Trisurface, Distribution, TimeSeries and Regression Elements would immediately support all three data formats.

The data format specific implementations of the API are implemented on utility classes, this will include:

* dimension_values: Returns an array, list or pandas series of one column
* range: Returns the minimum and maximum values along a dimension excluding NaNs.
* select: Same as ``__getitem__`` but works for one or multiple dimensions.
* groupby: Groups by the values along one or multiple dimensions returning an NdMapping type indexed by the grouped dimensions with Elements of the same type as values. (not implemented for Charts)
* dframe: Converts data to a dataframe
* collapse_data: Applies a function across a list of Element.data attributes
* reduce: Applies a reduce function across one or multiple axes
* sample: Returns a Table with of the samples specified as lists
* reindex: Reorder specified dimensions dropping unspecified dimensions. (not implemented for Charts)
* drop_dimension: Drops specified dimension(s)/column(s)
* add_dimension: Adds named column at specified position with supplied value(s)
* closest: Returns the closest samples given a list of coordinates.
* shape: NxD shape of data

and conversion methods, which can take an optional flag to return a clone of the Element with converted data:

* array: Returns data as an array.
* dframe: Returns data as a DataFrame.
* mapping: Returns data as NdElement.

Backwards incompatible changes:

* Various operations no longer return ``ItemTables``, the variable return type is confusing and makes the implementation very messy. All operations (select, reduce, sample) return one of two types of data either a modified Columns object or a single scalar.

Other features:

* Currently Elements only support slicing of key dimensions, this is a common annoyance mentioned by several people. I propose we extend the slicing behavior of our types to the following with the following proposal. The ``__getitem__`` method will now check whether the first or n-th index (where n is the number of key dimensions) matches an existing dimension by name, allowing the user to select particular columns. Passing a dimension as a first index will return the values along that dimension, this matches the behavior of DataFrames and will allow us to provide an easy interface for more complex Boolean slicing, e.g.: ``columns[np.logical_or(columns['x'] > 5, columns['y'] > 10)]``. Passing a dimension as the n-th index preserves the current behavior when slicing tables, allowing you to select a particular value dimension. If the non-key indexes don't match any existing dimension they are used to slice along the vdims.

* Collapse/reduce work via the same mechanism, they are basically a groupby operation over one or more key dimensions of an Element or list of Elements, then applying an aggregation function.

One of the major sticking points is how the constructor should behave for different data formats, currently this is the implemented behavior showing the mapping between an input type and the storage format that's used (some come with fallbacks if a certain condition isn't met):

* ndarray -> ndarray
* DataFrame -> DataFrame
* NdElement -> NdElement
* dict -> NdElement
* list of scalar values -> ndarray
* list of tuples for each row -> ndarray (if numeric) -> DataFrame (if available) -> NdElement
* tuple of arrays/lists -> ndarray (if numeric) -> DataFrame (if available) -> NdElement 
* list of key value tuples (i.e. tuple of tuples) -> NdElement

The current to-do list is as follows:

* [x] Get nosetests passing
* [x] Test whether all affected Element types can be constructed with all data types.
* [x] Test whether all affected Element types display using both plotting backends.
* [x] Unify TableConversion interface with DFrame conversions
* [x] Implement data-type agnostic comparisons for Columns.
* [x] Refactor utility classes using classmethods rather than using instances (easier for now)
* [x] Write comprehensive set of unit tests providing 100% coverage of Columns module (Okay maybe not 100%, let's settle for 99%)
* [x] Achieve feature-parity between backends:
   * [x] Constructors
   * [x] getitem/select
   * [x] dimension_values
   * [x] range
   * [x] shape
   * [x] sample
   * [x] reindex
   * [x] add_dimension
   * [x] closest
   * [x] dframe
   * [x] collapse
   * [x] reduce
   * [x] array
   * [x] aggregate
   * [x] sort
   * [x] Add index dimension to NdElement storage format to allow duplicate entries (numpy and pandas support this by behavior already).
   * [x] mapping
   * [x] concatenate
* [x] Get notebook tests passing:
   * [x] Composing Data
   * [x] Containers
   * [x] Continuous Coordinates
   * [x] Elements (8 sorting related failures)
   * [x] Exploring Data (2 NdMapping.info output changes)
   * [x] Exporting
   * [x] Introduction
   * [x] Options
   * [x] Pandas Conversion (16 sorting related failures)
   * [x] Sampling Data (18 sorting related failures)
   * [x] Showcase
* [x] Decide on good naming for the new module and classes
* [x] Document main Columns class and utility classes.

Improvements to plotting:

* [x] Support for datetime axes
* [x] Support for linked brushing when same DataFrame/NdElement is used across Layout.

That covers it for now. I won't have any time to work on this in a concentrated way so I'll be slowly ticking off the to-do list as I find time. In the mean-time we can think about the naming.

Documentation and notebooks:

* [Profiling notebook](http://philippjfr.com/work/work-in-progress/columns-profiling/)

### DIFF
--- a/doc/Tutorials/Sampling_Data.ipynb
+++ b/doc/Tutorials/Sampling_Data.ipynb
@@ -462,7 +462,7 @@
     "extents = (0, 0, 10, 10)\n",
     "img = hv.Image(np.random.rand(10, 10), bounds=extents)\n",
     "img_coords = hv.Points(img.table(), extents=extents)\n",
-    "img + img * img_coords * hv.Points([img.closest((5,5))])(style=dict(color='r')) + img.sample([(5, 5)])"
+    "img + img * img_coords * hv.Points([img.closest([(5,5)])])(style=dict(color='r')) + img.sample([(5, 5)])"
    ]
   },
   {
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -349,7 +349,7 @@ class DataColumns(param.Parameterized):
         elif isinstance(data, Columns):
             data = data.data
         elif isinstance(data, Element):
-            data = {d: data.dimension_values(d) for d in kdims+vdims}
+            data = tuple(data.dimension_values(d) for d in kdims+vdims)
 
         # Set interface priority order
         if datatype is None:
@@ -526,6 +526,10 @@ class DFColumns(DataColumns):
             vdims = vdims if vdims else vdim_param.default
             columns = [d.name if isinstance(d, Dimension) else d
                        for d in kdims+vdims]
+
+            if isinstance(data, dict):
+                data = OrderedDict([(d.name if isinstance(d, Dimension) else d, v)
+                                    for d, v in data.items()])
             if isinstance(data, tuple):
                 data = pd.DataFrame.from_items([(c, d) for c, d in
                                                 zip(columns, data)])
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -400,11 +400,7 @@ class ColumnarData(param.Parameterized):
     def _process_data(cls, data, paramobjs, **kwargs):
         params = {}
         if isinstance(data, Element):
-            params['kdims'] = data.kdims
-            params['vdims'] = data.vdims
-            params['label'] = data.label
-            if data.group != data.params()['group'].default:
-                params['group'] = data.group
+            params = util.get_param_values(data)
 
         if isinstance(data, NdElement):
             params['kdims'] = [d for d in params['kdims'] if d != 'Index']
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -58,7 +58,6 @@ class Columns(Element):
         super(Columns, self).__init__(data, **dict(kwargs, kdims=kdims, vdims=vdims))
         self.data = self._validate_data(self.data)
 
-
     def _validate_data(self, data):
         """
         Method that is often overridden in the implementation of
@@ -277,21 +276,6 @@ class Columns(Element):
         if group_type is None:
             group_type = type(self)
         return self.interface.groupby(self, dimensions, container_type, group_type, **kwargs)
-
-
-    @classmethod
-    def collapse_data(cls, data, function=None, kdims=None, **kwargs):
-        """
-        Class method utility function to concatenate the supplied data
-        and apply a groupby operation along the supplied key dimensions
-        then aggregates across the groups with the supplied function.
-        """
-        if isinstance(data[0], NdElement):
-            return data[0].collapse_data(data, function, kdims, **kwargs)
-        elif isinstance(data[0], np.ndarray):
-            return ArrayColumns.collapse_data(data, function, kdims, **kwargs)
-        elif util.is_dataframe(data[0]):
-            return DFColumns.collapse_data(data, function, kdims, **kwargs)
 
 
     @classmethod
@@ -896,7 +880,7 @@ class ArrayColumns(DataColumns):
         kdims = [kdim for kdim in columns.kdims if kdim not in reduce_dims]
         if len(kdims):
             reindexed = columns.reindex(kdims)
-            reduced = reindexed.collapse_data([reindexed.data], function, kdims)
+            reduced = cls.collapse_data([reindexed.data], function, kdims)
         else:
             if isinstance(function, np.ufunc):
                 reduced = function.reduce(columns.data, axis=0)
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -184,17 +184,14 @@ class Columns(Element):
         and vice versa.
         """
         if kdims is None:
-            key_dims = [d for d in self.kdims
-                        if d not in vdims]
+            key_dims = [d for d in self.kdims if d not in vdims]
         else:
             key_dims = [self.get_dimension(k) for k in kdims]
 
         if vdims is None:
-            val_dims = [d for d in self.vdims
-                        if d not in kdims]
+            val_dims = [d for d in self.vdims if d not in kdims]
         else:
             val_dims = [self.get_dimension(v) for v in vdims]
-
 
         data = self.interface.reindex(self, key_dims, val_dims)
         return self.clone(data, kdims=key_dims, vdims=val_dims)
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -56,14 +56,7 @@ class Columns(Element):
                                              datatype=kwargs.get('datatype'))
         (data, kdims, vdims, self.interface) = initialized
         super(Columns, self).__init__(data, **dict(kwargs, kdims=kdims, vdims=vdims))
-        self.data = self._validate_data(self.data)
-
-    def _validate_data(self, data):
-        """
-        Method that is often overridden in the implementation of
-        specific Elements for validating and transforming the input data
-        format."""
-        return self.interface.validate_data(self, data)
+        self.interface.validate(self)
 
 
     def __setstate__(self, state):
@@ -322,12 +315,13 @@ class Columns(Element):
             return dim_vals
 
 
-    def dframe(self, as_table=False):
+    def dframe(self, dimensions=None):
         """
-        Returns the data in the form of a DataFrame, if as_table is
-        requested the data will be wrapped in a Table object.
+        Returns the data in the form of a DataFrame.
         """
-        return self.interface.dframe(self, as_table)
+        if dimensions:
+            dimensions = [self.get_dimension(d).name for d in dimensions]
+        return self.interface.dframe(self, dimensions)
 
 
 
@@ -399,15 +393,15 @@ class DataColumns(param.Parameterized):
     def shape(cls, columns):
         return columns.data.shape
 
-
     @classmethod
     def length(cls, columns):
         return len(columns.data)
 
-
     @classmethod
-    def validate_data(cls, columns, data):
-        return data
+    def validate(cls, columns):
+        pass
+
+
 
 
 
@@ -440,10 +434,6 @@ class NdColumns(DataColumns):
             raise ValueError("NdColumns interface couldn't convert data.""")
         return data, kdims, vdims
 
-
-    @classmethod
-    def validate_data(cls, columns, data):
-        return data
 
     @classmethod
     def shape(cls, columns):
@@ -535,8 +525,8 @@ class DFColumns(DataColumns):
 
 
     @classmethod
-    def _validate(cls, columns):
-        if not all(c in data.columns for c in columns.dimensions(label=True)):
+    def validate(cls, columns):
+        if not all(c in columns.data.columns for c in columns.dimensions(label=True)):
             raise ValueError("Supplied dimensions don't match columns "
                              "in the dataframe.")
 
@@ -710,6 +700,8 @@ class ArrayColumns(DataColumns):
 
         if data is None or data.ndim > 2 or data.dtype.kind in ['S', 'U', 'O']:
             raise ValueError("ArrayColumns interface could not handle input type.")
+        elif data.ndim == 1:
+            data = np.column_stack([np.arange(len(data)), data])
 
         if kdims is None:
             kdims = eltype.kdims
@@ -719,10 +711,6 @@ class ArrayColumns(DataColumns):
 
 
     @classmethod
-    def validate_data(cls, columns, data):
-        if data.ndim == 1:
-            data = np.column_stack([np.arange(len(data)), data])
-        return data
 
 
     @classmethod
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -64,6 +64,11 @@ class Columns(Element):
 
 
     def _validate_data(self, data):
+        """
+        Method that is often overridden in the implementation of
+        specific Elements for validating and transforming the input
+        data format.
+        """
         return self.interface.validate_data(self, data)
 
 
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -168,7 +168,7 @@ class Columns(Element):
             return data
         else:
             return self.clone(data)
-        
+
 
     @property
     def interface(self):
@@ -342,7 +342,7 @@ class Columns(Element):
 
     @property
     def shape(self):
-        """Returns the shape of the data."""
+        "Returns the shape of the data."
         return self.interface.shape(self)
 
 
@@ -571,7 +571,7 @@ class ColumnarDataFrame(ColumnarData):
     def concat(columns_objs):
         return pd.concat([col.data for col in columns_objs])
 
-    
+
     @staticmethod
     def groupby(columns, dimensions, container_type, group_type, **kwargs):
         index_dims = [columns.get_dimension(d) for d in dimensions]
@@ -741,7 +741,7 @@ class ColumnarArray(ColumnarData):
         idxs = [columns.get_dimension_index(dim) for dim in by]
         return data[np.lexsort(np.flipud(data[:, idxs].T))]
 
-    
+
     @staticmethod
     def values(columns, dim):
         data = columns.data
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -1,6 +1,6 @@
 """
-The data module provides utility classes to interface with various
-data backends.
+The data module provides utility classes to interface with various data
+backends.
 """
 
 import sys
@@ -578,10 +578,9 @@ class DFColumns(DataColumns):
     @classmethod
     def reduce(cls, columns, reduce_dims, function=None):
         """
-        The aggregate function accepts either a list of Dimensions
-        and a function to apply to find the aggregate across
-        those Dimensions or a list of dimension/function pairs
-        to apply one by one.
+        The aggregate function accepts either a list of Dimensions and a
+        function to apply to find the aggregate across those Dimensions
+        or a list of dimension/function pairs to apply one by one.
         """
         kdims = [kdim.name for kdim in columns.kdims if kdim not in reduce_dims]
         vdims = columns.dimensions('value', True)
@@ -776,7 +775,8 @@ class ArrayColumns(DataColumns):
 
 
     @classmethod
-    def groupby(cls, columns, dimensions, container_type=HoloMap, group_type=None, raw=False, **kwargs):
+    def groupby(cls, columns, dimensions, container_type=HoloMap,
+                group_type=None, raw=False, **kwargs):
         data = columns.data
 
         # Get dimension objects, labels, indexes and data
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -50,8 +50,12 @@ class Columns(Element):
         the data fails to be understood).""")
 
     def __init__(self, data, **kwargs):
-        data, params, self.interface = DataColumns.initialize(type(self), data, kwargs)
-        super(Columns, self).__init__(data, **params)
+        initialized = DataColumns.initialize(type(self), data,
+                                             kwargs.get('kdims'),
+                                             kwargs.get('vdims'),
+                                             datatype=kwargs.get('datatype'))
+        (data, kdims, vdims, self.interface) = initialized
+        super(Columns, self).__init__(data, **dict(kwargs, kdims=kdims, vdims=vdims))
         self.data = self._validate_data(self.data)
 
 
@@ -349,14 +353,12 @@ class DataColumns(param.Parameterized):
     interfaces = {}
 
     @classmethod
-    def initialize(cls, eltype, data, kwargs):
+    def initialize(cls, eltype, data, kdims, vdims, datatype=None):
         # Process params and dimensions
-        params = {}
-        kdims, vdims = None, None
         if isinstance(data, Element):
-            params.update(util.get_param_values(data))
-        params.update(kwargs)
-        kdims, vdims = params.get('kdims'), params.get('vdims')
+            pvals = util.get_param_values(data)
+            kdims = pvals.get('kdims', kdims)
+            vdims = pvals.get('vdims', vdims)
 
         # Process Element data
         if isinstance(data, NdElement):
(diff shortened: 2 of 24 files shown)


jbednar:
This will be amazingly useful!

If one wants a raster or Image type with different datatypes on the x and y axes, it looks like that would only be supported with Pandas or as an OrderedDict?  I.e., there's no numpy-based mechanism provided for that, either as record arrays or as a list of arrays?

philippjfr:
While Image and Raster aren't part of this particular proposal (we will extend those with their own backend utilities), I get what you mean and the answer is unfortunately no. One of the major reasons why we decided to do this in the first place is precisely this reason, it isn't possible to easily extend the numpy based implementation to deal with heterogeneous types, so it will have to fall back to either the NdElement or Pandas based implementations.

jbednar:
Hmm.  Point taken about Image and Raster not being included.  I've never used recarray personally, but from http://docs.scipy.org/doc/numpy/reference/generated/numpy.recarray.html it looks like Numpy supports mixing types across columns already, even without recarray?  I didn't realize that.  Does that mean that there shouldn't be a problem with columns of different types, already?

philippjfr:
> I've never used recarray personally, but from http://docs.scipy.org/doc/numpy/reference/generated/numpy.recarray.html it looks like Numpy supports mixing types across columns already, even without recarray. I didn't realize that. Does that mean that there shouldn't be a problem with columns of different types, already?

That's true but it also changes the shape and indexing signature of the array:

```python
homogeneous = np.array([(1.0, 2), (3.0, 4)])
heterogeneous = np.array([(1.0, 2), (3.0, 4)], dtype=[('x', float), ('y', int)])
print homogenous.shape
print heterogeneous.shape
```
(2, 2)
(2,)
```

which means it wouldn't be compatible with our existing code.

**Comment by jbednar** [comment]
Ah.  Looks like it packs it up as something analogous to an array of tuples:

```python
>>> homogeneous
array([[ 1.,  2.],
       [ 3.,  4.]])

>>> heterogeneous
array([(1.0, 2), (3.0, 4)],
      dtype=[('x', '<f8'), ('y', '<i8')])

It's a shame that it can't just keep the same interface. :-(   Would it be possible to make an object that's a view of a heterogenous Numpy array that acts like the homogenous case?  There must be a reason it wasn't done already in Numpy (and pandas.dataframe.values), so presumably not?

**Comment by jlstevens** [comment]
I've just had a discussion with Philipp and here are the notes I've made:

* I'm happy if ``Table`` no longer supports ``ItemTable`` as a return type.
* We should preserve the input data type but if it is one of our fancy formats (e.g a dictionary or some other pure Python literal) we should have a flag to determine whether ``.data`` should become an ``NdElement`` or a pandas ``DataFrame`` (if available) by default. I say we don't use pandas unless the input is pandas.
* We need to update the documentation about key and value dimensions because it is no longer about indexing. I think we said that key dimensions are sample positions or categorical values whereas the value dimensions are the values recorded at those points. The point is that the new proposal *will* support indexing and slicing of value dimensions.
* ``NdElement`` needs a first ``'Index'`` dimension handled by the interface class (that needs to be written to support it!) in order to support multiple entries with duplicate keys. This index will be managed by the interface class.
* ``Columns``  will be unsorted by default but will have a ``.sort()`` method for lexical sort.
* For compatibility with pandas and numpy, we are considering another indexing mode: indexing with a single string would return the values of the associated dimension. In effect this is an alias to ``dimension_values`` that is familiar to pandas/numpy users. By default this would probably be a numpy array but I would consider a flag to change the return type (e.g list, xray-array, dataframe, another ``Columns`` instance). NOTE: In future we should be able to return another ``Column`` instance (the sensible thing!) because numpy will behave in a more sensible way (http://docs.scipy.org/doc/numpy-dev/release.html):

> * In the future Boolean array-likes (such as lists of python bools) will always be treated as Boolean 
> indexes and Boolean scalars (including python True) will be a legal boolean index

**Comment by jbednar** [comment]
That sounds good.  I too favor keeping the original data type where possible, and also not automatically converting to Pandas unless someone has specifically enabled that.

**Comment by philippjfr** [comment]
Added Profiling notebook

**Comment by philippjfr** [comment]
This PR is now almost ready to be merged, there are a few more fixes to make but nosetests are passing, coverage for the new module is >90% and the majority of the notebooks are passing or are failing only because the data is no longer sorted.

The major thing to discuss at this point is naming, I'll list all the new modules, classes and methods and you can tell me if they make sense or if you have a better suggestion, I'll mark the ones I'm not totally happy with, with a *:

Module(s):

* ``data.py``* - The new module is currently called data.py, I'd suggest renaming it to columns.py.

Classes:

* ``Columns`` - The new baseclass for all column based Element types. This is what we decided above, I'm happy with it.
* ``ColumnarData``* - Baseclass for the interface classes. (Very open to changing the naming scheme)
* ``ColumnarNdElement``* - Interface class for the NdElement based storage.
* ``ColumnarArray``* - Interface class for numpy ndarray based storage.
* ``ColumnarDataFrame``* - Interface class for DataFrame based storage.

Methods:

* ``array`` - Converts to numpy array, supports ``as_table`` Boolean argument
* ``mapping``* - Converts to NdElement, supports ``as_table`` Boolean argument
* ``dframe`` - Converts to DataFrame, supports ``as_table`` Boolean argument
* ``concat`` - Concatenates Columns objects of any type.
* ``aggregate`` - Familiar to pandas users, aggregates over specified dimensions with supplied function.

**Comment by jbednar** [comment]
That all sounds good to me, including the names.

**Comment by philippjfr** [comment]
Okay, all the tests are now passing the way I would like, the remaining failures are either changes in sorting or in two cases minor changes in the output of the NdMapping.info method. The next step is a code review by @jlstevens, then we'll need to document the new classes and methods and update all the remaining documentation to make sure they accurately reflect the new behavior.

I've also added a deprecation warning to the DFrame conversion methods and propose they should be deprecated in 1.5.

**Comment by jlstevens** [comment]
Looks great. I'll do the code review this evening.

One quick comment though, the method names above seem ok but not all the class names convey much to me. ``Columns`` is fine but after that I would consider dropping the verbose and redundant ``Columnar`` prefix for the other classes. 

One idea would be to make it clearer what these classes really are for e.g ``ColumnInterface``, ``NdElementInterface`` (``NdInterface``?) and ``ArrayInterface``. These names aren't great but give a sense of what these classes are for and the class hierarchy (and the API) makes the columnar aspect clear. 

Maybe there is some better name than ``Interface``? Maybe ``Columns`` could be shortened to ``Cols`` and made into a suffix e.g ``Columnar``, ``NdElementCols``, ``ArrayCols`` but I'm not terribly happy with this suggestion either.

As a last suggestion, this work is about supporting multiple data formats with the same interface so how about ``Columns``, ``ColumnFormat``, ``NdElementFormat`` and ``ArrayFormat``?

These are just some suggestions and even if you dislike them, I would like to try and come up with a clearer, shorter naming scheme.

**Comment by philippjfr** [comment]
Okay, I agree with you, I was hoping to find better names for the Columnar... classes. I think making it explicit that they are interface classes is a good idea so I'd suggest:

``ColumnInterface``, ``ArrayInterface``, ``DFInterface`` and ``NdInterface``

**Comment by jbednar** [comment]
Interface is even longer than Columnar, and seems more generic -- just about everything can be considered an interface.  Columnar is wordy, but at least it conveys what we're talking about, and is a word that other people in the scientific Python community will recognize, since so many people use columnar data stores.  I'm not saying Columnar is great, just that it actually conveys something about the intended usage of this class, while Interface doesn't convey much to me at all...
Jim

**Comment by jlstevens** [comment]
> just about everything can be considered an interface

I agree! Which is why if you name something 'interface' you probably means something specific. As you say, every API every is an interface (the 'I' part) but here we mean these classes are ways of accessing the data. Is ``Accessor`` a word?

**Comment by jlstevens** [comment]
I've discussed this with Philipp again and we agree that we should pick something semantically useful (but not too unwieldy) as these classes aren't user facing and these names aren't going to be used much outside the class hierarchy itself. 

As ``Interface`` isn't particularly helpful (especially since we might want a different style of interface for handling raster style data), the latest suggestion is: ``Columns``, ``DataColumns``,  ``DFColumns``, ``NdColumns``, ``ArrayColumns``.

I think this is an improvement over the old names but not too drastic a change in terms of semantics. Any other suggestions?

**Comment by jbednar** [comment]
Sounds fine to me.

**Comment by philippjfr** [comment]
Same, will do the renaming this weekend.

**Comment by philippjfr** [comment]
I'll be adding at least one new Tutorial notebook on working with Columnar data, which will demonstrate some of the new functionality. This also required me to make various fixes to (mainly bokeh) plotting code, hence why some seemingly unrelated commits have crept in. One example is the new GridMatrix class and gridmatrix operation, which takes advantage of being able to share NdElement and pandas data sources across plots allowing linked brushing.

**Comment by philippjfr** [comment]
[Here's the updated bokeh backend notebook](http://philippjfr.com/work/work-in-progress/bokeh-backend/), now also including shared data sources and datetime axes, both made possible by this PR.

**Comment by jlstevens** [comment]
I've pushed some improvements including the name changes and I am looking at the PR now. I can't get passed the constructor unfortunately! 

Firstly,  we want a clearer split between ``_process_data`` and ``_validate_data`` as currently both are able to change the data. Ideally validation should be checks only, raising exceptions as necessary meaning that no return value is required.

Secondly, ``_process_data`` is large and monolithic and doesn't seem to be factored into the different interfaces like the rest of the code. I'm talking to Philipp right now, and we are looking into making it cleaner with something roughly as follows:


```
@classmethod
def _process_data(cls, data, interfaces = [NdColumns, ArrayColumns, DFColumns]):
     for interface in interfaces:
         try:
              return interface.process_data(data)
         except:
              pass
      raise Exception('No interface able to process input')
```

We'll see what we cna come up with!

**Comment by jlstevens** [comment]
Ok, we think we can get the improved proposal working:

1. The ``Columns`` constructor can be written to avoid param wrangling by getting ``_process_data`` to only return the necessary kdims and vdims (if available!) from the input. 

2. The suggestion above should be possible as long as we pass ``self`` (the ``Column`` instance) along to have access to the dimensions (so an ``NdElement`` as ``.data`` has matching dimensions for instance).

3. Validation should be validation only with no setting of ``.data``. We want this at a more elementary level in the class hierarchy, probably at the ``LabelledData`` level.

4. We may need a post-processing hook that has access to the element parameters (something I am currently discussing with Philipp).

**Comment by jlstevens** [comment]
I'll keep adding comments to this post as I go through the PR. I won't delete any so feel free to add your response below as I make my way through all the code:

* The ``sort`` method returns ``self.clone(sorted_columns)`` i.e a copy instead of a side-effect. In idiomatic Python, shouldn't it be called ``sorted`` instead?

**Comment by jlstevens** [comment]
One thing worth considering (and currently under discussion) is the wisdom of keeping ``NdElement`` into the future. We currently need it for backwards compatibility but we now have a chance to deprecate it and it was not really designed for all the operations we are now using it for.

We need a pure-python alternative to pandas and we have a few possibilities instead of ``NdElement``:

* The record array suggestion (point 8 above) as numpy is already core dependency.

* A very simple Python literal to represent columns such as this dictionary format: 

```python
{'column1':[1,2,3], 'column2':['foo', 'bar', 'baz']}
```

* An easily installed,  pure-python alternative to pandas, e.g [EZtable](https://bitbucket.org/salimfadhley/toytable).

I prefer options 1 or 2 though I think the simple Python literal format is nicest as it is easily understandable by anyone who knows any Python at all (no numpy knowledge needed!) and is in fact a valid constructor format for pandas DataFrames. Also, seeing the format is so simple, hopefully it wouldn't be too hard to implement the operations we need relatively efficiently (unlike ``NdElement`` which abuses ``NdMapping``).

jlstevens:
Actually, there is yet another option to replace ``NdElement`` that might be even better than the dictionary list format:

```python
{'column1':np.array([1,2,3]), 'column2':np.array(['foo', 'bar', 'bazzar'])}
```

Using 1D numpy arrays per columns is more efficient than using lists. This is still very simple and understandable (simpler than record arrays) but still allows heterogenous types and column labels. Furthermore, there is an opportunity to share code with the regular ``ArrayColumns`` interface used for larger, heterogeneous arrays.

We aren't sure if one interface class should handle both formats for supplying numpy arrays or if we should have two interface classes. I suppose this new one should be called ``DictColumns`` as the outer datastructure is a dictionary of columns (which happen to by numpy arrays)? Also, we can still support the list format mentioned previously as casting a list to a 1D numpy array is trivial.

philippjfr:
> 1) **sort return a copy instead of a side-effect.** 

True although pandas uses also uses sort, which returns a new object by default.

> 2) **In select and a few other places it uses np.isscalar(data), could check isinstance(Dimensioned)**

True will fix, there's a few places where I can make this substitution.

> 3) **Columns doesn't have it's own array method implementation**. 

There is a general one for all Element types, but maybe it should move. 

> 4) a) **reduce has reduce_map = self._reduce_map(dimensions, function, reduce_map) where self._reduce_map doesn't seem to be part of the data API. Shouldn't it be?**

No, other Element types support reductions and use _reduce_map.

> 4) b) **Reduce iterates over the reduce_map calling self.interface.reduce each time. This means the signature is a bit different calling reduce at the Column level and at the data level. Maybe the iteration aspect and _reduce_map could be folded into ColumnData?**

Sure that's possible not sure it's any cleaner though.

> 5) collapse_data is on the element (i.e Column) but is a data operation so seems like it should only exist on the interfaces themselves, not the elements. As there are non-column elements that have this method, why not merge all this functionality into a simple function that handles all these cases? This would be a step towards unifying Column elements with other types (e.g Rasters) once those elements also have interfaces...

I'm not really in favor of this approach if it is one monolithic function it will grow over time and already has to support Columns, Rasters, Images, QuadMeshes and HeatMap. The other issue is that it's not extensible, subclassing the method makes this easy for new types.

> 6) The method concat... is this the standard name? Numpy at least seems to use the full word i.e 
concatenate. Not checked pandas though...

Pandas uses concat and people will be used to working with pandas for Tabular data.

> 7) The interfaces look very nice and clean! Most of the methods are about as short as I would hope except select on ArrayColumns. I would have though some boolean indexing in a loop would do the job but I haven't thought about this carefully yet...

Thanks, I'm actually rather happy with how ArrayColumns.select turned out. It basically builds up a mask and then applies it, this is generally faster than iteratively applying boolean indexing because it doesn't create a bunch of intermediate arrays.

> 8) A comment possibly for future reference: we might want two interface classes for Numpy as we could use this system to support numpy record arrays. Personally, I would lean towards using pandas before using record arrays but it is certainly a possibility...

Record arrays are an option but they're rather awkward to use, the dictionary approach you outline below makes more sense in the near future.

> 9) _process_data isn't a very helpful method name. What we are trying to convey is that we are putting some input (in varying possible formats) into a standard format. So maybe something along the lines of _standardize, _canonicalize or normalize instead of _process? Of maybe simply _format_data?

There's now one ``ColumnData.initialize`` method, which tries each registered interface in turn based on the priority list specified by the ``Columns.datatype`` parameter. Each interface implements a ``reshape`` method which attempts to convert the data and raises an exception if it can't manage.

jlstevens:
Ok, just to say I am happy with the Data API - it is a lot cleaner and nicer than what we had before! It should also let us expand HoloViews to do new and exciting things now we have a flexible way to switch between the core data structures.

There is quite a lot of stuff in this PR that is not related to the Data API as there was work on this branch before this work began. That said, everything I see should move to master (don't see any failed experiments or misfeatures) so I am happy to merge with master and continue work there.

As far as I'm concerned, the Data API is now proven to work and is a clear improvement over what we had before. As soon as the tests pass, I am happy to hit merge and continue working on it!

jlstevens:
Just a quick note about `` __setstate__``: we should check if ``interface`` exists and only then decide on a suitable interface based on the data type (for old pickles). I can fix this after the merge...

jlstevens:
One other thing to do after the merge: we should migrate ``HeatMap`` to be a ``Columns`` object. In addition, we should make sure that ``ItemTable`` can accept a row of a ``Table`` now the return type from indexing has changed slightly. I suppose the rows of the ``ItemTable`` are simply the value dimensions of a ``Table`` with a single row?

jlstevens:
Ok, all tests are green! Time to merge...

jbednar:
Yay!  Good work, guys!

jbednar:
Is there an issue open elsewhere that lists blockers for the next release, now that this has been merged?

philippjfr:
There's this milestone https://github.com/ioam/holoviews/milestones/v1.4.0.

Out of those, the following should be a quick fix (<1 hour each):

* #34 Arrow elements should support arrows in any direction
* #197 Sum across an axis
* #259 Widget type switches when switching between backends
* #264 Allow easy definition of sanitized substitutions
* #265 Table to errorbars (I have a new suggestion here)
* #263 Style pickling is currently broken
* #280 Export archive

These are issues where we simply have to come to a decision:

* #217 QuadMesh transposed?
* #221 Styles going missing when slicing NdOverlays
* #258 Switch to %reload_ext in the tutorials docs

These are items that might require a bit more work:

* #278 DynamicMap (could consider only offering basic cache control at first)
* #275 Bokeh Callbacks
* #282 API for static HTML export API feature

Then there's getting the build process streamlined and improving the docs:

* #289 Main repository has grown far too large
* #266 Document renderers and display function
* #180 Website building process needs improvement

jbednar:
On Mon, Nov 9, 2015 at 6:24 PM, Philipp Rudiger <notifications@github.com>
wrote:

> There's this milestone https://github.com/ioam/holoviews/milestones/v1.4.0
> .
>
> Out of those, the following should be a quick fix (<1 hour each):
>
>    - #34 <https://github.com/ioam/holoviews/issues/34> Arrow elements
>    should support arrows in any direction
>    - #197 <https://github.com/ioam/holoviews/issues/197> Sum across an
>    axis
>    - #259 <https://github.com/ioam/holoviews/issues/259> Widget type
>    switches when switching between backends
>    - #264 <https://github.com/ioam/holoviews/issues/264> Allow easy
>    definition of sanitized substitutions
>    - #265 <https://github.com/ioam/holoviews/issues/265> Table to
>    errorbars (I have a new suggestion here)
>    - #263 <https://github.com/ioam/holoviews/issues/263> Style pickling
>    is currently broken
>    - #280 <https://github.com/ioam/holoviews/issues/280> Export archive
>
> Those all seem worth fixing, though the Arrows one might not be urgent.

>
>
> These are issues where we simply have to come to a decision:
>
>    - #217 <https://github.com/ioam/holoviews/issues/217> QuadMesh
>    transposed?
>    - #221 <https://github.com/ioam/holoviews/issues/221> Styles going
>    missing when slicing NdOverlays
>    - #258 <https://github.com/ioam/holoviews/issues/258> Switch to
>    %reload_ext in the tutorials docs
>
> Well, you know my opinion on QuadMesh, so you two can be the deciding
votes.  I think we could postpone 221 again, as I still don't know the
answer, myself.  I'm not sure what needs to be decided about 258; it sounds
like you know you want reload_ext, so is the only issue whether to print
the message?  Again, you know my opinion on that, so just decide as you
like.

>
>
> These are items that might require a bit more work:
>
>    - #278 <https://github.com/ioam/holoviews/pull/278> DynamicMap (could
>    consider only offering basic cache control at first)
>    - #275 <https://github.com/ioam/holoviews/pull/275> Bokeh Callbacks
>    - #282 <https://github.com/ioam/holoviews/issues/282> API for static
>    HTML export API feature
>
> The last two seem a bit dangerous to add just before a release; seems like
we could either postpone them until after the release, or else put them in
pretty much as-is but marked experimental.  The DynamicMap stuff *does*
seem very important to include in this release, to me.

>
>
> Then there's getting the build process streamlined and improving the docs:
>
>    - #289 <https://github.com/ioam/holoviews/issues/289> Main repository
>    has grown far too large
>    - #266 <https://github.com/ioam/holoviews/issues/266> Document
>    renderers and display function
>    - #180 <https://github.com/ioam/holoviews/issues/180> Website building
>    process needs improvement
>
> These last three seem safe to postpone until after a release; important
but not required for the actual code release.

Apart from DynamicMap, the rest sounds like it adds up to a solid day's
work, in which case I think you ought to do it now.   How much for
DynamicMap?

jlstevens:
DynamicMap needs to be done as another PR. The basics are already working but we need to be careful that the design allows for the sort of interaction we *really* want (callbacks feeding information such as mouse position to the DynamicMap via something we have called Streams).

I am happy to work on that aspect which shouldn't take too long. The part I would have more difficulty with is the JavaScript side - we need a flexible caching system that supports 1. no caching 2. LRU 3. caching every nth frame. This is important for reproducibility and export as the caching determines what you get when exporting a DynamicMap to a regular HoloMap. For this, Philipp would need to suggest how long this would take to implement (I think this could be easy, LRU which accepts 0 or infinite frames together with an integer modulo would probably be enough i.e two parameters could do the job).

jbednar:
Sounds like DynamicMap should be released, since it's very useful and urgently needed, but marked experimental in the release notes since we're still polishing it.  

For caching, seems like you do need some support, but that can be part of what's marked experimental, and we can tune it later.  To me it seems like we do need at least two parameters: 

1. Something that limits the total amount cached (either in terms of memory, which seems tricky to calculate, or in terms of total number of frames, which seems easier), defaulting to some large number but potentially settable to infinity if people wish.  This would be a hard limit, regardless of setting 2.
2. Something that allows people to specify that only ever nth frame is exported, regardless of how many are visible interactively.  The skipped frames would never be exported, regardless of setting 1.   

This still leaves it undefined what happens when the limit in 1 is reached -- should it start deleting LRU, or start decimating frames?  We should probably just pick one behavior and leave it unconfigurable (at least for now), because if we started to add parameters to control it people would get confused between those option-1 parameters and those for option 2 (which is independent of option 1's limits).

PR #285, opened 2015-11-02
Minor fixes

maxalbert:


philippjfr:
Thanks for those, we really need to improve our documentation. Ideally we'd also have the tests passing but there's currently an issue with an update to the freetype library changing the fonts, which messes with our visual comparison tests, I'll try to get to that tonight. Since these are just changes to the docs, I'll just go ahead and merge the PR.

maxalbert:
Thanks!

PR #286, opened 2015-11-04
put the right sticker to this Milestone

stonebig:


stonebig:
would really be nice

philippjfr:
Sorry we didn't merge this right away. The v1.4.1a release was really just a stopgap for a project we were working on and doesn't include the major features we actually want to include in the 1.4 release. However, I see no reason not to merge this for the time being, @jlstevens thoughts?

jlstevens:
I'm not sure how this will interact with the versioning system we use and definitely won't work with ``setup.py upload`` or ``setup.py sdist`` though I would hope those are only needed when actually releasing.

I suppose my question is really why this needs to be changed. What purpose does it serve? We are still on 1.3.2 until 1.4.1 is released. I realize that this version string doesn't match the latest tag but why does it need to? These are genuine questions -  if there is a practical reason why making this change is helpful to people then I would be happy to merge.

stonebig:
I generated an Holoviews wheel for WinPython after you tagged a 1.4.1a
I obtained a '1.3.2' which is wrong, as it was '1.4.1a1 or better'
==> changing the dev. version on source to '1.4.1a1' before tagging it '1.4.1a1' would have been nice.

jlstevens:
> I generated an Holoviews wheel for WinPython

Thanks! I didn't know any one else was building releases for HoloViews. In this case, the PR does make sense and I'll merge it now.

Sorry for not merging it earlier - I didn't really understand what it was supposed to achieve (I guess I should have asked).

PR #296, opened 2015-11-19
fix variable name apearing with dollar signs around it

almarklein:


jlstevens:
Ah that one slipped through (MathJax didn't work out for us on the website). Would you like me to merge right away or leave this open so you can add more fixes if you find them?

almarklein:
I think you can merge it now.

PR #307, opened 2015-11-29
Cross backend pickle fixes

jlstevens:
This PR aims to address the problems with style pickling ahead of the 1.4 release and to allow a single object to retain styling across backends.

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -804,8 +804,7 @@ class Store(object):
         Equivalent to pickle.load except that the HoloViews trees is
         restored appropriately.
         """
-        cls.load_counter_offset = (max(max(d) for d in  cls._custom_options.values())
-                                   if cls.custom_options() else 0)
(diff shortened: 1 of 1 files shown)


jlstevens:
The tests are failing because style inheritance doesn't seem to be working for one example in the Pandas Conversion tutorial. It seems to be an issue with style inheritance even though there are unit tests for this and simple tests work fine for me in the notebook. Strange...

jlstevens:
I can confirm it is the last commit that stops inheritance working in this example. However, I need that commit to get the behavior we now want...

jlstevens:
Sorry about the accidental merge! The latest version of magit has a 'push quickly' option that went and did this... :-(

jlstevens:
On the plus side, the tests all passed so maybe it was time to merge these fixes anyway. :-)

I'm now going to try and add some new unit for cross backend styling.

PR #308, opened 2015-11-30
Refactor of IPython extension and rendering

philippjfr:
This PR will refactor the OutputMagic and display hooks such that they are backend independent and move some of the display machinery onto the Renderers. This will make the IPython extension cleaner and make it much easier to add new output formats and renders, and allow for static HTML exporting.

The PR is far from ready for merging.

### DIFF
--- a/holoviews/core/__init__.py
+++ b/holoviews/core/__init__.py
@@ -12,44 +12,6 @@ from .io import FileArchive
 
 archive = FileArchive()
 
-
-def displayable(obj):
-    """
-    Predicate that returns whether the object is displayable or not
-    (i.e whether the object obeys the nesting hierarchy
-    """
-    if isinstance(obj, HoloMap):
-        return not (obj.type in [Layout, GridSpace, NdLayout])
-    if isinstance(obj, (GridSpace, Layout, NdLayout)):
-        for el in obj.values():
-            if not displayable(el):
-                return False
-        return True
-    return True
-
-
-def undisplayable_info(obj, html=False):
-    "Generate helpful message regarding an undisplayable object"
-
-    collate = '<tt>collate</tt>' if html else 'collate'
-    info = "For more information, please consult the Composing Data tutorial (http://git.io/vtIQh)"
-    if isinstance(obj, HoloMap):
-        error = "HoloMap of %s objects cannot be displayed." % obj.type.__name__
-        remedy = "Please call the %s method to generate a displayable object" % collate
-    elif isinstance(obj, Layout):
-        error = "Layout containing HoloMaps of Layout or GridSpace objects cannot be displayed."
-        remedy = "Please call the %s method on the appropriate elements." % collate
-    elif isinstance(obj, GridSpace):
-        error = "GridSpace containing HoloMaps of Layouts cannot be displayed."
-        remedy = "Please call the %s method on the appropriate elements." % collate
-
-    if not html:
-        return '\n'.join([error, remedy, info])
-    else:
-        return "<center>{msg}</center>".format(msg=('<br>'.join(
-            ['<b>%s</b>' % error, remedy, '<i>%s</i>' % info])))
-
-
 # Define default type formatters
 Dimension.type_formatters[int] = "%d"
 Dimension.type_formatters[float] = "%.5g"
--- a/holoviews/ipython/magics.py
+++ b/holoviews/ipython/magics.py
@@ -175,9 +175,6 @@ class OptionsMagic(Magics):
         return items
 
 
-class OutputWarning(param.Parameterized):pass
-outputwarning = OutputWarning(name='Warning')
-
 
 @magics_class
 class OutputMagic(OptionsMagic):
@@ -308,8 +305,7 @@ class OutputMagic(OptionsMagic):
     def _validate(cls, options, linemagic):
         "Validation of edge cases and incompatible options"
         backend = Store.current_backend
-        Store.renderers[backend].validate(options)
-        return options
+        return Store.renderers[backend].validate(options)
 
 
     @line_cell_magic
--- a/holoviews/ipython/magics.py
+++ b/holoviews/ipython/magics.py
@@ -324,6 +324,9 @@ class OutputMagic(OptionsMagic):
             self._set_render_options(new_options)
             OutputMagic.options = new_options
         except Exception as e:
+            self.set_backend(self.last_backend)
+            OutputMagic.options = restore_copy
+            self._set_render_options(restore_copy)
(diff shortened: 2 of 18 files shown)


jlstevens:
I've skimmed through the PR and it looks pretty good to me. Nice to see deletion from the IPython extension without a corresponding increase elsewhere. The approach of using renderer instance to keep track of the options (i.e as parameters) is very nice.

What in particular isn't ready yet? Is there anything you would like me to look at in more detail?

holoviews/core/__init__.py
```diff
@@ -12,44 +12,6 @@
 
 archive = FileArchive()
 
-
```
jlstevens: Moving these two functions to ``util`` does make sense.

jlstevens:
Just a few comments (I'll extend this list as I go along):

1. ``render_object`` should just be called ``render``.
2. ``collate_object`` should just be called ``collate`` (the function as opposed to the methods)

holoviews/plotting/renderer.py
```diff
@@ -257,3 +284,10 @@ def state(cls):
         """
         yield
 
+
+    @classmethod
+    def validate(cls, options):
+        """
+        Validates a dictionary of options set on the backend.
```
jlstevens: I think the docstring should be something like ``"Validate an options dictionary for the renderer"'. The reason is that I would say that 'renderer' is a holoviews term (subclass of ``Renderer``) whereas the 'backend' is the library or 3rd party tool used by the renderer. Very closely related, I agree but not quite the same thing.

jlstevens: Also, why not make ``validate`` accept ``**kwargs`` instead of a dictionary of options? Ends up being the same thing but maybe a nicer interface for validation. Just something to consider.

holoviews/plotting/renderer.py
```diff
@@ -175,6 +187,21 @@ def html(self, obj, fmt=None, css={}):
         return tag.format(src=src, mime_type=mime_type, css=css)
 
 
+    def get_widget(self, plot, widget_type):
+        isuniform = plot.uniform
```
jlstevens: Does (or should?) widgets have an 'auto' option? I think it is a little weird you can specify a ``widget_type`` just to override it. If you have 'auto', you can turn that into a concrete widget but if you specify a widget and it doesn't make sense, shouldn't you raise an exception instead of overriding it?

holoviews/plotting/bokeh/renderer.py
```diff
@@ -42,7 +51,7 @@ def __call__(self, obj, fmt=None):
                 json = plotobj.vm_serialize(changed_only=True)
                 data[plotobj.ref['id']] = {'type': plotobj.ref['type'],
                                            'data': json}
-            return serialize_json(data), {'file-ext':json, 'mime_type':MIME_TYPES[fmt]}
```
jlstevens: Was it really setting the file extension to all the actual json data before? ;-p

holoviews/plotting/bokeh/renderer.py
```diff
@@ -42,7 +51,7 @@ def __call__(self, obj, fmt=None):
...
+            return serialize_json(data), {'file-ext':' json', 'mime_type':MIME_TYPES[fmt]}
```
jlstevens: Why is there a space in the ``file-ext``? Shouldn't it be ``'json'`` not `` json``?

holoviews/plotting/bokeh/renderer.py
```diff
@@ -93,3 +102,20 @@ def get_size(self_or_cls, plot):
...
+        Validates a dictionary of options set on the backend.
+        """
+        try:
+            import bokeh
+            import bokeh.io
+        except:
+            raise ImportError("Could not import one of bokeh, pandas or scipy.")
+
+        if not cls._loaded:
+            bokeh.io.load_notebook()
```
jlstevens: I am happy for this to be here as long as this is good behavior for injecting the bokeh javascript (i.e fewer problems with deleted cells, or repeating the same javascript over and over etc).

holoviews/ipython/magics.py
```diff
@@ -200,27 +194,15 @@ def list_backends():
...
-            assert sorted(mode_formats['fig'].keys()) == sorted(mode_formats['holomap'].keys())
-            for mode in mode_formats['fig'].keys():
-                key = '%s:%s' % (renderer.backend, mode) if mode != 'default' else renderer.backend
-                formats.update({key: (mode_formats['fig'][mode],
-                                      inbuilt_formats + mode_formats['holomap'][mode])})
-        return formats
-
-
+    def list_formats(format_type):
+        backend = Store.current_backend
```
jlstevens: Shouldn't this take a ``mode`` (or maybe ``backend``) argument, allowing it to be used in ``update_options``?

holoviews/ipython/magics.py
```diff
@@ -200,27 +194,15 @@ def list_backends():
...
+        renderer = Store.renderers[backend]
+        return renderer.params(format_type).objects
 
     # Lists: strict options, Set: suggested options, Tuple: numeric bounds.
-    allowed = {'backend'     : list_backends() + ['nbagg', 'd3'],
-               'fig'         : ['svg', 'png', 'repr', 'pdf', 'html'],
-               'holomap'     : inbuilt_formats,
+    allowed = {'backend'     : list_backends(),
+               'fig'         : list_formats('fig'),
+               'holomap'     : list_formats('holomap'),
```
jlstevens: If ``list_formats`` is updated as suggested above, the default mode (on initialization of the magic) could be made explicit. e.g ``list_formats('fig', 'mpl:default')`` (or similar).

holoviews/plotting/bokeh/renderer.py
philippjfr: We can do that after it's merged.

holoviews/ipython/magics.py
philippjfr: Done.

jlstevens:
Ok, I'm happy with this refactor. It really cleans things up and removing code from the ipython extension is always a good thing.

jlstevens:
It is now merged and working. There is one more thing I would like to fix before 1.4 - currently the constructor of ``DynamicMap`` makes sure one item is in ``.data``. 

For generators, that means using ``next(..)`` and consuming one step of the generator. This should be moved out into either the display hooks or (probably even better) the renderer.

I'm happy to implement this change if Philipp tells me where he thinks this initialization should go. I can definitely remove this initialization from the ``DynamicMap`` constructor once it is done.

PR #311, opened 2015-12-01
Static HTML Exporting

philippjfr:
This implements the suggestion for static HTML export in #282. It accepts an optional template argument with three optional fields js, css and html. It's not quite ready for merge but already works.

### DIFF
--- a/holoviews/plotting/renderer.py
+++ b/holoviews/plotting/renderer.py
@@ -50,6 +50,17 @@ MIME_TYPES = {
     'json':  None
 }
 
+static_template = """
+<html>
+  <head>
+    {css}
+    {js}
+  </head>
+  <body>
+    {html}
+  </body>
+</html>
+"""
 
 class Renderer(Exporter):
     """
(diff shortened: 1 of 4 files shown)


holoviews/plotting/renderer.py
```diff
@@ -187,6 +203,39 @@ def html(self, obj, fmt=None, css=None):
...
 
+    def static_html(self, obj, fmt=None, template=None):
+        """
+        Generates a static HTML with the rendered object in
+        the supplied format. Allows supplying a template
+        formatting string with fields to interpolate 'js',
+        'css' and the main 'html'.
+        """
+        cls_type = type(self)
+        renderers = (cls_type,) + cls_type.__bases__
```
jlstevens: I'm not sure I understand the need for this approach. Why not use this sort of pattern?

```python
class BaseClass(object):
       js_dependencies = {'jquery':'url1'}

class Subclass(BaseClass):
       js_dependencies = dict(BaseClass.js_dependencies, **{'d3':'url2'})
```

Then you could use ``self.js_dependencies``, ``self.css_dependencies`` etc and you'll be able to get all the dependencies directly from the class attribute without iterating through the class hierarchy like this. Maybe I am missing something here?

jlstevens:
Looks good! Other than the commit comment above, it looks fine. The only thing I would want before merging are a couple of (simple) unit tests.

jlstevens:
Ok, now all the tests are passing (including the new ones) it is time to merge!

Issue #264, opened 2015-09-10
Allow easy definition of sanitized substitutions


jlstevens:
For the next release, we should make sure it is easy for the user to define convenient substitutions for the sanitized strings, particularly dimension names (which can get very unwieldy, especially when LateX is involved)

 One suggestion is to allow definition of an alias at the dimension level e.g by specifying the name as a tuple. Alternatively, we should extend the formatter system to handle custom string sanitation and the idea of a tuple for a name could just be a shortcut to defining the appropriate formatter.

jlstevens:
I feel I should link to Issue #251 as that issue also deals with dimension formatters. I say we need to think careful about improving dimension formatters in general.

philippjfr:
They're different formatters though, one is for the values the other is for the name. Apart from making sure the naming makes it clear which is which there isn't much of a link between the two.

jlstevens:
The link is that we want a consistent system for defining multiple types of formatters:

* It should be easy to define these formatters.
* Naming needs to be consistent.
* We need to worry about pickling issues, especially when functions (especially lambdas) are involves.

 For this reason, I think issue #251 is related. There is also the issue of formatting units.

philippjfr:
Can we come to a decision here? Currently we have two formatter parameters on Dimensions:

* formatter - Defines the formatting of the dimension values and ticks
* format_string - Defines the display of the quantity along with the dimension, e.g. ``{name}: {val} {unit}`` becomes ``Weight: 10 kg``

Adding one additional formatter for the name itself along with some easy way to define this as an alias seems like the straightforward and sensible option here. By default this formatter would probably just be ``{name}`` so it mirrors the defined dimension name but can be overridden with any valid formatting string. I would suggest calling the parameter ``label_formatter`` and then allowing both the name and the alias/formatter to be defined as a tuple in the Dimension constructor, e.g. ``Dimension(('phi', '$\{name}$'))`` would define the name phi for easy access and the ``label_formatter``, which would render the label nicely in LaTeX. 

I also don't think we need to concern ourselves with dimension formatting any further for now. I know @jlstevens at least is concerned about proliferation of parameters on Dimensions but I think at least ``formatter`` and the new ``label_formatter`` are very well defined and clear, if we had those two I'd be happy to rename/rethink any other parameters controlling the labeling, such as the aforementioned ``format_string``.

jbednar:
[...]

PR #312, opened 2015-12-01
Easy definition of sanitized substitutions

jlstevens:
This PR aims to address issue #264.

### DIFF
--- a/doc/Tutorials/Exploring_Data.ipynb
+++ b/doc/Tutorials/Exploring_Data.ipynb
@@ -132,7 +132,7 @@
    },
    "outputs": [],
    "source": [
-    "date_dim = hv.Dimension(\"Date\", formatter=md.DateFormatter('%h %d %Y %H:%M UTC'), type=float)\n",
+    "date_dim = hv.Dimension(\"Date\", value_format=md.DateFormatter('%h %d %Y %H:%M UTC'), type=float)\n",
     "kdims = ['Frame', date_dim]"
    ]
   },
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -103,7 +103,14 @@ class Dimension(param.Parameterized):
             existing_params = dict(name.get_param_values())
         else:
             existing_params = {'name': name}
-        super(Dimension, self).__init__(**dict(existing_params, **params))
+
+        all_params = dict(existing_params, **params)
+        if isinstance(all_params['name'], tuple):
+            alias, long_name = all_params['name']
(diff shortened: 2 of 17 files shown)


jlstevens:
I think this PR is ready to merge. Here is an example of what it allows you to do:

```python
layout = (hv.Image(np.random.rand(10,10), 
         group=('Spectrum', 'Frequency spectrum'), 
         label=('Glucose', '$C_6H_{12}O_6$'))
+ hv.Image(np.random.rand(10,10), 
         group=('Spectrum', 'Frequency spectrum'), 
         label=('Water', '$H_2O$')))
```

Now you can select the two parts of the layout conveniently with ``layout.Spectrum.Glucose`` or ``layout.Spectrum.Water`` while showing the chemical compound names in the title (before you had to use some awkward, auto-generated sanitized string).

To make this less awkward, you can do this using the new ``Aliases`` helper function:

```python
al = hv.util.Aliases(Spectrum='Frequency spectrum', 
                     Water='$H_2O$', 
                     Glucose='$C_6H_{12}O_6$')

(hv.Image(np.random.rand(10,10), group=al.Spectrum,  label=al.Glucose)
+ hv.Image(np.random.rand(10,10), group=al.Spectrum, label=al.Water))
```

Lastly, the same principles apply setting dimension names using tuples:

```python
im =hv.Image(np.random.rand(10,10), 
             kdims=[('Lambda', '$\Lambda$'), ('Joules', 'Energy ($J$)')])
```

Which now makes methods such as select easier:

```python
im.select(Lambda=(-0.2, 0.2), Joules=(-0.3, 0.3))
```

Of course you can use the ``Aliases`` helper here too! 

After I add some tests, I *think* this is ready to merge and I definitely think it will help scientists (and anyone using LaTeX and special symbols really!) use HoloViews productively.

jlstevens:
I've now added unit tests and one they pass, I think it is ready to merge (unless you have any comments!).

It is worth mentioning one limitation of this new system, the group/label/dimension name stays a string (the long string, not the short alias) which means your alias definitions won't be pickled together with the object. 

Aliases are defined as dictionaries (the ``aliases`` parameter) on the three global functions: ``hv.util.label_sanitizer``, ``hv.util.group_sanitizer``, ``hv.util.dimension_sanitizer``. Restoring your aliases comes down to restoring these dictionaries on these three objects appropriately.

philippjfr:
Looks good, very happy to finally have a nice way to specify aliases. Since the tests are passing and I have no specific comments I'll go ahead and merge.

jbednar:
Excellent!  Looks great.

PR #314, opened 2015-12-02
NdMapping.__setitem__ now replaces rather than updates

philippjfr:
As the title says this commit makes sure that when assigning to an NdMapping using ``__setitem__`` it should replace rather than update the existing entry. It may be he case that some machinery relies on the old behavior so this isn't ready for merge until some testing has been done.

### DIFF
--- a/holoviews/core/ndmapping.py
+++ b/holoviews/core/ndmapping.py
@@ -145,7 +145,7 @@ class MultiDimensionalMapping(Dimensioned):
             raise KeyError('Key has to match number of dimensions.')
 
 
-    def _add_item(self, dim_vals, data, sort=True):
+    def _add_item(self, dim_vals, data, sort=True, update=True):
         """
         Adds item to the data, applying dimension types and ensuring
         key conforms to Dimension type and values.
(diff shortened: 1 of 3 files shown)


holoviews/core/ndmapping.py
```diff
@@ -145,7 +145,7 @@ def _item_check(self, dim_vals, data):
             raise KeyError('Key has to match number of dimensions.')
 
 
-    def _add_item(self, dim_vals, data, sort=True):
+    def _add_item(self, dim_vals, data, sort=True, update=True):
         """
         Adds item to the data, applying dimension types and ensuring
         key conforms to Dimension type and values.
```
jlstevens: Would be nice to know what the ``update`` argument does in the docstring.

holoviews/core/ndmapping.py
```diff
@@ -145,7 +145,7 @@ def _item_check(self, dim_vals, data):
             raise KeyError('Key has to match number of dimensions.')
 
 
-    def _add_item(self, dim_vals, data, sort=True):
+    def _add_item(self, dim_vals, data, sort=True, update=True):
```
jlstevens: Would be nice to know what the ``update`` argument does in the docstring.

jlstevens:
Other than describing the ``update`` argument in the docstring, it looks good to me (mainly because the tests are passing!). There isn't that much changed and I don't see anything that looks like it could go *too* wrong.

PR #315, opened 2015-12-02
Added load_notebook_fn to replace %load_ext

philippjfr:
Added a ParameterizedFunction to load the notebook resources and initialize the magics. Unlike %load_ext this version accepts arguments to include bokeh, suppress the logo and change the CSS settings such as the notebook width.


holoviews/plotting/bokeh/renderer.py
```diff
@@ -108,20 +108,3 @@ def get_size(self_or_cls, plot):
...
-    def validate(cls, options):
-        """
-        Validates a dictionary of options set on the backend.
-        """
-        try:
-            import bokeh
-            import bokeh.io
-        except:
-            raise ImportError("Could not import one of bokeh, pandas or scipy.")
-
```
jlstevens: Surely it is only the ``load_notebook`` bit that needs to go?

philippjfr: The baseclass implements the basics of validate, and if you can't import bokeh you wouldn't have gotten as far as calling BokehRenderer.validate anyway.

holoviews/ipython/__init__.py
```diff
@@ -85,18 +87,53 @@ def load_notebook():
...
-    global _loaded
-    if not _loaded:
-        _loaded = True
-        param_ext.load_ipython_extension(ip, verbose=False)
-        load_magics(ip)
-        OutputMagic.initialize()
-        set_display_hooks(ip)
-        load_notebook()
+    bokeh = param.Boolean(default=False, doc="""Toggle inclusion of BokehJS,
+        required to render plots using the Bokeh backend.""")
```
jlstevens: How about a ``js`` parameter as an ``ObjectSelector`` to state which javascript things are loaded? Default would be 'holoviews' but the allowed options would be ``['holoviews', 'bokeh']``.

jbednar:
Looks great to me!  Seems much more friendly than the previous approach.

holoviews/plotting/bokeh/renderer.py
jlstevens: Ok, that makes sense.

jlstevens:
I'm also very happy with this. 

My only concerns are is that we have no documentation/examples for the users. Fortunately, as the tests are passing and in this PR ``%load_ext`` is using ``load_notebook``, then the new behavior must match the old behavior meaning we could replace ``%load_ext`` with ``load_notebook()`` everywhere if we wanted to.

That said do we want to keep the old behavior (i.e matching what ``%load_ext`` used to do) shown in our notebooks?

If so, I think ``%%load_ext`` is still useful and valid and we can just mention the new ``hv.load_notebook`` in the text somewhere (probably across multiple tutorials).

If not, we should replace ``%load_notebook`` with ``load_notebook()`` where we feel the default behavior is improved and update all the notebooks to use it (which unfortunately might mean having to update all the tests).

Just something to think about...

philippjfr:
I think we should consistently use hv.load_notebook throughout the Tutorials. I'd keep the old %load_ext for compatibility for now and consider deprecating it eventually, since I don't see any benefit to it.

philippjfr:
As soon as this gets merged anyone using the bokeh backend will have to execute hv.load_notebook('bokeh') first, so we'll have to announce this somehow.

jbednar:
I vote for keeping %load_ext around for compatibility, but purging it from all examples and tutorials, pretending it doesn't exist henceforth.  

Officially, there's been no release with Bokeh support yet, so I think it's ok if we simply say in the release notes for the new version (since that is imminent) that there is now Bokeh support, and that it can be enabled with ``hv.load_notebook('bokeh')`` followed by the appropriate backend selection.  

I'm not completely happy with the name ``load_notebook`` -- to me that sounds like a mechanism for loading a notebook, which is not what this does.  Maybe ``notebook_support``?  ``notebook_extension``?  ``notebook_features``?

jlstevens:
Philipp and I discussed the name and we agree it isn't quite right but chose it to be consistent with what bokeh calls it. My suggestion for the correct name was ``initialize_notebook`` but ``notebook_extension`` is as good if not better.

jbednar:
Why is it important what Bokeh calls it?  HoloViews is not tied to Bokeh, and there's no need to propagate a confusing name.  notebook_extensions seems very clear to me.

jlstevens:
> Why is it important what Bokeh calls it?

I guess every project will have their own name for the thing that injects initial content into the notebook. :-)

I think ``notebook_extension`` is clear and I am also in favor of this name change.

jlstevens:
If we are going to change all the tutorials to use ``notebook_extension``  it would probably be worth making the change mentioned in #316 as well (standardizing on ``hv``).

jlstevens:
Ok, tests are passing and I think it is ready to merge. Any last things you wish to do @philippjfr?

Issue #208, opened 2015-07-07
Slicing and ellipsis


vascotenner:
In numpy one can do ```np.array[...,2]```

Can this functionality be implemented in hv?

jlstevens:
I've discussed this suggestion with Philipp and I don't see why this couldn't be supported. It might even be quite easy to implement!

I'll have a look into it and if it is simple enough to do, I'll implement it for the upcoming 1.4 release. If not, I think this would be something to add shortly after...

jlstevens:
I'll tentatively assign this to 1.4 and assign it to myself to make sure I have a look at it before the release. If it is tricky, I'll move it to the next milestone...

jlstevens:
I've made a new PR #317 that I hope can address this and a few other issues...

jlstevens:
The PR has now been merged so I think I can now close this issue. 

@vascotenner If you get a chance to try it out and have any feedback, I would love to hear it!

vascotenner:
This works nice now!

PR #317, opened 2015-12-02
Ellipsis and __getitem__ consistency

jlstevens:
This PR aims to address issue #208 which in turn requires our ``Element`` class (mostly ``Raster`` elements need fixing) to support consistent ``__getitem__`` semantics as ``Column``.

### DIFF
--- a/holoviews/element/raster.py
+++ b/holoviews/element/raster.py
@@ -537,6 +537,13 @@ class Image(SheetCoordinateSystem, Raster):
         if coords is () or coords == slice(None, None):
             return self
 
+        if len(coords) > (2 + self.depth):
+            raise Exception("Can only slice %d dimensions" % 2 + self.depth)
+        elif len(coords) == 3 and coords[-1] not in [self.vdims[0].name, slice(None)]:
+            raise Exception("Image only has single selectable channel %r" %
+                            self.vdims[0].name)
+
+        coords = coords[:2]
         if not any([isinstance(el, slice) for el in coords]):
             return self.data[self.sheet2matrixidx(*coords)]
         if all([isinstance(c, slice) for c in coords]):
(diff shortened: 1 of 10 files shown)


jlstevens:
This PR does seem to be working and the tests are passing. I'm not convinced the ellipses will work for *every* type of element (in particular ``Path`` type elements that need a refactor). That said, in the [Composing Data](http://holoviews.org/Tutorials/Composing_Data.html) tutorial:

```python
layout.Parameters.Sines[...,-2:2]
```


Which changes the extents at the level of the deepest element. This can be seen by comparing it to:

```python
layout.Parameters.Sines[...,-1:1]
```

jlstevens:
Ok, I've now added unit tests are passing, I think this PR is ready for a review/merge.

holoviews/element/raster.py
```diff
@@ -723,19 +746,27 @@ def __getitem__(self, coords):
...
+            channels = [el for el in values if isinstance(el, (str, Dimension))]
+            if len(channels) == 1:
+                sliced = super(RGB, self).__getitem__(coords[:self.ndims])
+                if channels[0] not in self.vdims:
+                    raise KeyError("%r is not an available value dimension"
+                                    % channels[0])
+                vidx = self.get_dimension_index(channels[0])
+                val_index = vidx - self.ndims
+                data = sliced.data[:,:, val_index]
+                return Image(data, **dict(self.get_param_values(onlychanged=True),
```
philippjfr: Not your fault as this was already wrong but this isn't quite right, could use core.util.get_param_values instead of self.get_param_values here.

philippjfr:
Apart from one minor comment this looks pretty good, glad the methods are now consistent and it's nice to Ellipsis support as well.

philippjfr:
Okay all passing now, merging.

PR #318, opened 2015-12-02
Add ArchiveComparison, allowing two archives to be compared.

ceball:
I'm opening this PR to ask for general feedback (mainly from @jlstevens) on archive comparisons: does the approach make sense?

You will see some CEBALERTs in the PR, just to note minor things I haven't gotten around to sorting out yet. If possible, please ignore those for the general feedback.
## ArchiveReader
- I've abstracted some operations on zip, tar, directory with 'ArchiveReaders'. How do you feel about that? I suppose there could be ArchiveWriters as well, but it depends what you feel about the idea of the Readers.
- ArchiveReaders are not explicitly tested (although they get tested by testarchivecomparison.py)
## ArchiveComparison
- I don't really like the `load_fns`
- I don't really like `ComparisonResult`. If it's staying, probably needs to be a class and have a repr etc. (They are handy for testing notebooks cell by cell.)
- As you know, a generator being returned by comparisons() seemed like a good idea immediately when I started. Not sure how I feel about that now, though.
## Also
- I feel some general unease about what will happen with large archives, what will happen with archives that aren't on the local file system, etc.

Anyway, just so you can see what's coming, and avert a disaster before it's too late (well, it's already too late, because I re-wrote the notebook testing with it ;) )

### DIFF
--- a/holoviews/core/io.py
+++ b/holoviews/core/io.py
@@ -16,6 +16,7 @@ from __future__ import absolute_import
 
 import re, os, time, string, zipfile, tarfile, shutil, itertools, pickle
 from collections import defaultdict
+from cStringIO import StringIO # CEBALERT
 
 from io import BytesIO
 from hashlib import sha256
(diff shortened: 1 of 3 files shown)


holoviews/core/io.py
```diff
@@ -840,3 +841,101 @@ def contents(self, maxlen=70):
...
+class ArchiveReader(object):
+    """
+    ArchiveReaders abstract operations over different archive formats,
+    e.g. directory vs. zip file.
+    """
+    def __init__(self,archive):
+        self.archive_location = archive
+        self.root,self.archive_name=os.path.split(self.archive_location)
+
+    def open_member(self,name):
```
jlstevens: Why not call this ``open_file``? As far as I can tell, all the ``ArchiveReaders`` contain files in some form or another and all these methods are concerned with files (even if they themselves are in a compressed archive). If this is true, I feel the generic word ``member`` obscures what is happening.

holoviews/core/io.py
```diff
@@ -840,3 +841,101 @@ def contents(self, maxlen=70):
...
+    def list_members(self):
+        return os.listdir(self.archive)
+
+    def open_member(self,name):
+        return open(self._fullname(name),'r')
+
+    def member_exists(self,name):
+        return os.path.exists(self._fullname(name))
+
+    def _fullname(self,name):
```
jlstevens: I would consider having a ``path`` attribute that points to the directory/archive file. Then you would have:

```python
def open_member(self,name):
        return open(os.path.join(self.path, name),'r')
```

holoviews/element/comparison.py
```diff
@@ -648,3 +656,140 @@ def __init__(self, *args, **kwargs):
...
+                    yield self._result(rmember,False,traceback.format_exc())
+                else:
+                    yield self._result(rmember,True,None)
+                
+
+
+    def _read_archive(self,archive_location):
+        if not os.path.exists(archive_location):
+            raise IOError("Cannot find archive %s"%archive_location)
+
```
jlstevens: I would stick the expected file extension (or more likely a list of file extensions) on the ``ArchiveReaders``. The I would just have a class attribute:

```python
class ArchiveComparison(param.Parameterized):
    readers = [Directory, ZipFile, TarFile]
```

Then ``_read_archive`` can iterate through to find an ``ArchiveReader`` it can use. As ``Directory`` isn't associated with a file extension (and you probably need to check a list of possible file extensions) I would probably check if the ``ArchiveReader`` can be used with an ``applicable`` method (or something like that returning a boolean).

holoviews/element/comparison.py
```diff
@@ -648,3 +656,140 @@ def __init__(self, *args, **kwargs):
...
             self.addTypeEqualityFunc(k, v)
+
+
+
+
+############################################################
+# Archive comparison
+
+ComparisonResult = namedtuple('ComparisonResult', ['ref_archive', 'test_archive', 'ref_member', 'match', 'msg'])
+
```
jlstevens: Just a minor comment, I would use the names: `` ['reference', 'test', 'file', 'match', 'msg']``

holoviews/element/comparison.py
```diff
@@ -648,3 +656,140 @@ def __init__(self, *args, **kwargs):
...
+        for rmember in matches:
+            if matches[rmember] is None:
+                yield self._result(rmember,False,'Test is missing member')
+            else:
+                load_fn = self.load_fns[os.path.splitext(rmember)[1].lower()]
+                r = load_fn(self.ref.open_member(rmember))
+                try:
+                    t = load_fn(self.test.open_member(matches[rmember]))
+                except:
+                    yield self._result(rmember,False,'Cannot open test and ref members with same function (%s): %s'%(lf,traceback.format_exc()))
```
jlstevens: Sounds like we want instances of some sort of parameterized function/class in order to avoid this sort of clash.

jlstevens:
This is still a work-in-progress but I do like the overall design!  Definitely the right idea in general...

The idea of ``ArchiveReaders`` seems quite sensible to me. I also agree that if we had a nice repr for ``ComparisonResult``, it would be worth having a class instead of a named tuple.

I don't like the ``load_fns`` much either as they currently stand (or the various lambda functions). Lastly, I am not decided on whether a  generator is the best interface. That said, I would leave it a generator until we have a better feeling for how it needs to be used.

jlstevens:
The PR referenced above raises a major limitation of the current testing that we should think about when implementing the new system.

philippjfr:
@jlstevens Any idea how far away from being useable this is?

jlstevens:
> @jlstevens Any idea how far away from being useable this is?

Sadly no. I remember thinking the idea was basically ok but I would probably need to adopt this PR from ceball to get it merged.

philippjfr:
@ceball Can we close this since you're now closing in on a better solution?

jlstevens:
I agree, I think we should close this PR now we have planned a better approach. I wouldn't delete the branch just yet though - on the slim chance that I see a way the archive comparison idea fits in somehow.

ceball:
Happy to close the PR (sorry...), but yes, let's not delete the branch unless archiving in general gets abandoned (I assume it won't).

Issue #322, opened 2015-12-02
Bug in Table indexing


jlstevens:
From the [Elements tutorial](http://holoviews.org/Tutorials/Elements.html#Table):

```python
from holoviews import Table
from collections import OrderedDict
keys =   [('M',10), ('M',16), ('F',12)]
values = [(15, 0.8), (18, 0.6), (10, 0.8)]
table =Table(zip(keys,values), 
          kdims = ['Gender', 'Age'], 
          vdims=['Weight', 'Height'])
table
```

When I index it as follows, I get:

```python
table[:,:, 'Height']
array([ 0.8,  0.6,  0.8])
```

Here I would expect a single column ``Table`` and not an array.

jlstevens:
The PR above has now been merged so I can now close this issue.

PR #323, opened 2015-12-02
Fixes to Columns indexing

philippjfr:
There were two bugs in the ``__getitem__`` method:

1) Indexing a value dimension by name would return that column as an array.

2) Boolean slicing was not working across backends.

I've fixed these issues and added some unit tests.


jlstevens:
Ok, good to make a PR to make sure the tests get run. If I understand it correctly, this PR will address issue #322.

jlstevens:
Ok, the tests are passing and this is definitely the behavior we wanted. I'll merge it now.

Issue #267, opened 2015-09-11
Standalone HTML export for HoloMaps


jlstevens:
The notebook archive facility currently doesn't export HoloMaps by default. Currently, you can output video (or GIF) if you want but generally these formats aren't particularly useful. In contrast, a standalone HTML file that includes interactive widgets could be useful and would work across both the matplotlib and the bokeh backends.

Another advantage of this adding this support (especially if tested!) is that it would help ensure HoloViews is tied to the HTML standard as opposed to anything specific to the notebook format. It might also open the possibility of generating documents which aren't notebooks (e.g. report documents containing interactive widgets).

As a side-issue, this might make us think of how we want to store data for such documents (as opposed to base64 encoding everything in the HTML file). This is a more open problem and something I know the IPython people are also thinking about.

philippjfr:
This has been on my radar for a while, there's nothing about our figure, video and widget rendering that's tied to the notebook and it'll be very useful to have something like this, for example I can imagine HoloViews running on a live server returning HTML to be injected into a website. As a quick demo we could have a simple web app that allows you to upload hvz files and returns the rendered HTML output.

In terms of storing the embedded data in a separate file I'd suggest sticking with a simple json file format, it's the format that the dynamic loading widgets already support and most databases can load in json files nowadays. The one issue I've had loading from json files is that only some browsers can open json files without a server (Chrome doesn't let you). One workaround is to define the json as an object in a js file, for some reason browsers trust js but not json. This is a bit of a hack but is the only way I've found to get around having a separate server to serve the json data.

jlstevens:
JSON is definitely the format that makes the most sense.

> One workaround is to define the json as an object in a js file, for some reason browsers trust js but not json.

Sounds reasonable to me.

That said I'm not sure whether this is really the 'right thing' to do. On one hand, this is circumventing the browser's security model and on the other, people are perfectly able to  inject malicious javascript  without needing our help!

As such, I don't think this is a particularly terrible workaround. All I would ask is that the ``.js`` file is just a JSON file in disguise with no extra JavaScript (unless absolutely necessary). If the file cannot be directly be parsed as JSON, I would want the minimal wrapping necessary to make it work.

philippjfr:
The PR #324 implementing these suggestions has now been merged. Closing.

PR #324, opened 2015-12-03
Allow exporting static widgets with the Renderer

philippjfr:
This turned out to be pretty straightforward and addresses the suggestion in #267. The widgets now have an option to export the data to a json file. This allows it to be displayed in the notebook without being embedded in the actual html and will also work when serving the notebook on a website. So you could enable this option, set the json files to be exported to a specific folder, run through the notebook and the convert it and ship the folder of json files along with the notebook. Cross-origin policies will prevent you from viewing the resulting HTML directly in chrome but it's a nice solution for hosting plots.

Here's an example using each of our exportable backends:

```python
import numpy as np
test = hv.HoloMap({i: hv.Image(np.random.rand(5,5)) for i in range(3)})

from holoviews.plotting.bokeh.widgets import BokehSelectionWidget
from holoviews.plotting.mpl.widgets import MPLSelectionWidget

BokehSelectionWidget.export_json = True
MPLSelectionWidget.export_json = True

%output backend='bokeh'
renderer = hv.Store.renderers['bokeh']
with open('Static_JSON_bokeh.html', 'w') as f:
    f.write(renderer.static_html(test))
```
[Output](http://philippjfr.com/ipynbs/Static_JSON.html)

```python
%output backend='matplotlib'
renderer = hv.Store.renderers['matplotlib']
with open('Static_JSON_mpl.html', 'w') as f:
    f.write(renderer.static_html(test))
```
[Output](http://philippjfr.com/ipynbs/Static_JSON_mpl.html)

```python
%output backend='matplotlib:mpld3'
renderer = hv.Store.renderers['matplotlib']
with open('Static_JSON_mpld3.html', 'w') as f:
    f.write(renderer.static_html(test))
```
[Output](http://philippjfr.com/ipynbs/Static_JSON_mpld3.html)

Still need to update the widget test SHAs but otherwise working.


jlstevens:
Very nice! There are very few changes to the Python code and the javascript changes look clean (e.g getting rid of the hidden ``server`` option).

Other than updating the SHAs, I have few comments:

1. May I see an example of the scrubber widget working this way? I assume you just set the format of ``holomap`` on the renderer.

2. As far as I can tell the json files will have unintelligible names based on ``uuid``. This isn't critical right now but it might mean it becomes hard to know which html file links to which json file. As a later feature, I can imagine zipping a single html file with the single json file it needs.

3. Instead of having to expose the widget settings all the way up to the renderer or output magic, something like this would be nice:

```python
renderer = hv.Store.renderers['matplotlib']
renderer.export_widget(test, widget_path, json_path='./json_figures', CDN =my_CDNs)
```

The signature is ``export_widget(object, html_path, **params)`` where the params are the widget params (we would now be exposing these to advanced users). This would check the renderer's ``holomap`` format is one of the widgets, otherwise it should complain.

Currently, I don't think you have access to the widget class at this level as you may have to briefly change the class parameters (as you show above) appropriately within this function. An extra ``widget_type`` parameter would be good and if it is an easy refactor the code to avoid fiddling the class parameters this way, then I would like to see that too.

This is a really important feature as it allows HoloViews to break out the notebook in a useful way. For this reason I would like it merged into 1.4 if you can address my comments above.

philippjfr:
Okay I'll address these issues then we can merge it.

> May I see an example of the scrubber widget working this way? I assume you just set the format of holomap on the renderer.

[Here](http://philippjfr.com/ipynbs/Static_JSON_bokeh_scrubber.html)

>As far as I can tell the json files will have unintelligible names based on uuid. This isn't critical right now but it might mean it becomes hard to know which html file links to which json file. As a later feature, I can imagine zipping a single html file with the single json file it needs.

Definitely, eventually this should be tied in with exporting facilities.

> Instead of having to expose the widget settings all the way up to the renderer or output magic, something like this would be nice:

```
renderer = hv.Store.renderers['matplotlib']
renderer.export_widget(test, widget_path, json_path='./json_figures', CDN =my_CDNs)
```

Can this just be in the static_html signature? While that method is general and applies to things other than widgets, you might as well just use ``.html`` well exporting videos and individual images.

jlstevens:
Just a note of something that I think fits into the PR.

In the Options tutorial, this isn't working properly:

```python
from holoviews.ipython.parser import OptsSpec
renderer.save(image + waves, 'example_V', 
              options=OptsSpec.parse("Image (cmap='gray')"))
```

The reason it that the object is a holomap and it is trying to save out a widget. In Python 3 on Travis you see an exception as it doesn't like writing the widget code to a file in mode ``'w'``.

We have two options:

1. Disallow widgets formats from ``renderer.save``.
2. Export a widget using sensible default settings. For control over the widget details, use  ``export_widget`` instead.

jlstevens:
I'm very happy with this! I've made a few comments above, but once they are addressed, I am happy to merge. The only suggestion I have is to have just a single SHA test of ``export_widget`` to make sure it keeps working in future.

jlstevens:
Ok, my comments have been addressed and the tests are passing. Merging.

jbednar:
The above all sounds great, but can you clarify how it interacts with our general exporting support?  Your comment above

> Definitely, eventually this should be tied in with exporting facilities.

seems to say that it's not currently interacting with exporting at all, but there seems like such a strong overlap between this support and what we do when exporting in general that it would be strange for them to be separate mechanisms.

philippjfr:
It's quite possible that it is working now since Renderer.save now supports it, should confirm it though.

Issue #321, opened 2015-12-02
DynamicMap consumes generators on initialization


jlstevens:
If you create a generator with ten steps (say) you'll find you can only call ``next()`` on it nine times. This is because one call is consumed in the constructor (to help initialize the plotting system).

This initial call is still needed but it should only happen on display (i.e in the ``Renderer``) and not in the constructor of ``DynamicMap`` itself.

jlstevens:
The PR mentioned above has been merged, addressing this issue.

PR #325, opened 2015-12-03
Initialization of DynamicMaps

jlstevens:
This PR aims to address issue #321.

It seems to be working but there is a bug whereby the appropriate widget is not being selected automatically, preventing display.


philippjfr:
Looks good, we'll make sure it displays correctly as part of #324. I'll merge.

PR #329, opened 2015-12-04
Implemented DynamicMap sampled mode

philippjfr:
DynamicMaps now have a sampled mode allowing defining them in closed mode without a fixed sampling. The user may sample from them explicitly or they may be displayed in a composite object containing HoloMaps that define a fixed number of samples.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -381,6 +381,12 @@ class DynamicMap(HoloMap):
        element will be cached and therefore accessible when casting to a
        HoloMap.  Applicable in open mode only.""")
 
+    sampled = param.Boolean(default=False, doc="""
+       Allows defining a DynamicMap in closed mode without defining the
+       dimension bounds or values. Useful for allowing to let a HoloMap
+       in a composite plot to define the dimension sampling.
+       """)
+
     def __init__(self, callback, initial_items=None, **params):
         super(DynamicMap, self).__init__(initial_items, callback=callback, **params)
         self.counter = 0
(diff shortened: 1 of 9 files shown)


holoviews/core/spaces.py
```diff
@@ -381,6 +381,12 @@ class DynamicMap(HoloMap):
        element will be cached and therefore accessible when casting to a
        HoloMap.  Applicable in open mode only.""")
 
+    sampled = param.Boolean(default=False, doc="""
+       Allows defining a DynamicMap in closed mode without defining the
+       dimension bounds or values. Useful for allowing to let a HoloMap
+       in a composite plot to define the dimension sampling.
```
jlstevens: stray 'to let' in this docstring.

holoviews/plotting/plot.py
```diff
@@ -426,7 +426,7 @@ def __init__(self, element, keys=None, ranges=None, dimensions=None,
         keys = keys if keys else list(self.hmap.data.keys())
         plot_opts = self.lookup_options(self.hmap.last, 'plot').options
 
-        dynamic = element.mode if isinstance(element, DynamicMap) else False
+        dynamic = False if not isinstance(element, DynamicMap) or element.sampled else element.mode
```
jlstevens: I think dynamic should be ``False`` or a string. i.e if sampled I think ``dynamic='sampled'``.

holoviews/plotting/renderer.py
```diff
@@ -147,6 +147,8 @@ def get_plot(self_or_cls, obj):
         # Initialize DynamicMaps with first data item
         dmaps = obj.traverse(lambda x: x, specs=[DynamicMap])
         for dmap in dmaps:
+            if dmap.sampled:
+                continue
```
jlstevens: Maybe add a comment to explain it is initialized in the plotting code itself (because that is where you know the sample keys).

jlstevens:
All looks good! Keeps ``DynamicMap`` simple in many cases but leaves flexibility when ``sampled=True``.

philippjfr:
Great, now that Travis has passed, I'll merge. Fingers crossed this is the last major thing before release.

PR #339, opened 2015-12-07
Bokeh charts

philippjfr:
This PR adds plotting classes for two Bokeh chart types, Bars, which already exist in matplotlib and box plots, which has also been added for matplotlib. Currently the Element name is BoxWhisker but I'd appreciate something better that's not BoxPlot (since that kind of breaks our naming convention).  The PR is not quite ready but it already mostly works.

Here are some demos of the new supported plots: [Bokeh Charts Notebook](http://philippjfr.com/work/work-in-progress/bokeh-charts/)


jlstevens:
Great! Target this for 1.4.1?

jbednar:
Looks good.  ``Box`` would follow the convention, but we've already used that for another Element.  BoxWhisker may be the best thing left; BoxAndWhiskers is too long, and BaW is too obscure.

philippjfr:
Now ready for merge.

jlstevens:
Ok, tests are passing now so I'll merge.

Issue #347, opened 2015-12-11
Data API equivalent for HoloMaps


jlstevens:
We've had the recent Data API PR merged for ``Chart`` elements and this issue suggests a similar thing could be implemented for ``HoloMaps`` and ``NdMappings``.

The API is not quite the same as for charts as ``HoloMap`` is designed to be dictionary like. However, instead of always using ordered dictionaries as ``.data``, you can imagine using pandas DataFrames instead. This could greatly improve the speed of certain operations (such as ``groupby``) involving holomaps.

philippjfr:
Note that the slow speed of ``groupby`` is the major driver here. I've just come up with an alternative implementation for the NdMapping groupby that temporarily converts to a DataFrame. Here's the performance profiling for a three-dimensional HoloMap grouped by two of the dimensions with both implementations:

![image](https://cloud.githubusercontent.com/assets/1550771/11733257/8bc39226-9fa3-11e5-9d70-efb6a488c3e5.png)

I think until we find a way around this we should allow HoloMap to use this implementation when pandas is available.

philippjfr:
Just some quick thoughts on this. Fundamentally the distinction between NdMapping and Columns types is in the way the data is indexed. NdMapping types are good for providing multi-dimensional indexing for dense chunks of data, e.g. Elements. Elements on the other hand hold dense chunks of data. Implementing a separate API for NdMapping types that can work using either the current OrderedDict based implementation or using pandas MultiIndexes therefore would make sense.

This second index based baseclass would also be useful for more powerful composite Element types, i.e. you could have a new Element baseclass where the data maps between multi dimensional keys and values that match the Columns data format, e.g. for a list of polygons for each country the data format would be {'Australia': {'x': xs, 'y': ys}, 'Austria': ...}. This would be very similar to what NdOverlay provides but without the need to nest the data and provide a much more optimized alternative for storing and plotting collections of artists.

This example from the pandas docs should make it clear what the data format is:

```
                     A         B         C
first second                              
bar   one     0.895717  0.410835 -1.413681
      two     0.805244  0.813850  1.607920
baz   one    -1.206412  0.132003  1.024180
      two     2.565646 -0.827317  0.569605
foo   one     1.431256 -0.076467  0.875906
      two     1.340309 -1.187678 -2.211372
qux   one    -1.170299  1.130127  0.974466
      two    -0.226169 -1.436737 -2.006747
```

[...]

PR #349, opened 2015-12-11
Speed up NdMapping.groupby

philippjfr:
This PR refactors the NdMapping.groupby operation into a separate function and provides an alternative implementation based on Pandas, which is significantly faster for large datasets. You can see the linear performance scaling by the old implementation and might just make out the sublinear performance of pandas, which becomes very significant for large datasets >10000 items. This is a temporary workaround until we come up with a general solution data API for NdMapping types that's being discussed in #347.

image

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -737,8 +737,11 @@ def unpack_group(group, getter):
 
 def ndmapping_groupby_pandas(ndmapping, dimensions, container_type,
                              group_type, sort=False, **kwargs):
-    idims = [dim for dim in ndmapping.kdims if dim not in dimensions
-             and dim != 'Index']
+    if 'kdims' in kwargs:
+        idims = [ndmapping.get_dimension(d) for d in kwargs['kdims']]
+    else:
+        idims = [dim for dim in ndmapping.kdims if dim not in dimensions]
+
     all_dims = [d.name for d in ndmapping.kdims]
     inds = [ndmapping.get_dimension_index(dim) for dim in idims]
     getter = operator.itemgetter(*inds) if inds else lambda x: tuple()
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -9,6 +9,11 @@ import numpy as np
 import param
 
 try:
+    from cyordereddict import OrderedDict
+except:
+    from collections import OrderedDict
+
+try:
     import pandas as pd
 except ImportError:
     pd = None
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -735,44 +735,56 @@ def unpack_group(group, getter):
             yield (wrap_tuple(key), obj)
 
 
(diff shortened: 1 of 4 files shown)


holoviews/core/util.py
```diff
@@ -710,3 +710,47 @@ def get_param_values(data):
...
+                              else [((), (v,))]), **kwargs))
+              for k, v in iterative_select(ndmapping, dim_names, selects)]
+    return container_type(groups, kdims=dimensions)
+
+
+try:
+    import pandas
+    ndmapping_groupby = ndmapping_groupby_pandas
+except:
+    ndmapping_groupby = ndmapping_groupby_python
```
jlstevens: I would make this a parameterized function (i.e a class) that uses one of two possible ``bothmethods`` in ``__call__``. My only other comment is that we need some docstrings here...

holoviews/core/util.py
```diff
@@ -710,3 +710,47 @@ def get_param_values(data):
...
+    for k, v in group.iterrows():
+        obj = v.values[0]
+        key = getter(k)
+        if hasattr(obj, 'kdims'):
+            yield (key, obj)
+        else:
+            yield [((), (v.ix[0],))]
+
+
+def ndmapping_groupby_pandas(ndmapping, dimensions, container_type, group_type, **kwargs):
```
jlstevens: From what I can see, this is actually used at the ``MultiDimensionalMapping`` level. Just a minor comment, as this name is certainly less unwieldy than ``multidimensional_groupby_pandas`` and I doubt this would cause much confusion.

jlstevens:
Definitely a very valuable PR: it cleans up ``groupby`` on ``MultiDimensionalMapping``, moves useful functionality into ``util`` and most importantly offers a major performance improvement (for people with pandas installed) with very little code.

I'm happy to implement my suggestion of turning ``ndmapping_groupby`` into a parameterized function once the tests are passing...

philippjfr:
I'm now done with this, I'd be happy if you refactored it into ParameterizedFunction, then we can merge.

jlstevens:
Great! The key thing is that all the tests are passing now...

My only comment now is whether you are happy for me to make this into a parameterized function? Or do you object to having a single parameterized function for groupby?

Shouldn't take long to do and I am happy to change it if you are busy.

philippjfr:
>My only comment now is whether you are happy for me to make this into a parameterized function? Or do you object to having a single parameterized function for groupby?

Must have missed my comment somehow:

>I'm now done with this, I'd be happy if you refactored it into [a] ParameterizedFunction, then we can merge.

jlstevens:
Sorry yes - I skimmed your reply too quickly. I'll do the final refactor now.

jlstevens:
If the tests pass now, I'll go ahead and merge.

jbednar:
Excellent!

Issue #350, opened 2015-12-11
Unicode in HoloMap labels


vascotenner:
HoloMaps do not render properly when the labels of the items are unicode.

Works:
```python
dim_frequency = hv.Dimension('Frequency', unit='Hz')
dim_powerspectrum = hv.Dimension('Power density', unit='dB/Hz', range=(-100,0))
hv.HoloMap({str(chr(i+65)+'a'): hv.Curve(([1,2,3],np.array([1,2,3])**i), kdims=[dim_frequency], 
                                vdims=[dim_powerspectrum]) for i in range(10)}, kdims=['Direction'])
```

Does not work:
```python
dim_frequency = hv.Dimension('Frequency', unit='Hz')
dim_powerspectrum = hv.Dimension('Power density', unit='dB/Hz', range=(-100,0))
hv.HoloMap({unicode(chr(i+65)+'a'): hv.Curve(([1,2,3],np.array([1,2,3])**i), kdims=[dim_frequency], 
                                vdims=[dim_powerspectrum]) for i in range(10)}, kdims=['Direction'])
```

```
Javascript error adding output!
SyntaxError: Unexpected string
See your browser Javascript console for more details.
```

Javascript console:
```
SyntaxError: Unexpected string(…)   outputarea.js:415
````

jlstevens:
Thanks for reporting this!

I see no reason why our javascript code should choke on unicode strings. Hopefully the fix will be relatively straight forward...

philippjfr:
This has been fixed in #376. Closing.

PR #352, opened 2015-12-11
Various unicode fixes for dimensions, Tables and widgets

philippjfr:
This PR addresses various unicode issues in Dimensions, Tables and Widgets. Opened to address #350 but fixes more wide-ranging issues.

Not yet ready for merge.


holoviews/core/dimension.py
```diff
@@ -3,6 +3,8 @@
 axis or map dimension. Also supplies the Dimensioned abstract
 baseclass for classes that accept Dimension values.
 """
+from __future__ import unicode_literals
+
```
jlstevens: Do you think we might need this anywhere else in HoloViews?

jlstevens:
Looks fine. I wish there were a good way to know when ``safe_unicode`` should be used (apart from writing as many different tests as possible!).

holoviews/core/dimension.py
philippjfr: Probably not a bad idea to add some of the ``__future__`` imports everywhere including ``division``, ``print_function`` and this one.

Issue #353, opened 2015-12-11
Addition option to specify fontsize in legend


RafalSkolasinski:
So far with magics one can adjust following fontsize's:
```
fontsize:          Specifies various fontsizes of the displayed text.
                   
                   Finer control is available by supplying a dictionary where any
                   unmentioned keys reverts to the default sizes, e.g:
                   
                      {'ticks':20, 'title':15, 'ylabel':5, 'xlabel':5}
```

It would be cool to also be able to set fontsize in legends.

jlstevens:
This issue is addressed in the PR mentioned above. I'll close this issue as soon as that PR is merged. 

If you have any issues using this feature after the merge, I'll be happy to reopen the issue.

philippjfr:
The PR has now been merged. Closing.

Issue #89, opened 2015-05-19
Options notebook tests are extremely fragile


jlstevens:
Everytime we make any change to a parameter name, parameter docstring or style option, the Options notebook tests fail (due to the `hv.help` output in that notebook). This means, we consistently have to:
- Generate updated test data (display data) for Python 2 and 3.
- Push the reference data.
- Update the submodule reference.
- Rerun the tests on Travis.

This is wasting a lot of time and causing a lot of meaningless commits (submodule reference updates). The only good thing about this situation is that when the Options notebook fails it typically means we have some API change (but it could also be a simple docstring change, e.g a typo fix!)

We would add ignore patterns to the testing code but there is too much output to try and match and ignore every time.

jbednar:
We could add support for something like the ``pyflakes-ignore`` tags we use in the source code, such that if there is a certain keyword in a comment in the input cell (``# nb-testing ignore``?), the output cell is ignored.

jbednar:
Oh, this is still open?  I thought this feature had been implemented ages ago.  Clearly wishful thinking! :-)

philippjfr:
At this point I'm going to close this, we're hoping to switch to nbval at some point and I haven't actually encountered any issues with Options tests in a long time.

PR #354, opened 2015-12-12
Added support for setting the legend font size

jlstevens:
This PR adds support for setting the fontsize in legends, addressing issue #353.

It uses the same system as everything else, i.e:

```python
%%opts NdOverlay [fontsize={'legend':14}]
hv.NdOverlay({i:hv.Curve(np.random.rand(10,2)) for i in range(3)})
```
image

### DIFF
--- a/doc/Tutorials/Options.ipynb
+++ b/doc/Tutorials/Options.ipynb
@@ -543,7 +543,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The color of the curve has been changed to red and the fontsizes of the x-axis label and all the tick labels has been modified. The ``fontsize`` is an important plot option and you can find more information about the available options in the ``fontsize`` documentation above.\n",
+    "The color of the curve has been changed to red and the fontsizes of the x-axis label and all the tick labels have been modified. The ``fontsize`` is an important plot option and you can find more information about the available options in the ``fontsize`` documentation above.\n",
     "\n",
     "The ``%%opts`` magic is designed to allow incremental customization which explains why the curve in the cell above has retained the increased thickness specified earlier. To reset all the customizations that have been applied to an object, you can create a fresh, uncustomized copy as follows:"
    ]
--- a/doc/Tutorials/Options.ipynb
+++ b/doc/Tutorials/Options.ipynb
@@ -228,7 +228,8 @@
    "cell_type": "code",
    "execution_count": null,
    "metadata": {
-    "collapsed": false
+    "collapsed": false,
+    "scrolled": false
    },
    "outputs": [],
    "source": [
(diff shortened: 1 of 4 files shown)


jbednar:
Looks very straightforward; I'd be happy to merge once the tests pass.  It would be nice to mention how to change fontsizes in the tutorials; if only that were a quick edit followed by clicking 'ok' to update the reference data!

jlstevens:
The commit above also make it possible to set the font size of the legend title using the key ``legend_title``. Note this currently applies to the matplotlib backend only as the bokeh backend doesn't have legend titles (Philipp is looking into this).

jlstevens:
> It would be nice to mention how to change fontsizes in the tutorials

Agreed. Which tutorial would you mention it in? Maybe the containers tutorial?

I need to update the test data anyway because of the options tutorial (see issue #89, it gets really annoying!).

sf-issues:
Probably the options tutorial, since this is clearly an option that
people will want to change.

-- 
The University of Edinburgh is a charitable body, registered in
Scotland, with registration number SC005336.

jlstevens:
``sf-issues`` is alive! It has gained sentience! ;-p

But yes, the options tutorial makes the most sense.

jlstevens:
It was tricky to find a nice play to insert an example of using ``fontsize`` in the Options tutorial. In the commit above, I think I found a way to insert it naturally into the tutorial while expanding on the content in a useful  way.

@jbednar @philippjfr I would appreciate it if you could have a look and give me feedback before I update the testing data.

doc/Tutorials/Options.ipynb
```diff
@@ -523,6 +524,45 @@
...
+   "source": [
+    "%%opts Curve (color='r') [fontsize={'xlabel':15, 'ticks':8}] \n",
+    "layout"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The color of the curve has been changed to red and the fontsizes of the x-axis label and all the tick labels has been modified. The ``fontsize`` is an important plot option and you can find more information about the available options in the ``fontsize`` documentation above.\n",
```
jbednar: s/has/have

jbednar:
That explains it!  In July I switched my github account to my new email address, but the effect was that I now got two copies of every single issue, one to the old and one to the new address.  I couldn't find any mention of my old address in my account, so I was completely mystified, but the reason must be that my old address is also listed in the sf-issues account!  Now I can finally get rid of that pesky deja-vu feeling that starts every day. :-)

doc/Tutorials/Options.ipynb
jlstevens: There is one curve shown so 'have' doesn't seem right to me...

jlstevens:
Glad I could help! :-)

I'm not quite sure about the comment, but as soon as that is resolved, I'll update the reference data and merge.

doc/Tutorials/Options.ipynb
jbednar: "fontsizes ... has" -> "fontsizes ... have"

jlstevens: I see it now, thanks! I'll fix that in a minute.

philippjfr:
Looks good, merging now.

Issue #351, opened 2015-12-11
PDF export via nbconvert


vascotenner:
I like to create a pdf file from my notebook. Before using holoviews I used nbconvert for that. However, hv graphs are not shown correctly, only the repr is shown.  This might have to do with all the javascript that is injected to show holoview plots.

As a semi-solution I now use the "print" function in ipython notebooks. However, in that case I loose syntax highlighting.

jlstevens:
Although PDF can support some [sort of Javascript](http://www.adobe.com/content/dam/Adobe/en/devnet/acrobat/pdfs/js_api_reference.pdf) this is often unsupported by PDF viewers and I'm not sure the API would make exporting widgets to PDF possible.

When exporting to anything other than HTML, you need to make sure the notebook doesn't contain any widgets. If you select a single element to display from your holomaps, those should export fine as static frames and if not, that is a bug we need to look into.

I think it may be possible to automatically turn a HoloMap into a frame using an nbconvert preprocessor but this is something each user of need to configure for themselves (HoloViews has no way of hooking into nbconvert). For more information, see this [stackoverflow post](http://stackoverflow.com/questions/22037910/converting-ipython-notebook-to-html-with-separate-images) for the sort of thing that can be done.

philippjfr:
I assume you mean plots in general not just HoloMaps? We'd have to look into supporting different reprs because currently all our output is wrapped in html, which probably doesn't export to pdf.

jlstevens:
Ah right, yes there is already a related issue about this: #183 

@philippjfr Given that the rendering system has been refactored even further, could this be a trivial change in the display hooks at this point? Looking at it, I think changing the ``render`` method to not necessarily output HTML and setting ``raw=True`` in ``display`` might do the trick.

The only issue then is how to register the display hooks correctly. *If* HTML takes precedence over other formats (as I think it should), you could register new formatters (png, svg) and just have an option that determines the output format (default would be HTML of course). Unfortunately, you would still need to rerun the notebook with the flag set for the change to take effect...

jlstevens:
> As a semi-solution I now use the "print" function in ipython notebooks. However, in that case I loose syntax highlighting.

@vascotenner Could you give an example of what you do here? If I understood this approach, maybe I could figure out a better solution...

vascotenner:
notebook menu: file->print preview + browser menu->print

philippjfr:
I've got an initial prototype working, at least for static figures, so it should be fairly straightforward.

[...]

PR #355, opened 2015-12-13
Support for multiple display output formats in the notebook

jlstevens:
This PR address issue #351 and makes our display system even more flexible.

You can now enable multiple display formats simultaneously (default is HTML only as before). For instance, you can now set:

```python
hv.notebook_extension(display_formats=['html','png'])
```

This will display everything as normal, but also render and save PNG data (for elements) into the notebook so that it can be used by nbconvert. As extra rendering is done, running the notebook is slower and of course HoloMaps cannot be exported as PNG unless you select a single frame to display.

You can also do:

```python
hv.notebook_extension(display_formats=['png'])
```

Which will force you to address HoloMaps for nbconvert and will also be faster to render. That said, your output will no longer be nicely centered due to how IPython displays images (out of our control).

The 'html' format is supported by everything but any other format is backend dependent. For instance, if you switch to Bokeh and ask for PNGs, you'll get text reprs instead (Bokeh currently doesn't export to PNG).

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -769,6 +769,10 @@ class Store(object):
     # types grouped by the backend. Set using the register method.
     registry = {}
 
+    # A list of formats to be published for display on the frontend (e.g
+    # IPython Notebook or a GUI application)
+    display_formats = ['html']
+
     # Once register_plotting_classes is called, this OptionTree is
     # populated for the given backend.
     _options = {}
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -105,6 +105,16 @@ class notebook_extension(param.ParameterizedFunction):
     width = param.Number(default=None, bounds=(0, 100), doc="""
         Width of the notebook as a percentage of the browser screen window width.""")
 
+    display_formats = param.List(default=['html'], doc="""
+        A list of formats that are rendered to the notebook where
+        multiple formats may be selected at once (although only one
+        format will be displayed).
+
+        Although the 'html' format is supported across backends, other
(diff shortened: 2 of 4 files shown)


philippjfr:
Looks good, great feature to have and it integrates pretty cleanly with the existing rendering and display system. Basically we now support output to different IPython backends and could easily implement implement display output to for new frontends.

jbednar:
Great!

philippjfr:
Just tested it using Atom and Hydrogen, using 'png' and 'svg' it displays plots nicely!

screen shot 2015-12-13 at 3 24 17 am

jlstevens:
@philippjfr Thats for testing it with Hydrogen! It is really cool to see HoloViews working in something other than Jupyter Notebook. In theory, Hydrogen could even support our javascript widgets as it is built on Atom that has a Javascript and rendering engine.

jbednar:
Neat!

majidaldo:
what would it take for hydrogen to display the ipython widgets?

jbednar:
ipywidgets is a separate project, and isn't used by HoloViews.  Not sure how much work it would take to get the JavaScript widgets that it does use working...

jlstevens:
I think in this case the question is about the HoloViews widgets.

A while back I do think I got a slider working with hydrogen. I can try again but my worry is that if I do get it working, none of use use hydrogen to make sure anything we add to HoloViews stays working and supported.

jlstevens:
I had another go at getting the widgets working with nteract but without success. When we get round to the widget Javascript code refactor we should look into some way of generating JS code that is self contained and includes all the necessary dependencies.

PR #356, opened 2015-12-13
Allowed specifying resources via *args in notebook_extension

jlstevens:
This is a small PR for something I felt might be useful. As ``notebook_extension`` is quite new, I found myself trying this syntax:

```python
```

When I should have done this:

```python
hv.notebook_extension(bokeh=True)
```

Given that I made this mistake in ``notebook_extension`` (I'm the one who implemented the keyword syntax!) this proposed syntax is definitely an obvious one. I saw no reason why both styles can't be supported (the keyword approach allows tab-completion and disabling HoloViewsJS) which is why I've submitted this PR.

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -119,8 +119,8 @@ class notebook_extension(param.ParameterizedFunction):
 
     _loaded = False
 
-    def __call__(self, **params):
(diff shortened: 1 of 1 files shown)


philippjfr:
Looks fine and very straightforward. Will merge.

PR #357, opened 2015-12-14
Streams API

philippjfr:
This PR adds an API to link dynamic state such as the Bokeh zoom level to the callbacks on DynamicMap. This allows the user to update plots based on the state of other plot objects, making complex interactive behaviors with bokeh widgets (as well as any other dynamic way to alter plots) very simple.

Edit: This is a prototype to see if I could get this working quickly, not ready review or merge.

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -14,6 +14,8 @@ except ImportError:
 
 import param
 
+from ...core.spaces import Stream
+from ..widgets import NdWidget
 from .plot import BokehPlot
 
 
@@ -137,7 +139,7 @@ class Callback(param.ParameterizedFunction):
             return self.serialize(objects)
 
 
-    def serialize(self, objects):
+    def serialize(self, objects, serialized={}):
         """
         Serializes any Bokeh plot objects passed to it as a list.
         """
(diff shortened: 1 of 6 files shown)


holoviews/plotting/bokeh/callbacks.py
```diff
@@ -148,9 +150,45 @@ def serialize(self, objects):
...
                 json = obj.to_json(False)
             data[obj.ref['id']] = {'type': obj.ref['type'], 'data': json}
+        data.update(serialized)
         return serialize_json(data)
 
 
+class StreamCallback(Callback, Stream):
+    """
+    The StreamCallbacks are a type of callback, which simply updates
+    the current state of the Stream and then.
```
jbednar: and then?

philippjfr: Not ready for review.

jlstevens:
It is nice to see something we've discussed for so long start to take shape!

Anyway to reiterate something we've been discussing privately, I suggest instead of having ``DynamicMap.streams`` as a dictionary, it should be a list. E.g something like:

```python
DynamicMap(callback, streams=[Range('range'), Zoom('zoom')])
```

This would make streams look similar to ``Dimension`` objects without actually being the same thing. Secondly, the names of the streams can be used to pass in the stream values into the callback as ``**kwargs``, instead of as a dictionary called ``streams``.

I have some new suggestions to make as well, based on a proposed example you showed me earlier: 

```python
range_stream = RangeStream.instance()
cb = Callbacks(ranges=range_stream)

%opts Curve [callbacks=cb]
hv.DynamicMap(test_fn, kdims=[hv.Dimension('X', range=(0,10)), 
                              hv.Dimension('Y', values=[0, 1, 2])], 
                              streams={'range': range_stream}) 
```

Other than simplifying the streams parameter, there are more improvements I think could be made. For a start, I think we can make this work without the ``%opts`` magic (hence making it more general). Anything the ``%opts`` magic can achieve, we can do behind the scenes (using ``OptionTrees`` and the same mechanism if necessary). We have the full power of the opts magic syntax using the dictionary format you can use in the ``__call__`` method if we need it.

Additionally, I think we might be able to do some of work defining the global ``Callbacks`` object easier. For instance, maybe something like this:

```python
callbacks=Callbacks()
...
streams=[RangeStream('range', callbacks=callbacks)]
```

If I remember correctly, the reason we need something along these lines is that there needs to be a global object to register all the callbacks and manage all the streams. I realize there are probably some subtle problems in the code we'll see to solve but an ideal API would remove the need for the user to declare ``Callbacks`` entirely (unless they want to!).

Perhaps declaring a set of ``Stream`` objects in a list could implicitly define them as part of the same ``Callbacks`` object?

I have a few other proposals for more important features (e.g modifying caching behavior as I've briefly discussed with Philipp) but this is enough for us to consider for now. :-)

PR #359, opened 2015-12-15
Various Optimizations

philippjfr:
This is a PR for various optimizations on some of the core components of HoloViews.

### DIFF
--- a/holoviews/core/ndmapping.py
+++ b/holoviews/core/ndmapping.py
@@ -14,7 +14,8 @@ import param
 from . import traversal, util
 from .dimension import OrderedDict, Dimension, Dimensioned, ViewableElement
 from .util import (unique_iterator, sanitize_identifier, dimension_sort,
-                   basestring, wrap_tuple, process_ellipses, itervalues)
+                   basestring, wrap_tuple, process_ellipses, itervalues,
+                   get_ndmapping_label)
 
 
 class item_check(object):
@@ -756,12 +757,7 @@ class UniformNdMapping(NdMapping):
             return self._group
         else:
             if len(self):
-                group = None
-                els = itervalues(self.data)
-                while group is None:
-                    el = next(els)
-                    if not el._auxiliary_component:
-                        group = el.group
+                group = get_ndmapping_label(self, 'group')
                 tp = type(el).__name__
                 if tp != group:
                     return group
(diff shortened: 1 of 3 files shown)


holoviews/core/ndmapping.py
```diff
@@ -778,10 +779,8 @@ def label(self):
         if self._label:
             return self._label
         else:
-            labels = {v.label for v in self.values()
-                      if not v._auxiliary_component}
-            if len(labels) == 1:
-                return list(labels)[0]
+            if len(self):
+                return next(itervalues(self.data)).label
```
jlstevens: I think you still need to check ``_auxiliary_component`` as that marks ``Annotations`` where you don't want to propagate the label upwards.

philippjfr: Bottom layer should usually not be an annotation but I can just throw a while loop in there that'll iterate until it finds a non-auxiliar component.

jlstevens: My understanding is that label should be the empty string if the first value (i.e the one we are inspecting) is an ``_auxiliary_component``. Before we were iterating over the NdMapping elements: when you say the bottom layer shouldn't be an annotation, I believe you are thinking about the elements of an Overlay. In other words, I think we need this check for HoloMaps of annotations.

jlstevens:
One quick comment: the code looks good but seems to repeat itself (almost exactly, but not quite) between group and label. I would consider a utility or this, either a function with an option to get the group or label, or maybe a class with two classmethods ``get_group`` and ``get_label``.

Otherwise it looks good and I am happy to merge.

jlstevens:
Looks good once the tests are passing. 

One thing you didn't mention at the start of the PR was how slow the old label/group processing was. I know this will offer some sort of performance improvement but I don't have a feeling for what kind of speed up I can expect and in which situations.

philippjfr:
This PR is ready for merge now that the tests are passing.

philippjfr:
>One thing you didn't mention at the start of the PR was how slow the old label/group processing was. I know this will offer some sort of performance improvement but I don't have a feeling for what kind of speed up I can expect and in which situations.

For very nested datastructures it saves a lot of time. I tested it one a HoloMap of 100 Overlays each containing 100 Curves, in that example it saved 120/130 seconds.

jlstevens:
Thanks. That is a plausible holomap and 120 seconds is a significant saving - although you didn't say how long it took in total. :-)

PR #360, opened 2015-12-15
Fix to how AttrTree handles name mangled attributes

jlstevens:
This PR is tiny but it constitutes a bug fix in an important part of the code that can potentially affect a lot of things. I want to make sure the tests pass here before touching master.


philippjfr:
Looks right to me and the tests seem to agree. Merging.

PR #362, opened 2015-12-16
Bokeh improvements

philippjfr:
This PR will include various improvements for the bokeh backend. First of all it now has proper support for blanking out frames with the empty option, which also allows downsampling to be applied without having to process all the data first. I still want to switch the VLine and HLine implementation to use BoxAnnotations to address zorder and other issues, so the PR isn't quite ready yet.

Note that while all the changes related to changing the BokehPlot get_data method to support an empty option might seem like overkill simply to get callbacks working it's also the first step to being able to reuse plots in an Overlay, which will eventually provide a major optimization.

### DIFF
--- a/holoviews/plotting/bokeh/__init__.py
+++ b/holoviews/plotting/bokeh/__init__.py
@@ -130,3 +130,8 @@ options.Image = Options('style', cmap='hot')
 options.Raster = Options('style', cmap='hot')
 options.QuadMesh = Options('style', cmap='hot')
 options.HeatMap = Options('style', cmap='RdYlBu_r', line_alpha=0)
+
+# Annotations
+options.HLine = Options('style', line_color='black', line_width=3, line_alpha=1)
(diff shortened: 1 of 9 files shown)


jlstevens:
It all sounds good and I don't see any huge blocks of new code to consider. You have a much better understanding of the current limitations of our Bokeh support so I am happy to let you merge whenever you think it is ready.

For ``VLine`` and ``HLine`, I've noticed that something you can see the ends of the lines when extents change (e.g in a ``DynamicMap``). If there much harm ( performance issues perhaps?) extending the lines say 100x the known extents to be safe? If Bokeh does effective clipping, I don't see why we need to try to set the extents of these lines to exactly what is visible.

philippjfr:
I've implemented the ``HLine`` and ``VLine`` using the new Bokeh BoxAnnotation (can replace it with LineAnnotation once bokeh 0.11 is released). This means it's extents always cover the full range of the plot and will always be on top of all the other Elements. With that implemented the PR is ready for merge.

PR #364, opened 2015-12-16
Fixing constant dimensions

philippjfr:
This PR fixes various issues with constant dimensions, in particular it ensures the the default call to .dimensions no longer returns the constant dimensions, which caused a wide range of issues.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -508,7 +508,7 @@ class Dimensioned(LabelledData):
 
     __abstract = True
     _sorted = False
-    _dim_groups = ['kdims', 'vdims', 'ddims']
+    _dim_groups = ['kdims', 'vdims', 'cdims', 'ddims']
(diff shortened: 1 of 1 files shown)


jlstevens:
The change looks fine but I am slightly concerned that it pushes constant dimensions to the side. We haven't thought too deeply about constant dimensions but there are cases where they are definitely useful.

I'm also worrying why you didn't just leave ``_dim_groups`` alone and then exclude 'cdims' and 'constant_dimensions' inside ``dimensions`` when ``selection='all'`` (the default). Does removing 'cdims' from ``_dim_groups`` have another effect you need in this fix?

philippjfr:
I'll readd them to dim_groups but until we've thought them through properly returning by default breaks all kinds of assumptions. We should open an issue to discuss how they should actually be used and then integrate them properly.

jlstevens:
Ok great! I'll merge as soon as the tests pass.

jlstevens:

PR #366, opened 2015-12-18
Improved tracebacks

jlstevens:
This pull request implements much nicer abbreviated traceback behavior. Abbreviated tracebacks only occur when the exception originates in a backend module (e.g matplotlib or bokeh) and not from HoloViews itself (unless a ``BackendError`` has been explicitly raised).

In addition, there is no need to worry about ``ABBREVIATE_TRACEBACKS`` anymore as you can just call ``ipython.show_traceback()`` after an abbreviated traceback has occurred in order to view the error in full (with syntax highlighting!). 

If the tests pass I'm happy to see this merged although maybe I should remove ``ABBREVIATE_TRACEBACKS`` entirely? Do we care about backward compatibility in this case?

Lastly, any feedback from anyone willing to try out this branch would be much appreciated.

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -42,6 +42,17 @@ from .tree import AttrTree
 from .util import sanitize_identifier, group_sanitizer,label_sanitizer
 from .pprint import InfoPrinter
 
+
+class BackendError(Exception):
+    """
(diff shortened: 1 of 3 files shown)


philippjfr:
Looks fine to me. Happy to merge.

PR #367, opened 2015-12-18
Dynamic overlay

philippjfr:
This PR makes it possible to use return Overlays with changing keys in a DynamicMap callback. It's a sensitive change, which will require some more testing.

Note that while it now displays everything certain things do not yet work:

* Legends do not yet change to reflect the new NdOverlay key or Overlay label
* In matplotlib styles will not consistently update.

And certain things cannot work:

* Cyclic styles cannot work with a DynamicMap as the total ordering is required during instantiation. Could consider raising a warning when an Element returned from a DynamicMap is given a cyclic style.
* You cannot return more Elements or different types of Elements than are supplied during instantiation.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -611,6 +611,14 @@ def sort_topologically(graph):
                                     (names_by_level.get(i, None)
                                      for i in itertools.count())))
 
+def get_overlay_spec(o, k, v):
+    """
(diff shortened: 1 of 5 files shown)


philippjfr:
We've opened some issues to improve the validation for DynamicMap return types. For now though this PR is ready for merge.

PR #368, opened 2015-12-19
Full support of the style system using DynamicMaps 

jlstevens:
This PR fixes a major issue with ``DynamicMaps``, namely that they did not properly support styling. This is because when a ``DynamicMap`` is created it may contain *no* elements meaning that  there is nothing to propagate an ``id`` to when ``%%opts`` or ``__call__`` is used. In other words, every time a value is returned by the callback, it was too late as styling had already been applied.

The approach used here seems to work well but there are things I am not entirely happy with:

* Using ``__class__.__name__`` to check for DynamicMap. I am fairly sure I'll encounter cyclic import issues if I try to import ``DynamicMap`` into ``options.py``.
* It is a little inefficient as I'm converting from an ``OptionTree`` to dictionary format and back to an ``OptionTree`` per frame.
* This is going to generate a ludicrous number of custom ids and option trees if you use the DynamicMap much (one per frame!). Many of these will be identical as well (until the style is changed, if ever).

Hopefully, I can address these issue but I think this PR is very important for the 1.4.1 release. Styling ``DynamicMap`` as easily as ``HoloMap`` is crucial!

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -969,6 +969,21 @@ class StoreOptions(object):
 
 
     @classmethod
+    def tree_to_dict(cls, tree):
+        """
+        Given an OptionTree, convert it into the equivalent dictionary format.
+        """
+        specs = {}
+        for k in tree.keys():
+            spec_key = '.'.join(k)
+            specs[spec_key] = {}
+            for grp in tree[k].groups:
+                kwargs = tree[k].groups[grp].kwargs
+                if kwargs:
+                    specs[spec_key][grp] = kwargs
+        return specs
+
+    @classmethod
     def propagate_ids(cls, obj, match_id, new_id, applied_keys):
         """
         Recursively propagate an id through an object for components
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -994,8 +994,10 @@ class StoreOptions(object):
             raise AssertionError("The set_ids method requires "
                                  "Store.custom_options to contain"
                                  " a tree with id %d" % new_id)
-        obj.traverse(lambda o: setattr(o, 'id', new_id)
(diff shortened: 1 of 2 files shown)


jlstevens:
Just a thought regarding the excess custom ids.... maybe the custom id/tree could be removed when popped from the cache? Then again, someone could still have a handle on an element that is no longer on the cache.

Python does offer the ability to know how many references there are to an object:

```
import sys
sys.getrefcount(object) #-- Returns the reference count of the object.
```

Maybe we could use this to implement a system to garbage collect the custom trees...

jlstevens:
I'm just brainstorming here but there is another approach to clean-up the custom trees. This is probably very over-engineered but I'm just trying to come up with ideas!

1. Create an object that looks like a dictionary but is in fact a compressed dictionary: for each unique item it holds, it keeps a list of associated ids.
2. Use tree comparisons to make sure the trees held are all unique. If not, 'compress' this dictionary object to eliminate duplicate option trees.

This would solve the duplicate option trees but there is no reason the list of integer ids wouldn't be able to grow without limit. This could only be solved by sweeping the namespace for all HoloViews objects, and checking with ids actually are being used (then cleaning up the trees associated with the ones no longer in use). 

Technically possible (using ``globals()`` or the IPython namespace) but really a horrible hack. Could be made into a utility I suppose that exists to be used only when needed...

philippjfr:
Looks good, glad this could be done fairly easily. The suggestions about squashing the OptionsTrees probably deserve their own issue. I'm happy to merge this now.

jlstevens:
>  The suggestions about squashing the OptionsTrees probably deserve their own issue.

Sure. I've opened an issue about this above. I think you can go ahead an merge!

Issue #331, opened 2015-12-04
Filtering help strings using regexps


maxalbert:
The help system is really awesome and comprehensive (huge thanks for documenting everything so well!). However, I find that it is sometimes difficult to find the correct piece of information in all the output. It would be fantastic if there was a quick way of only displaying those options whose name or description contain a certain keyword. For example, calling
```
hv.help(obj, regexp="ticks")
```
would display a filtered help string which only shows those options containing the pattern `"ticks"`.

This could be combined with the suggestion in #330 to quickly and succinctly find the relevant options on a composite object.

jlstevens:
Thanks for the more detailed explanation than what I mentioned in #330. Both this issue and #330 will probably be dealt with at the same time.

jlstevens:
This issue has been addressed in PR #370 (now merged) so I'll close the issue now.

@maxalbert Hopefully you'll find this a useful way to find the bits of the documentation you need. It even highlights your search terms which is a little detail I rather appreciate myself...

Issue #330, opened 2015-12-04
Concatenated help strings for composite objects


maxalbert:
If I have a nested structure such as
```
:NdLayout   [Frequency,Phase]
   :HoloMap   [Exponent]
      :NdOverlay   [Shift]
         :Curve   [x]   (y)
```
and I want to find the correct object to set an option (for example `xticks`), currently I need to call `hv.help` on each of the classes `NdLayout`, `HoloMap`, `NdOverlay` and `Curve` separately and search the docstrings.

It would be great to be able to call
```
hv.help(obj, recursive=True)
```
or
```
hv.help(..., all=True)
```
or something similar. This would display the docstrings for all four objects concatenated together in a single long docstring, which would make searching for the correct option using `Ctrl+F` in the browser much less cumbersome.

jlstevens:
Good idea. On gitter you also proposed support for regular expression search e.g ``hv.help(obj, regexp="ticks")`` which could also be very useful.

jlstevens:
This issue has been addressed in PR #370 so I'll close the issue now. 

@maxalbert If you could try it out and see if you have any feedback, that would be great!

PR #370, opened 2015-12-19
Improvements to the help system

jlstevens:
This PR implements some improvements to the help system, suggested in #331.

Supports filtering of parameters using regular expressions as well as highlight. For instance:

```python
hv.help(hv.Image, pattern='bounds')
```

image
 
```python
hv.help(hv.Image, visualization=False, pattern='bounds')
```
image

It seems to work well even though I needed to employ some clever tricks to filter parameters. Now I'll look into addressing issue #330 in this PR as well..


jlstevens:
I've now added a recursive option to address #330. This option will only work when given a nested object and will not do anything is a class is specified instead.

Once the tests pass, I think this is ready to merge.

philippjfr:
Looks good. Great to have a better help system. Merging.

maxalbert:
Great, this seems to work well. Many thanks!

I wonder whether it would be worth setting `recursive=True` by default to increase discoverability of options for composite objects (but I don't have a strong opinion on it).

It would also be good to document this new functionality in one of the tutorials (maybe the "Options" tutorial, since it already illustrates the `help` command?).

jlstevens:
> I wonder whether it would be worth setting recursive=True by default to increase discoverability of options for composite objects

Not a bad suggestion! I might do that after the website has been updated as changing it immediately will probably break notebook tests. 

> It would also be good to document this new functionality in one of the tutorials (maybe the "Options" tutorial, since it already illustrates the help command?).

Yes, that is definitely something I will do now. I can change Markdown text much more easily than output generated by code. Thank you for reminding me to do this!

jlstevens:
I've added a brief description of these parameters in commit 04f77e4 to the Options tutorial.

@philippjfr Maybe we should use the pattern argument in the actual example? This would mean updating the test data but I see three advantages:

1. Even ``hv.help(image)`` is generating too much output in the tutorial, making it hard to follow.
2. It would mean we are less likely to have to update the test data when parameter documentation is updated.
3. It would give a concrete example of ``pattern`` in use and why it is helpful.

What do you think? To be concrete, I'm considering showing the output of ``hv.help(image, pattern='bounds')``...

jbednar:
Sounds good to me.

maxalbert:
+1

jlstevens:
I've gone and implemented this suggestion in 5ed9eda.

Issue #334, opened 2015-12-06
Testing the bokeh Elements tutorial and matching the original


jlstevens:
I think it would be great to just copy the current Elements tutorial, set `hv.notebook_extension(bokeh=True)` at the top and make it available on the website (the tutorials could then be called `Elements [matplotlib]` and `Elements [bokeh]`. This would achieve a few things:
1. Show some official bokeh examples on or website.
2. Show how easy it is to use the same code and switch backends.
3. Test our bokeh support.
4. Show people the full range of things that can be done with the bokeh backend.

The main issues I can see are:
1. Setting this up on Travis (I think bokeh is already installed though).
2. Getting display tests working for bokeh. This may be tricky if there is a fair bit of randomness (e.g uuids) in the bokeh output given the same input.

jlstevens:
I forgot to mention you'll need ``%output backend='bokeh'`` too (maybe ``notebook_extension`` should switch for you?). I also notice that some things still aren't available (e.g ``VectorField``?) and when I run this cell, the rest of the notebook doesn't run:

```python
data = [('one',8),('two', 10), ('three', 16), ('four', 8), ('five', 4), ('six', 1)]
bars = hv.Bars(data, kdims=[hv.Dimension('Car occupants', values='initial')], vdims=['Count'])
bars + bars[['one', 'two', 'three']]
```

philippjfr:
I've already added a version of the Elements Tutorial using Bokeh. Do still need to get bokeh testing working though.

jlstevens:
Ah, yes... I see that commit fro 19 hours ago. :-)

Though it would be good if it were made as similar to the original tutorial as possible (e.g the issue above). What do you think about effectively calling the output magic to switch backends when using ``notebook_extension(bokeh=True)``? I guess it gets confusing if you load JS for multiple things because then it gets ambiguous...

jlstevens:
Ok, I changed the title of the issue. I'm sure you are already aware of some of the following problems but they are worth listing here anyway:

* The index isn't rendering properly.. e.g a line shows up as ``` [``Element``](#Element) ```. This is true for the normal tutorial too (dev section of the website).
* In ``Histogram`` you get a warning: ```WARNING:root:Options: Invalid options ['show_grid'], valid options are: ...```
[...]

PR #371, opened 2015-12-20
Bokeh Comms

philippjfr:
This PR fixes bokeh comms to retain the order of objects to update fixing issues with colormaps not updating correctly. Additionally the callbacks are also updated and now work with the dev release of bokeh and should be compatible with the upcoming v0.11 release.


philippjfr:
Ready for merge, was pretty straightforward.

jlstevens:
I'll trust you on this one! 

We don't have proper bokeh testing, so I can't rely on the fact that tests are passing (see #334). I'll merge now that you say it is ready.

PR #374, opened 2015-12-20
Notebook extension now switches backend automatically

jlstevens:
This PR addresses some of our later discussion in #334. Now when you use ``notebook_extension`` the backend associated with the last block of Javascript that gets loaded is activated.

The resource order is determined by the ``*args`` e.g ``hv.notebook_extension('bokeh', 'matplotlib')`` but if keywords are used, the order of the renderers registered in ``Store`` (i.e the order in which backends are registered) determines what will be activated.

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -34,6 +34,7 @@ Store:
 """
 import pickle
 from contextlib import contextmanager
+from collections import OrderedDict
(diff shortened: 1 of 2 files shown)


philippjfr:
Looks good to me, pretty straightforward, will merge.

PR #375, opened 2015-12-20
Disabled caching for live and dynamic modes

philippjfr:
This PR disable the JS cache for 'live' and dynamic modes, making the behavior more predictable.


jlstevens:
I agree with this change. Predictability is very important, I don't think caching on the client-side will speed things up very much (at least not when you are using Bokeh) and if we do want client-side caching, we need to be very careful to make it work predictably without the user needing to know about it.

I'm happy to merge once the tests pass.

jlstevens:
The tests are passing now.

PR #376, opened 2015-12-20
Unicode fixes for SelectionWidget

philippjfr:
This PR fixes various issues with unicode dimension labels and values in the SelectionWidget as seen in #350. It is a quick fix to deal with unicode specifically for the widgets, but in the long term we should switch to unicode everywhere as I started to do in #352.


philippjfr:
Tests are passing, this is ready to merge.

jlstevens:
Ok merging.

Issue #333, opened 2015-12-05
Substitution system not working in NotebookArchive


jlstevens:
I found issues with using the ``Substitute`` notebook preprocessor in ``NotebookArchive``. I tried fixing this for 1.4 but it is rather involved so I won't let this hold up the release.

The only side-effect of this not working is that archives may contain image files that are redundantly encoded as base64 in the exported HTML.

jlstevens:
This issue was fixed in the PR mentioned above (#378).

PR #378, opened 2015-12-20
Fix to substitution system in NotebookArchive

jlstevens:
This PR restores the functionality of HTML substitutions when using ``NotebookArchive``, addressing issue #333.

In addition, this PR fixes an issue where the same data was added multiple times to the archive due to the new support for multiple types of active display hook.

I've tested that it now works several times and if the Travis tests also pass, I think this can be merged.


jlstevens:

Issue #340, opened 2015-12-07
Skipping lines for pyflakes


jlstevens:
Currently we use comments like this for pyflakes to ignore:

```python
from .core import archive                               # pyflakes:ignore (API import)
```

However, this type of comment is only ignored by a custom pyflakes script we used to use in topographica. 

We should use a standard tool for this such as [flake8](https://flake8.readthedocs.org/en/latest/) that is supposed to skip lines with the ``# noqa`` comment. Maybe we can search and replace ``pyflakes:ignore`` with ``noqa`` everywhere and suggest people use flake8?

jbednar:
Sure, if there is now a standard tool for this, then great!  We've had to make our own solutions for so many things because we've been doing this for so long now, but anything we can offload to a standard solution is a big win for us!

jlstevens:
The PR switching us to use flake8 has now been merged. Closing this issue.

PR #379, opened 2015-12-21
Switching to flake8 for passive code checks

jlstevens:
Addresses issue #340.

As making pyflakes happy is a bigger priority for us than full PEP8 compliance, the basic test we want to pass without any warnings is this:

```python
flake8 --ignore=E,W   .
```

Running this in the main holoviews directory will run the checks. If we want to look at PEP8 errors and warnings, we only need to remove ``E`` or ``W`` from the ``--ignore`` option.

It will be nice to be able to use a standard tool for this. Ideally we would run ``flake8`` on Travis without making having flakes errors cause test failures.


jlstevens:
Phew! It was a fair bit of effort but there are only two flake warnings left:

```
./holoviews/plotting/bokeh/chart.py:94:73: F821 undefined name 'glyph'
./holoviews/plotting/mpl/chart.py:1035:13: F841 local variable 'clime' is assigned to but never used
```

The latter was probably supposed to be called ``clim`` but the former warning is definitely an actual bug.

@philippjfr Could you please check d8b0d71 in case there were some API (or side-effect driven) imports and also 6599443.

Soon holoviews will work properly with a standard static checker and will pass without errors (ignoring PEP8 of course)! :-)

jlstevens:
The tests are passing and I would be ready to merge except for the ``undefined name 'glyph'`` error. I hope @philippjfr will have a chance to look at it soon as this is the only think I'm currently aware of holding up the 1.4.1 release!

philippjfr:
Great thanks for going over this. I've just pushed the fixes you mentioned and the commits you pointed out look fine to me. Happy for you to merge this as I'll be traveling most of the day.

jlstevens:
Ok. Merging now.

jlstevens:
As I often refer to this issue when I want to remember how to invoke flake8, I'll note that things have changed recently. Now the flakes we are interested in are given by:

```bash
flake8 --ignore=E,W,F999  .
```

jlstevens:
Another one. Use:

```
flake8 --ignore=E,W,F999,F405  .
```

PR #383, opened 2015-12-26
python 3 compatibility fix

stonebig:
zip returns an iterable under python 3, and so 'Showcase.ipynb' doesn't work

### DIFF
--- a/holoviews/core/traversal.py
+++ b/holoviews/core/traversal.py
@@ -111,8 +111,8 @@ def hierarchical(keys):
     ndims = len(keys[0])
     if ndims <= 1:
         return True
-    dim_vals = zip(*keys)
-    combinations = (zip(*dim_vals[i:i+2])
+    dim_vals = list(zip(*keys))
+    combinations = (list(zip(*dim_vals[i:i+2]))
                     for i in range(ndims-1))
     hierarchies = []
     for combination in combinations:
--- a/holoviews/core/traversal.py
+++ b/holoviews/core/traversal.py
@@ -112,7 +112,7 @@ def hierarchical(keys):
     if ndims <= 1:
         return True
     dim_vals = list(zip(*keys))
-    combinations = (list(zip(*dim_vals[i:i+2]))
+    combinations = (zip(*dim_vals[i:i+2])
                     for i in range(ndims-1))
     hierarchies = []
     for combination in combinations:


jlstevens:
Thanks for this!

Now I'm wondering how the Python 3 notebook tests passed if this wasn't working?! Maybe the reference data was updated to something broken?

holoviews/core/traversal.py
```diff
@@ -111,8 +111,8 @@ def hierarchical(keys):
     ndims = len(keys[0])
     if ndims <= 1:
         return True
-    dim_vals = zip(*keys)
-    combinations = (zip(*dim_vals[i:i+2])
+    dim_vals = list(zip(*keys))
+    combinations = (list(zip(*dim_vals[i:i+2]))
                     for i in range(ndims-1))
```
philippjfr: Is the second list needed here? Seems like ``combination`` in the nested for loop below is only used as an iterator.

stonebig:
I did a hammer solution, maybe one is sufficient

philippjfr:
>Maybe the reference data was updated to something broken?

Must be in an untested cell since this PR doesn't break the tests.

jlstevens:
> Must be in an untested cell since this PR doesn't break the tests.

The only way something would be untested is if the HoloViews object is not displayed and the exception doesn't affect subsequent display. Certainly possible but definitely worth checking!

Anyway, the PR itself looks fine to me though I would go with Phillip's suggestion of just fixing ``dim_vals`` if that is sufficient.

jlstevens:
Ok, I can reproduce it. For me the cell that fails is:

```python
items = [(k, hv.Image(sine(grid, *k), vdims=['Amplitude'])) for k in keys]
circular_wave = hv.HoloMap(items, kdims=dimensions)
circular_wave
```

In other words, this should have been caught by the python 3 notebook tests with both data and display failures. We need to find out why the tests passed when they shouldn't have!

jlstevens:
Ok, the tests are passing (not that seems to say much!) so I will merge now.

jlstevens:
I've figured out the issue with the notebook tests. To compare display output, we look at the middle frame of the holomap because animated output formats such as gif or avi don't compare nicely bitwise (we could compare javascript if we removed the uuids appropriately).

This means that not quite the same code is exercised in the tests which is why the tests pass but the cell breaks when the notebook is actually run. The data test must have passed because the HoloMap object was created correctly.

This issue of picking display formats that can be compared bitwise and making sure the proper code is run is something to consider as part of the testing process. This is something we are trying to improve in PR #318 and we should discuss this issue there.

Issue #363, opened 2015-12-16
Improvement of scaling option for Points element


maxalbert:
*This is copied from a discussion on Gitter. Apologies for the length of this text, but I wanted to create a Github issue so that it's not forgotten. Unfortunately, I don't have time to to shorten it right now.*

I'm trying to plot some data as a collection of "bubbles" of varying sizes using the `Points` elements, but I'm having a hard time getting scaling to work.
The reason is that IMHO the `scaling_factor` option works in a very non-intuitive way.
Here is a small example to illustrate my point. First, let's create some random x/y positions and bubble sizes.
```
N = 100
np.random.seed(12345)
x = np.random.rand(N)
y = np.random.rand(N)
sizes = np.arange(N)
df = pd.DataFrame(dict(x=x, y=y, bubblesize=sizes), columns=['x', 'y', 'bubblesize'])
```
Let's plot this data:
```
%%opts Points [size_index=2 scaling_factor=1]
Points(df, kdims=['x', 'y'], vdims=['bubblesize'])
```
Just from looking at this cell I would read it as follows: "Plot a bunch of points at the x/y positions found in the first two columns of the data frame `df`, and set the point sizes according to the values in the third column; in doing so, use a scaling factor of 1, i.e. don't alter the values in the third column (**but still respect these values!**). I would expect that  chosing a scaling factor of 5 instead of 1 would result in every point being 5 times larger than before, but the *relative* point sizes should remain the same.

However, this is not how it works. The above cell produces a plot where all points have exactly the same size. At first I thought that this was perhaps because the scaling factor was too small and resulted in point sizes that were too small to be distingushable on screen. Thus I changed the scaling factor to 50 (the same value as in the example from the Holoviews documentation) but ended up with a plot that was uniformly blue. So I kept reducing the scaling factor to successively smaller values (10, 5, 2, 1.5, 1.2, 1.1), none of which made a difference. To be honest, at first I had no idea what was going on and tried fiddling with the plot for a good 15 minutes. Admittedly, I should have looked at the documentation much sooner but it didn't even occur to me that `scaling_factor` would do anything other than linearly scale the point size (either in terms of radius or area).

I finally ended up looking at the help string for `scaling_factor`, which at least made it clear that some sort of exponential scaling is being used (although the details of the method still escape me), although I must say that I still find the scaling behaviour rather unintuitive. But my subjective opinion aside, I think there are also significant objective drawbacks to the current behaviour.

[...]

PR #390, opened 2016-01-03
Decouple scaling method from scaling factor (as per issue #363).

maxalbert:
This PR implements a fix for #363. It introduces a new option `scaling_method` which determines whether `scaling_factor` should scale the point width or area (no other scaling methods are currently allowed, although they could certainly be added if there are other sensible alternatives).

If this PR is accepted then some of the notebooks will need updating as well (and the new option should be documented somewhere (possibly in the `Elements` notebook?). I haven't done this yet because I don't fully understand the notebook testing process and was worried that I'd break something.

Also, I couldn't find a unit test for the helper function `utils.compute_sizes()`. I'm not sure how fine-grained the unit testing process in holoviews is supposed to be so haven't added one, but let me know if you'd like me to do this.

### DIFF
--- a/doc/Tutorials/Elements.ipynb
+++ b/doc/Tutorials/Elements.ipynb
@@ -509,7 +509,7 @@
    "source": [
     "Thus the ``Scatter`` object expresses a dependent relationship between *x* and *y*, making it useful for combining with other similar ``Chart`` types, while the ``Points`` object expresses the relationship of two independent keys *x* and *y* with optional ``vdims`` (zero in this case), which makes ``Points`` objects meaningful to combine with the ``Raster`` types below.\n",
     "\n",
-    "Of course, the ``vdims`` need not be empty for ``Points``; here is an example with two additional quantities for each point, as ``value_dimension``s *z* and &alpha; visualized as the color and size of the dots, respectively:"
+    "Of course, the ``vdims`` need not be empty for ``Points``; here is an example with two additional quantities for each point, as ``value_dimension``s *z* and &alpha; visualized as the color and size of the dots, respectively. The point sizes can be tweaked using the option `scaling_factor`, which determines the amount by which each point width or area is scaled, depending on the value of `scaling_method`."
    ]
   },
   {
@@ -520,7 +520,7 @@
    },
    "outputs": [],
    "source": [
-    "%%opts Points [color_index=2 size_index=3 scaling_factor=50]\n",
+    "%%opts Points [color_index=2 size_index=3 scaling_method=\"width\" scaling_factor=50]\n",
     "np.random.seed(10)\n",
     "data = np.random.rand(100,4)\n",
     "\n",
--- a/doc/Tutorials/Pandas_Conversion.ipynb
+++ b/doc/Tutorials/Pandas_Conversion.ipynb
@@ -565,7 +565,7 @@
    },
    "outputs": [],
    "source": [
-    "%%opts Curve (color='k') Scatter [color_index=2 size_index=2 scaling_factor=10.0] (cmap='Blues' edgecolors='k')\n",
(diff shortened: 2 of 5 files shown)


philippjfr:
>This PR implements a fix for #363. It introduces a new option scaling_method which determines whether scaling_factor should scale the point width or area (no other scaling methods are currently allowed, although they could certainly be added if there are other sensible alternatives).

Great, the PR is nice and short and as discussed #363 I definitely agree this is the correct approach.

>If this PR is accepted then some of the notebooks will need updating as well (and the new option should be documented somewhere (possibly in the Elements notebook?). I haven't done this yet because I don't fully understand the notebook testing process and was worried that I'd break something.

We'll see what breaks after Travis is done, but it should only be a few examples. Unfortunately there has been no improvements to the testing process yet so I'll either have to walk you through it or do it myself after merge.

>Also, I couldn't find a unit test for the helper function utils.compute_sizes(). I'm not sure how fine-grained the unit testing process in holoviews is supposed to be so haven't added one, but let me know if you'd like me to do this.

HoloViews should have far more unit-testing something that has been hampered by the fact that there really are only two core developers and we're both still in get stuff done mode for our PhDs. In future we'll probably have some period of consolidating everything by 1) improving the testing process, 2) adding **A LOT** more unit tests and 3) improving documentation. We'd be very happy to see contributions with unit tests and at some point we'll make included unit tests mandatory.

philippjfr:
The tests indicate that only one example in the Elements tutorial and various examples in Pandas Conversions need to be updated. If you could explain the new ``scaling_method`` options in the Elements Tutorial and find some sensible settings for the examples in Pandas Conversions I'd be happy to merge and will update the tests myself.

maxalbert:
Ok, I have updated the tutorials. Let me know if this is satisfactory or whether you'd like me to tweak anything else. Thanks for taking care of the tests!

maxalbert:
By the way, I was wondering if there is an easy way to automatically pre-scale the points so that the by default the maximum point size is a certain fraction of the figure width (say, 5-10%). This would ensure that a lot of the time the user wouldn't need to apply a custom scaling factor because the points automatically have a reasonable size, and it would ensure that scaling factors would be close to 1 (I guess that values between 0.1 and 10 would be sufficient in almost all cases). This would make it much easier for the user to guess a suitable value and reduce the need for trial and error. Do you think this would be a good idea?

philippjfr:
>By the way, I was wondering if there is an easy way to automatically pre-scale the points so that the by default the maximum point size is a certain fraction of the figure width (say, 5-10%). This would ensure that a lot of the time the user wouldn't need to apply a custom scaling factor because the points automatically have a reasonable size, and it would ensure that scaling factors would be close to 1 (I guess that values between 0.1 and 10 would be sufficient in almost all cases). This would make it much easier for the user to guess a suitable value and reduce the need for trial and error. Do you think this would be a good idea?

I think it would a good idea although it would potentially be quite fiddly to get it to behave well in all cases. If it's something you'd want to tackle we'd certainly accept a PR.

maxalbert:
Thanks! I may have a play if I find the time.

Regarding the original PR, please hold off with merging and fixing the tests until I have double-checked it. I think my current implementation is slightly wrong because matplotlib scales point sizes by area so I probably need to reverse the logic in `util.compute_sizes()`.

philippjfr:
In the issue referenced above (#397) we noticed that point sizes differ between matplotlib and bokeh. I assume this is because matplotlib treats them as an area, while bokeh treats it as the size in screen units. Would be great if you could look into that as part of this PR.

maxalbert:
Yes, good point. I had planned to look into this anyway because I had seen the same discrepancy between matplotlib and bokeh in a different project before, but thanks for the heads-up.

philippjfr:
I forked your branch and created https://github.com/ioam/holoviews/pull/447 with the remaining fixes needed to get the right behavior. If you see any issues with the changes please let me know and please open a new PR if you feel like tackling scaling of sizes as a fraction of the screen size, but for now I'll close this PR.

PR #392, opened 2016-01-03
New GeoMap Element for plotting geographical maps

maxalbert:
This PR is only intended as a basis for discussion and certainly not ready to be merged.

I'm interested in plotting geographical maps using Holoviews. This PR is a quick-and-dirty implementation to get some basic functionality working quickly. It introduces a new element called `GeoMap` which can be used as follows.
```python
import holoviews as hv
from geopandas import GeoDataFrame
hv.notebook_extension()
```
```python
!wget https://raw.githubusercontent.com/MinnPost/simple-map-d3/master/example-data/world-population.geo.json
```
```python
df = GeoDataFrame.from_file('./world-population.geo.json')
```
```python
%%opts GeoMap [fig_inches=(14, None)] (linewidth=0.5)
hv.GeoMap(df)
```
map_of_world

The implementation is rather crude and I'm sure I'm doing a lot of things wrong or inefficiently. Also, there are certainly missing features. One of the key things I'd like to see is the ability to colour the regions according to the values in some column of the given GeoDataFrame (e.g. `POP2005` in this case to show population sizes), which I haven't figured out how to do yet although I'd imagine that if I subclassed the correct Holoviews element then it should just work (perhaps with minor tweaks). In the future I could also imagine support for different projection systems, adding layers from Google Maps/OpenStreetMap, etc.

I gather that @philippjfr has already done some work regarding geographical maps, so I didn't want to take this much further without feedback and opinions whether this is going in the right direction. Any comments, suggestions and criticism would be greatly appreciated. Thanks!

### DIFF
--- a/holoviews/plotting/mpl/path.py
+++ b/holoviews/plotting/mpl/path.py
@@ -116,3 +116,6 @@ class GeoMapPlot(PolygonPlot):
     aspect = param.Parameter(default='equal', doc="""
         GeoMap elements use an 'equal' aspect ratio by default but
         may be set to an explicit aspect ratio or to 'square'.""")
+
+    bgcolor = param.Parameter(default='white', doc="""
+        Background color of the map.""")
--- a/holoviews/plotting/mpl/path.py
+++ b/holoviews/plotting/mpl/path.py
@@ -103,3 +103,16 @@ class PolygonPlot(ColorbarPlot):
         elif value is not None and np.isfinite(value):
             collection.set_array(np.array([value]*len(element.data)))
             collection.set_clim(ranges[vdim.name])
+
+
+class GeoMapPlot(PolygonPlot):
+    """
+    GeoMapPlot draws paths representing the region boundaries in the
+    supplied GeoMap. If the GeoMap has an associated value the color of
+    regions will be drawn from the supplied cmap, otherwise the supplied
+    facecolor will apply.
+    """
+
+    aspect = param.Parameter(default='equal', doc="""
+        GeoMap elements use an 'equal' aspect ratio by default but
+        may be set to an explicit aspect ratio or to 'square'.""")
--- a/holoviews/plotting/mpl/path.py
+++ b/holoviews/plotting/mpl/path.py
@@ -119,3 +119,29 @@ class GeoMapPlot(PolygonPlot):
 
     bgcolor = param.Parameter(default='white', doc="""
         Background color of the map.""")
+
+    frame = param.Boolean(default=False, doc="""
+        Whether to show the frame around the plot.""")
+
+    def initialize_plot(self, ranges=None):
+        # TODO: Most of this is copied & pasted from PolygonPlot.
+        #       How can we avoid the code duplication?!?
+        element = self.hmap.last
+        key = self.keys[-1]
+        axis = self.handles['axis']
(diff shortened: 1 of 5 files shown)


philippjfr:
Thanks for putting together a concrete example. You're right that I developed a prototype to plot geographic data using HoloViews and matplotlib's basemap. The thing is I'm no longer sure that's the right approach. Using something more lightweight like cartopy and/or geopandas seems is probably the better approach. I won't be working on this over the next two weeks but support for geographic data is actually on our roadmap for late January/early February as we'll be using it for a project we are working on.

We'd also be really happy if you want to collaborate on this. From implementing prototypes for both matplotlib and bokeh I can offer some initial suggestions on how I see this working.

1) We will probably provide a GeoProjection type which allows defining the type of projection and should always be defined as the bottom layer in an Overlay. The creation of the appropriate axis/projection will be delegated to the bottom layer and all other plots simply draw onto that projection. I think we can get away with one type with a ``projection`` parameter defining the projection type to be used. For bokeh we should probably have distinct types that load data from various data sources, e.g. Google Maps or a bokeh tile server.

2) Things like colored polygons corresponding to countries, states etc., i.e. a chloropleth map, should simply be done using standard Element types, i.e. Polygons. For this we should improve the format of the path collection types, like Paths, Contours and Polygons. I've just made a new suggestion on how to do that in #347.

holoviews/plotting/mpl/path.py
```diff
@@ -103,3 +103,45 @@ def update_handles(self, axis, element, key, ranges=None):
...
+
+    bgcolor = param.Parameter(default='white', doc="""
+        Background color of the map.""")
+
+    frame = param.Boolean(default=False, doc="""
+        Whether to show the frame around the plot.""")
+
+    def initialize_plot(self, ranges=None):
+        # TODO: Most of this is copied & pasted from PolygonPlot.
+        #       How can we avoid the code duplication?!?
```
philippjfr: I'm planning a refactor of matplotlib Element plots, which would eliminate most of the duplication. For now this is fine, I'll simply clean it up when that happens.

maxalbert:
Thanks, that all sounds great. In principle I'd be very happy to collaborate on this but to be honest I anticipate having virtually no capacity to do so over the next few months. :-/ I'm also not a GIS person so don't have any knowledge how to deal properly with projections etc. It sounds like it will be best to let you take the lead on this but I'm happy to offer suggestions if I can, and to cross-check the design against my use cases (although currently I don't see them getting much more complicated than simple chloropleth plots).

Feel free to close this PR for now as it sounds like you have a much more well-thought-out approach in the pipeline.

jbednar:
FYI, the new version of Bokeh (0.11, due for release tomorrow), includes extensive and very solid support for tiled geographic maps, overlaying them in any combination with each other and with other datasets.  It essentially gives you a Google Maps interface that you can customize arbitrarily, with full interactivity.  I'm not sure how easily it can be integrated with HoloViews, but it's worth checking out once it gets released.

philippjfr:
I'll close this PR now. We have now created the https://github.com/ioam/geoviews/ package, which is based on Cartopy projections and provides a more appropriate place for geo specific element types. We'd be very happy to see a PR to support drawing chloropleths there though.

PR #393, opened 2016-01-05
Bokeh 0.11 Compatibility

philippjfr:
This PR adds various fixes to ensure HoloViews is compatible with the new 0.11 bokeh release.

It is now ready for merge.

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -15,6 +15,7 @@ except ImportError:
 import param
 
 from ...core.data import ArrayColumns
+from .util import models_to_json
 
 
 class Callback(param.ParameterizedFunction):
@@ -156,16 +157,7 @@ class Callback(param.ParameterizedFunction):
         if not old_bokeh:
             plot = self.plot[0]
             data['root'] = plot.state._id
-
-        for obj in objects:
-            if old_bokeh:
-                json = obj.vm_serialize(changed_only=True)
-            else:
-                json = obj.to_json(False)
-            json.pop('renderers', None)
-            data['data'].append({'id': obj.ref['id'], 'type': obj.ref['type'],
-                                 'data': json})
-        return serialize_json(data)
+        return serialize_json(models_to_json(objects))
 
 
 
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -6,11 +6,11 @@ from bokeh.models import CustomJS, TapTool, ColumnDataSource
 try:
     from bokeh.models import PlotObject
     from bokeh.protocol import serialize_json
-    old_bokeh = True
+    bokeh_lt_011 = True
 except ImportError:
     from bokeh.models import Component as PlotObject
     from bokeh.core.json_encoder import serialize_json
-    old_bokeh = False
+    bokeh_lt_011 = False
 
 import param
 
@@ -154,7 +154,7 @@ class Callback(param.ParameterizedFunction):
         Serializes any Bokeh plot objects passed to it as a list.
         """
         data = dict(data=[])
-        if not old_bokeh:
+        if not bokeh_lt_011:
             plot = self.plot[0]
             data['root'] = plot.state._id
         return serialize_json(models_to_json(objects))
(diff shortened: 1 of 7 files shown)


holoviews/plotting/bokeh/widgets.py
```diff
@@ -1,8 +1,20 @@
...
+from distutils.version import LooseVersion
+
 import param
+import bokeh
+from bokeh.io import push_notebook, Document
+
+if LooseVersion(bokeh.__version__) >= LooseVersion('0.11'):
+    old_bokeh = False
+    from bokeh.io import _CommsHandle
+    from bokeh.util.notebook import get_comms
```
jlstevens: How much can we trust that ``_CommsHandle`` is something we can rely on in future? From the name, it doesn't look like something we are expected to use directly...

holoviews/plotting/bokeh/renderer.py
```diff
@@ -56,23 +59,9 @@ def __call__(self, obj, fmt=None):
...
-            json_data = []
-            for plotobj in plotobjects:
-                if plotobj.ref['id'] in ids:
-                    continue
-                else:
-                    ids.append(plotobj.ref['id'])
-                if old_bokeh:
-                    json = plotobj.vm_serialize(changed_only=True)
-                else:
-                    json = plotobj.to_json(False)
```
jlstevens: If you find this pattern happening often, I would consider using a utility what does the version check and smooths over the differences between bokeh versions. I'll let you decide if this this worth it - if you only need it in this one instance, it probably doesn't need to be made into a utility right now.

holoviews/plotting/bokeh/widgets.py
philippjfr: Certainly possible that they will change it but there's no way around it. Bokeh notebook support is currently set up to only keep one CommsHandle open at the same so we can't rely on their core API. It's certainly worth keeping in mind though and replace it with a call to a higher-level API once/if that becomes possible.

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -9,12 +9,13 @@
     old_bokeh = True
```
jlstevens: ``old_bokeh`` isn't a great name as one day the latest version will become old! I would consider stating the version explicitly...i.e ``bokeh_XX``.

jlstevens:
I've made some comments above but I'm afraid I don't know enough details to do a very thorough review. I do have one general comment though...

The changes here seem to be of two types 1. actual changes in the bokeh API to smooth over 2. improvements to how the code works that is compatible across both bokeh versions (i.e understanding how to leverage bokeh better).

The second type of change seems general and I assume fairly stable. The first type of change is a lot of special cases and I worry that as bokeh keeps developing (and the API keeps getting tweaked) that the list of these fixes will keep growing. I'm not sure it is needed *now* but I might consider a class in util or even a file ``compat.py`` if we think the number of version dependent fixes will keep expanding in future.

holoviews/plotting/bokeh/widgets.py
jbednar: It's worth posting an issue on the Bokeh site asking whether this is an ok long-term solution, and/or request a better one be made available.

holoviews/plotting/bokeh/callbacks.py
jbednar: +1

holoviews/plotting/bokeh/widgets.py
philippjfr: I know that handling multiple Comms at once is on their radar and might even land in 0.11.1. If need be I can mirror CommsHandle in HoloViews, it's a tiny class that's literally just meant to hold a few attributes.

holoviews/plotting/bokeh/renderer.py
philippjfr: I think I ever repeat the same pattern multiple times.

jbednar: even, or never?

philippjfr:
Agree that we'll have to add a compat.py module if there are a lot more changes to the API coming. It could simply supply various version agnostic functions to do specific things. Currently it doesn't seem worth it though.

jlstevens:
Tests are all passing so I'll merge now.

PR #398, opened 2016-01-08
Plotly

philippjfr:
While nowhere near ready for merge I wanted to open this PR to document the progress on the backend and to discuss at what point we'll consider it "ready".

As of right now this PR lays most of the groundwork for this backend, i.e. it already supports displaying Elements, Overlays, Layouts and HoloMaps. It also already supports a wide range of Elements.

Major missing features:

- [x] Support for AdjointLayouts
- [x] Support for GridSpace
- [x] Add style options to Element plots
- [x] Issues in updating 3D plots in a HoloMap
- [x] Add experimental warning
- [x] Add unit tests

Here is my [WIP notebook](http://philippjfr.com/work/work-in-progress/plotly/)

Future work:

- Missing Elements: Spread, Spikes, Path, Polygons, Box, Arrow, RGB, HSV, Bounds, VLine, HLine, Contours, Ellipse, Table
- Add support for ``*rotation``, ``*ticks``, ``show_legend``, ``show_grid`` plot options


jbednar:
Is there a version of the Elements tutorial that would demonstrate (a) the current capabilities for the Plotly backend, and (b) what the behavior is for the things that don't work?  Ideally the same unmodified tutorial could be run on all backends, with some reasonable fallback behavior in all cases that are not properly supported.  Fallbacks could be dropping down to another backend for unsupported items, or even just printing a warning and moving on, as long as we could run the same code on all backends without aborting...

jlstevens:
One simple mechanism we currently have is a ``SkipRendering`` exception that falls back to the text repr (which will always work, even if no plotting libraries are installed). For this reason, I prefer this approach to falling back to some other plotting backend which may or may not be installed (and as a result, would result in inconsistent output that depends on the user's environment).

I agree that we want the same notebook to be runnable between backends, even if not everything is fully implemented. It would be possible to reraise all plotting exceptions (except for things like style option errors) as ``SkipRendering`` which would ensure notebooks should always be runnable even if the output isn't what is desired.

I'm not saying we should do this, as it  would make debugging even more difficult, masking errors and tracebacks even further.

philippjfr:
Warning looks like this and gets displayed when activating the notebook extension:

image

jbednar:
Looks good.  How do we suppress that when we need to?

philippjfr:
This would do it:

```python
hv.plotting.plotly.renderer.PlotlyRenderer._loaded = True
```

jlstevens:
The warning looks good... though are we sure we would notice if someone e-mailed holoviews@gmail.com? Personally, I don't check that account often....

jbednar:
If that's not an account that's checked often, presumably it should be forwarded to one that is?  You can forward it to my continuum email if you like, though you're more likely to be the ones making the decisions about any such offer.

philippjfr:
>The warning looks good... though are we sure we would notice if someone e-mailed holoviews@gmail.com? Personally, I don't check that account often....

Maybe we should set up a mail server for holoviews.org, for now I do monitor the gmail account daily.

jlstevens:
I'm happy to merge (the one test failure is due to a transient) but first, I feel it is worth making our stance regarding plotly clear:

* Maintaining matplotlib and bokeh backends keeps us plenty busy as it is!
* We have designed HoloViews to be modular - we want to allow backends to be added without having to introduce the code to the main repository.
* Plotly support could therefore have been a separate project that people could optionally install.

In addition, both bokeh and plotly overlap a lot in terms of what they offer - fast, interactive visualizations in the browser. This brings me to the main reason I am happy to merge this PR at this time: bokeh currently does not support 3D visualizations while plotly does.

Hopefully, the warning will help make our stance with respect to plotly clear. In addition, if there is interest in the plotly backend from users/contributors, I would urge us to separate the plotly backend support from the main repository regardless. I would also want to separate out the plotly support if bokeh fills in the missing functionality i.e by supporting 3D elements.

In short, if we find ourselves spending time maintaining this code, then it should be split out into a separate repository.

Merging.

Issue #396, opened 2016-01-07
Twin axis


vascotenner:
Is it possible to create overlays with two y-axis (a left and a right axis)? With matplotlib this is possible with ax1.twinx().

Twin axes are currently not supported bit yet. For now you could try to write a finalize_hook to do it.

vascotenner:
I have a really uggly solution:
```
%%opts Curve [finalize_hooks=[setglobal_hook], show_frame=True]
a = hv.Curve(([1,2], [3,21]), vdims=['A'])
b = hv.Curve(([1,2], [30,20]), vdims=['B'])

def setglobal_hook(plot, element):
    global fig, el
    fig = plot
    el = element

a
```

```
%%opts Curve [finalize_hooks=[overlay], show_frame=True]
b = hv.Curve(([1,2], [30,20]), vdims=['B'])

def overlay(plot, element):
    ax = plot.handles['axis']
    ax2 = ax.twinx()
    ax2.set_yticklabels([t.get_text() for t in fig.handles['axis'].get_yticklabels()])
    ax2.set_ylabel(fig.handles['axis'].get_ylabel())
    ax2.plot(*fig.handles['axis'].lines[0].get_data())
    ax2.lines[-1].set_color('r')
    
b
```

Is there a better way to get grips to the plotted version of a?

philippjfr:
I see, yes it's problematic because you have to create a new axis to plot on. I'll have to think about it a bit, one quick workaround to add support not just for ``finalize_hooks`` but also ``initialize_hooks``, that way you could simply replace the axis on the plot ensuring that the Element is drawn on the correct axis.

vascotenner:
A slightly cleaner solution:
```
def overlay(first, plot, element):
    fig = hv.Store.renderers['matplotlib'].get_plot(first)
    ax = plot.handles['axis']
    ax2 = ax.twinx()
    ax2.set_yticks(fig.handles['axis'].get_yticks())
    #ax2.set_yticklabels([t.get_text() for t in fig.handles['axis'].get_yticklabels()])
    ax2.set_ylabel(fig.handles['axis'].get_ylabel())
    ax2.plot(*fig.handles['axis'].lines[0].get_data())
    ax2.lines[-1].set_color('r')

def tmp(plot, element):
    overlay(a, plot, element)
```

```
%%opts Curve [finalize_hooks=[tmp], show_frame=True]
b = hv.Curve(([1,2], [30,20]), vdims=['B'])

b
```

[...]

PR #400, opened 2016-01-08
Added initial_hook and renamed finalize_hook to final_hook

philippjfr:
This PR implements the suggestion made in #396 to add an ``initial_hook`` and rename ``finalize_hook`` to ``final_hook``.

### DIFF
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -115,6 +115,12 @@ class ElementPlot(GenericElementPlot, MPLPlot):
         if isinstance(check, Element3D):
             self.projection = '3d'
 
+        for hook in self.initial_hooks:
+            try:
+                hook(self, element)
+            except Exception as e:
+                self.warning("Plotting hook %r could not be applied:\n\n %s" % (hook, e))
+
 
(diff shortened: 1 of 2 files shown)


jlstevens:
Looks good to me! 

Not much here to review and I assume this will be ready for an immediate merge...

philippjfr:
Fixing some docstrings, ready once those are pushed. Hopefully the Options Tutorial won't complain.

jbednar:
You can't just do self.finalize_hooks = self.final_hooks?

philippjfr:
Could do but shouldn't we warn users that they should switch to ``self.final_hooks``?

jbednar:
The original suggestion from Jean-Luc was "having an alias and eventually deprecating finalize_hooks", which I think is the right way to go about it.  I.e., immediately support the new way while making the old way continue to work, and then *later* deprecate the old way, in the following release.  Otherwise users have no time to update their code before they start seeing warnings, and thus they have to choose only a single HoloViews version for their code to support cleanly, which I think is unduly restrictive.

philippjfr:
I just removed the deprecation warning, now only warning when finalize_hooks are being overridden by final_hooks.

jlstevens:
Ok, tests are passing and looks ready to me. Merging.

PR #401, opened 2016-01-08
Doc cleanup

jbednar:
Partial cleanup, ready for comments (and merging if appropriate) but not yet including edits to the main tutorials.  It would be very good if Philipp or Jean-Luc could inspect my changes carefully, looking especially at:

1. I changed the tagline in the main site to match the one on Github.
2. Is the tone and level of detail right on Homepage.ipynb?  I tried to make it more explicit about the difference in philosophy between HoloViews and other libraries, but it's always hard to explain that.
3. I added links to the new Bokeh tutorials from the Tutorials index, and references to Bokeh throughout, but there are probably other places to mention it and guide people towards which backend is appropriate for their purposes.
4. Is the Pandas tutorial out of date?  It talks about converting Pandas dataframes, but I thought we were now supporting using them as-is, not copying or converting. Does this need to be clarified somewhere?
5. Can we simplify the GitHub README.rst?  Right now it has a redundant copy of features.rst, which I vote should just be replaced with a link, but in any case at least needs to be updated from features.rst.
6. There were a few locations, e.g. in the FAQ, where I wanted to document something (e.g. the aliases system), but the only documentation I could find was in a pull request, so I put in a link to the pull request.  We can either leave in such links as a pragmatic measure, or pull the information out of the pull request into something we can maintain over time.
7. I added "bokeh" to the conda install command, so that the result would be more like what happens for matplotlib (things just work).  But the right approach is presumably to make the conda package be completely independent of matplotlib (with the default backend falling back to bokeh when mpl is not installed), and then tell people to choose whichever command they wish:
  * conda install holoviews                            # no plotting backend 
  * conda install holoviews matplotlib            # default backend is mpl
  * conda install holoviews bokeh                 # default backend is bokeh
  * conda install holoviews matplotlib bokeh # default backend is mpl
8. I've run the tutorials, but don't know how to build the web site locally (which is a big pain and really discourages updating the web site!), so I can't tell if everything's still formatted properly.
9. It needs links to the SciPy paper in appropriate locations, because that has more philosophical discussion that makes good background reading.
10.  I still need to go through the other tutorials, but won't be able to do that today or tomorrow.

### DIFF
--- a/doc/FAQ.rst
+++ b/doc/FAQ.rst
@@ -5,9 +5,9 @@ Here is a list of questions we have either been asked by users or
 potential pitfalls we hope to help users avoid:
 
 
-**Q: Can I use HoloViews without IPython?**
+**Q: Can I use HoloViews without IPython/Jupyter?**
 
-**A:** Yes! The IPython support makes a lot of tasks easier, and
+**A:** Yes! The IPython/Jupyter notebook support makes a lot of tasks easier, and
 helps keep your data objects separate from the customization options,
 but everything available in IPython can also be done directly from
 Python.  For instance, in HoloViews 1.3.0 you can render an object 
@@ -20,28 +20,34 @@ directly to disk, with custom options, like this:
   renderer.save(my_object, 'example_I', style=dict(Image={'cmap':'jet'}))
 
 This process is described in detail in the 
-`Options tutorial <Tutorials/Options>`_, and some more information is 
-on `this wiki page
-<https://github.com/ioam/holoviews/wiki/HoloViews-without-IPython>`_.
+`Options tutorial <Tutorials/Options>`_.
 Of course, notebook-specific functionality like capturing the data in
-notebook cells or saving cleared notebooks is only for IPython.
+notebook cells or saving cleared notebooks is only for IPython/Jupyter.
 
 **Q: How should I use HoloViews as a short qualified import?**
 
-For a qualified import, we recommend importing HoloViews using ``import holoviews as hv``.
+We recommend importing HoloViews using ``import holoviews as hv``.
 
 **Q: My output looks different from what is shown on the website**
 
-**A:** Matplotlib supports a variety of plotting backends, and these
-can have inconsistent output. HoloViews will not switch the backend
-for you, but we strongly recommend selecting the 'agg' backend in
-general:
+**A:** HoloViews is organized as data structures that have
+corresponding plotting code implemented in different plotting-library
+backends, and each library will have differences in behavior.
+Moreover, the same library can give different results depending on its
+own internal options and versions.  For instance, Matplotlib supports
+a variety of internal plotting backends, and these can have
+inconsistent output. HoloViews will not switch Matplotlib backends for
+you, but when using Matplotlib we strongly recommend selecting the
+'agg' backend for consistency:
 
 .. code:: python
 
-  from matplotlib import pyplot as plt
-  plt.switch_backend('agg')
+  from matplotlib import pyplot
+  pyplot.switch_backend('agg')
 
+You can generally set options explicitly to make the output more
+consistent across HoloViews backends, but in general HoloViews tries
+to use each backend's defaults where possible.
 
 **Q: Help! I don't know how to index into my object!**
 
@@ -49,7 +55,7 @@ general:
 ``%%output fig='repr' holomap='repr'`` at the top of your code cell to
 see the structure of your object.
 
-In a regular Python session, you can look at ``print repr(obj)``. For
+In any Python session, you can look at ``print repr(obj)``. For
 an explanation of how this information helps you index into your
 object, see our `Composing Data tutorial <Tutorials/Composing_Data>`_.
 
@@ -57,11 +63,11 @@ object, see our `Composing Data tutorial <Tutorials/Composing_Data>`_.
 **Q: Help! How do I find out the options for customizing the
 appearance of my object?**
 
-**A:** If you are in the IPython Notebook you can use the cell magic
+**A:** If you are in the IPython/Jupyter Notebook you can use the cell magic
 ``%%output info=True`` at the top of your code cell. This will
 present the available style and plotting options for that object.
 
-The same information is also available using
+The same information is also available in any Python session using
 ``holoviews.help(obj)``. For more
 information on customizing the display of an object,
 see our `Options Tutorial <Tutorials/Options>`_.
@@ -72,7 +78,7 @@ through to matplotlib?**
 
 **A:** We have selected a subset of default allowable style options
 that are most commonly useful in order to hide the more arcane
-matplotlib options. If you do need to such an option to be passed to
+matplotlib options. If you do need such an option to be passed to
 the plotting system, you are welcome to declare that this is allowed.
 For instance, say you may want the ``'filternorm'`` option to be passed
 to matplotlib's ``imshow`` command when displaying an ``Image``
@@ -88,23 +94,26 @@ magic, including tab-completion!
 
 **Q: I still can't tweak my figure in exactly the way I want. What can I do?**
 
-HoloViews is designed to be as flexible as possible and should always
-support all the common visualization options. We intend to add a new 
-tutorial soon to explain how you can continue to 
-`tweak and extend HoloViews <https://github.com/ioam/holoviews/issues/19>`_.
-You can of course subclass any HoloViews object and modify 
-any of its behavior, but this tutorial will explain some simpler 
-and more maintainable alternatives once it is ready.
+The parameters provided by HoloViews should normally cover the most
+common plotting options needed.  In case you need further control, you
+can always subclass any HoloViews object and modify any of its
+behavior, and the object will still normally interact with other
+HoloViews objects (e.g. in Layout or Overlay configurations).  
 
-**Q: How do I get a legend on my figure?**
+There are also much simpler and more easily maintainable options for
+making customized modifications, and we expect to put together a new
+tutorial to explain the methods for `tweaking and extending HoloViews
+<https://github.com/ioam/holoviews/issues/19>`_.
+
+**Q: How do I get a legend on my overlay figure?**
 
 **A:** Legends are generated in two different ways, depending on the
-``Overlay`` type you are using. When using ``*`` to generate an ``Overlay``,
+``Overlay`` type you are using. When using ``*`` to generate a normal ``Overlay``,
 the legends are generated from the labels of the Elements.
 Alternatively, you can construct an ``NdOverlay``, where the key dimensions
-and values will become part of the legend. An example of an ``NdOverlay``
-in action may be `viewed here <Tutorials/Containers.html#NdOverlay>`_.
-
+and values will become part of the legend. The 
+`Containers tutorial <Tutorials/Containers.html#NdOverlay>`_.
+shows an example of an ``NdOverlay`` in action.
 
 **Q: I wish to use special characters in my title, but then attribute
 access becomes confusing.**
@@ -116,6 +125,20 @@ that let you refer to the object easily in the code, and then you can
 set the plot title directly, using the plot option
 ``title_format="my new title"``.
 
+For frequent cases you can define a dictionary of aliases, with short,
+Pythonic names for complicated typeset strings:
+
+```
+al = hv.util.Aliases(Spectrum='Frequency spectrum', 
+                     Water='$H_2O$',
+                     Glucose='$C_6H_{12}O_6$')
+
+(hv.Image(np.random.rand(10,10), group=al.Spectrum, label=al.Glucose) +
+ hv.Image(np.random.rand(10,10), group=al.Spectrum, label=al.Water))
+```
+
+See this `pull request <https://github.com/ioam/holoviews/pull/312>`_
+for more details.
 
 **Q: Where have my custom styles gone after unpickling my object?**
 
@@ -132,10 +155,12 @@ customization. You can import ``Store`` from the main namespace with
 my notebook?**
 
 **A:** It is very easy to visualize large volumes of data with
-HoloViews, and all available display data is embedded in the HTML
-snapshot when sliders are used. It is therefore worth being aware of
-file size when authoring a notebook to be made make public on the
-web. Useful tricks to reduce file size include:
+HoloMaps, and all available display data is embedded in the HTML
+snapshot when sliders are used so that the result can be viewed
+without using a Python server process. It is therefore worth being
+aware of file size when authoring a notebook or web page to be
+published on the web. Useful tricks to reduce file size of HoloMaps
+include:
 
 * Reducing the figure size.
 * Selecting fewer frames for display (e.g selecting a smaller number
@@ -144,8 +169,17 @@ web. Useful tricks to reduce file size include:
   ``webm``, ``mp4`` or animated ``gif``, while being aware that those
   formats may introduce visible artifacts.
 
+It is also possible to generate web pages that do not actually include
+all of the data shown, by specifying a `DynamicMap
+<https://github.com/ioam/holoviews/blob/babac9e2916384e03cd189865d5ca59a6b490708/holoviews/core/spaces.py#L332>`_
+rather than a HoloMap.  The DynamicMap will request data only as
+needed, and so requires a Python server to be running alongside the
+viewable web page.  Such pages are more difficult to share by email or
+on web sites, but much more feasible for large datasets.
+
+  
 **Q: How do I create a Layout or Overlay object from an arbitrary list?**
 
-You can supply the list of elements directly to the ``Layout`` and
+You can supply a list of ``elements`` directly to the ``Layout`` and
 ``Overlay`` constructors. For instance, you can use
 ``hv.Layout(elements)`` or ``hv.Overlay(elements)``.
--- a/doc/Homepage.ipynb
+++ b/doc/Homepage.ipynb
@@ -4,7 +4,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "HoloViews is a [Python](http://python.org) library that makes analyzing and visualizing scientific or engineering data much simpler, more intuitive, and more reproducible.  For instance, if you have a two-dimensional dataset like the fractal below ([mandelbrot.npy](https://github.com/ioam/holoviews/raw/master/doc/mandelbrot.npy)), it only takes one line of HoloViews code to mark it with a horizontal line indicating a cross section, add its histogram, sample to get the cross section, and display the result:"
+    "HoloViews is a [Python](http://python.org) library that makes analyzing and visualizing scientific or engineering data much simpler, more intuitive, and more easily reproducible.  Without HoloViews, there are typically many steps required before you can see your data, whether you use a GUI to build up a plot interactively or you write plotting code in Python to put together that specific type of plot.  HoloViews instead lets you store your data in an annotated format that is instantly visualizable, with immediate access to both the numeric data *and* its visualization.  For instance, if you wrap a two-dimensional dataset like the fractal below ([mandelbrot.npy](https://github.com/ioam/holoviews/raw/master/doc/mandelbrot.npy)) in a HoloViews ``Image`` object, you can view the data as an image (here annotated with a horizontal line), its histogram, and a slice of it at the indicated cross-section, without writing any plotting code:"
    ]
   },
   {
@@ -17,7 +17,7 @@
    "source": [
     "import numpy as np\n",
     "import holoviews as hv\n",
-    "%reload_ext holoviews.ipython\n",
+    "hv.notebook_extension('matplotlib')\n",
     "fractal = hv.Image(np.load('mandelbrot.npy'))\n",
     "\n",
     "((fractal * hv.HLine(y=0)).hist() + fractal.sample(y=0))"
@@ -27,7 +27,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "To use HoloViews, you first wrap your data in a HoloViews component along with optional metadata describing it.  It will then display itself automatically on its own or in combination with any other HoloViews component.  The separate [matplotlib](http://matplotlib.org) library does the plotting, but none of the data structures depend on the plotting code, so that you can easily create, save, load, and manipulate HoloViews objects from within your own programs.  HoloViews objects support arbitrary combination, selection, slicing, sorting, sampling, and animation, to allow you to focus on whatever aspect of your data you wish. Instead of writing or maintaining complex plotting code, just declare what data you want to see, and HoloViews will handle the rest."
+    "The original data always remains available in its native format (accessible via ``fractal.data`` in this case), but accessing it via the HoloViews object instead lets the data display itself, either alone (just type ``fractal`` in an IPython/Jupyter notebook) or alongside or overlaid with other HoloViews objects as shown above. The actual plotting is done using a separate library like [matplotlib](http://matplotlib.org) or [bokeh](http://bokeh.pydata.org), but all of the HoloViews objects can be used without any plotting library available, so that you can easily create, save, load, and manipulate HoloViews objects from within your own programs.  HoloViews objects support arbitrarily high dimensions, using continuous or discrete indexes and values, with flat or hierarchical organizations, and sparse or dense data formats.  The objects can then be flexibly combined, selected, sliced, sorted, sampled, or animated, all by specifying what data you want to see rather than by writing plotting code. The goal is to put the plotting code into the background, as an implementation detail to be written once and reused often, letting you focus clearly on your data in daily work."
    ]
   },
   {
@@ -41,7 +41,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Even extremely complex relationships between data elements can be expressed succinctly in HoloViews:"
+    "Even extremely complex relationships between data elements can be expressed succinctly in HoloViews, allowing you to explore them with ease:"
    ]
   },
   {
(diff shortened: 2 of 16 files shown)


philippjfr:
Thanks for going through it, looks good overall.

> (2) Is the tone and level of detail right on Homepage.ipynb? I tried to make it more explicit about the difference in philosophy between HoloViews and other libraries, but it's always hard to explain that.

Looks fine to me.

> (4) Is the Pandas tutorial out of date? It talks about converting Pandas dataframes, but I thought we were now supporting using them as-is, not copying or converting. Does this need to be clarified somewhere?

It is yes, large parts of it can be reused as material for either a Transforming Data or Columns Tutorial.

> (5) Can we simplify the GitHub README.rst? Right now it has a redundant copy of features.rst, which I vote should just be replaced with a link, but in any case at least needs to be updated from features.rst.

Not sure, it's nice to have a concrete list of features on the GitHub landing page especially since it actually receives more unique visitors than the website.

> (6) There were a few locations, e.g. in the FAQ, where I wanted to document something (e.g. the aliases system), but the only documentation I could find was in a pull request, so I put in a link to the pull request. We can either leave in such links as a pragmatic measure, or pull the information out of the pull request into something we can maintain over time.

We should open an issue to collect items that should get some documentation so we can prioritize.

> (7) I added "bokeh" to the conda install command, so that the result would be more like what happens for matplotlib (things just work). But the right approach is presumably to make the conda package be completely independent of matplotlib

It shouldn't be too hard to drop the matplotlib dependency for the bokeh backend and fallbacks shouldn't be too hard either. Of course any composite object will continue to simply warn about any unsupported Element types.

> (8) I've run the tutorials, but don't know how to build the web site locally (which is a big pain and really discourages updating the web site!), so I can't tell if everything's still formatted properly.

This should do it:

```
cd doc
conda install sphinx, runipy
make ipynb-doc
```

Problem the tests require very specific versions of freetype and matplotlib and will only work on Linux so it'll just abort. Should add an option to build the docs without tests.

jbednar:
Travis says https://travis-ci.org/ioam/holoviews/jobs/101181829 failed with display output mismatch, but I can't see what the output difference is...

jbednar:
BTW, the Introduction tutorial shows that a Numpy array will be preserved inside a HoloViews object; can the same be shown for a Pandas dataframe or other input types?  If so we should say that in that location, and show it somewhere (there if appropriate, or elsewhere).

philippjfr:
>BTW, the Introduction tutorial shows that a Numpy array will be preserved inside a HoloViews object; can the same be shown for a Pandas dataframe or other input types? If so we should say that in that location, and show it somewhere (there if appropriate, or elsewhere).

Should presumably go in the notebook about the Columns interface, which I'll adapt from the existing Pandas Conversion notebook.

philippjfr:
>Travis says https://travis-ci.org/ioam/holoviews/jobs/101181829 failed with display output mismatch, but I can't see what the output difference is...

The display tests for two cells in the Elements tutorial are still flakey, if you see a failure in cell 43 or 46 just restart the build.

jbednar:
Ok, I managed to cover two more of the tutorials; still lots to go, but the intro ones are the most important and have the most changes...

jbednar:
Oh, and I had some changes to About (and maybe others?) stashed away; any idea how to inspect the stash manually to cut and paste those out of there?  If not I'll have to recreate them...

philippjfr:
You can inspect the last stash with:

```
git stash show -p stash@{0}
```

I would seriously recommend getting magit for emacs though it makes working with git so much easier.

jlstevens:
Yes! I highly recommend magit!

Anyway, thanks for doing this and now I'll catch up and make some comments.

jlstevens:
It is really great to see a thorough update of the website!

Here are my replies to your queries:

> 1. I changed the tagline in the main site to match the one on Github.

Great!

> 2. Is the tone and level of detail right on Homepage.ipynb? I tried to make it more explicit about the difference in philosophy between HoloViews and other libraries, but it's always hard to explain that.

Yes, the tone is right and I like the updated text. It is a bit wordy but I think that is unavoidable as it quite hard to convey what HoloViews offers with text only.

> 5. Can we simplify the GitHub README.rst? Right now it has a redundant copy of features.rst, which I vote should just be replaced with a link, but in any case at least needs to be updated from features.rst.

I agree with Philipp. I use the README as the basis of the [PyPI page](https://pypi.python.org/pypi/holoviews) (with minor tweaks) and the [Anaconda Page](https://anaconda.org/ioam/holoviews). I also think it is important to have a nice list of features on the GitHub landing page. 

> 6. There were a few locations, e.g. in the FAQ, where I wanted to document something (e.g. the aliases system), but the only documentation I could find was in a pull request, so I put in a link to the pull request. We can either leave in such links as a pragmatic measure, or pull the information out of the pull request into something we can maintain over time.

Linking to PRs is fine and it is tricky to find a place for the miscellaneous odds and ends. The wiki is still a good place to record various tips and tricks as we encounter them but I do think we want a more visible place for newcomers to find them. I can imagine something in the style of the FAQ (which should be fore higher level questions) that is broken down into sections (rendering, styling, exporting etc). Whatever we do, it should be easily searchable.

> 7. I added "bokeh" to the conda install command, so that the result would be more like what happens for matplotlib (things just work). But the right approach is presumably to make the conda package be completely independent of matplotlib.

Are you suggesting the conda package include bokeh as a dependency (already the case I believe) but *not* matplotlib? I do agree the bokeh backend would ideally not require matplotlib but I have no found a good way to do the equivalent of the ``extras_require``  with conda. That is why the conda package includes everything (and conda can install more dependencies a lot quicker than pip anyway).

Anyway, I hope we will be able to shift our attention to improving the website building process now as well as author new tutorials and improve the documentation general. For the time being, I'll have a closer look at the rest of the updated text and make inline comments if I have anything to suggest...

doc/FAQ.rst
```diff
@@ -116,6 +125,20 @@ that let you refer to the object easily in the code, and then you can
...
+Pythonic names for complicated typeset strings:
+
+```
+al = hv.util.Aliases(Spectrum='Frequency spectrum', 
+                     Water='$H_2O$',
+                     Glucose='$C_6H_{12}O_6$')
+
+(hv.Image(np.random.rand(10,10), group=al.Spectrum, label=al.Glucose) +
+ hv.Image(np.random.rand(10,10), group=al.Spectrum, label=al.Water))
+```
```
jlstevens: I think documenting this in the FAQ isn't too bad in the end. It is probably things like a collection of  ``final_hooks`` examples that are hard to find a place for...

doc/FAQ.rst
```diff
@@ -144,8 +169,17 @@ web. Useful tricks to reduce file size include:
...
+rather than a HoloMap.  The DynamicMap will request data only as
+needed, and so requires a Python server to be running alongside the
+viewable web page.  Such pages are more difficult to share by email or
+on web sites, but much more feasible for large datasets.
+
+  
 **Q: How do I create a Layout or Overlay object from an arbitrary list?**
 
-You can supply the list of elements directly to the ``Layout`` and
+You can supply a list of ``elements`` directly to the ``Layout`` and
```
jlstevens: Shouldn't this be ``Elements``? Bit tricky as only the singular, ``Element`` is an actual class name but I think pluralizing it (with a capital) isn't too much of a stretch. To me ``elements`` looks a bit odd as I don't see why it should be formatted like code...

doc/Homepage.ipynb
```diff
@@ -17,7 +17,7 @@
    "source": [
     "import numpy as np\n",
     "import holoviews as hv\n",
-    "%reload_ext holoviews.ipython\n",
+    "hv.notebook_extension('matplotlib')\n",
```
jlstevens: It is nice to put ``'matplotlib'`` in there even it isn't necessary to emphasis we support multiple backends. What would you think about putting a comment here...maybe along the lines of ``# Or 'bokeh' if available``?

doc/Homepage.ipynb
```diff
@@ -27,7 +27,7 @@
...
+    "The original data always remains available in its native format (accessible via ``fractal.data`` in this case), but accessing it via the HoloViews object instead lets the data display itself, either alone (just type ``fractal`` in an IPython/Jupyter notebook) or alongside or overlaid with other HoloViews objects as shown above. The actual plotting is done using a separate library like [matplotlib](http://matplotlib.org) or [bokeh](http://bokeh.pydata.org), but all of the HoloViews objects can be used without any plotting library available, so that you can easily create, save, load, and manipulate HoloViews objects from within your own programs.  HoloViews objects support arbitrarily high dimensions, using continuous or discrete indexes and values, with flat or hierarchical organizations, and sparse or dense data formats.  The objects can then be flexibly combined, selected, sliced, sorted, sampled, or animated, all by specifying what data you want to see rather than by writing plotting code. The goal is to put the plotting code into the background, as an implementation detail to be written once and reused often, letting you focus clearly on your data in daily work."
```
jlstevens: Good to mention bokeh here but I think showing how it is enabled in the comment suggestion above might still be worth considering.

doc/features.rst
```diff
@@ -17,7 +17,7 @@ ________
   
 * Supports a truly reproducible workflow by minimizing the code needed for analysis and visualization.
 * Already used in a variety of research projects, from conception to final publication.
-* All HoloViews objects can be pickled and unpickled.
+* All HoloViews objects can be pickled and unpickled, with no plotting-library dependencies.
```
jlstevens: @philippjfr Is this always true? Would styles pickle/unpickle if a ``Palette`` is used in the style options for instance?

philippjfr: Don't see any reason why it wouldn't work, Palettes are just fancy cycles. There can be some issues pickling matplotlib objects (e.g. a custom ticker) but I think they were addressing that.

doc/latest_news.html
```diff
@@ -4,6 +4,7 @@
       December 22nd 2015: HoloViews 1.4.1 released and now available on 
       <a href="https://pypi.python.org/pypi/holoviews/">PyPI</a> and
       <a href="https://anaconda.org/anaconda/holoviews">Anaconda</a>.
+      Now includes full support for the <a href="http://bokeh.pydata.org">Bokeh</a> plotting library.
```
jlstevens: Not sure if 'full' is the right work as not all elements are supported. 

Maybe 'official' support is better? This means that as developers aim to support bokeh as standard, not that we are necessarily promising anything. :-)

doc/features.rst
jlstevens: Ok. So if you use a matplotlib object in style options, you will obviously have trouble unpickling the style if matplotlib is not available. This is a good reason to avoid using matplotlib objects in this way...

doc/FAQ.rst
jbednar: Just me not trying to get into too much explanation; obviously it needs more.  It's formatted like code to imply that it's a Python list, which I didn't feel was clear before, but if you didn't catch that, then it's *still* not clear.

doc/Homepage.ipynb
jbednar: Thanks for raising that; I meant to list it as an issue.  Yes, the reason I put 'matplotlib' explicitly, here on the homepage is to make it clear that (a) we are a way to access plotting libraries, not competing with them, (b) we support multiple backends, not just one, and (c) to imply that a user could type 'bokeh' here to switch the backend.  I'd rather avoid putting the comment explicitly, to keep the homepage as simple as possible; I'm pretty sure that if a reader looks at the rest of the page and sees that we support matplotlib and Bokeh, they will guess to try 'bokeh' or 'Bokeh' here, and of course they can also link on the Bokeh tutorials to verify that.

doc/features.rst
jbednar: Here I was referring to the objects, not the styles.  If we just use unpickle, we'll only get the objects, right?  And thus won't have to worry about the plotting library dependencies?  The point is to make it clear that the data will always, always be able to be extracted (as long as pickles are supported in general), and thus that it's a safe way to store the data.

doc/FAQ.rst
jlstevens: Ah ok, I didn't get that. The constructor accepts a list of elements but that argument isn't most likely called ``data`` because there are multiple ways it can be used.

doc/latest_news.html
jbednar: Well, what defines the list of "all elements"?  Can't the list of Bokeh elements be treated as the canonical ones, with the extra ones provided by Matplotlib as bonuses?  Seems like it's full support if we say it is. :-)  I'm reluctant to say "official", because that sounds like we have a team of paid support professionals at our disposal!  But whatever way you and Philipp want to say it is fine.  I do want to indicate that the Bokeh support is very different from the Plotly support -- the Plotly backend is *not* full support, right now at least.

doc/features.rst
philippjfr: That's a really good point. Maybe our ``hv.Store.load`` could somehow have a flag to toggle loading of the style options, afaik it will simply skip them if they can't be loaded already.

jbednar:
> >  (7) I added "bokeh" to the conda install command, so that the result would be more like what happens for matplotlib (things just work). But the right approach is presumably to make the conda package be completely independent of matplotlib.
> 
> Are you suggesting the conda package include bokeh as a dependency (already the case I believe) but not matplotlib? I do agree the bokeh backend would ideally not require matplotlib but I have no found a good way to do the equivalent of the extras_require with conda. That is why the conda package includes everything (and conda can install more dependencies a lot quicker than pip anyway).

No, I'm suggesting that neither bokeh *nor* matplotlib should be dependencies of the conda package, and that we ask users to specify whichever one or the other they wish to use, recommending that they install both but allowing them to install neither.

doc/features.rst
jlstevens: I agree that there should never be a case where you can't unpickle the data itself just because a plotting library is missing.

What I wanted to check is that styles (mostly) have the same behavior: although technically styles *aren't* part of the  holoviews elements (so what you say above is true if you replace the word 'object' with 'elements') a user might also expect styles to pickle/unpickle in a way that is agnostic to the plotting library.

jbednar: Such a toggle would be a good idea.

doc/latest_news.html
jlstevens: I don't think the plotly backend is going to be merged any time soon. I would just drop the word 'full' to keep it simple (I agree 'official' commits us to too much) so just 'Now includes support ...'

jlstevens:
> ... that we ask users to specify whichever one or the other they wish to use, recommending that they install both but allowing them to install neither.

I am ok with this and it makes sense to me. It does mean we need the user to install the plotting backend themselves but that makes sense if we are going to support more options (e.g plotly) in future.

I can make this change to the conda package for the next minor release i.e 1.4.2  (which will probably happen in a week or two).

doc/latest_news.html
philippjfr: Well the Plotly support has not been merged so no one should be under any illusion that it has full support. I did try to raise the issue of what it means for a backend to be ready for inclusion in HoloViews in that PR. As you point out in future there will likely not be one single backend that supports all the Elements. If we ever have cross-backend layout capabilities this won't be such an issue but for now people will have to very deliberately use backends for particular purposes. I think that's reasonable as long as it's easy for people to find out about the drawbacks and benefits of particular backends and what is supported. I don't think we should be overly restrictive about what constitutes a supported backend, since some backends might only support a few Element types but draw them very quickly.

jbednar:
Changing the conda package spec should be straightforward, and I don't think the change should affect any existing HoloViews users, because they will have already installed matplotlib previously.  But of course we do need to make sure that the codebase can run in some meaningful way without either matplotlib or bokeh (presumably a test of just creating, saving, and restoring each Element and Container type, without plotting them?).  And we also need to make the Bokeh plotting work when Matplotlib is not installed, which I believe requires making our own copies of a few Matplotlib objects that we use as defaults (e.g. colormaps).

doc/features.rst
jbednar: Sure, hopefully styles will normally unpickle also; we just have to make clear that in some cases there might be an object used as a value in those styles whose class is defined by the plotting library, and thus that particular option won't be able to be restored without the plotting library (nor would it presumably make any sense in that case).

doc/latest_news.html
jbednar: Ok, I'll remove "full".  Looking at it in context, I decided that "extensive" might be a good replacement -- I do want to convey that this is not just a couple of things that are supported, but that there's been reallly a lot of work involved and it's completely viable for use in projects.

jbednar: Regarding cross-backend layouts, when Philipp and I were discussing that, I briefly thought that PhosphorJS coupled with some custom stitching of SVG into larger SVGs might make us be able to solve layout problems once, for all backends.  But it's hard to see how we could do that when considering aligning plots by their active area (main plot), not just their bounding boxes (including axes, labels, etc.)  *Maybe* we could extract that information from all backends and supply it to single layout engine, but that seems ambitious.  So even if we do have some support for cross-backend layouts, it seems unlikely we'll be able to use that for everything...

jbednar:
BTW, is there already an issue to fix the table of contents from the Elements tutorials online, which are not formatting and linking correctly?  They seem formatted fine in my own IPython session, but online are a mess: http://holoviews.org/dev/Tutorials/Elements

philippjfr:
> BTW, is there already an issue to fix the table of contents from the Elements tutorials online, which are not formatting and linking correctly? They seem formatted fine in my own IPython session, but online are a mess: http://holoviews.org/dev/Tutorials/Elements

Not sure what happened there, might be a recent version of nbconvert or Sphinx that broke it. May have to create a pure HTML index.

jbednar:
What's the status of the Transforming Data tutorial currently linked from the sidebar?  Should I just delete that?  Or is there some less-than-perfect version we can slap up there?  Anything is better than nothing, even if it's just some code with no text...

philippjfr:
Yes, please delete it. It doesn't exist and has been subsumed by the Sampling Data and now the Columnar Data tutorial. We do still need a Tutorial about operations but that should probably be called something different.

jbednar:
Ok, will do.  So the Operations tutorial is now the missing one?  Do you at least have a list of code cells with one example per operation?  That's enough to be worth putting in.

philippjfr:
Not really no and there isn't that many operations defined so I don't think that would be very useful anyway. The Operations Tutorial should really just explain different types of operations (ElementOperation, MapOperation, TreeOperations) and then go on to implement a few basic examples.

jbednar:
The Elements tutorial states that "Points is very similar to Scatter, but it is sliceable in both x and y, not just x", but Scatter *is* sliceable in x and y:

image

Any suggestions for how to edit the Elements tutorial description to fix that?

jbednar:
Similarly, **Slicing and indexing in 2D** in the Sampling Data tutorial says "For data defined in a 2D space, there are 2D equivalents of the 1D Curve and Scatter types. A ``Path``, for example, can be thought of as a line in a 2D space. It may therefore be sliced along both dimensions".  Yet Scatter supports everything that's being shown for Path:

image

So isn't a Path just the connected version of Scatter?  If not, is there an example that would illustrate the difference?

Also, is there a good reason ``Scatter(paths)`` (and vice versa?) doesn't work?

philippjfr:
>The Elements tutorial states that "Points is very similar to Scatter, but it is sliceable in both x and y, not just x", but Scatter is sliceable in x and y:

That's no longer true. Really we need to convey the difference between key and value dimensions very clearly somewhere, i.e. that key dimensions are bins, samples, raw positions, or categories while the value dimensions represent some associated value with that sample.

>Similarly, Slicing and indexing in 2D in the Sampling Data tutorial says "For data defined in a 2D space, there are 2D equivalents of the 1D Curve and Scatter types. A Path, for example, can be thought of as a line in a 2D space. It may therefore be sliced along both dimensions". Yet Scatter supports everything that's being shown for Path.

Again it's only recently become true that Scatter supports the same thing. The difference is in the semantics, Paths/Polygons are defined in a 2D space, while Scatter is meant to represent a sampling in 1D with associated values for each sample. There is nothing enforcing this distinction but that's how the user should think of it.

>Also, is there a good reason Scatter(paths) (and vice versa?) doesn't work?

Yes, Path types can define multiple paths, while there is no easy way to store multiple paths in a Columns object without either inserting delimiting rows (kind of ugly and would probably require a lot of fiddling with the existing columns interface), or by inserting an ID for each path (which would be fairly redundant). In #347 I made some comments about the possibility about a new interface to store Path types in a format that would be very similar to NdMappings, and which could also serve as the future replacement for the relatively slow NdMapping implementation (again using an optional pandas interface). The example I give there is of a list of Paths/Polygons for individual countries: {'Australia': {'x': xs, 'y': ys}, 'Austria': ...}, where {'x': xs, 'y': ys} is a valid definition for an individual Curve. In many respects that's the same as NdOverlay but without the (fairly large) overhead of creating a separate ``Curve`` element for each path and mirroring that on the plotting side by creating hundreds/thousands of individual matplotlib artists or bokeh glyphs (which was a major performance concern in our recent project).

jbednar:
Hmm.  Ok, I'll update the tutorials to remove claims about slicing not being supported in those cases.  

What are the semantics of Scatter()[0.5] vs Points()[0.5]?  It looks like [0.5] is indexing for Scatter but an empty slice for Points; is that true?  Is it a general rule that [] falls back to slicing for cases where indexing is not defined?

philippjfr:
>What are the semantics of Scatter()[0.5] vs Points()[0.5]? It looks like [0.5] is indexing for Scatter but an empty slice for Points; is that true? Is it a general rule that [] falls back to slicing for cases where indexing is not defined?

In a 1D space (e.g. Scatter) indexing will return the closest sample, for >=2D Elements (e.g. Points) the index needs to be an exact match else you'll get an empty Element.

jbednar:
Why an empty *Element*, not a number, tuple, etc?  E.g. when I manage to get an exact match, I still get an Element instead of something I could use directly:

image

Apologies if this is something we discussed ages ago that I've forgotten. :-)

philippjfr:
>Why an empty Element, not a number, tuple, etc? E.g. when I manage to get an exact match, I still get an Element instead of something I could use directly:

That's a bug. Will have to investigate.

jlstevens:
> Should it maybe be an IndexError?

Tricky. 

When you don't know if a sample falls within some bounds, you don't necessarily want your code to fail just because there doesn't happen to be any data there. I see arguments both ways but yes, exact matches shouldn't be returning empty elements...

philippjfr:
>I see arguments both ways but yes, exact matches shouldn't be returning empty elements...

That's not happening, an exact match should be returning a scalar, currently it returns a length 1 Element. No match at all is returning empty Elements.

jbednar:
Right.  

In any case, seems like IndexError is reasonable in cases of non-exact matches, though it's arguably a  NotImplementedError.  I.e. the error isn't that it's not a valid index semantically, it's just not an index that we've implemented support for (presumably because that support would require multidimensional searching).

philippjfr:
>In any case, seems like IndexError is reasonable in cases of non-exact matches, though it's arguably a NotImplementedError. I.e. the error isn't that it's not a valid index semantically, it's just not an index that we've implemented support for (presumably because that support would require multidimensional searching).

I'm not multi-dimensional search is that desirable in any case, in the 1D case it's easy enough to reason about the sample you're requesting, once we have multiple dimensions we have to consider distance metrics and even with simple euclidean distance it will be pretty hard to reason about since you have no indication of where in the space you landed (unless you call ``closest`` as well).

jlstevens:
> ... currently it returns a length 1 Element

That actually makes sense then. Consider a overlay that you are slicing, there is nothing wrong with having a single point appear at a certain position. I guess this is what I have in mind as good behavior:

* Slices should always return elements (empty, single element..doesn't matter).
* Concrete indexing should return a value (i.e a scalar) if it is an exact match, ``IndexError`` otherwise.
* I agree with Philipp that multi-dimensional search may be too complicated and confusing to bother with.

Maybe I am forgetting something we decided on previously but right now, this seems like the right thing to me...

jbednar:
Sure; I'm not arguing for multidimensional search to be implemented, just that it *could* meaningfully be implemented, and hasn't yet been (and probably never will be, unless we come up with a clear concrete use case).  Probably more meaningful would be indexing as interpolation, really, using Delauney triangles, but that's way off topic for this thread, and somewhat magical.

jbednar:
I'm not sure what you mean about a length-1 Element "making sense" -- it's being returned from an indexing operation, not a slice.  For slices it would make sense, but not for indexing, right?

philippjfr:
* Slices should always return elements (empty, single element..doesn't matter).
* Concrete indexing should return a value (i.e a scalar) if it is an exact match, IndexError otherwise.

Yes, that's the what the behavior we decided on, will have to fix the length 1 bug for indexes and add an IndexError.

>I'm not sure what you mean about a length-1 Element "making sense" -- it's being returned from an indexing operation, not a slice. For slices it would make sense, but not for indexing, right?

Correct.

jlstevens:
Ok, we are agreed then! Good that we have decided on what is supposed to be happening and what exactly is a bug and not a feature. :-)

jbednar:
Anyway, as you can probably see, I'm working on the Sampling Data tutorial, which I don't think I've ever previously edited, so it's going slowly.  I think slicing is making sense mostly, barring the bug above, but I'm having trouble understanding what the different high-level categories of object really are and what is supported for each meaningful category.  E.g. the Sampling Data tutorial has a section "Regularly sampled or binned data in 1D", but I can't see anything about that section that requires regular sampling.  E.g. the first example in it works fine if I change the sampling to irregular:

image

So I can't figure out what the real breakdown into important categories are -- slicing just works, and so isn't informative of category, discrete vs. continuous seems to be used inconsistently, regular sampling isn't needed for the cases it's said to be needed, etc.  -- what's the real story?  I'm trying to work it out...

philippjfr:
>So I can't figure out what the real breakdown into important categories are -- slicing just works, and so isn't informative of category, discrete vs. continuous seems to be used inconsistently, regular sampling isn't needed for the cases it's said to be needed, etc. -- what's the real story? I'm trying to work it out...

Histograms are different again since their data is defined in terms of bin boundaries. We ideally also want to switch those to be Columns types eventually. Their current indexing behavior should return the value for the current bin you're in, while slicing has to include the bin boundaries.

You're also right that it doesn't really have anything to do with regular sampling, nothing about Columns, Histogram and Path types assumes or enforces anything about the data being regularly sampled. So I guess you should remove references to that.

The only real distinction that some types are treated as discrete samples of a continuous space and therefore snap. This is currently to restricted 1D Columns types (Curve, Scatter, ErrorBars) and Histograms. All other types are agnostic to their sampling.

jbednar:
> You're also right that it doesn't really have anything to do with regular sampling, nothing about Columns, Histogram and Path types assumes or enforces anything about the data being regularly sampled. So I guess you should remove references to that.

Ok, will do.

> The only real distinction that some types are treated as discrete samples of a continuous space and therefore snap. This is currently to restricted 1D Columns types (Curve, Scatter, ErrorBars) and Histograms. All other types are agnostic to their sampling.

But don't Image and other Raster types also support snapping?  In that case the regular grid *is* important, because it allows you to compute the nearest neighbor in 2D with no search and a well-bounded uncertainty on the point position.

philippjfr:
>But don't Image and other Raster types also support snapping? In that case the regular grid is important, because it allows you to compute the nearest neighbor in 2D with no search and a well-bounded uncertainty on the point position.

True, I was only referring to Columns types and the other types that could eventually become Column like. Raster types are another example of Elements that are represent discrete samples of a continuous space and therefore snap (this time in 2D). QuadMesh is again discrete but not necessarily regular.

Here's a breakdown of the way I see the different Elements:

0D Continuous: Distribution, Spikes, BoxWhisker*
1D discrete with snapping: Scatter, Curve, Histogram, ErrorBars
1D discrete without snapping: Bars*
2D discrete with snapping: Raster, Image, Surface
2D discrete without snapping: HeatMap
2D general: Points, Paths, Polygons
3D general: Scatter3D, Trisurface

* - can actually be multi-dimensional

jbednar:
Thanks.  Is it safe to say this?

  All Element types support slicing using a syntax like ``e[a:b]``, which will return another Element of the same type, with the data from the specified range [*a*,*b*).  

  Some Elements also support indexing, using a syntax like ``e[a]``, which will return scalar values representing the ``vdims`` (`value_dimensions`) of the nearest data point.  Specifically, the ``Scatter``, ``Curve``, ``Histogram``, and ``ErrorBars`` 1D Element types, and all of the 2D ``Raster``-based types (``Image``, ``RGB``, ``HSV``, ``QuadMesh``) support indexing, because in each case finding the nearest data point to the requested coordinate is straightforward.  Other types do *not* support indexing, because doing so would represent a poorly defined search in multidimensional space, rather than just computing bin location in 2D or nearest neighbors in 1D.

philippjfr:
>Thanks. Is it safe to say this?

Yes, looks good.

jbednar:
I'm also confused -- shouldn't the Sampling Data tutorial also define what ``.select()`` does?  select is used in only one location, without comment, but the conclusion of the tutorial says that we now know how to select data.

philippjfr:
>I'm also confused -- shouldn't the Sampling Data tutorial also define what .select() does? select is used in only one location, without comment, but the conclusion of the tutorial says that we now know how to select data.

Hmm, wonder if I had something and deleted it. The Exploring Data tutorial discusses it a little bit but we should probably present some simple examples here, showing that it's basically equivalent to indexing and slicing but can be applied to any datastructure.

jbednar:
Oh dear.  No wonder I've never been able to keep indexing/sampling/selecting/slicing straight.  Ok, is it safe to say this?

``
  In addition to slicing using a syntax like ``e[a:b]`` (supported for all ``Element`` types) and indexing using a syntax like ``e[a]`` (supported for some ``Element`` types as described above), HoloViews elements also support a separate method ``.select()``, which can do both slicing and indexing but is more verbose.
``

If .select is supported in some cases that regular [] slicing and indexing is not, I should say what those are.  But there do seem to be cases where .select doesn't work where [] does, which implies that .select isn't more general. E.g. .select doesn't seem to work for Histograms:

image

image

So I'm not sure what I can say about .select.

jbednar:
Also, the section on "The .table and .dframe methods" seems out of place in the middle of the Sampling Data tutorial; it doesn't seem to say anything about sampling data, just converting it. It's true that the subsequent section relies on it, but that doesn't mean it can go here.  Will it fit into the Pandas or Columnar tutorials?

philippjfr:
>If .select is supported in some cases that regular [] slicing and indexing is not, I should say what those are. But there do seem to be cases where .select doesn't work where [] does, which implies that .select isn't more general. E.g. .select works for Curves, but doesn't seem to work for Histograms:

It isn't more general just more convenient in some deeply nested cases and Histogram **should** support select. Another bug I'll have to look into.

>Also, the section on "The .table and .dframe methods" seems out of place in the middle of the Sampling Data tutorial; it doesn't seem to say anything about sampling data, just converting it. It's true that the subsequent section relies on it, but that doesn't mean it can go here. Will it fit into the Pandas or Columnar tutorials?

True, it's not really about sampling, moving them to Columnar Data might make sense but we should make it clear that most Element types can be converted to a Table.

philippjfr:
Just a quick side-note I was just looking at some of the new docs for Bokeh and came across their new charts documentation, which has a good section about tall and wide data (see [here](https://notebooks.anaconda.org/nroth/deep-dive)). We use the same format for our Columns so it might be worth providing a similar explanation.

philippjfr:
Should we merge soon? We wanted to get the website updated asap, so unless there are any half-finished changes I'd say merge now, I'll update the website and then you can open another PR to go through the rest of the Tutorials.

jbednar:
I think it's fine to merge now.  Please check my latest commits, and then merge.  I'll still have more commits, but nothing else major is ready yet.

philippjfr:
Something is going wrong in testing the Composing Data tutorial under Python 3, there's 8 of these errors:

```
ERROR: test_Composing_Data_data_000 (__main__.NBTester)
  File "/home/travis/build/ioam/holoviews/doc/nbpublisher/nbtest.py", line 437, in data_comparison
    ref_data,  ref_code =  pickle.load(ref_file)
ImportError: No module named 'UserString'
```

jbednar:
That's odd; I don't know what UserString is.  I did notice some print statements in your recently added tutorials that lacked () that I meant to fix for Python 3 usage, but I don't recall that being in Composing Data, and in any case I didn't change them...

philippjfr:
Seeing the same errors on master so it's not your changes. Will have to look into it.

philippjfr:
Traced down the bugs in the tests. Miniconda got upgraded from Python 3.4.3-2 to 3.4.4, which apparently breaks unpickling, pretty unsettling they're breaking the pickling protocol in bug fix releases. Will have to pin it to 3.4.3-2 for now.

philippjfr:
Python 2 tests are passing here and Python 3 tests are now passing again on master so I'll go ahead and merge.

PR #404, opened 2016-01-10
Columns fixes and improvements

philippjfr:
Various fixes that were overlooked previously. Also sets 'dataframe' interface to be preferred over 'dictionary' if pandas is available.

Also adds an initial draft of the [Columnar Data](http://philippjfr.com/work/work-in-progress/columnar-data/) Tutorial, which will replace the Pandas Conversion Tutorial.

### DIFF
--- a/doc/Tutorials/Columnar_Data.ipynb
+++ b/doc/Tutorials/Columnar_Data.ipynb
@@ -4,9 +4,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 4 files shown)


jbednar:
Looks good at a skim.  If Columnar Data supersedes (portions of?) other tutorials, it would be good to include deleting those as part of this PR.

jbednar:
Unless there is some reason not to, can we go ahead and merge this so that I can update my own PR from the master instead of this branch?  Seems like this tutorial will be better than not having one, in any case!

philippjfr:
Sure, go ahead and merge. I'll delete the apppropriate sections from the Pandas Conversion Tutorial (or maybe leave just a stub redirecting people to the new Tutorial) tomorrow.

Issue #418, opened 2016-01-26
Partial export


vascotenner:
The archive options of hv are great. It gives an easy access to all the plots, without having to save them 1 by 1. Furthermore it gives a way to quickly look what is inside a notebook, because the html export loads much faster than the ipynb file (and does not require an running ipython instance and trust of the nb).

In my workflow, the archive machinery can be improved even further. Quite often, I have notebooks that take quite some time to run. Often, I add (in a new ipython session) some cells at the end, that do not depend on ALL other cells in the notebook. When saving the archive, only the new cells are saved, while the old plots are deleted.

I am not sure if this should be fixed in the archive module, this module is way more powerfull than only saving rendered plots. Extracting images from an ipython notebook can also be done using nbconvert: http://stackoverflow.com/questions/22037910/converting-ipython-notebook-to-html-with-separate-images

What already might help, is not flushing the contents of an archive while saving.

jlstevens:
Firstly, I really glad you like the archive system! Personally, I haven't used it much but @jbednar convinced me that it was an important feature. It was also a lot of effort to implement!

If I understand correctly, the simple thing I could do for now is to add an option to keep accumulating to an archive. I.e if the archive folder/zip/tar has the same name, perform a merge instead of replacing it.

This seems like a reasonable feature request and I'll look into it when I get a chance.

jlstevens:
The PR has now been merged so I think this issue has been resolved. Do say so if there is something else about this you want to see addressed!

PR #419, opened 2016-01-26
Added option to avoid flushing the archive after export

vascotenner:
Feature request in #418


philippjfr:
Looks reasonable to me, but @jlstevens knows more about exporting so I'll let him review.

jlstevens:
I haven't tested it but the change looks simple and correct. 

The default of ``flush_archive=True`` means the code is effectively identical to what it was before (unless the parameter is changed). For this reason, I can't see how this change could have any negative effects which is why I'll go ahead and merge it right away.

PR #428, opened 2016-01-29
Fixed ticks in 'bare' x/yaxis plot option in matplotlib

philippjfr:
Relatively minor fix to tick labels that changes a few display tests.


jlstevens:
Ok. This would be a good place to test the new workflow for updating reference data! Should be ready soon...

philippjfr:
Ready for merge. Very minor fix, which unfortunately will touch quite a bit of reference data.

jlstevens:
> Very minor fix, which unfortunately will touch quite a bit of reference data.

For this reason, I would like to merge a few other PRs first to make sure reference data is updated as we expect...

jlstevens:
The change looks simple so I am happy to merge. As I understand it, we don't expect the ``push`` build to pass but the ``pr`` build is passing and that is what matters. Merging.

Issue #432, opened 2016-02-02
Alter plot output before display


vascotenner:
I want to combine a holoviews plot with an existing svg image, such that in my layout A is a drawing, and B and C are plots. In order to obtain this I propose another hook: before_display_hook, where I can alter the raw data.

philippjfr:
As described in #433 we've now added ``post_render_hooks`` allowing you to specify post-processing functions per output format. I'll close this issue now.

PR #433, opened 2016-02-02
Add extra hook before_display_hook to alter plotted data before display.

vascotenner:
See #432 

Example to add a second SVG file to the plot:
```python
%output fig='svg'
import xml.etree.ElementTree as et
from svg_merge import remove_meta, combine

def combiner(self, obj, data, fmt):
    if fmt == 'svg':
        data = combine([et.fromstring(data), remove_meta('/tmp/b.svg')])
    return data

hv.Store.renderers['matplotlib'].before_display_hooks=[combiner]

hv.Curve(zip([0,1,2], [0,1,.5]))
```

c

### DIFF
--- a/holoviews/plotting/mpl/renderer.py
+++ b/holoviews/plotting/mpl/renderer.py
@@ -55,6 +55,11 @@ class MPLRenderer(Renderer):
          The 'mpld3' mode uses the mpld3 library whereas the 'nbagg' uses
          matplotlib's the experimental nbagg backend. """)
 
+    before_display_hooks = param.HookList(default=[], doc="""
+        Optional list of hooks called before displaying a plot.
+        The hook is passed renderer, the plot object, the binary plotted object
+        and the fmt. It can be used to alter the raw data.""")
+
 
     # <format name> : (animation writer, format,  anim_kwargs, extra_args)
     ANIMATION_OPTS = {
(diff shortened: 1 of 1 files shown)


jlstevens:
Certainly an interesting proposal!

I've worked with SVGs quite a lot in the past so I am interested. However, to understand what is going on (and to decide whether I think this is the right approach) I would like to know more about ``combine``. I'm wondering if we could have SVG support such that normal overlays work.

In the past (before HoloViews), I created SVG templates where that let me substitute dynamic SVG output generated by matplotlib. Something like that looks like it would have done the job here...

I can see how this functionality is useful but this suffers the same issue as the other plot hooks: they tie you down to a specific backend and in this case, a specific output format (SVG). I'm not against such hooks as the user has to explicitly decide to use them (as long as they are aware of the limitations) which means normally you can change backend and format without any issue.

I'll have to think about it more to see if I can think of a better way. However, this is a fairly small change so it might be worth having anyway (e.g if you could think of cases where this is useful for formats other than SVG).

vascotenner:
I just tried to use this hook in combination with a slightly more complex svg file. The way I combine them now does not work well. For me the easiest option is to use inkscape to copy/paste my image in place.

This plotting hook is still usefull: the matplotlib svgs contain clipping-paths, that are not imported (opening works!) will in inkscape. I have now a 1 line sed script to remove these clipping-paths, but with this plotting option that can be automated easier.

jlstevens:
Thanks for the suggestion, I've now implemented it in b4113e9, 523ca4a and 12fd534.

I made a few changes that were agreed upon with Philipp. The hooks are called ``post_render_hooks`` which I think is more appropriate and it is implemented on the base ``Renderer`` class. These hooks have a slightly different signature and are applied according to their declared format type. For instance:

```python
%output fig='svg'
def combiner(data, obj):
    # Process data here
    return data

hv.Store.renderers['matplotlib'].post_render_hooks={'svg':[combiner]}
hv.Curve(zip([0,1,2], [0,1,.5]))
```

This seems to make more sense - the hooks will be more dependent on the output format type (e.g svg versus png) than the renderer used to generate that data. This also simplifies the signature of the hooks themselves.

This should be able to do exactly the same thing as you what you proposed in this PR. Sorry we didn't end up using your code but as we are aiming to release 1.4.2 tomorrow, I wanted to get this functionality into master now. I hope this approach remains useful to you!

PR #434, opened 2016-02-02
Cleaned up and improved legend handling for all backends

philippjfr:
As the title suggests this is a refactor of the legend code so that a) it behaves more similarly across backends and b) to clean up the code a little bit.

Concretely this PR addresses the following issues:

* Legends not updating per frame in mpl
* Legend zorder being too low in mpl
* Allow toggling legends consistently at the Element and Overlay level, providing finer control. Previously some Element types could be toggled others could not.

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -264,6 +264,9 @@ class SpikesPlot(PathPlot):
     position = param.Number(default=0., doc="""
       The position of the lower end of each spike.""")
 
+    show_legend = param.Boolean(default=True, doc="""
+        Whether to show legend for the plot.""")
+
     style_opts = (['color', 'cmap', 'palette'] + line_properties)
 
     def get_extents(self, element, ranges):
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -602,46 +602,39 @@ class OverlayPlot(LegendPlot, GenericOverlayPlot):
         for subplot in self.subplots.values():
             subplot._finalize_artist(key)
 
-    def _adjust_legend(self, axis):
+    def _adjust_legend(self, overlay, axis):
         """
         Accumulate the legend handles and labels for all subplots
         and set up the legend
         """
-
         title = ''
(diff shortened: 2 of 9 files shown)


philippjfr:
This PR is now ready. While not perfect, legend behavior should now be much more consistent and I've tidied up the matplotlib implementation a little bit.

holoviews/plotting/bokeh/chart.py
```diff
@@ -264,6 +264,9 @@ class SpikesPlot(PathPlot):
     position = param.Number(default=0., doc="""
       The position of the lower end of each spike.""")
 
+    show_legend = param.Boolean(default=True, doc="""
```
jlstevens: This is a parameter available for the matplotlib backend IIRC. In which case, it is good to see this support added to the Bokeh backend.

philippjfr: It's always been there I'm just being explicit about which plots have it enabled and disabled by default.

holoviews/plotting/mpl/element.py
```diff
@@ -658,6 +651,7 @@ def _adjust_legend(self, axis):
             frame.set_facecolor('1.0')
             frame.set_edgecolor('0.0')
             frame.set_linewidth('1.0')
+            leg.set_zorder(10e6)
```
jlstevens: Will this break if someone uses a z-order of ``10e7`` somewhere else? ;-p

holoviews/plotting/bokeh/chart.py
jlstevens: Ok good.

holoviews/plotting/mpl/element.py
philippjfr: Yes it will, could make it some very large but more obscure number I guess.

jlstevens:
Looks good and I am happy to merge. 

This is the first PR to use the new documentation building system. I'm really happy with how easy it was to update the reference data for this PR. Note that when I click merge, this should trigger a merge hook but this will be the first time it is tested - we may have a failure on master if something goes wrong with the test data update.

Issue #425, opened 2016-01-28
Turn off dimension label for a specific HoloMap


basnijholt:
I have a `HoloMap` with custom labels:
<img width="784" alt="screen shot 2016-01-28 at 16 03 47" src="https://cloud.githubusercontent.com/assets/6897215/12648141/cd545ba2-c5d8-11e5-9376-f317a7ff7f19.png">

I would like to turn off the label that is automatically generated by the `HoloMap` for this specific one, not all by doing `hv.core.dimension.title_format = ''"`.

jlstevens:
Just to be clear, is there any reason you cannot just disable the title in the plot options?

philippjfr:
That's true, the main problem is when you want group and label to be shown and not the dimension values. The title_format plot option should ideally provide control over that.

jlstevens:
Seems to me this should also be done via the plot option.

The way I see it, the use of ``dim_title`` (``dim_title = self._frame_title(key, separator=separator)``) in ``GenericElementPlot._format_title`` should also be controlled via the ``title_format`` plot option.

philippjfr:
The title_format argument now accepts a new keyword, by default it is:

```
title_format='{group} {label}\n{dimensions}'
```

By removing the dimensions part you can disable dimensions being displayed.

PR #436, opened 2016-02-02
title_format plot option now includes dimensions

philippjfr:
As the title says this allows more customizing the plot title_format as suggested in #425.

### DIFF
--- a/holoviews/plotting/mpl/plot.py
+++ b/holoviews/plotting/mpl/plot.py
@@ -916,7 +916,7 @@ class LayoutPlot(GenericLayoutPlot, CompositePlot):
 
         # Create title handle
         if self.show_title and len(self.coords) > 1:
-            title = self.handles['fig'].suptitle('', **self._fontsize('title'))
+            title = self.handles['fig'].suptitle('', y=1.05, **self._fontsize('title'))
             self.handles['title'] = title
 
         return layout_subplots, layout_axes, collapsed_layout
--- a/holoviews/plotting/mpl/plot.py
+++ b/holoviews/plotting/mpl/plot.py
@@ -916,7 +916,7 @@ class LayoutPlot(GenericLayoutPlot, CompositePlot):
 
         # Create title handle
         if self.show_title and len(self.coords) > 1:
-            title = self.handles['fig'].suptitle('', y=1.05, **self._fontsize('title'))
+            title = self.handles['fig'].suptitle('', **self._fontsize('title'))
             self.handles['title'] = title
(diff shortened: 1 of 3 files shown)


philippjfr:
The ``title_format`` plot option now allows enabling or disabling dimensions in the title. The only way I was able to do so without losing flexibility was to let the title accept a separator variable: "{label} {group}{separator}{dimensions}". Since bokeh doesn't support newline characters I'm tempted to remove this again and simply set different default title_format, for matplotlib I'd have "{label} {group}\n{dimensions}" and for bokeh I'd have "{label} {group} - {dimensions}".

holoviews/plotting/mpl/plot.py
```diff
@@ -916,7 +916,7 @@ def _compute_gridspec(self, layout):
 
         # Create title handle
         if self.show_title and len(self.coords) > 1:
-            title = self.handles['fig'].suptitle('', **self._fontsize('title'))
+            title = self.handles['fig'].suptitle('', y=1.05, **self._fontsize('title'))
```
jlstevens: Is this a magic number? I assume this makes it look better.

philippjfr: Yea, generally avoids overlap between the plot title and the overall layout title. Not sure how best to provide control over it.

holoviews/plotting/plot.py
```diff
@@ -145,8 +145,9 @@ class DimensionedPlot(Plot):
     show_title = param.Boolean(default=True, doc="""
         Whether to display the plot title.""")
 
-    title_format = param.String(default="{label} {group}", doc="""
-        The formatting string for the title of this plot.""")
+    title_format = param.String(default="{label} {group}\n{dimensions}", doc="""
+        The formatting string for the title of this plot, allows defining
+        a label group separator and dimension labels.""")
```
jlstevens: Definitely an improvement as long as the old tests pass (i.e backwards compatible).

philippjfr: It probably won't be, there may be fewer whitespace now and things may have shifted around a little bit.

jlstevens: Ok. That means there will be quite a lot of test data to update. We need to be careful about transient glitches and need to review the display output thoroughly.

philippjfr:
Looks good for merging, the display changes are the deletion of trailing whitespace in the titles when no label is supplied.

jlstevens:
Ok. Tests are passing. Merging.

Issue #435, opened 2016-02-02
Store.dump and fileobject


vascotenner:
In the definition of Store.dump (in core/options.py), it asks for a filename as argument. However, inside the function it wants to have a file object!

jlstevens:
Well spotted! The argument name should be renamed to ``file`` (consistent with ``pickle.dump``).

philippjfr:
PR got merged. Closing.

PR #437, opened 2016-02-02
Changed argument of dump function to file instead of filename

vascotenner:
Fixing #435


philippjfr:
Looks good to me, merging as soon as tests are passing.

PR #438, opened 2016-02-03
Refactored matplotlib ElementPlots by adding higher-level API

philippjfr:
This refactor avoids a lot of the duplication that has spread across all existing matplotlib plots. It should also make it simpler for a user to create a new plot. I'll summarize the overall API once I've worked on it a little bit more.


jlstevens:
Great! Will this put us in a position where we can document these classes and their methods properly? Maybe even write some sort of developer documentation about it?

philippjfr:
The general idea behind this PR is to simplify the implementation of matplotlib plotting classes to reduce redundancy and make it easier to implement a new Element class.

Previously the main methods to implement a plotting class were:

* initialize_plot: Would have to handle everything from unpacking the data in the element, getting ranges, getting styles, calling the the plotting function, storing the artist on handles and call finalize_axis. This meant that there were ~10 lines of boilerplate per plotting class.
* update_handles: This method would update existing artists by accessing the handles and manually setting the new data. This had to be implemented if just to tear down the existing artist and redraw it from new. It also required boilerplate to get the styles and unpack the data in a very similar way as the initialize_plot call would already have done.

To make this old API workable a lot of plotting classes implemented custom workarounds, e.g. so that each plot didn't have two separate implementations to unpack the element. However in developing the bokeh backend I came up with a better solution. I've now ported some of the ideas from the bokeh backend to matplotlib, it's important to mention though because the matplotlib API is sometimes awkward this new API is entirely optional and some plots will retain the old API for now.

The general idea is to separate the unpacking of the data and keyword arguments from the element from the actual plotting. The ElementPlot baseclass now has a default implementation of ``initialize_plot``, which does all the boilerplate stuff mentioned above, then calls a new method called ``get_data`` and passes the output from that method to another new method called ``init_artist``. These two methods are generally enough to implement a plot type but if you want to manually update the handles on a plot instead of recreating them you also implement a ``update_handles`` method, which can itself call ``get_data`` to get the latest data, if not the previous artist will be torn down by a method called ``teardown_handles``.

The signature for these methods is the following:

- ``get_data``:
   - Inputs: ``element``, ``ranges``, ``style``
   - Outputs: plot_args, plot_kwargs, axis_kwargs
   - This method should unpack the element and return the appropriate arguments for the plotting call. Additionally it may also supply axis_kwargs, which allow supplying custom ticks and axis labels (e.g. think of a HeatMap where the x-axis and y-axis labels are derived from the data.

- ``init_artist``:
   - Inputs: ``axis``, ``plot_args``, ``plot_kwargs``
   - Output: Dictionary of artists
   - This method is given an axis to draw onto as well as plot_args and plot_kwargs to pass to the plotting call, should return a dictionary of artists, by convention the main artist should have the key artist.

- ``update_handles``
   - Optional method to update existing handles, if not implemented will simply recreate artist for each frame. Otherwise should use ``get_data`` method to unpack required data and then update artists manually with the data.
   - Inputs: ``axis``, ``element``, ``ranges``, ``style``
   - Output: Optional dictionary of axis arguments, e.g. xticks, yticks etc. 

- ``teardown_handles``
  - Only required if some custom teardown procedure is required, i.e. if no update_handles method has been implemented and the plot does not consist of a simple matplotlib artist or collection. Should go through handles and remove them from the axes.
  - Inputs: None
  - Outputs: None

Overall implementing all this roughly halves the plotting code required and in future when matplotlib provides a better interface to update existing artists, it should be possible to feed he  output of ``get_data`` straight to the artist without manually updating various attributes. As you can see this almost halves the length of the various plotting implementations I have refactored but there's still full flexibility to override ``initialize_plot`` and  ``update_handles`` completely as is still the case for Histogram and Bar plots, where more flexibility is required.

philippjfr:
I'm now done with the refactor itself. There will be a few output changes due to changes in handling edgecolors in 3d scatter plots, which is now basically a very small extension to the 2d scatter plot where I changed the edgecolor handling already. To make sure I didn't break anything else I made a copy of the Elements tutorial and turned each cell into a HoloMap. While I probably wouldn't want that as a Tutorial it makes a good test case and maybe we can consider running it independently of the regular Tutorials somehow.

Now that the refactor itself is done I'll start putting together some documentation on the plotting baseclasses and then two separate Tutorials on extending the matplotlib and bokeh backends.

philippjfr:
Backwards incompatible changes:

1) The signature of the ``update_handles`` method has changed
2) ``DrawPlot`` has been deprecated in favor of implementing a regular ``ElementPlot`` without an ``update_handles`` method.

philippjfr:
Okay I was wrong the 3d stuff wasn't the only change. There are three display changes:

1) 3d plot edgecolors are handled better now resulting in some changes in Elements tutorial.
2) HeatMap annotations now drawn slightly differently resulting in one display change in Continuous Coordinates.
3) VectorFields thickness has changed and I can't figure out why. I checked that all the arguments to the ``quiver`` call are the same in both implementations. Mysterious but not visually noticeable.

The first two are fixes/improvements and I really don't know what to do about 3). All three are very minor and purely aesthetic changes, so I'll focus on documenting now.

holoviews/plotting/mpl/chart.py
```diff
@@ -115,26 +115,10 @@ class CurvePlot(ChartPlot):
...
     style_opts = ['alpha', 'color', 'visible', 'linewidth', 'linestyle', 'marker']
 
-    def initialize_plot(self, ranges=None):
-        element = self.hmap.last
-        axis = self.handles['axis']
-        key = self.keys[-1]
-
-        ranges = self.compute_ranges(self.hmap, key, ranges)
-        ranges = match_spec(element, ranges)
+    def init_artist(self, ax, plot_data, plot_kwargs):
```
jlstevens: Should be called ``init_artists`` if multiple artists can be initialized here.

philippjfr: Agreed, will rename.

holoviews/plotting/mpl/chart.py
```diff
@@ -578,59 +525,49 @@ class PointPlot(ChartPlot, ColorbarPlot):
...
-        legend = points.label if self.show_legend else ''
-        scatterplot = axis.scatter(xs, ys, zorder=self.zorder, label=legend,
-                                   edgecolors=edgecolor, **style)
-        self.handles['artist'] = scatterplot
+        style['edgecolors'] = style.pop('edgecolors', style.pop('edgecolor', 'none'))
 
-        return self._finalize_axis(self.keys[-1], ranges=ranges)
+        if element.get_dimension(self.size_index):
+            sizes = element.dimension_values(self.size_index)
+            ms = style.pop('s') if 's' in style else plt.rcParams['lines.markersize']
```
jlstevens: Happy to use ``rcParams`` here? I thought ``rcParams`` could sometimes be tricky...

holoviews/plotting/mpl/chart.py
```diff
@@ -679,104 +616,73 @@ def _get_map_info(self, vmap):
...
-        return (xs, ys, list((radians / np.pi) * 180),
-                magnitudes, colors, input_scale / min_dist)
+        return np.min([self._get_min_dist(vfield) for vfield in vmap])
 
 
     def _get_min_dist(self, vfield):
         "Get the minimum sampling distance."
-        xys = np.array([complex(x,y) for x,y in zip(vfield.dimension_values(0),
-                                                    vfield.dimension_values(1))])
+        xys = vfield.array([0, 1]).view(dtype=np.complex128)
```
jlstevens: Does 128 matter? Won't any complex datatype do and if so, why specify 128?

holoviews/plotting/mpl/chart.py
philippjfr: Hmm I don't think there are any particular issues but I guess we could require that a 's' is supplied to apply points scaling.

holoviews/plotting/mpl/chart.py
```diff
@@ -1065,18 +954,20 @@ def get_data(self, element, ranges):
...
+        style['array'] = array
+        style['clim'] = clim
+        return [np.array(data)], style, {}
 
 
-    def update_handles(self, axis, element, key, ranges=None):
+    def update_handles(self, key, axis, element, ranges, style):
         artist = self.handles['artist']
-        data, array, clim = self.get_data(element, ranges)
+        (data,), kwargs, axis_kwargs = self.get_data(element, ranges, style)
```
jlstevens: A unituple? I assume there are instances where this tuple is longer?

holoviews/plotting/mpl/chart.py
philippjfr: Good point, it's relying on the two columns being 64-bit, will have to double check this or revert it.

holoviews/plotting/mpl/chart3d.py
```diff
@@ -109,45 +109,30 @@ class Scatter3DPlot(Plot3D, PointPlot):
...
-        self.handles['axis'].add_collection(scatterplot)
-        self.handles['artist'] = scatterplot
-
-        return self._finalize_axis(key, ranges=ranges)
-
-    def update_handles(self, axis, points, key, ranges=None):
+    def get_data(self, element, ranges, style):
+        xs, ys, zs = (element.dimension_values(i) for i in range(3))
+        self._compute_styles(element, ranges, style)
+        # Temporary fix until color handling is deterministic in py3
```
jlstevens: Is it python 3 or matplotlib the issue here?

holoviews/plotting/mpl/chart.py
philippjfr: It's the arguments to the plotting call and you're right they should be consistent.

holoviews/plotting/mpl/chart3d.py
```diff
@@ -109,45 +109,30 @@ class Scatter3DPlot(Plot3D, PointPlot):
...
+        ax.add_collection(scatterplot)
+        return {'artist': scatterplot}
+
+    def update_handles(self, key, axis, element, ranges, style):
         artist = self.handles['artist']
-        artist._offsets3d = tuple(points[d] for d in points.dimensions())
-        cdim = points.get_dimension(self.color_index)
-        style = self.style[self.cyclic_index]
+        offsets, style, _ = self.get_data(element, ranges, style)
+        artist._offsets3d = offsets
```
jlstevens: You don't seem to use ``offset`` later so I would assign it straight to ``artist._offsets3d``.

holoviews/plotting/mpl/chart3d.py
philippjfr: The behavior seems to be deterministic in python 2 but not python 3, not sure if it's down to dictionary ordering or something else.

holoviews/plotting/mpl/chart3d.py
```diff
@@ -214,24 +187,13 @@ class TrisurfacePlot(Plot3D):
...
+    def get_data(self, element, ranges, style):
         dims = element.dimensions(label=True)
         vrange = ranges[dims[2]]
+        style['vmin'] = vrange[0]
+        style['vmax'] = vrange[1]
         x, y, z = [element.dimension_values(d) for d in dims]
-        artist = axis.plot_trisurf(x, y, z, vmax=vrange[1],
-                                   vmin=vrange[0], **style_opts)
-        self.handles['artist'] = artist
+        return (x, y, z), style, {}
```
jlstevens: Now back to returning data as a tuple...

philippjfr: Right as above, they should be consistent.

holoviews/plotting/mpl/element.py
```diff
@@ -711,40 +749,3 @@ def update_frame(self, key, ranges=None, element=None):
             self._adjust_legend(element, axis)
 
         self._finalize_axis(key, ranges=ranges)
-
-
-
-class DrawPlot(ElementPlot):
```
jlstevens: Good to see a class go. Think we have any backward compatibility concerns about this though?

philippjfr: Bas' Sphere thing is the only example of this that I know of.

jlstevens: I also had a networkX prototype that used this but that never went anywhere...

jlstevens:
Made a few minor comments but overall it does seem like a nice improvement. As it is a fairly large PR, I have a few general comments:

* We need to be very careful when updating the test data. We must make sure we know exactly what changed and that it is ok.
* Probably worth running flakes (i.e flake8) on your branch to see if it catches anything before merging. Sometimes flakes helps catch actual bugs...
* Originally the handles dictionary was meant for matplotlib artists. Seems handles can now hold almost any matplotlib component...

Anyway, I'm happy to merge once these issues are addressed. Frankly, I expect we will find bugs related to this refactor but I do think this is worth having cleaner, more easily understandable code.

holoviews/plotting/mpl/element.py
philippjfr: Basically all you do now is define a ``teardown_handles`` method containing:

```
if self.zorder == 0: self.handles['axis'].cla()
```

and then define a ``get_data`` and ``init_artist`` method, if you ever figure out how to update the plot by changing an existing artist you just replace the ``teardown_handles`` method with an ``update_handles`` method and it will use that instead.

philippjfr:
>We need to be very careful when updating the test data. We must make sure we know exactly what changed and that it is ok.

I've carefully reviewed the test data changes and listed the 3 things that have changed above. The only thing I'm sort of worried about is the VectorField changes as their width has changed slightly and I can't figure out why.

>Probably worth running flakes (i.e flake8) on your branch to see if it catches anything before merging. Sometimes flakes helps catch actual bugs.

Definitely a good idea.

>Originally the handles dictionary was meant for matplotlib artists. Seems handles can now hold almost any matplotlib component...

We've always been putting the axis and figure in it, and apart from that it's generally only used for artists. The artist key is treated somewhat specially as that is where the legend code expects to find the artist and by default that is what gets removed if you don't define an ``update_handles`` method so that it can simply be redrawn completely.

holoviews/plotting/mpl/chart.py
philippjfr: I think it's best to postpone this to another PR as to avoid changing any more test data.

jlstevens:
I understand this is ready for a merge and the tests are passing. I'm glad that the plotting code will be clearer to understand after this!

PR #439, opened 2016-02-03
Dimensioned.clone always transfers id to new object

philippjfr:
This PR will make sure styles are transferred when applying various operations to an object including slices. This will probably affect some existing tests but it is better behavior.


jlstevens:
Ok, the tests are passing and I do think this is a good change. Merging.

PR #443, opened 2016-02-04
Normalizing aspects when computing Layout fig_inches

philippjfr:
This fixes a bug when adjusting aspects in a Layout and setting the aspect_weight to a non-zero value. The previous bug made it incredibly fiddly to get the plot spacing in a gridded layout with non-square aspects right. Now when setting aspect_weight=1 you should always get a plot without excessive whitespace.

### DIFF
--- a/holoviews/plotting/mpl/plot.py
+++ b/holoviews/plotting/mpl/plot.py
@@ -820,6 +820,8 @@ class LayoutPlot(GenericLayoutPlot, CompositePlot):
         # Compute and set the plot size if not explicitly supplied
         col_ars = [ar for ars in col_aspect_ratios for ar in ars]
         row_ars = [ar for ars in row_aspect_ratios for ar in ars]
+        col_ars = [ar/max(col_ars) if ar else 0 for ar in col_ars]
+
         width = len(col_ars[::2]) + sum(col_ars[1::2])
         yscale = sum(col_ars)/sum(row_ars)
         xinches, yinches = None, None


jbednar:
Wow, does this already address what I just said was possible at the start of my comment in https://github.com/ioam/holoviews/pull/444?  I.e., will the plot Jean-Luc drew at the top be what actually happens automatically when combining things that have the shapes (aspect ratios) shown there?

jlstevens:
@jbednar This will really help make it easier to use the current system but this is not quite the same as properly nested Layouts. Right now, Layouts do not nest by design (they simple concatenate). 

Otherwise, yes this makes it much easier to define the sort of Layout I showed in the PR - in fact that is exactly the layout we were trying to get working better!

philippjfr:
>Wow, does this already address what I just said was possible at the start of my comment in #444? I.e., will the plot Jean-Luc drew at the top be what actually happens automatically when combining things that have the shapes (aspect ratios) shown there?

Yes as long as those ``aspect_weight`` has been set to 1 and the aspect ratios have been set explicitly as I do not try to infer data derived aspect ratios. Example:

```python
%%opts Area [aspect=3] Image [aspect=0.7] Layout [aspect_weight=1]
np.random.seed(13)
img = hv.Image(np.random.rand(10,10))
area = hv.Area(np.random.rand(10))
(img + img + area + area).cols(2)
```

image

jbednar:
Excellent!  That's a big step forward in usability (e.g. for many of Jean-Luc's thesis plots). I guess the tick/label generation code doesn't know about the axis squashing, and thus generates too many ticks and labels?

philippjfr:
Bit annoyed as I'd thought I'd done this when I originally implemented all this but it seems I didn't commit it and it's not actually been working properly ever since (this was pre-1.3). Also the only reason ``aspect_weight`` isn't a boolean ``respect_aspects`` or similar is because this bug made it really hard to achieve a good layout. Glad it's finally fixed though.

jbednar:
Makes sense.  I always thought this should be easily fixable!  (Not that I ever dived into the code to try to fix it myself...)

philippjfr:
>I guess the tick/label generation code doesn't know about the axis squashing, and thus generates too many ticks and labels?

Right, it's easy enough to change them yourself and I'd rather not complicate the layout code much more for now, better to wait until we decided what we decide on to improve layouts in general.

jbednar:
Makes sense, though I actually don't know how to change them, so it's not easy for me.  Plus when people start changing them, their code will have those specific fixes in each plot, causing forwards compatibility problems for them when they later upgrade to where we fix them more generally.

philippjfr:
>Plus when people start changing them, their code will have those specific fixes in each plot, causing forwards compatibility problems for them when they later upgrade to where we fix them more generally.

Explicitly setting xticks or yticks will always take precedence over some automatically derived setting. The valid formats for xticks are integers for the number of ticks, a list of positions, or a list of tuples specifying (position, label).

philippjfr:
Tests are passing, which means we're not actually testing this anywhere. I'll add an example to Containers once this is merged.

jlstevens:
Ok, I'm really happy this is working well (again?). Definitely a great thing to have in 1.4.2.

PR #444, opened 2016-02-04
Layout templates

jlstevens:
This PR is designed to address the difficulties we have with customizing Layouts, e.g setting non-square aspects, positioning subfigures and insets. What is proposed here was first suggested on Gitter.

The basic idea is that HoloViews makes visualizing things easy. So why not use our existing elements and operators to define a 'layout template'? We already have a ``Bounds`` element that is pretty ideal for this:

```python
hv.Bounds((0.05, 0.3, 0.9, 0.95), extents=(0, 0, 2, 1)) *\
hv.Bounds((0.05, 0.05, 0.9, 0.25)) *\
hv.Bounds((1.05, 0.3, 1.9, 0.95)) *\
hv.Bounds((1.05, 0.05, 1.9, .25))
```

image

The idea is that the overlay order defines the subfigure ordering. This would let us do insets as well as you could position these anywhere you like. We might want a special layout template element/object that could do something like this but number the subfigures.

There isn't much here yet as we will need to discuss this proposal carefully before much code is written. Discussing this with Philipp, it sounds like this would be feasible to support in the plotting code.

### DIFF
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -789,6 +789,9 @@ class GenericLayoutPlot(GenericCompositePlot):
     """
     A GenericLayoutPlot accepts either a Layout or a NdLayout and
     displays the elements in a cartesian grid in scanline order.
+
+    The positioning of the elements of the layout can be customized
+    with a layout template.
     """
 
     def __init__(self, layout, **params):


jbednar:
This sounds like something for after 1.4.2, as I think it will take some iteration to make it be general and usable.  

It does sound like a useful alternative, allowing precise control when needed.  But I would really hope people would rarely need such a mechanism.  I'd guess that most people would only want to use this as they approach a finished (and fairly high-profile) publication.  It gives them very precise absolute positioning, but is only feasible if they are not expecting to change things afterwards, because it only works with very specific sizes and arrangements.  

I'd be much happier to see things that make the normal HoloViews layout process more powerful in general.  For starters, the layout above *should* just work already, using +; I don't see any reason we can't make the extra vertical space that's currently generated for the second row go away, without major changes to our current approach, i.e. just using `(a+b+c+d).cols(2)`. 

Plus it also seems like we should be able to handle nesting well, as in  `((a + b).cols(2) + c + d + (e+f+g).cols(3)).cols(2)`  to specify a layout like:

```
   ---a---  ---b---  --------c--------
   --------d-------  --e-- --f-- --g--
```

or `((a + b + c).cols(3) + (d + (e+f+g).cols(3)).cols(2))).cols(1)` to specify:

```
   -----a-----  -----b-----  ------c------
   ---------d--------  ---e-- ---f-- ---g-
```

It would need a further extension to handle different relative sizes of things that aren't just evenly divided as above.   E.g. we could declare an `hstretch` parameter (defaulting to 1.0) for each element or group of elements, which would add control over the relative size of things in a given group, to allow layouts like:
 `((a + b.hstretch(2.0) + c).cols(3) + (d + (e+f+g).cols(3).hstretch(1.5)).cols(2))).cols(1)` to specify:


```
   ----a---  --------b--------  ----c-----
   -------d-----  ---e--- ---f--- ----g---
```

I.e., any plot that stretches would make the other plots in the same group get smaller or bigger so that they fit into whatever size is available for that whole group, when the whole context is considered.

Note that this all treats horizontal space as the driver for everything (as in a word processor), and then vertical layout just has to follow on from there.  Of course, that won't work well when people want to lay out plots starting in the upper left and going down, because they would be setting relative sizes horizontally when they really want to control things vertically.   Supporting that would require a corresponding set of stretch and grouping operations for vertical arrangements, but I guess that might work too (.rows() and .vstretch).

jlstevens:
> This sounds like something for after 1.4.2, as I think it will take some iteration to make it be general and usable.

I agree.

>  It gives them very precise absolute positioning, but is only feasible if they are not expecting to change things afterwards, because it only works with very specific sizes and arrangements.

I think it is somewhat flexible. It can handle any number of subfigures (well...at least up to the number you've defined in the template). As the template defines the size of the axes so I don't see why it would be tied to sizes. What *is* restrictive about templates is that with non-square aspects, there are probably some plot types that make more sense squashed/stretched than others i.e the position of the element in the layout may affect how good it looks.

> I'd be much happier to see things that make the normal HoloViews layout process more powerful in general. 

I also agree here. What you suggest here is something we've discussed before (I think we considered an ``ncols`` method before to define such a nesting process). Although I really do like the idea, I think that would be appropriate for a separate PR (this one is only about the templating idea). Philipp would be the one to evaluate how difficult such a proposal would be...

One advantage of the templating suggestion is that it would be easier to implement. Because it is explicit, there is a lot less computation we need to do to figure out a good layout (implementing a layout engine is  hard!).

philippjfr:
I guess I'll expand a little bit on some of the possibilities this idea would allow. Basically what Jean-Luc proposed would provide a very low-level interface to specify complex layouts. As you point out this is not the interface a user will usually want to use, but it gives us huge freedom to create custom layouts. The idea would be that we would have various helper functions to generate complex layout templates and combine and validate them against an existing layout. 

E.g. combining the layout template from above with a Layout of the form:

```
:Layout
   .Image.A :Image   [x,y]   (z)
   .Image.B :Image   [x,y]   (z)
   .Curve.A :Curve   [x]   (y)
   .Curve.B :Curve   [x]   (y)
```

would give you a view something like this:

image

Various helper functions would generate layout templates using Bounds objects so you can view the resulting layout immediately and then validate it against an existing layout. You should also be able to get a template from an existing layout, do some fine-tweaking of the bounds and then combine them again.

I think this would be a very HoloViews way of working with layouts, ``annotate your data and let it visualize itself``. All the variations of nested layouts could be implemented much more easily by relying on this low level implementation, we'd simply have a helper function which generates a template from a nested layout. So we'd start with implementing a version of LayoutPlot which lays out plots according to a Bounds template and then we can work on generating bounds templates based on (nested) Layout objects.

jlstevens:
I'll also add that this system would allow insets by deliberately overlapping the subfigures. This is not something that would be possible to specify purely by nesting layouts without such a template (or some way of specifying how the subfigures overlap).

jbednar:
I may be missing some of your emphasis here, but from Philipp's description it sounds like this approach could be useful for creating a new Element-like chunk that then later gets laid out alongside other Elements in the usual way?  I.e., to group several bits into a single unit with fixed internal spacing, but then the result goes into a larger figure that has the usual flexible spacing?  If so, it does sound more useful -- people could set up specific chunks that have such absolute positioning, but then still be able to flexibly combine them with other chunks.

Allowing insets is indeed useful, though I was thinking of that as an overlaying problem and not a layout problem.  If approached as layout, I guess it would need an offset x,y hint akin to the stretch hint above (defaulting to 0,0), which could actually be cool too -- it would let people adjust relative spacing when needed, while still laying things out automatically.

I agree that what I'm proposing here ought to be in a separate PR, and if I were a good person I would move it there. :-)

philippjfr:
>a new Element-like chunk that then later gets laid out alongside other Elements in the usual way? I.e., to group several bits into a single unit with fixed internal spacing, but then the result goes into a larger figure that has the usual flexible spacing?

Not quite no, this would specify the overall layout of everything. A helper function would just automatically generate a layout template (i.e. an overlay of bounds objects) in the background and then create the axes based on that layout template. In most cases this is an implementation detail that's hidden from you, and simply provides a general specification of the layout of axes. But you could also construct an (optionally nested) Layout the usual way, get a hold of the template, and make specific adjustments before plotting the combined ``template*layout``. Basically it's a proposal for a very low-level layout specification, which happens to visualize itself.

It requires an implementation in two stages:

1) Plotting classes which translate an overlay of Bounds elements into a set of axes positioned according to the specification.
2) A helper function which automatically generates an Overlay of Bounds Elements from an existing Layout, which takes into account aspects, nesting etc..

On it's own step 1 isn't very useful, but once we have that in place we can add all kinds of functionality to allow you to quickly generate, visualize and modify these templates, including the usual ``+`` and ``*`` operators and the proposed nesting functionality.

philippjfr:
Can we reopen this as an issue and close this PR?

jlstevens:
> Can we reopen this as an issue and close this PR?

Sure. Just make sure to clearly link to the discussion above.

jbednar:
Ok, will do.

Issue #291, opened 2015-11-12
Invalid keys in options dictionary are silently ignored (but should print a warning instead)


maxalbert:
Executing the two IPython notebook cells below illustrates the issue. The second one produces a plot, but the (invalid) fontsize option `labels` is silently ignored. Instead, it should at least print a warning. The current behaviour is a usability issue because it makes it much harder to experiment and discover what the invalid options are. Tested with HoloViews 1.3.2 (rev. 72d0b41).

```
%reload_ext holoviews.ipython
from holoviews import Curve
```
```
%%opts Curve [fontsize={'labels':30}]
Curve([(1, 1), (2, 2)])
```

jlstevens:
I am fairly sure HoloViews used to warn about invalid options in the past and that we explicitly decided to get rid of this behavior. The reason is that when you switch plotting backends, the valid set of options changes too (e.g switching from matplotlib to bokeh). 

In order to allow the same notebook to work  when switching backends, we decided that invalid options would definitely not cause errors. We also decided not to generate warnings because we thought that could get annoying (switching backend and rerunning  a notebook would generate a ton of warnings). As long as the keywords don't clash, you can currently style your cell for multiple backends without too much trouble.

If you don't consider the different backends (i.e matplotlib only), I definitely agree warnings would be a good idea. Maybe there is still a good way to support this? Perhaps, we could consider a 'validation' option for the output magic? E.g ``validation=None``, ``validation='warn'`` and maybe ``validation='error'``? 

Although there is a good reason for the current behavior, that doesn't mean  we can't improve it to also address this issue (suggestions welcome!)

philippjfr:
I don't think that's accurate, we absolutely do still warn if the style or plot option you supply is invalid. The issue here is the validation of the actual value that has been supplied, in this case the dictionary of fontsizes, apart from the direct validation we get from param, i.e. if the type doesn't match, we've never had validation for the values but probably should. This also fits in with the suggestion of tab-completing the values brough up in #16. To do both validation and completion robustly would probably be a lot of work, but it should definitely be on our roadmap.

maxalbert:
@jlstevens Thanks for the explanation! Yes, I can see that warnings don't mix well with switching backends and can easily get annoying in this case. I also agree that invalid options should not cause errors, but I'd highly be  in favour of a good warning system. From what I can see the key use cases are:

(i) New users who are trying to get to grips with Holoviews and don't know about the valid options yet.

[...]

PR #446, opened 2016-02-04
Improved validation of fontsize keys

jlstevens:
This PR addresses an issue raised in #291.

There it was pointed out that there is no warning when an invalid fontsize key is supplied and such a warning could help users discover which keys are valid.


jlstevens:
I think this is ready for merging now.

PR #447, opened 2016-02-05
Fixed points scaling in matplotlib and bokeh

philippjfr:
This PR supercedes @maxalbert's PR #390, by switching the function to return the area of points and for bokeh to convert those areas into widths. Changes a few tests but should be ready for merge once the data is updated.

### DIFF
--- a/doc/Tutorials/Columnar_Data.ipynb
+++ b/doc/Tutorials/Columnar_Data.ipynb
@@ -853,7 +853,7 @@
    },
    "outputs": [],
    "source": [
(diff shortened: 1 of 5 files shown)


philippjfr:
All tests passing, ``scaling_method`` now works correctly and matplotlib and bokeh points scale in the same way. Ready to merge.

jlstevens:
As I understand it, this really makes point scaling more sensible and consistent across the backends. As the tests are passing, I'll merge.

PR #451, opened 2016-02-05
Fixed matplotlib figure closing bug

philippjfr:
Fixes a small bug I introduced recently causing figures not to be closed correctly. Should cause no failures.


philippjfr:
Seems to be fine, will merge.

PR #454, opened 2016-02-05
RGB: more flexibility

vascotenner:
RGB: Allow also NdOverlay as input. When ranges are not set, assume normalized data.


jlstevens:
Seems reasonable to allow ``NdOverlays`` although I find it a bit odd that you would be using ``NdOverlays`` in the first place. 

With images, generally you only see the top layer so overlaying images isn't useful (they can't have alpha channels like RGBA) and ``im1 * im2 * im3`` seems more convenient than using the ``NdOverlay`` constructor. I suppose one reasonable argument is that people might like the indexing of ``NdOverlay`` more than ``Overlay``.

In other words, I don't object to the overall idea but I am really not sure I am happy about assuming that the ranges are normalized if not set. Normalization is a really tricky issue that needs a lot of care (generally we never normalize data in the elements if we can avoid it).

vascotenner:
I agree that Overlays are a bit strange to use. Maybe every indexable Container would do? HoloMap, Layout etc.

For the normalization I agree that this is tricky. It might be good to use the same normalization that should be used by default the specific imaging type. Or is that also tricky because this setting depends backend?

On the other hand, if it goes wrong, than the user can set it via the range option!

philippjfr:
Sorry we've let this linger so long. Could you rebase against master, then we can finally get this merged.

vascotenner:
I am still no git master yet. I hope I did not mess up git again. But the rebase should be there now.

philippjfr:
Something did go screwy there, can't quite figure out what though. I think you merged master in rather than rebasing it. Think you could hard reset to current master and then reapply your changes?

philippjfr:
Superceded by #759

Issue #421, opened 2016-01-26
reload(sys) and help


vascotenner:
When trying to use unicode characters in plots, the hv.help functions breaks down.

To use unicode I add the following lines to my notebook:
```python
from __future__ import unicode_literals, division
import sys; reload(sys)
sys.setdefaultencoding("utf-8")
```

However, the reload(sys) breaks the help, hv.help is not shown in the browser, but in the terminal where jupyter notebook is running

Works:
```python
from __future__ import unicode_literals, division
import sys
%matplotlib inline
import holoviews as hv
hv.notebook_extension()
hv.help(hv.Image)
```

Does not work:
```python
from __future__ import unicode_literals, division
import sys; reload(sys)
%matplotlib inline
import holoviews as hv
hv.notebook_extension()
hv.help(hv.Image)
```

jlstevens:
If I remember, people use ``reload(sys)`` to set unicode in Python 2. I assume that is what you are doing...

Have you tried setting the ``PYTHONIOENCODING`` environment variable to ```'utf8'``? Then you could just stick it into your bashrc and not worry about doing this in Python all the time.

vascotenner:
When setting the PYTHONIOENCODING environment variable to 'utf8' I can print unicode characters, however, I cannot create plots with unicode characters:

```hv.Points([1,1], kdims=['x',hv.Dimension('γ', unit='α')])```

```python
---------------------------------------------------------------------------
UnicodeEncodeError                        Traceback (most recent call last)
/usr/local/lib/python2.7/dist-packages/IPython/core/formatters.pyc in __call__(self, obj)
    695                 type_pprinters=self.type_printers,
    696                 deferred_pprinters=self.deferred_printers)
--> 697             printer.pretty(obj)
    698             printer.flush()
    699             return stream.getvalue()

/usr/local/lib/python2.7/dist-packages/IPython/lib/pretty.pyc in pretty(self, obj)
    381                             if callable(meth):
    382                                 return meth(obj, self, cycle)
--> 383             return _default_pprint(obj, self, cycle)
    384         finally:
    385             self.end_group()

/usr/local/lib/python2.7/dist-packages/IPython/lib/pretty.pyc in _default_pprint(obj, p, cycle)
    501     if _safe_getattr(klass, '__repr__', None) not in _baseclass_reprs:
    502         # A user-provided repr. Find newlines and replace them with p.break_()
--> 503         _repr_pprint(obj, p, cycle)
    504         return
[...]

PR #455, opened 2016-02-05
Unicode issues

jlstevens:
This PR aims to address issue #421 as well as a few other unicode issues.


jlstevens:
In my local testing, these commits do address issue #421 and I *think* they are correct improvements overall. I had to update the test data (strings became byte strings, prefixed with ``b'`` in a few places) so if that is ok, I think we can merge.

philippjfr:
Sounds alright, there's still more work to be done to make sure we're handling unicode correctly everywhere but I think this is an improvement and addresses a few existing issues.

PR #457, opened 2016-02-05
Reimplemented Layout aspects handling

philippjfr:
The previous PR improved aspect handling in Layouts but did not really fix it. This new implementation actually does what it's supposed to. I'd appreciate it if you could try some complex layout arrangements and then fiddling with the aspects of the individual components.

Here's some that I've tried:

```python
%%opts Layout [aspect_weight=1.0 absolute_scaling=True] Image.A [aspect=1.5] Image.C [aspect=0.5]
np.random.seed(42)
(hv.Image(np.random.rand(25, 25), group='A') +
 hv.Image(np.random.rand(25, 25), group='B') +
 hv.Image(np.random.rand(25, 25), group='C'))
```

image

Here absolute scaling controls whether the plots are scaled in absolute or relative terms, i.e. in the example relative scaling means that plot A (aspect=1.5) is the width of a usual plot and everything else is scaled down relative to that and absolute scaling means plot A is actually 1.5 times wider than it would usually be.

A more complex example is this:

```python
%%opts Layout [aspect_weight=1 sublabel_format='']
%%opts Image [aspect=1] Area.A [invert_axes=True aspect=0.4] Area.B [aspect=3]
img_a = hv.Image(np.random.rand(25, 25), group='A')
img_b = hv.Image(np.random.rand(25, 25), group='B')
img_a.hist() + hv.Area(img_b.reduce(x=np.mean), group='A') + img_b + hv.Area(img_a.reduce(x=np.mean), group='A') +\
hv.Area(img_b.reduce(y=np.mean), group='B').hist() + hv.Empty() + hv.Area(img_b.reduce(y=np.mean), group='B') + hv.Empty()
```

image

### DIFF
--- a/holoviews/plotting/mpl/plot.py
+++ b/holoviews/plotting/mpl/plot.py
@@ -17,6 +17,7 @@ from ...core import traversal
 from ..plot import DimensionedPlot, GenericLayoutPlot, GenericCompositePlot
 from ..util import get_dynamic_mode, initialize_sampled
 from .renderer import MPLRenderer
+from .util import compute_ratios
 
 
 class MPLPlot(DimensionedPlot):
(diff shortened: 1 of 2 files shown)


philippjfr:
Good to go I think. It's still not perfect but it's the most robust implementation I've gotten so far. Hopefully by 1.5 or 1.6 we'll have redesigned specifying and plotting layouts and won't have to rely on trying to infer a good arrangement based on the aspects of the individual plots.

jlstevens:
I do think this is a big improvement. No doubt we'll encounter new bugs but it does look like the behaviour is going to be a lot better on average. Merging.

Issue #389, opened 2016-01-01
Better warning message for dimension_values


jlstevens:
If I run this line where ``profile`` is a ``Curve`` with key dimension 'Duration':

```python
profile.dimension_values('Time')
```

I get the following error which isn't very helpful:

```python
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-12-0ed0e938563e> in <module>()
----> 1 profile.dimension_values('Time')

/Users/jstevens/Desktop/development/holoviews/holoviews/core/data.pyc in dimension_values(self, dim, unique)
    342         values are requested will return only unique values.
    343         """
--> 344         dim = self.get_dimension(dim).name
    345         dim_vals = self.interface.values(self, dim)
    346         if unique:

AttributeError: 'NoneType' object has no attribute 'name'
```

It would be good to report an error that states that the key dimension doesn't exist and maybe list the names of the key dimensions that do exist.

jlstevens:
This issue has been resolved in the PR mentioned above (now merged).

PR #463, opened 2016-02-06
Better exceptions in dimension_values

jlstevens:
This PR aims to address #389. I'm starting off by changing ``get_dimension`` as that would be a much smaller change than updating all the ``dimension_values`` methods. Note, this PR is the same as #450 which was automatically closed by GitHub when I tried to do a clean update with master.


philippjfr:
We'll have to do some more work on this for 1.5 but it's already a huge improvement and the PR build is passing so I'll go ahead and merge.

PR #471, opened 2016-02-08
Improvements for slider widget

philippjfr:
Slider widget now hides dependent dimensions which have a 1-to-1 mapping to previous dimension, like in the Exploring Data example where each frame maps onto a single date.


jlstevens:
Changes look sensible to me and I'm glad we can fix this. I think we should consider pushing a fixed version of 1.4.2 to pypi/conda. Together with the dynamic map sliders bug (now fixed) I don't think 1.4.2 should be left as it is.

philippjfr:
Agreed.

philippjfr:
Ready for merge.

PR #472, opened 2016-02-08
Fixed matplotlib Point clims and sizes

philippjfr:
Title says it all. Simple fix, turns out plt.scatter accepts clim but really expects the ranges to be set via vmin and vmax.


jlstevens:
Ok. The changes look fine and I've been asked to merge.

Issue #473, opened 2016-02-08
Escape characters confuse the escaping code


philippjfr:
Just came across a bug in the dimension escaping code. When constructing an NdOverlay like this:

```python
ndoverlay = hv.NdOverlay({i: hv.Contours([np.random.rand(20,2)]) for i in range(10)},
                         kdims=[hv.Dimension('$\alpha$')])
ndoverlay
```

I get the following error:

```
  File "/Users/prudiger/topographica/external/holoviews/holoviews/plotting/plot.py", line 501, in _get_frame
    selection = self.hmap.select((HoloMap, DynamicMap), **select)
  File "/Users/prudiger/topographica/external/holoviews/holoviews/core/dimension.py", line 722, in select
    select[self.get_dimension_index(dim)] = val
  File "/Users/prudiger/topographica/external/holoviews/holoviews/core/dimension.py", line 652, in get_dimension_index
    (dim, self.__class__.__name__))
Exception: Dimension Frame not found in HoloMap.
```

This seems to be because:

```python
ndoverlay.kdims
```

```
[Dimension('$\x07lpha$')]
```

And the ``\x07`` unicode character cannot be escaped by the escaping code in:

```
File "/Users/prudiger/topographica/external/holoviews/holoviews/core/util.py", line 249, in shortened_character_name
name = unicodedata.name(c).lower()
ValueError: no such name
```

jlstevens:
Ok, thankfully this issue was quick to identify.

The problem is that ``\a`` is an escape code, which according to [this](http://python-reference.readthedocs.org/en/latest/docs/str/escapes.html) is ``ASCII Bell (BEL)``. The solution is to use a raw string i.e:

```python
ndoverlay = hv.NdOverlay({i: hv.Contours([np.random.rand(20,2)]) for i in range(10)},
                         kdims=[hv.Dimension('$\alpha$')])
```

This is a tricky gotcha right now. I suppose I can try to detect escape codes, apparently I can do this by checking if ``unicodedata.type`` is ``"Cn"``. After that, it would need some sort of special handling to undo the escape codes (but not all of them).

So in short, I don't think this is a bug as such but maybe a feature request. I'll have a quick look to see what I might be able to do...

jlstevens:
I'll note that ``'$\\alpha$'`` does work so maybe that is what I should aim to do if I detect an escape code.

jlstevens:
[...]

PR #476, opened 2016-02-08
Improved exception for control codes

jlstevens:
This PR aims to address issue #473 by raising an exception early if any of the '\a','\b','\v','\f' or '\v' control codes are detected. These control codes are almost never used and cause errors in matplotlib/holoviews when users try using something simple like ``'$\alpha$'`` or ``'$\beta$'``.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -246,7 +246,14 @@ class sanitize_identifier_fn(param.ParameterizedFunction):
         (as a list of tokens) by applying the eliminations,
         substitutions and transforms.
         """
-        name = unicodedata.name(c).lower()
+        ccmap = {'\a':unicodedata.name(u'a')}
+        if unicodedata.category(c) == 'Cc': # Handle control codes
+            invalid = {'\a':'a','\b':'b','\v':'v','\f':'f','\r':'r'}
+            if c in invalid:
+                raise Exception("Please use a raw string or escape control code '\%s'"
+                                % invalid[c])
+        else:
+            name = unicodedata.name(c).lower()
         # Filtering
         for elim in eliminations:
             name = name.replace(elim, '')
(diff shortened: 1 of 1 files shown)


holoviews/core/util.py
```diff
@@ -246,7 +246,14 @@ def shortened_character_name(self_or_cls, c, eliminations=[], substitutions={},
         (as a list of tokens) by applying the eliminations,
         substitutions and transforms.
         """
-        name = unicodedata.name(c).lower()
+        ccmap = {'\a':unicodedata.name(u'a')}
```
philippjfr: Is ccmap ever used?

holoviews/core/util.py
```diff
@@ -246,7 +246,14 @@ def shortened_character_name(self_or_cls, c, eliminations=[], substitutions={},
...
+        if unicodedata.category(c) == 'Cc': # Handle control codes
```
philippjfr: You mentioned the category is 'Cn' earlier, is 'Cc' correct?

philippjfr:
Made two small comments. Otherwise looks good.

holoviews/core/util.py
jlstevens: Oops. That is a stray declaration from an earlier test. Removing it now.

holoviews/core/util.py
jlstevens: See [here](http://stackoverflow.com/questions/4324790/removing-control-characters-from-a-string-in-python). Seems 'Cc' is correct and makes sense to me if it stands for 'control code'.

jlstevens:
If you review the code and think it is ok this PR can be merged.

philippjfr:
Looks good to me now. Will merge.

PR #477, opened 2016-02-08
Small Bokeh tutorial/code improvements

philippjfr:
Minor fixes for the Bokeh tutorials removing additional Element examples, which do not yet work, i.e. sidebar of ErrorBars and overlay of VectorFields. Also added bootstrap based warning messages to warn that Tutorials can take a minute to render because there's so many plots on the page.

### DIFF
--- a/doc/Tutorials/Bokeh_Backend.ipynb
+++ b/doc/Tutorials/Bokeh_Backend.ipynb
@@ -7,6 +7,10 @@
     "id": "e6a399c2-931d-445e-a80a-30ea27653f27"
    },
    "source": [
+    "<div class=\"alert alert-info\" role=\"alert\">\n",
+    "    This tutorial contains a lot of bokeh plots, which may take a little while to load and render.\n",
+    "</div>\n",
+    "\n",
     "One of the major design principles of HoloViews is that the declaration of data is completely independent from the plotting implementation. This means that the visualization of HoloViews data structures can be performed by different plotting backends. As part of the 1.4 release of HoloViews, a [Bokeh](http://bokeh.pydata.org) backend was added in addition to the default ``matplotlib`` backend. Bokeh provides a powerful platform to generate interactive plots using HTML5 canvas and WebGL, and is ideally suited towards interactive exploration of data.\n",
     "\n",
     "By combining the ease of generating interactive, high-dimensional visualizations with the interactive widgets and fast rendering provided by Bokeh, HoloViews becomes even more powerful.\n",
--- a/doc/Tutorials/Bokeh_Backend.ipynb
+++ b/doc/Tutorials/Bokeh_Backend.ipynb
@@ -471,7 +471,7 @@
     "heatmap_data = {(chr(65+i),chr(97+j)):i*j for i in range(5) for j in range(5) if i!=j}\n",
     "data = [np.random.normal() for i in range(10000)]\n",
     "hist = np.histogram(data, 20)\n",
-    "hv.Points(error, vdims=['Error']) + hv.HeatMap(heatmap_data) + hv.Histogram(hist)"
+    "hv.Points(error) + hv.HeatMap(heatmap_data).sort() + hv.Histogram(hist)"
    ]
   },
   {
@@ -503,8 +503,8 @@
    },
    "outputs": [],
    "source": [
-    "%%opts Points [tools=['box_select', 'lasso_select', 'poly_select']] (s=10 unselected_color='red' color='blue')\n",
-    "hv.Points(error, vdims=['Error'])"
+    "%%opts Points [tools=['box_select', 'lasso_select', 'poly_select']] (size=10 unselected_color='red' color='blue')\n",
+    "hv.Points(error)"
    ]
   },
   {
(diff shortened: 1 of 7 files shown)


doc/Tutorials/Bokeh_Backend.ipynb
```diff
@@ -7,6 +7,10 @@
     "id": "e6a399c2-931d-445e-a80a-30ea27653f27"
    },
    "source": [
+    "<div class=\"alert alert-info\" role=\"alert\">\n",
+    "    This tutorial contains a lot of bokeh plots, which may take a little while to load and render.\n",
+    "</div>\n",
+    "\n",
```
jlstevens: This does help but I was also thinking of submitting an issue to suggest some sort of loading placeholder for Bokeh plots. Would this be feasible/worthwhile?

holoviews/plotting/bokeh/__init__.py
```diff
@@ -116,11 +118,11 @@
 
 # Charts
 options.Curve = Options('style', color=Cycle(), line_width=2)
-options.Scatter = Options('style', color=Cycle())
+options.Scatter = Options('style', color=Cycle(), size=np.sqrt(6))
 options.ErrorBars = Options('style', color='black')
 options.Spread = Options('style', fill_color=Cycle(), fill_alpha=0.6, line_color='black')
 options.Histogram = Options('style', fill_color="#036564", line_color="#033649")
-options.Points = Options('style', color=Cycle())
+options.Points = Options('style', color=Cycle(), size=np.sqrt(6))
```
jlstevens: ``np.sqrt(6)``? I assume you are trying to match something but it might be worth having a comment stating what exactly.

holoviews/plotting/bokeh/element.py
```diff
@@ -538,17 +538,44 @@ def __init__(self, element, plot=None, **params):
 
     def initialize_plot(self, ranges=None, plot=None, plots=None):
         self.mplplot.initialize_plot(ranges)
-        plot = mpl.to_bokeh(self.mplplot.state)
+
```
jlstevens: How many ``BokehMPLWrapper`` plots are there? These plots depend on both matplotlib **and** bokeh. This isn't typically a problem but I would like to know how many of these there are.

holoviews/plotting/bokeh/util.py
```diff
@@ -177,3 +179,33 @@ def hsv_to_rgb(hsv):
     rgb = clist[order[i], np.arange(np.prod(shape))[:,None]]
 
     return rgb.reshape(shape+(3,))
+
+
+def update_plot(old, new):
```
jlstevens: Seems like something that would ideally be offered by Bokeh itself (if I understand it right).

jlstevens:
I've made a few comments but otherwise it looks ok to me. That said, there are a number of code changes that I assume will affect more than just the Bokeh tutorial.

holoviews/plotting/bokeh/element.py
philippjfr: This is a general purpose wrapper class so there's only this one and the one that converts matplotlib plots to RGB images (which we should probably remove). Currently we only register the seaborn Element types for actual conversion and have the matplotlib 3D plots registered to use the RGB conversion. I'll actually go ahead and delete these:

```                Surface: PlotSelector(lambda x: 'bokeh',
                                      [('mpl', SurfacePlot),
                                       ('bokeh', BokehMPLRawWrapper)], True),
                Scatter3D: PlotSelector(lambda x: 'bokeh',
                                        [('mpl', Scatter3DPlot),
                                         ('bokeh', BokehMPLRawWrapper)], True),
                Trisurface: PlotSelector(lambda x: 'bokeh',
                                         [('mpl', TrisurfacePlot),
                                          ('bokeh', BokehMPLRawWrapper)], True)},
```

 as part of this PR.

holoviews/plotting/bokeh/__init__.py
philippjfr: True, will add it. Point sizes in bokeh are widths while matplotlib uses area and has a default of 6.

holoviews/plotting/bokeh/util.py
philippjfr: Agreed, they're working on it. Will make a note that it should be replaced after bokeh 0.12.

doc/Tutorials/Bokeh_Backend.ipynb
philippjfr: Does sound nice but I also don't know how feasible it is.

holoviews/plotting/bokeh/element.py
jlstevens: Ok great.

There probably isn't much value in having Bokeh wrap matplotlib as you will always be limited by the speed of rendering on the matplotlib side. In general nesting backends probably doesn't make sense and it is probably best to simply state that a particular element isn't supported by a given backend.

philippjfr:
Ready to merge now.

jlstevens:
Ok. Tests are passing so I'll merge now.

PR #478, opened 2016-02-09
Ensured initial dynamicmap values are escaped correctly in widgets

philippjfr:
The widgets require slightly different escaping when using DynamicMap, this PR ensures they are escaped correctly to avoid a bug where all sliders would have to be moved from their default position for DynamicMap to work.


jlstevens:
Doesn't look too bad. I'm happy to merge as soon as the tests pass.

philippjfr:
Ready to merge.

jlstevens:
Ok. Merging now.

PR #480, opened 2016-02-09
Allowed DynamicMaps to work with dimension subsets

philippjfr:
Allows DynamicMaps to be used as long as their dimensions are overlapping. In sampled mode now raises exception if the DynamicMap dimensions aren't a subset of the HoloMap dimensions, which define the sampling.

### DIFF
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -460,8 +460,10 @@ class GenericElementPlot(DimensionedPlot):
             self.current_key = key
             return self.current_frame
         elif self.dynamic:
+            dims = {d.name: k for d, k in zip(self.dimensions, key)
+                    if d in self.hmap.kdims}
             if isinstance(key, tuple):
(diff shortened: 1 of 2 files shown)


jlstevens:
Changes look ok to me.

Ideally we'll be able to lift this restriction in future. What exactly would that take? In the documentation (i.e the DynamicMap tutorial), should I say this is a 'current' restriction or simply a 'restriction'?

philippjfr:
I agree, I think we can lift the dimension subset restriction more generally. Each HoloMap/DynamicMap simply gets the keys for the dimensions it has and ignores any others. I don't actually think there's any major obstacle to this.

Making a mixture of continuous vs. sampled dimensions work is a bit more complicated but in the end should also be possible. For now just say it's a current restriction.

philippjfr:
Tests are passing, ready to merge.

jlstevens:
Merging.

PR #483, opened 2016-02-09
Validating dimension ranges in dynamic sampled mode

philippjfr:
As the title says this makes sure that when plotting a DynamicMap in sampled mode, i.e. next to a HoloMap that provides the sampling, that all the keys are actually within the ranges specified on the DynamicMap dimensions.


jlstevens:
Tests are passing and I believe it is ready to merge.

PR #486, opened 2016-02-10
MPLWidgets now always display first frame from cache

philippjfr:
As the title says the matplotlib widgets now display the first frame from the cache even in dynamic/live mode. This allows exported examples of DynamicMaps to at least display the initial image.


jlstevens:
This is @philippjfr very fast response to the issues with the tutorial in #470. :-)

I'm happy to merge whenever you are ready and I'll test it both with a local export and with a website build later tonight.

philippjfr:
Go ahead and merged. Very happy there was such an easy (and correct) fix for the issue.

PR #487, opened 2016-02-10
Avoid encoding matplotlib data multiple times

philippjfr:
Fixes minor bug causing matplotlib plots to be unnecessarily encoded multiple times.


jlstevens:
Sounds like a potential performance improvement. Tests are passing so I'll merge.

PR #488, opened 2016-02-10
Implemented slicing semantics for bounded DynamicMaps

jlstevens:
If you start using slices in bounded mode, you have to use slices throughout.

Slicing sets/unsets the dimension soft_ranges which can be used to restrict the explorable region. I'll be documenting this in the new Dynamic Map tutorial.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -551,6 +551,21 @@ class DynamicMap(HoloMap):
         return self.clone(data)
 
 
+    def _slice_bounded(self, tuple_key):
+        """
+        Slices bounded DynamicMaps by setting the soft_ranges on key dimensions.
+        """
+        cloned = self.clone(self)
+        for i, slc in enumerate(tuple_key):
+            (start, stop) = slc.start, slc.stop
+            if start and start < cloned.kdims[i].range[0]:
+                raise Exception("Requested slice below defined dimension range.")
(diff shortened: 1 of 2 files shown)


jlstevens:
Seems to work fine in the examples I've tried. If it looks ok to you, could you please merge?

holoviews/core/spaces.py
```diff
@@ -551,6 +551,21 @@ def _cross_product(self, tuple_key, cache):
...
+    def _slice_bounded(self, tuple_key):
+        """
+        Slices bounded DynamicMaps by setting the soft_ranges on key dimensions.
+        """
+        cloned = self.clone(self)
+        for i, slc in enumerate(tuple_key):
+            (start, stop) = slc.start, slc.stop
+            if start and start < cloned.kdims[i].range[0]:
+                raise Exception("Requested slice below defined dimension range.")
+            if stop and stop > cloned.kdims[i].range[1]:
```
philippjfr: Make sure you check ``start is not None`` and ``stop is not None`` explicitly unless it'll fall through for zeros.

philippjfr:
Made one comment, looks fine otherwise.

holoviews/core/spaces.py
jlstevens: Good point. Fixed in the last commit.

philippjfr:
Looks good, will merge as soon as tests pass.

PR #490, opened 2016-02-10
Website building saves widget data to json

philippjfr:
This PR updates the website building submodules which include changes to save widget data to json. This means the main JS thread doesn't have to parse all the data at once when loading the website, instead setting up independent async calls to load the data for each widget, which should speed up the download of larger tutorials.


philippjfr:
This works now, you can see the website with the data loaded from json here: http://hvtesting.s3-website-eu-west-1.amazonaws.com/

You should notice that the tutorials load considerably faster.

philippjfr:
It also avoids a bug I encountered quite a bit for large tutorials, where only part of the page would load.

jbednar:
Excellent!  Yes, I'd encountered that too, even on the homepage (when on a mobile network).  Hope this helps!

jlstevens:
I'm happy to merge as no one should worry too much about how the website is powered. It will be nice to have the improved performance!

I do have some general questions:

* How do we make this available to other people who want to use JSON files on their website? What is the current API to use this?
* How similar is this approach to what the Jupyter people are looking into for storage? Maybe there will be a way to standardize?
* Can this be used without a web server? I assume the same origin policy won't let this work unless a web server is running.

Other than those general queries, I am happy to merge once you say it is ready.

philippjfr:
>How do we make this available to other people who want to use JSON files on their website? What is the current API to use this?

Currently the API isn't as convenient as it might be. In our building profile I have put the following:

```

from holoviews.plotting.widgets import NdWidget
NdWidget.export_json=True
NdWidget.json_load_path = '/json'
NdWidget.json_save_path = './'
```

This saves the json files in the same folder as the ipynb that generated them. You can also just specify a save path, which means that the files always have to be in the right location relative to the notebook, e.g.:

```
from holoviews.plotting.widgets import NdWidget
NdWidget.export_json=True
NdWidget.json_save_path = '.'
```

This example also works from within the notebook.

>How similar is this approach to what the Jupyter people are looking into for storage? Maybe there will be a way to standardize?

I don't know yet tbh, I don't think there's been much progress on that front but I'll have to double check.

>Can this be used without a web server? I assume the same origin policy won't let this work unless a web server is running.

Your intuition is correct this requires a live server because of same origin policies, so you can use it from within the notebook or on a webserver.

>What about where the JSON files are stored? I assume they need a naming scheme and have to be stored relative to the HTML.

As I showed above you have control over where they are stored but if you want it to work in the notebook and on a different server, i.e. you export a notebook you just ran, you have to use relative paths. For naming it currently just uses uuids because I've found no robust way of communicating the notebook name yet. I'm sure we can come up with something better though, but not for this release.

philippjfr:
Also, yes it's ready to merge.

PR #491, opened 2016-02-10
Re-raising exception in display hooks

jlstevens:
Previously the display hook would simply printing the exception traceback which allowed the notebook to continue executing instead of stopping at that cell.


jlstevens:
Tests are passing so if you think this is fine, please merge.

philippjfr:
Looks fine. Merging.

PR #492, opened 2016-02-10
Fixes for initialization of overlays returned from DynamicMaps

philippjfr:
As the title says, makes sure that initialization of DynamicMaps returning Overlays doesn't get confused.


philippjfr:
Ready to merge, bit of a hacky fix though, so I might file an issue to rethink how Overlays in DynamicMaps could be handled better.

jlstevens:
Merging.

PR #501, opened 2016-02-11
Change homepage static plot to use bokeh

philippjfr:
This PR changes the static example on the homepage to use bokeh. I think this makes the homepage a lot more compelling overall. I've changed the static example because it's easier to enable bokeh with notebook_extension and then switch with output magic. Secondly the image is rather large and matplotlib handles that better than bokeh for now. But the static example in itself makes a great showcase because it shows automatic linking of the x-axis across plots and this gets the point across how easy it is to switch. It's literally 1 extra line in the notebook.


philippjfr:
The image might still be a bit too big to display, so we'd probably have to downsample the array by half or so but visually that makes virtually no difference anyway since the array is much larger than the matplotlib plot.

jlstevens:
I think this is a good idea although I think it might be worth having at least one line in the text to explicitly mention the switch. It might already be mentioned in general but now this would be a concrete example of using multiple backends.

jbednar:
I think it might be a bit overwhelming to have the initial plot be so complex, with toolbars, dragging, etc.  I like the way the matplotlib version of the first plot is clean, simple, and looks publishable -- it establishes who this library is meant for (serious people!)  If we show the Bokeh one, it has more features and more interactivity, but those seem like secondary things we'd want to convey.  Plus it makes my browser hang at first, which is disconcerting.  Maybe this would be best handled by building the Bokeh version in a separate document, and provide a link underneath this figure that shows how easy it is to switch the backends, where we mention Bokeh already?

BTW, the default black line of annotations isn't very effective with our default Hot colormap; shouldn't we make it blue by default like in mpl?

jlstevens:
I think I agree with @jbednar as I wasn't hugely keen on introducing bokeh so quickly, even though I do agree with @philippjfr  that it is a nice feature that should be made more prominent than it is now.

When I saw this PR, I also thought of exactly the same idea: why not have a clearly visible link from the homepage to a bokeh equivalent to make the point that it is easy to switch backends?

philippjfr:
If the plot loaded quicker I'd disagree with you but as you point out it makes the page annoyingly slow to load. My issue with simply mentioning bokeh is that not that many people know about bokeh yet so simply mentioning it might be meaningless to some, so something should demonstrate what it can do for you and how easy it is to switch. Adding a link might be a good idea for now but we should think of some visually compelling way to demonstrate support for different backends and all the interactivity that provides.

jbednar:
Bokeh would be fine for the  second plot, where we want more razzle dazzle, if it weren't slow...

philippjfr:
Agreed, maybe in future I can detect that the Image is unchanged through the HoloMap and only send the data once.

jlstevens:
> Bokeh would be fine for the second plot, where we want more razzle dazzle, if it weren't slow...

That was what I thought initially but I want to make it clear that all the sliders and interaction are available with matplotlib. I still think a bokeh version of the homepage is a nice idea but we might also want to add a tutorial dedicated to the different backends at some point.

jlstevens:
I see you've closed the issue but I think it would be worth opening a new PR/issue to emphasise the ability to switch backends on the homepage (in whatever way we think works well).

doc/Homepage.ipynb
```diff
@@ -52,10 +53,11 @@
    },
    "outputs": [],
    "source": [
+    "%%output backend='matplotlib'\n",
     "%%opts Points [scaling_factor=50] Contours (color='w')\n",
     "dots = np.linspace(-0.45, 0.45, 19)\n",
     "\n",
-    "hv.HoloMap({y: (fractal * fractal.sample([(i,y) for i in dots]).to.points(['x','y'], 'z') +\n",
+    "hv.HoloMap({y: (fractal * hv.Points(fractal.sample([(i,y) for i in dots])) +\n",
```
jbednar: I like this change; I think it's easier to understand, and a bit shorter.  Can you please apply that to master?

philippjfr: Sure, will do.

PR #502, opened 2016-02-11
Bokeh Elements only send data once if completely static

philippjfr:
Often times when exploring some dataset using the widgets some Elements are completely static, while others change when dragging the slider such as in the homepage example where there is a large image that is completely static. Currently that data gets sent multiple times, this PR detects if a frame is completely static and ensures that the data is only sent when first initializing the plot. This speeds up and reduces the memory footprint of the homepage example by a factor of ~40.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -14,7 +14,8 @@ except ImportError:
     mpl = None
 import param
 
-from ...core import Store, HoloMap, Overlay, DynamicMap, CompositeOverlay
+from ...core import (Store, HoloMap, Overlay, DynamicMap,
+                     CompositeOverlay, Element)
 from ...core import util
 from ...element import RGB
 from ..plot import GenericElementPlot, GenericOverlayPlot
@@ -149,6 +150,8 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         self.current_ranges = None
         super(ElementPlot, self).__init__(element, **params)
         self.handles = {} if plot is None else self.handles['plot']
+        element_ids = self.hmap.traverse(lambda x: id(x), [Element])
+        self.static = len(set(element_ids)) == 1 and len(self.keys) == len(self.hmap)
 
 
     def _init_tools(self, element):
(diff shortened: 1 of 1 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -149,6 +150,8 @@ def __init__(self, element, plot=None, show_labels=['x', 'y'], **params):
         self.current_ranges = None
         super(ElementPlot, self).__init__(element, **params)
         self.handles = {} if plot is None else self.handles['plot']
+        element_ids = self.hmap.traverse(lambda x: id(x), [Element])
```
jlstevens: Traversing a holomap should be quick but won't be O(1). Still going to be a lost faster than what we had before though!

jlstevens:
I think I understand what is going on and this optimization makes sense. Though I think this is a special case of a more general optimization where you detect blocks of constant ids (e.g you could have one of ``id0`` followed by a million of ``id1``).

As the general optimization is more tricky and this will improve most common cases, I think this very worthwhile.

jlstevens:
I should also add that this optimization makes an assumption we use throughout HoloViews: elements are treated as immutable. If you modify an element with a given id, this optimization would be invalid.

This is worth mentioning as immutability is our default assumption but nothing *stops* people from mutating data in an element and I certainly can think of cases where this would be more efficient performance-wise.

philippjfr:
>That said, I do think this is a special case of a more general optimization where you detect blocks of constant ids (e.g you could have one of id0 followed by a million of id1).

Agreed, that's definitely an optimization worth considering, now that bokeh server has settled down a little bit we can consider hooking into their events API and designing a storage format which will allow us to dynamically update plot components on the front end. The simplest approach would probably to send a dictionary of references along with the minimized data. On each frame the widget looks up the appropriate references in the data store and updates the plot with the data.

philippjfr:
Ready to merge. Can open another issue to discuss how we can remove redundant data in static bokeh plots if you think it's worth it.

jlstevens:
I've just realized that mutation isn't an issue: once it is in the plotting code, the user won't be able to do such things. As the tests are passing, I'll merge now.

PR #506, opened 2016-02-12
Layout.clone copies id and cols settings

philippjfr:


jlstevens:
Makes sense. If this PR in response to a specific issue that was raised?

philippjfr:
Yes two of them, it seems Overlays don't retain their id on relabel for some reason (as reported on Gitter) and @jbednar complained that the cols were being reset when using ``__call__``.

philippjfr:
Not quite sure why there's no PR build on travis.

jlstevens:
The tests are passing so I'll merge it now.

PR #510, opened 2016-02-16
Matplotlib now supports custom projections as plot option

philippjfr:
This PR allows custom projections to be specified on an Element. Also includes a number of fixes to ensure that extents are computed correctly in those cases.

### DIFF
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -641,11 +641,11 @@ class OverlayPlot(LegendPlot, GenericOverlayPlot):
 
     def __init__(self, overlay, ranges=None, **params):
         if 'projection' not in params:
-            projs = self._deep_options(overlay, 'plot', ['projection'],
-                                       [Element])['projection']
+            projs = self._traverse_options(overlay, 'plot', ['projection'],
+                                           [Element])['projection']
             if len(set(projs)) > 1:
                 raise Exception("A single axis may only be assigned one projection type")
-            else:
+            elif len(projs):
                 params['projection'] = projs[0]
         super(OverlayPlot, self).__init__(overlay, ranges=ranges, **params)
 
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -640,8 +640,13 @@ class OverlayPlot(LegendPlot, GenericOverlayPlot):
     _passed_handles = ['fig', 'axis']
 
     def __init__(self, overlay, ranges=None, **params):
-        if overlay.traverse(lambda x: x, (Element3D,)):
-            params['projection'] = '3d'
+        if 'projection' not in params:
+            projs = self._deep_options(overlay, 'plot', ['projection'],
+                                       [Element])['projection']
+            if len(set(projs)) > 1:
+                raise Exception("A single axis may only be assigned one projection type")
+            else:
+                params['projection'] = projs[0]
         super(OverlayPlot, self).__init__(overlay, ranges=ranges, **params)
 
     def _finalize_artist(self, key):
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -641,13 +641,9 @@ class OverlayPlot(LegendPlot, GenericOverlayPlot):
 
     def __init__(self, overlay, ranges=None, **params):
         if 'projection' not in params:
-            projs = self._traverse_options(overlay, 'plot', ['projection'],
-                                           [Element])['projection']
-            if len(set(projs)) > 1:
-                raise Exception("A single axis may only be assigned one projection type")
-            elif len(projs):
-                params['projection'] = projs[0]
+            params['projection'] = self._get_projection(overlay)
         super(OverlayPlot, self).__init__(overlay, ranges=ranges, **params)
+
 
     def _finalize_artist(self, key):
         for subplot in self.subplots.values():
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -232,28 +232,36 @@ class ElementPlot(GenericElementPlot, MPLPlot):
         extents = self.get_extents(view, ranges)
         if extents and not self.overlaid:
             coords = [coord if np.isreal(coord) else np.NaN for coord in extents]
+            valid_lim = lambda c: util.isnumeric(c) and not np.isnan(c)
             if isinstance(view, Element3D) or self.projection == '3d':
                 l, b, zmin, r, t, zmax = coords
                 zmin, zmax = (c if util.isnumeric(c) and not np.isnan(c) else None
                               for c in (zmin, zmax))
-                if not zmin == zmax:
-                    axis.set_zlim((zmin, zmax))
+                if zmin != zmax:
+                    if valid_lim(zmin):
+                        axis.set_zlim(bottom=zmin)
+                    if valid_lim(zmax):
+                        axis.set_zlim(top=zmax)
             else:
                 l, b, r, t = [coord if np.isreal(coord) else np.NaN for coord in extents]
             if self.invert_axes:
                 l, b, r, t = b, l, t, r
-            l, r = (c if util.isnumeric(c) and not np.isnan(c) else None
-                    for c in (l, r))
+
             if self.invert_xaxis or any(p.invert_xaxis for p in subplots):
                 r, l = l, r
-            if not (l is None and r is None) and not l == r:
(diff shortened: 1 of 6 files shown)


holoviews/plotting/mpl/plot.py
```diff
@@ -963,11 +964,11 @@ def _create_subplots(self, layout, positions, layout_dimensions, ranges, axes={}
...
 
             # Determine projection type for plot
-            components = view.traverse(lambda x: x)
-            projs = ['3d' if isinstance(c, Element3D) else
-                     self.lookup_options(c, 'plot').options.get('projection', None)
-                     for c in components]
-            projs = [p for p in projs if p is not None]
+            projs = self._deep_options(view, 'plot', ['projection'],
+                                      [CompositeOverlay])['projection']
+            if projs:
```
jlstevens: Little confused...is this supposed to be ``if not projs``? Seems you would try to look for the options on the elements *if* they couldn't be found at the overlay level, especially since we only allow one projection anyway...

philippjfr: You're right.

holoviews/plotting/mpl/element.py
```diff
@@ -234,22 +234,25 @@ def _finalize_limits(self, axis, view, subplots, ranges):
...
-            if not l == r:
+            if not (l is None and r is None) and not l == r:
                 axis.set_xlim((l, r))
-            b, t = (c if np.isfinite(c) else None for c in (b, t))
+            b, t = (c if util.isnumeric(c) and not np.isnan(c) else None
+                    for c in (b, t))
             if self.invert_yaxis or any(p.invert_yaxis for p in subplots):
                 t, b = b, t
-            if not b == t:
+            if not (b is None and t is None) and not b == t:
```
jlstevens: Was there an issue with the components of the extent ``(l,b,r,t)``` values being ``None`` before? If so, this is actually a bug fix as None was supposed to be supported...

philippjfr: It depends on the type of axes a little bit, setting the actual xlim or ylim to None causes errors sometimes. I'll have to look at this again as I don't think the logic is quite correct.

holoviews/plotting/mpl/element.py
```diff
@@ -637,8 +640,13 @@ class OverlayPlot(LegendPlot, GenericOverlayPlot):
     _passed_handles = ['fig', 'axis']
 
     def __init__(self, overlay, ranges=None, **params):
-        if overlay.traverse(lambda x: x, (Element3D,)):
-            params['projection'] = '3d'
+        if 'projection' not in params:
+            projs = self._deep_options(overlay, 'plot', ['projection'],
+                                       [Element])['projection']
+            if len(set(projs)) > 1:
+                raise Exception("A single axis may only be assigned one projection type")
```
jlstevens: Is this check happening twice? There is an  ``if len(set(opts['projection'])) > 1:`` check in ``MPLPlot`` as well. I'm trying to think if there is one place that could handle both...

holoviews/plotting/mpl/plot.py
```diff
@@ -354,9 +352,12 @@ def _create_subplots(self, layout, axis, ranges, create_axes):
             # Create axes
             kwargs = {}
             if create_axes:
-                threed = issubclass(vtype, Element3D)
-                subax = plt.subplot(self._layoutspec[r, c],
-                                    projection='3d' if threed else None)
+                opts = self._deep_options(view, 'plot', ['projection'], [CompositeOverlay])
```
jlstevens: I think ``_deep_options`` is a useful method but I'm not really happy with the name.

holoviews/plotting/mpl/element.py
philippjfr: I can make it a little function but I can't make it happen in one place.

jlstevens: Ok, I'm happy to leave it as it is given that there is no single place this check can be carried out.

philippjfr: No you had a point, it was also calling traverse twice which is unnecessary so I've made ``traverse_options`` a bit more powerful and factored out a ``_get_projection`` method.

jlstevens:
There are now conflicts but I was happy to merge once you had checked that the lbrt logic was correct...

philippjfr:
I'll resolve the conflict soon. For now this solution is fine but I may open an issue to make some of the methods on Plot into utility functions.

holoviews/plotting/mpl/plot.py
```diff
@@ -82,8 +82,7 @@ class MPLPlot(DimensionedPlot):
     sublabel_size = param.Number(default=18, doc="""
          Size of optional subfigure label.""")
 
-    projection = param.ObjectSelector(default=None,
-                                      objects=['3d', 'polar', None], doc="""
+    projection = param.Parameter(default=None, doc="""
```
jlstevens: Do you validate the possible strings? I've not read the code below but we should make sure if a string is supplied it is validate...

jlstevens: Oh and I would like some information about what a projection specification is if it isn't None or one of the strings...

holoviews/plotting/plot.py
```diff
@@ -166,7 +166,9 @@ class DimensionedPlot(Plot):
         of plotting. Allows selecting normalization at different levels
         for nested data containers.""")
 
-    projection = param.ObjectSelector(default=None)
+    projection = param.Parameter(default=None, doc="""
```
jlstevens: Again, I would like the docstring to say a bit more about what is required to set a custom projection...

holoviews/plotting/mpl/plot.py
philippjfr: Doesn't validate yet but could do, although some matplotlib extensions allow new projection types to be specified as strings.

holoviews/plotting/plot.py
```diff
@@ -393,15 +395,66 @@ def _compute_group_range(group, elements, ranges):
...
-                            {o: cls.lookup_options(x, opt_type).options.get(o, None)
-                             for o in opts})
-        return dict(obj.traverse(lookup, specs))
+        in opts for the specified opt_type and specs. If
+        a keyfn is supplied the returned options will be
+        nested by the returned key. If defaults is True
+        it will also look up the default plot options
+        specified on the matching plot type, appending
+        the default values with ``_default``.
+        """
```
jlstevens: Docstring mostly makes sense though I don't get what keyfn is needed for. Also, shouldn't ``defaults=True`` just fill in options that aren't explicitly specified with the defaults instead of adding the '_default' suffix?

jlstevens:
I've made a few more comments but once those are addressed, I think this PR looks good.

holoviews/plotting/plot.py
philippjfr: keyfn can be get_spec to group the options by type.group.label spec, for the projection lookup I just use it to tell apart Overlays and Elements. I think you're right about defaults and what you suggest will work but I'll have to test it again.

holoviews/plotting/mpl/plot.py
philippjfr: To be honest creating the axes is one of the first things that happens and the traceback matplotlib provides is reasonable:

```python
ValueError [Call ipython.show_traceback() for details]
Unknown projection 'Test'
```

jlstevens:
Thanks for improving the docstrings!

At the ``DimensionedPlot`` level we now have a projection parameter even though we can't say concretely what it is. I would say it is a concept that cuts across backends but as far as I know, we only have a concrete example in matplotlib. Is that right? Do you know if plotly has this concept for instance?

philippjfr:
>I would say it is a concept that cuts across backends but as far as I know, we only have a concrete example in matplotlib. Is that right? Do you know if plotly has this concept for instance?

Yes, plotly has this concept, I think there it only provides control over cartesian or polar axes and 3d plots are a separate concept but we can smooth over that inconsistency with our API. I do think it's a general concept and the reason why it's a parameter at the Plot level is because some of general extents finding code has been written to handle 3d axes. In future hopefully Bokeh will also support custom projections including various geographic projections, polar plots and maybe 3d plots as well.

holoviews/plotting/mpl/plot.py
jlstevens: Yes, that is very reasonable.

philippjfr:
Ready to merge now. I'll open a new issue to discuss which bits on Plot can become utilities.

jlstevens:
Looks good!

PR #511, opened 2016-02-16
Matplotlib version compatibility fix for Scatter3D facecolors

philippjfr:
Matplotlib handling of colors continues to be the bane of my life, it seems that facecolor handling for 3d scatter plots has changed between 1.4.3. and 1.5.0 and that was the source of various transients. Hoping this will finally put that issue to rest and we can remove the handling once we upgrade to 1.5 or whenever we decide to stop supporting earlier versions.

### DIFF
--- a/holoviews/plotting/mpl/chart3d.py
+++ b/holoviews/plotting/mpl/chart3d.py
@@ -1,5 +1,8 @@
+from distutils.version import LooseVersion
+
 import numpy as np
 import param
+import matplotlib as mpl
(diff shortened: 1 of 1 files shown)


jlstevens:
> the bane of my life...

Yes -we are all sick of these damn color issues! 

As you say, hopefully we can forget about this once we can switch to matplotlib 1.5.

philippjfr:
Ready to merge once the tests are passing.

PR #515, opened 2016-02-18
Made color handling for all matplotlib plots consistent

philippjfr:
Small fix for ColorbarPlots, moving ``symmetric`` plot option to the base class where it is used.


jlstevens:
Looks good. Removing redundant parameter definitions is always a good thing!

philippjfr:
Agreed, although it was also an actual bug because some classes didn't have the parameter so it could fail in some cases. I've expanded the scope of the PR a little bit by consistently using the ColorbarPlot method to set color ranges so they all now support logarithmic and symmetric color ranges.

philippjfr:
Tests are passing and I confirmed the options actually work on the different plot types. Ready to merge.

jlstevens:
Great!

PR #518, opened 2016-02-21
Refactored matplotlib axis setting for easier reusage

philippjfr:
Not ready yet but this PR begins to refactor some of the code in setting up the axes limits, labels, ticks, aspect etc. to make reusing this code easier in custom extensions with different types of axes.

### DIFF
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -11,7 +11,7 @@ from ...core import (OrderedDict, Collator, NdOverlay, HoloMap, DynamicMap,
                      CompositeOverlay, Element3D, Columns, NdElement, Element)
 from ...element import Table, ItemTable
 from ..plot import GenericElementPlot, GenericOverlayPlot
-from ..util import dynamic_update
+from ..util import dynamic_update, dim_axis_label
 from .plot import MPLPlot
 from .util import wrap_formatter
 
@@ -208,11 +208,19 @@ class ElementPlot(GenericElementPlot, MPLPlot):
         and Elements. Sets the axes position as well as tick positions,
         labels and fontsize.
         """
+        ndims = len(dimensions)
+        xdim = dimensions[0] if ndims else None
+        ydim = dimensions[1] if ndims > 1 else None
+
         # Tick formatting
-        self._set_axis_formatter(axis.xaxis, dimensions[0])
-        self._set_axis_formatter(axis.yaxis, dimensions[1])
+        if xdim:
+            self._set_axis_formatter(axis.xaxis, xdim)
+        if ydim:
+            self._set_axis_formatter(axis.yaxis, ydim)
         if self.projection == '3d':
-            self._set_axis_formatter(axis.zaxis, dimensions[2])
+            zdim = dimensions[2] if ndims > 2 else None
+            if zdim:
+                self._set_axis_formatter(axis.zaxis, zdim)
 
         xticks = xticks if xticks else self.xticks
         self._set_axis_ticks(axis.xaxis, xticks, log=self.logx,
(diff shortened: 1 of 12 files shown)


jlstevens:
All these changes look good.

@philippjfr  That said, can we agree to have a freeze on the plotting code once this is merged?

I really feel we need a chance to document as much of the API (including internal API) that you consider stable. This will take the pressure off you and allow other people to start contributing by writing their own plotting classes. I don't think you should be the sole maintainer of the plotting module forever!

In other words, I think all these refactors have pushed your current [bus factor](https://en.wikipedia.org/wiki/Bus_factor) a bit too high. :-)

jlstevens:
I would also like to link to PR #438. That seems the right place to start thinking about the API in order to finally document it properly.

philippjfr:
Yes, there will be a freeze on the plotting code until we have sufficiently documented the current implementation. Both the matplotlib and bokeh element plots should now be pretty stable and after documenting both we can consider how to improve other aspects of the plotting code including improved layouts, unifying some of the options across backends and then finally document how to create an entire backend.

philippjfr:
The current test failures are as follows and are all expected:

* Columnar Data:
  - 15/16/22: Axis label changes
* Composing_Data:
  - 0/3: GridSpace gridlines reenabled
* Containers:
  - 9: GridSpace gridlines reenabled
* Elements:
  - 8/10: Axis label changes
  - 11/21: Tiny axis frame change

The axis label change is to provide more consistent handling for axes representing more than one dimension (e.g. on Bars and ErrorBars types). The GridSpace line change reverts an earlier change where I'd accidentally removed gridlines completely when the axis itself is hidden. The axis frame change is down to adjoined plots, where the edge where the adjoined plot attaches was automatically added even if not specified. I'll rebuild the data soon and then this PR is ready to review and merge.

philippjfr:
Tests passing, ready to review.

holoviews/plotting/mpl/element.py
```diff
@@ -105,7 +107,8 @@ class ElementPlot(GenericElementPlot, MPLPlot):
     # Element Plots should declare the valid style options for matplotlib call
     style_opts = []
 
-    _suppressed = [Table, NdElement, Collator, Columns, ItemTable]
+    # Whether plot has axes, disables setting axis limits, labels and ticks
+    _axes = True
```
jlstevens: I would consider renaming this attribute ``_has_axes`` to indicate it is a boolean predicate.

holoviews/plotting/util.py
```diff
@@ -1,10 +1,12 @@
+from __future__ import unicode_literals
```
jlstevens: Not sure how this import relates to the PR but seems like a good idea anyway...

holoviews/plotting/mpl/element.py
philippjfr: Good plan, I considered adding a separate baseclass, but didn't want to change too much.

holoviews/plotting/util.py
```diff
@@ -1,10 +1,12 @@
...
+
```
philippjfr: The separator declared in the function below has to be unicode.

jlstevens:
I've had a look through and I am happy to merge if you are happy with it. My only other suggestion is to run flakes on your branch because there is quite a bit changed and flakes might help you catch bugs.

philippjfr:
Good plan, I'll apply your ``has_axes`` suggestion and run a few linters over it.

philippjfr:
Made that fix and ran pyflakes, apart from superfluous imports there were a few issues that were introduced in a previous PR I think.

jlstevens:
Ok..the tests are passing. Merging!

PR #523, opened 2016-02-27
Developer Manual section documenting plotting API

jlstevens:
It is time for us to slow down the changes to the internal plotting API so we can document the plotting system. The purpose of this proposed documentation is to help explain how a plotting class is structured to allow new developers to extend HoloViews themselves.

Although documenting the plotting system is the biggest job and highest priority, we will probably need a section on adding new elements as well in order for this to be useful. I imagine this documentation will serve as a guide to the core API which is already documented in the reference manual.

There is a fair bit of work to do here but I don't think we can postpone it any longer!

### DIFF
--- a/doc/Developer_Manual/index.rst
+++ b/doc/Developer_Manual/index.rst
@@ -1,0 +1,6 @@
+****************
+Developer Manual
+****************
+
+The is the developer manual where the design and API of HoloViews is
+described to help new developers contributing to the project.
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -116,6 +116,7 @@ showing how to fix the bug or implement the feature!
    Features <features>
    Tutorials <Tutorials/index>
    Examples <Examples/index>
+   Developer Manual <Developer_Manual/index>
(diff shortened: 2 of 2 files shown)


jlstevens:
I'm making some progress and will have new commits soon. I just wanted to mention the problem I had described in issue #525. Should I make use of such links and assume this problem can be fixed shortly or should I avoid them for now?

jbednar:
I'd go ahead and use the links, because even if they don't work, they are at least a very precise record of what entity is being discussed.  And they may work someday!

jlstevens:
Thankfully, Philipp managed to implement a fix pretty quickly! #525

philippjfr:
Anything ever happen here? I started a some documentation about the renderer and plotting API, which I hope to get merged before 1.7, but it seems like this PR isn't going anywhere. Can I close?

jlstevens:
Yes, I think you can close.

Generally, the idea now is to push this kind of material to contrib. If you have a notebook explaining these topics, it would be good to submit to contrib.

PR #524, opened 2016-02-27
Fixed bug in bokeh hover tool for unsanitized dimensions

philippjfr:
The bokeh hover tool currently breaks if a dimension contains a space or other special character, the PR adds a small method to generate the hover data correctly.


jlstevens:
Also looks like the new method removes redundancy which is always good. Happy to merge if it is ready.

philippjfr:
Ready to merge whenever.

jlstevens:
Now then! :-)

Issue #528, opened 2016-03-02
Ensure that holoviews imports without any plotting backend


philippjfr:
As the title says, HoloViews used to require only a minimal set of packages to import. This is now broken because the ``__init__.py`` attempts to import ipython, which in turn requires at least the matplotlib backend to be loaded. We should ensure this either fails gracefully or get rid of this requirement entirely. Additionally we should also ensure that each backend works independently from all the others, allowing bokeh to be used without matplotlib being installed.

jlstevens:
> .. which in turn requires at least the matplotlib backend to be loaded

It is worth pointing out that *I believe* what is happening is that it is expecting a renderer to exist when none may be defined (no plotting backends available).

I've not looked at the code just yet but I think you meant 'at least one backend to be loaded'...unless we've hardcoded matplotlib as the default backend. In that case, the default backend should work by ordering where the first backend in the renderers list is the default.

philippjfr:
That's one issue but there's more, after resolving that issue I ran into at least two more issues:

1) Notebook archive requires nbformat (usually shipped with the notebook)
2) Parser requires pyparsing (usually shipped with matplotlib)

jlstevens:
Really our ipython extension really means the ipython notebook extension which is why we can assume jinja2 and nbformat. The pyparsing issue we've known about and I do think it is safe to say that most users will have matplotlib (and can easily install pyparsing installed otherwise).

In conclusion, I think those other imports are ok (nbformat and pyparsing) and the main problem is you shouldn't have to have matplotlib as long as you are happy to install pyparsing to get the magics working.

philippjfr:
No, the IPython extension gets loaded as long as you have IPython installed, either we only import it when all the notebook dependencies are available or we make sure that it imports even without those dependencies, which is what I did in #529.

jlstevens:
The approach in #529 is good but what what we could have done is simply try to import notebook and not IPython as that would automatically ensure the dependencies are satisfied (except for pyparsing).

philippjfr:
HoloViews now avoids importing the plotting backends until they are requested by the notebook extension. Closing.

PR #529, opened 2016-03-02
Removed dependency on matplotlib, nbformat and pyparsing in ipython

philippjfr:
Addresses issues mentioned in #528 partially by removing all hard dependencies on notebook and matplotlib in the ipython module. This ensures holoviews can be imported in IPython even without matplotlib and the notebook.

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -2,7 +2,7 @@ import os
 from unittest import SkipTest
 
 import param
-import jinja2
+from IPython import version_info
 from IPython.display import HTML
 
 import holoviews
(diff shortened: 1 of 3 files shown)


holoviews/ipython/__init__.py
```diff
@@ -11,17 +11,24 @@
...
+try:
+    if version_info[0] >= 4:
+        import nbformat
+    else:
+        from IPython import nbformat
+    from .archive import notebook_archive
+    holoviews.archive = notebook_archive
+except ImportError:
+    pass
+
```
jlstevens: I suspect we'll want to add some ``# noqa`` lines here for pyflakes...

philippjfr: True, the nbformat lines will need it.

jlstevens:
Looks good and does what I expect. We only need jinja2 if widgets are requested and a conditional import of archive makes sense in case pyparsing is missing.

philippjfr:
>We only need jinja2 if widgets are requested

Also used to display the ``notebook_extension`` output, but again that's only needed when working in the notebook anyway.

>conditional import of archive makes sense in case pyparsing is missing.

Archive only requires nbformat. It's the parser that requires pyparsing which was already mostly working except that although the magics were wrapping the import in a try except the __init__.py was not.

jbednar:
Does this address the issues that Ilan found when he made conda packages for the main channel, for python3?  Hopefully those extra dependencies that he found would no longer be needed?

philippjfr:
Yes it does, although we should still add notebook as a dependency to our conda package.

jlstevens:
Yes, I was confusing the use of pyparsing for the magics with nbformat for archive. The conditional import still makes sense though...

philippjfr:
Ready to merge once the tests pass.

jlstevens:
Ok the tests are passing so I'll merge it now.

PR #532, opened 2016-03-04
Various plotting fixes

philippjfr:
A number of fixes to plotting in particular for the bokeh backend. 

The fixes include:

* Empty datasources cause issues in 0.11.1, now setting the glyph invisible instead
* Datetime axes limits were being inferred incorrectly on OverlayPlots
* Fixed some outdated method signatures and unreferenced variables
* Fixed a bug with selection_specs in Columns.select that caused issues in plotting
* Fixed a bug in widget templating resulting in a unicode string with the `u` prefix being templated (causing JS syntax errors)

### DIFF
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -173,7 +173,7 @@ class Columns(Element):
         supplied, which will ensure the selection is only applied if the
         specs match the selected object.
         """
-        if selection_specs and not self.matches(selection_specs):
+        if selection_specs and not any(self.matches(sp) for sp in selection_specs):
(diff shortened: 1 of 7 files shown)


jlstevens:
I've had a look through and all the fixes look good. Thankfully this is more a collection of small, localized fixes than a refactor! If it is ready, I'm happy to merge once the tests pass.

philippjfr:
As you say just a bunch of small changes and tests are passing. Ready to merge.

jlstevens:
Merged.

PR #534, opened 2016-03-05
Fixing urls

basnijholt:
Links missed `.html`


philippjfr:
Sorry should have gotten to these earlier. They should end in ``.ipynb`` so you can switch between them when running the notebook. We replace it with ``.html`` automatically when building the docs.

philippjfr:
Superceded by #757. Closing.

PR #537, opened 2016-03-06
Switched Bokeh LineAnnotationPlot to use Span annotation

philippjfr:
Previously we had to use the bokeh BoxAnnotation to render VLine and HLine Elements but since 0.11 bokeh ships with a Span annotation which is more appropriate. This can be merged once https://github.com/bokeh/bokeh/issues/3993 in bokeh is fixed and 0.12 is released.


philippjfr:
Ready to merge now, we don't support bokeh <0.12 anymore anyway.

PR #538, opened 2016-03-06
Renderer instances keep track of last plot

philippjfr:
This PR adds a ``last_plot`` attribute to ``Renderer`` instances, which let's the user easily get a handle on the last thing they plotted, which is very helpful for debugging, understanding how plots are structured, and in order to directly customize a plot, e.g. to prototype an ``initial_hook`` or ``final_hook``.

### DIFF
--- a/holoviews/plotting/renderer.py
+++ b/holoviews/plotting/renderer.py
@@ -143,6 +143,7 @@ class Renderer(Exporter):
     backend_dependencies = {}
 
     def __init__(self, **params):
+        self.last_plot = None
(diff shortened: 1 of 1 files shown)


jlstevens:
As we've discussed, ``Renderer`` is the right place to put this and it is a small change. I would be happy to merge but I'm wondering what is up with Travis right now...

philippjfr:
Restarted the tests, ready to merge when it goes green.

jlstevens:
Tests are passing. Merging.

Issue #541, opened 2016-03-07
Columns row/column based indexing API


philippjfr:
As we further broaden the scope of the Columns interfaces and provide support for various out-of-memory data backends it's becoming clear that the current interface isn't quite sufficient. When accessing a Columns datastructure it is sometimes necessary to do so by index, e.g. when displaying a Table and only a small subset of rows should be shown or when trying to randomly sampling from the data. Currently this requires each column to be loaded into memory in its entirety to index into them. Instead there should be an API on the Columns class to slice the row and column directly. I could imagine something like this `table.cells[0:10, 0:2]`, where cells is a small object that allows slicing by index.

jbednar:
Would this API provide Raster-like access for Image and RGB objects?

philippjfr:
That would also benefit from an interface to slice into the array by index but I'm not sure if the two interfaces should be shared.

jbednar:
Seems like they ought to have the same spelling, at least...

philippjfr:
Tabular and nd-array indexing interfaces have been added in https://github.com/ioam/holoviews/pull/1435.

PR #542, opened 2016-03-07
Adding a general interface for N-dimensional gridded data

philippjfr:
In HoloViews we now have an interface to hold data in a columnar format. This is provides a very powerful interface for some kinds of data, however when exploring dense high-dimensional arrays it is wasteful because it expands the coordinates of the key dimensions. An alternative format used by xarray, iris and in some limited ways pandas, stores the index values or coordinates (as they are sometimes called) separately from the value dimension data, which is stored as an n-dimensional array.

Instead of storing the cartesian product of all key dimension values we store only the outer indices. Often working in a Columnar format is a lot easier because merging or adding new or derived dimensions is considerably easier, however it is not only inefficient in terms of space but is also considerably slower for various operations, particularly for groupby, aggregation and reduce operations.

### The proposal

The HoloViews Columns interface actually provides a very general interface to work with structured data and in theory it does not actually restrict the format of the data. In this notebook I will outline a suggestion to add additional interfaces for the Columns type, which works with N-D gridded from hereon referred to as dense data. I will set out to show that not only is this format more efficient for various operations but the implementation is actually fairly simple and fits into our current system.

##### The datastructure

The current Columns interfaces already have different datastructures, which all fundamentally represent an array of the shape``Row x Column``, where the rows represent the total number of samples and the columns the combined key and value dimensions, this is fundamentally no different to the COO (Coordinate) sparse matrix format (except the r, c indices are actually values). The new format would provide a dense equivalent and would differ from these existing implementations in the following ways:

1) The data would clearly distinguish between key dimensions, which provide the indices/coordinates to index into the value dimensions, and how the actual value dimensions are stored, some example data would look like this:

```python
hv.Table({'x': range(10), 'y': range(100), 'z': np.random.rand(10, 100)},
         kdims=['x', 'y'], vdims=['z'])
```

Here the x and y arrays provide the indexes along the first and second axis of the z-array. Using the current formats this would have to be specified as:

```python
xs, ys = np.meshgrid(range(10), range(100))
hv.Table({'x': xs, 'y': ys, 'z': np.random.rand(10, 100)}, kdims=['x', 'y'], vdims=['z'])
```

Instead of storing the cartesian product as computed by meshgrid, the internal representation stores just the outer indices. The interface then expands these indices if required (which would generally be pretty rare).

2) There are two possible ways to represent value dimensions, either we can have one array for all value dimensions which simply stacks the arrays or we can expand the value dimensions out into the separate arrays, i.e. multiple value dimensions could be specified like this:

```python
hv.Table({'x': range(10), 'y': range(100), 'array': np.random.rand(10, 100, 2)}, kdims=['x', 'y'], vdims=['a', 'b'])
```

or like this:

```python
hv.Table({'x': range(10), 'y': range(100), 'a': np.random.rand(10, 100), 'b': np.random.rand(10, 100)},
         kdims=['x', 'y'], vdims=['a', 'b'])
```

This comes down to whether the interface should support heterogeneous value dimension types. The current proposal works on the first suggestion but it would be trivial to automatically expand the first format into the second format and store the value dimensions separately internally.

## Pros vs Cons

Pros:

+ More memory efficient
+ Considerably faster for gridded data
+ Columns and Raster types would share the same interface, the only other data structures required would be for Annotations and Paths.
+ Avoids having two completely separate implementations for gridded data and columnar data, which is important because even our current Chart types could sometimes benefit from a denser representation.
+ Provides a template for further interfaces based on xarray and iris Cubes.

Cons:

- Requires some changes to existing interfaces to access both the expanded and compressed formats easily.
- Columns becomes a misnomer (not a major obstacle since renaming baseclasses is easy).
- Gridded data is more restricted than columnar data, so value dimension indexing and swapping key and value dimensions is not directly supported.

## Obstacles/Problems

* Need to establish clear interfaces to access the dimension values both in the expanded cartesian product format and in the dense format. I would propose that ``dimension_values`` accepts a ``product`` argument (or similar) that defaults to ``False`` returning the full cartesian product by default. Additionally it would also support a ``flat`` argument defaulting to ``True`` to retain a consistent backward compatible interface. 
* Consider how to deal with Raster and Image types where the key dimension values are implicit. My current suggestion would be either to generate the coordinates in the constructor or to introduce proxy objects, which lazily compute the indexes, e.g. to describe an Image you could do something like:

```python
hv.Image({'x': BoundCoords(-0.5, 0.5, 100), 'y': BoundCoords(-0.5, 0.5, 100), 'z': np.random.rand(100, 100)})
```

* Gridded data would be more restricted than Columnar data, e.g. before slicing value dimensions or reindexing the Element it would first have to be converted to the expanded format.
* Since the sparse and dense representation have potentially overlapping signatures I believe it should be a parameter on the Columns object, you should explicitly declare that the data you are passing in is dense and the dense backends will try to parse that data.
* We can have ``as_dense`` and ``as_sparse`` methods that convert between dense and sparse representations. The as_sparse representation is obviously very straightforward as it's just the cartesian product of the dense representation. The ``as_dense`` implementation requires that the data has been aggregated already. After that it's most easily implemented by combining the sparse columns with a cartesian product of the key dimensions inserting NaNs for all values, aggregating, sorting and reshaping.
* Certain methods when applied to a dense Element return a sparse representation (currently only sample would do so).

To-do list:

- [x] Establish API for accessing data in column based and array based formats using ``dimension_values`` (interface proposed in #541 postponed).
- [x] Complete the interface by deciding whether to support sorting or whether dense sorting should be fixed. (Require sorted key data in constructor for now)
- [x] Adapt HeatMap to support both formats (postponed)
- [x] Decide what to do about Raster and Image key dimensions values (postponed).
- [x] Decide whether to implement Histogram and QuadMesh using the interface (postponed).
- [x] Add unit tests
- [x] Update docstrings

[Notebook with examples and profiling](https://github.com/ioam/holoviews/issues/541)

### DIFF
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -44,7 +44,7 @@ class Columns(Element):
     of aggregating or collapsing the data with a supplied function.
     """
 
-    datatype = param.List(['array', 'dataframe', 'dictionary', 'ndelement'],
+    datatype = param.List(['array', 'dataframe', 'dictionary', 'grid', 'ndelement'],
         doc=""" A priority list of the data types to be used for storage
         on the .data attribute. If the input supplied to the element
         constructor cannot be put into the requested format, the next
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -858,7 +858,7 @@ class DFColumns(DataColumns):
 
 
     @classmethod
-    def values(cls, columns, dim, expanded, flat):
+    def values(cls, columns, dim, expanded=True, flat=True):
         data = columns.data[dim]
         if util.dd and isinstance(data, util.dd.Series):
             data = data.compute()
@@ -986,7 +986,7 @@ class ArrayColumns(DataColumns):
 
 
     @classmethod
-    def values(cls, columns, dim, expanded, flat):
+    def values(cls, columns, dim, expanded=True, flat=True):
         data = columns.data
         dim_idx = columns.get_dimension_index(dim)
         if data.ndim == 1:
@@ -1238,7 +1238,7 @@ class DictColumns(DataColumns):
         return OrderedDict([(d, v[sorting]) for d, v in columns.data.items()])
 
     @classmethod
-    def values(cls, columns, dim, expanded, flat):
+    def values(cls, columns, dim, expanded=True, flat=True):
         values = np.array(columns.data.get(columns.get_dimension(dim).name))
         if not expanded:
             return util.unique_array(values)
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -442,7 +442,7 @@ class DataColumns(param.Parameterized):
         # Iterate over interfaces until one can interpret the input
         for interface in prioritized:
             try:
-                (data, kdims, vdims) = interface.reshape(eltype, data, kdims, vdims)
+                (data, kdims, vdims) = interface.init(eltype, data, kdims, vdims)
                 break
             except:
                 pass
@@ -583,7 +583,7 @@ class NdColumns(DataColumns):
     datatype = 'ndelement'
 
     @classmethod
-    def reshape(cls, eltype, data, kdims, vdims):
+    def init(cls, eltype, data, kdims, vdims):
         if isinstance(data, NdElement):
             kdims = [d for d in kdims if d != 'Index']
         else:
@@ -712,7 +712,7 @@ class DFColumns(DataColumns):
         return columns.data.dtypes[idx].type
 
     @classmethod
-    def reshape(cls, eltype, data, kdims, vdims):
+    def init(cls, eltype, data, kdims, vdims):
         element_params = eltype.params()
         kdim_param = element_params['kdims']
         vdim_param = element_params['vdims']
@@ -908,7 +908,7 @@ class ArrayColumns(DataColumns):
         return columns.data.dtype.type
 
     @classmethod
-    def reshape(cls, eltype, data, kdims, vdims):
+    def init(cls, eltype, data, kdims, vdims):
         if kdims is None:
             kdims = eltype.kdims
         if vdims is None:
@@ -1127,7 +1127,7 @@ class DictColumns(DataColumns):
         return columns.data[name].dtype.type
 
     @classmethod
-    def reshape(cls, eltype, data, kdims, vdims):
+    def init(cls, eltype, data, kdims, vdims):
         odict_types = (OrderedDict, cyODict)
         if kdims is None:
             kdims = eltype.kdims
@@ -1346,7 +1346,7 @@ class GridColumns(DictColumns):
     datatype = 'grid'
 
     @classmethod
-    def reshape(cls, eltype, data, kdims, vdims):
+    def init(cls, eltype, data, kdims, vdims):
         if kdims is None:
             kdims = eltype.kdims
         if vdims is None:
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -1330,7 +1330,7 @@ class DictColumns(DataColumns):
 
 
 
-class NdArrayColumns(DictColumns):
+class GridColumns(DictColumns):
     """
     Interface for simple dictionary-based columns format. The dictionary
     keys correspond to the column (i.e dimension) names and the values
@@ -1339,9 +1339,7 @@ class NdArrayColumns(DictColumns):
 
     types = (dict, OrderedDict, cyODict)
 
-    datatype = 'ndarray'
-
-    dense = True
+    datatype = 'grid'
 
     @classmethod
     def reshape(cls, eltype, data, kdims, vdims):
@@ -1573,6 +1571,6 @@ class NdArrayColumns(DictColumns):
 DataColumns.register(DictColumns)
 DataColumns.register(ArrayColumns)
 DataColumns.register(NdColumns)
-DataColumns.register(NdArrayColumns)
+DataColumns.register(GridColumns)
 if pd:
     DataColumns.register(DFColumns)
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -1341,9 +1341,18 @@ class DictColumns(DataColumns):
 
 class GridColumns(DictColumns):
     """
-    Interface for simple dictionary-based columns format. The dictionary
-    keys correspond to the column (i.e dimension) names and the values
-    are collections representing the values in that column.
+    Interface for simple dictionary-based columns format using a
+    compressed representation that uses the cartesian product between
+    key dimensions. As with DictColumns, the dictionary keys correspond
+    to the column (i.e dimension) names and the values are NumPy arrays
+    representing the values in that column.
+
+    To use this compressed format, the key dimensions must be orthogonal
+    to one another with each key dimension specifiying an axis of the
+    multidimensional space occupied by the value dimension data. For
+    instance, given an temperature recordings sampled regularly across
+    the earth surface, a list of N unique latitudes and M unique
+    longitudes can specify the position of NxM temperature samples.
     """
 
     types = (dict, OrderedDict, cyODict)
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -51,9 +51,6 @@ class Columns(Element):
         format listed will be used until a suitable format is found (or
         the data fails to be understood).""")
 
-    dense = param.Boolean(default=False, doc="""Whether the supplied data
-        is in a dense format.""")
-
     # In the 1D case the interfaces should not automatically add x-values
     # to supplied data
     _1d = False
@@ -254,7 +251,7 @@ class Columns(Element):
         matching the key dimensions, returning a new object containing
         just the selected samples.
         """
-        return self.clone(self.interface.sample(self, samples), dense=False)
+        return self.clone(self.interface.sample(self, samples))
 
 
     def reduce(self, dimensions=[], function=None, spreadfn=None, **reduce_map):
@@ -464,6 +461,11 @@ class DataColumns(param.Parameterized):
             raise ValueError("Supplied data does not contain specified "
                              "dimensions, the following dimensions were "
                              "not found: %s" % repr(not_found))
+
+
+    @classmethod
+    def check_dense(cls, arrays):
+        return any(array.shape not in [arrays[0].shape, (1,)] for array in arrays[1:])
 
 
     @classmethod
@@ -748,6 +750,9 @@ class DFColumns(DataColumns):
                     data = tuple(data[:, i]  for i in range(data.shape[1]))
 
             if isinstance(data, tuple):
+                data = [np.array(d) if not isinstance(d, np.ndarray) else d for d in data]
+                if cls.check_dense(data):
+                    raise ValueError('Dimension values must all be the same shape.')
                 data = pd.DataFrame.from_items([(c, d) for c, d in
                                                 zip(columns, data)])
             else:
@@ -1156,6 +1161,8 @@ class DictColumns(DataColumns):
             raise ValueError("DictColumns interface couldn't convert data.""")
         elif isinstance(data, dict):
             unpacked = [(d, np.array(data[d])) for d in data]
+            if cls.check_dense([d[1] for d in unpacked]):
+                raise ValueError('Dimension values must all be the same shape.')
             if isinstance(data, odict_types):
                 data.update(unpacked)
             else:
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -1591,11 +1591,6 @@ class GridColumns(DictColumns):
 
     @classmethod
     def add_dimension(cls, columns, dimension, dim_pos, values, vdim):
-        raise NotImplementedError
-
-
-    @classmethod
-    def add_dimension(cls, columns, dimension, dim_pos, values, vdim):
         if not vdim:
             raise Exception("Cannot add key dimension to a dense representation.")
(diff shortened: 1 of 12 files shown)


jlstevens:
I think this will be great!

Just to summarize the discussion I've just had with Philipp for future reference:

* I think instead of ``product`` the new keyword argument in ``dimension_values`` could be called ``expanded`` or something similar.

* I'm not so sure about the ``'array'`` key for specifying vdims, mainly as numpy arrays (at least) would not allow heterogeneous types. We can come back to this bit later as it is optional but I can say I would prefer a key called ``'vdims'``.

*  I am happy with a new interface based on ``DictColumns`` which I would call ``GridColumns``. I would want to make sure we always try the simple ``DictColumns`` interface first before ``GridColumns``: this makes sure there is effectively one dictionary format to reason about even though there are actually two interface classes behind the scenes.

* To keep the semantics easier to understand, the rule I would recommend is that ``DictColumns`` is used if all the array values have the same shape. Otherwise, you fall back to ``GridColumns`` as suggested in the previous point. This means if you want a Cartesian product for kdims only where all the arrays are the same shape, you would explicitly have to remove ``GridColumns`` from the interface list as a user.

Anyway, these are my general comments for now. I'm happy to review the code in more detail once you think it is ready.

jbednar:
I'm very happy with this proposal.  I think it really will help HoloViews work well in a broad range of other applications, and is worth taking the effort to work on now.

philippjfr:
Okay as far as I can tell I'm now done with this PR. @jlstevens said he'd go through and document the class and methods so he can get a better idea about the implementation. So I'll wait on that to make any more changes, as I'm sure he'll find some further issues.

philippjfr:
Note that a lot of the work to allow Raster, Image, Histogram and QuadMesh types to use dense interfaces has been postponed and is not part of this PR. Hopefully for version 1.5 we can unify all these types together leaving only Path and Annotation types with custom data formats.

jlstevens:
I'm going to go through this PR carefully now, making sure I understand it, making comments and updating docstrings as necessary. Then once those issues are addressed I think it can be merged.

holoviews/core/data.py
```diff
@@ -469,6 +464,11 @@ def validate(cls, columns):
 
 
     @classmethod
+    def check_dense(cls, arrays):
+        return any(array.shape not in [arrays[0].shape, (1,)] for array in arrays[1:])
```
jlstevens: If I understand this code correctly ``check_compressed`` might be a better name...

holoviews/element/raster.py
```diff
@@ -615,13 +615,14 @@ def dimension_values(self, dim, unique=False):
...
-            else:
+            if expanded:
                 Y, X = np.meshgrid(d2lin, d1lin)
                 return Y.flatten() if dim_idx else X.flatten()
+            else:
+                return d2lin if dim_idx else d1lin
         elif dim_idx == 2:
-            return np.flipud(self.data).T.flatten()
+            data = np.flipud(self.data).T
+            return data.flatten() if flat else data
```
jlstevens: The values are the flattened, transposed ``flipud`` of the data array? This may well be correct but we will need to know why this is necessary and document it.

holoviews/core/data.py
```diff
@@ -714,7 +717,6 @@ def reshape(cls, eltype, data, kdims, vdims):
         kdim_param = element_params['kdims']
```
jlstevens: We should probably renamed ``reshape`` to ``init`` as it is confusing. I got confused with reshape in numpy!

holoviews/core/data.py
```diff
@@ -1308,9 +1334,288 @@ def aggregate(cls, columns, kdims, function, **kwargs):
 
 
 
+class GridColumns(DictColumns):
+    """
+    Interface for simple dictionary-based columns format. The dictionary
+    keys correspond to the column (i.e dimension) names and the values
+    are collections representing the values in that column.
+    """
```
jlstevens: I'll update the docstring to say why this isn't the same as ``DictColumns``.

holoviews/core/data.py
```diff
@@ -1308,9 +1334,288 @@ def aggregate(cls, columns, kdims, function, **kwargs):
...
+        if issubclass(container_type, NdMapping):
+            with item_check(False), sorted_context(False):
+                return container_type(grouped_data, kdims=dimensions)
+        else:
+            return container_type(grouped_data)
+
+
+    @classmethod
+    def coord_mask(cls, columns, arr, ind):
+        if isinstance(ind, tuple):
```
jlstevens: The docstring should say how this related to ``select_mask`` as it is very similar. I think it might be called ``select_mask_kdim`` where ``ind`` is the ``kdim`` index. I'm not sure whether we need both ``columns`` and ``arr`` in the signature either...

holoviews/core/data.py
```diff
@@ -1308,9 +1334,288 @@ def aggregate(cls, columns, kdims, function, **kwargs):
...
+
+
+    @classmethod
+    def length(cls, columns):
+        return np.product([len(columns.data[d.name]) for d in columns.kdims])
+
+
+    @classmethod
+    def values(cls, columns, dim, expanded=True, flat=True):
+        if dim in columns.kdims:
```
jlstevens: The docstring will need to explain that the cartesian product is only applied to key dimensions.

holoviews/core/data.py
```diff
@@ -1308,9 +1334,288 @@ def aggregate(cls, columns, kdims, function, **kwargs):
...
+            if not expanded:
+                return columns.data[dim]
+            prod = util.cartesian_product([columns.data[d.name] for d in columns.kdims])
+            idx = columns.get_dimension_index(dim)
+            return prod[:, idx]
```
jlstevens: The ``flat`` argument doesn't seem to do anything here. Either the call to ``flatten`` can be done outside this method or maybe the key dimension values should take the shape of the value dimension arrays if ``flat`` is set to ``False``?

holoviews/core/data.py
```diff
@@ -1308,9 +1334,288 @@ def aggregate(cls, columns, kdims, function, **kwargs):
...
+        if kdims != columns.kdims:
+            axes = [columns.kdims.index(d) for d in kdims]
+            for vdim in vdims:
+                data[vdim.name] = np.transpose(data[vdim.name], axes)
+        return data
+
+
+    @classmethod
+    def add_dimension(cls, columns, dimension, dim_pos, values, vdim):
+        raise NotImplementedError
```
jlstevens: It seems to be implemented just below. :-)

holoviews/core/data.py
```diff
@@ -1308,9 +1334,288 @@ def aggregate(cls, columns, kdims, function, **kwargs):
...
+        return dict(columns.data, **{dim: values})
+
+
+    @classmethod
+    def sort(cls, columns, by=[]):
+        if not by or by in [columns.kdims, columns.dimensions()]:
+            return columns.data
+        else:
+            raise Exception('Dense representation cannot be sorted, either instantiate '
+                            'in the desired order or use a sparse format.')
```
jlstevens: It isn't clear what you mean by sparse format. I think 'expanded format' might be clearer as they are still dense arrays but you don't need to worry about the cartesian product...

holoviews/core/data.py
```diff
@@ -1308,9 +1334,288 @@ def aggregate(cls, columns, kdims, function, **kwargs):
...
+            data[vdim.name] = np.atleast_1d(function(columns.data[vdim.name],
+                                                     axis=axes, **kwargs))
+
+        return data
+
+
+    @classmethod
+    def reindex(cls, columns, kdims, vdims):
+        if len(kdims) != columns.ndims:
+            raise ValueError('Cannot drop dimension varying dimension.')
```
jlstevens: Not sure what this exception message means...

holoviews/core/data.py
```diff
@@ -1308,9 +1334,288 @@ def aggregate(cls, columns, kdims, function, **kwargs):
...
+        return mask
+
+
+    @classmethod
+    def select(cls, columns, selection_mask=None, **selection):
+        dimensions = columns.dimensions('key', label=True)
+        val_dims = [vdim for vdim in columns.vdims if vdim in selection]
+        if val_dims:
+            raise IndexError('Cannot slice value dimensions on dense '
+                             'data, convert to sparse format first')
```
jlstevens: Similar comment to something I said below...I think 'sparse format' is confusing. As far as I know, you mean 'expanded' format (i.e all kdims have the same shape, no cartesian product applied).

holoviews/core/data.py
```diff
@@ -1308,9 +1334,288 @@ def aggregate(cls, columns, kdims, function, **kwargs):
...
+
+        dimensions = [d.name if isinstance(d, Dimension) else
+                      d for d in kdims + vdims]
+        if isinstance(data, tuple):
+            data = {d: v for d, v in zip(dimensions, data)}
+        elif not isinstance(data, dict):
+            raise ValueError('GridColumns must be instantiated as a '
+                             'dictionary or tuple')
+
+        if 'vdims' in data:
```
jlstevens: I think we might want to change this. Although the functionality is useful, I don't like the arbitrary name 'vdims'. Instead I propose that the dictionary key is allowed to be a tuple of the actual value dimension names e.g ``('height', 'pressure', 'temperature').

holoviews/core/data.py
```diff
@@ -1308,9 +1334,288 @@ def aggregate(cls, columns, kdims, function, **kwargs):
...
+            name = dim.name if isinstance(dim, Dimension) else dim
+            if name not in data:
+                raise ValueError("Values for dimension %s not found" % dim)
+            if not isinstance(data[name], np.ndarray):
+                data[name] = np.array(data[name])
+
+        kdim_names = [d.name if isinstance(d, Dimension) else d for d in kdims]
+        vdim_names = [d.name if isinstance(d, Dimension) else d for d in vdims]
+        expected = [len(data[kd]) for kd in kdim_names]
+        for vdim in vdim_names:
```
jlstevens: I am wondering if this bit shouldn't be in ``validate``. Looks to me like ``validate`` is always called when the interface is constructed. Does an exception in ``validate`` trigger a fallback?

jlstevens:
I think I'm done making comments for now and I only have a few docstrings to update as the API is consistent with what we had before. There are a few name changes and most of what I suggest should be quite quick to fix. Once these fixes are done, I'll run flakes and update the docstrings I mentioned.

holoviews/core/data.py
philippjfr: No that's part of the problem, ``validate`` only gets called once the object has been fully instantiated, which is too late for a fallback.

holoviews/core/data.py
philippjfr: I'm happy to take this out for now.

holoviews/core/data.py
philippjfr: Oops, will fix it.

holoviews/core/data.py
philippjfr: Oops, must have duplicated itself in a copy/paste.

holoviews/core/data.py
philippjfr: I think for consistency the key dimensions should also be in the vdim shape. Don't have any current use for it but it's consistent and you can could evaluate a new function on the same grid easily.

holoviews/element/raster.py
philippjfr: Will add some comments.

holoviews/core/data.py
philippjfr: Done.

holoviews/element/raster.py
philippjfr: Done.

holoviews/core/data.py
philippjfr: Done.

holoviews/core/data.py
philippjfr: I've renamed it to key_select_mask because that's what made most sense to me. I'm still passing through the values array because it avoids getting it multiple times and I don't want to remove columns as an argument because it would be the only inconsistent method in that regard.

holoviews/core/data.py
philippjfr: Implemented this for both the gridded interface and our Raster types now.

holoviews/core/data.py
philippjfr: Deleted.

holoviews/core/data.py
philippjfr: Done.

holoviews/core/data.py
philippjfr: Sounds good!

holoviews/core/data.py
philippjfr: Will leave that for you.

philippjfr:
Okay, I've gone through an made all the fixes you suggested and tests should pass in a minute. If you could go through it and add docstrings then I think this is ready to merge. Only other thing we should decide is whether to add 'grid' as a potential interface.

holoviews/core/data.py
jlstevens: I've filed issue #550 as a reminded to reconsider this functionality.

jlstevens:
I've had a go updating the class docstring for ``GridColumns`` and I've made a few other small changes. In the end, I've decided not to do all the docstrings now as all the methods on ``DataColumns`` should be given docstrings and there are a lot of them! ``DataColumns`` is what defines the API so that should go in a separate PR.

Once the pr tests pass, I'm happy to merge.

jlstevens:
Ok, the pr build is passing. Time to merge!

PR #544, opened 2016-03-08
Fixing Travis issue with libgfortran

jlstevens:
After a couple unsuccessful attempts to fix this mysterious issue on master, this PR will allow me to try as many different things as necessary till the tests are working again.


philippjfr:
Very odd issue. Looks like it it's passing now though, will merge.

PR #555, opened 2016-03-14
Fixed setting of aspect on log axes

philippjfr:
Previously setting of aspects on plots with log axes was not working so I just disabled it. However I just found out through an issue brought up by @basnijholt that this is merely because aspects have to be set before the axes have had log scaling applied. This PR simply changes the order in which these settings are applied.


philippjfr:
Closing as rejected. Requires a more subtle approach.

PR #556, opened 2016-03-14
Fixed generator generator check on DataColumns

philippjfr:
The check to expand ``zip``, ``range`` and generator expressions on Columns relied on excluding other types rather than checking explicitly for those types. This explicitly checks for (i)zip, (x)range and generator types across python versions.


jlstevens:
It is a simple fix but an important one. The tests are passing so I'll merge it now.

PR #557, opened 2016-03-14
Simplified cartesian_product utility

philippjfr:
As the title says this greatly simplifies the ``cartesian_product`` utility now returning arrays for each dimension in the shape defined by the dimension values.


jlstevens:
I understand this is ready to merge now the tests are passing.

PR #560, opened 2016-03-19
Deprecated cyclic options on Curve plots

philippjfr:
Curve plots have various options to apply very specific operations to cyclic Curve data including centering it on the highest value, applying relative labels and formatting radian values as degrees. These options were inherited from old topographica plotting code and made sense in that context, however within the general framework of holoviews they don't make much sense and are better implement as operations on the data and by setting the formatter or ticks directly. The other issue is that these options were broken when applied to Curves in Overlays. This PR removes these options.

### DIFF
--- a/holoviews/plotting/mpl/chart.py
+++ b/holoviews/plotting/mpl/chart.py
@@ -23,60 +23,6 @@ class ChartPlot(ElementPlot):
     show_legend = param.Boolean(default=True, doc="""
         Whether to show legend for the plot.""")
 
-    def __init__(self, data, **params):
-        super(ChartPlot, self).__init__(data, **params)
-        key_dim = self.hmap.last.get_dimension(0)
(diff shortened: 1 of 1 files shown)


jlstevens:
Seems reasonable as I don't think anyone would have been using this...especially as it wasn't documented. I agree this is worth removing especially as it was broken!

I understand it is ready to merge and the tests are passing so I'll do that now.

jbednar:
Renamed, since it removes these options, it doesn't just deprecate them.

At least for tuning curve plots in Topographica, it's important to be able to center them, so is that code moving to Topographica?

philippjfr:
I've added a ``center_cyclic`` operation to featuremapper. I can move it  to or make it available somewhere in topographica as well though.

jbednar:
Featuremapper sounds like a good place for that.

PR #563, opened 2016-03-22
Let GridColumns.reindex expand format as required

philippjfr:
Currently the newly added ``GridColumns`` interface restricts how the data can be reindexed, ensuring that you're not dropping non-redundant dimensions. This is because gridded data requires unique indices for each position in the grid. This is not an issue when working with the expanded format because individual samples do not have to be unique. For full generality reindex on GridColumns and other grid based interfaces should allow any requested reindexing even if it requires falling back to the expanded format (which is now very cheap because the expansion uses striding tricks).

### DIFF
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -1589,11 +1589,10 @@ class GridColumns(DictColumns):
     @classmethod
     def reindex(cls, columns, kdims, vdims):
         dropped_kdims = [kd for kd in columns.kdims if kd not in kdims]
-        if dropped_kdims and any(len(columns.data[kd.name]) > 1 for kd in dropped_kdims):
-            raise ValueError('Compressed format does not allow dropping key dimensions '
-                             'which are not constant.')
-        if (any(kd for kd in kdims if kd not in columns.kdims) or
-            any(vd for vd in vdims if vd not in columns.vdims)):
+        if dropped_kdims and all(len(columns.data[kd.name]) == 1 for kd in dropped_kdims):
+            pass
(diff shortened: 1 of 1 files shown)


philippjfr:
Ready to merge.

jlstevens:
> For full generality reindex on GridColumns and other grid based interfaces should allow any requested reindexing even if it requires falling back to the expanded format 

Would it be appropriate to at least warn? I can imagine cases where the memory consumption explodes if you expand a high dimensional space.

philippjfr:
I'd have to profile it but since the cartesian product expansion works with striding tricks I don't think there will any significant increase in memory consumption.

jlstevens:
Ah. That's good but I think that means there is a difference between actually expanded (i.e consuming memory) and something that appears expanded (using striding tricks).

philippjfr:
True, we still need to make it clear that you shouldn't modify underlying datastructure inplace. It's always a bad idea since it can be shared across many objects. As long we're clear about that I don't think there's any functional difference between the actual expansion and the strided version.

philippjfr:
Just tried it, turns out I'm wrong, for the expanded format it has to flatten it, which means it does expand it in memory. So yes, I'm happy to add a warning. I think I can also rewrite the conversion interface to avoid reindexing before a groupby in most cases, which is probably the most common case where this occurs. 

The only time a full expansion should be required is when grouping by a value dimension, and you really shouldn't groupby unbinned continuous variables.

philippjfr:
Closing as rejected.

PR #568, opened 2016-03-23
Add binder

basnijholt:


philippjfr:
Just tested it, seems to work well. Merging now.

PR #570, opened 2016-03-23
Collector catches Task failures and warns

philippjfr:
When executing tasks using ``Collector`` a single exception will cause the entire run to fail. When executing a bunch of long-running tasks in sequence it's usually preferable to simply warn when an individual task fails. Alternatively we could add a parameter to control whether it is strict or not.


jlstevens:
Seems like a reasonable change to me. I'm happy to merge once you think it is ready.

philippjfr:
Ready.

jlstevens:
Ok merging now.

PR #572, opened 2016-03-24
Fixed DynamicMap error introduced by e3a0658e

tjol:
Closes ioam/holoviews#571


philippjfr:
Thanks, @tjol I should have realized my changes would affect dynamic mode and we desparately need tests to avoid regression since the widgets are probably one of most fragile components in HoloViews. Have you checked whether the DynamicMap tutorial works with this change? I think it might require some more adjustments.

jlstevens:
I also wish we could test widgets better! Unfortunately, everything I can think of is rather complicated e.g using selenium.

philippjfr:
Superseded by #575. Closing.

PR #574, opened 2016-03-24
Ad hoc fix of the \xi bug

### LINKED ISSUE #573: Dimension $\xi$ breaks HoloMap
Calling a dimension of a HoloMap or DynamicMap `$\xi$` fails with a javascript error
  `SyntaxError: malformed hexadecimal character escape sequence`

As far as I can tell, this is a question of insufficient escaping. Holoviews produces code like
``` javascript
    widget_ids[0] = "_anim_widget0f74e25011d543c28d2ea4d20346da74_$\xi$";`
```

which obviously isn't good JS.

tjol:
By escaping backslashes in the Jinja template this fixes #573, but it would be more elegant to have a specific filter for making things JavaScript-safe.

### DIFF
--- a/holoviews/plotting/widgets/__init__.py
+++ b/holoviews/plotting/widgets/__init__.py
@@ -272,10 +272,17 @@ class SelectionWidget(NdWidget):
                         dim_vals = escape_list(escape_vals(dim.values))
                         widget_type = 'dropdown'
                     init_dim_vals.append(dim_vals[0])
+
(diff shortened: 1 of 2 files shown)


philippjfr:
Looks like I overlooked this when escaping dimensions a while back, I've added escaping using our default dimension escaping function as part of #575. Thanks again for bringing this to my attention.

Issue #571, opened 2016-03-24
UnboundLocalError when using a DynamicMap


tjol:
Displaying a `DynamicMap` fails with the following error: 
  `UnboundLocalError("local variable 'value_labels' referenced before assignment",) `

The error is raised in `holoviews.plotting.widgets.SelectionWidget.get_widgets`. The fix is coming to a pull request near you in just a minute.

philippjfr:
This has been fixed in #575, please reopen if you encounter any issues.

PR #575, opened 2016-03-24
Fixed dimension escaping and bug setting dynamic widget labels

philippjfr:
When implementing improved widget value formatting in #562 I didn't handle the dynamic widget case causing some issues as described in #571 and partially fixed in #572. I've also included another fix for escaping dimension names that was overlooked previously and partially fixed in #574.


philippjfr:
Ready to merge.

jlstevens:
Good to fill in holes in the validation. Merging.

PR #577, opened 2016-03-28
Fix to nbagg widgets ensuring that tight fig_inches are applied

philippjfr:
While the nbagg backend works fairly well it does not seem to take into account the ``tight`` bbox_inches property, which means that a lot of whitespace is left over when plotting anything that is non-square. Since we already have a function to determine the tight bounding box we can simply apply that manually to avoid large white margins.

Ready to merge as soon as tests pass.


jlstevens:
Ok the tests are passing. Merging now.

PR #578, opened 2016-03-29
If plot ranges cannot be determined use mpl autoscale view

philippjfr:
This ensures that the xlim and ylim on a plot correspond to the actual data limits at all times, allowing calculations based on these values to be performed.

### DIFF
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -290,6 +290,7 @@ class ElementPlot(GenericElementPlot, MPLPlot):
         Compute extents for current view and apply as axis limits
         """
         # Extents
+        scalex, scaley = True, True
         extents = self.get_extents(view, ranges)
         if extents and not self.overlaid:
             coords = [coord if np.isreal(coord) else np.NaN for coord in extents]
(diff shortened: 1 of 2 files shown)


holoviews/plotting/mpl/element.py
```diff
@@ -314,16 +315,21 @@ def _set_axis_limits(self, axis, view, subplots, ranges):
...
             if self.invert_yaxis or any(p.invert_yaxis for p in subplots):
                 t, b = b, t
             if b != t:
                 if valid_lim(b):
                     axis.set_ylim(bottom=b)
+                    yset = True
                 if valid_lim(t):
                     axis.set_ylim(top=t)
+                    yset = True
+        axis.autoscale_view(scalex=not xset, scaley=not yset)
```
jlstevens: Seems like a reasonable change although why not invert the booleans and use the same names as matplotlib ``scalex`` and ``scaley``? That would seem clearer to me...

philippjfr: Agreed.

philippjfr:
Apparently some ``Spikes`` plots are affected by this, will have to investigate that before this can be merged.

philippjfr:
Ready to merge.

philippjfr:
Not sure why the reference data is complaining but this is still ready to merge.

jlstevens:
With the pr build passing, I'll merge it now.

PR #579, opened 2016-03-29
Disabled sorting optimization in Column interface groupby methods

philippjfr:
The sorting optimization in the Columns interface groupby methods cause issues because NdMapping types are always expected to be sorted. This disables the optimization so the resulting NdMapping types are always sorted appropriately.

### DIFF
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -24,7 +24,7 @@ import param
 from .dimension import Dimension, Dimensioned
 from .element import Element, NdElement
 from .dimension import OrderedDict as cyODict
-from .ndmapping import NdMapping, item_check, sorted_context
+from .ndmapping import NdMapping, item_check
 from .spaces import HoloMap
 from . import util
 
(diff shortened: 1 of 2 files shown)


jlstevens:
I think it would be worth summarizing the status of sorted/unsorted containers. Here is what I remember and it seems to have gotten confused again:

* Early on in HoloViews, ``NdMappings`` were always sorted.
* When ``Table`` was introduced sorting was switched off, at least for some of the time.
* With the new interface classes, I was under the impression that sorting was now off by default.

In this PR is seems that sorting is still assumed in some places which is why I think we need to carefully sort out when and where we sort. :-)

philippjfr:
It's always been well defined, NdMappings are always sorted, while Columns is not. When you groupby a Columns type the keys you're grouping over have to be sorted, hence this change. Table was inherited from NdElement previously, which specifically extended NdMapping so we could allow unsorted and duplicate entries.

jlstevens:
Thanks for the summary! 

It does makes sense now and this PR seems safe: I don't think removing the context manager will have any impact other than leaving the sorting behavior enabled for ``NdMappings``.

I'm happy to merge once the tests pass.

philippjfr:
It seems I did make a half-hearted attempt at actually supporting unsorted NdMappings but at this point a lot of the code in HoloViews assumes NdMappings are sorted so it didn't really work. If we do want to support it in future we'd have to carefully review where it causes issues. I do think we should decide to commit to one or the other but if we decide to keep things sorted we should make use of the fact things are sorted to speed up NdMappings in a number of ways.

For now this is the correct fix and it should hopefully pass in a minute.

jlstevens:
I think always having sorted ``NdMappings`` is fine as long as they assumption that the items are sorted doesn't creep in elsewhere (which shouldn't happen as interfaces keep different data formats separate).

philippjfr:
Ready to merge now.

Issue #580, opened 2016-03-29
Fix lexicographical sort for heterogeneous types on Columns interfaces


philippjfr:
Currently the Columns interfaces (except DFColumns) use lexicographical sorting, which is often not desirable. We should ensure that numbers are sorted by their value not the lexicographical value or actually provide control over the kind of sorting that is applied. Note that this is not trivial as there is no efficient numpy sorting implementation that I'm aware of, which does not rely on constructing a structured array.

jlstevens:
What is wrong with [lexical sorting](http://mathworld.wolfram.com/LexicographicOrder.html)?

It just means that the first column (kdim) is sorted, then for equal values in that first column, the second column is then sorted etc.

philippjfr:
As you rightly point out lexicographical sorting is what we want, it's just that heterogeneous types on DictColumns doesn't currently work as they are being cast to a homogeneous array of object type, which cannot be sorted correctly.

jlstevens:
Ok the PR is merged now so I'll close the issue.

PR #583, opened 2016-03-29
Correctly sort heterogeneous types in DictColumns

philippjfr:
As suggested in #580 this replaces the lexicographical sorting implemented on ArrayColumns and DictColumns with more useful general sorting as is already in place on the DFColumns and NdColumns interfaces. In the case of ArrayColumns we can simply create a record array view into the actual array, which we can apply the sorting to. For DictColumns we need to actually construct a record array but this shouldn't be any more expensive than the existing implementation, which already creates an intermediate array to compute the sort order on.

This should be ready to merge once the tests pass.

### DIFF
--- a/holoviews/core/data.py
+++ b/holoviews/core/data.py
@@ -985,8 +985,13 @@ class ArrayColumns(DataColumns):
     @classmethod
     def sort(cls, columns, by=[]):
         data = columns.data
-        idxs = [columns.get_dimension_index(dim) for dim in by]
-        return data[np.lexsort(np.flipud(data[:, idxs].T))]
+        if len(by) == 1:
+            sorting = cls.values(columns, by[0]).argsort()
+        else:
+            dtypes = (columns.data.dtype,)*columns.data.shape[1]
+            sort_fields = tuple('f%s' % columns.get_dimension_index(d) for d in by)
+            sorting = columns.data.T.view(dtypes).argsort(order=sort_fields)
+        return data[sorting]
 
 
     @classmethod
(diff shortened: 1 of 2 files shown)


holoviews/core/data.py
```diff
@@ -557,6 +557,19 @@ def array(cls, columns, dimensions):
         return Element.array(columns, dimensions)
 
     @classmethod
+    def recarray(cls, columns, dimensions):
+        if dimensions is None:
```
jlstevens: Seems to be an odd method to introduce just for sorting purposes, especially as it is only needed to figure out a sort order (via ``argsort`` on the returned record array).

Generally I was hoping to avoid record arrays entirely and wouldn't want them to spread elsewhere. That said, the implementation here is fine and right now this method isn't used in a way that is likely to cause problems later. I don't yet understand why sorting record arrays is a simpler approach but I assume you wouldn't introduce them without a good reason!

jlstevens:
I've made a comment but now I'm wondering what was wrong with the originally intended (lexicographical) sorting behavior...

jlstevens:
In Lancet, I use lexical sorting without record arrays. The default sort in Python is stable so cycles of sorting do the job. Here is the [relevant code in Lancet](https://github.com/ioam/lancet/blob/f807dfcc89de9cf00e095120a8608bd3c298661d/lancet/core.py#L430).

Not saying it is an efficient approach but it is simple, works for heterogenous types and doesn't use record arrays.

philippjfr:
Okay, so the problem wasn't lexicographical sorting but the fact that it only works on homogeneous arrays. I think the record array approach is still the way to go but I'm happy to make the recarray method into a function that specifically computes the lexsorted indices.

jlstevens:
Ok, that sounds reasonable. I would make it a utility function and maybe call it ``lexsorted_indices``?

After that, I'm happy to merge once the tests pass.

PR #588, opened 2016-04-01
Added support for dynamic operations and overlaying

philippjfr:
As the title says, this adds a ``dynamic`` parameter to the ``ElementOperation`` class, which let's the user apply lazy analyses to a HoloMap (or GridSpace of HoloMaps), which will only be evaluated when the user requests a specific key. This makes it trivial to apply complex analyses to large datasets without having to apply the analysis everywhere at once and avoids having to manually wrap the analysis in a dynamic callback.

### DIFF
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -72,33 +72,42 @@ class Operation(param.ParameterizedFunction):
 
 class DynamicOperation(Operation):
     """
-    Dynamically applies a function to the elements of a HoloMap
+    Dynamically applies an operation to the elements of a HoloMap
     or DynamicMap. Will return a DynamicMap wrapping the original
     map object, which will lazily evaluate when a key is requested.
+    The _process method should be overridden in subclasses to apply
+    a specific operation, DynamicOperation itself applies a no-op,
+    making the DynamicOperation baseclass useful for converting
+    existing HoloMaps to a DynamicMap.
     """
 
-    def __call__(self, map_obj, function, **kwargs):
-        callback = self._dynamic_operation(map_obj, function, **kwargs)
+    def __call__(self, map_obj, **params):
+        self.p = param.ParamOverrides(self, params)
+        callback = self._dynamic_operation(map_obj)
         if isinstance(map_obj, DynamicMap):
             return map_obj.clone(callback=callback, shared_data=False)
         else:
             return self._make_dynamic(map_obj, callback)
 
 
-    def _dynamic_operation(self, map_obj, function, **kwargs):
+    def _process(self, element):
+        return element
+
+
+    def _dynamic_operation(self, map_obj):
         """
         Generate function to dynamically apply the operation.
         Wraps an existing HoloMap or DynamicMap.
         """
         if not isinstance(map_obj, DynamicMap):
             def dynamic_operation(*key):
-                return function(map_obj[key], **kwargs)
+                return self._process(map_obj[key])
             return dynamic_operation
 
         def dynamic_operation(*key):
             key = key[0] if map_obj.mode == 'open' else key
             _, el = util.get_dynamic_item(map_obj, map_obj.kdims, key)
-            return function(el, **kwargs)
+            return self._process(el)
 
         return dynamic_operation
 
@@ -112,6 +121,26 @@ class DynamicOperation(Operation):
         params = util.get_param_values(hmap)
         kdims = [d(values=list(values)) for d, values in zip(hmap.kdims, dim_values)]
         return DynamicMap(dynamic_fn, **dict(params, kdims=kdims))
+
+
+
+class DynamicFunction(DynamicOperation):
+    """
+    Dynamically applies a function to the Elements in a DynamicMap
+    or HoloMap. Must supply a HoloMap or DynamicMap type and will
+    return another DynamicMap type, which will apply the supplied
+    function with the supplied kwargs whenever a value is requested
+    from the map.
+    """
+
+    function = param.Callable(default=lambda x: x, doc="""
+        Function to apply to DynamicMap items dynamically.""")
+
+    kwargs = param.Dict(default={}, doc="""
+        Keyword arguments passed to the function.""")
+
+    def _process(self, element):
+        return self.p.function(element, **self.p.kwargs)
 
 
 
@@ -176,7 +205,7 @@ class ElementOperation(Operation):
             processed = GridSpace(grid_data, label=element.label,
                                   kdims=element.kdims)
         elif dynamic:
-            processed = DynamicOperation(element, self, **params)
+            processed = DynamicFunction(element, function=self, kwargs=params)
         elif isinstance(element, DynamicMap):
             if any((not d.values) for d in element.kdims):
                 raise ValueError('Applying a non-dynamic operation requires '
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -106,7 +106,6 @@ class DynamicOperation(Operation):
                 el = map_obj[key]
             elif key >= map_obj.counter:
                 el = next(map_obj)
-                key = map_obj.keys()[-1]
             return function(el, **kwargs)
 
         return dynamic_operation
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -5,12 +5,18 @@ the purposes of analysis or visualization.
 from functools import reduce
 import param
 
+try:
+    from itertools import izip as zip
+except:
+    pass
+
 from .dimension import ViewableElement
 from .element import Element, HoloMap, GridSpace, Collator
 from .layout import Layout
 from .overlay import NdOverlay, Overlay
+from .spaces import DynamicMap
 from .traversal import unique_dimkeys
-
+from . import util
 
 
 class Operation(param.ParameterizedFunction):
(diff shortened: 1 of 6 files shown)


holoviews/core/operation.py
```diff
@@ -115,19 +125,41 @@ def __call__(self, element, **params):
...
-            mapped_items = [(k, self._process(el, key=k))
-                            for k, el in element.items()]
-            refval = mapped_items[0][1]
-            processed = element.clone(mapped_items,
-                                      group=refval.group,
-                                      label=refval.label)
+            if self.p.dynamic:
+                def dynamic_operation(key):
+                    return self(element[key], **params)
+                processed = self._make_dynamic(element, dynamic_operation)
```
jlstevens: Minor suggestion: I might consider using the signature ``self._make_dynamic(element, key, params)`` and moving the nested function into that method (or using a lambda).

This is a very mild preference and if you prefer to leave it as it is, I don't mind either.

holoviews/core/operation.py
```diff
@@ -115,19 +125,41 @@ def __call__(self, element, **params):
...
+                refval = mapped_items[0][1]
+                processed = element.clone(mapped_items,
+                                          group=refval.group,
+                                          label=refval.label)
         else:
             raise ValueError("Cannot process type %r" % type(element).__name__)
         return processed
 
 
+    def _make_dynamic(self, hmap, dynamic_fn):
```
jlstevens: Actually, how about making this a general utility to turn ``HoloMaps`` into ``DynamicMaps``? That seems like it could be useful and would justify keeping the existing signature (overriding my suggestion above).

jlstevens:
Looks good and I can see how it would be useful. I've made two small comments.

There is only one other thing I think is necessary for this to be complete (though this PR can be merged as a step towards this goal). If the input is a ``DynamicMap`` I would expect it to return a ``DynamicMap`` by default. The question then is what the ``dynamic`` parameter means. The behaviour I would expect:

* ``HoloMap`` -> ``HoloMap`` (default)
* ``HoloMap`` -> ``DynamicMap`` (if explicitly requested) 
* ``DynamicMap`` -> ``DynamicMap`` (default)
* ``DynamicMap`` -> ``HoloMap`` (if explicitly requested)

A single ``dynamic`` parameter doesn't capture both those defaults. I would rather avoid having to always supply ``dynamic=True`` when working with ``DynamicMap``. How about some other parameter, something like:

```python
cast = param.ObjectSelector(default=None, objects=[None, 'static', 'dynamic'])
```

Then instead of ``operation(hmap, dynamic=True)`` it would be ``operation(hmap, cast='dynamic')``?

I think another way would be to specify the container class eg ``operation(hmap, container=hv.HoloMap)`` which might be nicer in some ways but could potentially be more verbose. At least ``hv.HoloMap`` and ``hv.DynamicMap`` would tab complete!

philippjfr:
Don't particularly like ``cast`` as a name because it's not casting anything. ``container`` is also confusing to me as you could be passing in a GridSpace of HoloMaps and it would turn the HoloMaps into DynamicMaps but it's not obvious what container refers to in that case. I'd prefer sticking with ``dynamic`` and set it to either ``None`` or ``'default'``by default, then it defaults to what you pass in but you can explicitly override it.

philippjfr:
I've now factored out a DynamicOperation, which takes a HoloMap/DynamicMap and a function as input and returns a DynamicMap, which will apply that function lazily to each Element as it is loaded. The ``ElementOperation`` simply makes use of this to apply itself to its input. I've gone with setting dynamic to 'default' and allowing the user to explicitly force it to True/False.

Here's a nice little example:

```python
hv.operation.contours(hv.DynamicMap(lambda x: hv.Image(np.random.rand(10,10))),
                      levels=[0, 0.25, 0.5, 0.75, 1])
```

image

philippjfr:
As the title says, I've also folded in dynamic overlaying which turned out not to be a major problem having implemented the DynamicOperation. I think have the basic operators work is pretty important and we can focus on the remaining methods on DynamicMap at another point. This is ready to merge once you've reviewed it.

holoviews/core/operation.py
```diff
@@ -64,6 +70,50 @@ def get_overlay_bounds(cls, overlay):
...
 
+class DynamicOperation(Operation):
+    """
+    Dynamically applies a function to the elements of a HoloMap
+    or DynamicMap. Will return a DynamicMap wrapping the original
+    map object, which will lazily evaluate when a key is requested.
+    """
+
+    def __call__(self, map_obj, function, **kwargs):
+        callback = self._dynamic_operation(map_obj, function, **kwargs)
```
jlstevens: I would prefer it if ``__call__`` has the same signature as ``Operation`` and if instead of passing in ``function`` as an argument you use the ``_process`` method. This would make ``DynamicOperation`` more consistent with ``Operation`` and you could always create a very generic ``DynamicOperation`` subclass that takes in a function from the outside to achieve the same functionality.

holoviews/core/operation.py
```diff
@@ -74,6 +124,13 @@ class ElementOperation(Operation):
     ElementOperation may turn overlays in new elements or vice versa.
     """
 
+    dynamic = param.ObjectSelector(default='default',
+                                   objects=['default', True, False], doc="""
+       Whether the operation should be applied dynamically when a
+       specific frame is requested, specified as a Boolean. If set to
+       'default' the mode will be determined based on the input type,
+       i.e. if the data is a DynamicMap it will stay dynamic.""")
+
```
jlstevens: I'm happy with this approach.

holoviews/core/spaces.py
```diff
@@ -97,6 +98,56 @@ def _dimension_keys(self):
                 for k in self.keys()]
 
 
+    def _dynamic_mul(self, dimensions, other, keys):
+        """
```
jlstevens: Glad to see this functionality but sadly it makes ``__mul__`` into even more of a monster. I've always felt that ``__mul__`` should be implemented once as a separate utility/operation one day. If you agree, I can file an issue to suggest this.

holoviews/plotting/plot.py
```diff
@@ -515,19 +515,7 @@ def _get_frame(self, key):
...
-            elif key < self.hmap.counter:
-                key_offset = max([key-self.hmap.cache_size, 0])
-                key = self.hmap.keys()[min([key-key_offset, len(self.hmap)-1])]
-                frame = self.hmap[key]
-            elif key >= self.hmap.counter:
-                frame = next(self.hmap)
-                key = self.hmap.keys()[-1]
-            else:
-                frame = None
+            key, frame = util.get_dynamic_item(self.hmap, self.dimensions, key)
```
jlstevens: Nice to see this as a utility!

holoviews/core/util.py
```diff
@@ -886,3 +886,27 @@ def arglexsort(arrays):
         recarray['f%s' % i] = array
     return recarray.argsort()
 
+
+def get_dynamic_item(map_obj, dimensions, key):
+    """
+    Looks up an item in a DynamicMap given a list of dimensions
+    and a corresponding key. The dimensions must be a subset
+    of the map_obj key dimensions.
+    """
```
jlstevens: An example would be nice in the docstring here. Doesn't have to be an actual doctest if that is too tricky to get working...

tests/testdynamic.py
```diff
@@ -78,3 +79,55 @@ def test_sampled_bounded_resample(self):
         dmap=DynamicMap(fn, sampled=True)
         self.assertEqual(dmap[{0, 1, 2}].keys(), [0, 1, 2])
 
+
+class DynamicTestOperation(ComparisonTestCase):
```
jlstevens: Great! Thanks for adding these tests!

holoviews/core/spaces.py
```diff
@@ -146,6 +202,11 @@ def __mul__(self, other):
                     items.append((new_key, Overlay([other[other_key]])))
             return self.clone(items, kdims=dimensions, label=self._label, group=self._group)
         elif isinstance(other, self.data_type):
+            if isinstance(self, DynamicMap):
+                from .operation import DynamicOperation
+                def dynamic_mul(element):
+                    return element * other
+                return DynamicOperation(self, dynamic_mul)
```
jlstevens: This inner function is so small, why not use a lambda?

philippjfr: Prefer to use a proper function because at least it is named. Don't really care though.

holoviews/core/spaces.py
philippjfr: I do agree, a separate utility makes sense to me, so please do open an issue.

jlstevens:
Looks good and I'm very happy with the new functionality here. I've made some comments, none of which should take long to implement (or respond to). Once addressed, I'm happy to merge.

holoviews/core/operation.py
philippjfr: Sounds good, happy to do this and shouldn't take long.

holoviews/core/util.py
philippjfr: Will do, can add a unit test too.

holoviews/core/spaces.py
jlstevens: Done. See issue #600.

jlstevens:
I'm happy with the latest changes and I am ready to merge. I wish I could think of a name better than ``DynamicFunction`` though! 

If you have any suggestions, it would be good to get that in now. If now, we can always rename it in a later PR after the merge.

philippjfr:
Don't have any immediate suggestion. As you say, we can always rename it later as it's mostly an internal detail for now.

jlstevens:
I understand it is now ready to merge.

PR #596, opened 2016-04-08
Added dynamic throttling of events to live/dynamic widgets

philippjfr:
This PR adds dynamic throttling of events to the widgets. The first time a widget dynamically requests a frame from the Python kernel it will instantiate ``this.time``, which will then be used to calculate the time taken to generate that frame. Each subsequent time a new frame is requested the dynamic_update method will first check whether sufficient time has passed since the last frame was requested, if not it will queue the event. Whenever a callback completes it will execute the last queued event if there is any. This approach ensures that events generated faster than they can be processed are thrown away, while still ensuring that the last event that the user generated is processed, which means that the slider and the plot stay in sync.

Here's what I used for testing:

```python
def a(i):
    time.sleep(1)
    return hv.Image(np.random.rand(10,10))

hv.DynamicMap(a, kdims=[hv.Dimension('x', range=(0, 1.))])
```

This also let me measure the speed of plotting of updates for the first time. For this example, if you disable the sleep, I'm getting ~30 ms per frame in matplotlib and ~35ms per frame using bokeh.

### DIFF
--- a/holoviews/plotting/bokeh/bokehwidgets.js
+++ b/holoviews/plotting/bokeh/bokehwidgets.js
@@ -45,12 +45,6 @@ var BokehMethods = {
 		if (current === undefined) {
 			return
 		}
-		if (this.time === undefined) {
-			// Do nothing the first time
-		} else if (this.timer === undefined | ((this.time + this.timer) > Date.now())) {
-			this.queue.push(current);
-			return
-		}
 		this.time = Date.now()
 		if(this.dynamic) {
 			current = JSON.stringify(current);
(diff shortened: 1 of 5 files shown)


philippjfr:
Doesn't yet work very well for the scrubber widget. Will have to think about how to deal with that.

philippjfr:
Okay, I've now added handling for the scrubber widget as well and made some fixes to the queuing. The scrubber widget now simply waits until a frame is ready to request a new one on play.

philippjfr:
This is ready to review now, it would be best if whoever reviews it also tries it out. I've tested all the permutations of live, dynamic and embedded widget with matplotlib and bokeh but since there are quite a few changes I'd feel more comfortable if someone independently confirmed things are still working as expected.

jlstevens:
I'm happy to review it but in order to spend some testing it and making sure it is ok, it probably won't be merged before Monday.

jlstevens:
As promised, I'm not testing this PR. It is certainly a huge improvement to responsiveness!

There is one issue I've noticed though (tested on JupyterHub). Trying the ``DynamicMap`` tutorial, I am using huge arrays to slow things down:

```python
x,y = np.mgrid[-1000:1001, -1000:1001] * 0.1

def sine_array(phase, freq):
    return np.sin(phase + (freq*x**2+freq*y**2))

def sine_image(phase, freq):
    return hv.Image(np.sin(phase + (freq*x**2+freq*y**2)))

dmap = hv.DynamicMap(sine_image, kdims=[hv.Dimension('phase',range=(0, np.pi)),
                                 hv.Dimension('frequency', range=(0.01,np.pi))])
dmap
```

I notice the throttling but I notice I can drag the slider quickly to the end (to ``np.pi``) and the final image doesn't show that - it shows the image at some intermediate point. In other words, I see this:

screen shot 2016-04-11 at 3 46 25 pm

When I think it ought to be something more like this:

screen shot 2016-04-11 at 3 47 24 pm

philippjfr:
>I notice the throttling but I notice I can drag the slider quickly to the end (to np.pi) and the final image doesn't show that - it shows the image at some intermediate point. In other words, I see this:

Seems to be a completely unrelated issue with the Dimension ranges, somewhere I assume is checking whether the value is out of bounds with ``val < dim.range[1]`` which fails in this case. So whenever you hit the upper bound an exception is raised which clears the event queue. I think you'll see exactly the same behavior without this PR, so I still think this can be merged asap.

jlstevens:
Well at least we know what is going wrong in this case!

Given my tests worked well (except for the example above) and everything is certainly more responsive without all the events queuing up, I will merge this PR now.

Issue #602, opened 2016-04-14
NameError: global name 'get_ipython' is not defined.


mforbes:
The following code fails when trying to update the widget controls:

```python
import holoviews as hv
hv.notebook_extension()
import ipywidgets
import numpy as np

@ipywidgets.interact(w=(1,10,0.01))
def f(w=1.0):
    x = np.linspace(0,1,100)
    return hv.Curve((x, np.sin(6*w*x)))
```

```python
NameError                                 Traceback (most recent call last)
.../IPython/core/formatters.pyc in __call__(self, obj)
    337                 pass
    338             else:
--> 339                 return printer(obj)
    340             # Finally look for special method names
    341             method = _safe_get_formatter_method(obj, self.print_method)

.../holoviews/ipython/display_hooks.py in pprint_display(obj)
    219 
    220     # If pretty printing is off, return None (fallback to next display format)
--> 221     ip = get_ipython()  #  # noqa (in IPython namespace)
    222     if not ip.display_formatter.formatters['text/plain'].pprint:
    223         return None

NameError: global name 'get_ipython' is not defined
```

Not exactly sure what the issue is, but simply importing this fixes the issue:

```python
from IPython import get_ipython
```

Is there something deeper that is broken here?

Dev version of holoviews `Version((1, 4, 3)` from master and version 4.1.1 of `ipywidgets`.

philippjfr:
Just merged #603 to fix this. Closing.

PR #603, opened 2016-04-14
BUG: Resolve issue #602 - import get_ipython.

mforbes:


philippjfr:
Assume this is a recent change in IPython, ``get_ipython`` always used to magically appear in the namespace as long as you were running in an IPython context.

philippjfr:
Will go ahead and merge, certainly safer than assuming that it just appears in the namespace magically.

PR #606, opened 2016-04-15
Added bbox_extra_artists to improve bounding box calculation

philippjfr:
This PR adds a default entry to the ``MPLPlot`` handles to register any additional artists, which are not automatically tracked by the matplotlib figure. These are then taken into account when computing a tight bounding box on the renderer. This fixes various cases where colorbars, axis labels and legends were ignored and therefore completely or partially cut-off.

### DIFF
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -563,6 +563,8 @@ class ColorbarPlot(ElementPlot):
             self._adjust_cbar(cbar, label, dim)
             self.handles['cax'] = cax
             self.handles['cbar'] = cbar
+            ylabel = cax.yaxis.get_label()
+            self.handles['bbox_extra_artists'] += [cax, ylabel]
             ax_colorbars.append((artist, cax, spec, label))
 
         for i, (artist, cax, spec, label) in enumerate(ax_colorbars[:-1]):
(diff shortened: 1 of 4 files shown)


jlstevens:
Looks good and the code makes sense.

Is there any rule to know what artists will be automatically tracked in the bounding box calculation on the matplotlib figure? Or is this simply found via trial and error?

I would hope we won't need to think much about adding to ``bbox_extra_artists`` in the future, assuming we've managed to address the problematic cases. There is also the question of *why* matplotlib isn't tracking these artists for us. Is this fix simply compensating for a matplotlib bug?

philippjfr:
>Is there any rule to know what artists will be automatically tracked in the bounding box calculation on the matplotlib figure? Or is this simply found via trial and error?

There's a method on figures called ``get_children`` which determines which objects are included. This should include all axes and artists but in some cases does not seem to. The other thing is that ``get_bbox_default_artists`` ignores invisible artists, which makes sense when you compute a new bounding box for each frame, but doesn't make sense for determining a global bounding box for all frames, as we have done, to ensure the figure doesn't jump around.

jlstevens:
Is this ready to merge?

philippjfr:
Yes.

Issue #590, opened 2016-04-05
Setting integer fontsize


philippjfr:
This used to be supported but no longer seems to be:

```python
%%opts Image [fontsize=10]
hv.Image(np.random.rand(10,10), group='A')
```

This is because it no longer seems to handle integer input at all:

```python
/Users/philippjfr/topographica/external/holoviews/holoviews/plotting/plot.pyc in _fontsize(self, key, label, common)
    275     def _fontsize(self, key, label='fontsize', common=True):
    276         if not self.fontsize: return {}
--> 277         unknown_keys = set(self.fontsize.keys()) - set(self._fontsize_keys)
    278         if unknown_keys:
    279             msg = "Popping unknown keys %r from fontsize dictionary.\nValid keys: %r"

AttributeError: 'int' object has no attribute 'keys'
```

jlstevens:
Ok, I'm happy to look into this. 

So do floats still work because if they do (but integers don't) then that would be very odd!

philippjfr:
As far as I can tell only dictionaries work.

philippjfr:
Fixed in PR referenced above. Closing.

PR #607, opened 2016-04-15
Fixed fontsize option when supplying a single numeric value

jlstevens:
This PR addresses issue #590 by re-enabling numeric values for the fontsize option. It seems to be working again although I do seem to have some trouble changing the tick label fontsizes. I'll have to investigate and make sure it is hooked up properly...

### DIFF
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -274,21 +274,24 @@ class DimensionedPlot(Plot):
 
     def _fontsize(self, key, label='fontsize', common=True):
         if not self.fontsize: return {}
+
+        if not isinstance(self.fontsize, dict):
+            return {label:self.fontsize} if common else {}
+
         unknown_keys = set(self.fontsize.keys()) - set(self._fontsize_keys)
         if unknown_keys:
             msg = "Popping unknown keys %r from fontsize dictionary.\nValid keys: %r"
             self.warning(msg %  (list(unknown_keys), self._fontsize_keys))
             for key in unknown_keys: self.fontsize.pop(key, None)
 
-        if isinstance(self.fontsize, dict):
-            if key in self.fontsize:
-                return {label:self.fontsize[key]}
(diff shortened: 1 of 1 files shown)


jlstevens:
The 'ticks' fontsize uses ``common=False`` which means it is unaffected by a common, scalar fontsize value. It can still be specified explicitly with the dictionary format though.

I can't quite remember why we chose to do this but I suspect it might be because scaling the tick label size together with everything else rarely works well.

Maybe there should some catch all dictionary option for all  the non-tick labels? What would it be called? After grepping the code, it seems ticks is the only thing that doesn't follow ``common=True``...

jlstevens:
The full set of fontsize keys is 'xlabel','ylabel', 'labels', 'ticks', 'title', 'legend', 'legend_title'.

Using labels to set both axis labels together this reduces to 'ticks', 'title', 'legend', 'legend_title'. As result I am tempted to say ticks should use ``common=True`` to help make the user switch to the explicit dictionary syntax. 

This would break backwards compatibility though so maybe we could just add support for a ``'common'`` key for everything other than ticks? This name isn't very good though...semantically we want something that means 'everything other than ticks'.

Lastly, I'll note that the fontsize for ticks on colorbars don't seem to be customizable yet.

philippjfr:
The ticks issue is difficult, it's confusing that it doesn't respond to a global value but you generally don't want the tick labels to be the same size as axis labels and titles and specifying all of them is annoying.

>Edit: Perhaps the real distinction is font size formatting for numeric/non-numeric text?

Don't think that's necessarily the distinction, you can have categorical string based tick labels and since those are usually longer the need to have these be smaller than the rest of the text is usually even greater. As you suggest a common keyword meaning everything except ticks is probably the best approach, can't think of a good suggestion right now though.

jlstevens:
I'm happy to continue working on this PR if you have further suggestions for improvements. Otherwise, I think it can be merged right away just as a bug fix.

philippjfr:
Okay, looks good to me. Merging.

PR #610, opened 2016-04-17
Correctly compute RasterGridPlot widths and heights

philippjfr:
Previously RasterGridPlot only considered one column when computing the row and column heights, which caused issues if the space is sparsely populated. This correctly computes the row heights and column widths by considering all Elements.


philippjfr:
Ready for review and merge.

jbednar:
Hard to argue with it; seems ok to merge to me!

philippjfr:
Okay, I'll just go ahead and do so. It fixes bugs, so definitely worth doing.

PR #612, opened 2016-04-17
Added strict checking of dimensions on Columns.groupby

philippjfr:
As the title says this makes Columns.groupby use strict checking when looking up dimensions, ensuring that proper exception messages are raised.


jlstevens:
Great!

When the strict option was added it was tricky to figure out when it should be used. This is definitely a good place to enable strict access.

I'm happy to merge as soon as the tests pass.

PR #614, opened 2016-04-17
Avoid passing default group in clone if new_type is supplied

philippjfr:
Avoids inheriting of default groups in clone when new_type is supplied avoiding inappropriate group labels.


jlstevens:
This is a simple fix and it will be good to get rid of those incorrect titles. I'll merge as soon as the tests pass.

jlstevens:
The pr build is passing so I'll merge it now.

PR #615, opened 2016-04-17
Always apply tick rotation even if no explicit ticks are supplied

philippjfr:
As the title says, in a recent refactor I only applied tick rotations if the ticks have been supplied explicitly, this always applies the rotation.


jlstevens:
The pr build is passing and I understand it is ready to merge.

PR #616, opened 2016-04-17
Switched from bootstrap slider sizing to explicit width values

philippjfr:
Fixes bugs when widgets on websites.


jlstevens:
No need to worry about the tests as none of these CSS changes would be tested anyway. Merging.

PR #617, opened 2016-04-18
fix to signature for hv.Element

corinnebosley:
Updated line to correct signature in the description of hv.Element


jlstevens:
Thanks!

I'll merge this right away now the tests have passed. In addition, I'll need to update the website and the online notebook zip archive. 

Updating the notebooks in the conda/pypi packages will have to wait till the next release which should happen in a few weeks time.

PR #620, opened 2016-04-19
Renamed Columns Element to Dataset

philippjfr:
As the title says this is a simple renaming from Columns to Dataset. This is because data interfaces are no longer required to be columnar, also supporting gridded data for instance. A general baseclass called ``Dataset`` will be very useful to wrap very high-dimensional datasets, which do not have a clear visual representation.


philippjfr:
Ready to merge, preferably via the new squash commit functionality.

philippjfr:
Need to look into why the data cache is complaining even though the tests are passing at some point.

jlstevens:
> Need to look into why the data cache is complaining, even though the tests are passing, at some point.

Ok, sounds liked you can do that in a different PR. Merging.

PR #621, opened 2016-04-19
Splitting core/data module into smaller files

philippjfr:
Depends on #620 and simply does as the title says as the core/data.py module was soon going to exceed 2000 lines.


philippjfr:
Ready to merge once #620 has been merged.

jlstevens:
Did doc/Tutorials/Columnar_Data.ipynb get missed in the previous renaming? In the files tab I see the ``Columns`` to ``Dataset`` renaming in that notebook.

Other than that, this is a good idea as the file was really getting too long. Interfaces tend to be entirely independent of one another and make use of different libraries etc so splitting them up makes a lot of sense.

Merging.

PR #622, opened 2016-04-19
Conversion refactor

philippjfr:
Depends on #620 and #621 and moves the conversion interface from Table Element to ``Dataset`` class. This exposes this API on a lower level, allows overriding the conversion interface and splits it into a baseclass for general conversion and a separate class to implement methods for specific Elements.


jlstevens:
I now see how these PRs just extend one another. It does make it harder to review as the files tab shows changes from the PRs I just merged as well so I have to look for the changes specific for this PR through 20 files worth of changes!

Anyway ``TableConversion`` to ``DataConversion`` makes sense, as does generalizing conversions off of ``Table``. I also like the way ``ElementConversion`` adds support for new elements - very nice and clean!

Looking at the code, it does seem it should work for everything and it avoids having to convert things to tables as we've done in the past. That said, maybe we should add some unit tests to cover the new possibilities? I want to make sure it really is as general as it seems to be!

I don't want that to hold up this PR though so I'll merge it now.

PR #623, opened 2016-04-20
Remove unused variables.

rhattersley:


philippjfr:
Thanks @rhattersley, looks like I just copied the matplotlib example without checking it properly. Merging now.

PR #624, opened 2016-04-20
Added interface class for iris Cube datasets

philippjfr:
This PR adds an interface for iris Cube datasets to work with large out-of-memory datasets. The interface isn't entirely complete but supports all the basic functionality to work with such a dataset.

In future we can think about also implementing the ``sample`` and ``aggregate`` methods but it should not hold up this PR.

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -15,10 +15,20 @@ from .dictionary import DictInterface
 from .grid import GridInterface
 from .ndelement import NdElementInterface
 
+datatypes = ['array', 'dictionary', 'grid', 'ndelement']
+
 try:
     import pandas as pd # noqa (Availability import)
     from .pandas import PandasInterface
+    datatypes = ['array', 'dataframe', 'dictionary', 'grid', 'ndelement']
     DFColumns = PandasInterface
+except ImportError:
+    pass
+
+try:
+    import iris # noqa (Availability import)
+    from .iris import CubeInterface
+    datatypes.append('cube')
 except ImportError:
     pass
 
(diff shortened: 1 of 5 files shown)


holoviews/core/data/__init__.py
```diff
@@ -15,13 +15,23 @@
 from .grid import GridInterface
 from .ndelement import NdElementInterface
 
+datatypes = ['array', 'dictionary', 'grid', 'ndelement']
+
```
jlstevens: Is this because you are considering setting datatypes with ``hv.core.data.datatypes = [ 'dictionary']`` for example? Makes sense, for convenience but I would like to know if there was any another reason for moving it off the class itself.

philippjfr: So I can populate it depending on the availability of the various data backends. Happy to change the way this is handled though.

jlstevens: No, this is perfectly fine - just seems more flexible this way.

tests/testirisinterface.py
```diff
@@ -0,0 +1,64 @@
...
+except ImportError:
+    raise unittest.SkipTest("Could not import iris, skipping iris interface "
+                            "tests.")
+
+from holoviews.core.data import Dataset
+from holoviews.core.data.iris import coord_to_dimension
+from holoviews.element.comparison import ComparisonTestCase
+
+class TestCube(ComparisonTestCase):
+
```
jlstevens: Great! Maybe we can get some people who are familiar with Iris tests to add some more of their own. :-)

holoviews/core/data/iris.py
```diff
@@ -0,0 +1,248 @@
...
+from .interface import Interface
+from .grid import GridInterface
+from ..ndmapping import (NdMapping, item_check, sorted_context)
+from ..spaces import HoloMap, DynamicMap
+from .. import util
+
+from iris.util import guess_coord_axis
+from holoviews.core.dimension import Dimension
+
+
```
jlstevens: The three functions below are fine. I just hope we don't accumulate *too* many functions floating outside the class itself.

jlstevens:
Looks good! Happy to merge once the tests pass and you are happy with the number of tests for the time being.

philippjfr:
Ready to merge, I think the push build is getting confused about the updated data, but the PR build passed.

jlstevens:
The pr build is the one that counts. Merging.

PR #626, opened 2016-04-20
Allow specific underscore attributes to be ignored in AttrTree

jlstevens:
The underscore was checked after santizing the identifier, allowed ``A_repr_html`` to make an unwelcome reappearance.

### DIFF
--- a/holoviews/core/tree.py
+++ b/holoviews/core/tree.py
@@ -19,6 +19,8 @@ class AttrTree(object):
     >>> t.Example.Path                             #doctest: +ELLIPSIS
     1
     """
+    _disabled_prefixes = [] # Underscore attributes that should be
+                            # ignored instead of escaped.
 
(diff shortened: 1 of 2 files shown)


jlstevens:
As typical when working with  ``AttrTree`` my first approach didn't work. :-(

I'm now trying to see if a better solution is to toggle sanitization at the class level. Right now it is applied everywhere but if the tests pass this time, I'll try disabling it just for ``AttrTree``.

jlstevens:
It looks like the tests will pass this time. If they do and you are happy with this approach, please merge.

philippjfr:
Looks fine to me. Push build will fail since the test data has changed and the latest changes haven't been merged in. Since the PR build has passed I'll merge.

jlstevens:
Ready to merge if you are happy with it.

PR #627, opened 2016-04-20
Added _dir_mode option to AttrTrees

jlstevens:
If you set ``t._dir_mode = 'user'`` on an ``AttrTree`` called ``t`` then only the keys added by the user are returned by \_\_dir\_\_ and tab-completed.

### DIFF
--- a/holoviews/core/tree.py
+++ b/holoviews/core/tree.py
@@ -34,8 +34,8 @@ class AttrTree(object):
 
     def __dir__(self):
         """
-        The _dir_mode may be 'default' or 'user' in which case only the
-        child nodes add by the user are listed.
+        The _dir_mode may be set to 'default' or 'user' in which case
+        only the child nodes added by the user are listed.
         """
         dict_keys = self.__dict__.keys()
         if self.__dict__['_dir_mode'] == 'user':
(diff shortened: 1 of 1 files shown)


jlstevens:
The tests are passing but I'm not sure this is ready to merge just yet. The last step involving ``__setattr__`` isn't being captured by ``__dir__``...

jlstevens:
I'm not hugely happy with the last commit but it does seem to work. It also brings up another issue...shouldn't we be sanitizing the final identifier as well?

```python
tree = AttrTree()
tree.aaa.bbb.ccc = 3
u.Aaaa.Bbb.ccc
3
```

Note that ``ccc`` hasn't been capitalized.

philippjfr:
I have no better suggestion and it looks fine to me. Happy to merge if you think it's ready.

jlstevens:
I've done another push force, getting rid of the messing around with ``_terminals``.

I'll test it more to check ``__dir__`` works the way I expect and then if the tests pass, I think it can be merged.

jlstevens:
I think this is ready to merge. In the end, I also added a small change to where the sanitizer is defined (makes it easier to override in subclasses).

philippjfr:
Looks fine and worked for me. Merging.

Issue #589, opened 2016-04-04
Avoid hard dependency on pandas in Bokeh backend 


philippjfr:
The charts interface in bokeh depends on pandas and since it is import it, it raises an exception causing all of holoviews to fail to import. This is only triggered when bokeh is installed and pandas isn't. We should ensure that all that happens is that Elements depending on the chart interface are simply not registered.

This was originally brought up in https://github.com/ContinuumIO/cube-explorer/pull/2

PR #628, opened 2016-04-20
Fixed hard dependencies on pandas

philippjfr:
There were two hard dependencies on pandas, which are resolved by this PR. The first involved import the PandasInterface, which is not available unless pandas was installed. The other was an issue in the bokeh backend outlined in #589.


philippjfr:
Ready to merge.

PR #629, opened 2016-04-20
Small fix for updating HeatMap labels

philippjfr:
The HeatMap value labels weren't correctly added to the handles which resulted in them never being cleared, resulting in drawing more and more overlapping text labels.


philippjfr:
Ready to merge.

jlstevens:
Almost not worth having a PR to just add something that was missing in the handles!

That said, good to know the tests ran and passed. Merging.

PR #631, opened 2016-04-20
Fixed bug in determining axis labels

philippjfr:
In the recent refactor I changed how axis labels are determined, in particular Overlays set the axis labels by getting the first Element that is found via a traverse. However some Elements don't have a simple mapping between the dimensions and the axes. Since the bottom layer of an Overlay already sets the axis label only to be overridden by the Element this simply stops OverlayPlot from setting axis labels. The behavior should be identical in all cases except when the axis labels were previously wrong.

### DIFF
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -136,7 +136,7 @@ class ElementPlot(GenericElementPlot, MPLPlot):
         """
         element = self._get_frame(key)
         self.current_frame = element
-        if not dimensions and element:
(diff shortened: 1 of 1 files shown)


philippjfr:
Ready to merge if the tests pass.

jlstevens:
Looks straightforward enough. 

If I understand it right, you are essentially disabling the labelling process when it is likely to get it wrong, knowing that it will be done correctly elsewhere.

Tests are passing now so I'll merge.

PR #633, opened 2016-04-21
Temporarily disabled certificate validation in travis.yml

philippjfr:


philippjfr:
Let's just merge and see if it helps. It seems like it has correctly updated the ``s3-reference-data-cache`` at least so it seems like it's working.

jlstevens:
Ok sure.

PR #634, opened 2016-04-21
Fixed for matplotlib ColorbarPlot with no solids

philippjfr:
Some plot types have only discrete colors, which means the colorbar has no solids. This ensures that the solid edgecolor is only set if there is actually a ``cbar.solids`` object to set it on.


jlstevens:
Looks ready to merge to me. Should I go ahead?

philippjfr:
Yes please.

PR #636, opened 2016-04-22
Added plot option to control which axis labels to display

philippjfr:
This PR adds a plot option to control which axis labels to display. Currently we provide no control over whether to display specific axis labels. While you usually want axis labels there are cases where you have complex plot arrangements with shared axes when you do want control over it. I didn't want to add separate parameters for each so I have implemented it as a list of ``['x', 'y', ('z')]``. I'm very open to other suggestions but I do think we should offer some control over this.

Also includes a small addition to GridSpace plots to ensure that only one axis label is shown for the shared axes of those plots, which uses this mechanism.

Edit: Another option would be to contract it to a single string so you just say set 'xy', 'x' or 'y'.

### DIFF
--- a/holoviews/plotting/mpl/chart3d.py
+++ b/holoviews/plotting/mpl/chart3d.py
@@ -33,6 +33,9 @@ class Plot3D(ColorbarPlot):
     bgcolor = param.String(default='white', doc="""
         Background color of the axis.""")
 
+    labelled = param.List(default=['x', 'y', 'z'], doc="""
+        Whether to plot the 'x', 'y' and 'z' labels.""")
+
     projection = param.ObjectSelector(default='3d', objects=['3d'], doc="""
         The projection of the matplotlib axis.""")
 
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -45,6 +45,9 @@ class ElementPlot(GenericElementPlot, MPLPlot):
     invert_zaxis = param.Boolean(default=False, doc="""
         Whether to invert the plot z-axis.""")
 
+    labelled = param.List(default=['x', 'y'], doc="""
+        Whether to plot the 'x' and 'y' labels.""")
+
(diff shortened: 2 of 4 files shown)


jlstevens:
Bit tricky as those strings almost sound like the axis labels themselves (which they are not). I also want to specify this as a single parameter so I'll need to think about it.

philippjfr:
>Bit tricky as those strings almost sound like the axis labels themselves (which they are not)

Good point, I'm sure we can come up with a clearer name.

jlstevens:
Maybe a simple change to the parameter name would do the trick. ``labelled_axes`` would make it clearer that this indicates the axes that are labelled and not the axis labels themselves.

philippjfr:
Yes, ``labelled_axes`` would be sensible, only other thing I can think of is ``display_labels``, which sounds more like a boolean.

philippjfr:
Ready if you think ``labelled_axes`` is a reasonable parameter name.

jlstevens:
Would just ``labelled`` be ok? If not, I'm happy with ``labelled_axes``.

What else could be labelled in a plot? We already say 'legend' when it is a legend and annotations are annotations (i.e elements).

Is ``labelled=['x','y','z']`` good enough? If so, I prefer the shorter name.

holoviews/plotting/mpl/chart3d.py
```diff
@@ -33,6 +33,9 @@ class Plot3D(ColorbarPlot):
     bgcolor = param.String(default='white', doc="""
         Background color of the axis.""")
 
+    labelled = param.List(default=['x', 'y', 'z'], doc="""
+        Specifies the list of axes to display the labels on.""")
+
```
jlstevens: Minor docstring improvement. Should probably read "Whether to plot the 'x', 'y' and 'z' labels" for consistency.

philippjfr:
Renamed and tests passing, please merge.

PR #640, opened 2016-04-26
Render matplotlib output to canvas avoiding flicker

philippjfr:
Initial prototype to render matplotlib output to canvas to completely avoid flicker. Before merge should decide whether rendering SVG to canvas always makes sense since it will no longer be vectorized. Possibly support switching between canvas and regular rendering? Also consider whether to override save-as menu to allow saving the original image not the rasterized canvas version. Finally also need to set canvas size once, rather than on each update, perhaps with some margin.

### DIFF
--- a/holoviews/plotting/mpl/mplwidgets.jinja
+++ b/holoviews/plotting/mpl/mplwidgets.jinja
@@ -1,5 +1,10 @@
 {% extends widget_template %}
 
+{% block init_frame %}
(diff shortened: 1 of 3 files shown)


jlstevens:
After discussing this with Philipp, we came to the conclusion this approach might be a good one that we want to pursue. If this PR is rebased, I would be happy to merge this approach after a bit more review.

philippjfr:
I'm closing this at last. Once we have gone through with a widget refactor the matplotlib plot updates will happen on a bokeh canvas.

PR #641, opened 2016-04-26
Allows closest to automatically capitalize group and label lookup

jlstevens:
As OptionTree already capitalizes by default this will help elements with lowercase group and label strings to lookup their options correctly.


jlstevens:
It turns out this behavior is already implemented but was shadowed by a subtle and annoying bug:  #642

As this fix isn't actually required in the end, I'm closing this PR.

PR #642, opened 2016-04-26
Instances of sanitize functions were incorrectly sharing a cache

jlstevens:
A subtle bug that was annoying and very hard to pin down!


jlstevens:
Tests are passing so I think it is ready to merge.

philippjfr:
Important bug fix. Good we caught this one as it has lots of potential for confusing behavior. Merging.

PR #644, opened 2016-04-26
Initial prototype for Dimension preset system

jlstevens:
This approach is very simple and I rather like it:

* Entirely optional.
* 6 lines of code and could have been only 4 lines of code if I only wanted to support ``(name, unit)`` lookups
* Works completely transparently with no changes to the code required anywhere else. Create a dimension and if there is a matching preset defined, the other dimension parameters are filled in.
* Works from specific to general. If a preset with a ``(name, unit)`` match is found, that is used first. If a preset with only ``name`` defined is found, that is then used, allowing the unit to be filled in from the dimension name (useful if only one possible unit makes sense).
* ``hv.Dimension.presets`` is nice an convenient for access/definition (and now has a nice repr!)

It doesn't allow sharing but I don't see an issue. If you aren't happy with a preset, make sure you define it the way you want it (e.g at the top of the notebook) before building your elements. You are free to change the preset definitions at any time.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -91,6 +91,7 @@ class Dimension(param.Parameterized):
     # Defines default formatting by type
     type_formatters = {}
     unit_format = ' ({unit})'
+    presets = {} # A dictionary-like mapping name or (name, unit) to a preset
 
     def __init__(self, name, **params):
         """
@@ -98,6 +99,11 @@ class Dimension(param.Parameterized):
         """
         if isinstance(name, Dimension):
             existing_params = dict(name.get_param_values())
+        elif (name, params.get('unit', None)) in self.presets.keys():
+                preset = self.presets[(name, params['unit'])]
+                existing_params = dict(preset.get_param_values())
+        elif name in self.presets.keys():
+            existing_params = dict(self.presets[name].get_param_values())
         else:
             existing_params = {'name': name}
 
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -91,7 +91,8 @@ class Dimension(param.Parameterized):
     # Defines default formatting by type
     type_formatters = {}
     unit_format = ' ({unit})'
-    presets = {} # A dictionary-like mapping name or (name, unit) to a preset
+    presets = {} # A dictionary-like mapping name, (name,) or
+                 # (name, unit) to a preset Dimension object
 
     def __init__(self, name, **params):
         """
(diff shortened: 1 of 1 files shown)


jbednar:
Can you explain "it doesn't allow sharing"?

jlstevens:
@philippjfr suggested that it would be useful in some way for all the elements to point to the same dimenion object (if it is a preset). This would allow you to change the range (for instance) for existing elements already using that preset.

I can see cases where that could be useful but I also think it is stateful and relies on mutation (and therefore may have unexpected results). I see dimension objects as being cheap metadata that every element is allowed to have its own copies of.

philippjfr:
> I can see cases where that could be useful but I also think it is stateful and relies on mutation (and therefore may have unexpected results). I see dimension objects as being cheap metadata that every element is allowed to have its own copies of.

In practice dimensions are shared across many objects all the time anyway. I can see how a globally shared object that automatically gets applied might lead to confusion though. I think we really want an easier way of overriding a particular dimension on a Dataset without also having to supply all the other dimensions. Then you can just supply the specific dimension you want to override to the Dataset and it will automatically be shared among all the individual Elements when you apply a groupby. That can wait though because we already support this pattern (and currently have no need to specify key dimension ranges, which would get far more verbose):

```
temp = hv.Dimension('Temperature', range=(0, 100))
dataset = hv.Dataset(data, vdims=[temp])
images = data.to.image(['longitude', 'latitude'])
temp.range = (0, 50)
```

jlstevens:
@philippjfr I think this PR is ready to be merged now. Tests are passing so please review it and merge if you are happy with it.

holoviews/core/dimension.py
```diff
@@ -91,13 +91,22 @@ class Dimension(param.Parameterized):
...
+                 # (name, unit) to a preset Dimension object
 
     def __init__(self, name, **params):
         """
         Initializes the Dimension object with the given name.
         """
         if isinstance(name, Dimension):
             existing_params = dict(name.get_param_values())
+        elif (name, params.get('unit', None)) in self.presets.keys():
+                preset = self.presets[(str(name), str(params['unit']))]
```
philippjfr: What's up with the indentation here?

philippjfr:
Looks fine to me apart from a small indentation issue I've commented on. Would also be nice to see one or two unit tests since they should be pretty straightforward to write but that won't stop me from merging if you don't have the time right now.

jlstevens:
Please merge it. I'll make a local branch right away to add unit tests...that way I won't forget to add them later!

holoviews/core/dimension.py
jlstevens: Well spotted! Fixing it now.

philippjfr:
Thanks for the fix. Merging.

Issue #647, opened 2016-04-27
Issue with groupby and conversion interface


bjlittle:
Using this cube, there are issues using groupby and conversion:

```
air_temperature / (K)               (time: 2; model_level_number: 70; grid_latitude: 810; grid_longitude: 621)
     Dimension coordinates:
          time                           x                      -                  -                    -
          model_level_number             -                      x                  -                    -
          grid_latitude                  -                      -                  x                    -
          grid_longitude                 -                      -                  -                    x
     Auxiliary coordinates:
          forecast_period                x                      -                  -                    -
          level_height                   -                      x                  -                    -
          sigma                          -                      x                  -                    -
     Scalar coordinates:
          forecast_reference_time: 2016-02-07 21:00:00
     Attributes:
          Conventions: CF-1.5
          STASH: m01s16i004
          source: Data from Met Office Unified Model
          um_version: 10.1
```

Trying this:

```python
cube = hv.Dataset(air_temperature_cube)
cube.to.curve('model_level_number', dynamic=True)
```

Shows the first frame correctly but afterwards matplotlib complains because the two input arrays have very different lengths. Looking at the sliced cube, all the dimensions are still there.

The same issue exists when using groupby:

```
cube = hv.Dataset(air_temperature_cube)
cube.groupby(['time', 'grid_longitude', 'grid_latitude'], group_type=hv.Curve, dynamic=True)
```

Or the explicit conversion syntax:

```
cube = hv.Dataset(air_temperature_cube)
cube.to(hv.Curve, ['model_level_number'], dynamic=True)
```
What is really weird is how the first frame looks right but then it breaks!

philippjfr:
I don't have access to the same cube here so I can't reproduce it. Could you print the summary of one of the cubes that are extracted using the groupby?

bjlittle:
Here is an example of a Curve generated by the groupby for one of the keys that fails:

```
air_temperature / (K)               (time: 2; model_level_number: 70; grid_latitude: 810; grid_longitude: 621)
     Dimension coordinates:
          time                           x                      -                  -                    -
          model_level_number             -                      x                  -                    -
          grid_latitude                  -                      -                  x                    -
          grid_longitude                 -                      -                  -                    x
     Auxiliary coordinates:
          forecast_period                x                      -                  -                    -
          level_height                   -                      x                  -                    -
          sigma                          -                      x                  -                    -
     Scalar coordinates:
          forecast_reference_time: 2016-02-07 21:00:00
[...]

PR #649, opened 2016-04-27
Sampled dynamic widgets now work by index, avoiding precision issues

philippjfr:
Currently when working in sampled dynamic mode the values will loose precision and potentially break various lookups (see #647). This PR avoids sending the float values to Javascript working by index in the dimension values instead. The way the dynamic values are sent to JS now is now redundant (a mapping from index to index) but refactoring it would take a while so I suggest postponing that for now.

### DIFF
--- a/holoviews/plotting/mpl/widgets.py
+++ b/holoviews/plotting/mpl/widgets.py
@@ -73,7 +73,8 @@ class MPLWidget(NdWidget):
 
     def update(self, key):
         if self.plot.dynamic == 'bounded' and not isinstance(key, int):
-            key = tuple(key)
+            key = tuple(dim.values[k] if dim.values else k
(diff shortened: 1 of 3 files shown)


jlstevens:
Looks good and I am happy to merge as only a few lines have changed.

Please file an issue pointing out the need for a refactor as some of the code structure/variable names are rather odd after this change.

philippjfr:
Forgot to deal with dropdowns, fix for that coming shortly.

philippjfr:
Okay that's fixed now. Just don't look at the indentation in the jinja file 😦

philippjfr:
I've now also opened #650 to discuss how best to communicate widget ranges, values and labels to the JS code.

jlstevens:
Ok. Ready to merge?

philippjfr:
Yep, good to go.

PR #657, opened 2016-05-04
Support for rotated ellipses

vascotenner:
```
hv.NdOverlay({angle: hv.Ellipse(0,0,5,aspect=2, angle=angle) for angle in np.linspace(0,np.pi/2, 6)}, kdims=['Angle'])
```

download

### DIFF
--- a/holoviews/element/path.py
+++ b/holoviews/element/path.py
@@ -178,8 +178,8 @@ class Box(BaseShape):
 class Ellipse(BaseShape):
     """
     Draw an axis-aligned ellipse at the specified x,y position with
-    the given width and aspect ratio, with a given angle between the long axis
-    and the horizontal axis. By default draws a circle (aspect=1).
+    the given width, aspect ratio and orientation. By default 
+    draws a circle (aspect=1).
 
     Note that as a subclass of Path, internally an Ellipse is a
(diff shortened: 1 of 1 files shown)


holoviews/element/path.py
```diff
@@ -193,6 +193,8 @@ class Ellipse(BaseShape):
 
     aspect= param.Number(default=1.0, doc="The aspect ratio of the ellipse.")
 
+    angle = param.Number(default=0, doc="The angle in radian between the long axis and the horizontal.")
+
```
jbednar: To be compatible with our other packages, such a parameter should probably be called `orientation`, and the description could be "Orientation in the Cartesian coordinate system, with zero at 3 o'clock and increasing counterclockwise".

philippjfr:
Apart from Jim's comment this looks like a good addition and I'll be happy to merge once you've renamed the parameter.

jlstevens:
Looks good! 

I agree the docs need to be updated to show an example of an oriented ellipse but that doesn't necessarily have to be part of this PR. I think @philippjfr or myself can easily add such an example the next time we find ourselves updating the website.

I'm also happy to see this merged once you've completed the suggested renaming.

philippjfr:
Looks good. Merging.

Issue #662, opened 2016-05-07
unique_array broken without pandas


jlstevens:
I tried running some code on a fresh desktop install and I found that the ``sample`` method was broken. I was trying the following line on a ``HoloMap`` of ``Images``:

```python
 hmap.sample((11,11), bounds=(-0.45, -0.45, 0.45, 0.45))
```

And every time, this exception was raised:

```python
TypeError: 'only integer arrays with one element can be converted to an index'
```

It turns out the problem is in ``core.util.unique array``:

```python
def unique_array(arr):
    """
    Returns an array of unique values in the input order
    """
    if pd:
        return pd.unique(arr)
    else:
        _, uniq_inds = np.unique(arr, return_index=True)
        return arr[np.sort(uniq_inds)]
```

I have no idea what is going on but then I realized that it *seems* to work only because I normally have pandas installed. We aren't supposed to be depending on pandas! 

This is a good example of why we should try to set up some Travis tests without pandas installed.

philippjfr:
This was fixed in #663, ``unique_array`` is meant to be used with arrays, which it wasn't doing in this case. Closing.

PR #663, opened 2016-05-07
Fix for HoloMap.sample without pandas

philippjfr:
Fix for #662, unique_array is meant to handle arrays and the closest call returns a list. Switched to ``unique_iterator`` to keep samples sorted.


jlstevens:
Happy to merge if it is ready.

philippjfr:
Yep, ready.

PR #664, opened 2016-05-08
Implemented DynamicMap groupby methods

philippjfr:
Also prototyped this a while back and it would be nice to get it into this release. This allows applying groupby operations to DynamicMaps as long as the Dimensions that will become part of a non-dynamic type define a fixed sampling:

```python
%%opts GridSpace [show_legend=True]
xs = np.linspace(0, 2*np.pi)
def sin(ph, f, amp):
    return hv.Curve((xs, np.sin(xs*f+ph)*amp))

hv.DynamicMap(sin, kdims=[hv.Dimension('ph', range=(0, np.pi)),
                          hv.Dimension('f', values=[0.1, 1, 2]),
                          hv.Dimension('amp', values=[0.5, 1])]).overlay('f').grid('amp')
```

image

Here we define the frequency and amplitude dimensions with a fixed sampling and then set them to be overlaid and gridded respectively. Since the phase dimension stays in a DynamicMap it can define a continuous range. The displayed object is a GridSpace of DynamicMaps, which return NdOverlays. This will be very useful to explore very large but non-continuous spaces, as we do in some of our projects.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -1,10 +1,11 @@
-from numbers import Number
 import itertools
-from itertools import groupby
-import numpy as np
-
-import param
 import types
+from numbers import Number
+from itertools import groupby
+from functools import partial
+
+import numpy as np
+import param
 
 from . import traversal, util
 from .dimension import OrderedDict, Dimension, ViewableElement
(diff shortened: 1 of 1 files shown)


philippjfr:
Also dropped requirement for either range or value to be defined. There are many cases where you might want to define a range on a dimension, e.g. cyclic dimensions, and specify a fixed sampling. The fixed sampling defined by the values will always override the range.

jbednar:
Looks very useful!

jlstevens:
Looks good. Is it ready to merge now?

philippjfr:
Yes, ready to merge, will need to add tests once I have time again.

jlstevens:
> currently these methods are just inherited from HoloMap

Any other methods that should be overridden to raise ``NotImplementedError``?

philippjfr:
>Any other methods that should be overridden to raise NotImplementedError?

That's a good idea, I'll open a new PR to do that.

PR #667, opened 2016-05-11
Added Tutorial section to explain DynamicMap groupby methods

philippjfr:


jlstevens:
Great! Thanks for updating the docs!

Ready to merge?

philippjfr:
Ready.

doc/Tutorials/Dynamic_Map.ipynb
```diff
@@ -601,6 +601,106 @@
...
+   "outputs": [],
+   "source": [
+    "%%opts GridSpace [xaxis=None yaxis=None] Path [bgcolor='w' xaxis=None yaxis=None]\n",
+    "spiral_dmap.groupby(['f', 'ph'], group_type=hv.NdOverlay, container_type=hv.GridSpace)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
```
jbednar: Line 714 below says "need to check" -- what's the answer to that question?

philippjfr: Good question, looks like Jean-Luc forgot to address his own question :smile: I think the statement is correct as is, as long as the dimension is unbounded on either end it will default to open mode.

jbednar: Ok, I'll remove the qualifier while I'm doing an editing pass on this tutorial to clean up the new changes.

jlstevens: In that case we can remove that stray comment...oops!

I can make an edit directly on GitHub if that is easiest now it is merged...

jbednar: Would "At least one of the declared key dimensions must have an unbounded range (i.e., with no upper bound specified)." be correct?

jlstevens: I think so...though Philipp suggests it works if either bounds is None, including the lower bound? I just wonder how the scrubber handles that...does it decrease starting from the upper bound?

jbednar: It just ignores all the bounds, at present; all that matters is that one bound is not specified.  I've clarified it to say that, but we're still deciding on the larger issue of how to describe these various modes.

PR #668, opened 2016-05-11
Avoid hard dependencies on plotting backends

philippjfr:
Addresses #528 in two ways. The main holoviews ``__init__.py`` no longer imports the ipython module automatically, which avoids initializing everything immediately. This means HoloViews can now be imported without all the plotting side-effects that causes, which was an issue raised by some users, e.g. see https://github.com/ioam/imagen/issues/48. Instead everything gets initialized when the notebook extension is loaded or the ipython and plotting modules are imported manually. Additionally I've made sure that the IPython extension and the magics correctly handle the case where no plotting library is available.

The only issue I see with this approach is that the notebook extension is now a simple wrapper, which does not define the parameters or a docstring. A better solution would be very welcome.

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -13,8 +13,7 @@ from ..core.options import Store, Cycle, Palette
 from ..element.comparison import ComparisonTestCase
 from ..interface.collector import Collector
 from ..plotting.renderer import Renderer
-from ..plotting import * # noqa (API import - register plotting backends)
-from .magics import load_magics
+from .magics import load_magics, list_formats
 from .display_hooks import display  # noqa (API import)
 from .display_hooks import set_display_hooks, OutputMagic
 from .widgets import RunProgress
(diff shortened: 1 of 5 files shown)


holoviews/__init__.py
```diff
@@ -32,7 +32,10 @@
 
 try:
     import IPython  # noqa (API import)
-    from .ipython import notebook_extension
+    class notebook_extension(param.ParameterizedFunction):
+        def __call__(self, *args, **opts):
+            from .ipython import load_extension
+            load_extension(*args, **opts)
```
jlstevens: Do we really need to rename to ``load_extension``? What about a qualified import, ``ipython.notebook_extension(*args, **opts)``. I would rather avoid renaming it in ``.ipython`` if possible.

philippjfr: True, that'd be fine.

jlstevens:
After discussing this with Philipp, we think it might be a good idea to defer only the plotting related imports to when the extension is called (so the related side-effects only happen if requested). 

This seems like a cleaner approach and ``hv.notebook_extension`` will still be nicely parameterized.

holoviews/ipython/__init__.py
```diff
@@ -140,8 +139,21 @@ class notebook_extension(param.ParameterizedFunction):
...
+
     def __call__(self, *args, **params):
-        resources = self._get_resources(args, params)
+        imports = [(name, b) for name, b in self._backends.items()
+                   if name in args or params.get(name, False)]
+        if not imports:
+            imports.append(('matplotlib', 'mpl'))
+        for backend, imp in imports:
+            try:
+                __import__('.'.join(['holoviews.plotting', imp]))
```
jlstevens: Don't see why you would use join instead of a simple formatting string in this instance.

holoviews/ipython/__init__.py
```diff
@@ -140,8 +139,21 @@ class notebook_extension(param.ParameterizedFunction):
 
     _loaded = False
 
+    _backends = {'matplotlib': 'mpl',
```
jlstevens: Maybe a comment here would be good saying this is a dictionary from resource name to plotting subpackage import name.

jlstevens:
If this works, then I'm much happier with this approach! I'm happy to merge once my suggestions are addressed and the tests pass.

jbednar:
Sounds great to me!  Thanks for doing this.

philippjfr:
Although the tests are failing this is ready to merge. It's the GridMatrix changes that are failing, since that updated the test data.

Issue #619, opened 2016-04-19
Support predicate functions in select method


jlstevens:
Currently we support literals, tuples (ranges) and sets in select. We could also support callables (conditions) where we would assume boolean outputs.

philippjfr:
This feature was merged.

PR #669, opened 2016-05-11
Added support for slicing/indexing with predicate functions

philippjfr:
Implements the suggestion in #619 supporting predicate functions which return Booleans for slicing on both NdMapping types and the data interfaces.


jlstevens:
Definitely useful and not much code required. Happy to merge once the tests pass.

philippjfr:
Ready to merge.

Issue #637, opened 2016-04-22
Transpose GridInterface data format


philippjfr:
The current implementation of the GridInterface expects the data to be in the following format:

```python
xs = range(10)
ys = range(20)
zs = np.ones(10, 20)
data = {'x': xs, 'y': ys, 'z': zs}
```

i.e., the value arrays must match the shape of the key dimensions in the order they are specified. However in discussing the iris format and thinking about it again I realized that generally the convention is for the shape of a 2D array to be:

```python
ylength, xlength = array.shape
```

This is also more consistent with the ways we already store the array in our ``Raster`` types (although still flipped upside down). We should therefore change the ``GridInterface`` to the transposed format before we release and users officially depend on it.

philippjfr:
@jbednar @jlstevens Thoughts on this? Need to decide on the correct orientation of the data in the grid interface before releasing 1.5.

jbednar:
I'm not sure what approach is best; the whole world seems confused about how to store matrices and rasters.  But yes, r,c is normally how the shape is specified in numpy, i.e. len(ys),len(xs).

In any case, I also can't quite tell how to interpret your example; how does anyone know that xs came before ys, given that a Python dictionary is not ordered?

philippjfr:
>In any case, I also can't quite tell how to interpret your example; how does anyone know that xs came before ys, given that a Python dictionary is not ordered?

When specifying in the dictionary format you always have to explicitly declare the dimensions as well. You can always supply a tuple though and it automatically expand it to the dictionary format:

```
hv.Dataset((xs, ys, zs), kdims=['x', 'y'], vdims=['z'])
```

Optionally we could also automatically infer dimensions if supplied an OrderedDict as we already do for the pandas interface.

>I'm not sure what approach is best; the whole world seems confused about how to store matrices and rasters. But yes, r,c is normally how the shape is specified in numpy, i.e. len(ys),len(xs).

The benefit of having it in the current format (column, row) is that when you flatten it, it corresponds directly to the default Column order. Using r, c instead seems to be convention elsewhere (like iris), while our raster are r, c but flipped upside down. I'm really not sure what to suggest but am tending toward regular r, c since that's generally expected, the inconsistency with our existing Rasters bothers me though.

philippjfr:
The PR was just merged, we're now stuck with this format. I do think it is the correct format since it is what other grid based interfaces like iris and xarray will expect.

PR #670, opened 2016-05-11
Transposed grid interface data format

philippjfr:
Addresses #637 to ensure that the new Grid interface is more consistent with the usual conventions of matrix shapes.

### DIFF
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -68,7 +68,7 @@ class GridInterface(DictInterface):
         expected = tuple([len(data[kd]) for kd in kdim_names])
         for vdim in vdim_names:
             shape = data[vdim].shape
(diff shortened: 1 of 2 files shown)


jlstevens:
Is this now inconsistent with our Rasters?

I'm happy to merge but it is always tricky to know what to do when faced with different conventions. If we use this convention we have to stick to it and my understanding is that we will just follow what everyone else seems to be using?

philippjfr:
No, it's still flipped upside down compared to our Rasters but it is the general convention used when working with ndarrays and almost everywhere else including the iris backend.

jlstevens:
Ok, I'll merge it now given that this seems to be the right thing and this is a new class which won't have many users just yet. As long as we support people using the grid interface through this change, it is the right time to make this fix.

PR #671, opened 2016-05-11
Improvements to installation and setup.py

### LINKED ISSUE #527: Dependencies missing
Hi,

I think that the holoviews pip package does not correctly state its dependencies. These are packages that holoviews complained about not finding when I tried importing it:

* jinja2 
* nbformat 
* nbconvert 
* matplotlib

After installing them manually via pip, I can import holoviews fine.

jlstevens:
This PR aims to fix #527 and implement any other small improvements we can to setup.py and the installation process.

### DIFF
--- a/meta.yaml
+++ b/meta.yaml
@@ -1,0 +1,28 @@
+package:
+  name: holoviews
+  version: 1.4.3
+
+source:
+  path: .
+
+requirements:
+  build:
+    - python
+    - param
+    - numpy
+    - setuptools
+  run:
+    - python
+    - param
+    - numpy
+    - matplotlib
+    - bokeh
+    - jupyter
+    - notebook
+    - ipython
+    - lancet
+
+about:
+  home: www.holoviews.org
+  summary: Stop plotting your data - annotate your data and let it visualize itself.
+  license: BSD
--- a/setup.py
+++ b/setup.py
@@ -101,4 +101,22 @@ if __name__=="__main__":
             import holoviews
             holoviews.__version__.verify(setup_args['version'])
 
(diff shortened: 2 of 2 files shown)


jlstevens:
Trying to detect whether the users was already using extras proved to be tricker than it was worth.

Here is an example of what you see when running ``python setup.py install``:

```
==================================
HOLOVIEWS INSTALLATION INFORMATION
==================================

HoloViews supports the following installation types:

holoviews[recommended]
holoviews[extras]
holoviews[all]
holoviews[notebook-dependencies]

Users should consider using one of these options.

By default only a core installation is performed and
only the minimal set of dependencies are fetched.


For more information please visit http://holoviews.org/install.html

==================================
```

I tried to make it quite visible as when installing this way you get a lot of messages of what is copied where. I have no idea what it looks like installing from PyPI but I assume this output will be printed.

Currently, I'm only printing this if ``install`` is in ``sys.argv``. As far as I know, that covers what we need.

philippjfr:
Makes sense to me, not sure how we can increase visibility any further. We recommend installing via conda anyway so I think this is fine.

jlstevens:
If we agree this is ok, please merge once the tests pass.

jlstevens:
Actually, it is time I add the ``meta.yaml`` file for conda seeing it has become our recommended installation route. I don't want to add ``build.sh`` yet unless it becomes a bit more than:

```bash
#!/bin/bash
$PYTHON setup.py install
```

philippjfr:
Looks good to me and PR build passed. Merging.

Issue #595, opened 2016-04-07
Raster and reduce and axes direction


vascotenner:
When I reduce the x axis of a Raster object, the direction of the Y axis is swapped.

Reducing the Y axis works fine:
```
%%opts Raster [colorbar=True]
img = hv.Raster(np.rot90(np.meshgrid(*[np.arange(5)]*2)[1][::-1]))
img+img.sample(y=2)+img.reduce(y=np.mean).to.curve('x','z')
```
![download 1](https://cloud.githubusercontent.com/assets/3776190/14356591/ca57d4f8-fce5-11e5-9707-f05b60b10617.png)

While reducing the X axis fails:
```
%%opts Raster [colorbar=True]
img = hv.Raster(np.meshgrid(*[np.arange(5)]*2)[1][::-1])
img+img.sample(x=2)+img.reduce(x=np.mean).to.curve('y','z')
```
![download](https://cloud.githubusercontent.com/assets/3776190/14356590/ca552442-fce5-11e5-8273-602e3b4f1b38.png)

philippjfr:
Fixed in #672. Closing.

PR #672, opened 2016-05-11
Fixed Raster.reduce for non Image types

philippjfr:
Fix for #595, handles the reduction of the x-axis for the Raster class itself correctly. The reduce method is currently shared between all Raster and Image classes and it only handled the Image case correctly. In future when we have integrated Raster types with the grid interfaces this should be refactored but it works fine for now.


jlstevens:
> In future when we have integrated Raster types with the grid interfaces this should be refactored but it works fine for now.

It might be worth mentioning this in a comment as a reminder.

Otherwise looks fine, the pr build is passing and I'm happy to merge.

philippjfr:
Not sure it needs a reminder, we will be forced to refactor these methods when they become part of an interface anyway.

philippjfr:
Ready to merge, PR build passed.

Issue #507, opened 2016-02-13
Reraise plotting exceptions with original traceback 


philippjfr:
Currently when a plotting call raises an exception the IPython display hook wrapper catches it and reraises the exception, but with a new traceback relative to where the error was reraised (which is not very helpful). The error should be reraised with the original traceback as described in this [StackOverflow post](http://stackoverflow.com/questions/8760267/re-raise-python-exception-and-preserve-stack-trace).

jlstevens:
Yes, I agree the traceback is wrong and it shouldn't be hard to fix using the approach you link to.

philippjfr:
I tested that approach and it didn't work for me, but it's definitely possible.

jlstevens:
This has been addressed in the PR mentioned above. Closing.

PR #674, opened 2016-05-11
Improved abbreviated exception and traceback processing

jlstevens:
PR to address #507 which eliminates the hackish code that didn't work particularly well and replaces it with a more targeted solution. We now need to use the ``abbreviated_exception`` context manager in the plotting code as appropriate. All other exceptions and tracebacks now remain unaffected.

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -33,6 +33,7 @@ Store:
 
 """
 import pickle
+import traceback
 from contextlib import contextmanager
 from collections import OrderedDict
 
@@ -109,8 +110,14 @@ class AbbreviatedException(Exception):
     def __str__(self):
         abbrev = '%s: %s' % (self.etype.__name__, self.msg)
         msg = ('To view the original traceback, catch this exception '
-               'and access the traceback attribute.')
+               'and call print_traceback() method.')
         return '%s\n\n%s' % (abbrev, msg)
+
+    def print_traceback(self):
+        """
+        Print the traceback of the exception wrapped by the AbbreviatedException.
+        """
+        traceback.print_exception(self.etype, self.value, self.traceback)
 
 
 class abbreviated_exception(object):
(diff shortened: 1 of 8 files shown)


philippjfr:
Looks good. I think this behavior will be much better than before.

jlstevens:
One issue with this approach is that Python users outside the notebook will occasionally see ``AbbreviatedExceptions`` instead of the normal ones.

Maybe I should just add ``__str__`` and ``__repr__`` methods to the new exception class to make it clear the original exception and traceback are inside?

jlstevens:
I've added the ``__str__`` method, so here is something a pure Python user might see something like this:

```
AbbreviatedException: Abbreviated ValueError: Colormap jetz is
not recognized. Possible values are: Accent, Accent_r, Blues,
Blues_r, BrBG, BrBG_r, BuGn, BuGn_r, BuPu, BuPu_r, CMRmap,
CMRmap_r, Dark2, Dark2_r, GnBu, GnBu_r, Greens, Greens_r, Greys, ...

To view the original traceback, catch this exception and access
the traceback attribute.
```

The traceback above will lead up to the point where the context manager is called.

jlstevens:
If the tests pass and you are happy with this, please merge. I do think this system will be a lot better!

jlstevens:
This new exception is related to ``BackendError`` introduced in #366. I think the new approach is better but I've tried to support the one place where ``BackendError`` was used.

jlstevens:
I think it is actually ready to merge once the tests pass. @philippjfr Are you happy with these changes?

jlstevens:
The pr build is passing while the push build has been consistently failing. As the pr build is what counts, please merge if you think it is ready.

philippjfr:
Looks good to me. I think I'll open a separate issue to remind us to go through our code and make a concerted effort to raise more and better exceptions.

PR #675, opened 2016-05-11
Cleanup of unused and undefined variables

philippjfr:
As the title says this includes all the unused and undefined variables highlighted by pylint.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -176,7 +176,7 @@ class Dimension(param.Parameterized):
         hashes are equal, all the parameters of the Dimensions are
         also equal.
         """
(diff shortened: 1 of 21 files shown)


jlstevens:
Aren't we using pyflakes? Specifically flake8?

I assume there will be a lot of overlap but let's fix flakes before 'lints'...

philippjfr:
Since this is all undefined and unused variables pylint and flake8 should have a large overlap in this case.

jlstevens:
I'll assume this is fine given the tests are passing and I'll run flake8 to check there is nothing left. Again, please lets stick with flake8 first with:

```
flake8 --ignore=E,W   .
```

PR #677, opened 2016-05-11
Dynamic Map tutorial updates

jbednar:
Did a full pass on this tutorial to fold in recent updates and clarify.  Would be worth someone looking at the changes and checking that the clarifications were in fact correct, and not misunderstandings!

One issue: I find the name of "sampled mode" confusing, as in its purest form it's not actually sampled (yet).  Would it be fair to call it "unbounded mode", in comparison to "bounded mode"?

### DIFF
--- a/doc/Tutorials/Dynamic_Map.ipynb
+++ b/doc/Tutorials/Dynamic_Map.ipynb
@@ -6,11 +6,11 @@
    "source": [
     "The [Containers](Containers.ipynb) Tutorial introduced the [HoloMap](Containers.ipynb#HoloMap), a core HoloViews data structure that allows easy exploration of parameter spaces. The essence of a HoloMap is that it contains a collection of [Elements](Elements.ipynb) (e.g. Images and Curves) that you can easily select and visualize.\n",
     "\n",
-    "HoloMaps hold Elements at specific sampled points in a multidimensional space. Although this property makes them useful for exploring high-dimensional parameter spaces, HoloMaps can very quickly consume huge amounts of memory for such spaces. For instance, a hundred samples along four orthogonal dimensions would need a HoloMap containing a hundred *million* Elements, each of which could be a substantial object that takes time to create and costs memory to store. Thus ``HoloMaps`` have some clear limitations:\n",
+    "HoloMaps hold fully constructed Elements at specifically sampled points in a multidimensional space. Although HoloMaps are useful for exploring high-dimensional parameter spaces, they can very quickly consume huge amounts of memory to store all these Elements. For instance, a hundred samples along four orthogonal dimensions would need a HoloMap containing a hundred *million* Elements, each of which could be a substantial object that takes time to create and costs memory to store. Thus ``HoloMaps`` have some clear limitations:\n",
     "\n",
-    "* HoloMaps may require the generation of millions of elements before they are fully defined.\n",
+    "* HoloMaps may require the generation of millions of Elements before the first element can be viewed.\n",
     "* HoloMaps can easily exhaust all the memory available to Python.\n",
-    "* HoloMaps can simultaneously exhaust all the memory in the browser when displayed.\n",
+    "* HoloMaps can even more easily exhaust all the memory in the browser when displayed.\n",
     "* Static export of a notebook containing HoloMaps can result in impractically large HTML files.\n",
     "\n",
     "The ``DynamicMap`` addresses these issues by computing and displaying elements dynamically, allowing exploration of much larger datasets:\n",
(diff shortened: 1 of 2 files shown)


philippjfr:
How about 'sampling' mode? Doesn't imply it's sampled but indicates you can sample it.

doc/Tutorials/Dynamic_Map.ipynb
```diff
@@ -15,25 +15,33 @@
...
     "* DynamicMaps do not require fixed sampling, allowing exploration of parameters with arbitrary resolution.\n",
     "* DynamicMaps are lazy in the sense they only compute only as much data as the user wishes to explore.\n",
     "\n",
     "The limitations of ``DynamicMaps`` are:\n",
     "\n",
-    "* DynamicMaps require a live notebook server and cannot be directly exported to static HTML.\n",
-    "* DynamicMaps do not hold data which reduces the utility of a pickling DynamicMap.\n",
-    "* DynamicMaps are often stateful as are their element caches.\n",
+    "* DynamicMaps require a live notebook server and cannot be fully exported to static HTML.\n",
+    "* DynamicMaps do not store copies of the underlying data, which reduces the utility of pickling a DynamicMap.\n",
```
philippjfr: The point this is making holds but technically it does hold copies of the data in the cache.

philippjfr:
The changes all look good, except for one minor comment. Happy to merge if we decide on a name for sampled mode.

jlstevens:
I agree "sampling" mode is better than "unbounded mode".

It is a little hard to convey as a concept: it is when you need to specify the keys yourself (or have another HoloMap do it) to generate an element.

jbednar:
Sampling mode makes sense, as long as sampling is the _only_ thing you can do in it.  E.g. if you combine a sampling-mode DynamicMap with a bounded-mode DynamicMap, will it then act as if it's bounded (rather than sampled)?  If so sampling isn't a good word, and unbounded is better.

jlstevens:
I don't know if there is a good word that means 'unspecified' or 'uninstantiated'. My problem with 'unbounded' is that is sounds like it is about the ranges which may in fact be bounded by a limited ranges. Essentially it is a ``DynamicMap`` that can't visualize itself unless you request something from it...

Your comment just appeared and I think that is how it does behave so maybe 'sampling' is right after all! :-)

jbednar:
>  My problem with 'unbounded' is that is sounds like it is about the ranges which may in fact be bounded by a limited ranges. Essentially it is a DynamicMap that can't visualize itself unless you request something from it...

Obviously I'm still confused -- I thought that if the ranges are bounded, then it *can* be visualized, because that lets HoloViews provide sliders.

doc/Tutorials/Dynamic_Map.ipynb
jbednar: Is it holding the data, or the Elements?  It sounds like your statement applies both to the original and the changed text?

philippjfr:
>Obviously I'm still confused -- I thought that if the ranges are bounded, then it can be visualized, because that lets HoloViews provide sliders.

No you're correct.

>if you make a Layout of a sampling-mode DynamicMap with a bounded-mode DynamicMap, will the "sampling-mode" map then act as if it's bounded (rather than sampled)

The sampling is provided by whatever you lay it out with, but I see that's a difficult concept to convey. I'd be okay with either.

doc/Tutorials/Dynamic_Map.ipynb
philippjfr: Yes, it applies to the original text too. It holds the Elements, which themselves hold the data. The cache holds up to 500 Elements.

jbednar: Ok, I'll clarify that; sounds like it is doing what I had expected.

jbednar:
> I believe this is true:
> >  if you combine a sampling-mode DynamicMap with a bounded-mode DynamicMap, will it then act as if it's bounded
>
> In which case isn't 'sampling' the better name?

There still seems to be confusion, but I'm not sure where.  To me, "sampling" means selecting a specific set of locations in a multidimensional space, which is what a HoloMap does.  "bounded" means that we have delineated a region of that space, but have *not* (yet) taken a specific set of locations.  So, if we have a HoloMap `hmap` and a Dynamic_Map `dmap` in what's currently called "sampled mode" and do `dmap + hmap`, dmap effectively gets sampled.  But if we do `dmap + dmap2`, where `dmap2` is a "bounded" Dynamic_Map, then what I hope would happen is that the combination will have sliders covering the ranges specified in dmap2.  At that point, `dmap` will get sampled, it's true, based on the instantaneous values of the sliders, but it's no more sampled than `dmap2` in that sense, so calling `dmap` "sampling mode" doesn't make sense to me if `dmap2` isn't also considered to be in sampling mode (in addition to being in bounded mode).  Is my confusion at least clearly articulated now? :-)

jlstevens:
I think I accidentally contributed to the confusion:

> if you combine a sampling-mode DynamicMap with a bounded-mode DynamicMap, will it then act as if it's bounded

I read that as:

> if you combine a sampling-mode DynamicMap with a bounded-mode *HoloMap*, will it then act as if it's bounded

The latter case is the way I think of 'sampled' mode. It can derive its samples from the *HoloMap* keys.

Hope that helps clear things up a bit. I still don't think this property has anything to do with bounds...

philippjfr:
> It can derive its samples from the HoloMap keys.

It can also derive it from a bounded DynamicMap, which is what @jbednar is getting at. The difference is that it doesn't define a range or values of it's own so I'd be okay with unbounded.

jlstevens:
I'm not sure. I agree.

Unbounded is easily confused with 'open' mode which is also unbounded! But I think what you said gets to the core of it:

> The difference is that it doesn't define a range or values of it's own ...

It is *unspecified*.

jbednar:
I don't think `unspecified` would convey anything.  

Could we perhaps say that `open` and `counter` modes are special cases of `unbounded` mode?  I.e., a DynamicMap is always either in `bounded` or `unbounded` mode, and that there are three types of `unbounded` mode (fully unbounded (not visualizable), `open` mode (values from a generator), or `counter` mode (values generated based on a shared counter)?

philippjfr:
That seems backwards to me. The main distinction is between a mode where you can request discrete samples in a bounded or unbounded and continuous or discrete space, and generator mode where you request the next sample, which is either indexed by a count or returns a key that specifies where in a space it is.

jlstevens:
Maybe another way I think of it is as a 'request' mode - it needs to have a request before it does anything. The requests could be bounded, from a finite set...e.g a HoloMap or you can request elements forever. It isn't about the range, it is that it doesn't know enough to visualize itself. In a way it is a 'naked' ``DynamicMap`` you can use in any bounded or unbounded range you like.

jbednar:
Maybe "mode" is just confusing things here?  Mode seems like something that can be set explicitly to change behavior, whereas what you are discussing above sounds more like predicates that are either true or not true of a given object at a given point.

philippjfr:
Any renaming here would be backwards compatibility breaking and require significant changes to the code. So maybe we should move the mode renaming discussion to a separate issue and just decide on a way to make it clearer in the tutorial in the meantime.

philippjfr:
>Maybe "mode" is just confusing things here? Mode seems like something that can be set explicitly to change behavior, whereas what you are discussing above sounds more like predicates that are either true or not true of a given object at a given point.

That's true as well but I think grouping them conceptually into just bounded and unbounded is confusing for that reason.

jbednar:
With predicates, we can say whether various things are true about an object without having to make just one grouping.  I.e., we can say that a given Dynamic Map is:

- bounded or unbounded:  It's unbounded if it has at least one range with at least one end not specified.
- using data from a function of parameter values, or from a generator
- if using a generator, whether the generator takes a counter argument or not

jlstevens:
Everything you say there is fine and I agree. My problem with 'unbounded' is that it will get really confusing when thinking of 'open' mode. I think the bullet point missing from your list is this:

* Self-visualized or not: Some DynamicMaps need the particular sample value they are shown to be requested from outside, or specified by another HoloMap/DynamicMap.

This to me is orthogonal to what you wrote...except you can't use a generator as you need to ask for an element for a particular sample position (i.e arguments to a callable).

philippjfr:
True but my point was that bounded and unbounded is not a useful conceptual grouping. You can have a bounded sampling map, and a bounded generator map and they will behave very differently. In the latter case all it'll do is check that the returned keys are within the specified ranges, whether it's fully bounded or not therefore doesn't really matter.

jlstevens:
Maybe we can merge this PR now and defer this discussion to another PR? It would be nice to get all of @jbednar's other improvements into the notebooks bundled with this release.

jbednar:
Sure, merge away.

jlstevens:
Ok. Though please submit an issue linking to this PR! Then we can continue the discussion there.

PR #678, opened 2016-05-11
Fix Bokeh QuadMesh example

philippjfr:
I've updated the index of the Bokeh Elements tutorial to indicate it's supported. Additionally I'm now raising an SkipRendering error for non-rectangular QuadMeshes which are not yet supported.


philippjfr:
Ready to merge.

PR #679, opened 2016-05-11
Fixed all flake8 warnings

jlstevens:
Title says it all. Let's try not to introduce any flakes before the release!


philippjfr:
Looks like that caught a few bugs. Tests passed, merging.

PR #680, opened 2016-05-11
Removed BackendError and replaced single usage with SkipRendering

philippjfr:


jlstevens:
This is exactly the change I was expecting. I'm tempted to merge right away but I suppose I should wait for the tests...

jlstevens:
I'm going to take a risk and merge it now - I'm pretty certain it is fine. I'll make sure it goes green on master though...

PR #683, opened 2016-05-11
Fixed initial frame Renderer bounding box calculation

philippjfr:
Very annoying the bounding box calculation has been broken for the initial frame for a while. I assume this will change some test data as well. The regression occurred in this commit: https://github.com/ioam/holoviews/commit/637fe63a475943e03be57a41db6fc62d6de394e1


philippjfr:
PR build passed, ready to merge.

PR #684, opened 2016-05-11
Fixes for dependent widget values

philippjfr:
When adding some of the improvements for dynamic dropdowns and widget labels I didn't correctly update code to handle the dependent widget values where the state of one widget determines the state of next widget. These are some small fixes to make sure that works again.


philippjfr:
Ready to merge.

PR #685, opened 2016-05-12
Clarified homepage examples

jbednar:
This is a bit speculative, but I've noticed that people have a hard time parsing the second example on the homepage -- the result looks good, but it's hard to see how the bits add up to it.  I've reorganized the code slightly and changed the explanation to match; what do you think?


jbednar:
I can't see why the push build would have failed...

doc/Homepage.ipynb
```diff
@@ -4,7 +4,7 @@
...
    ]
```
philippjfr: Hmm rereading the original, I'm wondering about this:

```
HoloViews instead lets you store your data in an annotated format that is instantly visualizable, with immediate access to both the numeric data *and* its visualization.
```

After reading that first sentence I've now gotten the impression that HoloViews is a way of storing your data that makes it instantly visualizable, which is true but a fairly narrow perspective and might turn some people away. In our academic work we often store HoloViews objects and it's a very useful and reproducible way to store a plot but in our other projects we've seen that ingesting external data sources like pandas and iris (and eventually xarray) using the data interfaces is a large part of how holoviews is used in practice. So I think we should get that across in this paragraph somehow.

philippjfr:
The edits look good. It's a pretty complex thing to convey especially because some of the things it's doing aren't particularly meaningful and serve just to demonstrate freedom to express complex relationships.

philippjfr:
Push build failure was a transient. They're very frequent again since I reintroduced something I accidentally deleted before but I know where they're coming from now, the matplotlib tight bounding box code on the MPLRenderer seems to fail sometimes on travis.

jbednar:
>  It's a pretty complex thing to convey especially because some of the things it's doing aren't particularly meaningful

I had to be very careful in how I worded it, because I didn't want to try to explain that the dictionary of Layout objects becomes a HoloMap of Layout objects and then is collated into a Layout of HoloMap objects, which will just make peoples' heads explode.  So I focused on the fact that it was a collection of items at different y values, that we then get a slider for. :-)

doc/Homepage.ipynb
jbednar: I agree that that aspect isn't coming through right now, but it's difficult to see how to make it clear.  People have a hard enough time understanding even the basic idea of how HoloViews works and how it's different from standard plotting APIs, and it's hard to see how one could convey that while also trying to convey that it can also be used to make visualizations of bits of some larger external data structure.  Strictly speaking, I think it does cover the Pandas and xarray cases already as written, in the sense that those objects remain accessible as .data even after wrapping, right?  It's just that it's a different workflow, where only part of the object is exposed as a visualization at a time, rather than having the entire collection of data each always individually and collectively visualizable.

If you can think of how to say it, maybe try editing it and pushing to this branch?

jlstevens: I'm a bit torn on this issue. On one hand I know for a fact that HoloViews offers a very powerful way of working that is hard to put into words. Trying to capture all HoloViews offers in an introductory sentence is really hard...which is why I might consider making the obvious sell first and maybe hinting at/linking to some other material that expands more on how HoloViews is different from other libraries.

philippjfr:
Ready to merge?

jbednar:
Sure; I thought it had been merged ages ago. :-)

PR #700, opened 2016-05-31
Add missing import to Element init

mrksr:
The import is needed in line 44 of the same file.


philippjfr:
Thanks @mrksr. There's something wrong with our Travis setup atm, which I'll have to fix next week. This PR is safe though, so I'll merge.

PR #709, opened 2016-06-04
Iris interface handles standard constructors enabling unit tests

philippjfr:
This PR allows the iris interface to handle the standard tuple and dict based constructors making it equivalent to the grid interface. This means all the unit tests, except for aggregation and sampling (which are not yet implemented), can be shared across the two interfaces.

### DIFF
--- a/holoviews/core/data/iris.py
+++ b/holoviews/core/data/iris.py
@@ -265,8 +265,9 @@ class CubeInterface(GridInterface):
         """
         constraint = cls.select_to_constraint(selection)
         pre_dim_coords = [c.name() for c in dataset.data.dim_coords]
+        indexed = cls.indexed(dataset, selection)
         extracted = dataset.data.extract(constraint)
-        if not extracted.dim_coords:
+        if indexed and not extracted.dim_coords:
             return extracted.data.item()
         post_dim_coords = [c.name() for c in extracted.dim_coords]
         dropped = [c for c in pre_dim_coords if c not in post_dim_coords]
--- a/holoviews/core/data/iris.py
+++ b/holoviews/core/data/iris.py
@@ -75,16 +75,18 @@ class CubeInterface(GridInterface):
             kdim_names = [kd.name for kd in eltype.kdims]
 
         if not isinstance(data, iris.cube.Cube):
+            ndims = len(kdim_names)
+            kdims = [kd if isinstance(kd, Dimension) else Dimension(kd)
+                     for kd in kdims]
+            vdim = vdims[0].name if isinstance(vdims[0], Dimension) else vdims[0]
             if isinstance(data, tuple):
-                coords = [iris.coords.DimCoord(vals, long_name=kd)
-                          for kd, vals in zip(kdim_names, data)]
                 value_array = data[-1]
-                vdim = vdims[0].name if isinstance(vdims[0], Dimension) else vdims[0]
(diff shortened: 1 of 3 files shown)


philippjfr:
Fixes various inconsistencies and adds better coverage. Ready to merge.

tests/testdataset.py
```diff
@@ -480,3 +480,35 @@ def test_dataset_sort_vdim_hm(self):
...
+    """
+
+    def setUp(self):
+        import iris
+        self.restore_datatype = Dataset.datatype
+        Dataset.datatype = ['cube']
+        self.data_instance_type = iris.cube.Cube
+        self.init_data()
+
+    def test_dataset_add_dimensions_values_hm(self):
```
jlstevens: I assume you are doing this to disable some of the inherited tests. A docstring explaining why these tests are not enable would be good...

tests/testirisinterface.py
```diff
@@ -70,7 +70,7 @@ def test_select_index(self):
 
     def test_select_slice(self):
         cube = Dataset(self.cube)
-        self.assertEqual(cube.select(longitude=(0, 1)).data.data,
+        self.assertEqual(cube.select(longitude=(0, 1.01)).data.data,
```
jlstevens: Is the extra .01 due to inclusive vs exclusive bounds?

jlstevens: If so, it might be worth defining ``self.epsilon`` and using that.

tests/testdataset.py
philippjfr: Yes, sounds good. They should eventually be remove once those methods are implemented.

holoviews/core/data/iris.py
```diff
@@ -232,7 +253,7 @@ def select_to_constraint(cls, selection):
             if isinstance(constraint, slice):
                 constraint = (constraint.start, constraint.stop)
             if isinstance(constraint, tuple):
-                constraint = iris.util.between(*constraint)
+                constraint = iris.util.between(*constraint, rh_inclusive=False)
```
jlstevens: Ah, this must be related to the bounds issue I noticed in the tests (see below).

philippjfr: Yes, so far we've used non-inclusive upper bounds in holoviews, so this just makes it consistent. Might be worth having the discussion which is more appropriate, since iris and xarray both seem to use inclusive bounds.

jlstevens:
Looks good and good to support those constructors. My only worry is won't changing the way it does inclusive/exclusive bounds be an issue for backwards compatibility? Perhaps I'm missing something...

philippjfr:
>My only worry is won't changing the way it does inclusive/exclusive bounds be an issue for backwards compatibility?

Backwards compatibility for the Iris backend? I guess so, but it was a bug since it was inconsistent with the other interfaces.

jlstevens:
Ok, as long as that is clear. I agree consistency is more important, especially as this interface is new.

I'm happy to merge now, although maybe you might like to implement my ``self.epsilon`` suggestion? Defining something called ``epsilon`` helps make it clear that all you want is a small increment to include something in the bounds that would otherwise be excluded.

philippjfr:
I'll do that quickly in addition to adding a comments about the disabled tests.

philippjfr:
Done. Ready to merge.

jlstevens:
Looks good. Merging.

PR #712, opened 2016-06-06
Skip rendering when no matching plotting class is found

philippjfr:
Rather than raising an error when a nested object containing objects which cannot be renderered is displayed (e.g. a HoloMap of ``Datasets``) rather than raising a useless error message this will simply raise a ``SkipRendering`` error, which just raises a warning and then prints the repr, which is more useful.


philippjfr:
Ready to merge.

jlstevens:
Agreed this makes sense. Merging.

Issue #714, opened 2016-06-07
Provide convenient way to override dimensions


philippjfr:
One of the major recurring issues we are seeing is that it is fairly difficult to change any options which have been set on a Dimension object, including ranges, formatters, units etc. One easy solution to that would be to provide a convenience method on all Dimensioned objects, which traverses the object and replaces the dimensions, much like the ``relabel`` method allows setting the label and group on a nested object.

Deciding on the name and appropriate signature is the major issue here but we should also decide whether it should allow renaming dimensions, which I believe would be very useful too.

Ignoring the name of the method for now (I'll use redimension but I'm sure we can come up with something better), here are some signatures I can see being supported:

Provide a dictionary mapping from the old name to the new Dimension settings:

```python
# As a dictionary
obj.redimension({'x': dict(name='Time', unit='s', range=(0, 100))})

# As a dimension object
obj.redimension({'x': hv.Dimension(name='Time', unit='s')})
```

Alternatively we can do it directly as kwargs:

```python
# As a dictionary
obj.redimension(x=dict(name='Time', unit='s', range=(0, 100)))

# As a dimension object
obj.redimension(x=hv.Dimension(name='Time', unit='s'))
```

Passing a dictionary would override individual attributes, while passing a dimension would completely override the previous dimension. It could also have a depth argument (just like relabel), but I think that should be set to be infinite by default.

jlstevens:
Sounds good, and I like the keyword syntax.

I know ``redim`` is not a word but then neither is ``redimension`` really. I think ``redim`` as analogous to ``relabel`` is nice and short. Why we are at it, anything else ``X`` you might want to change on a HoloViews object that makes sense to have a ``reX`` method?

philippjfr:
I think ``relabel`` and ``redim`` are the main ones, since those are the two main things that make up a Dimensioned object.

philippjfr:
Just a quick note that implementing this is all very trivial except for one aspect, if we want to allow renaming of dimensions, which I'd argue for strongly, we'll have to provide a way for the data interfaces to change the name of a dimension in the underlying datastructure.

jbednar:
I vote for redim.  Changing the name of a dimension also sounds important.  The kw version seems far more convenient, though it presumably only works if the name can be a Python attribute name, which I guess must be true in this case?

> I know redim is not a word but then neither is redimension really. 

Well... [redimension, v. t.](https://en.wiktionary.org/wiki/redimension)

[...]

PR #715, opened 2016-06-08
Implemented redim method on Dimensioned objects

philippjfr:
This PR adds the ``redim`` method to easily replace dimensions or certain attributes on dimensions in a nested object (following the proposal in #714). The dimension overrides are supplied as keyword arguments to the method with the keyword matching the existing dimension by name and the value supplying one of three overrides:

1. A simple string to rename the dimension, e.g. ``curve.redim(x='Time')``.
2. A dictionary of parameter settings, e.g. ``curve.redim(x={'range': (0, 10), 'name': 'Time'})``
3. A new Dimension object, e.g. ``curve.redim(x=hv.Dimension('Time'))``.

The objects the redim applies to can be narrowed down with a specification, which is already in use on the ``traverse``, ``map`` and ``select`` method. This allows only applying the ``redim`` to certain objects in  three different ways:

1. The type of the object, e.g. ``layout.redim(hv.Curve, x='Time')``
2. A string specification of the form type{{.group}.label}, e.g. ``layout.redim('Curve.GDP', x='Time')``
3. A function, e.g. ``layout.redim(lambda x: x.vdims[0].name == 'GDP', x='Time')``

The implementation exists in two places, on ``Dimensioned`` and on ``Dataset``. The implementation on Dataset is required because a change to the dimension names has to be reflected in the underlying datasource. The ``rename`` implementation on all the interfaces so far is pretty trivial and I think it's good functionality to have anyway. Overall I'm pretty happy with this and will be very useful to have, as now it'll be possible to quickly rename dimensions and adjust their ranges.

I still need to work on the docstrings, docs and add unit tests.

### DIFF
--- a/holoviews/core/data/pandas.py
+++ b/holoviews/core/data/pandas.py
@@ -152,8 +152,9 @@ class PandasInterface(Interface):
 
 
     @classmethod
-    def rename(cls, dataset, renames):
-        return dataset.data.rename(columns=renames)
+    def redim(cls, dataset, dimensions):
+        column_renames = {k: v.name for k, v in dimensions.items()}
+        return dataset.data.rename(columns=column_renames)
 
 
     @classmethod
--- a/holoviews/core/data/pandas.py
+++ b/holoviews/core/data/pandas.py
@@ -152,6 +152,11 @@ class PandasInterface(Interface):
 
 
     @classmethod
+    def rename(cls, dataset, renames):
+        return dataset.data.rename(columns=renames)
+
+
+    @classmethod
     def sort(cls, columns, by=[]):
         import pandas as pd
         if not isinstance(by, list): by = [by]
@@ -210,7 +215,7 @@ class PandasInterface(Interface):
     @classmethod
     def dframe(cls, columns, dimensions):
         if dimensions:
-            return columns.reindex(columns=dimensions)
+            return columns.reindex(dimensions).data.copy()
         else:
             return columns.data.copy()
 
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -394,9 +394,8 @@ class Dataset(Element):
         """
         Replace dimensions on the dataset and allows renaming
         dimensions in the dataset. Dimension mapping should map
-        between the old dimension name and either a dictionary of the
-        new attributes or a completely new dimension to replace it
-        with.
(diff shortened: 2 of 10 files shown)


holoviews/core/data/pandas.py
```diff
@@ -210,7 +215,7 @@ def add_dimension(cls, columns, dimension, dim_pos, values, vdim):
     @classmethod
     def dframe(cls, columns, dimensions):
         if dimensions:
-            return columns.reindex(columns=dimensions)
```
philippjfr: This change snuck in and is unrelated, but I've been meaning to commit it for a while anyway.

philippjfr:
I've now added unit tests, so the PR is ready for review.

jbednar:
Looks good to me!

jlstevens:
Ok, it makes sense. It definitely has to be on ``Dimensioned`` and the reason you need some implementation on the interfaces is also clear.

My only question is why is it called ``rename`` on the interfaces too? Why not also call it ``redim``? Maybe I'm missing something but I think it might be more consistent to have one name to consider denoting one concept. Even if the signature isn't identical etc, to me ``redim`` makes sense on an interface and well as on a dimensioned object.

Other than that one comment, I like this PR and am happy to merge.

philippjfr:
>My only question is why is it called rename on the interfaces too? Why not also call it redim? Maybe I'm missing something but I think it might be more consistent to have one name to consider denoting one concept. Even if the signature isn't identical etc, to me redim makes sense on an interface and well as on a dimensioned object.

I think I agree with you and may even say that the signatures should be the same. Some interfaces like iris also supplying units so those should also be updated. I'll go ahead and make that change.

philippjfr:
Okay, my latest commit has renamed the interface methods from ``rename`` to ``redim`` for consistency. I think this is ready to merge now.

jlstevens:
Ah, the push build passes but the pr one didn't? Is it a transient error?

philippjfr:
Looks like it. Restarting.

philippjfr:
So this makes no sense, one of the new unit tests is failing but only the PR build. Both the push build and my local copy work just fine.

philippjfr:
Finally passing, turned out to be my fault. Part of the tests in this PR depended on the other iris PR.

jlstevens:
Ok. Looks good! Merging.

Issue #705, opened 2016-06-03
Current state of DynamicMap


vascotenner:
Is there an easy way to get the parameters of the showed item in a DynamicMap?

dmap.keys() of dmap.cache shows all the keys available in the cache, not the current selected frame.

When using ipython.widgets.interact, one can do: object.kwargs or something.

This addition would it also make it possible to easilier to save the last state of a widget in a notebook.

jlstevens:
I think this is a reasonable suggestion. 

The attribute could be called ``args`` as that is how arguments are passed into the callback (if not a generator), following the key dimension order. We have some plans to use ``kwargs`` for something else in future...

jlstevens:
I've made a PR (linked above) to add an ``args`` attribute. I'll wait for @philippjfr  to review it but first I better test it works as expected...

jlstevens:
It seems to work fine. I noticed one instance where the ``args`` didn't match the state of the sliders (a bit disturbing!) but I can't seem to reproduce that issue now...

jlstevens:
@vascotenner This should now be addressed by PR #1453 which has been merged.

You'll find an ``args`` and ``kwargs`` attribute on ``dmap.callback`` with the information you need. Closing.

PR #719, opened 2016-06-13
Made last set of positional arguments available in DynamicMap

jlstevens:
This PR is intended to address issue #705 

@philippjfr It is a simple change but it would be good if you could review it anyways.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -456,6 +456,7 @@ class DynamicMap(HoloMap):
     def __init__(self, callback, initial_items=None, **params):
         super(DynamicMap, self).__init__(initial_items, callback=callback, **params)
         self.counter = 0
+        self.args = None # Last set of args passed to callback
         if self.callback is None:
             raise Exception("A suitable callback must be "
                             "declared to create a DynamicMap")
@@ -541,6 +542,7 @@ class DynamicMap(HoloMap):
         if self.call_mode == 'generator':
             retval = next(self.callback)
         else:
+            self.args = args
             retval = self.callback(*args)
 
         if self.call_mode=='key':
@@ -601,6 +603,7 @@ class DynamicMap(HoloMap):
             product = itertools.product(*args)
 
         data = []
+        args = self.args
         for inner_key in product:
             key = util.wrap_tuple(inner_key)
             if key in cache:
@@ -608,6 +611,7 @@ class DynamicMap(HoloMap):
             else:
                 val = self._execute_callback(*key)
             data.append((key, val))
+        self.args = args
         return self.clone(data)
 
 


philippjfr:
Not sure I'm entirely happy with the name ``args``, it really doesn't tell me anything, ``last_args`` might be okay but really it's the last generated key, calling it args seems confusing. Maybe it would make sense to change the semantics of ``last`` and ``last_key`` on DynamicMaps to point to the most recently generated objects rather than the last one in the sorting order?

jlstevens:
> Maybe it would make sense to change the semantics of last and last_key on DynamicMaps to point to the most recently generated objects rather than the last one in the sorting order?

I would prefer to keep ``DynamicMap`` as similar to ``HoloMap`` as possible. I'm happy to call it ``last_args`` if you prefer. 

The reason I think ``args`` is the correct name is that ``args`` and ``kwargs`` are the standard names when considering callables. Talking about 'keys' in this case seems like it would be confusing to me...

philippjfr:
>The reason I think args is the correct name is that args and kwargs are the standard names when considering callables. Talking about 'keys' in this case seems like it would be confusing to me...

Don't really agree, our ``NdMapping`` based datastructures are about keys and adding another name for what is really just the last generated tuple key makes it seem like a separate concept when it really isn't. The DynamicMap is asked for a key and that key is passed as the args to the callback, so the args is a concept the callback worries about but at the DynamicMap level we should be talking about keys.

jlstevens:
> ... at the DynamicMap level we should be talking about keys.

If you like, but I don't think should change the semantics of ``last`` or ``last_key``. Do you have another suggestion for a name?

philippjfr:
>If you like, but I don't think we should change the semantics of last or last_key.

Yea, you're right, probably not a good idea. I'll think about it.

jbednar:
`most_recent_key`?  `current_key`?

jlstevens:
``last_executed_key``?

jlstevens:
I think that streams make this a bit more complicated as there are now ``*args`` (anything in kdims, whether streams or not) or ``**kwargs`` which are streams not in the kdims. Only the *args part could be considered a key...

philippjfr:
Should we close this? With addition of Callables this is probably quite out of date. I wouldn't mind if we made the last key and stream values available on the ``Callable`` though.

jlstevens:
> I wouldn't mind if we made the last key and stream values available on the Callable though.

That would be reasonable and yes, I agree this original approach is now out-of-date. I don't think the Callable should know about keys versus streams but the arguments for the last call should be made available somehow.

philippjfr:
Opened an issue for the suggestion and closing this PR as outdated.

PR #721, opened 2016-06-13
Added start of a new Collector tutorial

jlstevens:
Title says it all! Note this is only a start to a proper tutorial...

I quite like the example I came up with... it shows a real time example by tracking the International Space Station :-)

I needed this tutorial to test my other PR, showing that you can interrupt the collector without losing data (#722)

### DIFF
--- a/doc/Tutorials/Collector.ipynb
+++ b/doc/Tutorials/Collector.ipynb
@@ -1,0 +1,141 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "import urllib2, json, time, datetime\n",
+    "import numpy as np\n",
+    "import param\n",
+    "\n",
+    "import holoviews as hv\n",
+    "from holoviews.interface import Collector\n",
+    "hv.notebook_extension()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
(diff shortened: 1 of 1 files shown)


philippjfr:
Not sure why this was held up for so long? Do we want to keep ``Collector``? Should this be merged?

jbednar:
Collector does seem very useful to me, unless you guys have a replacement for it?

jlstevens:
I have mixed feelings about these classes. 

They are useful but 1. quite complicated 2. can be difficult to understand 3. using NdElement which we want to deprecate. I almost feel there ought to be a separate project for advanced holoviews tools but there isn't quite enough material to justify that yet. 

If nothing else, I feel we ought to have another look at these utilities with as fresh an eye as possible to re-evaluate what we think of them.

philippjfr:
Agree on all counts, I already opened an issue about deprecating NdElement. It marks a dark part of holoviews history when we didn't know how to deal with mixed types, but unfortunately is entangled in both Collator and Collector. Collator we need as a core part of HoloViews and should be reimplemented but I'm not sure what do about Collector. It is still the most convenient way to work with topographica, in a way that would be fairly hard to replace.

philippjfr:
Actually now that I look at it Collector is independent of NdElement, it's just when collating multiple hvz files that you need it in form of a Collator.

philippjfr:
I think we decided a while ago that we'd (re)move Collector. It can either go into topographica or we can write a replacement for it based on ``DynamicMap``. I'd therefore suggest we close this.

philippjfr:
Collector now lives in featuremapper so if appropriate this should move there.

Issue #658, opened 2016-05-07
Handle KeyboardInterrupt in Collector


jlstevens:
The ``Collector`` is a useful (although currently undocumented!) feature. One problem is that interrupting the kernel loses all your data, even if a lot of data has been collected.

For instance, if you run ``data = Collector()(times=[0,500,1000,5000,10000])`` and interrupt after 5000, you get no data back at all. Using the following pattern, it would be possible to return whatever data has been collected before the ``KeyboardInterrupt``.

```python
import time, sys

def loop():
    x = 1
    try:
        for i in range(20):
            print x
            time.sleep(.3)
            x += 1
        return x
    except KeyboardInterrupt:
        return x
    
data = loop()
```

jbednar:
That seems like an excellent idea, and will save some heartache.

jlstevens:
I'm going to implement this now although I don't know who will be using ``Collector`` other than Philipp and myself (now that we have submitted our theses!).

If anyone else is going to make use of it, it will need to be documented.

jlstevens:
Implemented in the PR mentioned above. I'll close this issue as soon as that one is merged.

philippjfr:
Has been merged, closing.

PR #722, opened 2016-06-13
Interrupting a Collector now returns all available data

jlstevens:
Title says it all: ``KeyboardInterrupt`` no longer loses the available data. Shame I didn't implement this before I actually needed it!

Addresses #658

### DIFF
--- a/holoviews/interface/collector.py
+++ b/holoviews/interface/collector.py
@@ -559,7 +559,7 @@ class Collector(AttrTree):
                     if update_progress:
                         interval_hook.percent_range = (completion[i],
                                                        completion[i+1])
-                #interval_hook(0)
+                interval_hook(0)
 
             (self.fixed, attrtree.fixed) = (True, True)
             return attrtree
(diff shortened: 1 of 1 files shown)


jlstevens:
I will note that the last commit uncomments a line ``interval_hook(0)`` which is something I think is a bit of a mistake. The problem is that I can't remember if it is really needed or just something we added to get progress bars working...

Without this line commented, the ``advance_time`` hook alternates between actually advancing time with pointless time advances of zero in between. This is bad when you have side effects such as printing (as in the tutorial).

philippjfr:
Looks fine, going to restart the push build and then merge.

jlstevens:
Ok all the tests have passed now. Ready to merge.

jlstevens:
Though maybe we should discuss the line ``interval_hook(0)``? I think it shouldn't be there but maybe that is an issue for a different PR?

philippjfr:
>I think it shouldn't be there but maybe that is an issue for a different PR?

Yes sounds like a separate PR issue, perhaps either reverting my changes or completely rehauling the progress bar.

philippjfr:
Merging.

PR #724, opened 2016-06-17
Improvements to bokeh callbacks

philippjfr:
This PR now includes various improvements to the bokeh callbacks system.

1. Instead of sending the data to the browser via the regular display machinery it now hooks into the WebSocket Comms that is also used by our widgets. This is cleaner and avoids having to handle the JSON string manually in the callback JS code. 

2. Instead of using some arbitrary throttling value it makes use of blocking callbacks until the previous callback has confirmed it has completed. Additionally it maintains a queue to ensure that the last scheduled event actually gets executed. As a final safeguard it also has a timeout to ensure that if anything goes wrong the callback does not stay in a permanently blocked state.

3. Small changes to make DownsampleColumns a bit faster and avoid having to convert a pandas dataframe to a numpy array for downsampling.

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -8,6 +8,8 @@ from .util import compute_static_patch, models_to_json
 
 from bokeh.models import CustomJS, TapTool, ColumnDataSource
 from bokeh.core.json_encoder import serialize_json
+from bokeh.io import _CommsHandle
+from bokeh.util.notebook import get_comms
 
 
 class Callback(param.ParameterizedFunction):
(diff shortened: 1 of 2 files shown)


philippjfr:
@jbednar @jlstevens Ready for review.

jbednar:
This looks great to me.  There are a few cases where I think it should be redrawing less often, but generally it's about right, and it never seems *terribly* wrong (never queuing up tens of redraws), so I think it's ready to merge.

jlstevens:
It all sounds good to me although I never feel too enthusiastic when I see the Javascript code grow in size. :-(

Other than that, the new approach seems like a nice improvement and I'll merge it now.

PR #725, opened 2016-06-20
Fixes for bokeh 0.12 compatibility

philippjfr:
As the title says this PR makes HoloViews compatible with the upcoming bokeh 0.12 release. Since bokeh is going through rapid development cycles and we do not want to accumulate endless compatibility layers we are also dropping compatibility with 0.10 in this PR. The major component of this PR is in changing the way json patches are computed when using the widgets and callbacks, which has now been unified to use the same approach everywhere, bypassing the relatively slow patch diffing that bokeh uses by default and also dropping various hacks, which I've had to add to make things work across 0.10 and 0.11. This PR also does not yet add support for any of the new features in bokeh 0.12, those should follow in future PRs, for now this just makes sure all existing functionality continues to work.

### DIFF
--- a/holoviews/plotting/bokeh/bokehwidgets.js
+++ b/holoviews/plotting/bokeh/bokehwidgets.js
@@ -25,20 +25,8 @@ var BokehMethods = {
 			var data = this.frames[current];
 		}
 		if (data !== undefined) {
-			if (data.root !== undefined) {
-				var doc = Bokeh.index[data.root].model.document;
-			}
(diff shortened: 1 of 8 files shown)


philippjfr:
Ready to merge.

jlstevens:
I'm happy to see  0.10 compatibility dropped and I like seeing Javascript removed. Unifying the approach (and also getting a speed up!) sounds great.

Merging.

PR #728, opened 2016-06-22
Added support for tick formatters in bokeh

philippjfr:
As the title says this adds support for supplying arbitrary Python functions as tick formatters for the axes. A simple example is a formatter that displays seconds as minutes.

```python
def formatter(seconds):
    m, s = divmod(seconds, 60)
    s = ('%d' % s) if s > 9 else ('0%s' % s)
    return "%d:%s" % (m, s)

hv.Curve(np.random.rand(300)).redim(x=dict(value_format=formatter, unit='min'))
```
image

Note that this depends on ``flexx`` being installed because compiling the python function to JS requires pyscript. Currently it will warn if flexx can't be imported or the function cannot be compiled.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -6,7 +6,6 @@ import bokeh.plotting
 from bokeh.models import Range, HoverTool, Renderer
 from bokeh.models.tickers import Ticker, BasicTicker, FixedTicker
 from bokeh.models.widgets import Panel, Tabs
-from distutils.version import LooseVersion
 
 try:
     from bokeh import mpl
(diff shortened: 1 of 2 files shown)


canavandl:
Conversely - it supports writing the function in JS or Coffeescript too (without flexx).

philippjfr:
That's true although I'm not sure how or even if we should expose that in HoloViews. The benefit of supplying a Python function is that it should work for both bokeh and matplotlib and will show up correctly in sliders and titles.

jbednar:
Looks good to me!  Not exposing the JS options sounds appropriate; we don't normally expose anything JS in HoloViews, but Bokeh does it throughout, so it's fine that we make a different choice.  Happy to have this merged.

philippjfr:
Just added a small change so that a single argument function is supported in matplotlib. We now consistently support single argument functions everywhere and you can also supply a function with an **optional** second argument, which will be passed the matplotlib ``pos`` argument. It's worth noting that matplotlib and bokeh will likely never be consistent here because the ``pos`` argument depends on the zoom level and is therefore pretty much useless in bokeh. Ready to review and merge.

jlstevens:
Ok, I reviewed it earlier and it looked fine except for that one issue. Merging.

PR #730, opened 2016-06-27
Handle backend JS/CSS dependencies for static export better

philippjfr:
Automatically adds backend JS/CSS dependencies to ``static_html`` output and handles inlined bokeh dependencies if CDN is not available (e.g. for dev builds of bokeh).

### DIFF
--- a/holoviews/plotting/bokeh/renderer.py
+++ b/holoviews/plotting/bokeh/renderer.py
@@ -38,7 +38,8 @@ class BokehRenderer(Renderer):
     widgets = {'scrubber': BokehScrubberWidget,
                'widgets': BokehSelectionWidget}
 
-    backend_dependencies = {'js': CDN.js_files, 'css': CDN.css_files}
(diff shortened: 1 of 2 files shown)


philippjfr:
Ready to review and merge.

jbednar:
Sounds good!  

What does "CDN is not available" cover?  Does it just cover custom/dev builds where there are no corresponding files on CDN at all, or does it also handle transient unreachability of CDN?  I can't see what the mechanism is from the diff.

jlstevens:
Looks good. My only worry is that``html_assets`` has the potential to get hairier and hairier over time but for now it is still ok.

Merging.

philippjfr:
>What does "CDN is not available" cover? Does it just cover custom/dev builds where there are no corresponding files on CDN at all, or does it also handle transient unreachability of CDN (e.g. working offline)?

Only covers dev builds where there is no CDN version, in that case it just supplies the JS and CSS as raw text. However we could/(should?) also consider provide control over that.

PR #735, opened 2016-06-28
Correctly convert glyph properties before updating them

philippjfr:
Tiny fix to ensure that colors are processed correctly. Most of this processing should be moved out of the ``mpl_to_bokeh`` method into one that is independent of matplotlib soon, but that's a separate PR.


jlstevens:
Looks fine as it is a one line change. Merging.

PR #736, opened 2016-06-28
Skip rendering of nested Layouts

philippjfr:
This ensures that a layout that is nested rather than combined raises SkipRendering exception causing the display of the object to be skipped and a warning to be emitted.


philippjfr:
Will fix the PR in the morning.

jlstevens:
Simple and sensible change. Merging.

PR #737, opened 2016-06-28
Added bokeh model priority order

philippjfr:
This is needed because some models trigger a rerender while others do not, which has the potential to cause a redraw before a certain property has been updated.

### DIFF
--- a/holoviews/plotting/bokeh/util.py
+++ b/holoviews/plotting/bokeh/util.py
@@ -40,6 +40,11 @@ markers = {'s': {'marker': 'square'},
 # and can therefore be safely ignored.
 IGNORED_MODELS = ['LinearAxis']
 
+# Model priority order to ensure some types are updated before others
+MODEL_PRIORITY = ['Range1d', 'Title', 'Image', 'LinearColorMapper',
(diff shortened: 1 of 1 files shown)


holoviews/plotting/bokeh/util.py
```diff
@@ -40,6 +40,11 @@
 # and can therefore be safely ignored.
 IGNORED_MODELS = ['LinearAxis']
 
+# Model priority order to ensure some types are updated before others
+MODEL_PRIORITY = ['Range1d', 'Title', 'Image', 'LinearColorMapper',
+                  'Plot', 'Range1d', 'LinearAxis', 'ColumnDataSource']
```
jlstevens: Looks like a list of things we may need to worry about maintaining over time.

jlstevens:
Seems fine as it is only putting things in priority order. It is another list of things to maintain over time, but I don't see a way to avoid that. Merging.

holoviews/plotting/bokeh/util.py
philippjfr: Yes, although I've not so far encountered any issues apart from the color mapper one, but we may well run into others at some point.

PR #741, opened 2016-06-29
Ensure nan is not set on bokeh axis range

philippjfr:
Straightforward fix which avoids NaN errors when plotting annotations or other plots which do not provide an extent.


jlstevens:
As you say, looks straightforward enough. Merging.

PR #743, opened 2016-06-29
Cast mpl data to string before updating live widget

philippjfr:
Currently a random quote appears in live matplotlib widgets because the data is unicode. Eventually we'll want to replace this with a websocket, so for now I'm just casting back to string. This is safe because all the data is base64 encoded (and it is what we used to do). This bug only affected Python2 because Python 3 does not have the ``u'`` prefix for string literals.


jlstevens:
Fix seems fine although the bug itself is annoying. Ready to merge?

philippjfr:
Just restarted the tests, merge once it passes.

jlstevens:
Tests passed. Merging.

PR #745, opened 2016-06-30
Fix bokeh event sorting in py3

philippjfr:
Fixes python 3 bug where events could not be sorted:

```python
TypeError: unorderable types: dict() < dict().
```

Now sorting on just the priorities.


jbednar:
Works for me; ok to merge?

philippjfr:
Yes, go ahead.

PR #746, opened 2016-07-01
Added function to strip bokeh update events from specific models

philippjfr:
Fixes previous approach to stripping certain models from bokeh update events. Both LinearAxis and LogAxis complain when updated saying that ``computed_bounds`` cannot be changed. Previously I was crudely looking for these objects, as I found out this is not sufficient in some cases. I've now added a function that traverses the model data and makes sure all references to those models are deleted. I'll investigate this bug with bokeh, but in the meantime it doesn't seem to do any harm since we never update the axis name or switch out an axis anyway.


jlstevens:
Ok, the tests are green so I'll merge now.

PR #750, opened 2016-07-01
Disable shared datasources on Layouts by default

philippjfr:
Sharing of datasources can cause issues when it is updated in one plot but not the other. In Grids this will not occur but on a Layout it can, so sharing of datasources should only be enabled when specifically requested.


jlstevens:
Sounds reasonable. Are there any backwards compatibility implications?

philippjfr:
Possibly, we should make a note in the release notes.

jlstevens:
> Doubt very many people were relying on this though.

I agree, which is why I'm happy to merge once the tests go green.

PR #754, opened 2016-07-01
Removed bokeh resize tool until v1.0 release

philippjfr:
The bokeh resize tool is broken since the layout work they did, hopefully it will be fixed in v1.0 but for now let's remove it.


jlstevens:
Looks ready to merge right away.

PR #758, opened 2016-07-01
Added Cycle and Palette to top-level namespace

philippjfr:


jlstevens:
Good idea!

jlstevens:
The pr builds pass so I'll merge this now.

Issue #756, opened 2016-07-01
Unicode and table


vascotenner:
I try to print unicode labels in an ItemTable. I start jupyter from an envirment with `PYTHONIOENCODING='utf8'`:

```
from __future__ import  division #unicode_literals,
%matplotlib inline
import holoviews as hv
from collections import OrderedDict
hv.notebook_extension()

#works
hv.ItemTable(OrderedDict({'b':'α'}))
#broken
hv.ItemTable(OrderedDict({'α':'b'}))
```
Traceback:

```
---------------------------------------------------------------------------
UnicodeEncodeError                        Traceback (most recent call last)
/home/tenner/.local/lib/python2.7/site-packages/IPython/core/formatters.pyc in __call__(self, obj)
    337                 pass
    338             else:
--> 339                 return printer(obj)
    340             # Finally look for special method names
    341             method = _safe_get_formatter_method(obj, self.print_method)

/home/tenner/src/holoviews/holoviews/ipython/display_hooks.pyc in pprint_display(obj)
    214     if not ip.display_formatter.formatters['text/plain'].pprint:
    215         return None
--> 216     return display(obj, raw=True)
    217 
    218 

/home/tenner/src/holoviews/holoviews/ipython/display_hooks.pyc in display(obj, raw, **kwargs)
    196         html = grid_display(obj)
    197     elif isinstance(obj, (CompositeOverlay, ViewableElement)):
--> 198         html = element_display(obj)
    199     elif isinstance(obj, (Layout, NdLayout, AdjointLayout)):
    200         html = layout_display(obj)

/home/tenner/src/holoviews/holoviews/ipython/display_hooks.pyc in wrapped(element)
    101             html = fn(element,
    102                       max_frames=OutputMagic.options['max_frames'],
--> 103                       max_branches = OutputMagic.options['max_branches'])
    104 
    105             # Only want to add to the archive for one display hook...

/home/tenner/src/holoviews/holoviews/ipython/display_hooks.pyc in element_display(element, max_frames, max_branches)
    142         return None
    143     renderer = Store.renderers[backend]
--> 144     return renderer.html(element, fmt=renderer.fig)
    145 
    146 

/home/tenner/src/holoviews/holoviews/plotting/renderer.pyc in html(self, obj, fmt, css)
    247         Renders plot or data structure and wraps the output in HTML.
    248         """
--> 249         plot, fmt =  self._validate(obj, fmt)
    250         figdata, _ = self(plot, fmt)
[...]

PR #760, opened 2016-07-01
Unicode fix for ItemTable

jlstevens:
This PR aims to address the unicode issues raised in #756.

### DIFF
--- a/holoviews/plotting/mpl/tabular.py
+++ b/holoviews/plotting/mpl/tabular.py
@@ -89,7 +89,7 @@ class TablePlot(ElementPlot):
         elif isinstance(value, str):
             formatted = safe_unicode(value)
         else:
-            formatted = str(value)
+            formatted = safe_unicode(value)
 
         if len(formatted) > self.max_value_len:
             return formatted[:(self.max_value_len-3)]+'...'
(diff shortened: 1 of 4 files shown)


holoviews/plotting/mpl/tabular.py
```diff
@@ -89,7 +89,7 @@ def pprint_value(self, value):
         elif isinstance(value, str):
             formatted = safe_unicode(value)
         else:
-            formatted = str(value)
```
philippjfr: ``safe_unicode`` will not cast a non-string type to unicode, so if something downstream of this is expecting a string this won't do it.

jlstevens: Right, this change makes the `` elif isinstance(value, str):`` condition pointless. What would you recommend? Simply using ``str()`` results in unicode errors.

jlstevens: ah...I think I should just use...

```
elif isinstance(value, (str, unicode)):
```

I'll try that.

philippjfr: Hmm, changing the previous block with:

```
elif isinstance(value, basestring):
```

where basestring is import from holoviews.core.util, should do it.

philippjfr:
Looks fine, let's see about tests.

PR #761, opened 2016-07-01
Ignore all axis and ticker models when updating bokeh plots

philippjfr:
Expanded the list of models ignored when sending updates to bokeh plots. This is still because bokeh axes raise errors when their ``computed_bounds`` are updated. Therefore all axis models and now also their tickers and tick formatters have to be excluded. For now this is not an issue, because we don't allow changing anything about those types between frames, but I will still investigate whether this can be solved in bokeh.


jlstevens:
Ok, I'm hoping this can be resolved on the bokeh side of things in future.

PR #762, opened 2016-07-03
Correctly set batched plot options

philippjfr:
Correctly respect the plot options set on the batched elements. Previously inherited all plot options from the NdOverlay, which a) could cause unsupported parameters being set and b) ignored plot options set on the Element. This appropriately combines the plot options from set on the NdOverlay and Element plot.

### DIFF
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -503,10 +503,10 @@ class GenericElementPlot(DimensionedPlot):
                                kdims=['Frame'], id=element.id)
         else:
             self.hmap = element
+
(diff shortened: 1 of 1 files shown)


jlstevens:
Is this ready for review?

jbednar:
It certainly sounds like an improvement, and I'd be in favor of merging.

philippjfr:
Yes, ready to review/merge.

jlstevens:
I notice this is implemented on ``GenericElementPlot`` but the tests have passed, the changes look reasonable so I think it is all ok. Merging.

PR #763, opened 2016-07-04
Fixes for Bokeh path plots

philippjfr:
Fixes for hover tools on polygons and adding batching for Path plots. This PR is based off https://github.com/ioam/holoviews/pull/762 and should be merged after that one.

### DIFF
--- a/holoviews/plotting/bokeh/path.py
+++ b/holoviews/plotting/bokeh/path.py
@@ -8,7 +8,7 @@ from bokeh.models import HoverTool
 from ...core import util
 from ..util import map_colors
 from .element import ElementPlot, line_properties, fill_properties
-from .util import get_cmap
+from .util import get_cmap, rgb2hex
 
 
 class PathPlot(ElementPlot):
--- a/holoviews/plotting/bokeh/path.py
+++ b/holoviews/plotting/bokeh/path.py
@@ -17,13 +17,33 @@ class PathPlot(ElementPlot):
         Whether to show legend for the plot.""")
 
     style_opts = ['color'] + line_properties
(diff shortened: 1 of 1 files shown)


philippjfr:
Just rebased the PR, now ready to merge.

holoviews/plotting/bokeh/path.py
```diff
@@ -17,14 +17,34 @@ class PathPlot(ElementPlot):
...
+            eldata, elmapping = self.get_data(el, ranges, empty)
+            for k, eld in eldata.items():
+                data[k].extend(eld)
+            zorder = self.get_zorder(element, key, el)
+            val = style[zorder].get('color')
+            elmapping['color'] = 'color'
+            if isinstance(val, tuple):
+                val = rgb2hex(val)
+            data['color'] += [val for _ in range(len(eldata['xs']))]
+        if len(set(data.get('color'))) == 1:
```
jlstevens: I'm a little confused by this bit. In the loop, you populate ``data`` with ``'color'`` but here you pop it off again?

jlstevens:
Other than one comment (which is only asking for clarification) this looks fine and I am happy to merge.

holoviews/plotting/bokeh/path.py
philippjfr: Yes, if all the paths are the same color there's no need to set it individually for each path. It'll simply be set once separately from the data.

jlstevens: Ok, thanks for clarifying!

philippjfr: I'll get into the habit of adding more comments for non-obvious things like this.

PR #769, opened 2016-07-08
Fixed extent handling in batched mode

philippjfr:
While introducing batched mode I did not correctly handle computing the extents. This PR consists of two main fixes, a) the GenericOverlayPlot now correctly computes the extents for a batched plot, and b) the bokeh backend now correctly detects framewise normalization modes (via a property) and always makes sure no NaNs are set on the axis ranges (which can cause the plot to blow up).

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -181,10 +181,6 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         # Get the bottom layer and range element
         el = element.traverse(lambda x: x, [Element])
         el = el[0] if el else element
-        if self.batched and not isinstance(self, OverlayPlot):
-            range_el = el
-        else:
-            range_el = element
 
         dims = el.dimensions()
         xlabel, ylabel, zlabel = self._get_axis_labels(dims)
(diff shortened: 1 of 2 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -583,21 +587,23 @@ def current_handles(self):
...
                 handles += [plot.x_range, plot.y_range]
         return handles
 
 
+    @property
+    def framewise(self):
+        """
+        Property to determine whether the current frame should have
+        framewise normalization enabled.
+        """
```
jlstevens: Docstring should say why this is a bokeh specific property - that it is about generating events to update the bokeh plots when it makes sense. Otherwise, having this property makes sense.

holoviews/plotting/bokeh/element.py
```diff
@@ -178,8 +178,14 @@ def _get_hover_data(self, data, element, empty=False):
...
 
     def _axes_props(self, plots, subplots, element, ranges):
+        # Get the bottom layer and range element
         el = element.traverse(lambda x: x, [Element])
         el = el[0] if el else element
+        if self.batched and not isinstance(self, OverlayPlot):
+            range_el = el
+        else:
+            range_el = element
+
```
jlstevens: 1. This could be defined lower down closer to where ``range_el`` is used.
2. As ``range_el`` is only used once, I would consider inlining it:

```python
l, b, r, t = self.get_extents(el if self.batched 
        and not isinstance(self, OverlayPlot) else element, ranges)

philippjfr: May not inline it, but I can move it closer at least.

jlstevens:
Other than my two comments (one of which you have already addressed) this looks fine and I am happy to merge.

philippjfr:
Addressed both comments now, ready to merge whenever.

jlstevens:
Looks good and seems to have fixed the notebooks I was working on. 

As this involves a change to ``get_extents`` which is fairly basic plotting functionality, I think it is worth making sure the tests pass. Once that happens, I will merge.

jlstevens:
Ok great! The tests passed. Merging.

PR #770, opened 2016-07-08
Added 'LogTickFormatter' to the set of ignored Bokeh models

jlstevens:
Simple PR that fixes using ``logy=True`` with Bokeh plots (at least for ``Curve`` plots).


philippjfr:
Thanks. Won't affect tests so I'll just merge.

PR #772, opened 2016-07-09
Small fixes for bokeh OverlayPlot updates

philippjfr:
Handles condition where an Overlay is missing for one or more frames.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -821,28 +821,27 @@ class OverlayPlot(GenericOverlayPlot, ElementPlot):
             self.current_frame = element
             self.current_key = key
 
+        items = element.items() if element else []
         if isinstance(self.hmap, DynamicMap):
             range_obj = element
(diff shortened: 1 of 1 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -821,28 +821,27 @@ def update_frame(self, key, ranges=None, element=None, empty=False):
...
                 empty = idx is None
                 if not empty:
                     _, el = items.pop(idx)
             subplot.update_frame(key, ranges, element=el, empty=(empty or all_empty))
 
-        if isinstance(self.hmap, DynamicMap) and items:
-            raise Exception("Some Elements returned by the dynamic callback "
-                            "were not initialized correctly and could not be "
-                            "rendered.")
+        if isinstance(self.hmap, DynamicMap) and items and element:
```
jlstevens: Do we need ``and element`` here?

jlstevens:
Looks good and the condition makes sense. Merging.

PR #773, opened 2016-07-09
Fixed bokeh shared_datasource setting

philippjfr:
I recently disabled the ``shared_datasource`` option because it can potentially cause problems. However I meant to disable it by default on Layouts only. This sets the behavior I had originally meant to apply.


jlstevens:
Straightforward change that I think is unlikely to cause backwards compatibility issues. Merging.

PR #774, opened 2016-07-09
Fix for single row grid plots in bokeh

philippjfr:
The logic in bokeh GridPlots was wrong, which was causing single dimensional layouts to be the wrong shape and not appropriately hide the axes.


jlstevens:
Good to know that the simpler version is correct! Merging.

PR #775, opened 2016-07-09
Ensure non-None values aren't ignored in bokeh json messages

philippjfr:
The code that deletes references which cannot easily be updated in bokeh has a small bug where values that evaluate as False were deleted, when it really should just have been deleting Nones. This caused events to set the visibility of glyphs to False to be deleted for example.


jlstevens:
Nice simple fix. Merging.

PR #776, opened 2016-07-09
Avoid mutating bokeh data source mapping

philippjfr:
In some cases the mapping returned by the ``get_data`` method on bokeh plots is mutated, which can cause side-effects because it is originally a class attribute. This PR ensures that the class attribute is never mutated.

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -206,7 +206,7 @@ class SpreadPlot(PolygonPlot):
         upper = mean + pos_error
         band_x = np.append(xvals, xvals[::-1])
         band_y = np.append(lower, upper[::-1])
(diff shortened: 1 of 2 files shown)


holoviews/plotting/bokeh/chart.py
```diff
@@ -206,7 +206,7 @@ def get_data(self, element, ranges=None, empty=None):
         upper = mean + pos_error
         band_x = np.append(xvals, xvals[::-1])
         band_y = np.append(lower, upper[::-1])
-        return dict(xs=[band_x], ys=[band_y]), self._mapping
+        return dict(xs=[band_x], ys=[band_y]), dict(elf._mapping)
```
jlstevens: ``elf._mapping``? :-)

philippjfr: Good catch!

jlstevens:
Looks good. Merging.

PR #777, opened 2016-07-09
Handle renderer options size and webgl in bokeh

philippjfr:
Adds support for global scaling of bokeh plots via ``size`` output magic option and setting webgl option based on global renderer.


jlstevens:
There are really three separate changes here but they all look good. Merging.

PR #778, opened 2016-07-11
Fixed batched plotting for bokeh PathPlot

philippjfr:
This PR fixes batched plotting of Path, Spikes and ErrorBars (which all have plot classes which subclass from PathPlot). The previous approach a) did not consider that Spikes and ErrorBars also use PathPlot and b) removes an optimization where it would not send color data if only a single color was set, which caused issues when some frames had multiple colors and some had only a single color.


jlstevens:
Nice to see the simpler version is the one that works correctly. Merging.

PR #779, opened 2016-07-11
Add support for inverting axes and adjoining all Element types

philippjfr:
This PR adds support for almost all Element types to be used as adjoints in bokeh by correctly implementing axis inversions on the plotting classes.


jbednar:
This looks amazing!  Let's merge right away!  I can't tell why the tests are taking so long.

philippjfr:
Opened the issue above to ensure that we add support for inverting the axes on Raster types sometime in the future as well. For now this is ready to merge as it massively improves our existing support for adjoining plots and inverting axes.

jlstevens:
Looks good! It will be good to finish this off so it is supported everywhere but this is already a nice improvement. Merging.

PR #784, opened 2016-07-12
Ensure empty frames don't break plots in bokeh

philippjfr:
Fixes bug where empty frame breaks plot updating.


jlstevens:
Nice simple fix. Merging.

PR #785, opened 2016-07-12
Reenable layout constraint solver workaround in bokeh

philippjfr:
In bokeh 0.12 the bokeh layout constraint solver was finally updated to correctly handle empty parts of a plot. However apparently it does not correctly handle the size of the empty plots. This PR reenables the workaround that explicitly inserts an empty plot.


jlstevens:
Shame the layout engine isn't working as expected. Merging.

PR #786, opened 2016-07-12
Ensure that empty overlay frame does not break

philippjfr:
As the title says, when there's an empty frame in bokeh's OverlayPlot, it should not try to compute and set extents.


jlstevens:
Looks good. Merging.

PR #788, opened 2016-07-14
Add GridImage Element which will eventually replace Image

philippjfr:


jlstevens:
Looks clean and is a nice step towards the improved system that will unify gridded data with rasters. Merging.

PR #792, opened 2016-07-18
Make dependency optional

tinloaf:
The DFrame -> TablePlot association should only be registered if there is a DFrame to be imported...


jlstevens:
Looks good to me!

I won't merge till Philipp gives his feedback...

philippjfr:
Looks good, thanks for the fix!

PR #793, opened 2016-07-18
Do not fail if .info is called on an empty holomap

tinloaf:


jlstevens:
Looks good and thanks for the fix! 

I've tested it and it works fine -  I'll merge it once you confirm that you are done with it.

tinloaf:
I think I'm happy with it. :)

PR #794, opened 2016-07-18
Grid based interface fixes

philippjfr:
Ensure that the flat and non-flat arrays returned by the grid interfaces are consistent and of the correct orientation. Also ensures the constructors work correctly and all process tuples of the coordinates and arrays correctly. Also adds small fixes for the GridImage Element, updates test and fixes issues with dynamic groupby, which caused the grid interface to reject the data in some cases. Addresses https://github.com/ioam/holoviews/issues/789 and more.

Still want to add a number of tests to ensure consistency and stability of the interfaces.

- [x] Standardized interface to get canonical views of coordinates and data arrays
- [x] Fixed constructors ensuring they handle standard tuple and dict based gridded data correctly
- [x] Fixed GridImage density calculation
- [x] Fixes for dynamic groupby with gridded interfaces (and allowing passing of parameters to groups).
- [x] Added unit tests to check output of gridded data ``dimension_values`` is consistent.

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -393,13 +393,15 @@ class Dataset(Element):
 
         if dynamic:
             group_dims = [d.name for d in self.kdims if d not in dimensions]
+            group_kwargs = dict(util.get_param_values(self))
+            group_kwargs.update(kwargs)
             def load_subset(*args):
                 constraint = dict(zip(dim_names, args))
                 group = self.select(**constraint)
                 if np.isscalar(group):
                     return group_type(([group],), group=self.group,
                                       label=self.label, vdims=self.vdims)
-                return group_type(group.reindex(group_dims))
+                return group_type(group.reindex(group_dims), **group_kwargs)
             dynamic_dims = [d(values=list(self.interface.values(self, d.name, False)))
                             for d in dimensions]
             return DynamicMap(load_subset, kdims=dynamic_dims)
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -393,8 +393,8 @@ class Dataset(Element):
 
         if dynamic:
             group_dims = [d.name for d in self.kdims if d not in dimensions]
-            group_kwargs = dict(util.get_param_values(self))
-            group_kwargs.update(kwargs)
+            group_kwargs = dict(util.get_param_values(self), **kwargs)
+            group_kwargs['kdims'] = [self.get_dimension(d) for d in group_dims]
             def load_subset(*args):
                 constraint = dict(zip(dim_names, args))
                 group = self.select(**constraint)
--- a/holoviews/core/data/dictionary.py
+++ b/holoviews/core/data/dictionary.py
@@ -56,7 +56,8 @@ class DictInterface(Interface):
             data = {k: data[:,i] for i,k in enumerate(dimensions)}
         elif isinstance(data, list) and np.isscalar(data[0]):
             data = {dimensions[0]: np.arange(len(data)), dimensions[1]: data}
-        elif not any(isinstance(data, interface) for interface in cls.interfaces):
+        elif not any(isinstance(data, tuple(t for t in interface.types if t is not None))
+                     for interface in cls.interfaces.values()):
             data = {k: v for k, v in zip(dimensions, zip(*data))}
         elif isinstance(data, dict) and not all(d in data for d in dimensions):
             dict_data = zip(*((util.wrap_tuple(k)+util.wrap_tuple(v))
--- a/holoviews/core/data/dictionary.py
+++ b/holoviews/core/data/dictionary.py
@@ -87,7 +87,7 @@ class DictInterface(Interface):
             raise ValueError('Following dimensions not found in data: %s' % not_found)
         lengths = [len(dataset.data[dim]) for dim in dimensions]
         if len({l for l in lengths if l > 1}) > 1:
-            raise ValueError('Length of dataset do not match')
+            raise ValueError('Length of columns do not match')
 
 
     @classmethod
@@ -135,8 +135,8 @@ class DictInterface(Interface):
         cast_objs = cls.cast(dataset_objs)
         cols = set(tuple(c.data.keys()) for c in cast_objs)
         if len(cols) != 1:
-            raise Exception("In order to concatenate, all Column objects "
-                            "should have matching set of dataset.")
+            raise Exception("In order to concatenate, all Dataset objects "
+                            "should have matching set of columns.")
         concatenated = OrderedDict()
         for column in cols.pop():
             concatenated[column] = np.concatenate([obj[column] for obj in cast_objs])
--- a/holoviews/core/data/dictionary.py
+++ b/holoviews/core/data/dictionary.py
@@ -56,6 +56,8 @@ class DictInterface(Interface):
             data = {k: data[:,i] for i,k in enumerate(dimensions)}
         elif isinstance(data, list) and np.isscalar(data[0]):
             data = {dimensions[0]: np.arange(len(data)), dimensions[1]: data}
+        # Ensure that interface does not consume data of other types
+        # with an iterator interface
         elif not any(isinstance(data, tuple(t for t in interface.types if t is not None))
                      for interface in cls.interfaces.values()):
             data = {k: v for k, v in zip(dimensions, zip(*data))}
--- a/holoviews/core/data/dictionary.py
+++ b/holoviews/core/data/dictionary.py
@@ -56,7 +56,7 @@ class DictInterface(Interface):
             data = {k: data[:,i] for i,k in enumerate(dimensions)}
         elif isinstance(data, list) and np.isscalar(data[0]):
             data = {dimensions[0]: np.arange(len(data)), dimensions[1]: data}
-        elif not isinstance(data, dict):
+        elif not any(isinstance(data, interface) for interface in cls.interfaces):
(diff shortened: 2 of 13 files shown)


tests/testirisinterface.py
```diff
@@ -51,10 +51,10 @@ def test_dimension_values_kdim(self):
     def test_dimension_values_vdim(self):
         cube = Dataset(self.cube, kdims=['longitude', 'latitude'])
         self.assertEqual(cube.dimension_values('unknown', flat=False),
-                         np.flipud(np.array([[ 0,  4,  8],
-                                             [ 1,  5,  9],
-                                             [ 2,  6, 10],
-                                             [ 3,  7, 11]], dtype=np.int32).T))
```
jlstevens: As long as you are sure this is definitely correct now... :-)

holoviews/plotting/mpl/raster.py
```diff
@@ -183,7 +183,7 @@ def update_handles(self, key, axis, element, ranges, style):
 class ImagePlot(RasterPlot):
 
     def get_data(self, element, ranges, style):
-        data = element.dimension_values(2, flat=False)
+        data = np.flipud(element.dimension_values(2, flat=False))
```
jlstevens: How come this doesn't flip ``Image`` given ``dimension_values`` for numpy arrays won't be based on interface classes?

holoviews/element/raster.py
```diff
@@ -648,16 +651,17 @@ class GridImage(Dataset, Element2D):
...
+    vdims = param.List(default=[Dimension('z')], bounds=(1, 1))
 
     def __init__(self, data, **params):
         super(GridImage, self).__init__(data, **params)
         (l, r), (b, t) = self.interface.range(self, 0), self.interface.range(self, 1)
         (ys, xs) = self.dimension_values(2, flat=False).shape
-        xsampling = (float(r-l)/xs)/2.
-        ysampling = (float(t-b)/ys)/2.
+        xsampling = (float(r-l)/(xs-1))/2.
+        ysampling = (float(t-b)/(ys-1))/2.
```
jlstevens: There was an off-by-one error in the bounds? How come this wasn't flagged up by the tests...or did you update the test data? If not, why isn't bounds being checked in the data comparisons?

holoviews/element/raster.py
```diff
@@ -648,16 +651,17 @@ class GridImage(Dataset, Element2D):
 
     group = param.String(default='GridImage', constant=True)
 
-    kdims = param.List(default=['x', 'y'], bounds=(2, 2))
+    kdims = param.List(default=[Dimension('x'), Dimension('y')],
+                       bounds=(2, 2))
 
-    vdims = param.List(default=['z'], bounds=(1, 1))
+    vdims = param.List(default=[Dimension('z')], bounds=(1, 1))
```
jlstevens: I thought strings were automatically promoted to ``Dimension`` objects and that we favored strings by default?

holoviews/element/raster.py
```diff
@@ -396,6 +397,8 @@ def __init__(self, data, extents=None, **params):
 
 
     def _compute_raster(self):
+        if issubclass(self.interface, GridInterface):
+            return self, np.flipud(self.dimension_values(2, flat=False))
```
jlstevens: Perhaps all interfaces could have a property to check to see if this return can be done directly instead of using ``isinstance``?

holoviews/core/data/__init__.py
```diff
@@ -393,13 +393,15 @@ def groupby(self, dimensions=[], container_type=HoloMap, group_type=None,
...
+            group_kwargs['kdims'] = [self.get_dimension(d) for d in group_dims]
             def load_subset(*args):
                 constraint = dict(zip(dim_names, args))
                 group = self.select(**constraint)
                 if np.isscalar(group):
                     return group_type(([group],), group=self.group,
                                       label=self.label, vdims=self.vdims)
-                return group_type(group.reindex(group_dims))
+                return group_type(group.reindex(group_dims), **group_kwargs)
             dynamic_dims = [d(values=list(self.interface.values(self, d.name, False)))
```
jlstevens: I think this looks right, but if I am not confused this is a separate fix from the rest of the PR? Looks like the kwargs weren't being passed into the group container when dynamic....

holoviews/element/raster.py
philippjfr: This was on GridImage, and is currently untested.

holoviews/core/data/dictionary.py
```diff
@@ -56,7 +56,8 @@ def init(cls, eltype, data, kdims, vdims):
             data = {k: data[:,i] for i,k in enumerate(dimensions)}
         elif isinstance(data, list) and np.isscalar(data[0]):
             data = {dimensions[0]: np.arange(len(data)), dimensions[1]: data}
-        elif not isinstance(data, dict):
+        elif not any(isinstance(data, tuple(t for t in interface.types if t is not None))
```
jlstevens: I would just add a comment to say this is needed as dictionary can incorrectly accept xarray (duck typing fails here!). The comment might also mention that a cleaner approach would be welcome...

holoviews/element/raster.py
philippjfr: Not as a default.

holoviews/core/data/__init__.py
philippjfr: That caused weird bugs on the interfaces before I fixed the grid interfaces. So it's related at least.

holoviews/core/data/grid.py
```diff
@@ -102,18 +102,79 @@ def length(cls, dataset):
...
+        Returns the coordinates along a dimension.
+        """
+        data = dataset.data[dim]
+        if ordered and np.all(data[1:] < data[:-1]):
+            data = data[::-1]
+        return data
+
+
+    @classmethod
+    def expanded_coords(cls, dataset, dim):
```
jlstevens: Can't this be folded into ``get_coords`` above using an ``expanded=False`` default argument?

holoviews/core/data/__init__.py
jlstevens: I'm happy to have the fix! I just wanted to make sure I understood it...

holoviews/element/raster.py
jlstevens: Ah ok, that makes sense.

holoviews/element/raster.py
jlstevens: Looking at various examples in ``element`` I see you are right. Now I'm wondering how it worked at all before when using the default!

holoviews/core/data/grid.py
```diff
@@ -102,18 +102,79 @@ def length(cls, dataset):
...
+        idx = dataset.get_dimension_index(dim)
+        return util.cartesian_product(arrays)[idx]
+
+
+    @classmethod
+    def canonicalize(cls, dataset, data, coord_dims=None):
+        """
+        Canonicalize takes an array of values as input and
+        reorients and transposes it to match the canonical
+        format expected by plotting functions. In addition
```
jlstevens: It is probably too much of a pain to stick in the docstring but it would be good to describe our canonical format properly somewhere in our docs. Not something that should hold up this PR though!

jlstevens:
Looks good!

I've made most of my key comments now and in general I am happy with this. I didn't look in detail at the changes in the interface methods that already exist - I assume those are fixes as opposed to API changes.

philippjfr:
Thanks for the review.

>I assume those are fixes as opposed to API changes.

Yes mostly changes to correctly handle constant dimensions, actually integrating constant dimensions properly might actually be helpful here in future.

jlstevens:
>  actually integrating constant dimensions properly might actually be helpful here in future.

Does xarray, iris etc have a corresponding concept? In a separate PR, we might want to add a minimal API for setting/accessing constant dimensions on interfaces. 

It is a little tricky in some cases e.g wrapping numpy arrays + bounds as in such a case, I'm not sure where this state would actually live.

holoviews/core/data/grid.py
jlstevens: Also, I've never been a fan of ``get_`` as a prefix. How about simple ``coords`` or ``coordinates``?

philippjfr: Sure, just coords is fine.

philippjfr:
>Does xarray, iris etc have a corresponding concept? As I think they might, in a separate PR, we might want to consider a minimal API for setting/accessing constant dimensions on interfaces.

Iris does have explicit support and xarray at least highlights which dimensions are constant in the repr. The special handling is mostly about removing the constant axes in the array, e.g. reducing an (100, 50, 1, 1) shaped array to (100, 50). But it would be nice not to throw those constant coordinates away and turn them into constant dimensions instead.

philippjfr:
I have now applied all your comments except for folding in ``expanded_coords``. As it is now it can be implemented once on the GridInterface baseclass, if I fold it into the coords method I'll be duplicating it for each of the ``coords`` implementations.

jlstevens:
> if I fold it into the coords method I'll be duplicating it for each of the coords implementations.

Would that be true if you made ``expanded_coords`` into a utility? I can imagine each ``coords`` implementation having the signature ``def coords(cls, dataset, dim, ordered=False, expanded=False):`` and ending with something like: ``return util.expanded_coords(data, expanded)``.

philippjfr:
Okay done that as well now.

jlstevens:
Looks good! I'll merge as soon as the tests pass.

jlstevens:
Tests passed. Merging.

Issue #646, opened 2016-04-27
opts magic inheritance cmap


marqh:
```
import iris
import holoviews as hv
import geoviews as gv
import cube_browser
import cartopy
```

```
cube = iris.load_cube(iris.sample_data_path('rotated_pole.nc'))
```

with a specification of Dataset, the color map is preserved:
```
%%opts FilledContours.Air_pressure_at_sea_level [levels=7]

airp = hv.Dataset(cube)
airp.to.filledcontours(['grid_longitude', 'grid_latitude'], dynamic=True)  * gv.Feature(cartopy.feature.COASTLINE)
```

but, if the FilledContours does not specifiy the Air_pressure...., the colormap is overridden 
```

%%opts FilledContours.Air_pressure_at_sea_level [levels=7]

airp = hv.Dataset(cube)
airp.to.filledcontours(['grid_longitude', 'grid_latitude'], dynamic=True)  * gv.Feature(cartopy.feature.COASTLINE)
```
this last call should not alter the colormap

philippjfr:
I can't tell the difference between your two examples.

bjlittle:
@jlstevens says: yes, it is fine to assign it to him and you are right that the example is wrong (your version is right)

philippjfr:
It's a pretty tricky problem and I know roughly how to fix it but actually implementing the correct solution will still be pretty difficult. Hopefully @jlstevens will have the time to tackle it this week, otherwise I'll try to spend an hour or two on it over the weekend.

philippjfr:
This has finally been fixed in https://github.com/ioam/holoviews/pull/796. Closing.

PR #796, opened 2016-07-25
Dynamic inheritance in OptionTrees

jlstevens:
This PR is primarily to find out whether disabling ``merge_keywords`` has an effect on the test. If the tests continue to pass, then I propose keeping it disabled.

**Edit**: This PR turned into a fix for dynamic inheritance.

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -542,7 +542,7 @@ class OptionTree(AttrTree):
         components = (obj.__class__.__name__,
                       group_sanitizer(obj.group),
                       label_sanitizer(obj.label))
-        return self.find(components).options(group)
+        return self.find(components).options(group, target='.'.join(components))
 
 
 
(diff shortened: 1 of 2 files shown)


jlstevens:
Thanks for that last fix! Hope the tests pass this time...

jlstevens:
The existing tests passed and I am currently adding more unit tests. I should also mention that there is no need to change ``merge_keywords``and that this PR addresses issue #646.

jlstevens:
I looked into the possibility of defining a utility to replace this code for generating a string specification:

```python
components = (obj.__class__.__name__,
                        group_sanitizer(obj.group),
                        label_sanitizer(obj.label))
 '.'.join([c for c in components if c])
```

But decided this pattern doesn't appear often enough in the codebase to bother with after grepping for ``'.'.join(``. For tuple specifications, there is already ``get_spec(obj)`` which seems sufficient as tuple specs are recommended anyway.

Having added unit tests to ensure the new dynamic inheritance keeps working, I think this PR is now ready for a review/merge.

jlstevens:
I added one more unit test to simulate the %%opts magic. Ready to review/merge.

jlstevens:
I just updated  the title of the PR as I confused myself thinking it is about options on ``DynamicMap``instead of dynamic inheritance!

philippjfr:
Thanks for adding the tests. Very happy we finally fixed this.

PR #799, opened 2016-07-26
Made xarray and iris optional dependencies for dataset tests

jlstevens:
Simple PR to allow tests to be run without requiring iris or xarray to be installed.


philippjfr:
Looks good.

PR #800, opened 2016-07-26
Generalized FileArchive add method signature

jlstevens:
Simple PR to make FileArchive signature of add method compatible with ``NotebookArchive``. This means that tests should be able to run if nbformat isn't available i.e when ``FileArchive`` is set because ``NotebookArchive`` is unavailable.

### DIFF
--- a/holoviews/core/io.py
+++ b/holoviews/core/io.py
@@ -640,7 +640,7 @@ class FileArchive(Archive):
         except: raise Exception("Timestamp format invalid")
 
 
-    def add(self, obj=None, filename=None, data=None, info={}):
+    def add(self, obj=None, filename=None, data=None, info={}, **kwargs):
         """
         If a filename is supplied, it will be used. Otherwise, a
         filename will be generated from the supplied object. Note that


jlstevens:
Ready to merge.

philippjfr:
Looks fine. One question though, if you're expecting it to do a notebook export but you don't have nbformat available will it give you any indication if that's the case? Someone might think they've enabled proper exporting but they forgot to install nbformat, so export some crucial bit of work but don't realize until months later that it actually only did a FileArchive export not a NotebookArchive one.

jlstevens:
> One question though, if you're expecting it to do a notebook export but you don't have nbformat available will it give you any indication if that's the case?

Good point. I should definitely issue a warning.

jlstevens:
Actually, I think it is ok again.

What happened is that I tried running the tests before installing ``notebook`` which is why I installed ``nbformat`` explicitly. The correct thing isn't to install ``nbformat`` but to install ``notebook`` which has ``nbformat`` as a dependency.

Anyone loading the IPython extension will definitely have to have ``notebook`` and should therefore have ``nbformat`` installed. Therefore I don't think there is need for a warning here...sorry for the confusion!

philippjfr:
Sounds good. Merging.

PR #801, opened 2016-07-27
Added support for adding legends outside plot in bokeh

philippjfr:
As the title says, this allows placing legends outside the axis itself with options that mirror those also present in matplotlib. Positioning is still a bit fiddly and doesn't work perfectly for all plots but the behavior is pretty good overall. Here are some simple examples:

```python
hv.Layout([(hv.Curve(np.random.rand(10,2), label='A') * hv.Curve(np.random.rand(10,2), label='B'))(plot=dict(legend_position=pos)).relabel(pos)
           for pos in ['left', 'right', 'top', 'bottom']]).cols(2)
```

image

Additionally I also added the legend_columns plot options. In matplotlib this option is an integer to control the number of columns in a legend, bokeh only supports an orientation, which is equivalent to setting the number of columns higher than the number of entries in the legend. For consistency between the backends I'd still argue that naming them the same thing and treating any non-zero ``legend_columns`` value as a horizontal legend layout makes sense:

```python
%%opts Overlay [legend_cols=True]
hv.Layout([(hv.Curve(np.random.rand(10,2), label='A') * hv.Curve(np.random.rand(10,2), label='B'))(plot=dict(legend_position=pos)).relabel(pos)
           for pos in ['left', 'right', 'top', 'bottom']]).cols(2)
```

screen shot 2016-07-27 at 2 45 27 pm

Overall this is an improvement, it's just a bit annoying that bokeh treats legends, axes and anything else that's placed outside the main plot as part of the plot area, which means it won't preserve aspects and squish plots as you see above, but I'm hoping that's something they'll address soon.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -705,12 +705,17 @@ class OverlayPlot(GenericOverlayPlot, ElementPlot):
     legend_position = param.ObjectSelector(objects=["top_right",
                                                     "top_left",
                                                     "bottom_left",
-                                                    "bottom_right"],
+                                                    "bottom_right",
+                                                    'right', 'left',
+                                                    'top', 'bottom'],
                                                     default="top_right",
                                                     doc="""
         Allows selecting between a number of predefined legend position
         options. The predefined options may be customized in the
         legend_specs class attribute.""")
+
+    legend_cols = param.Integer(default=False, doc="""
+       Whether to lay out the legend as columns.""")
 
     tabs = param.Boolean(default=False, doc="""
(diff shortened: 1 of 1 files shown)


jlstevens:
I agree with using ``legend_cols`` in this way as I want to avoid different parameters between backends whenever possible. The aspect changes due to the legend are out of control and I also hope that Bokeh handles this better in future.

Looks good. Ready to merge?

philippjfr:
Glad you agree, might file an issue on bokeh to support columns properly, horizontal and vertical orientations are too restrictive. For now it's ready to merge, hopefully we can make the remaining improvements once bokeh has tackled them.

PR #802, opened 2016-07-27
Filter style keywords according to backend

jlstevens:
I think this functionality is best expressed with a screenshot:

screen shot 2016-07-27 at 2 54 03 pm

Note that ``linewidth`` applies to matplotlib and ``line_width`` applies to Bokeh.

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -296,6 +296,17 @@ class Options(param.Parameterized):
         super(Options, self).__init__(allowed_keywords=allowed_keywords,
                                       merge_keywords=merge_keywords, key=key)
 
+    def filtered(self, allowed):
+        """
+        Return a new Options object that is filtered by the specified
+        list of keys. Mutating self.kwargs to filter is unsafe due to
+        the option expansion that occurs on initialization.
+        """
+        kws = {k:v for k,v in self.kwargs.items() if k in allowed}
+        return self.__class__(key=self.key,
+                              allowed_keywords=self.allowed_keywords,
+                              merge_keywords=self.merge_keywords, **kws)
+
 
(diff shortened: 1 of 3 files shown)


jlstevens:
In the last commit, I skip filtering if ``style_opts`` is an empty list.

The assumption is that things like ``OverlayPlot`` don't support styles (although they may support plot options). Applying filtering on ``OverlayPlot`` was part of why the tests failed.

Is this assumption reasonable?

philippjfr:
>In the last commit, I skip filtering if style_opts is an empty list.

Sounds reasonable, yes. Not sure why not doing that would make it fail though.

jlstevens:
Without it, I saw odd things in the Elements notebook...for instance the first overlay with errorbars ended up having red errorbars instead of the usual black ones.

jlstevens:
Thanks for the last fix! Ready to merge once the tests pass.

philippjfr:
Okay, tests passing at last. Merging.

jbednar:
In the future, in cases where multiple backends use the same keyword to mean different things, should we support adding an identifier to specify the backend?  E.g. 'bokeh:linewidth=5" or "mpl:linewidth=10"?  This could also be used if we wanted to customize the two plots separately, e.g. to choose one color that looks good against mpl's default gray background, and a different color for Bokeh...

jlstevens:
> E.g. 'bokeh:linewidth=5" or "mpl:linewidth=10"?

Seems like a good suggestion - please file an issue for this as a feature request.

philippjfr:
That does sounds like a good suggestion, although I'd go for something like ``mpl:(linewidth=5 facecolor='red') bokeh:(line_width=5 fill_color='red')`` to separate them more cleanly.

PR #807, opened 2016-08-03
Improved bokeh json patch computation

philippjfr:
Improved code that tidies up bokeh json patches to update plots. Should result in an overall reduction of 30-50% in the size of the json containing the plot updates.

Things to do:

- [x] Test on a wide range of plots to ensure nothing breaks

I've now tested this on every bokeh based notebook I could find and have not encountered any issues. Very happy I could condense the plot updates this much!

### DIFF
--- a/holoviews/plotting/bokeh/util.py
+++ b/holoviews/plotting/bokeh/util.py
@@ -1,6 +1,7 @@
+import itertools
 from distutils.version import LooseVersion
-
 from collections import defaultdict
+
 import numpy as np
 
 try:
@@ -40,12 +41,8 @@ markers = {'s': {'marker': 'square'},
 # and can therefore be safely ignored. Axes currently fail saying
 # LinearAxis.computed_bounds cannot be updated
 IGNORED_MODELS = ['LinearAxis', 'LogAxis', 'DatetimeAxis',
-                  'CategoricalAxis' 'BasicTicker', 'BasicTickFormatter',
+                  'CategoricalAxis', 'BasicTicker', 'BasicTickFormatter',
                   'FixedTicker', 'FuncTickFormatter', 'LogTickFormatter']
-
-# Where to look for the ignored models
-LOCATIONS = ['new', 'below', 'right', 'left',
-             'renderers', 'above', 'attributes', 'plot', 'ticker']
 
 # Model priority order to ensure some types are updated before others
 MODEL_PRIORITY = ['Range1d', 'Title', 'Image', 'LinearColorMapper',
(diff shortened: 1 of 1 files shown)


philippjfr:
This is ready to be merged now, I've been running off this branch for almost two weeks now without any issues and it really does represent a huge saving in the amount of data that is sent to the browser.

holoviews/plotting/bokeh/util.py
```diff
@@ -176,22 +173,40 @@ def refs(json):
     return result
 
 
-def compute_static_patch(document, models, json=None):
+def get_ids(obj):
+    """
+    Returns a list of all ids in the supplied object.
+    Useful for determining if a json representation contains
+    references to other objects.
```
jlstevens: Might be worth mentioning why this knowing if the JSON has references is useful information (if it can be summarized in a simple way).

holoviews/plotting/bokeh/util.py
```diff
@@ -226,15 +242,14 @@ def delete_refs(obj, locs, delete):
             return None
         new_obj = {}
         for k, v in list(obj.items()):
-            if k in locs:
-                ref = delete_refs(v, locs, delete)
-                if ref is not None:
-                    new_obj[k] = ref
-            else:
-              new_obj[k] = v
+            if k in ['data', 'palette'] or (k == 'attributes' and not get_ids(v)):
```
jlstevens: Will this list of hardcoded strings potentially change? What do they correspond to? It might be worth specifying them as an optional argument (tuple).

holoviews/plotting/bokeh/util.py
```diff
@@ -40,13 +41,9 @@
 # and can therefore be safely ignored. Axes currently fail saying
 # LinearAxis.computed_bounds cannot be updated
 IGNORED_MODELS = ['LinearAxis', 'LogAxis', 'DatetimeAxis',
-                  'CategoricalAxis' 'BasicTicker', 'BasicTickFormatter',
+                  'CategoricalAxis', 'BasicTicker', 'BasicTickFormatter',
                   'FixedTicker', 'FuncTickFormatter', 'LogTickFormatter']
 
-# Where to look for the ignored models
-LOCATIONS = ['new', 'below', 'right', 'left',
```
jlstevens: Good to get rid of this!

jlstevens:
Ok, hopefully two weeks worth of testing means it is robust!

I am happy to merge once you've decided to either ignore/implement the two suggestions I've made in the comments.

holoviews/plotting/bokeh/util.py
philippjfr: Agreed, I don't envision them changing but breaking them out probably makes sense regardless. As a quick explanation, there are two bits of data that are sent to update a plot:

1) The events: These contain the actual attributes that have changed and need to be updated.
2) The references: These contain references to the objects in the plot but also duplicate the data.

This PR filters filters the various data attributes in the references leaving only the id references that are actually needed. This means that the ``data``, the ``palette`` and all attributes that do not contain references to other objects are deleted. Passing in the list of attributes that should be dropped explicitly makes sense so I'll do that now.

holoviews/plotting/bokeh/util.py
philippjfr: Will do.

philippjfr:
Just implemented your suggestions and improved various docstrings to make it a bit clearer how this stuff actually works. Ready to merge now.

jlstevens:
Looks good! Merging.

PR #814, opened 2016-08-16
Ensure that NdOverlay is propagated correctly in dynamic batched mode

philippjfr:
When an OverlayPlot is used in dynamic mode it is responsible for delegating the layers in the Overlay/NdOverlay to the appropriate subplots. This handling works fine when each layer in an NdOverlay is assigned its own plot (as is the case in regular non-batched mode), what was missing was handling for dynamic mode + batched mode in which case it just passes the whole NdOverlay to the subplot. This PR makes sure that happens.


jlstevens:
Happy to merge although I would like to have tests for this sort of thing once we have the Bokeh testing infrastructure in place...

PR #816, opened 2016-08-17
Datetime64 fixes

philippjfr:
Matplotlib does not currently support plotting numpy datetime64 types directly. This PR adds a compatibility layer ensuring these types are converted appropriately for plotting in matplotlib.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -5,6 +5,7 @@ baseclass for classes that accept Dimension values.
 """
 from __future__ import unicode_literals
 import re
+import datetime as dt
 from operator import itemgetter
 
 import numpy as np
@@ -13,7 +14,7 @@ import param
 from ..core.util import (basestring, sanitize_identifier,
                          group_sanitizer, label_sanitizer, max_range,
                          find_range, dimension_sanitizer, OrderedDict,
-                         safe_unicode, unicode)
+                         safe_unicode, unicode, dt64_to_dt)
 from .options import Store, StoreOptions
 from .pprint import PrettyPrinter
 
@@ -178,7 +179,11 @@ class Dimension(param.Parameterized):
             if callable(formatter):
                 return formatter(value)
             elif isinstance(formatter, basestring):
-                if re.findall(r"\{(\w+)\}", formatter):
+                if isinstance(value, dt.datetime):
+                    return value.strftime(formatter)
+                elif isinstance(value, np.datetime64):
+                    return dt64_to_dt(value).strftime(formatter)
+                elif re.findall(r"\{(\w+)\}", formatter):
                     return formatter.format(value)
                 else:
                     return formatter % value
--- a/holoviews/plotting/mpl/chart.py
+++ b/holoviews/plotting/mpl/chart.py
@@ -6,15 +6,15 @@ import numpy as np
 from matplotlib import cm
 from matplotlib import pyplot as plt
 from matplotlib.collections import LineCollection
-from matplotlib.dates import date2num
+from matplotlib.dates import date2num, DateFormatter
 
 import param
 
-from ...core import OrderedDict
+from ...core import OrderedDict, Dimension
 from ...core.util import (match_spec, unique_iterator, safe_unicode,
                           basestring, max_range, unicode)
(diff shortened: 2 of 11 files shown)


holoviews/plotting/mpl/chart.py
```diff
@@ -61,6 +63,14 @@ def get_data(self, element, ranges, style):
         ys = element.dimension_values(1)
         return (xs, ys), style, {}
 
+    def init_artists(self, ax, plot_args, plot_kwargs):
+        xs, ys = plot_args
+        if xs.dtype.kind == 'M':
```
jlstevens: Might be worth adding a comment mentioning what the 'M' code means. I assume it is the numpy datetime64 type?

holoviews/plotting/util.py
```diff
@@ -270,3 +272,10 @@ def dim_axis_label(dimensions, separator=', '):
     if not isinstance(dimensions, list): dimensions = [dimensions]
     return separator.join([safe_unicode(d.pprint_label)
                            for d in dimensions])
+
+def dt64_to_dt(dt64):
```
jlstevens: Seems like a useful utility!

jlstevens:
Looks good. Ready to merge?

philippjfr:
Think there might be a few more places where datetime64 support can be improved. Will investigate a bit more.

philippjfr:
I've now overhauled the PR again. The main changes here are that you can now register datetime formatters for datetime64 and datetime types, which means you can now do this:

```python
%%output size=100
%%opts Curve [xrotation=15 xticks=5]
import numpy as np
from datetime import date
import pandas as pd

start = date(2012, 1, 15)
end = date(2012, 1, 30)
hv.Dimension.type_formatters[np.datetime64] = '%m/%d'
hv.Curve((pd.date_range(start, end, freq='D'), np.random.rand(16)))
```

Additionally I fixed various bugs in xarray date range handling and made formatting of grid tick labels consistent with everything else. Ready to merge now, but there's a definite possibility tests will break due to Grid tick formatting changes.

philippjfr:
Ready to review, updated test data will pass shortly.

holoviews/plotting/mpl/plot.py
```diff
@@ -262,9 +262,6 @@ class GridPlot(CompositePlot):
     show_legend = param.Boolean(default=False, doc="""
         Legends add to much clutter in a grid and are disabled by default.""")
 
-    tick_format = param.String(default="%.2f", doc="""
```
jlstevens: So this parameter is now deprecated?

philippjfr: Hmm, yes, we have a system for tick formatting using ``Dimension.value_format`` and ``Dimension.type_formatters``, so this is redundant. You think we need to go through a deprecation cycle for this?

holoviews/core/dimension.py
```diff
@@ -178,7 +179,11 @@ def pprint_value(self, value):
             if callable(formatter):
                 return formatter(value)
             elif isinstance(formatter, basestring):
-                if re.findall(r"\{(\w+)\}", formatter):
+                if isinstance(value, dt.datetime):
+                    return value.strftime(formatter)
+                elif isinstance(value, np.datetime64):
+                    return dt64_to_dt(value).strftime(formatter)
+                elif re.findall(r"\{(\w+)\}", formatter):
                     return formatter.format(value)
```
jlstevens: So the datetime64 objects will use datetime style formatting? That is fine, assuming datetime64 doesn't have its own formatting conventions.

jlstevens:
Looks good. I've made two comments and once you have addressed those, I am happy to merge.

holoviews/core/dimension.py
philippjfr: Yes, I think that's nice behavior, datetime64 can't be formatted directly so it converts to regular datetime first and then applies the formatter.

holoviews/plotting/mpl/plot.py
jlstevens: It will at least need to be mentioned in the CHANGELOG. Other than that I think it is fine.

PR #817, opened 2016-08-17
Fixed handling of constant dimensions in xarray

philippjfr:
This PR fixes handling of constant dimensions in the xarray interface of HoloViews. By default a select operation does not reduce the dimensionality of a HoloViews object, i.e. the dimensionality stays unchanged. Until we have proper support for constant dimensions this is probably the right approach, however xarray does convert coordinates to constant coordinates, which changes the shape of the array. This PR ensures that the shape of the arrays stays unchanged.


jlstevens:
Looks ok, but I think we need to have a proper plan regarding constant dimensions across HoloViews. Not sure when we'll find the appropriate time to address that though...

Merging.

PR #818, opened 2016-08-17
Ensure that iris does not override supplied dimensions

philippjfr:
Previously the iris interface was simply overriding any dimensions supplied by the user, this ensures user supplied dimensions are handled correctly.


jlstevens:
Looks like a simple fix. Merging.

PR #820, opened 2016-08-18
Fix for xarray interface add_dimension implementation

philippjfr:
Fixes bug in handling add_dimension on xarray based datasets. Also fixes small bug when slicing dataset with no upper bound.

### DIFF
--- a/holoviews/core/data/xarray.py
+++ b/holoviews/core/data/xarray.py
@@ -191,7 +191,8 @@ class XArrayInterface(GridInterface):
             if isinstance(v, set):
                 validated[k] = list(v)
             elif isinstance(v, tuple):
-                validated[k] = slice(v[0], v[1]-sys.float_info.epsilon*10)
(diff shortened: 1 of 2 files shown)


holoviews/core/data/xarray.py
```diff
@@ -191,7 +191,8 @@ def select(cls, dataset, selection_mask=None, **selection):
             if isinstance(v, set):
                 validated[k] = list(v)
             elif isinstance(v, tuple):
-                validated[k] = slice(v[0], v[1]-sys.float_info.epsilon*10)
+                upper = None if v[1] is None else v[1]-sys.float_info.epsilon*10
```
jlstevens: Why ``sys.float_info.epsilon*10``? Is ``sys.float_info.epsilon`` not enough of a difference?

philippjfr: Yea apparently not, bit annoying.

jlstevens:
Looks fine. I made one comment about ``sys.float_info.epsilon``. Ready to merge?

PR #821, opened 2016-08-18
Fixed bug computing extents of mpl SpreadPlot

philippjfr:
The extent of Spread Elements wasn't being correctly computed in matplotlib. This provides a safe (but not perfect) implementation, computing the absolute largest possible extent, because there is no direct way to compute it precisely across multiple frames at the moment, something that will only become possible when we make the normalization system more general.


philippjfr:
Reviewed the updated test date, rebuilding now. Then ready to merge.

philippjfr:
Had to rebuild the test data hence why the push build is failing. Otherwise ready to merge.

jlstevens:
Looks good! Merging.

PR #822, opened 2016-08-18
Optimization for bokeh comms messages in dynamic mode

philippjfr:
Very simple optimizations for dynamic mode, including caching of the previous frame to check whether it is unchanged. Happy to hear a cleaner approach to caching the previous frame, but it's probably not worth factoring out that logic into some other object, since it's so trivial.

### DIFF
--- a/holoviews/plotting/bokeh/util.py
+++ b/holoviews/plotting/bokeh/util.py
@@ -45,7 +45,7 @@ IGNORED_MODELS = ['LinearAxis', 'LogAxis', 'DatetimeAxis',
                   'FixedTicker', 'FuncTickFormatter', 'LogTickFormatter']
 
 # List of attributes that can safely be dropped from the references
-IGNORED_ATTRIBUTES = ['data', 'palette']
+IGNORED_ATTRIBUTES = ['data', 'palette', 'image', 'x', 'y']
 
 # Model priority order to ensure some types are updated before others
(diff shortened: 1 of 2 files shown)


holoviews/plotting/bokeh/util.py
```diff
@@ -239,7 +239,8 @@ def compute_static_patch(document, models):
             update_types[obj['type']].append(key)
     events = [delete_refs(e, IGNORED_MODELS)
               for _, e in sorted(events, key=lambda x: x[0])]
-    events = [e for e in events if all(i in requested_updates for i in get_ids(e))]
+    events = [e for e in events if all(i in requested_updates for i in get_ids(e))
+              if 'new' in e]
```
jlstevens: Looks like anything with the substring 'new' will be included. Even if it is safe, it would be good to at least have a comment justifying this condition.

philippjfr: Will do. The events are dictionaries and only events which actually include a new value need to be sent.

jlstevens: And 'new' is in there as a string? Ok!

philippjfr: 'new' is the key containing the updated value. Not sure why there are events without a 'new' value in the first place since they have no effect anyway.

philippjfr:
PR build passed, push build is failing because the branch is behind master slightly. Ready to merge.

jlstevens:
Looks good. Merging.

PR #825, opened 2016-08-19
Fixed small bug in GridInterface.coords

philippjfr:
Small bug when a dimension object rather than a string is passed to the GridInterface.coords method.


philippjfr:
Just going to merge as it's a small bugfix.

PR #827, opened 2016-08-23
Allowed partial dimension overlap in DynamicMaps

philippjfr:
The code on DimensionedPlot was not robust to partial overlap in dimensions when plotting DynamicMap types. This PR ensures that partial keys (i.e. keys that are not of the same dimensionality as the overall plot) are padded with Nones before being inserted into the list of keys, ensuring that all keys are of the right length.


jlstevens:
Looks good! Merging.

PR #828, opened 2016-08-23
Avoid updating data source if element data is unchanged

philippjfr:
Instead of checking whether the Element has changed it now checks whether the element data attribute has changed, when deciding whether to send the data source data to bokeh plot.


jlstevens:
Looks good and the tests are passing. I'll merge now.

The only thing to mention is why this change was needed: ``DynamicMaps`` use ``__call__`` to propagate style information on the fly. This changes the id of the returned elements which is why the original approach was flawed. Using the id of the ``.data`` should be more robust.

PR #830, opened 2016-08-24
Allow None in keys for partial dimension overlap

philippjfr:
We allow dimensions with partial overlaps in their dimensions to be displayed together. The code will internally pad any dimension keys that need it with Nones so that the overall object has a consistent set of keys that it can be indexed by. However since NdMappings check that all dimension values must be in the list of Dimension.values the None padding will raise issues. This PR means that None is always allowed as a dimension value.


philippjfr:
Ready to merge, if you agree with the fix.

jlstevens:
Although this PR can't be merged right now due to conflicts, I'm happy to discuss this proposal again now that ``None`` values have be enabled in ``DynamicMap`` keys for handling dimensioned stream values. As the plotting code has also been updated to handle dimensioned streams, I think it will be easier to see if we want to allow None values in this case...

philippjfr:
Let's merge this. It changes nothing but validation code so I don't think there's anything else to worry about.

jlstevens:
Ok, I think we should be ok with None values now. Merging.

PR #832, opened 2016-08-25
Introduction of Streams API

jlstevens:
This is the first PR of many that will introduce an exciting new feature to HoloViews 1.7, namely interactive streams. Everything is very much work-in-progress at this stage.

Streams will make it easy to enable interactive visualizations with HoloViews by making it possible for HoloViews plots to react to events. These events may be generated on the server-side from Python or on the client-side via JavaScript. 

In this PR, the focus is on the former (no JavaScript) and future PRs will enable interaction in the browser (using the Bokeh backend). Note that the design of streams ensures they are backend agnostic as possible (like everything else in HoloViews!).

The key points about streams:

* Streams are simply Parameterized objects whose parameters change over time in response to events.
* These events involve specifying new parameter values using the ``update`` method.
* A list of stream instances can be passed to ``DynamicMap`` to enable interactivity. The stream parameters are passed as keyword arguments to the ``DynamicMap`` callable when update events occur.
* Streams have their own callbacks (e.g user defined Python functions) that are also invoked with the stream parameter values.

This PR is currently work-in-progress but here is a quick preview. Note that this example will work with the matplotlib backend after a future PR is merged to update the plotting system:

image

Right now, a dummy kdim ('Ignored') is declared which won't be needed in future.

Next the following setup is currently needed. In future, this will happen automatically and will be hidden from the user:

image

Now when you call update on the stream:

```python
m1.update(dict(y=-0.5, x=-0.5))
```

The plot responds to this event:

image

We plan to include the following streams:
``AxisRangeX``, ``AxisRangeY``, ``AxisRangeXY``, ``BoxSelection`` as well as more specialized streams such as ``ElementSelector`` and ``DataSelector``. We'll also make it as easy as possible for users to define their own custom stream implementations.
 
Currently some of our plans are outlined on the wiki [here](https://github.com/ioam/holoviews/wiki/Streams) and [here](https://github.com/ioam/holoviews/wiki/Streams-implementation). These pages are likely to go stale quickly as the implementation evolves.

Exciting stuff and obviously a lot more to come!

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -436,6 +436,15 @@ class DynamicMap(HoloMap):
         simulation time across the layout).
     """)
 
+    streams = param.List(default=[], doc="""
+       List of Stream instances to associate with the DynamicMap. The
+       set of parameter values across these streams will be supplied as
+       keyword arguments to the callback when the events are received,
+       updating the streams.
+
+       Note that streams may only be used with callable callbacks (i.e
+       not generators).""" )
+
     cache_size = param.Integer(default=500, doc="""
        The number of entries to cache for fast access. This is an LRU
        cache where the least recently used item is overwritten once
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -556,8 +556,10 @@ class DynamicMap(HoloMap):
         if self.call_mode == 'generator':
             retval = next(self.callback)
         else:
-            retval = self.callback(*args)
-
+            # Additional validation needed to ensure kwargs don't clash
+            kwarg_items = [s.value.items() for s in self.streams]
+            flattened = [el for kws in kwarg_items for el in kws]
+            retval = self.callback(*args, **dict(flattened))
         if self.call_mode=='key':
             return self._style(retval)
 
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -559,6 +559,11 @@ class DynamicMap(HoloMap):
             # Additional validation needed to ensure kwargs don't clash
             kwarg_items = [s.value.items() for s in self.streams]
             flattened = [el for kws in kwarg_items for el in kws]
+            klist = [k for k,_ in flattened]
+            clashes = set([k for k in klist if klist.count(k) > 1])
+            if clashes:
+                self.warning('Parameter name clashes for keys: %r' % clashes)
+
             retval = self.callback(*args, **dict(flattened))
         if self.call_mode=='key':
             return self._style(retval)
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -668,6 +668,8 @@ class DynamicMap(HoloMap):
 
         # Cache lookup
         try:
+            if self.streams:
+                raise KeyError('Using streams disables DynamicMap cache')
             cache = super(DynamicMap,self).__getitem__(key)
             # Return selected cache items in a new DynamicMap
             if isinstance(cache, DynamicMap) and self.mode=='open':
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -463,13 +463,13 @@ class DynamicMap(HoloMap):
        """)
 
     def __init__(self, callback, initial_items=None, **params):
+        super(DynamicMap, self).__init__(initial_items, callback=callback, **params)
 
         # Set source to self if not already specified
-        for stream in params.get('streams',[]):
+        for stream in self.streams:
             if stream.source is None:
                 stream.source = self
 
-        super(DynamicMap, self).__init__(initial_items, callback=callback, **params)
         self.counter = 0
         if self.callback is None:
             raise Exception("A suitable callback must be "
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -463,6 +463,12 @@ class DynamicMap(HoloMap):
        """)
 
     def __init__(self, callback, initial_items=None, **params):
+
+        # Set source to self if not already specified
+        for stream in params.get('streams',[]):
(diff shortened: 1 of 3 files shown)


jlstevens:
I think this PR is ready for review. Of course, this is only a first cut at streams and more work is needed to integrate them with the plotting system etc. What this PR should do is make streams available so that integration work can begin.

holoviews/streams.py
```diff
@@ -0,0 +1,166 @@
...
+        some event.
+
+        If trigger is enabled, the trigger classmethod is invoked on
+        this particular Stream instance.
+        """
+        for param in self.params().values():
+            param.constant = False
+        self.set_param(**kwargs)
+        for param in self.params().values():
+            param.constant = True
```
philippjfr: So parameters can't be set directly, ever? How would this play with paramnb, where the widgets set the parameter values directly. I was hoping I could simply attach Stream.update as a callback on the paramnb widgets and have it just work, but if the params are constant it will never work.

jlstevens: I think that is fine: just subclass from ``Stream`` without declaring the parameters as constant to create a parameterized object to use with paramnb. 

What I think I can fix is to restore the ``constant`` state to what it was originally. That way,  declaring stream parameters them as constant is recommended but not mandatory and they suddenly become constant when update is used.

philippjfr: Sounds reasonable.

jbednar: I'm a bit a alarmed by the suggestion to subclass from Stream for creating a parameterized object for use with paramnb -- the whole point of paramnb is to be able to display Jupyter-editable parameter widgets for parameters that are defined independently of any GUI or plotting system.  The underlying parameterized object sometimes is created just for the GUI, but in many cases is something that doesn't know or need to know anything about anything to do with GUIs or plotting.  Maybe in this particular case it is fine, because people using paramnb together with HoloViews streams are already doing things tied to plotting, but it does alarm me!

philippjfr: It's just one way of doing it, you can easily create a callback to attach to the paramnb widgets which copies the parameterized objects parameters over to a stream.

holoviews/core/spaces.py
```diff
@@ -454,6 +463,12 @@ class DynamicMap(HoloMap):
        """)
 
     def __init__(self, callback, initial_items=None, **params):
+
+        # Set source to self if not already specified
+        for stream in params.get('streams',[]):
```
philippjfr: This should happen after the parameters have been instantiated. I wouldn't recommend it but someone could set a number of default streams directly on DynamicMap.

holoviews/streams.py
```diff
@@ -0,0 +1,195 @@
...
+        Given a list of streams, collect all the stream parameters into
+        a dictionary and pass it to the union set of subscribers.
+
+        Passing multiple streams at once to trigger can be useful when a
+        subscriber may be set multiple times across streams but only
+        needs to be called once.
+        """
+        # Union of stream values
+        items = [stream.value.items() for stream in streams]
+        union = dict(kv for kvs in items for kv in kvs)
```
philippjfr: What happens when there are clashes? Shouldn't it warn at least?

holoviews/core/spaces.py
```diff
@@ -541,8 +556,10 @@ def _execute_callback(self, *args):
         if self.call_mode == 'generator':
             retval = next(self.callback)
         else:
-            retval = self.callback(*args)
-
+            # Additional validation needed to ensure kwargs don't clash
+            kwarg_items = [s.value.items() for s in self.streams]
```
philippjfr: Same as below, no warning for clashing kwargs?

holoviews/streams.py
jlstevens: Sure. I was planning to do that and then forgot.

Once issue is that ``self.warning`` seems to be showing up in the console and not the notebook...

philippjfr: Somehow it works for Jim, not sure why.

philippjfr:
Apart from my three comments it looks good. Doesn't need to be finalized right now anyway. Once we've hooked everything up we can still reconsider some of the API.

holoviews/core/spaces.py
jlstevens: Agreed. I'll fix this now.

jlstevens:
I believe I've address your three comments. Ready for final review/merge.

holoviews/streams.py
```diff
@@ -0,0 +1,220 @@
...
+        union = [kv for kvs in items for kv in kvs]
+        klist = [k for k,_ in union]
+        clashes = set([k for k in klist if klist.count(k) > 1])
+        if clashes:
+            param.main.warning('Parameter name clashes for keys: %r' % clashes)
+
+        # Currently building a simple set of subscribers
+        groups = [stream.subscribers + stream._hidden_subscribers for stream in streams]
+        subscribers = set(s for subscribers in groups for s in subscribers)
+        for subscriber in subscribers:
```
philippjfr: Hmm, shouldn't some order be maintained, so that the subscribers defined on an individual stream are at least executed in sequence and the _hidden_subscribers are executed after all others?

jlstevens: Yes, that makes sense mainly because we know that ``_hidden_subscribers`` is the sort of thing we want to batch and execute last.

philippjfr: Right, currently it is not well defined when the plot redraw would happen.

philippjfr:
Looks good, tests will pass. Now we just need to hook it all up, exciting!

holoviews/streams.py
jbednar: It doesn't *always* work; sometimes I see messages from param on the console.  But usually they go to the notebook.

PR #834, opened 2016-08-29
Fixed coverage reporting

philippjfr:
The coverage module introduced a backward compatibility breaking change in v4.2 (see https://coverage.readthedocs.io/en/coverage-4.2/changes.html#version-4-2-2016-07-26). This PR ensures coverage reports are combined correctly again.


jlstevens:
It will be great to have a correctly reported coverage again!

PR #835, opened 2016-08-30
Fix for bokeh plot updates when frame is empty

philippjfr:


jlstevens:
I checked out this branch and it did fix the issue I was having. Thanks!

PR #837, opened 2016-08-30
Allowed hiding legends on bokeh BarPlot

philippjfr:


jlstevens:
Looks like a simple fix. Merging.

PR #838, opened 2016-08-30
Add Comms to allow updating plots dynamically

philippjfr:
This PR introduces Comms classes which allow bidirectionally communicating between the python process and a frontend. Additionally it gives Plot instances access to the current renderer and instantiates a Comms class, allowing updates to plots to be pushed to the frontend using the new ``refresh`` and ``push`` methods. Supports updating regular matplotlib plots, nbagg plots, mpld3 plots and bokeh plots.

### DIFF
--- a/holoviews/plotting/bokeh/widgets.py
+++ b/holoviews/plotting/bokeh/widgets.py
@@ -5,7 +5,7 @@ from bokeh.io import _CommsHandle
 from bokeh.util.notebook import get_comms
 
 from ..widgets import NdWidget, SelectionWidget, ScrubberWidget
-from .util import compute_static_patch, serialize_json
+from .util import serialize_json
 
 class BokehWidget(NdWidget):
 
@@ -39,7 +39,7 @@ class BokehWidget(NdWidget):
             if fig_format == 'html':
                 msg = self.renderer.html(self.plot, fig_format)
             else:
-                json_patch = self.renderer.patch(self.plot, serialize=False)
+                json_patch = self.renderer.diff(self.plot, serialize=False)
                 msg = dict(patch=json_patch, root=self.plot.state._id)
                 msg = serialize_json(msg)
             return msg
--- a/holoviews/plotting/bokeh/widgets.py
+++ b/holoviews/plotting/bokeh/widgets.py
@@ -36,22 +36,15 @@ class BokehWidget(NdWidget):
         """
         self.plot.update(idx)
         if self.embed or fig_format == 'html':
-            html = self.renderer.html(self.plot, fig_format)
-            return html
-        else:
-            doc = self.plot.document
-            if hasattr(doc, 'last_comms_handle'):
-                handle = doc.last_comms_handle
+            if fig_format == 'html':
+                msg = self.renderer.html(self.plot, fig_format)
             else:
-                handle = _CommsHandle(get_comms(doc.last_comms_target),
-                                      doc, doc.to_json())
-                doc.last_comms_handle = handle
-
-            plotobjects = [h for handles in self.plot.traverse(lambda x: x.current_handles)
-                           for h in handles]
-            msg = compute_static_patch(doc, plotobjects)
-            handle.comms.send(json.dumps(msg))
-            return 'Complete'
+                json_patch = self.renderer.patch(self.plot, serialize=False)
+                msg = dict(patch=json_patch, root=self.plot.doc._id)
+            return msg
+        else:
+            self.plot.push()
+            return "Complete"
 
 
 class BokehSelectionWidget(BokehWidget, SelectionWidget):
--- a/holoviews/plotting/bokeh/widgets.py
+++ b/holoviews/plotting/bokeh/widgets.py
@@ -5,7 +5,7 @@ from bokeh.io import _CommsHandle
 from bokeh.util.notebook import get_comms
 
 from ..widgets import NdWidget, SelectionWidget, ScrubberWidget
-from .util import compute_static_patch
+from .util import compute_static_patch, serialize_json
 
 class BokehWidget(NdWidget):
 
@@ -40,7 +40,8 @@ class BokehWidget(NdWidget):
                 msg = self.renderer.html(self.plot, fig_format)
             else:
                 json_patch = self.renderer.patch(self.plot, serialize=False)
-                msg = dict(patch=json_patch, root=self.plot.doc._id)
+                msg = dict(patch=json_patch, root=self.plot.state._id)
+                msg = serialize_json(msg)
             return msg
         else:
             self.plot.push()
--- a/holoviews/plotting/comms.py
+++ b/holoviews/plotting/comms.py
@@ -74,6 +74,7 @@ class JupyterComm(Comm):
     template = """
     <script>
       function msg_handler(msg) {{
+        var data = msg.content.data;
         {msg_handler}
       }}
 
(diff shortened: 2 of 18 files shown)


jbednar:
Can you briefly summarize what new functionality this will support, and/or what types of simplifications to the codebase will be possible, depending on your motivation here?

philippjfr:
Sure, probably the right place to put this even though it's only a first step in that direction. 

Eventually the aim of Comms is to allow any bidirectional communication between a Python process, whether that is a webserver or a Jupyter kernel, and a frontend. This will allow defining complex interactions using the upcoming Streams interface but should also eventually replace the existing mechanisms on our widgets to communicate widget state to python and update a plot on the frontend. One remaining obstacle will be to appropriately throttle events sent from the frontend to the python process.

jlstevens:
At a higher level, I think of comms as offering a way to standardize communication with HoloViews visualizations that live in the browser (not necessarily the notebook). This should help simplify things and make it easier to implement new functionality whereby plots get updated over some communication channel (Jupyter Comms, websockets, some raw TCP socket, pipes, RPC etc).

jbednar:
And Bokeh server?

jlstevens:
> And Bokeh server?

Most probably! We will definitely find a way to communicate with Bokeh visualizations without relying on Jupyter comms.

holoviews/plotting/mpl/renderer.py
```diff
@@ -135,6 +141,19 @@ def get_size(self_or_cls, plot):
         return (w*dpi, h*dpi)
 
 
+    def patch(self, plot):
```
jlstevens: I think ``diff`` would be a clearer name.

holoviews/plotting/mpl/comms.py
```diff
@@ -0,0 +1,87 @@
...
+"""
+
+mpld3_msg_handler = """
+target = $('#fig_el{comms_target}');
+target.children().each(function () {{ $(this).remove() }});
+mpld3.draw_figure("fig_el{comms_target}", data);
+"""
+
+class WidgetCommSocket(CommSocket):
+    """
```
jlstevens: I suggest calling this ``NbAggCommSocket``...

jlstevens:
Made a couple of comments regarding naming.

Overall, I'm very pleased! This looks a lot clearer to me, the APIs look clean and I feel that I understand how it all works.

I realize it is still work-in-progress but it is already looking good!

holoviews/plotting/mpl/comms.py
```diff
@@ -0,0 +1,87 @@
...
+    CustomCommSocket provides communication between the IPython
+    kernel and a matplotlib canvas element in the notebook.
+    A CustomCommSocket is required to delay communication
+    between the kernel and the canvas element until the widget
+    has been rendered in the notebook.
+    """
```
jlstevens: I believe ``CustomCommSocket`` must be an outdated name. This docstring should mention that ``CommSocket`` is a matplotlib concept that you are customizing to support NbAgg.

holoviews/plotting/comms.py
```diff
@@ -0,0 +1,165 @@
...
+        """
+        Decode received message before passing it to on_msg callback
+        if it has been defined.
+        """
+        if self._on_msg:
+            self._on_msg(self.decode(msg))
+
+
+class SimpleJupyterComm(Comm):
+    """
```
jlstevens: How about calling it ``JupyterPushComm``? From the sound of it, we may not need it in future one bokeh works with bi-directional comms.

holoviews/plotting/bokeh/widgets.py
```diff
@@ -36,22 +36,16 @@ def _plot_figure(self, idx, fig_format='json'):
...
             else:
-                handle = _CommsHandle(get_comms(doc.last_comms_target),
-                                      doc, doc.to_json())
-                doc.last_comms_handle = handle
-
-            plotobjects = [h for handles in self.plot.traverse(lambda x: x.current_handles)
-                           for h in handles]
-            msg = compute_static_patch(doc, plotobjects)
-            handle.comms.send(json.dumps(msg))
-            return 'Complete'
```
jlstevens: I would consider a better name than 'Complete'. Maybe 'Transferred', 'Transfer Complete' or maybe just 'Sent'. Not too important though...

holoviews/plotting/mpl/comms.py
```diff
@@ -0,0 +1,89 @@
...
+img = $('<div />').html(data);
+target.children().each(function () {{ $(this).remove() }})
+target.append(img)
+"""
+
+mpld3_msg_handler = """
+target = $('#fig_el{comms_target}');
+target.children().each(function () {{ $(this).remove() }});
+mpld3.draw_figure("fig_el{comms_target}", data);
+"""
```
jlstevens: I don't see a ``bokeh_msg_handler`` anywhere in the file. Is that because it is already set up for you?

philippjfr: Yes, both registration and message handling is provided by bokeh itself, which is why I'm stuck using the JupyterPushComm for now at least.

holoviews/plotting/mpl/comms.py
```diff
@@ -0,0 +1,89 @@
...
+from mpl_toolkits.mplot3d import Axes3D
+
+from ..comms import JupyterComm
+
+mpl_msg_handler = """
+target = $('#{comms_target}');
+img = $('<div />').html(data);
+target.children().each(function () {{ $(this).remove() }})
+target.append(img)
+"""
```
jlstevens: Given this is the body of the function and not the complete function, I think a comment could help make it clearer. E.g:

mpl_msg_handler = """
/* Body of the msg_handler function that is passed the message */
target = $('#{comms_target}');
img = $('<div />').html(data);
target.children().each(function () {{ $(this).remove() }})
target.append(img)
"""

I think it should be Javascript comment as it is about Javascript code. I think this will help anyone copying and pasting this to write a new message handler.

holoviews/plotting/mpl/renderer.py
```diff
@@ -81,6 +82,11 @@ class MPLRenderer(Renderer):
     widgets = {'scrubber': MPLScrubberWidget,
                'widgets': MPLSelectionWidget}
 
+    # Define comm targets by mode
+    comms = {'default': (JupyterComm, mpl_msg_handler),
+             'nbagg':   (NbAggJupyterComm, None),
+             'mpld3':   (JupyterComm, mpld3_msg_handler)}
```
jlstevens: In a later PR it would be good to have a nice way for the user to select the appropriate comms. E.g using ``WebsocketComms`` if not using the notebook.

tests/testplotinstantiation.py
```diff
@@ -23,6 +24,11 @@ class TestPlotInstantiation(ComparisonTestCase):
     def setUp(self):
         if pyplot is None:
             raise SkipTest("Matplotlib required to test plot instantiation")
+        self.default_comm = renderer.comms['default']
+        renderer.comms['default'] = JupyterPushComms
+
+    def teardown(self):
+        renderer.comms['default'] = self.default_comm
```
jlstevens: My last comment is that it might be nice to have as many unit tests as possible. I realize it is tricky without the notebook and Javascript but maybe the decode method can be tested (if made into a classmethod). Hopefully, ``__init__`` can also be tested in some cases...

jlstevens:
I've now reviewed this PR and made all the comments I want to make. Overall, I'm really pleased with all of this! Once these comments are addressed, I am happy to merge.

PR #840, opened 2016-08-31
Cleaned up CompositePlot instantiation

philippjfr:
So I was about to add some code to different plots classes to attach streams and I got fed up with all the duplication in the composite plot constructors. I've moved everything to the base class cutting down a lot of the mess.


philippjfr:
Ready to merge.

PR #841, opened 2016-08-31
Allowed NdMappings with no kdims

philippjfr:
This fixes indexing for 0D NdMappings and allows displaying them as a single frame.

### DIFF
--- a/holoviews/core/ndmapping.py
+++ b/holoviews/core/ndmapping.py
@@ -573,6 +573,8 @@ class NdMapping(MultiDimensionalMapping):
                 raise IndexError("Boolean index must match length of sliced object")
             selection = zip(indexslice, self.data.items())
             return self.clone([item for c, item in selection if c])
+        elif indexslice == () and not self.kdims:
+            return self.data[()]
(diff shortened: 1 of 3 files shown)


philippjfr:
Ready to merge.

jlstevens:
I think it might be worth using param to ensure the length of the kdim list is one or more for all NdMapping types except ``DynamicMap`` where we can override it and allow an empty list. That can then make use of this PR without making things more flexible elsewhere (where it isn't needed) that could cause problems.

Otherwise, looks fine.

jlstevens:
Just to summarize a discussion I had with Philipp:

This PR allows some fairly odd things such as 0D HoloMaps. As it turns out, they need to be possible for 0D DynamicMaps to display. 

In the end, we decided to allow these odd but semantically valid HoloMaps because they are semantically valid, used in the plotting code and it seems unlikely that users will ever encounter them. For instance, to declare one, you would need to explicit use an empty tuple key and then declare ``kdims=[]`` as there is one kdim declared by default.

PR #842, opened 2016-08-31
Correctly initialize comms and connect streams

philippjfr:
This PR ensures that only the top-level plot instance creates a Comm instance and then attaches the plot.refresh method as a ``_hidden_subscriber`` to the stream. A simple test I used:

```python
posx = hv.streams.PositionX()
dmap = hv.DynamicMap(lambda x: hv.Image(np.random.rand(x+1,10)), kdims=[], streams=[posx])
```

```python
dmap + hv.Image(np.random.rand(10,10))
```

image

```python
for i in range(10):
    posx.update(x=i)
```

The plot updates in response to the position updates.


jlstevens:
Looks good. Merging.

PR #844, opened 2016-09-01
Param values stream

jlstevens:
This PR introduced the ``ParamValues`` stream which allows a stream interface to be used to manipulate the parameters of an object that is not the stream itself.

### DIFF
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -223,3 +223,60 @@ class PositionXY(Stream):
     def __init__(self, preprocessors=[], source=None, subscribers=[], **params):
         super(PositionXY, self).__init__(preprocessors=preprocessors, source=source,
                                         subscribers=subscribers, **params)
+
+
+
+
+class ParamValues(Stream):
+    """
+    A Stream based on the parameter values of some other parameterized
+    object, whether it is a parameterized class or a parameterized
+    instance.
+
+    The update method enables the stream to update the parameters of the
+    specified object.
+    """
+
+    def __init__(self, obj, **params):
+        self._obj = obj
+        super(ParamValues, self).__init__(**params)
+
+
+    @property
+    def value(self):
+        if isinstance(self._obj, type):
+            remapped={k: getattr(self._obj,k)
+                               for k in self._obj.params().keys() if k!= 'name'}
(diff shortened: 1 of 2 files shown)


philippjfr:
Looks good!

holoviews/streams.py
```diff
@@ -223,3 +223,60 @@ class PositionXY(Stream):
...
+
+class ParamValues(Stream):
+    """
+    A Stream based on the parameter values of some other parameterized
+    object, whether it is a parameterized class or a parameterized
+    instance.
+
+    The update method enables the stream to update the parameters of the
+    specified object.
+    """
```
jbednar: Can the docstring be improved?  The PR makes it sound like the stream is updating the parameters, primarily, but what I thought was that it went the other way around, and here it seems to do both.  Are these two very different usages?  Seems like something to explain.

jlstevens: Sure it can be improved but I don't think anything surprising is going on. Reading values from parameters is how streams specify their value and the update method is how they update those values. In other words this is exactly the behavior you would expect given how streams work in general and what is needed by their API.

philippjfr: Well, in the case of paramnb the widgets set the parameter values directly and the update method won't really do anything, so there's two ways the parameters can be set. HoloViews doesn't care which one is actually used though.

jlstevens: Right. Though if you aren't using update to actually update parameters, you can use ``Stream.trigger([callback])`` instead. For paramnb it would be more awkward than update as you would need to wrap that in a lambda or something.

jbednar: I'm still confused; seems like Philipp is saying that for paramnb update doesn't do anything, and Jean-Luc is saying that if not using update you'd have to do something else?  I guess Philipp is saying that update doesn't do anything but triggering, with paramnb?

What seems like would help is if the docstring outlined at least two different expected usages for this object, one where the object updates the parameter values itself, which is what triggers changes elsewhere, and presumably one where it needs to consume changes to the parameters made elsewhere.  Anyway, there's definitely something I'm confused about, but I may be confused about precisely what that is, so this may not be the real thing that needs explaining.

PR #845, opened 2016-09-02
Renamed Stream.value to 'contents' to avoid name clashes

jlstevens:
A parameter called ``value`` would be very natural to use in a stream making it likely to clash with the ``value`` property. Renaming this property to ``contents`` should help avoid potential name clashes and confusion.

Ready to merge.


philippjfr:
Looks good, ``contents`` is much less likely to clash than ``value``. I'll merge when tests pass.

jlstevens:
Push build has passed. I had to restart one of the pr builds due to a transient error - you can wait for that to pass but I don't think you need to!

PR #846, opened 2016-09-02
Dimensioned streams

jlstevens:
This PR implements dimensioned streams i.e support for streams with parameters listed in the key dimensions. This generalizes streams fully so that everything you can achieve with a regular DynamicMap can now be achieved using a stream (i.e our sliders could now use a streams based approach for dynamic maps).

I think this PR is mostly complete but it isn't quite ready yet. I would like to test it more, taking time to ensure that the counter and generator ('open') modes work correctly. In addition, I want to write some unit tests for the three new utility functions I've introduced as well as for dynamic maps with dimensioned streams.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -729,7 +729,7 @@ class DynamicMap(HoloMap):
         (key, val) = (retval if isinstance(retval, tuple)
                       else (self.counter, retval))
 
-        key = util.wrap_tuple(key)
+        key = util.wrap_tuple_streams(key, self.kdims, self.streams)
         if len(key) != len(self.key_dimensions):
             raise Exception("Generated key does not match the number of key dimensions")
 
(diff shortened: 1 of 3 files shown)


jlstevens:
I thought I might as well give an example of how this is used:

image

As the stream parameters 'x' and 'y' are listed in the kdims, you can index the ``DynamicMap`` with ``__getitem__`` normally. You can specify values explicitly or ``None`` which means 'get the value from the stream'. As this dynamic map is now dimensioned, it has a cache that can fill with keys and data. This means you can cast it to a ``HoloMap`` which can be statically exported.

jlstevens:
Tests have passed except for one transient. There are several issues we are aware of but they won't be addressed in this PR.

* Multiple calls to the dynamic map per plotting refresh.
* Dimensioned streams do not mix well with regular key dimensions where sliders are required.

Addressing these involves distinct tweaks to DynamicMap so it is probably worth merging this PR to continue with this work.

philippjfr:
Looks good to me and only affects DynamicMaps with streams. PRs to fix multiple dynamic calls and widgets to follow in separate PRs.

Issue #813, opened 2016-08-16
Restoring option state on DynamicMaps


jlstevens:
This issue is in two parts, one requiring an easy fix and a second part that is more involved.

1. When an exception occurs, the display hooks attempt to restore the option state using ``StoreOptions.state(element, state=optstate)``. When a ``DynamicMap`` is involved, this fails with ``StopIteration``. This exception should be caught and silenced as ``StopIteration`` indicates there are no elements to traverse and therefore nothing to restore state on.

2. Actually restoring the option state on ``DynamicMaps`` properly. ``DynamicMap`` requires special handling as the option state is held on the container and dynamically propagated to the elements as they are generated. ``StoreOptions.state`` therefore needs to be ``DynamicMap`` aware to handle this restoration properly.

jlstevens:
The above PR addresses part 1. though not in the way I thought it would. Catching ``StopIteration`` wouldn't be a good solution as it would mean that the style restoration would stop as soon as it encountered a DynamicMap.

Part 2. above will need to be addressed in a separate PR.

philippjfr:
Will you tackle this for 1.7? Otherwise please postpone to 1.8.

jlstevens:
That depends on when 1.7 is ready: as soon as we have proper integration with Bokeh events, I feel we should release. This particular issue shouldn't block 1.7 but I would like to tackle it before then if I have time.

philippjfr:
Doesn't seem likely this will happen, and there's higher priority issues.

philippjfr:
@jlstevens Is this still needed? I don't quite remember if this is specific to the magics or not.

PR #848, opened 2016-09-04
Disable DynamicMap._deep_indexable in display hook exceptions

jlstevens:
I am not sure this is the *correct* fix (setting DynamicMap as not being ``_deep_indexable) but it is a simple fix for the issue with style restoration on DynamicMaps in the display hooks.

The issue was that ``StopIteration`` was being raised in ``traverse`` on dynamic maps. There are two other possible approaches I can think of:

1. Give ``DynamicMap`` an alternative implementation of ``traverse``. I don't like this much as we rely on ``traverse`` a lot and expect it to always behave in a consistent way.
2. Temporarily declare ``DynamicMap`` as not being ``deep_indexable`` during style restoration.

I might look into option 2. now as it would be a less dramatic change.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -601,7 +601,7 @@ class Dimensioned(LabelledData):
 
     def _valid_dimensions(self, dimensions):
         """Validates key dimension input
-        
+
         Returns kdims if no dimensions are specified"""
         if dimensions is None:
             dimensions = self.kdims
--- a/holoviews/ipython/display_hooks.py
+++ b/holoviews/ipython/display_hooks.py
@@ -122,8 +122,7 @@ def display_hook(fn):
             sys.stderr.write("Rendering process skipped: %s" % str(e))
             return None
         except AbbreviatedException as e:
-            try:
-                option_state(element, state=optstate)
+            try: option_state(element, state=optstate)
             except: pass
             FULL_TRACEBACK = '\n'.join(traceback.format_exception(e.etype,
                                                                   e.value,
@@ -134,8 +133,7 @@ def display_hook(fn):
             return "<b>{name}</b>{msg}<br>{message}".format(msg=msg, **info)
 
         except Exception as e:
-            try:
-                option_state(element, state=optstate)
+            try: option_state(element, state=optstate)
             except: pass
             raise
     return wrapped
--- a/holoviews/ipython/display_hooks.py
+++ b/holoviews/ipython/display_hooks.py
@@ -89,6 +89,13 @@ def last_frame(obj):
 # Display hooks #
 #===============#
 
+def option_state(element, state=None):
+    # Temporary fix to avoid issues with DynamicMap traversal
+    DynamicMap._deep_indexable = False
+    optstate = StoreOptions.state(element,state=state)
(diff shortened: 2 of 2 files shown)


jlstevens:
The following diff also fixes the issue. I'm not sure I want to commit it as it is fairly ugly and maybe we do want to disable deep indexing for DynamicMaps?

```diff
         global FULL_TRACEBACK
         if Store.current_backend is None:
             return
+        DynamicMap._deep_indexable = False
         optstate = StoreOptions.state(element)
+        DynamicMap._deep_indexable = True
         try:
             html = fn(element,
                       max_frames=OutputMagic.options['max_frames'],
@@ -115,7 +117,10 @@ def display_hook(fn):
             sys.stderr.write("Rendering process skipped: %s" % str(e))
             return None
         except AbbreviatedException as e:
-            try:    StoreOptions.state(element, state=optstate)
+            try:
+                DynamicMap._deep_indexable = False
+                StoreOptions.state(element, state=optstate)
+                DynamicMap._deep_indexable = True
             except: pass
             FULL_TRACEBACK = '\n'.join(traceback.format_exception(e.etype,
                                                                   e.value,
@@ -126,7 +131,10 @@ def display_hook(fn):
             return "<b>{name}</b>{msg}<br>{message}".format(msg=msg, **info)
 
         except Exception as e:
-            try:    StoreOptions.state(element, state=optstate)
+            try:
+                DynamicMap._deep_indexable = False
+                StoreOptions.state(element, state=optstate)
+                DynamicMap._deep_indexable = True
             except: pass
             raise
     return wrapped
```

jlstevens:
I have now gone for the latter approach, setting ``_deep_indexable`` to ``False`` temporarily in the display hooks.

philippjfr:
>I have now gone for the latter approach, setting _deep_indexable to False temporarily in the display hooks.

That sounds okay although it is not the correct fix. The problem is that the options state that was initially recorded for the displayed object is no longer long enough because the DynamicMap has potentially created more items. The only way I can see of doing this correctly is to collect both the id and the objects in the initial traverse and then additionally collect all the objects that were styled dynamically. That way you would collect all the objects and their original ids, which you can iterate over to restore the original ids.

philippjfr:
If you want to leave it at this fix for now that's okay, but we should keep an issue open to make sure we implement the proper fix at some point whatever form that will take.

jlstevens:
> The problem is that the options state that was initially recorded for the displayed object is no longer long enough because the DynamicMap has potentially created more items. 

Yes, that's right. ``optstate = option_state(element)`` runs before ``html = fn(element, ...)`` which in the case of ``DynamicMap`` creates new elements.

I agree this is a temporary fix and there is already an issue to track this #813. In future there will be a PR to fix this properly - I think it will be tricky as DynamicMaps already have their own approach to styling elements.

jlstevens:
If this quick fix isn't too ugly, I think it is ready to merge. I've assigned the corresponding issue to the 1.7 milestone to make sure we don't forget to get a better fix in place before then.

philippjfr:
As a temporary fix this is fine, since it's currently broken in a number of ways and is actually likely to restore the wrong id to various objects. Merging.

PR #849, opened 2016-09-05
Make HeatMap more general

philippjfr:
The HeatMap Element has gone through a lot of redesign, initially being defined as a Raster type and then allowing columnar data. This PR refactors HeatMap to improve the simplify aggregation step and allow any number of value dimensions which can appear as hover information in bokeh. Still a WIP and I need to ensure the code is backward compatible (it should be).

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -610,11 +610,11 @@ def sort_topologically(graph):
     [set([1, 2]), set([3, 4]), set([5, 6])]
     """
     levels_by_name = {}
-    names_by_level = defaultdict(set)
+    names_by_level = defaultdict(list)
 
     def add_level_to_name(name, level):
         levels_by_name[name] = level
-        names_by_level[level].add(name)
+        names_by_level[level].append(name)
 
 
     def walk_depth_first(name):
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -649,7 +649,10 @@ def sort_topologically(graph):
 
 
 def is_cyclic(graph):
-    """Return True if the directed graph g has a cycle."""
+    """
+    Return True if the directed graph g has a cycle. The directed graph
+    should be represented as adictionary mapping of edges for each node.
+    """
     path = set()
 
     def visit(vertex):
@@ -664,9 +667,11 @@ def is_cyclic(graph):
 
 
 def one_to_one(graph, nodes):
-    """Return True if graph contains only one to one mappings.
-    Pass a graph as a dictionary mapping of edges for each node and
-    a list of all nodes."""
+    """
+    Return True if graph contains only one to one mappings. The
+    directed graph should be represented as a dictionary mapping of
+    edges for each node. Nodes should be passed a simple list.
+    """
     edges = itertools.chain.from_iterable(graph.values())
     return len(graph) == len(nodes) and len(set(edges)) == len(nodes)
 
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -663,19 +663,12 @@ def is_cyclic(graph):
     return any(visit(v) for v in graph)
 
 
-def one_to_one(graph):
-    """Return True if graph contains only one to one mappings."""
-    start, end = set(), set()
-    for s, e in graph.items():
-        if len(e) == 0:
-            start.add(s)
-            continue
-        e = e[0]
-        if len(e) > 1 or s in start or e in end:
-            return False
-        start.add(s)
-        end.add(e)
-    return True
+def one_to_one(graph, nodes):
+    """Return True if graph contains only one to one mappings.
+    Pass a graph as a dictionary mapping of edges for each node and
+    a list of all nodes."""
+    edges = itertools.chain.from_iterable(graph.values())
+    return len(graph) == len(nodes) and len(set(edges)) == len(nodes)
 
 
 def get_overlay_spec(o, k, v):
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -996,3 +996,13 @@ def dt64_to_dt(dt64):
     """
     ts = (dt64 - np.datetime64('1970-01-01T00:00:00Z')) / np.timedelta64(1, 's')
     return dt.datetime.utcfromtimestamp(ts)
+
+
+def is_nan(x):
+    """
+    Checks whether value is NaN on arbitrary types
+    """
+    try:
+        return np.isnan(x)
+    except:
+        return False
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -607,7 +607,7 @@ def sort_topologically(graph):
     }
 
     sort_topologically(graph)
-    [set([1, 2]), set([3, 4]), set([5, 6])]
+    [[1, 2], [3, 4], [5, 6]]
     """
     levels_by_name = {}
     names_by_level = defaultdict(list)
@@ -646,6 +646,37 @@ def sort_topologically(graph):
     return list(itertools.takewhile(lambda x: x is not None,
                                     (names_by_level.get(i, None)
                                      for i in itertools.count())))
+
+
+def is_cyclic(graph):
+    """Return True if the directed graph g has a cycle."""
+    path = set()
+
+    def visit(vertex):
+        path.add(vertex)
+        for neighbour in graph.get(vertex, ()):
+            if neighbour in path or visit(neighbour):
+                return True
+        path.remove(vertex)
+        return False
+
+    return any(visit(v) for v in graph)
+
+
+def one_to_one(graph):
+    """Return True if graph contains only one to one mappings."""
+    start, end = set(), set()
+    for s, e in graph.items():
+        if len(e) == 0:
+            start.add(s)
+            continue
+        e = e[0]
+        if len(e) > 1 or s in start or e in end:
+            return False
+        start.add(s)
+        end.add(e)
+    return True
+
 
 def get_overlay_spec(o, k, v):
     """
--- a/holoviews/element/raster.py
+++ b/holoviews/element/raster.py
@@ -14,7 +14,7 @@ from ..core.sheetcoords import SheetCoordinateSystem, Slice
 from ..core.util import pd
 from .chart import Curve
 from .tabular import Table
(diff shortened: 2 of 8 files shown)


philippjfr:
@jlstevens The PR is now ready for review, I'll rebuild the test data shortly. The test data will have to be updated because ``HeatMap`` no longer pads with ``NaN`` values in the constructor, instead it computes a 2D gridded aggregate, which is used for display. I've also readded the previously supported ``raster`` as a property so it can go through a deprecation cycle. Representing the HeatMap as a gridded aggregate is much more flexible since it allows multiple value dimensions.

philippjfr:
Still need to add thorough unit tests for the aggregation. I'm also now wondering whether I should add a warning when you pass non-aggregated data to a HeatMap, i.e. if you had two clashing values for the same 2D index, e.g. there are two different values for index ('A', 'a') in the heatmap, it would just silently ignore the second value. Really it should warn you that you should aggregate your data using some function (mean, max, min) before passing it to the HeatMap.

philippjfr:
Also deprecates support for unpickling the original ``HeatMap`` format (which was replaced in 1.4).

holoviews/element/util.py
```diff
@@ -30,3 +39,60 @@ def compute_edges(edges):
...
+    """
+    values = x.values if pd and isinstance(x, pd.Series) else x
+    for v in values:
+        if not is_nan(v):
+            return v
+    return np.NaN
+
+
+def get_2d_aggregate(obj):
+    """
```
jlstevens: Perhaps this would be better expressed as an operation? Then maybe it could have a minimal docstring example in the class docstring?

holoviews/element/raster.py
```diff
@@ -383,85 +381,18 @@ class HeatMap(Dataset, Element2D):
...
-
-
-    def dframe(self, dense=False):
-        if dense:
-            keys1, keys2 = self.dense_keys()
-            dense_map = self.clone({(k1, k2): self._data.get((k1, k2), np.NaN)
-                                 for k1, k2 in product(keys1, keys2)})
-            return dense_map.dframe()
-        return super(HeatMap, self).dframe()
+        self.gridded = get_2d_aggregate(self)
```
jlstevens: Nice to see how much ``HeatMap`` has been simplified!

holoviews/element/raster.py
```diff
@@ -383,85 +381,18 @@ class HeatMap(Dataset, Element2D):
 
     vdims = param.List(default=[Dimension('z')])
 
-    def __init__(self, data, extents=None, **params):
+    depth = 1
```
jlstevens: I might have forgotten...what is this ``depth`` class attribute?

holoviews/element/raster.py
jlstevens: That said, it isn't immediately obvious that ``gridded`` is now a ``Dataset``. Not sure I am necessarily recommending changing the name as ``gridded_dataset`` is awkward...

holoviews/plotting/bokeh/raster.py
```diff
@@ -130,26 +136,31 @@ class HeatmapPlot(ColorbarPlot):
     def _axes_props(self, plots, subplots, element, ranges):
         dims = element.dimensions()
         labels = self._get_axis_labels(dims)
-        xvals, yvals = [element.dimension_values(i, False)
+        agg = element.gridded
+        xvals, yvals = [unique_array(agg.dimension_values(i, False))
```
jlstevens: I thought gridded ``Datasets`` have the 1D coordinate arrays available. Is the uniqueness being applied over the 2D set of samples or the 1D sequence?

holoviews/plotting/bokeh/raster.py
```diff
@@ -130,26 +136,31 @@ class HeatmapPlot(ColorbarPlot):
...
-            zvals = np.rot90(element.raster, 3).flatten()
-            xvals, yvals = [[str(v) for v in element.dimension_values(i)]
+            zvals = aggregate.dimension_values(z)
+            xvals, yvals = [[str(v) for v in aggregate.dimension_values(i)]
                             for i in range(2)]
             data = {x: xvals, y: yvals, z: zvals}
 
+        if 'hover' in self.tools+self.default_tools:
+            for vdim in element.vdims[1:]:
+                data[vdim.name] = ['' if is_nan(v) else v
```
jlstevens: Wondering if an empty string really suggests NaN. 'NaN' would be explicit but might look noisy.

holoviews/plotting/bokeh/raster.py
philippjfr: Yes, good point, no longer any need for the unique_array here.

holoviews/plotting/bokeh/raster.py
philippjfr: Good point, I'm now using masked arrays to represent the data, in matplotlib the NaNs are therefore represented by ``-``, which might be better.

holoviews/plotting/mpl/raster.py
```diff
@@ -156,18 +153,25 @@ def init_artists(self, ax, plot_args, plot_kwargs):
...
         _, style, axis_kwargs = super(HeatMapPlot, self).get_data(element, ranges, style)
-        mask = np.logical_not(np.isfinite(element.raster))
-        data = np.ma.array(element.raster, mask=mask)
-        style['annotations'] = self._annotate_values(element)
+        aggregate = element.gridded
+        data = np.flipud(aggregate.dimension_values(2, flat=False))
+        shape = data.shape
+        cmap_name = style.pop('cmap', None)
+        cmap = copy.copy(plt.cm.get_cmap('gray' if cmap_name is None else cmap_name))
+        cmap.set_bad('w', 1.)
```
jlstevens: Might want to make this a plot option at some point instead of hard coding 'w'.

holoviews/plotting/bokeh/raster.py
jlstevens: Yes, I think ``-`` might be a good compromise.

holoviews/plotting/mpl/raster.py
philippjfr: Again good point, indeed we already expose this via ``clipping_colors``, should hook that in here.

holoviews/element/raster.py
philippjfr: I think this may be wrong now, will have to look into it.

holoviews/plotting/mpl/raster.py
jlstevens: I also find it curious that you are using ``copy.copy`` on a colormap - which suggests you are mutating it. I guess ``set_bad`` must have side-effects which explains the copying...

jlstevens:
Ok, I've made my comments for now (and you have already replied to most of them). The biggest suggestion is that ``get_2d_aggregate`` might be better expressed as an operation (if that makes sense).

philippjfr:
>The biggest suggestion is that get_2d_aggregate might be better expressed as an operation (if that makes sense).

Yes, ``get_2d_aggregate`` is basically a fairly crude approximation of datashader aggregation for categorical key dimensions (i.e. 2D aggregation without the binning).

holoviews/element/raster.py
philippjfr: Wasn't needed at all in the end, removed it.

philippjfr:
@jlstevens Once tests are passing this is ready for a second review and then merge.

philippjfr:
@jlstevens Tests now passing on the PR build.

holoviews/core/util.py
```diff
@@ -647,6 +647,30 @@ def walk_depth_first(name):
                                     (names_by_level.get(i, None)
                                      for i in itertools.count())))
 
+
+def is_cyclic(graph):
+    """Return True if the directed graph g has a cycle."""
```
jlstevens: What is the representation of the graph? A list of edges as tuples? Would be good to mention in the docstring.

jlstevens: I'm guessing the representation is similar as in ``one_to_one``...even so, probably worth mentioning..

holoviews/element/util.py
```diff
@@ -30,3 +44,98 @@ def compute_edges(edges):
...
+    Aggregation function to get the first non-zero value.
+    """
+    values = x.values if pd and isinstance(x, pd.Series) else x
+    for v in values:
+        if not is_nan(v):
+            return v
+    return np.NaN
+
+
+class categorical_aggregate2d(ElementOperation):
```
jlstevens: Looks great! I was just wondering if you want to keep this class in ``util`` or move it to ``operations.element``?

holoviews/element/util.py
```diff
@@ -30,3 +44,98 @@ def compute_edges(edges):
...
+
+    datatype = param.List(['xarray', 'grid'] if xr else ['grid'], doc="""
+        The grid interface types to use when constructing the gridded Dataset.""")
+
+    def _process(self, obj, key=None):
+        """
+        Generates a categorical 2D aggregate by inserting NaNs at all
+        cross-product locations that do not already have a value assigned.
+        Returns a 2D gridded Dataset object.
+        """
```
jlstevens: Quite a long method...if you see chunks that could be split up into helper methods, that might be sensible. Up to you though!

jlstevens:
I made three more comments for you to reply to. Otherwise looks good and I expect this will be merged very soon!

holoviews/element/util.py
philippjfr: It's imported there but can't be moved, cyclical imports again.

holoviews/core/util.py
philippjfr: Right, all three methods here (``sort_topologically``, ``cyclical`` and ``one_to_one``) use the same representation, which is mapping between nodes and edges, will add the docstring.

holoviews/element/util.py
philippjfr: Happy to split it up.

holoviews/element/util.py
jlstevens: Ok, having it available for ``operation.element`` is fine.

philippjfr:
Latest comments addressed, ready to merge when tests pass.

jlstevens:
Tests passed. Merging!

PR #850, opened 2016-09-07
Removed caching of Table values

jlstevens:
This PR allows Tables to be updated with streams by disabling the caching of cell values in TablePlot. I've tested that streams now work but I won't be confident that everything is working as it should until the tests pass.

### DIFF
--- a/holoviews/plotting/mpl/tabular.py
+++ b/holoviews/plotting/mpl/tabular.py
@@ -48,7 +48,7 @@ class TablePlot(ElementPlot):
 
     def __init__(self, table, **params):
         super(TablePlot, self).__init__(table, **params)
-        self.cell_values, self.cell_widths = self._format_table()
+        self.cell_widths = self._format_table()
 
 
     def _format_table(self):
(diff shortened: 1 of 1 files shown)


jlstevens:
One consequence of using streams is that it isn't possible for ``TablePlot`` to adjust the size of the table based on future (unknown) values. This will always be an issue using ``DynamicMap`` so I only want to point out this problem as I'm not sure how it can be solved.

holoviews/plotting/mpl/tabular.py
```diff
@@ -117,7 +115,7 @@ def initialize_plot(self, ranges=None):
             for col in range(element.cols):
                 if summarize and row > half_rows:
                     adjusted_row = (element.rows - self.max_rows + row)
-                cell_value = self.cell_values[self.keys[-1]][(row, col)]
+                cell_value = element.pprint_cell(row, col)
```
philippjfr: Shouldn't both of these calls also pass the return value through the ``pprint_value`` method on TablePlot, I'm not sure why both the element and the plot need a pprint method but I assume there was a reason.

jlstevens: Well spotted! I've just pushed a commit to fix this.

philippjfr:
Looks good now. Happy to merge when tests pass.

PR #851, opened 2016-09-07
Fixed updating of bokeh tables

philippjfr:
Ensures the bokeh TablePlot returns the appropriate plot handles to update the table in HoloMaps/DynamicMaps.


jlstevens:
Great!

With this, streams have been tested and shown to work for tables both with matplotlib and bokeh.

Merging.

PR #852, opened 2016-09-07
Applying dimension formatters to Bokeh Table values

jlstevens:
This PR fixes the value formatting in Bokeh Tables that were previously unformatted.


jlstevens:
This PR links to #853 which is about making tables in bokeh more consistent with matplotlib.

philippjfr:
Looks good. We'll revisit once we've decided what to do about table indexing.

PR #855, opened 2016-09-09
Fix inverted coords on iris interface

philippjfr:
The iris interface does not correctly canonicalize the array orientation in all cases, causing the plot to be inverted. Since the tests still seem to be passing there appears to be a hole in the tests that I'll patch before this is ready to merge.


jlstevens:
Tests passed. Merging.

PR #856, opened 2016-09-11
Fixed python3 bug in GenericOverlayPlot

philippjfr:
Simple python3 bug causing batched plotting to fail in python3.


jlstevens:
Looks like a very straightforward fix and I'll merge as soon as the tests pass. 

My only other suggestion is to write some unit tests so we can detect such problems earlier in future.

PR #858, opened 2016-09-12
DynamicMap with streams force callback on first lookup after refresh

philippjfr:
Currently the plotting code will request a particular key from a single DynamicMap multiple times during plotting which should be avoided due to potentially random state.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -710,7 +710,7 @@ class DynamicMap(HoloMap):
         if self.mode == 'open' and (self.counter % self.cache_interval)!=0:
             return
         if len(self) >= cache_size:
-            first_key = next(self.data.iterkeys())
+            first_key = next(k for k in self.data)
             self.data.pop(first_key)
         self.data[key] = val
 
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -477,6 +477,7 @@ class DynamicMap(HoloMap):
 
         self.call_mode = self._validate_mode()
         self.mode = 'bounded' if self.call_mode == 'key' else 'open'
+        self._stream_cache_lookup = False
 
 
     def _initial_key(self):
@@ -672,7 +673,8 @@ class DynamicMap(HoloMap):
 
         # Cache lookup
         try:
-            if util.dimensionless_contents(self.streams, self.kdims):
+            if (util.dimensionless_contents(self.streams, self.kdims) and
+                not self._stream_cache_lookup):
                 raise KeyError('Using dimensionless streams disables DynamicMap cache')
             cache = super(DynamicMap,self).__getitem__(key)
             # Return selected cache items in a new DynamicMap
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -477,7 +477,7 @@ class DynamicMap(HoloMap):
 
         self.call_mode = self._validate_mode()
         self.mode = 'bounded' if self.call_mode == 'key' else 'open'
-        self._stream_cache_lookup = False
+        self._dimensionless_cache = False
 
 
     def _initial_key(self):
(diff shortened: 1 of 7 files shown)


holoviews/core/spaces.py
```diff
@@ -672,7 +673,8 @@ def __getitem__(self, key):
 
         # Cache lookup
         try:
-            if util.dimensionless_contents(self.streams, self.kdims):
+            if (util.dimensionless_contents(self.streams, self.kdims) and
+                not self._stream_cache_lookup):
                 raise KeyError('Using dimensionless streams disables DynamicMap cache')
```
jlstevens: From our discussion, we think it ought to be something like:

```
if (util.dimensionless_contents(self.streams, self.kdims)
                  raise KeyError('Using dimensionless streams disables DynamicMap cache')
elif self._disable_cache_lookup:
                  raise KeyError('Cache lookup disabled')
```

That way the variable isn't about streams and you can disable cache lookup on dynamic maps not matter how they are (or aren't) using streams.

holoviews/core/util.py
```diff
@@ -787,7 +787,7 @@ def stream_parameters(streams, no_duplicates=True, exclude=['name']):
     If no_duplicates is enabled, a KeyError will be raised if there are
     parameter name clashes across the streams.
     """
-    param_groups = [s.params().keys() for s in streams]
+    param_groups = [s.contents.keys() for s in streams]
     names = [name for group in param_groups for name in group]
```
jlstevens: Thanks for the fix!

holoviews/core/traversal.py
```diff
@@ -128,3 +128,29 @@ def hierarchical(keys):
                 store1[v2].append(v1)
         hierarchies.append(store2 if hierarchy else {})
     return hierarchies
+
+
+class enable_streams_cache(object):
+    """
```
jlstevens: Given the discussion about, I would call it ``disable_cache_lookup`` with constructor:
```
def __init__(self, obj):
```

I don't think it needs an ``enable`` or ``disable`` flag as ``enable`` is the default behaviour.  Unless you can think of situations where you want to be able to disable the context manager with a boolean? If you can think of such a situation (or just think it makes it more flexible) then

```
def __init__(self, obj, disable=True):
```

Would be reasonable. The issue with using either ``enable`` or ``disable`` is it isn't clear if you are enabling/disabling the cache lookup of enabling/disabling the context manager that is enabling/disabling the cache lookup! 

For this reason, maybe something like this would be clearer:

```
def __init__(self, obj, allow_cache_lookup=False):
```

Just some things to think about...

holoviews/plotting/plot.py
```diff
@@ -258,6 +260,15 @@ def traverse(self, fn=None, specs=None, full_breadth=True):
...
 
+    @property
+    def force(self):
+        return self._force
+
+
+    @force.setter
+    def force(self, value):
+        self.traverse(lambda x: setattr(x, '_force', value))
+
```
jlstevens: From what I see below, I think having ``self.force`` be an attribute without ``self._force`` would be simpler (i.e no properties). The reason is that in the context_manager you just need to access``self.force``.  Then you set ``self.force`` to True and False in one place in this file. As a result, I'm not sure it is worth introducing properties and an underscore attribute when you can just inline two fairly short traversal expressions.

holoviews/core/spaces.py
philippjfr: I think we decided in the end that it is about streams.

holoviews/core/traversal.py
philippjfr: >I don't think it needs an enable or disable flag as enable is the default behaviour. Unless you can think of situations where you want to be able to disable the context manager with a boolean? If you can think of such a situation (or just think it makes it more flexible) then

The issue is that making context manager conditional is awkward, which is why I moved the boolean onto the object. I agree with renaming the argument and making it optional though.

holoviews/plotting/plot.py
philippjfr: We decided on replacing this with a ``traverse_setter`` utility.

jlstevens:
All looks good now! Happy to merge once the tests pass.

jlstevens:
Tests are passing. Merging.

PR #861, opened 2016-09-13
Bokeh colorbars

philippjfr:
Since version 0.12.2 bokeh finally support colorbars and client-side colormapping for all kinds of glyphs. This PR adds a ColorbarPlot mixin class which handles the creation of a Colorbar and a ColorMapper. Additionally it replaces the manual colormapping that was previously done in Python for various plot types. I have also moved the toolbar to the top so it does not overlap with a colorbar by default and because it is more consistent.

To Do:

* [x] Test each changed Element in more detail
* [x] Implement color_mapper range updating
* [x] Update HeatMapPlot in #849 once this has been merged.
* [x] Add some unit tests to ensure colorbars get instantiated correctly

Examples:

Different ``colorbar_position``s:

screen shot 2016-09-14 at 12 54 54 am

Log colormapping on points:

screen shot 2016-09-14 at 12 45 16 am

QuadMesh:

screen shot 2016-09-14 at 12 45 48 am

Polygons (more sensible on a Choropleth):

screen shot 2016-09-14 at 12 43 54 am

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -308,7 +308,7 @@ class ErrorPlot(PathPlot):
         return (data, dict(self._mapping))
 
 
-class SpikesPlot(PathPlot):
+class SpikesPlot(PathPlot, ColorbarPlot):
 
     color_index = param.ClassSelector(default=1, class_=(basestring, int), doc="""
       Index of the dimension from which the color will the drawn""")
@@ -346,22 +346,14 @@ class SpikesPlot(PathPlot):
             xs, ys = zip(*(((x[0], x[0]), (pos+height, pos))
                            for x in element.array(dims[:1])))
 
-        if not empty and self.invert_axes: keys = keys[::-1]
+        if not empty and self.invert_axes: xs, ys = ys, xs
         data = dict(zip(('xs', 'ys'), (xs, ys)))
-
-        cmap = style.get('palette', style.get('cmap', None))
         cdim = element.get_dimension(self.color_index)
-        if cdim and cmap:
-            map_key = 'color_' + cdim.name
-            mapping['color'] = map_key
-            if empty:
-                colors = []
-            else:
-                cmap = get_cmap(cmap)
-                cvals = element.dimension_values(cdim)
-                crange = ranges.get(cdim.name, None)
-                colors = map_colors(cvals, crange, cmap)
-            data[map_key] = colors
+        if cdim:
+            mapper = self._get_colormapper(cdim, element, ranges, style)
+            data[cdim.name] = [] if empty else element.dimension_values(cdim)
+            mapping['color'] = {'field': cdim.name,
+                                'transform': mapper}
 
         if 'hover' in self.tools+self.default_tools and not empty:
             for d in dims:
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -12,12 +12,12 @@ from ...element import Raster, Points, Polygons, Spikes
 from ...core.util import max_range, basestring, dimension_sanitizer
 from ...core.options import abbreviated_exception
 from ..util import compute_sizes, get_sideplot_ranges, match_spec, map_colors
-from .element import ElementPlot, line_properties, fill_properties
+from .element import ElementPlot, ColorbarPlot, line_properties, fill_properties
 from .path import PathPlot, PolygonPlot
 from .util import get_cmap, mpl_to_bokeh, update_plot, rgb2hex, bokeh_version
 
 
-class PointPlot(ElementPlot):
+class PointPlot(ColorbarPlot):
 
     color_index = param.ClassSelector(default=3, class_=(basestring, int),
                                       allow_None=True, doc="""
@@ -55,21 +55,12 @@ class PointPlot(ElementPlot):
         mapping = dict(x=dims[xidx], y=dims[yidx])
         data = {}
 
-        cmap = style.get('palette', style.get('cmap', None))
         cdim = element.get_dimension(self.color_index)
-        if cdim and cmap:
-            map_key = 'color_' + cdim.name
-            mapping['color'] = map_key
-            if empty:
-                data[map_key] = []
-            else:
-                cmap = get_cmap(cmap)
-                colors = element.dimension_values(self.color_index)
-                if colors.dtype.kind in 'if':
-                    crange = ranges.get(cdim.name, element.range(cdim.name))
-                else:
-                    crange = np.unique(colors)
-                data[map_key] = map_colors(colors, crange, cmap)
+        if cdim:
+            mapper = self._get_colormapper(cdim, element, ranges, style)
+            data[cdim.name] = [] if empty else element.dimension_values(cdim)
+            mapping['color'] = {'field': cdim.name,
+                                'transform': mapper}
 
         sdim = element.get_dimension(self.size_index)
         if sdim:
@@ -98,7 +89,7 @@ class PointPlot(ElementPlot):
             eldata, elmapping = self.get_data(el, ranges, empty)
             for k, eld in eldata.items():
                 data[k].append(eld)
-            if 'color' not in eldata:
+            if 'color' not in elmapping:
                 zorder = self.get_zorder(element, key, el)
                 val = style[zorder].get('color')
                 elmapping['color'] = 'color'
@@ -128,6 +119,8 @@ class PointPlot(ElementPlot):
         else:
             plot_method = self._plot_methods.get('batched' if self.batched else 'single')
             renderer = getattr(plot, plot_method)(**dict(properties, **mapping))
+        if self.colorbar and 'color_mapper' in self.handles:
+            self._draw_colorbar(plot, self.handles['color_mapper'])
         return renderer, renderer.glyph
 
 
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -232,7 +232,7 @@ class HistogramPlot(ElementPlot):
         return (data, mapping)
 
 
-class SideHistogramPlot(HistogramPlot):
+class SideHistogramPlot(HistogramPlot, ColorbarPlot):
 
     style_opts = HistogramPlot.style_opts + ['cmap']
 
@@ -255,19 +255,20 @@ class SideHistogramPlot(HistogramPlot):
             data = dict(top=element.values, left=element.edges[:-1],
                         right=element.edges[1:])
 
-        dim = element.get_dimension(0).name
+        dim = element.get_dimension(0)
         main = self.adjoined.main
-        range_item, main_range, dim = get_sideplot_ranges(self, element, main, ranges)
-        vals = element.dimension_values(dim)
+        range_item, main_range, _ = get_sideplot_ranges(self, element, main, ranges)
         if isinstance(range_item, (Raster, Points, Polygons, Spikes)):
             style = self.lookup_options(range_item, 'style')[self.cyclic_index]
         else:
             style = {}
 
         if 'cmap' in style or 'palette' in style:
-            cmap = get_cmap(style.get('cmap', style.get('palette', None)))
-            data['color'] = [] if empty else map_colors(vals, main_range, cmap)
-            mapping['fill_color'] = 'color'
+            main_range = {dim.name: main_range}
+            mapper = self._get_colormapper(dim, element, main_range, style)
+            data[dim.name] = [] if empty else element.dimension_values(dim)
+            mapping['fill_color'] = {'field': dim.name,
+                                     'transform': mapper}
         self._get_hover_data(data, element, empty)
         return (data, mapping)
 
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -265,10 +265,10 @@ class SideHistogramPlot(HistogramPlot, ColorbarPlot):
 
         if 'cmap' in style or 'palette' in style:
             main_range = {dim.name: main_range}
-            mapper = self._get_colormapper(dim, element, main_range, style)
+            cmapper = self._get_colormapper(dim, element, main_range, style)
             data[dim.name] = [] if empty else element.dimension_values(dim)
             mapping['fill_color'] = {'field': dim.name,
-                                     'transform': mapper}
+                                     'transform': cmapper}
         self._get_hover_data(data, element, empty)
         return (data, mapping)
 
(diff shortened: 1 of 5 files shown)


jbednar:
Looks great!

jlstevens-datacamp:
I agree it looks great! 

The only thing that is bothering me slightly is that I would expect a black frame around the colorbars, especially if you think about colormaps that end in white (like the last two examples above).

jlstevens:
I agree it looks great! The only thing that is bothering me slightly is that I would expect a black frame around the colorbars, especially if you consider colormaps ending in white (like the last two examples shown above).

philippjfr:
>The only thing that is bothering me slightly is that I would expect a black frame around the colorbars, especially if you consider colormaps ending in white (like the last two examples shown above).

I agree, it's a design choice the bokeh team decided on but I think enabling a black border makes sense.

holoviews/plotting/bokeh/element.py
```diff
@@ -675,7 +675,9 @@ def _draw_colorbar(self, plot, color_mapper):
             ticker = LogTicker()
         else:
             ticker = BasicTicker()
-        cbar_opts = self.colorbar_specs[self.colorbar_position]
+        cbar_opts = dict(self.colorbar_specs[self.colorbar_position],
+                         bar_line_color='black', label_standoff=8,
+                         major_tick_line_color='black')
```
jlstevens: Great! Glad there is an option for this...

philippjfr:
Ready for review.

jbednar:
Are there issues on the Bokeh site already where I can go to request that there be a tiny, tiny, black outline (since colorbars starting or ending in white really are not uncommon, e.g. grayscale), and to plead to have the plot not change shape when a colorbar is added?  Both of those are quite important for usability!

jbednar:
Doesn't moving the toolbar to the top conflict with the title now?  I thought that's why the tools moved to the right.

philippjfr:
>Are there issues on the Bokeh site already where I can go to request that there be a tiny, tiny, black outline (since colorbars starting or ending in white really are not uncommon, e.g. grayscale)

I've added black outlines by default in holoviews.

> and to plead to have the plot not change shape when a colorbar is added?

No way to do this although I could approximate the percentage of the width of the colorbar and offset, not sure I want to go down that route though.

>Doesn't moving the toolbar to the top conflict with the title now? I thought that's why the tools moved to the right.

Yes, sorry should have updated my description, the toolbar location is now configurable with the toolbar plot option which accepts ['above', 'below', 'left', 'right', None]. Should I change above and below to top and bottom for consistency?

holoviews/plotting/bokeh/element.py
```diff
@@ -618,6 +638,138 @@ def framewise(self):
...
+        color_bar = ColorBar(color_mapper=color_mapper, ticker=ticker,
+                             **dict(cbar_opts['opts'], **self.colorbar_opts))
+
+        plot.add_layout(color_bar, pos)
+        self.handles['colorbar'] = color_bar
+
+
+    def _get_colormapper(self, dim, element, ranges, style):
+        low, high = ranges.get(dim.name)
+        palette = mplcmap_to_palette(style.pop('cmap', 'viridis'))
```
jlstevens: Not sure I am happy about hard coding 'viridis' as a default, especially as it is a matplotlib colormap (unless bokeh now has viridis?).

jbednar: Bokeh does now have viridis.

holoviews/plotting/bokeh/element.py
```diff
@@ -104,6 +113,13 @@ class ElementPlot(BokehPlot, GenericElementPlot):
     tools = param.List(default=[], doc="""
         A list of plugin tools to use on the plot.""")
 
+    toolbar = param.ObjectSelector(default='right',
+                                   objects=["above", "below",
+                                            "left", "right", None],
+                                   doc="""
+        The toolbar location, must be one of 'above', 'below',
+        'left', 'right', None.""")
```
jlstevens: I assume ``None`` can be a sensible choice even if the user has added tools (e.g to temporarily hide the toolbar). Otherwise, adding tools with toolbar=None could issue a warning...

jbednar:
above and below could be top and bottom; sure.

holoviews/plotting/bokeh/chart.py
```diff
@@ -55,21 +55,12 @@ def get_data(self, element, ranges=None, empty=False):
...
-                if colors.dtype.kind in 'if':
-                    crange = ranges.get(cdim.name, element.range(cdim.name))
-                else:
-                    crange = np.unique(colors)
-                data[map_key] = map_colors(colors, crange, cmap)
+        if cdim:
+            mapper = self._get_colormapper(cdim, element, ranges, style)
+            data[cdim.name] = [] if empty else element.dimension_values(cdim)
+            mapping['color'] = {'field': cdim.name,
+                                'transform': mapper}
```
jlstevens: Looks much cleaner!

holoviews/plotting/bokeh/element.py
philippjfr: Matplotlib has hardcoded defaults, which is viridis now and bokeh does have viridis now as well.

holoviews/plotting/bokeh/chart.py
```diff
@@ -128,6 +119,8 @@ def _init_glyph(self, plot, mapping, properties):
         else:
             plot_method = self._plot_methods.get('batched' if self.batched else 'single')
             renderer = getattr(plot, plot_method)(**dict(properties, **mapping))
+        if self.colorbar and 'color_mapper' in self.handles:
+            self._draw_colorbar(plot, self.handles['color_mapper'])
         return renderer, renderer.glyph
```
jlstevens: When could you request a colorbar but not have a color_mapper available?

holoviews/plotting/bokeh/element.py
jlstevens: Great!

holoviews/plotting/bokeh/chart.py
philippjfr: They might have enabled colorbar by default but not set a ``color_index``.

holoviews/plotting/bokeh/element.py
jbednar: Personally, I like that viridis is perceptually uniform, but I don't actually like it overall, because it has no intuitive ordering.  Hot is clearly ordered in a way that people can appreciate at first glance, as are cool colormaps (black->blue->white), but viridis just has to be memorized.  In that sense its as bad as jet (but only that sense).

holoviews/plotting/bokeh/element.py
```diff
@@ -618,6 +638,138 @@ def framewise(self):
...
+
+
+    def _get_colormapper(self, dim, element, ranges, style):
+        low, high = ranges.get(dim.name)
+        palette = mplcmap_to_palette(style.pop('cmap', 'viridis'))
+        colormapper = LogColorMapper if self.logz else LinearColorMapper
+        cmapper = colormapper(palette, low=low, high=high)
+        if 'color_mapper' not in self.handles:
+            self.handles['color_mapper'] = cmapper
+        return cmapper
```
jlstevens: I can see the logic of returning ``cmapper`` but this function could also just be called for the side-effect of adding 'color_mapper' to the handles. I had to check the code here to make sure it is indeed the same thing...

holoviews/plotting/bokeh/raster.py
```diff
@@ -176,24 +150,23 @@ class QuadMeshPlot(ElementPlot):
...
             xs, ys = cartesian_product([xvals, yvals])
             ws, hs = cartesian_product([widths, heights])
-            data = {x: xs.flat, y: ys.flat, z: zvals, 'color': colors,
+            data = {x: xs.flat, y: ys.flat, z: zvals,
                     'widths': ws.flat, 'heights': hs.flat}
 
-        return (data, {'x': x, 'y': y, 'fill_color': 'color',
+        return (data, {'x': x, 'y': y,
+                       'fill_color': {'field': z, 'transform': cmapper},
                        'height': 'heights', 'width': 'widths'})
```
jlstevens: Here you call it ``cmapper`` but further down you sometimes call it ``mapper``. I prefer the former....

holoviews/plotting/bokeh/element.py
philippjfr: It's just a convenient way to disable the toolbar altogether. I could either disable it, or warn ``tools`` or warn for ``tools`` and ``default_tools``.

holoviews/plotting/bokeh/element.py
```diff
@@ -618,6 +638,138 @@ def framewise(self):
...
+            ticker = BasicTicker()
+        cbar_opts = dict(self.colorbar_specs[self.colorbar_position])
+
+        # Check if there is a colorbar in the same position
+        pos = cbar_opts['pos']
+        if any(isinstance(model, ColorBar) for model in getattr(plot, pos, [])):
+            return
+
+        opts = dict(cbar_opts['opts'], bar_line_color='black',
+                    label_standoff=8, major_tick_line_color='black')
```
jlstevens: I wouldn't necessarily make these parameters but maybe they could be specified as a dictionary in the class attributes. Alternatively, you could just make them into the default for ``colorbar_opts`` which might make more sense...

holoviews/plotting/bokeh/raster.py
```diff
@@ -56,28 +46,12 @@ def _glyph_properties(self, plot, element, source, ranges):
...
-            palette = mplcmap_to_palette(properties.pop('cmap', None))
-        colormapper = LogColorMapper if self.logz else LinearColorMapper
-        cmap = colormapper(palette, low=low, high=high)
-        properties['color_mapper'] = cmap
-        if 'color_mapper' not in self.handles:
-            self.handles['color_mapper'] = cmap
+        val_dim = [d for d in element.vdims][0]
+        properties['color_mapper'] = self._get_colormapper(val_dim, element, ranges,
+                                                           properties)
         return properties
```
jlstevens: I'm just wondering why in this case ``self._get_colormapper`` is used to add the color mapper to the properties (in ``_glyph_properties``) but is added to the data in ``get_data`` everywhere else...

holoviews/plotting/bokeh/chart.py
jlstevens: Ok, makes sense, thanks.

holoviews/plotting/bokeh/raster.py
philippjfr: The image glyph accepts a colormapper directly while in other cases it is used as a transform that maps color to a particular column.

holoviews/plotting/bokeh/element.py
jlstevens: I think being able to disable the toolbar without issuing warnings could be convenient so I am happy to leave it as it is.

holoviews/plotting/bokeh/element.py
jlstevens: Isn't hot what we decided is the holoviews default colormap? I agree that viridis has some nice properties but being intuitive isn't necessarily one of them!

holoviews/plotting/bokeh/element.py
philippjfr: When you add them to the ``colorbar_opts`` they all get overridden when the user supplies new options and ``bar_line_color='black'`` just looks silly without ``major_tick_line_color='black'``, so I wanted them to behave like defaults. Happy to make them into a class attribute though.

holoviews/plotting/bokeh/raster.py
jlstevens: Thanks for the explanation!

holoviews/plotting/bokeh/element.py
jlstevens: A class attribute would be good in that case.

holoviews/plotting/bokeh/element.py
philippjfr: Yes, but if you forget to set a default via the options system it will fall back to what's in the matplotlib rc, which will be viridis in matplotlib >1.5. Also hot isn't included in bokeh afaik.

jlstevens: That makes sense but also points to a possible annoyance. Once we make the bokeh backend work without matplotlib entirely, then I guess 'hot' will be our default colormap for matplotlib and 'viridis' will be the default for bokeh. Given the change in matplotlib > 1.5,  we could use that as an excuse to make viridis our default colormap (given it is supported by matplotlib and bokeh) but I also prefer hot for the reason Jim mentions above!

holoviews/plotting/bokeh/element.py
philippjfr: Yes, there's some annoyances I have with this, ideally the handles would be set elsewhere but since it is now called from get_data, which is the main method the user has to implement and is duplicated everywhere I didn't want to move it out. The other thing is only the first colormapper that's created is actually used while the rest are simply used to update the existing colormapper. I'll leave it up to you if you think I should find a cleaner solution otherwise I'll make sure to leave a comment.

holoviews/plotting/bokeh/element.py
jbednar: It is easy to provide a Hot for bokeh:

Hot = ["black", "maroon", "darkred", "red", "orangered", "darkorange", "orange", "gold", "yellow", "white"]

I actually prefer this definition to the one in mpl, because it's more perceptually uniform; the one in mpl makes all the top few values be indistinguishably white on the monitors I have tested.  But it is slightly different than mpl's hot.

We could take mpl's hot and make that available for Bokeh easily by just copying the values from matplotlib/_cm_listed.py and converting them to a palette using `[(int(r*256),int(g*256),int(b*256)) for r,g,b in cm_data]`.  The advantage of doing this is that "hot" will mean the same thing in mpl as in HoloViews, and there's no existing "hot" in Bokeh to conflict.  So this way everyone would agree about the meaning of "hot".

But we can also have our own hot as listed above, perhaps renamed, to use as our default on both backends, with the advantage of being more perceptually uniform than mpl's hot, and being our own thing that won't have to be confused (at least not by name) with mpl's.  But because it's only *more* perceptually uniform, and isn't truly uniform, I'm not sure if that's the right thing to do.

jbednar: Plus that would be a (very slight) change to the current defaults.  Probably few people besides me could tell the difference, though.

jlstevens: And we would have to come up with a name for it! :-)

Definitely good suggestions though and something we should consider doing for 1.7.

jbednar: Mpl's hot definitely has regions where it plateaus and conveys little information about changes in value:

image

It looks good to me up until it turns yellow, then it's got a huge yellow stretch with little change as intensity varies.  It wouldn't be hard to do a better job, and I'd be happy to do that.

jbednar: Looks like mpl's hot came from matlab originally? http://ab-initio.mit.edu/wiki/index.php/Color_tables_in_h5topng

jbednar: > And we would have to come up with a name for it!

Mpl has hot, afmhot, and gist_heat, so ours could be hhot. :-)

jlstevens:
Tests are passing and I understand this PR is complete. Merging.

jbednar:
The colorbars above are squishing the plots when they are added, which is an issue I just filed for bokeh: https://github.com/bokeh/bokeh/issues/5186

PR #867, opened 2016-09-18
matplotlib2.0.0b4 compatibility fix (1/2)

stonebig:
https://github.com/ioam/holoviews/issues/866

### DIFF
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -2,6 +2,7 @@ import math
 
 import param
 import numpy as np
+import matplotlib as mpl
 import matplotlib.pyplot as plt
 from matplotlib import ticker
 from matplotlib import colors
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -15,6 +15,7 @@ from ..plot import GenericElementPlot, GenericOverlayPlot
 from ..util import dynamic_update
 from .plot import MPLPlot
 from .util import wrap_formatter
+from distutils.version import LooseVersion
 
 
 class ElementPlot(GenericElementPlot, MPLPlot):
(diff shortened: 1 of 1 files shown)


jlstevens:
Requested change looks fine to me. I'll let @philippjfr decide whether it can be merged right away...

holoviews/plotting/mpl/element.py
```diff
@@ -15,6 +16,7 @@
 from ..util import dynamic_update
 from .plot import MPLPlot
 from .util import wrap_formatter
+from distutils.version import LooseVersion
```
jlstevens: I'm not sure we should assume distutils is available (which might explain the current test failures).

holoviews/plotting/mpl/element.py
```diff
@@ -149,8 +151,11 @@ def _finalize_axis(self, key, title=None, dimensions=None, ranges=None, xticks=N
         subplots = list(self.subplots.values()) if self.subplots else []
         if self.zorder == 0 and key is not None:
             if self.bgcolor:
-                axis.set_axis_bgcolor(self.bgcolor)
-
+                if LooseVersion(mpl.__version__) <= '1.5.9':
+                    axis.set_axis_bgcolor(self.bgcolor)
```
jlstevens: To avoid relying on ``LooseVersion``, maybe we could check that ``2`` is the major version instead? i.e something like ``if mpl.__version__[0] == '2':``? That would also make it clear that the change is occurring in the matplotlib 2+.

philippjfr:
Distutils is in the standard library no? We also use LooseVersion in several places already. The test failures are down to it trying to import PyQt4 for some reason. Something going wrong setting the agg backend?

jlstevens:
Looks like distutils is in the standard library so I have no idea why this PR would make the tests fail.

jlstevens:
Hmm, looks like it is currently a problem on master too: https://travis-ci.org/ioam/holoviews

philippjfr:
Test failures are unrelated and this fix looks fine. Going to merge now and then fix the tests separately.

stonebig:
in the mean time bryevdv suggests to do it a different way`(for bokeh similar fix), avoiding to use LooseVersion:
````
if "get_facecolor" in dir(ax):
````

... maybe next time

Issue #866, opened 2016-09-18
deprecation warning with Matplotlib-2.0.0b4 / Holoviews-1.6.2


stonebig:
Hi,
I get this deprecation warning with Matplolib-2.0.0b4 Holoviews 1.6.2

````
import holoviews as hv
%load_ext holoviews.ipython
fractal = hv.Image(image)

((fractal * hv.HLine(y=0.16)).hist() + fractal.sample(y=0.16))
`````

````
C:\WinPython\basedir35\buildQt5\winpython-64bit-3.5.x.3\python-3.5.2.amd64\lib\site-packages\holoviews\plotting\mpl\element.py:152: MatplotlibDeprecationWarning: The set_axis_bgcolor function was deprecated in version 2.0. Use set_facecolor instead.
  axis.set_axis_bgcolor(self.bgcolor)
````
![holoviews-20160918a](https://cloud.githubusercontent.com/assets/4312421/18615166/78b8c284-7d9f-11e6-86fd-160df701ea5c.PNG)

stonebig:
https://github.com/matplotlib/matplotlib/blob/master/lib/matplotlib/axes/_base.py#L2717..L2726

jlstevens:
Thanks for reporting this and thanks for submitting a PR!

I haven't personally tried out the matplotlib 2.0 beta yet though it is definitely we need to test properly soon.

philippjfr:
Just merged the PRs. Thanks for making those fixes, closing.

PR #868, opened 2016-09-18
matplotlib2.0.0b4 compatibility fix (1/2)

stonebig:
#866


jlstevens:
This PR looks like it might be redundant with #867 (which seems more complete as it isn't missing some of the necessary imports).

stonebig:
it's not the same file. whith github interface, you modify on file at a time

philippjfr:
Same as #867, tests failing due to anaconda pyqt issues. Going to merge and then fix the tests with #871.

jlstevens:
> it's not the same file. whith github interface, you modify on file at a time

Ah, I see. Fairly awkward to have to build PRs that way! :-(

PR #880, opened 2016-09-23
Fixed bokeh Callbacks

philippjfr:
Updates bokeh Callbacks to correctly use comms and added comms to all subplots. The latter change will also be necessary for making streams work correctly.


jlstevens:
Ready to merge?

I believe the test failures are due to an issue that has now been fixed (on master).

philippjfr:
Yes, please do.

PR #889, opened 2016-09-28
Connect streams with bokeh callbacks

philippjfr:
This finally allows connecting Streams with interactions on a bokeh plot, which will allow for all kinds of interactive features becoming trivial to implement. The PR is currently a WIP and only works for a DynamicMap returning a single element.

The PR works by defining Callback classes which define the attributes that should be returned from the bokeh plot and the handles the callback will be attached to. The Callback then auto-generates the code to look up those attributes and sends them back via a JupyterCommJS.

The next part I will work on is how the source of a stream is looked up to make sure it also works with overlays and layouts.

### Example

Having laid out what still needs work I will outline a concrete example of what does work. The PR registers the following callback:

```python
class PositionXYCallback(Callback):

    attributes = {'x': 'cb_data.geometry.x', 'y': 'cb_data.geometry.y'}

    handles = ['hover']


callbacks = Stream._callbacks['bokeh']
callbacks[PositionXY] = PositionXYCallback
```

When creating a PositionXY stream on a DynamicMap it will now create a hovertool on the source of the stream, which sends its x and y position back to the Stream.

```python
dmap = hv.DynamicMap(lambda x, y: hv.Points([(x**i, y**i) for i in np.linspace(0, 3)],
                                            extents=(0, 0, 10, 10)),
                     kdims=[], streams=[PositionXY()])
dmap
```

xy

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -221,5 +221,5 @@ callbacks[PositionY] = PositionYCallback
 callbacks[RangeXY] = RangeXYCallback
 callbacks[RangeX] = RangeXCallback
 callbacks[RangeY] = RangeYCallback
-callbacks[Box] = BoundsCallback
+callbacks[Bounds] = BoundsCallback
 callbacks[Selection1D] = Selection1DCallback
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -53,10 +53,11 @@ class Callback(param.Parameterized):
     handles = []
     msgs=[]
 
-    def __init__(self, plot, streams, **params):
+    def __init__(self, plot, streams, source, **params):
         self.plot = plot
         self.streams = streams
         self.comm = JupyterCommJS(plot, on_msg=self.on_msg)
+        self.source = source
 
 
     def initialize(self):
@@ -91,11 +92,15 @@ class Callback(param.Parameterized):
     def set_customjs(self, cb_obj):
         # Generate callback JS code to get all the requested data
         self_callback = self.js_callback.format(comms_target=self.comm.target)
-        attributes = get_attributes(self.attributes)
+        attributes = attributes_js(self.attributes)
         code = 'var data = {};\n' + attributes + self.code + self_callback
 
+        handles = dict(self.plot.handles)
+        plots = [self.plot] + (self.plot.subplots.values()[::-1] if self.plot.subplots else [])
+        for plot in plots:
+            handles.update(plot.handles)
         # Set cb_obj
-        cb_obj.callback = CustomJS(args=self.plot.handles, code=code)
+        cb_obj.callback = CustomJS(args=handles, code=code)
 
 
 
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -93,12 +93,13 @@ class Callback(object):
 
     # The plotting handle(s) to attach the JS callback on
     handles = []
-    msgs=[]
+
+    _comm_type = JupyterCommJS
 
     def __init__(self, plot, streams, source, **params):
         self.plot = plot
         self.streams = streams
-        self.comm = JupyterCommJS(plot, on_msg=self.on_msg)
+        self.comm = self._comm_type(plot, on_msg=self.on_msg)
         self.source = source
 
 
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -60,11 +60,19 @@ class Callback(param.Parameterized):
 
 
     def initialize(self):
-        for handle in self.handles:
-            if handle not in self.plot.handles:
-                self.warning('Plotting handle for JS callback not found')
-                continue
-            self.set_customjs(self.plot.handles[handle])
+        plots = [self.plot]
+        if self.plot.subplots:
+            plots += list(self.plot.subplots.values())
+
+        found = []
+        for plot in plots:
+            for handle in self.handles:
+                if handle not in plot.handles or handle in found:
+                    continue
+                self.set_customjs(plot.handles[handle])
+                found.append(handle)
+        if len(found) != len(self.handles):
+            self.warning('Plotting handle for JS callback not found')
 
 
     def on_msg(self, msg):
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -5,7 +5,8 @@ import param
 import numpy as np
 from bokeh.models import CustomJS
 
-from ...streams import Stream, PositionXY
+from ...streams import (Stream, PositionXY, RangeXY, BoxSelect, RangeX,
+                        RangeY, PositionX, PositionY)
 from ..comms import JupyterCommJS
 
 
@@ -97,5 +98,75 @@ class PositionXYCallback(Callback):
     handles = ['hover']
 
 
+class PositionXCallback(Callback):
+
+    attributes = {'x': 'cb_data.geometry.x'}
+
+    handles = ['hover']
+
+
+class PositionYCallback(Callback):
+
+    attributes = {'y': 'cb_data.geometry.y'}
+
+    handles = ['hover']
+
+
+class RangeXYCallback(Callback):
+
+    attributes = {'x0': 'x_range.attributes.start',
+                  'x1': 'x_range.attributes.end',
+                  'y0': 'y_range.attributes.start',
+                  'y1': 'y_range.attributes.end'}
+
+    handles = ['x_range', 'y_range']
+
+    def _process_msg(self, msg):
+        return {'x_range': (msg['x0'], msg['x1']),
+                'y_range': (msg['y0'], msg['y1'])}
+
+
+class RangeXCallback(Callback):
+
+    attributes = {'x0': 'x_range.attributes.start',
+                  'x1': 'x_range.attributes.end'}
+
+    handles = ['x_range']
+
+    def _process_msg(self, msg):
+        return {'x_range': (msg['x0'], msg['x1'])}
+
+
+class RangeYCallback(Callback):
+
+    attributes = {'y0': 'y_range.attributes.start',
+                  'y1': 'y_range.attributes.end'}
+
+    handles = ['y_range']
+
+    def _process_msg(self, msg):
+        return {'y_range': (msg['y0'], msg['y1'])}
+
+
+class BoxCallback(Callback):
+
+    attributes = {'x0': 'cb_data.geometry.x0',
+                  'x1': 'cb_data.geometry.x1',
+                  'y0': 'cb_data.geometry.y0',
+                  'y1': 'cb_data.geometry.y1'}
+
+    handles = ['box_select']
+
+    def _process_msg(self, msg):
+        return {'bounds': (msg['x0'], msg['y0'], msg['x1'], msg['y1'])}
+
+
 callbacks = Stream._callbacks['bokeh']
+
 callbacks[PositionXY] = PositionXYCallback
+callbacks[PositionX] = PositionXCallback
+callbacks[PositionY] = PositionYCallback
+callbacks[RangeXY] = RangeXYCallback
+callbacks[RangeX] = RangeXCallback
+callbacks[RangeY] = RangeYCallback
+callbacks[BoxSelect] = BoxCallback
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -10,7 +10,11 @@ from ...streams import (Stream, PositionXY, RangeXY, Selection1D, RangeX,
 from ..comms import JupyterCommJS
 
 
-def get_attributes(attributes):
+def attributes_js(attributes):
+    """
+    Generates JS code to look up attributes on JS objects from
+    an attributes specification dictionary.
+    """
     code = ''
     for key, attr_path in attributes.items():
         data_assign = "data['{key}'] = ".format(key=key)
@@ -23,6 +27,34 @@ def get_attributes(attributes):
 
 
 class Callback(param.Parameterized):
+    """
+    Provides a baseclass to define callbacks, which return data from
+    bokeh models such as the plot ranges or various tools. The callback
+    then makes this data available to any streams attached to it.
+
+    The defintion of a callback consists of a number of components:
+
+    * handles    :  The handles define which object the callback will be
+                    attached on.
+
+    * attributes :  The attributes define which attributes to send back
+                    to Python. They are defined as a dictionary mapping
+                    between the name under which the variable is made
+                    available to Python and the specification of the
+                    attribute. The specification should start with the
+                    variable name that is to be accessed and the
+                    location of the attribute separated by periods.
+                    All plotting handles such as tools, the x_range,
+                    y_range and (data)source can be addressed in this
+                    way.
+    * code        : Defines any additional JS code to be executed,
+                    which can modify the data object that is sent to
+                    the backend.
+
+    The callback can also define a _process_msg method, which can
+    modify the data sent by the callback before it is passed to the
+    streams.
+    """
 
     code = param.String(default="", doc="""
         Custom javascript code executed on the callback. The code
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -14,6 +14,12 @@ def attributes_js(attributes):
     """
     Generates JS code to look up attributes on JS objects from
     an attributes specification dictionary.
+
+    Example:
+
+    Input  : {'x': 'cb_data.geometry.x'}
+
+    Output : data['x'] = cb_data['geometry']['x']
     """
     code = ''
     for key, attr_path in attributes.items():
@@ -26,7 +32,7 @@ def attributes_js(attributes):
     return code
 
 
-class Callback(param.Parameterized):
+class Callback(object):
     """
     Provides a baseclass to define callbacks, which return data from
     bokeh models such as the plot ranges or various tools. The callback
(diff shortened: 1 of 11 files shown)


jlstevens:
Fantastic!

It is probably worth noting that in the final version, users won't even need to define ``PositionXYCallback`` as we will have all those defined and registered already. This means, all that will need to be done is to import ``PositionXY`` and define the ``DynamicMap``.

I haven't reviewed the PR yet but I am assuming the mechanism that will register these callbacks for each backend (in such a way the user doesn't need to worry about it) is something we will still need to discuss.

philippjfr:
>It is probably worth noting that in the final version, users won't even need to define PositionXYCallback as we will have all those defined and registered already. This means, all that will need to be done is to import PositionXY and define the DynamicMap.

That's already the case, I just pulled it out to demonstrate how it works and how little code is involved. Currently there is a registry on ``Stream._callbacks``, indexed by backend and Stream type.

jlstevens:
Great!

I'll review the PR when you feel it is ready and have hopefully added a few more examples.

jbednar:
Excellent!  Looks really clean.  And somehow you did it by *removing* 300 lines of code.

philippjfr:
I've replaced the old callback system with this, which means some of the old downsampling callbacks should become operations. There is some worry about backward compatibility, but I'd prefer not to have two separate systems around.

jlstevens:
> There is some worry about backward compatibility, but I'd prefer not to have two separate systems around.

I agree. I don't think we need to worry too much as the old plotting callback system wasn't properly documented so I think we were probably the only people using it.

holoviews/plotting/bokeh/path.py
```diff
@@ -49,25 +49,6 @@ class PolygonPlot(ColorbarPlot, PathPlot):
     style_opts = ['color', 'cmap', 'palette'] + line_properties + fill_properties
     _plot_methods = dict(single='patches', batched='patches')
 
-    def _init_tools(self, element):
```
jlstevens: Good to get rid of this! I guess it now inherits it from the superclass.

holoviews/plotting/comms.py
```diff
@@ -119,7 +119,7 @@ def send(self, data):
 
 
 
-class JupyterCommJS(Comm):
+class JupyterCommJS(JupyterComm):
     """
```
jlstevens: The new version makes more sense...was this a bug?

philippjfr: Yep, but the class wasn't used.

holoviews/streams.py
```diff
@@ -213,18 +211,54 @@ class PositionXY(Stream):
...
+
+    bounds = param.NumericTuple(default=(0, 0, 1, 1), constant=True)
+
+
+class Selection1D(Stream):
+    """
+    A stream representing a 1D selection of objects by their index.
+    """
 
+    index = param.List(default=[])
```
jlstevens: All these new stream classes look good - they will need docstrings on their parameters though....

philippjfr: Agreed, will add them shortly.

holoviews/streams.py
```diff
@@ -184,10 +190,6 @@ class PositionX(Stream):
     x = param.Number(default=0, doc="""
            Position along the x-axis in data coordinates""", constant=True)
 
-    def __init__(self, preprocessors=[], source=None, subscribers=[], **params):
-        super(PositionX, self).__init__(preprocessors=preprocessors, source=source,
-                                        subscribers=subscribers, **params)
-
```
jlstevens: Thanks for getting rid of these unnecessary constructors!

holoviews/streams.py
```diff
@@ -121,14 +115,26 @@ def __init__(self, preprocessors=[], source=None, subscribers=[], **params):
...
+    @property
+    def source(self):
+        return self._source
+
+    @source.setter
+    def source(self, source):
+        if self._source:
+            raise Exception('source has already been defined on stream.')
+        self._source = source
+        self.registry[source].append(self)
```
jlstevens: Perhaps this should be ``id(source)``?

holoviews/streams.py
```diff
@@ -121,14 +115,26 @@ def __init__(self, preprocessors=[], source=None, subscribers=[], **params):
         datastructure that the stream receives events from, as supported
         by the plotting backend.
         """
-        self.source = source
+        self._source = source
         self.subscribers = subscribers
         self.preprocessors = preprocessors
         self._hidden_subscribers = []
 
         self.uuid = uuid.uuid4().hex
```
jlstevens: Looks like the ``uuid`` attribute might no longer b necessary...

holoviews/streams.py
```diff
@@ -75,7 +75,9 @@ class Stream(param.Parameterized):
     """
 
     # Mapping from uuid to stream instance
-    registry = OrderedDict()
+    registry = defaultdict(list)
+
+    _callbacks = defaultdict(dict)
```
philippjfr: Yea definitely need to add docstrings in this file still. Basically it defines the registry of callbacks for each backend which supply the stream information.

holoviews/streams.py
philippjfr: Will remove it.

holoviews/plotting/bokeh/util.py
```diff
@@ -193,6 +163,39 @@ def get_ids(obj):
     return list(itertools.chain(*ids))
 
 
+def replace_models(obj):
+    """
+    Processes references replacing Model and HasProps objects.
+    """
```
jlstevens: Maybe 'Recursively processes references, replacing Models with there .ref values and HasProps objects with their property values'. At least that is how I understand the code!

philippjfr: Yep, will update it.

holoviews/plotting/bokeh/element.py
```diff
@@ -55,10 +57,6 @@
 
 class ElementPlot(BokehPlot, GenericElementPlot):
 
-    callbacks = param.ClassSelector(class_=Callbacks, doc="""
-        Callbacks object defining any javascript callbacks applied
-        to the plot.""")
-
```
jlstevens: Great to get rid of this! Though we mustn't forget to mention its removal in the CHANGELOG.

holoviews/plotting/bokeh/util.py
```diff
@@ -193,6 +163,39 @@ def get_ids(obj):
...
+    else:
+        return obj
+
+
+def to_references(doc):
+    """
+    Convert the document to a dictionary of references.  Avoids
+    converting document to json and the performance penalty that
+    involves.
+    """
```
jlstevens: Perhaps 'Avoids unnecessary JSON serialization/deserialization within Python and the corresponding performance penalty'. If you don't want to use that, just make it clear that this is all within Python.

holoviews/plotting/bokeh/element.py
```diff
@@ -168,18 +166,56 @@ def __init__(self, element, plot=None, **params):
         self.current_ranges = None
         super(ElementPlot, self).__init__(element, **params)
         self.handles = {} if plot is None else self.handles['plot']
-        element_ids = self.hmap.traverse(lambda x: id(x), [Element])
```
jlstevens: Was ``element_ids`` not needed?

philippjfr: I realized that even if a HoloMap of Overlays has only a single element it's still static.

holoviews/plotting/bokeh/element.py
```diff
@@ -168,18 +166,56 @@ def __init__(self, element, plot=None, **params):
...
+        """
+        Initializes any callbacks for streams which have defined
+        the plotted object as a source.
+        """
+        if not self.static or isinstance(self.hmap, DynamicMap):
+            source = self.hmap
+        else:
+            source = self.hmap.last
+        streams = Stream.registry.get(source, [])
+        registry = Stream._callbacks['bokeh']
```
jlstevens: Ah, so ``_callbacks`` contains the backend specific code? Worth clarifying where ``_callbacks`` is defined on ``Stream`` (see comment below).

holoviews/plotting/bokeh/element.py
```diff
@@ -168,18 +166,56 @@ def __init__(self, element, plot=None, **params):
...
+        callbacks = {(registry[type(stream)], stream) for stream in streams
+                     if type(stream) in registry and streams}
+        cbs = []
+        for cb, group in groupby(sorted(callbacks), lambda x: x[0]):
+            cb_streams = [s for _, s in group]
```
jlstevens: ``cb_streams`` doesn't seem to be used...

philippjfr: Thanks that's a bug.

holoviews/plotting/bokeh/element.py
```diff
@@ -168,18 +166,56 @@ def __init__(self, element, plot=None, **params):
...
+            dims = list(self.overlay_dims.keys())
+        dims += element.dimensions()
+        tooltips = [(d.pprint_label, '@'+util.dimension_sanitizer(d.name))
+                    for d in dims]
+
+        callbacks = callbacks+self.callbacks
+        cb_tools = []
+        for cb in callbacks:
+            for handle in cb.handles:
+                if handle and handle in known_tools:
```
jlstevens: If I am following this right, this could be simplified if there was a single (custom) tool for capturing events? I believe this code is about finding the right tool that the callback needs...

holoviews/plotting/bokeh/element.py
```diff
@@ -512,7 +552,7 @@ def initialize_plot(self, ranges=None, plot=None, plots=None, source=None):
         self.handles['plot'] = plot
 
         # Get data and initialize data source
-        empty = self.callbacks and self.callbacks.downsample
+        empty = False
         if self.batched:
```
jlstevens: Is this temporary? If not, there is no reason for ``empty`` to appear as I don't see it being set a different value anywhere...

holoviews/plotting/bokeh/element.py
```diff
@@ -933,14 +971,14 @@ def _init_tools(self, element):
...
-                    el_tools = subplot._init_tools(el)
-                    el_tools = [t for t in el_tools
-                                if not (isinstance(t, HoverTool) and hover)]
-                    tools += el_tools
-                    if any(isinstance(t, HoverTool) for t in el_tools):
-                        hover = True
+            el = element.get(key)
+            if el is not None:
+                el_tools = subplot._init_tools(el, self.callbacks)
+                el_tools = [t for t in el_tools
```
jlstevens: I guess ``OverlayPlot`` is merging the set of tools? Did it always have to do this or is this just for callbacks? If it is the latter, then I think this is another reason to have our own custom tool. :-)

holoviews/plotting/bokeh/element.py
philippjfr: Yes, I'm going to end up getting rid of empty everywhere I think. But probably in a separate PR.

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -1,446 +1,225 @@
...
-    custom javascript code.
-
-    The user should define any plot_attributes and cb_attributes
-    he wants access to, which will be supplied in the form of a
-    dictionary to the call method. The call method can then apply
-    any processing depending on the callback data and return the
-    modified bokeh plot objects.
+    Generates JS code to look up attributes on JS objects from
+    an attributes specification dictionary.
+    """
```
jlstevens: Could the docstring have a simple example of input and the corresponding JS output? It is hard to tell just looking at the code...

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -1,446 +1,225 @@
...
+    Provides a baseclass to define callbacks, which return data from
+    bokeh models such as the plot ranges or various tools. The callback
+    then makes this data available to any streams attached to it.
+
+    The defintion of a callback consists of a number of components:
+
+    * handles    :  The handles define which object the callback will be
+                    attached on.
+
```
jlstevens: What kind of object? HoloViews objects? If so, it should say so. Also, 'definition' is misspelt above...

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -1,446 +1,225 @@
...
+    * attributes :  The attributes define which attributes to send back
+                    to Python. They are defined as a dictionary mapping
+                    between the name under which the variable is made
+                    available to Python and the specification of the
+                    attribute. The specification should start with the
+                    variable name that is to be accessed and the
+                    location of the attribute separated by periods.
+                    All plotting handles such as tools, the x_range,
+                    y_range and (data)source can be addressed in this
+                    way.
```
jlstevens: After reading this description, I *think* I understand but I think a short example would really help here....

holoviews/plotting/bokeh/element.py
philippjfr: It's always been merging them.

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -1,446 +1,225 @@
...
+    * code        : Defines any additional JS code to be executed,
+                    which can modify the data object that is sent to
+                    the backend.
+
+    The callback can also define a _process_msg method, which can
+    modify the data sent by the callback before it is passed to the
+    streams.
     """
```
jlstevens: Having described these three things (handles, attributes and code) is there a performance reason that only ``code`` is a parameter and the other two are class attributes? I'm also a little confused as to why they are class attributes and not instance attributes...

holoviews/plotting/bokeh/element.py
```diff
@@ -533,9 +573,9 @@ def initialize_plot(self, ranges=None, plot=None, plots=None, source=None):
             self._update_glyph(glyph, properties, mapping)
         if not self.overlaid:
             self._update_plot(key, plot, style_element)
-        if self.callbacks:
-            self.callbacks(self)
-            self.callbacks.update(self)
+
+        for cb in self.callbacks:
+            cb.initialize()
         if not self.overlaid:
```
jlstevens: It is a little confusing that there is a ``_init_callbacks`` method which actually constructs the callback instances (based on my understanding of it) and then an ``initialize`` method as well. Maybe the former should be called ``_construct_callbacks`` instead...

holoviews/plotting/bokeh/callbacks.py
philippjfr: The user can supply some custom code to the instance to be executed alongside the callback but each Callback really only applies to certain handles (e.g. ``RangeX`` and ``x_range``) and specific attributes on those handles (e.g. ``x_range.attributes.x``). I have no objection making them all parameters though, you originally suggested making them class attributes.

jlstevens: In fact ``code`` seems to be a class parameter as you can't set it through the constructor! Nothing wrong with that although I think my question above still holds...

jlstevens: Actually, ``code`` seems to be the only reason this class is parameterized at all...

holoviews/plotting/bokeh/callbacks.py
jlstevens: I now think they are in fact 'plotting handles'.

holoviews/plotting/bokeh/element.py
philippjfr: Agreed.

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -1,446 +1,225 @@
...
+    msgs=[]
+
+    def __init__(self, plot, streams, source, **params):
+        self.plot = plot
+        self.streams = streams
+        self.comm = JupyterCommJS(plot, on_msg=self.on_msg)
+        self.source = source
+
+
+    def initialize(self):
```
jlstevens: If I understand this right, the docstring could say something like 'Iterates through the plot (and any subplots), registering the necessary JS code across all relevant plotting handles'. Again, that is what I expect it to be doing from the code!

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -1,446 +1,225 @@
...
-    def serialize(self, models):
-        """
-        Serializes any Bokeh plot objects passed to it as a list.
-        """
-        for plot in self.plots:
-            msg = compute_static_patch(plot.document, models)
-            plot.comm.send(serialize_json(msg))
-        return 'Complete'
 
+    def set_customjs(self, cb_obj):
```
jlstevens: Should ``cb_obj`` simply be called ``handle`` for consistency?

holoviews/plotting/bokeh/element.py
```diff
@@ -168,18 +166,56 @@ def __init__(self, element, plot=None, **params):
...
+            for handle in cb.handles:
+                if handle and handle in known_tools:
+                    if handle == 'hover':
+                        tool = HoverTool(tooltips=tooltips)
+                    else:
+                        tool = known_tools[handle]()
+                    cb_tools.append(tool)
+                    self.handles[handle] = tool
+
+        tools = cb_tools + self.default_tools + self.tools
```
philippjfr: In some cases an actual tool is necessary e.g. to get the geometry of a selection, but I agree a hidden tool would be preferable for hover and clicking callbacks.

holoviews/plotting/bokeh/callbacks.py
jlstevens: > The user can supply some custom code to the instance to be executed alongside the callback

Sure! In that case you'll need to call ``super`` :-)

In the end, I do think class attributes are fine now I see how they are defined on each concrete Callback...

jlstevens: Though as the plotting code creates these things, I don't see how code could be customized per instance...

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -1,446 +1,225 @@
...
-from bokeh.core.json_encoder import serialize_json
-from bokeh.io import _CommsHandle
-from bokeh.util.notebook import get_comms
+from ...streams import (Stream, PositionXY, RangeXY, Selection1D, RangeX,
+                        RangeY, PositionX, PositionY, Bounds)
+from ..comms import JupyterCommJS
 
 
-class Callback(param.ParameterizedFunction):
+def attributes_js(attributes):
```
jlstevens: In the end, I am not convinced that this class needs to be parameterized.

Lastly, it might be nice to define ``Callback`` to be backend agnostic. These methods look fairly independent from bokeh in their semantics to me:

```
def __init__(self, plot, streams, source, **params):
      
def initialize(self):

def on_msg(self, msg):

def _process_msg(self, msg):
```

No need to abstract it out in this PR though...

jlstevens:
I understand this is now ready to merge!

Overall, I'm really pleased with how the streams system is turning out. It was simple to implement streams in core/plotting and this PR shows how you can get access to events from Bokeh in a nice clean way.

There are still some outstanding things to address in future PRs:

* I think some of the implementation here could be simplified using a custom Bokeh tool to captures some of the events.
* Throttling needs to be implemented on the JS side with the intent of eventually making it backend agnostic.
* Overlays need to be handled properly. 
* We need to replace the plotting callback system e.g re-introduce downsampling as an operation.

Together with some nice documentiation and examples on mybinder, I think this will warrant the 1.7 release!

PR #893, opened 2016-09-30
Fixed updating of colormappers

philippjfr:
The way color mappers were updated was a bit hacky and in fact didn't work in some cases. This ensures existing color mapper instances are simply updated.


philippjfr:
Ready to merge when tests pass (there are bokeh colormapper tests).

jlstevens:
Looks good. Merging.

Issue #812, opened 2016-08-15
Datashader Operations


philippjfr:
[Datashader](https://github.com/bokeh/datashader/) provides a great way to rasterize tabular data through 2D aggregation and optionally provides powerful tools for transforming and colormapping the aggregates. By providing an interface to datashader in holoviews we can make it trivial to explore very large datasets meaningfully. In HoloViews we already provide ElementOperations, which can provide parameterized interfaces to transform one or more elements into one or more other Elements. In private discussion we agreed it might make sense to add datashader operations, I'll try to outline how I propose to wrap different parts of the API by going through the different stages of the pipeline so we can discuss it.

### Aggregation

The aggregation step simply provides simple (but fast) 2D binning (at least when using the points glyph):

```python
cvs = ds.Canvas(plot_width=w, plot_height=h, x_range=x_range, y_range=y_range)
agg = cvs.points(df, 'longitude', 'latitude',  ds.count())
```

##### HoloViews API:

```python
class Aggregate(ElementOperation):

    aggregator = param.ClassSelector(class_=ds.reductions.Reduction,
                                     default=ds.count())

    height = param.Integer(default=800)

    width = param.Integer(default=600)

    x_range  = param.NumericTuple(default=(0, 1))

    y_range  = param.NumericTuple(default=(0, 1))
...
```

The idea would be that the general operation would call lower level aggregation operation that execute the aggregation that makes the most sense given the types of the Elements passed to it, i.e. Curves and Path Elements (and NdOverlays of them) are aggregated as paths while points/scatter (and every other >=2D Element) are aggregated as points. 

### Shading

The shading portion of the pipeline simply takes the 2D aggregate and provides custom color mapping.

```python
img = tf.interpolate(agg, cmap=Hot, how='eq_hist')
```

##### HoloViews API

```python
class Shade(ElementOperation):

    cmap = param.ClassSelector(class_=(basestring, bokeh.Palette, mpl.colormap, list))

    how = param.ObjectSelector(objects=['linear', 'log', 'eq_hist'])
```

Input: Image/GridImage or NdOverlay of Images (for categorical colorization)

Output: RGB

Other transforms like dynamic spreading would be exposed in similar ways.

Once we have implemented these operations they can immediately be used directly. The real value will come once we have exposed streams where it should then be possible to easily apply an operation as part of a stream for example to aggregate based on the current zoom level in a bokeh plot.

[...]

PR #894, opened 2016-10-04
Add Datashader operations

philippjfr:
As suggested in #812 this adds datashader operations to HoloViews, making it trivial to aggregate any Element type into an Image or stack of images and then shade it. The PR currently adds two main operations:

* Aggregate: Takes any 2D dataset with or without values and generates a ``GridImage`` aggregate wrapping an xarray DataArray.
* Shade: Takes the output of an aggregate and applies color interpolation and a transfer function ('linear', 'log', 'eq_hist'), returning an RGB element.

I've not yet written docstrings or any tests yet, but I do have the initial skeleton of a tutorial: https://anaconda.org/philippjfr/holoviews_datashader/notebook

### DIFF
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -70,81 +70,6 @@ class Operation(param.ParameterizedFunction):
             raise ValueError("Extents across the overlay are inconsistent")
 
 
-class Dynamic(Operation):
-    """
-    Dynamically applies a callable to the Elements in any HoloViews
-    object. Will return a DynamicMap wrapping the original map object,
-    which will lazily evaluate when a key is requested. By default
-    Dynamic applies a no-op, making it useful for converting HoloMaps
-    to a DynamicMap.
-
-    Any supplied kwargs will be passed to the callable and any streams
-    will be instantiated on the returned DynamicMap.
-    """
-
-    callable = param.Callable(default=lambda x: x, doc="""
-        Function or ElementOperation to apply to DynamicMap items
-        dynamically.""")
-
-    kwargs = param.Dict(default={}, doc="""
-        Keyword arguments passed to the function.""")
-
-    streams = param.List(default=[], doc="""
-        List of streams to attach to the returned DynamicMap""")
-
-    def __call__(self, map_obj, **params):
-        self.p = param.ParamOverrides(self, params)
-        callback = self._dynamic_operation(map_obj)
-        if isinstance(map_obj, DynamicMap):
-            dmap = map_obj.clone(callback=callback, shared_data=False)
-        else:
-            dmap = self._make_dynamic(map_obj, callback)
-        if isinstance(self.p.callable, ElementOperation):
-            return dmap.clone(streams=[s() for s in self.p.streams])
-        return dmap
-
-
-    def _process(self, element):
-        if isinstance(self.p.callable, Operation):
-            return self.p.callable.process_element(element, **self.p.kwargs)
-        else:
-            return self.p.callable(element, **self.p.kwargs)
-
-
-    def _dynamic_operation(self, map_obj):
-        """
-        Generate function to dynamically apply the operation.
-        Wraps an existing HoloMap or DynamicMap.
-        """
-        if not isinstance(map_obj, DynamicMap):
-            def dynamic_operation(*key, **kwargs):
-                self.p.kwargs.update(kwargs)
-                return self._process(map_obj[key])
-            return dynamic_operation
-
-        def dynamic_operation(*key, **kwargs):
-            key = key[0] if map_obj.mode == 'open' else key
-            self.p.kwargs.update(kwargs)
-            _, el = util.get_dynamic_item(map_obj, map_obj.kdims, key)
-            return self._process(el)
-
-        return dynamic_operation
-
-
-    def _make_dynamic(self, hmap, dynamic_fn):
-        """
-        Accepts a HoloMap and a dynamic callback function creating
-        an equivalent DynamicMap from the HoloMap.
-        """
-        if isinstance(hmap, ViewableElement):
-            return DynamicMap(dynamic_fn, kdims=[])
-        dim_values = zip(*hmap.data.keys())
-        params = util.get_param_values(hmap)
-        kdims = [d(values=list(set(values))) for d, values in
-                 zip(hmap.kdims, dim_values)]
-        return DynamicMap(dynamic_fn, **dict(params, kdims=kdims))
-
-
 
 class ElementOperation(Operation):
     """
@@ -204,9 +129,10 @@ class ElementOperation(Operation):
             processed = GridSpace(grid_data, label=element.label,
                                   kdims=element.kdims)
         elif dynamic:
+            from ..util import Dynamic
             streams = getattr(self, 'streams', [])
             processed = Dynamic(element, streams=streams,
-                                callable=self, kwargs=params)
+                                operation=self, kwargs=params)
         elif isinstance(element, ViewableElement):
             processed = self._process(element)
         elif isinstance(element, DynamicMap):
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -82,7 +82,7 @@ class ElementOperation(Operation):
     An ElementOperation can be set to be dynamic, which will return a
     DynamicMap with a callback that will apply the operation
     dynamically. An ElementOperation may also supply a list of Stream
-    classes on the streams attribute, which can allow dynamic control
+    classes on a streams parameter, which can allow dynamic control
     over the parameters on the operation.
     """
 
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -78,6 +78,12 @@ class ElementOperation(Operation):
     input, a processed holomap is returned as output where the
     individual elements have been transformed accordingly. An
     ElementOperation may turn overlays in new elements or vice versa.
+
+    An ElementOperation can be set to be dynamic, which will return a
+    DynamicMap with a callback that will apply the operation
+    dynamically. An ElementOperation may also supply a list of Stream
+    classes on the streams attribute, which can allow dynamic control
+    over the parameters on the operation.
     """
 
     dynamic = param.ObjectSelector(default='default',
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -70,28 +70,40 @@ class Operation(param.ParameterizedFunction):
             raise ValueError("Extents across the overlay are inconsistent")
 
 
-class DynamicOperation(Operation):
+class DynamicFunction(Operation):
     """
-    Dynamically applies an operation to the elements of a HoloMap
-    or DynamicMap. Will return a DynamicMap wrapping the original
-    map object, which will lazily evaluate when a key is requested.
-    The _process method should be overridden in subclasses to apply
-    a specific operation, DynamicOperation itself applies a no-op,
-    making the DynamicOperation baseclass useful for converting
-    existing HoloMaps to a DynamicMap.
+    Dynamically applies an operation the Elements in any HoloViews
+    object.  Will return a DynamicMap wrapping the original map
+    object, which will lazily evaluate when a key is requested.  By
+    default DynamicFunction itself applies a no-op, making the
+    DynamicFunction baseclass useful for converting existing HoloMaps
+    to a DynamicMap.
     """
+
+    function = param.Callable(default=lambda x: x, doc="""
+        Function to apply to DynamicMap items dynamically.""")
+
+    kwargs = param.Dict(default={}, doc="""
+        Keyword arguments passed to the function.""")
+
 
     def __call__(self, map_obj, **params):
         self.p = param.ParamOverrides(self, params)
         callback = self._dynamic_operation(map_obj)
         if isinstance(map_obj, DynamicMap):
-            return map_obj.clone(callback=callback, shared_data=False)
+            dmap = map_obj.clone(callback=callback, shared_data=False)
         else:
-            return self._make_dynamic(map_obj, callback)
+            dmap = self._make_dynamic(map_obj, callback)
+        if isinstance(self.p.function, ElementOperation) and self.p.function.streams:
+            return dmap.clone(streams=[s() for s in self.p.function.streams])
+        return dmap
 
 
     def _process(self, element):
-        return element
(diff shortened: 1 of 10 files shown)


holoviews/operation/datashader.py
```diff
@@ -0,0 +1,208 @@
...
+
+def uint32_to_uint8(img):
+    return np.flipud(img.view(dtype=np.uint8).reshape(img.shape + (4,)))
+
+
+class Shade(ElementOperation):
+
+    cmap = param.ClassSelector(class_=(Iterable, Callable))
+
+    how = param.ObjectSelector(default='eq_hist', objects=['linear', 'log', 'eq_hist'])
```
jlstevens: Not sure ``how`` is a great parameter name. Maybe something like ``color_mapping`` or ``shading_mode``?

holoviews/operation/datashader.py
```diff
@@ -0,0 +1,208 @@
...
+                df = element.dframe()
+                if isinstance(obj, NdOverlay):
+                    df = df.assign(**dict(zip(obj.dimensions('key', True), key)))
+                paths.append(df)
+            kdims += element.kdims
+            vdims = element.vdims
+        elif isinstance(obj, Element):
+            line = isinstance(obj, Curve)
+            paths.append(obj.dframe())
+        if line:
```
jbednar: It's never nice to see a bunch of if-else blocks to support different types, but I don't see a cleaner way to propose.

holoviews/operation/datashader.py
```diff
@@ -0,0 +1,208 @@
...
+        cvs = ds.Canvas(plot_width=width, plot_height=height,
+                        x_range=(xstart, xend), y_range=(ystart, yend))
+
+        if line:
+            return cvs.line(data, x, y, self.p.aggregator)
+        else:
+            return cvs.points(data, x, y, self.p.aggregator)
+            
+
+def uint32_to_uint8(img):
```
jlstevens: Given this is used in a single place, I would consider inlining it or at least making it a classmethod of Shade...

holoviews/operation/datashader.py
```diff
@@ -0,0 +1,208 @@
...
+            height = int(min([(y_range/self.p.y_sampling), height]))
+
+        cvs = ds.Canvas(plot_width=width, plot_height=height,
+                        x_range=(xstart, xend), y_range=(ystart, yend))
+
+        if line:
+            return cvs.line(data, x, y, self.p.aggregator)
+        else:
+            return cvs.points(data, x, y, self.p.aggregator)
+
```
jbednar: Maybe instead of a boolean `line`, this could be handled more generally by setting `glyph=cvs.points` or `glyph=cvs.line` above, then doing `return getattr(cvs,glyph)(data, x, y, self.p.aggregator)` here?

holoviews/operation/datashader.py
```diff
@@ -0,0 +1,208 @@
...
+        if isinstance(element, NdOverlay):
+            bounds = element.last.bounds
+            element = self.concatenate(element)
+        else:
+            bounds = element.bounds
+
+        array = element.data[element.vdims[0].name]
+        kdims = element.kdims
+        shade_opts = dict(how=self.p.how)
+        if element.ndims > 2:
```
jbednar: A comment would be helpful here to summarize what this fairly convoluted processing of colormaps is for.

holoviews/operation/datashader.py
philippjfr: It's consistent with datashader at least, I'd prefer it over either of those.

holoviews/operation/datashader.py
philippjfr: Sure.

holoviews/operation/datashader.py
```diff
@@ -0,0 +1,208 @@
...
+
+    y_range  = param.NumericTuple(default=None, length=2,
+                                  allow_None=True)
+
+    x_sampling = param.Number(default=None, doc="""
+        Specifies the smallest allowed sampling interval along the y-axis.""")
+
+    y_sampling = param.Number(default=None, doc="""
+        Specifies the smallest allowed sampling interval along the y-axis.""")
+
```
philippjfr: They let you define a minimum sampling interval. It's one of my major annoyances with datashader you can zoom in beyond the sampling resolution, giving you useless speckly aggregates.

holoviews/operation/datashader.py
philippjfr: Sounds good.

holoviews/operation/datashader.py
```diff
@@ -0,0 +1,208 @@
...
+def uint32_to_uint8(img):
+    return np.flipud(img.view(dtype=np.uint8).reshape(img.shape + (4,)))
+
+
+class Shade(ElementOperation):
+
+    cmap = param.ClassSelector(class_=(Iterable, Callable))
+
+    how = param.ObjectSelector(default='eq_hist', objects=['linear', 'log', 'eq_hist'])
+
```
philippjfr: Hmm, does it even support tab-completion? Either way cbrt and any other callable should be allowed.

holoviews/operation/datashader.py
philippjfr: Will add docstrings and comments next.

holoviews/operation/datashader.py
jbednar: It's actually a normalization function, so `normalizer`? `normalization`?

holoviews/operation/datashader.py
jbednar: Should that be added to datashader instead?

holoviews/operation/datashader.py
jbednar: Calling it "glyph" is a bit confusing, as it's actually a method name rather than an actual glyph object, but "method" seems meaningless...

holoviews/operation/datashader.py
jlstevens: So ``how`` is an existing parameter name in datashader then? I suspected as much...

holoviews/operation/datashader.py
jlstevens: I made a comment about this below (my fault for reviewing PRs bottom up!). I see the  argument regarding consistency but I still don't like a parameter called 'how'. Personally, I find 'normalization' to be a much nicer name...

holoviews/core/operation.py
```diff
@@ -70,28 +70,45 @@ def get_overlay_bounds(cls, overlay):
...
+    object. Will return a DynamicMap wrapping the original map object,
+    which will lazily evaluate when a key is requested. By default
+    Dynamic applies a no-op, making it useful for converting HoloMaps
+    to a DynamicMap.
+
+    Any supplied kwargs will be passed to the callable and any streams
+    will be instantiated on the returned DynamicMap.
     """
 
+    callable = param.Callable(default=lambda x: x, doc="""
```
jlstevens: In discussion, we decided ``operation`` would be an even better name...

holoviews/core/operation.py
```diff
@@ -70,28 +70,45 @@ def get_overlay_bounds(cls, overlay):
             raise ValueError("Extents across the overlay are inconsistent")
 
 
-class DynamicOperation(Operation):
+class Dynamic(Operation):
```
jlstevens: Should probably be a ``ParameterizedFunction`` and moved to some other file as a utility..

holoviews/core/operation.py
```diff
@@ -182,7 +182,7 @@ def _process(self, view, key=None):
         raise NotImplementedError
 
 
-    def process_element(self, element, key, **params):
+    def process_element(self, element, key=None, **params):
```
jlstevens: Why is ``key`` now optional? From memory, the purpose of this method was to always supply a key and get the corresponding result...

holoviews/core/operation.py
```diff
@@ -197,16 +197,18 @@ def __call__(self, element, **params):
...
-        elif isinstance(element, GridSpace):
+        if isinstance(element, GridSpace):
             # Initialize an empty axis layout
             grid_data = ((pos, self(cell, **params))
                          for pos, cell in element.items())
             processed = GridSpace(grid_data, label=element.label,
                                   kdims=element.kdims)
         elif dynamic:
-            processed = DynamicFunction(element, function=self, kwargs=params)
+            streams = getattr(self, 'streams', [])
```
jlstevens: This feature probably needs mentioning in the docstring wherever ``dynamic`` is mentioned.

holoviews/operation/datashader.py
jbednar: I don't much like it either.  I might change it in datashader as well, so I'd vote for using  `normalization` here.

holoviews/operation/datashader.py
```diff
@@ -68,74 +69,91 @@ def dataset_pipeline(dataset, schema, canvas, glyph, summary):
 
 
 class Aggregate(ElementOperation):
+    """
+    Aggregate implements 2D binning for any valid HoloViews Element
+    type using datashader. By default it will simply count the number
+    of values in each bin but custom aggregators can be supplied
+    implementing mean, max, min and other reduction operations.
```
jbednar: I wouldn't call them "custom"; they are just other aggregators.  So maybe "bin but other aggregators can be supplied explicitly".

holoviews/operation/datashader.py
```diff
@@ -68,74 +69,91 @@ def dataset_pipeline(dataset, schema, canvas, glyph, summary):
...
+
+    The bins of the aggregate are defined by the width and height and
+    the x_range and y_range. If x_sampling or y_sampling are supplied
+    the operation will ensure that a bin is no smaller than the
+    minimum sampling distance.
+    """
```
jbednar: How?

holoviews/operation/datashader.py
```diff
@@ -68,74 +69,91 @@ def dataset_pipeline(dataset, schema, canvas, glyph, summary):
...
     aggregator = param.ClassSelector(class_=ds.reductions.Reduction,
                                      default=ds.count())
 
-    height = param.Integer(default=800)
+    height = param.Integer(default=800, doc="""
+       The height of the aggregated image in pixels.""")
 
-    width = param.Integer(default=600)
+    width = param.Integer(default=600, doc="""
+       The width of the aggregated image in pixels.""")
```
jbednar: Should it be square by default, since the default hv image size is square in aspect?

philippjfr: True, I'll also make it a bit smaller since our plots aren't nearly this big by default.

holoviews/operation/datashader.py
```diff
@@ -148,32 +166,52 @@ def _process(self, element, key=None):
...
-            
-
-def uint32_to_uint8(img):
-    return np.flipud(img.view(dtype=np.uint8).reshape(img.shape + (4,)))
 
 
 class Shade(ElementOperation):
+    """
+    Shade applies a normalization function to the data and then
+    applies colormapping to an Image or NdOverlay of Images, returning
```
jbednar: Maybe explain what "the data" is?  I.e. is it an xarray, either 2D or if 3D then a stack of categories?

holoviews/operation/datashader.py
jbednar: Can you just make it respect the normal holoviews dpi, etc., so that there are no explicit height and width parameters here at all?

philippjfr: Hmm, the user still might want a specific height/width for export. Adding a stream for the bokeh plot width and height would be possible though, then it's even responsive to resizing.

jbednar: But isn't that true for hv in general, not just datashader?  In any case, respecting the bokeh plot height and width however that could be done would be fabulous, because that will ensure pixel-accurate rendering with no wasted pixels.  Not sure how you can get that information out of bokeh, though.

philippjfr: Easily enough, I should be able to send it along with the ranges.

holoviews/operation/datashader.py
```diff
@@ -172,21 +179,28 @@ def _process(self, element, key=None):
...
-                                         doc="""
-        The normalization operation applied before colormapping.""")
-
+    normalization = param.ClassSelector(default='eq_hist',
+                                        class_=(basestring, Callable),
+                                        doc="""
+        The normalization operation applied before colormapping.
+        Valid options include 'linear', 'log', 'eq_hist', 'cbrt',
+        and any valid transfer function that acces data, mask, nbins
+        arguments.""")
```
jbednar: acces -> has?

holoviews/operation/datashader.py
jbednar: I thought one could only get the overall size of the plot, not the size of the data area?  Anyway, either is better than nothing!

philippjfr: Ah that's true. If you disable axes it'll be fairly accurate I guess.

holoviews/core/operation.py
```diff
@@ -152,6 +78,12 @@ class ElementOperation(Operation):
     input, a processed holomap is returned as output where the
     individual elements have been transformed accordingly. An
     ElementOperation may turn overlays in new elements or vice versa.
+
+    An ElementOperation can be set to be dynamic, which will return a
+    DynamicMap with a callback that will apply the operation
+    dynamically. An ElementOperation may also supply a list of Stream
+    classes on the streams attribute, which can allow dynamic control
+    over the parameters on the operation.
```
jlstevens: I would have said 'on a streams parameter' instead of 'on the streams attribute'...

holoviews/util.py
```diff
@@ -0,0 +1,86 @@
...
+    which will lazily evaluate when a key is requested. By default
+    Dynamic applies a no-op, making it useful for converting HoloMaps
+    to a DynamicMap.
+
+    Any supplied kwargs will be passed to the callable and any streams
+    will be instantiated on the returned DynamicMap.
+    """
+
+    operation = param.Callable(default=lambda x: x, doc="""
+        Callable to apply to DynamicMap items dynamically.""")
```
jlstevens: How about 'Operation or user-defined callable to apply dynamically'?

jlstevens:
Looks good to me!

There are a few other things we have discussed:

* Making ``Shade`` into a compositor.
* Updating the compositor system to become a plot option
* Generalizing (compositors are really operations+predicates) and documenting them.

I also expect there will be some additional PRs for small fixes and general clean-up but this one is now ready to merge.

PR #896, opened 2016-10-05
Fixed extents on bokeh HistogramPlot

philippjfr:
The extents of the Histogram were not correctly computed previously showing just the region between the min and max bin value. Histograms bins should instead reach down (or up) to zero.


jlstevens:
Looks good. Merging.

PR #899, opened 2016-10-05
Avoid deprecation warnings in bokeh 0.12.3

philippjfr:
When setting font sizes you now have to wrap it in a value object otherwise you get a deprecation warning.


jlstevens:
Looks fine. Merge?

jlstevens:
It seems ``value`` has been around for a while so it should remain backwards compatible. Merging.

PR #902, opened 2016-10-05
Fixed empty NdOverlay on initialization in batched mode

philippjfr:
Previously some parts of the plotting code would receive empty NdOverlays on initialization causing it to break. Now the plot is initialized on the last non-empty frame. Since the plot is updated subsequently with the correct frame this should not cause any issues.


jlstevens:
Tests have passed. Merging.

PR #903, opened 2016-10-05
Added downsample_columns operation

philippjfr:


jlstevens:
Looks good and I am much happier with the new name. Merging.

PR #904, opened 2016-10-05
Allow multiple streams attached to the same plot model

philippjfr:
Currently when multiple streams are attached to the same plot model they will override each other. This PR ensures that  either the python callback or the JS callback are merged depending on whether the callback is of the same type. This is mostly a problem if two separate plots attach distinct streams to a set of linked axes. It also includes a fix for Layouts with multiple streams of the same type across the subplots, which would previously complain about parameter overlap. Finally it allows supplying stream instances to operations instead of instantiating a Stream instance internally, which can be useful to get a handle on the stream.

Here's a demo:

```
points = hv.Points(np.random.multivariate_normal((0,0), [[0.1, 0.1], [0.1, 1.0]], (1000000,)))
Datashade(points) + decimate(points)
```

linked_datadecimate

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -133,6 +133,8 @@ class Callback(object):
     # Timeout if a comm message is swallowed
     timeout = 20000
 
+    _callbacks = {}
+
     def __init__(self, plot, streams, source, **params):
         self.plot = plot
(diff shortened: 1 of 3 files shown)


jbednar:
It certainly looks and sounds good!

jlstevens:
Looks good. Merging.

PR #905, opened 2016-10-05
Avoid traversing data when computing bokeh plot update

philippjfr:
Currently there are some cases where the code is traversing large arrays or lists when trying to compute the required plot update for a bokeh plot. This PR simply makes sure they are no longer traversed in the remaining cases where this was not yet handled.


jlstevens:
Looks fine. Merging.

PR #906, opened 2016-10-05
Improved error reporting on comms

philippjfr:
Improves the traceback reported in the console on error by including filename, error type, line number and function.

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -79,6 +79,9 @@ class Callback(object):
           var comm = HoloViewsWidget.comms["{comms_target}"];
           var comm_state = HoloViewsWidget.comm_state["{comms_target}"];
           if (msg.msg_type == "Ready") {{
+            if (msg.content) {{
+              console.log("Python callback returned following output:", msg.content);
+            }}
             if (comm_state.event) {{
               comm.send(comm_state.event);
             }} else {{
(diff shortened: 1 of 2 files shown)


jlstevens:
I'm very happy with this, especially as everything stays simple on the Javascript side.

My only comments:

* ``Capture`` needs a docstring to explain why it is needed - to capture ``stdout`` (e.g print output) and make it available on the web console for debugging etc.
* It probably wouldn't be too hard to capture ``stderr`` at the same time and I would consider doing this for consistency.
* Lastly, if you do catch ``stderr`` as well, I might consider renaming ``Capture`` to something like ``StandardOutput``. Then the context manager can be used with:

```python
with StandardOutput() as stdout:
     ...
if stdout: ... 
```

You'll need two context managers if you want to also handle stderr - at least if you want use each one this way by inheriting from ``list``:

```python
with StandardOutput() as stdout, StandardError() as stderr:
     ...
if stdout: ...
if stderr: ... 
```

The redundancy between the  ``StandardOutput`` and  ``StandardError`` context managers is a little annoying but maybe you could parameterize it to have ``StandardIO('stdout)'`` and ``StandardIO('stderr')``...

jlstevens:
Great, tests are passing! Merging.

jlstevens:
I know this PR is merged now but I want to just state one thing that is still niggling at me. How will a regular user even know the check the Javascript console for messages? Both of us know that the output is going there now but it is invisible to everyone else until they look at the docs to find out.

Not sure how best to handle this. Javascript alerts would be very obvious, but also very annoying...though maybe for stdout, using alerts might make sense even if exceptions don't?

PR #912, opened 2016-10-07
Dashboard comms fixes

philippjfr:
A fix to get streams working on Jupyter dashboards and another fix to ensure events are processed when an error is received.

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -85,7 +85,9 @@ class Callback(object):
           }}
           comm_state.timeout = Date.now();
           comm_state.event = undefined;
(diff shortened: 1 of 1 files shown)


jlstevens:
Great - though I assume we still need to use a dev version of ipywidgets for the dashboard comms to work?

philippjfr:
>Great - though I assume we still need to use a dev version of ipywidgets for the dashboard comms to work?

Yes, I also remembered you have to satisfy the missing jQuery dependency. We'll be able to expose that in the ``notebook_extension`` though.

jlstevens:
Thanks for clarifying. Ready to merge?

philippjfr:
Yep

PR #913, opened 2016-10-07
Default Bounds stream to None

philippjfr:
Any arbitrary default values are unlikely to really make sense and Bounds should also allow None in other cases.

### DIFF
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -254,7 +254,7 @@ class Bounds(Stream):
     tuple of the left, bottom, right and top coordinates.
     """
 
-    bounds = param.NumericTuple(default=(0, 0, 1, 1), constant=True,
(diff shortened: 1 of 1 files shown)


jlstevens:
> Any arbitrary default values are unlikely to really make sense

Well, I already made use of these values in a Bounds stream I used - my code snapped to the nearest available key which meant any value would do. 

I also have a vague memory of the default of NumericTuple being specified as ``None`` actually becomes ``(None, None)`` or a similar tuple of the specified length. If that is true, then the length should be specified as four and the default is really ``(None,None,None,None)``.

I might be wrong but this is worth checking. If it really is None, I can merge and then update the code that made use of this default value.

philippjfr:
It really is None.

jlstevens:
Must be remembering an old param bug then. Merging.

PR #919, opened 2016-10-10
Add Versioneer for pep480 compliance

basnijholt:
I keep having issues with installing holoviews and not getting the correct `__version__`.

This pull request implements the usage of [`versioneer`](https://github.com/warner/python-versioneer), a package that creates the version from git tags.

It is used in many packages, amongst them: [pandas](https://github.com/pydata/pandas/blob/master/versioneer.py), [numpy](https://github.com/ContinuumIO/gulinalg/blob/master/setup.py#L11), [matplotlib](https://github.com/matplotlib/matplotlib/blob/master/versioneer.py), and many more!

Now installing from a dirty commit will result in: 
```
import holoviews
holoviews.__version__
'v1.7dev1+102.gee129cc.dirty'
```
where the `v1.7dev1` is the version tag, `+102` means 102 commits after last tag, `gee129cc` the short hash of the commit you installed, and `dirty` means that I have changes I didn't commit.


jlstevens:
I am working on fixing ``param.Version`` to make sure it works (it should!). I assure you once I fix whatever is wrong with``param.Version``, it is much cleaner and much saner than using versioneer (it is only nice if you don't look at the horrible code inside it!)! :-)

jlstevens:
I've fixed the bug in ``param.Version`` in ioam/param#135  addressing the issue mentioned above.

I'll make sure the updated version of param is released before 1.7 and maybe we should consider pinning a more recent version of param in setup.py to make sure everyone receives the fix.

jbednar:
Thanks, @jlstevens.  Including 1800 lines of versioneer code for updating the version would be a bit much!  But I too have found having the repr not just being a plain string to be annoying. It seems to work well now in param!

jlstevens:
@jbednar I'm happy with the fix and I tested that it works properly. My only question is whether we should release a new version of param immediately or wait a bit?

jbednar:
I guess if we released param now then the other packages would immediately see versions strings start to work better, for any new or updated installation?  That's a good argument for updating now, though I believe there is at least one outstanding issue on Param raised by ChrisB that we might want to address before a release.

jlstevens:
> I guess if we released param now then the other packages would immediately see versions strings start to work better, for any new or updated installation? 

Right. We would have to pin the version to greater than 1.4.1 to update existing installations but as you say, new installations could benefit earlier.

jbednar:
Even if you did pin the version, it presumably wouldn't affect anything until someone updated something else, so it's maybe not that crucial to do it.  But sure, it won't hurt!

PR #922, opened 2016-10-10
Dynamic operations only pass valid params

philippjfr:
When creating a Dynamic ElementOperation with streams sometimes you want to supply additional streams, which do not supply parameters to the operation. This ensures that Dynamic only passes along parameters that are accepted by the operation.

### DIFF
--- a/holoviews/util.py
+++ b/holoviews/util.py
@@ -53,7 +53,9 @@ class Dynamic(param.ParameterizedFunction):
 
     def _process(self, element, key=None):
         if isinstance(self.p.operation, ElementOperation):
-            return self.p.operation.process_element(element, key, **self.p.kwargs)
+            kwargs = {k: v for k, v in self.p.kwargs
+                      if k in self.p.operation.params()}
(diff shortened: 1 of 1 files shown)


jlstevens:
Looks fine and the change makes sense. I still don't quite get the motivation though:

> When creating a Dynamic ElementOperation with streams sometimes you want to supply additional streams, which do not supply parameters to the operation.

What do you mean? That said, I see how this change would make using streams more robust....

philippjfr:
By default an operation can supply streams classes which supply parameters for the operation, e.g. think RangeXY which supplies ``x_range`` and ``y_range`` for the datashader operation. The user can however supply additional streams, which don't supply operation parameters, so this let's you do this:

```python
bounds = Bounds(bounds=None)
decimate(datapoints, streams=[bounds, RangeXY()])
```

Without getting a bunch of warnings about ``bounds`` being passed to the decimate operation.

jlstevens:
Ok. Makes sense and the tests are passing. Merging.

PR #923, opened 2016-10-10
Ensured dynamic callback is only called once

philippjfr:
A bug in the ``get_dynamic_item`` function ended up making clones of DynamicMaps which resulted in the dynamic callback being called multiple times. This PR fixes this and adds unit tests for all three backends to ensure the plot is appropriately updated.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -957,7 +957,8 @@ def get_dynamic_item(map_obj, dimensions, key):
     dmaps = map_obj.traverse(lambda x: x, ['DynamicMap'])
     dmap = dmaps[0] if dmaps else map_obj
     if key == () and (not dimensions or not dmap.kdims):
-        return key, map_obj.map(lambda x: x[()], ['DynamicMap', 'HoloMap'])
+        return key, map_obj.map(lambda x: x[()], ['DynamicMap', 'HoloMap'],
+                                clone=False)
     elif isinstance(key, tuple):
         dims = {d.name: k for d, k in zip(dimensions, key)
                 if d in map_obj.kdims}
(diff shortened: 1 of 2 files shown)


jlstevens:
Great!

The PR referenced above is an example notebook that needs this fix. It is also good to have those unit tests to make sure this bug is squashed for good. I'll now review the PR...

holoviews/core/util.py
```diff
@@ -967,7 +968,8 @@ def get_dynamic_item(map_obj, dimensions, key):
         key_offset = max([key-map_obj.cache_size, 0])
         key = map_obj.keys()[min([key-key_offset,
                                   len(map_obj)-1])]
-        el = map_obj.map(lambda x: x[key], ['DynamicMap'])
+        el = map_obj.map(lambda x: x[key], ['DynamicMap'],
+                         clone=False)
```
jlstevens: Very minor nitpick: I don't see why ``clone=False`` needs to be on the next line in this case. Even if it is over 80 characters long it looks a bit weird...

tests/testplotinstantiation.py
```diff
@@ -87,6 +88,19 @@ def test_errorbar_test(self):
         plot = mpl_renderer.get_plot(errorbars)
         plot.initialize_plot()
 
+    def test_stream_callback_single_call(self):
+        def history_callback(x, history=deque(maxlen=10)):
+            history.append(x)
+            return Curve(list(history))
```
jlstevens: Just like the price stream example notebook in contrib!

jlstevens:
The actual fix is very small and the tests looks good. Happy to merge if you feel it is ready...

philippjfr:
Once the tests pass let's merge it.

jlstevens:
Tests passed. Merging.

PR #926, opened 2016-10-11
Switched AttrTree dir mode to 'user'

jlstevens:
By design, `AttrTrees` are supposed to allow easy tab-completion with capitalized names coming first and lowercase attributes coming later. This is the reason we capitalize the path components of Layouts and Overlays.

In some recent version of IPython, this behavior has been broken:

image

Infuriatingly, the tab-completer even ignores the order of the values output by `__dir__`.

As a result, I've switched the dir_mode to 'user' to make AttrTrees work as intended. It means you won't see tab completion of the methods (e.g select) but at least you will see the items of the AttrTree properly.

### DIFF
--- a/holoviews/core/tree.py
+++ b/holoviews/core/tree.py
@@ -43,7 +43,7 @@ class AttrTree(object):
         else:
             return dir(type(self)) + list(dict_keys)
 
-    def __init__(self, items=None, identifier=None, parent=None, dir_mode='default'):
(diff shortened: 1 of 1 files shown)


philippjfr:
Should we close this now you've snuck the proper fix into IPython?

jlstevens:
Yes. 

I've closed this now we have our own mechanism to fix completion in older IPython versions and now that it will also be fixed in the next IPython release.

PR #927, opened 2016-10-11
Ensure get_dynamic_item does not modify object

philippjfr:
So this was a pretty nasty bug, turns out my usage of ``.map`` with ``clone=False`` was modifying the object causing various weird bugs. This PR is a bit of an ugly workaround updating the item in DynamicMaps using a ``map`` call that doesn't modify the object and updates it and a separate map call to actually get the item. Can't think of a better to solution to update composite objects (i.e. Layouts and AdjointLayout) containing dimensionless DynamicMaps since ``select`` won't work.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -957,8 +957,8 @@ def get_dynamic_item(map_obj, dimensions, key):
     dmaps = map_obj.traverse(lambda x: x, ['DynamicMap'])
     dmap = dmaps[0] if dmaps else map_obj
     if key == () and (not dimensions or not dmap.kdims):
-        return key, map_obj.map(lambda x: x[()], ['DynamicMap', 'HoloMap'],
-                                clone=False)
+        map_obj.traverse(lambda x: x[()], ['DynamicMap'])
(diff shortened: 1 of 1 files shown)


jlstevens:
Would calling ``select`` without arguments having the same effect as ``[()]`` for dimensionless dynamicmaps help (or at least appear more consistent)?

Otherwise, I'm happy to merge although I have to admit I have difficulty following this code (because it is a complicated thing to do..not because I think it is badly written).

philippjfr:
>Would calling select without arguments having the same effect as [()] for dimensionless dynamicmaps help (or at least appear more consistent)?

I would like it if there was some way of doing this via select but I currently don't see select without arguments working in that case because ``select`` ends up being called for all kinds of things, often without specifying any valid dimensions for that particular object. Let's merge for now, and I'll keep thinking about it, this fixes some fairly nasty bugs so I'd like it merged now. Let's revisit when we look at counter mode, which faces similar hairy issues as dimensionless DynamicMaps.

jlstevens:
Sure. Merging!

PR #929, opened 2016-10-12
Added log option to histogram operation

philippjfr:
This adds a log option to the histogram operation, which automatically generates logarithmic sampling for the histogram bins. Also includes a small fix to fix bokeh histograms with inverted axes.

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -479,7 +479,7 @@ class histogram(ElementOperation):
       Specifies whether the histogram will be rescaled for each Raster in a UniformNdMapping.""")
 
     log = param.Boolean(default=False, doc="""
-      Whether to apply log scaling """)
+      Whether to use base 10 logarithmic samples for the bin edges.""")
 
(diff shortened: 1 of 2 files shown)


holoviews/operation/element.py
```diff
@@ -478,6 +478,9 @@ class histogram(ElementOperation):
     individually = param.Boolean(default=True, doc="""
       Specifies whether the histogram will be rescaled for each Raster in a UniformNdMapping.""")
 
+    log = param.Boolean(default=False, doc="""
+      Whether to apply log scaling """)
```
jlstevens: I would at least mention that it is log10....

philippjfr: True, that's what logspace uses, but it should still be explained.

jlstevens:
Other than a small comment about the docstring, this PR looks fine and I am happy to merge.

jlstevens:
Other than one unrelated test failure, the tests seem to be ok. Merging.

PR #932, opened 2016-10-14
Removed extra copies of options in messages

jbednar:
RIght now, if you try to set an option with an invalid name, you get a huge list of all options, with three copies of everything.  There's probably a way to avoid having three copies, and if anyone wants to track that down, feel free!  Otherwise, we could merge this PR and at least make the messages easier to understand.

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -74,7 +74,7 @@ class OptionError(Exception):
 
     def message(self, invalid_keyword, allowed_keywords, group_name, path):
         msg = ("Invalid option %s, valid options are: %s"
-               % (repr(invalid_keyword), str(list(set(allowed_keywords)))))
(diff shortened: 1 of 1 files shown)


jbednar:
The test failures look unrelated to this change, and related to https://github.com/ContinuumIO/anaconda-issues/issues/1068

jlstevens:
The fix looks good and the test failures are annoying.

My only suggestion is that maybe it would be good to sort the set of keywords alphabetically?

jbednar:
Ok, sorted (pun intended).  I'm actually not sure how to trigger this message anymore, though, as I only saw it with parameter misspellings combined with various combinations of packages (param, jupyter, etc.) that were installed when I wrote the PR, and those are long gone.  It's hard to see how this would fail, but I do have to admit that I haven't tested it.

philippjfr:
Looks good and tests passed. Merging.

PR #933, opened 2016-10-14
Added stream activate and deactivate methods

philippjfr:
Adds ``Stream.activate`` and ``Stream.deactivate`` methods to streams to be triggered just before updating parameters and just after triggering the stream respectively. This allows defining custom actions after an event, e.g. in order to reset the value after an event has been processed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,7 +24,7 @@ install:
   - conda update -q conda
   # Useful for debugging any issues with conda
   - conda info -a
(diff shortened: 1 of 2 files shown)


jlstevens:
I think it is a bit odd that ``activate`` is called only if you use ``update``. This means:

* If you use ``update`` with ``trigger=False``, then ``activate`` is called but ``deactivate`` isn't.
* If you use the ``trigger`` classmethod directly then ``deactivate`` is called by ``activate`` isn't.

Can both ``activate`` and ``deactivate`` be used in the trigger classmethod?

jlstevens:
Looks good. Merging.

PR #934, opened 2016-10-14
Fixed setting of styles on batched bokeh plots

philippjfr:
Batched plots were not correctly respecting styles set on individual Elements. This fixes that.


jlstevens:
Changes make sense - you want to use the highest granularity when accessing style options.

Merging.

PR #936, opened 2016-10-17
Website update

philippjfr:


jlstevens:
Ah, I said I would add a bit of text describing mybinder and holoviews-contrib. Don't merge till I've done that!

philippjfr:
@jlstevens Any update on this?

jlstevens:
@philippjfr  I'll take a look at it now.

jlstevens:
Sorry I didn't get to this earlier! My first change is this:

image

I'm not sure this is the best approach but I think a link to the dev part of the website would be useful.

jlstevens:
My next change is to try and shorten the first paragraph in the Homepage notebook while also linking to holoviews-contrib. I think we agree that the original homepage had a bit too much text before presenting the first visualization. I also think the following paragraph covers the key points.

As this is an edit to the first thing people see on the homepage, we should all make sure we are happy with it before making it live!

jlstevens:
I've also linked to ``holoviews-contrib`` from the tutorials index. I think this conclude the most urgent changes I wanted to make to the website at this time.

We should review all these changes carefully and if we are happy, we can then merge and update the website.

doc/Tutorials/index.rst
```diff
@@ -9,7 +9,7 @@ you may also try out live copies on `mybinder.org
 explore many other examples in our `contrib repository
 <https://github.com/ioam/holoviews-contrib>`_. Lastly, for the most
 responsive experience, you can install HoloViews and try out the
-`tutorial notebooks`_ for yourself.
+tutorial notebooks for yourself.
```
jbednar: I don't know what this was meant to link to, but it wasn't linking to anything anymore.

jlstevens: This isn't a broken link - it gets replaced with the right link based on the current version in setup.py. See the replacement [here](https://github.com/ioam/holoviews/blob/master/doc/conf.py#L31).

jlstevens: It works just fine which is why the dev version is correctly pointing to 1.6.2 - the issue is that we need to copy the files from the dev version over to the main website on release. I *think* what happened is that the website building infrastructure broke so that the files must have been stale.

What I do know is that there are tricky bits to updating the docs/website but thankfully this isn't one of them! :-)

philippjfr:
Thanks for the updates. Merging and updating the website.

PR #939, opened 2016-10-18
Suppress resources warning for failed backend imports

philippjfr:
When a backend can't be imported the resources warning will be suppressed ensuring that only one warning is raised.


jlstevens:
Looks good. Merging.

PR #940, opened 2016-10-18
Ensured Dimensioned.range respects one-sided Dimension.range

philippjfr:
After just fixing one-sided ``Dimension.range`` specifications on ``Dataset`` types I just noticed that ``Dimensioned`` itself does not handle this condition correctly either. This ensures handling for ranges which are None on either side are handled correctly.


jlstevens:
Good to have spotted the issue in both places. Merging.

PR #941, opened 2016-10-19
Bokeh notebook tests

philippjfr:
Ready to merge now. Bokeh_Elements is now tested in Python2 at least.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ install:
   - python setup.py install
   - pip install git+git://github.com/ioam/param.git
   - pip install path.py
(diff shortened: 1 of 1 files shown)


jbednar:
Looks good to me.

jlstevens:
Certainly better than no bokeh tests at all! What would it take to get the bokeh tests working with Python 3?

philippjfr:
Getting it to sort predictably for comparison. I've not really figured out why that's so difficult.

jlstevens:
Python 2's heterogenous sorting vs Python 3's homogenous only sorting?

Issue #601, opened 2016-04-11
Rounding of DynamicMap range causes bug


philippjfr:
Currently due to precision differences in Python vs. Javascript we round values to 10 decimal places. This ensures that the values will match between Python and Javascript, however rounding can result in rounding down or up, which can cause problems. When the upper limit of a dimension is rounded up for example, the rounded value will be outside the limit defined by the dimension range. When moving the slider to the maximum position it will therefore request a frame from Python which falls outside of its defined range and will raise an exception. We should ensure that the rounded values sent to JS are within the original range defined on the dimension.

basnijholt:
I am encountering this bug now. It took me a while to figure out it was not because I was doing something stupid.

Any recommended workarounds?

philippjfr:
No I don't know of any workarounds, but this is a high priority issue and I will try to have it fixed asap.

jlstevens:
I imagine this will be an issue if you have something like ``range=(0, np.pi)`` and after rounding the value goes up, outside that range? If so, I think the solution is to use one function to consistently keep the Python and JS values matching while staying within range. In other words, this issue is annoying but it also doesn't sound like it should be too hard to solve.

basnijholt:
Sorry, I didn't commit the nonworking code, so it's not easy for me to check.

If you like I'll try to reproduce on Monday :)

philippjfr:
I'll close, please reraise the issue if needed.

PR #944, opened 2016-10-21
Fix rounding issues in DynamicMap

philippjfr:
Fixes issue #601 involving a bug where dimension values were rounded up outside a Dimension range before being sent to JS, causing the DynamicMap to throw out of range errors. I now round to 11 decimal places and then throw away the last digit.


philippjfr:
PR build doesn't include the nbpublisher update for the bokeh tests yet. Push build should pass though.

jlstevens:
Throwing the last digit should always reduce the ``abs()`` value so this approach seems fairly sensible. I'm happy to merge if you feel it is ready and tested.

philippjfr:
Yes, ready to merge.

Issue #930, opened 2016-10-13
Allow definition of color palettes using RGB-tuples


mrksr:
Seaborn offers an API to create new color palettes on the fly. These APIs return lists of RGB-Tuples:
```python
In[1]: sns.color_palette()
Out[1]: 
[(0.2980392156862745, 0.4470588235294118, 0.6901960784313725),
 (0.3333333333333333, 0.6588235294117647, 0.40784313725490196),
 (0.7686274509803922, 0.3058823529411765, 0.3215686274509804),
 (0.5058823529411764, 0.4470588235294118, 0.6980392156862745),
 (0.8, 0.7254901960784313, 0.4549019607843137),
 (0.39215686274509803, 0.7098039215686275, 0.803921568627451)]
```
It would be very nice to be able to use them in Holoviews directly:
```python
%%opts Curve (color=hv.Cycle(values=sns.color_palette()))

hv.Overlay([curve1, curve2, curve3])
```
However, `Cycle` (and `Palette`) assume colors to be defined as hex strings (like `#FF0000`) or names (like `green`).
It is possible to convert the seaborn palette like so:
```python
hv.Cycle(values=[hv.plotting.bokeh.util.rgb2hex(c) for c in sns.color_palette()])
```
But this is quite cumbersome. Ideally, Holoviews would just accept RGB tuples anywhere it accepts hex strings.

philippjfr:
@mrksr I just tried it and it seems to work fine for me even without converting to hex. I did fine that declaring the Cycle directly in the output magic doesn't work, e.g.:

``%%opts Curve (color=hv.Cycle(values=sns.color_palette('Set1')))``

Seems to throw an error, but if you declare it separately:

``cycle = hv.Cycle(values=sns.color_palette('Set1'))``

```
%%opts Curve (color=cycle)
hv.Overlay([hv.Curve(np.arange(10)*i) for i in range(3)])
```

It seems to work fine in both matplotlib and bokeh. Could you confirm?

philippjfr:
I've got a fix for the parser so that ``%%opts Curve (color=hv.Cycle(values=sns.color_palette('Set1')))`` just works. Will make a PR later today.

mrksr:
Indeed, defining the cycle separately seems to work fine. Thank you for the fix!

philippjfr:
Once the parser fixes referenced above are merged you'll be able to define it inline in the magic as well. I'll close this issue now though.

Issue #270, opened 2015-09-16
Make OptsSpec parser more robust


philippjfr:
The `OptsSpec` parser for the IPython magic works well for simple types of data but often breaks when you try to pass it more complex data such as dictionaries with string values string keys and values. It is probably a larger project to handle cases like this and won't be tackled any time soon. This issue should be seen as a place to gather examples where the parser gets things wrong so we have a bit more data to work with when someone decides to look at this. 

One example was highlighted in issue #262:

``` python
%%opts Scatter [fontsize={'title':' '10pt'}]
hv.Scatter(((1,2,3),(1,10,100)), group='A')
```

vascotenner:
Can support for comment be added?

    %%opts Scatter [logy=True] #[logx=False]

jlstevens:
Support for comments would be fairly trivial although it means the ``#`` character can't be used anywhere as everything after it would be considered a comment.  I would have to think a bit of the value of supporting comments versus the value of being able to use ``#`` when necessary.

Alternatively, I can imagine that there could be a different comment syntax (e.g ``##`` ) to make such clashes less likely. Right now I am undecided on this issue.

jbednar:
Ooh, that's tricky.  Using a different comment character would be confusing for people, so I would hope that's not necessary.  What about allowing # only if it is escaped (\#)?

jlstevens:
> What about allowing # only if it is escaped (\#)? That way it could still be used if that were ever important...

That seems like a fairly reasonable approach that could be implemented quite easily...

maxalbert:
@jlstevens In which positions do you expect `#` to appear if it wasn't to introduce a comment? Would it only be inside strings? In that case it sound like with a sufficiently flexible parser you should be able to  easily ignore `#` when it occurs inside strings and otherwise treat it as starting a comment. But I may of course be missing something.

maxalbert:
Btw, I hit the same problem which @philippjfr mentioned above, but with a number instead of a string as the value (using the matplotlib backend):
```
%%opts Curve [fontsize={'xlabel': 30}]
Curve([(1, 1), (2, 2)])
```
Without having looked at the code it appears to me that the options parser is trying to re-implement parsing of basic Python syntax, but in a somewhat ad-hoc way which misses some corner cases. Is there any value in using `ast.literal_eval()` for this (see [here](https://docs.python.org/3.5/library/ast.html) for documentation of the `ast` module)?

jlstevens:
[...]

PR #948, opened 2016-10-24
Parser fixes

philippjfr:
Two small fixes for the parser, the first fixes passing dictionary literals by eliminating duplicate commas generated by the parser, fixing part of the issue described in #270, e.g. this is now allowed:

```python
%%opts Scatter [fontsize={'title': '20pt'}]
hv.Scatter(np.random.rand(100,2), group='A')
```

Secondly it fixes a bug when a opts definition includes nested bracketing as described in #930, which described an issue where:

```
%%opts Curve (color=hv.Cycle(values=sns.color_palette('Set1')))
hv.Overlay([hv.Curve(np.arange(10)*i) for i in range(3)])
```

Was not evaluated correctly. This is now also correctly processed.

Note that this does not fix all issues mentioned in #270, e.g. ``aspect=dict(x=1, x=2, z=3)`` still fails to evaluate, but it provides a significant improvement in some cases.

### DIFF
--- a/holoviews/ipython/parser.py
+++ b/holoviews/ipython/parser.py
@@ -47,12 +47,13 @@ class Parser(object):
         recursed = []
         for tok in token:
             if isinstance(tok, list):
-                new_tok = [recurse_token(t, inner) if isinstance(t, list) else t
-                           for t in tok]
+                new_tok = [s for t in tok for s in
+                           (cls.recurse_token(t, inner)
+                            if isinstance(t, list) else [t])]
                 recursed.append((inner % ''.join(new_tok)))
             else:
                 recursed.append(tok)
-        return recursed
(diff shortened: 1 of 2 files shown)


jlstevens:
Parser improvements, yay!

All I will ask is that you add a bunch of unit tests. There are already a [bunch here](https://github.com/ioam/holoviews/blob/master/tests/testparsers.py) but you cannot have too many parser tests.

philippjfr:
Agreed, happy to add tests.

philippjfr:
Added some tests now, which ended up helping me figure out further issues. Still not perfect but it actually let's you do a fair number of more complex expressions now.

jlstevens:
Great to have more fixes and more tests! Merging.

PR #950, opened 2016-10-24
Mapping types complain when supplied with vdims

philippjfr:
As suggested in https://github.com/ioam/holoviews/issues/811, ``NdMapping`` types will now error when you supply any value dimensions (vdims).


jlstevens:
Hopefully this will help avoid confusion. Merging.

Issue #895, opened 2016-10-04
Callable API for stream-enabled Bokeh overlays


jlstevens:
The new bokeh streams system is coming along nicely and working really well for a ton of simple examples. This issue documents discussion I've had with @philippjfr about how we can use bokeh streams in more complex examples. 

I'll start by listing the specific hole in the system we have currently:
- If you use Bokeh streams to get interaction information from the browser...
- ... and if your visualization involves overlays

... then HoloViews faces a problematic ambiguity. Alternatively, if you have two dynamic maps using streams (implicitly defined source elements) you face the same issue, even if they only return simple elements e.g:

``` python
dmap1 * dmap2
```

In essence, there HoloViews has to find the full set of applicable streams, making sure that the relevant information comes from the appropriate part of the plot and is directed to the appropriate DynamicMap.

The true core of this issue, is that a callback is a black box (code) that can return complicated datastructures (e.g overlays). We need to look into that black box to establish the relationships between the applicable streams and the various components of the output.

Our proposal is to define an API for DynamicMap callbacks (specifically callables) that enable Bokeh interactive streams to be used with composite returns (i.e Overlays). We cannot get the information we need from a regular Python function, but an appropriate callable class could expose the API that we require. The overall notion is something like `traverse` for `DynamicMap` - a way to dig into the structure that will be returned by a callback.

For instance, `dmap1 * dmap2` already returns a new `DynamicMap`. After this proposal, the callback of this DynamicMap would be an instance of `OverlayCallable` obeying the `Callable` API. This callback has something like the `traverse` method that lets you collect all the nested streams. For instance, `dmap1` might have been composed from  `dmapA * dmapB`.

 HoloViews would automatically use this `Callable` API as necessary so it is (mostly) invisible to users. Users can continue to use generators, lambda functions or any other callable they want with `DynamicMap`, but if they want to output overlays and streams with Bokeh interaction, they need to follow the `Callable` API. Note, if their callables output simple elements, `dmap1 * dmap2` would now work automatically, with HoloViews handling this machinery behind the scenes.

This could be enforced by an exception message raised by the Bokeh backend when a DynamicMap returns an `Overlay` and has streams specified. The message could be something like: 

'Callback API not found. This API is needed by DynamicMaps that have Bokeh streams enabled and that also return Overlays. For more information about this API see ....'

[...]

PR #951, opened 2016-10-24
Dynamic Callable API

philippjfr:
As described in #895, we needed a way to access streams on DynamicMaps wrapping other DynamicMaps or overlays containing one or more DynamicMaps. This PR implements a ``Callable`` object, which is used to wrap the DynamicMap callbacks generated by all internal operations. This makes it possible to access streams and stream sources on nested DynamicMaps and hook them up appropriately in the plotting code. 

Currently the Callable object is very straightforward, exposing a ``callable_function`` and ``objects`` parameter, which correspond to the callback and the list of objects the callback wraps around, e.g. in an overlaying operation between a ``DynamicMap`` and a ``Points`` Elements, the ``callable_function`` would be a function which overlays the two objects and the ``objects`` a list containing both objects.

Using two helper functions ``get_streams`` and ``get_sources`` we can then look up all the streams and sources that are associated with a particular plot and more specifically a specific layer in an Overlay.

The PR is still lacking docstrings and tests but it already works.

The best example to confirm it is working is this example of two separate selection callbacks attached to two separate data sources:

```python
%%opts Points [tools=['box_select', 'lasso_select', 'tap']]
points = hv.Points(np.random.multivariate_normal((0, 0), [[1, 0.1], [0.1, 1]], (1000,)))
points2 = hv.Points(np.random.multivariate_normal((0, 0), [[1, 0.1], [0.1, 1]], (1000,)))

sel1 = Selection1D()
sel2 = Selection1D()
hv.DynamicMap(lambda index: points, kdims=[], streams=[sel1]) *\
hv.DynamicMap(lambda index: hv.HLine(points['y'][index].mean() if index else -10), kdims=[], streams=[sel1]) *\
hv.DynamicMap(lambda index: points2, kdims=[], streams=[sel2]) *\
hv.DynamicMap(lambda index: hv.HLine(points2['y'][index].mean() if index else -10), kdims=[], streams=[sel2])
```

dual_selection

### DIFF
--- a/holoviews/core/overlay.py
+++ b/holoviews/core/overlay.py
@@ -29,7 +29,7 @@ class Overlayable(object):
                 element = other[args]
                 return self * element
             callback = Callable(callable_function=dynamic_mul,
-                                objects=[self, other])
+                                inputs=[self, other])
             return other.clone(shared_data=False, callback=callback,
                                streams=[])
         if isinstance(other, UniformNdMapping) and not isinstance(other, CompositeOverlay):
--- a/holoviews/core/overlay.py
+++ b/holoviews/core/overlay.py
@@ -24,10 +24,14 @@ class Overlayable(object):
 
     def __mul__(self, other):
         if type(other).__name__ == 'DynamicMap':
-            from ..util import Dynamic
-            def dynamic_mul(element):
+            from .spaces import Callable
+            def dynamic_mul(*args, **kwargs):
+                element = other[args]
                 return self * element
-            return Dynamic(other, operation=dynamic_mul)
+            callback = Callable(callable_function=dynamic_mul,
+                                objects=[self, other])
+            return other.clone(shared_data=False, callback=callback,
+                               streams=[])
         if isinstance(other, UniformNdMapping) and not isinstance(other, CompositeOverlay):
             items = [(k, self * v) for (k, v) in other.items()]
             return other.clone(items)
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -144,7 +144,7 @@ class HoloMap(UniformNdMapping):
             except KeyError:
                 pass
             return Overlay(layers)
-        callback = Callable(callable_function=dynamic_mul, objects=[self, other])
+        callback = Callable(callable_function=dynamic_mul, inputs=[self, other])
         if map_obj:
             return map_obj.clone(callback=callback, shared_data=False,
                                  kdims=dimensions, streams=[])
@@ -211,7 +211,7 @@ class HoloMap(UniformNdMapping):
                     element = self[args]
                     return element * other
                 callback = Callable(callable_function=dynamic_mul,
-                                    objects=[self, other])
+                                    inputs=[self, other])
                 return self.clone(shared_data=False, callback=callback,
                                   streams=[])
             items = [(k, v * other) for (k, v) in self.data.items()]
@@ -401,18 +401,18 @@ class HoloMap(UniformNdMapping):
 
 class Callable(param.Parameterized):
     """
-    Callable allows wrapping callbacks on one or more DynamicMaps such
-    that the operation and the objects that are part of the operation
-    are made available. This allows traversing a DynamicMap that wraps
-    multiple operations and is primarily used to extracting any
-    streams attached to the object.
+    Callable allows wrapping callbacks on one or more DynamicMaps
+    allowing their inputs (and in future outputs) to be defined.
+    This makes it possible to wrap DynamicMaps with streams and
+    makes it possible to traverse the graph of operations applied
+    to a DynamicMap.
     """
 
     callable_function = param.Callable(default=lambda x: x, doc="""
          The callable function being wrapped.""")
 
-    objects = param.List(default=[], doc="""
-         The objects the callable function is processing.""")
+    inputs = param.List(default=[], doc="""
+         The list of inputs the callable function is wrapping.""")
 
     def __call__(self, *args, **kwargs):
         return self.callable_function(*args, **kwargs)
@@ -425,7 +425,7 @@ def get_streams(dmap):
     layer_streams = list(dmap.streams)
     if not isinstance(dmap.callback, Callable):
         return layer_streams
-    for o in dmap.callback.objects:
+    for o in dmap.callback.inputs:
         if isinstance(o, DynamicMap):
             layer_streams += get_streams(o)
     return layer_streams
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -120,12 +120,13 @@ class HoloMap(UniformNdMapping):
             map_obj = self if isinstance(self, DynamicMap) else other
             mode = map_obj.mode
 
-        def dynamic_mul(*key):
+        def dynamic_mul(*key, **kwargs):
             key = key[0] if mode == 'open' else key
             layers = []
             try:
                 if isinstance(self, DynamicMap):
-                    _, self_el = util.get_dynamic_item(self, dimensions, key)
+                    safe_key = () if not self.kdims else key
+                    _, self_el = util.get_dynamic_item(self, dimensions, safe_key)
                     if self_el is not None:
                         layers.append(self_el)
                 else:
@@ -134,7 +135,8 @@ class HoloMap(UniformNdMapping):
                 pass
             try:
                 if isinstance(other, DynamicMap):
-                    _, other_el = util.get_dynamic_item(other, dimensions, key)
+                    safe_key = () if not other.kdims else key
+                    _, other_el = util.get_dynamic_item(other, dimensions, safe_key)
                     if other_el is not None:
                         layers.append(other_el)
                 else:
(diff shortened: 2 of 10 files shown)


jlstevens:
I've skimmed the PR and now I'm looking at it in detail. Overall it looks fine but I need to think about the implications for the user quite carefully.

My first thought was...could we offer a decorator that users can easily use to wrap any callable? A decorator would look nice and clean and would be very easy for the user to add at any point. In fact, couldn't this transformation be applied dynamically? For instance, the wrapped callback could be an underscore attribute in ``DynamicMap`` as soon as it is declared so that the user doesn't even have to worry or know about any of this?

Perhaps you have already done all this this and I'll find out once I look a bit more closely!

holoviews/core/spaces.py
```diff
@@ -393,6 +399,37 @@ def hist(self, num_bins=20, bin_range=None, adjoin=True, individually=True, **kw
                 return histmaps[0]
 
 
+class Callable(param.Parameterized):
+    """
+    Callable allows wrapping callbacks on one or more DynamicMaps such
+    that the operation and the objects that are part of the operation
+    are made available. This allows traversing a DynamicMap that wraps
+    multiple operations and is primarily used to extracting any
+    streams attached to the object.
```
jbednar: to extract

jlstevens:
@philippjfr I would be interested to see the same example you posted above as ``DynamicMap * DynamicMap``. I think that would give me a better insight into how it works as well as a taste of what users might want to do in some situations.

holoviews/core/spaces.py
```diff
@@ -393,6 +399,37 @@ def hist(self, num_bins=20, bin_range=None, adjoin=True, individually=True, **kw
...
+         The list of inputs the callable function is wrapping.""")
+
+    def __call__(self, *args, **kwargs):
+        return self.callable_function(*args, **kwargs)
+
+
+def get_streams(dmap):
+    """
+    Get streams from DynamicMap with Callable callback.
+    """
```
jlstevens: Might say something like 'Get all (potentially nested) streams ...'

holoviews/core/spaces.py
```diff
@@ -689,7 +726,8 @@ def __getitem__(self, key):
 
         # Cache lookup
         try:
-            dimensionless = util.dimensionless_contents(self.streams, self.kdims)
+            dimensionless = util.dimensionless_contents(get_streams(self),
+                                                        self.kdims, False)
             if (dimensionless and not self._dimensionless_cache):
```
jlstevens: Using ``no_duplicates=False`` would be clearer here...

holoviews/core/spaces.py
```diff
@@ -393,6 +399,37 @@ def hist(self, num_bins=20, bin_range=None, adjoin=True, individually=True, **kw
...
+
+    inputs = param.List(default=[], doc="""
+         The list of inputs the callable function is wrapping.""")
+
+    def __call__(self, *args, **kwargs):
+        return self.callable_function(*args, **kwargs)
+
+
+def get_streams(dmap):
+    """
```
jlstevens: Would ``get_nested_streams`` or ``nested_streams`` be a better name?

holoviews/plotting/util.py
```diff
@@ -295,11 +296,43 @@ def attach_streams(plot, obj):
...
             stream._hidden_subscribers.append(plot.refresh)
     return obj.traverse(append_refresh, [DynamicMap])
 
 
+def get_sources(obj, index=None):
+    """
+    Traverses Callable graph to resolve sources on
+    DynamicMap objects, returning a list of sources
+    indexed by the Overlay layer.
+    """
```
jlstevens: Can't this be simplified to:

```python
if isinstance(obj, DynamicMap) and isinstance(obj.callback, Callable):
    return [(index, obj)]
```

leaving the rest of the function to handle ``DynamicMaps`` with ``Callable`` callbacks?

holoviews/plotting/widgets/__init__.py
```diff
@@ -108,7 +108,11 @@ def __init__(self, plot, renderer=None, **params):
         super(NdWidget, self).__init__(**params)
         self.id = plot.comm.target if plot.comm else uuid.uuid4().hex
         self.plot = plot
-        self.dimensions, self.keys = drop_streams(plot.streams,
```
jlstevens: I assume this bit is simply a bug fix and otherwise unrelated?

holoviews/core/spaces.py
philippjfr: Sure.

holoviews/util.py
```diff
@@ -69,15 +72,17 @@ def _dynamic_operation(self, map_obj):
...
-            return self._process(el, key)
-
-        return dynamic_operation
+        else:
+            def dynamic_operation(*key, **kwargs):
+                key = key[0] if map_obj.mode == 'open' else key
+                self.p.kwargs.update(kwargs)
+                _, el = util.get_dynamic_item(map_obj, map_obj.kdims, key)
+                return self._process(el, key)
+        if isinstance(self.p.operation, ElementOperation):
```
jlstevens: Couldn't ``self.p.operation`` be folded into ``callable_function`` so you only need ``Callable`` and not ``OperationCallable``?

holoviews/core/overlay.py
```diff
@@ -24,10 +24,14 @@ class Overlayable(object):
...
-            from ..util import Dynamic
-            def dynamic_mul(element):
+            from .spaces import Callable
+            def dynamic_mul(*args, **kwargs):
+                element = other[args]
                 return self * element
-            return Dynamic(other, operation=dynamic_mul)
+            callback = Callable(callable_function=dynamic_mul,
+                                inputs=[self, other])
+            return other.clone(shared_data=False, callback=callback,
```
philippjfr: >Wondering whether it should be self.clone, or other.clone or maybe a new DynamicMap declaration entirely. I see this is in the condition where other is a DynamicMapbut is this definitely right in terms of kdims?

Yes, this is the condition where self is a single Element or Overlay.


>If all the logic regarding dynamic could move to _dynamic_mul, that would be cleaner...

This is the ``__mul__`` implementation on Overlayable, it doesn't have ``_dynamic_mul``, because I'd like to avoid inline imports.

holoviews/plotting/bokeh/element.py
```diff
@@ -177,15 +177,18 @@ def _construct_callbacks(self):
...
+                       if i in [None, self.zorder]]
         else:
-            source = self.hmap.last
-        streams = Stream.registry.get(id(source), [])
-        registry = Stream._callbacks['bokeh']
-        callbacks = {(registry[type(stream)], stream) for stream in streams
-                     if type(stream) in registry and streams}
+            sources = [(self.zorder, self.hmap.last)]
+        cb_classes = set()
+        for _, source in sources:
```
jlstevens: So the key difference seems to be that you can now have multiple sources whereas before there was only one. And the sources are now retrieved recursively via ``get_sources`` which works via the ``Callable`` instances...

holoviews/plotting/widgets/__init__.py
philippjfr: Before this wasn't an issue, so not unrelated.

holoviews/plotting/util.py
philippjfr: Not quite but it could definitely be refactored more nicely.

holoviews/core/spaces.py
philippjfr: Sure.

holoviews/plotting/util.py
jlstevens: Ah...

```
 +   if isinstance(obj, DynamicMap):
 +        if isinstance(obj.callback, Callable):
                  ....
 +        else:
 +            return [(index, obj)]
 +    else:
 +        return [(index, obj)]
```

I think I meant:

```
if not isinstance(obj, DynamicMap) or not isinstance(obj.callback, Callable):
    return [(index, obj)]
```

holoviews/plotting/widgets/__init__.py
jlstevens: Ok...seems to be a result of ``plot.streams`` potentially having many more instances due to the nesting than before...

philippjfr: Right in these cases it just needs to know if a kdim has a corresponding stream or not, there is no actual clash at the DynamicMap level because each level of wrapping resolves it's own streams.

holoviews/util.py
jlstevens: I'm a little confused. ``_dynamic_operation`` seems to be called once [here](https://github.com/ioam/holoviews/blob/ca945bf92e41bcfe9d3975bd53fbdbf4719f01a3/holoviews/util.py#L36) and I don't see the ``operation`` parameter then being inspected/used on the ``callback`` variable anywhere in ``Dynamic``. I am yet to spot where the ``operation`` parameter of ``OperationCallable`` is used...so I'm assuming I've missed it.

philippjfr:
Made the changes you suggested.

holoviews/util.py
philippjfr: It is never used, but is useful information and will in future be used to look up the ``outputs``.

philippjfr: Ah sorry, I'm confused different bit of code.

philippjfr:
```python
if isinstance(self.p.operation, ElementOperation):
            kwargs = {k: v for k, v in self.p.kwargs.items()
                      if k in self.p.operation.params()}
            return self.p.operation.process_element(element, key, **kwargs)
```

jlstevens:
Ok, I feel I mostly understand it now... the most magical bit (i.e the bit I couldn't figure out!) seems to be in ``get_sources`` but everything else makes sense.

I'll merge now on the understanding that we now have a plan to improve the user API side of things. This PR shows that this approach will allow nested ``DynamicMaps`` to work properly with streams.

Issue #530, opened 2016-03-02
Overlays should support multiple HoloMaps


philippjfr:
While Overlays should not generally contain HoloMaps during plotting it is often useful to construct them that way and then let collation handle the correct nesting during display. However currently Overlays fail entirely when passed more than one HoloMap as is the case in this example:

```python
xs, ys, zs = np.meshgrid(range(30), range(250), range(25))
data = np.cumsum(np.random.normal(size=(30, 250, 25, 2)), axis=1)
columns = np.column_stack([xs.flat, ys.flat, zs.flat, data.reshape(-1, 2)])

dtable = hv.Table(columns, kdims=["Time", "Particle", "Horizon"], vdims=list("ab"))

hv.Overlay([
    dtable.reduce(['Particle'], fn).to.curve('Horizon', 'a', label=name)
    for name, fn in [('max', np.max), ('mean', np.mean), ('min', np.min)]
])
```

```python
d:\repos\holoviews\holoviews\core\tree.py in __setattr__(self, identifier, val)
    179         # Getattr is skipped for root and first set of children
    180         shallow = (self.parent is None or self.parent.parent is None)
--> 181         if identifier[0].isupper() and self.fixed and shallow:
    182             raise AttributeError(self._fixed_error % identifier)
    183 

IndexError: invalid index to scalar variable.
```

The Overlay should just work and the Collator should resolve the improper nesting during plotting.

Note there is a workaround using the mul operator to combine the HoloMaps of Elements into HoloMaps of Overlays:

```python
import operator
reduced = [dtable.reduce(['Particle'], fn).to.curve('Horizon', 'a', label=name)
           for name, fn in [('max', np.max), ('mean', np.mean), ('min', np.min)]]
reduce(operator.mul, reduced)

philippjfr:
The PR referenced above ensures that HoloMaps and GridSpaces can now be nested inside an Overlay. Using a new ``collate`` method on Overlays it is possible to then get them to be organized according to the recommended nesting structure. Closing.

PR #952, opened 2016-10-25
Overlay collate

philippjfr:
This PR addresses issue #530, which suggested that creating an Overlay of HoloMaps to make it easy to declare certain datastructures. This PR reimplements the ``Layout.from_values`` method to be more efficient by not creating a bunch of intermediate objects, which also makes it possible to create an Overlay of HoloMaps and GridSpaces. Additionally it adds a trivial ``collate`` implementation which simply mirrors the old ``from_values`` implementation by reducing all the objects with the ``__mul__`` operator. Finally it sets up Overlays of HoloMaps and GridSpaces to be automatically collated and issue a warning like we already do for other invalid nesting structures.

### DIFF
--- a/holoviews/core/layout.py
+++ b/holoviews/core/layout.py
@@ -367,10 +367,6 @@ class Layout(AttrTree, Dimensioned):
 
 
     @classmethod
-    def _from_values(cls, val):
-        return reduce(lambda x,y: x+y, val).display('auto')
-
-    @classmethod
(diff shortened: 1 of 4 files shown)


jlstevens:
Having things work but issue a warning is something we already partially implemented and this PR completes that (or at least, I can't currently think of any other similar collation+warning scenario).

The increased efficiency is also nice! Merging.

PR #953, opened 2016-10-25
Correctly deduplicate tools in bokeh OverlayPlots

philippjfr:
Currently when the user or a stream callback passes a bokeh tool instance to plots in an overlay duplicates can occur issuing bokeh warnings. This PR ensures callbacks on the tools are appropriately merged.


philippjfr:
Bokeh tests have changed because the tool order is now well defined. Will try to rebuild.

jlstevens:
Good to see those annoying duplicate tools are fixed. Tests are still failing though...

philippjfr:
The approach taken here is fundamentally broken. Closing.

Issue #565, opened 2016-03-23
Setting colorbar ticks


vascotenner:
I tried to set the number of ticks on a colorbar, using the mpl backend:
```python
%%opts Image [colorbar=True, cbar_ticks=5]
hv.Image(np.array([[1,.5],[0,.9]]))
```

This raises in the following error:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/home/tenner/.local/lib/python2.7/site-packages/IPython/core/formatters.pyc in __call__(self, obj)
    337                 pass
    338             else:
--> 339                 return printer(obj)
    340             # Finally look for special method names
    341             method = _safe_get_formatter_method(obj, self.print_method)

/home/tenner/src/holoviews/holoviews/ipython/display_hooks.pyc in pprint_display(obj)
    222     if not ip.display_formatter.formatters['text/plain'].pprint:
    223         return None
--> 224     return display(obj, raw=True)
    225 
    226 

/home/tenner/src/holoviews/holoviews/ipython/display_hooks.pyc in display(obj, raw, **kwargs)
    204         html = grid_display(obj)
    205     elif isinstance(obj, (CompositeOverlay, ViewableElement)):
--> 206         html = element_display(obj)
    207     elif isinstance(obj, (Layout, NdLayout, AdjointLayout)):
    208         html = layout_display(obj)

/home/tenner/src/holoviews/holoviews/ipython/display_hooks.pyc in wrapped(element)
    136                        'Call ipython.show_traceback() for details')
    137                 return msg % e
--> 138             raise e
    139 
    140     return wrapped

AttributeError: Colorbar instance has no attribute 'set_major_locator'
```

philippjfr:
Thanks, will look into it.

philippjfr:
Fixed in the PR referenced above, closing.

PR #954, opened 2016-10-26
Correctly set matplotlib colorbar tick locators

philippjfr:
Fixes issue #565, the locators were being set using a non-existent method.


jlstevens:
Looks good. Merging.

PR #956, opened 2016-10-27
Fixed confusing aliasing

jbednar:
The attribute "colors" was being used in the same function both to refer to the matplotlib colors module and to a local dictionary.  In some cases (not easily reproducible) this led to complaints about variables being referenced before they were assigned. In any case this change makes it much clearer what is meant by "colors" each time.


jbednar:
The tests failed for Python3, but that seems to be unrelated to this PR, as it's a failure in a bokeh file and this is a change to the mpl backend.

philippjfr:
Looks fine, thanks for the fix. The test failure was is a transient issue in the bokeh backend.

PR #959, opened 2016-10-31
Bokeh server support

philippjfr:
Adds support for deploying any holoviews object as a bokeh server app.
- [x] BokehRenderer initializes server document
- [x] Initial support for handling stream events

To do:
- [x] Bokeh widget implementation to support DynamicMaps and HoloMaps with standard streams
- [x] Ensure that bokeh stream callbacks relying on `cb_data` and `cb_obj` work correctly.
- [x] Add example apps.

Optional/Future:
- Provide easy way to tie a stream to any bokeh widget.
- Support bokeh widgets in a notebook context.

Example app:

```python
import numpy as np
import holoviews as hv
import holoviews.plotting.bokeh

hv.Store.current_backend = 'bokeh'
renderer = hv.Store.renderers['bokeh'].instance(mode='server', holomap='server')
points = hv.HoloMap({i: hv.Points(np.random.multivariate_normal((0,0), [[0.1, 0.1], [0.1, 1.0]], (1000000,))) for i in range(10)})
datashaded = datashade(points, x_sampling=0.01, y_sampling=0.01)
doc, _ = renderer(datashaded)
doc.title = 'HoloViews Datashade'
```

hv_datashade_server


jlstevens:
Nice to have the examples!

That said, I feel they belong in ``contrib`` which is an entire repo dedicated to various examples. I would put them in the ``scripts`` directory [here](https://github.com/ioam/holoviews-contrib/tree/master/scripts).

philippjfr:
I do like the idea of contrib, but until it's been integrated into our doc building process and referenced throughout our documentation it just seems like no one will ever find it.

jlstevens:
Given that we want to split out the bokeh backend eventually and this is definitely bokeh related, it should be in ``plotting.bokeh.examples`` if you don't want it in ``contrib``.

philippjfr:
>it should be in plotting.bokeh.examples if you don't want it in contrib.

I did think ``contrib`` was the right place, but ``plotting.bokeh.examples`` might actually be even better if we are going to split it out.

jlstevens:
I think it would be nice to keep all the examples together in ``contrib`` as it would show the breadth of what you can do with HoloViews. There is no restriction in ``contrib`` on backends or any other libraries necessary to make the examples work.

For this reason, I think we should move these examples to ``contrib`` right now - the bokeh backend repo could always link to the contrib examples to ensure people can discover them.

philippjfr:
@jlstevens, @jbednar I've rebased this PR with the latest events PR and it's working as I've wanted it to. The question is now about scope, I'll happily move the current examples to contrib, but my main question is about bokeh widgets. As of right now this PR ships with a small class that emulates our regular Dimension based widgets, and you can easily connect streams to other widgets and use the regular bokeh API to compose your new set of widgets into a layout with the plot(s).

Should we ship a basic widget class which generates widgets for HoloMap/DynamicMap key dimensions? You can see what that looks like in the gif example above. The implementation is fairly straightforward (~130 LOC) and means you can directly deploy standard HoloMaps/DynamicMaps you used and defined in your browser without manually having to hook up widgets for key dimensions.

jbednar:
That all sounds good, but is there a reason not to just use Bokeh widgets all the time when we are using the Bokeh backend, even in the notebook? That seems like the way to have seamless interoperability with bokeh in its various possible usages.

philippjfr:
If we had already done the widget refactor and implemented a general widget/comm manager as we are planning to do I'd say yes. However for the time being I'd have to come up with some ad hoc solution which I'm strongly against. So my answer is that once we have that we can easily retrofit this widget implementation to support it. 

In the long run we should find an easy and general way to map both dimensions and streams to specific widgets.

philippjfr:
Waiting on https://github.com/bokeh/bokeh/pull/6062 so that the ``cb_obj`` can be handled the same for the ``js_on_event`` and ``on_event`` cases.

I've also implemented an event queue on the python end which works fairly well. Once the PR has been merged I'll have to extend that implementation to handle different event types as I did on the JS end.

philippjfr:
@jlstevens This is now ready for review. All streams now work for both bokeh server and in the notebook. I'm happy to move the example apps elsewhere though.

jlstevens:
My first comment is that I think ``examples/apps/crossfilter.py`` should either go to contrib for now or move to ``plotting/bokeh/examples``. These examples are very bokeh specific so make I prefer the latter option as it would also make sense once we split out the plotting backends.

holoviews/plotting/bokeh/widgets.py
```diff
@@ -1,12 +1,178 @@
...
+from bokeh.layouts import layout, gridplot, widgetbox, row, column
 
+from ...core import Store, NdMapping, OrderedDict
+from ...core.util import drop_streams, unique_array, isnumeric, wrap_tuple_streams
 from ..widgets import NdWidget, SelectionWidget, ScrubberWidget
 from .util import serialize_json
 
+
+
+class BokehServerWidgets(param.Parameterized):
```
jlstevens: There needs to be a base class with at least ``basejs`` and ``extensionjs`` for widgets. Inheriting from ``param.Parameterized`` made me think this was a new class type and that the code using these parameters was a new mechanism.

philippjfr: There is, I can probably inherit from it without problem.

holoviews/plotting/renderer.py
```diff
@@ -298,7 +299,9 @@ def get_widget(self_or_cls, plot, widget_type, **kwargs):
         if not isinstance(plot, Plot):
             plot = self_or_cls.get_plot(plot)
         dynamic = plot.dynamic
-        if widget_type == 'auto':
+        if widget_type == 'server':
```
jlstevens: I would rather avoid introducing a new widget type - why not just switch to bokeh widgets if the renderer is in 'server' mode?

philippjfr: Agreed, would definitely be good to find a better solution here.

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -278,7 +298,7 @@ def _filter_msg(self, msg, ids):
             else:
                 filtered_msg[k] = v
         return filtered_msg
-
+
```
jlstevens: Trailing whitespace...

holoviews/plotting/bokeh/widgets.py
```diff
@@ -1,12 +1,178 @@
...
+                                for d, item in self.lookups.items()}
+        self.subplots = {}
+        if self.plot.renderer.mode == 'default':
+            self.attach_callbacks()
+        self.state = self.init_layout()
+        self._event_queue = []
+
+
+    def get_widgets(self):
+        # Generate widget data
```
jlstevens: What I find vexing about ``get_widgets`` is that I see what I expect to be entirely general purpose code for turning a dimension into some sort of widget instance mixed up with other logic. This is probably not specific to this PR but ``get_widgets`` wherever it is implemented.

 I would like there to be an entirely general classmethod/utility that takes a single dimension and returns an appropriate widget. Looking at the code below, I *think* that would be possible if it weren't for the ``lookup``/``lookups`` thing that I don't fully understand yet.

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -229,39 +229,59 @@ class Callback(object):
...
                     continue
-                handle = handles[handle_name]
-                js_callback = self.get_customjs(requested)
-                self.set_customjs(js_callback, handle)
-                self.callbacks.append(js_callback)
+                handle = self.plot_handles[handle_name]
+
+                # Hash the plot handle with Callback type allowing multiple
+                # callbacks on one handle to be merged
+                cb_hash = (id(handle), id(type(self)))
```
jlstevens: Note that the ``id`` isn't really a hash but it is supposed to be unique...for a given Python session on a given architecture. I suspect it is ok here as I doubt this information is expected to persist in anyway between sessions.

holoviews/plotting/bokeh/widgets.py
```diff
@@ -1,12 +1,178 @@
...
+            widgets[dim.pprint_label] = (label, widget)
+            if lookup:
+                lookups[dim.pprint_label] = OrderedDict(lookup)
+        return widgets, lookups
+
+
+    def init_layout(self):
+        widgets = [widget for d in self.widgets.values()
+                   for widget in d if widget]
+        wbox = widgetbox(widgets, width=200)
```
jlstevens: Just noting the magic number 200...

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -330,6 +350,75 @@ def _get_stream_handle_ids(self, handles):
...
+                raise Exception('Bokeh plot attribute %s could not be found' % path)
+            for p in attr_path[1:]:
+                if p == 'attributes':
+                    continue
+                if isinstance(attr_val, dict):
+                    attr_val = attr_val.get(p)
+                else:
+                    attr_val = getattr(attr_val, p, None)
+            values[attr] = {'id': model_obj.ref['id'],
+                            'value': attr_val}
```
jlstevens: I think there can be a utility here. Something like:

```python
get_message(attributes, events=True)
```

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -330,6 +350,75 @@ def _get_stream_handle_ids(self, handles):
...
+        self.on_msg(values)
+        self.plot.document.add_timeout_callback(self.trigger, 50)
+
+
+    def set_onchange(self, handle):
+        """
+        Set up on_change events for bokeh server interactions.
+        """
+        if self.events and bokeh_version >= '0.12.5':
```
jlstevens: ``self.events`` should be ``self.on_events`` and similarly for ``on_change``.

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -330,6 +350,75 @@ def _get_stream_handle_ids(self, handles):
...
+                    attr_val = attr_val.get(p)
+                else:
+                    attr_val = getattr(attr_val, p, None)
+            values[attr] = {'id': model_obj.ref['id'],
+                            'value': attr_val}
+        self.on_msg(values)
+        self.plot.document.add_timeout_callback(self.trigger, 50)
+
+
+    def set_onchange(self, handle):
```
jlstevens: The name is misleading now, should be something like ``set_bokeh_callback``

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -330,6 +350,75 @@ def _get_stream_handle_ids(self, handles):
...
+        """
+        Process bokeh UIEvents adding timeout to process multiple concerted
+        value change at once rather than firing off multiple plot updates.
+        """
+        self._event_queue.append((event))
+        if self.trigger not in self.plot.document._session_callbacks:
+            self.plot.document.add_timeout_callback(self.trigger, 50)
+
+
+    def trigger(self):
```
jlstevens: I would consider splitting this into ``trigger_change`` and ``trigger_event``.

holoviews/plotting/bokeh/callbacks.py
jlstevens: Given the suggestion of splitting trigger, maybe ``get_event_message(attributes)`` and ``get_change_message(attributes)``.

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -330,6 +350,75 @@ def _get_stream_handle_ids(self, handles):
...
+        """
+        Trigger callback change event and triggering corresponding streams.
+        """
+        if not self._event_queue:
+            return
+        if self.events:
+            event = self._event_queue[-1]
+        self._event_queue = []
```
jlstevens: Probably worth just calling it ``queue`` as the word ``event`` is now confusing...

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -330,6 +350,75 @@ def _get_stream_handle_ids(self, handles):
...
+
+
+    def on_event(self, event):
+        """
+        Process bokeh UIEvents adding timeout to process multiple concerted
+        value change at once rather than firing off multiple plot updates.
+        """
+        self._event_queue.append((event))
+        if self.trigger not in self.plot.document._session_callbacks:
+            self.plot.document.add_timeout_callback(self.trigger, 50)
```
jlstevens: The only difference between these is how the queue is used. I think it is possible to simplify. e.g

``def debounce(change_spec=None, event_spec=None):
   if change_spec: # Use trigger_change and add to queue appropriately
        callback = self.trigger_change
   elif event_spec: # Use trigger_change and add to queue appropriately
       callback = self.trigger_event
   self.plot.document.add_timeout_callback(callback, 50)
```

Comment (philippjfr) [inline holoviews/plotting/bokeh/callbacks.py]: No, can't do that, bokeh validates the signature of callbacks.

Comment (jlstevens) [inline holoviews/plotting/bokeh/callbacks.py]: Won't ``self.triggr_change`` and ``self.trigger_event`` have the appropriate signatures?

File: holoviews/plotting/bokeh/callbacks.py [hunk 13]
```diff
@@ -187,7 +187,7 @@ class Callback(object):
         }}
 
         // Add current event to queue and process queue if not blocked
```
philippjfr: Worked out quite well.

jbednar: Great!

holoviews/plotting/bokeh/widgets.py
philippjfr: Will make it a parameter.

holoviews/plotting/renderer.py
jlstevens: Have you found a way to avoid introducing this widget type?

holoviews/plotting/bokeh/widgets.py
jlstevens: Was there a problem? It is still inheriting from ``param.Parameterized``...

holoviews/plotting/renderer.py
jlstevens: Looks like this was addressed.

holoviews/plotting/bokeh/callbacks.py
jlstevens: Fixed.

holoviews/plotting/bokeh/widgets.py
jlstevens: This was fixed.

holoviews/plotting/bokeh/widgets.py
philippjfr: Hmm, I guess I can create the baseclass but it's all kind of pointless, while Renderer is expecting these attributes they're not actually used in the server case.

jlstevens: Ok it doesn't matter so much - it is basically a mix-in class now.

holoviews/plotting/bokeh/widgets.py
jlstevens: This was addressed.

holoviews/plotting/bokeh/callbacks.py
jlstevens: Addressed.

holoviews/plotting/bokeh/widgets.py
```diff
@@ -1,12 +1,208 @@
...
+        self.mock_obj = NdMapping([(k, None) for k in self.keys],
+                                  kdims=self.dimensions)
+        self.widgets, self.lookups = self.get_widgets()
+        self.reverse_lookups = {d: {v: k for k, v in item.items()}
+                                for d, item in self.lookups.items()}
+        self.subplots = {}
+        if self.plot.renderer.mode == 'default':
+            self.attach_callbacks()
+        self.state = self.init_layout()
+        self._event_queue = []
```
jlstevens: I would still rename this from ``_event_queue`` to just ``queue`.

jlstevens: Ah, this is for the widgets now.. so it might be ok.

philippjfr: Although maybe better for consistency? I'll rename.

holoviews/plotting/bokeh/callbacks.py
jlstevens: Addressed.

holoviews/plotting/bokeh/callbacks.py
jlstevens: Addressed. There is an ``_event_queue`` in the widget class though...

holoviews/plotting/bokeh/callbacks.py
jlstevens: Addressed.

jlstevens: Addressed.

holoviews/plotting/bokeh/callbacks.py
jlstevens: Addressed.

holoviews/plotting/bokeh/callbacks.py
jlstevens: Addressed - I now see why this won't work.

holoviews/plotting/bokeh/widgets.py
jlstevens: This is in widgets now so my original comment was addressed. Might be worth renaming here too - up to you.

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -4,115 +4,143 @@
...
 
-    * events      : If the Callback should listen to bokeh events this
-                    should declare the types of event as a list (optional)
 
-    * change      : If the Callback should listen to model attribute
-                    changes on the defined ``models`` (optional)
+    def _init_plot_handles(self):
+        """
+        Find all requested plotting handles and cache them along
+        with the IDs of the models callbacks will be attached to.
```
jlstevens: Either an an extra word here or words missing. Either '... along with the IDs the model callbacks will be attached to' or '...along with the IDs of the model callbacks that they will be attached to'.

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -350,78 +303,258 @@ def get_customjs(self, references):
...
+                model_obj = self.plot_handles.get(self.models[0])
+                msg[attr] = self.resolve_attr_spec(path, event, model_obj)
+            self.on_msg(msg)
+        self.plot.document.add_timeout_callback(self.process_on_event, 50)
+
+
+    def process_on_change(self):
+        if not self._queue:
+            return
+        self._queue = []
```
jlstevens: How come the ``_queue`` isn't used for ``process_on_change``?

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -350,78 +303,258 @@ def get_customjs(self, references):
...
+                    changes on the defined ``models`` (optional)
+
+    If either on_events or on_changes are declared the Callback will
+    be registered using the on_event or on_change machinery, otherwise
+    it will be treated as a regular callback on the model.  The
+    callback can also define a _process_msg method, which can modify
+    the data sent by the callback before it is passed to the streams.
+    """
+
+    def initialize(self):
```
jlstevens: *Much* better! :-)

holoviews/plotting/bokeh/callbacks.py
philippjfr: Neither, it's missing a ``the``:

>along with the IDs of the models the callbacks will be attached to.

jlstevens: That works too. I knew something was wrong!

holoviews/plotting/bokeh/callbacks.py
philippjfr: Because process_on_change just looks up the model values directly, that avoids having separate callbacks for each model change, which would be required otherwise.

holoviews/plotting/bokeh/renderer.py
```diff
@@ -28,22 +29,38 @@ class BokehRenderer(Renderer):
...
 
+    holomap = param.ObjectSelector(default='auto',
+                                   objects=['widgets', 'scrubber', 'server',
+                                            None, 'auto'], doc="""
+        Output render multi-frame (typically animated) format. If
+        None, no multi-frame rendering will occur.""")
+
+    mode = param.ObjectSelector(default='default',
+                                objects=['default', 'server'], doc="""
+         Whether to render the DynamicMap in regular or server mode. """)
```
jlstevens: Is it only rendering DynamicMaps? Might want a bit more to say what server mode is about ...

philippjfr: Hmm, not sure why I wrote that. It handles anything.

holoviews/plotting/bokeh/widgets.py
```diff
@@ -1,12 +1,209 @@
...
+                                for d, item in self.lookups.items()}
+        self.subplots = {}
+        if self.plot.renderer.mode == 'default':
+            self.attach_callbacks()
+        self.state = self.init_layout()
+        self._queue = []
+
+
+    @classmethod
+    def create_widget(self, dim, holomap=None):
```
jlstevens: Much nicer as a classmethod! Good for testing...

jlstevens:
Fantastic!

Other than one docstring fix I am now very happy with this PR. The functionality is great and the code is now structured in a much more sensible way.

I'll merge once that is done.

jlstevens:
All done! Merging.

PR #962, opened 2016-10-31
Add short timeout to bokeh stream callbacks

philippjfr:
The short timeout handles events which are meant to be processed together but execute in rapid succession, e.g. x_range and y_range events.


jlstevens:
Looks good. I guess you can always try to play with the debounce value a little but 20 milliseconds seems reasonable. Merging.

PR #963, opened 2016-10-31
Small fixes and optimizations for datashader operations

philippjfr:
Includes various improvements/optimizations for the datashader operations avoiding making copies of the data, allowing different Element types to be used and providing an explicit color key dictionary.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -22,13 +22,20 @@ from ..element import GridImage, Path, Curve, Contours, RGB
 from ..streams import RangeXY
 
 
+def is_df(obj):
+    """
+    Check if the Element is a Dataset type wrapping DataFrame-like data.
+    """
+    return isinstance(obj, Dataset) and obj.interface is PandasInterface
+
+
 @dispatch(Element)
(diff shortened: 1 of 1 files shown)


holoviews/operation/datashader.py
```diff
@@ -259,6 +277,8 @@ def _process(self, element, key=None):
             categories = array.shape[-1]
             if not self.p.cmap:
                 pass
+            elif isinstance(self.p.cmap, dict):
+                shade_opts['color_key'] = self.p.cmap
             elif isinstance(self.p.cmap, Iterable):
                 shade_opts['color_key'] = [c for i, c in
                                            zip(range(categories), self.p.cmap)]
```
jbednar: Hmm; datashader keeps the cmap and color_key options separate so that there can be a default for each one of them separately, but it looks like you are using cmap for both?

philippjfr: We don't have a default ``cmap`` so it just ends up using the appropriate datashader default.

jbednar: Ok, I guess that works out then.

PR #964, opened 2016-11-01
Fixed bug with duplicate streams

philippjfr:
Ensure that plotting code is not given any duplicate streams.


jlstevens:
I tested this branch out on the notebook I was having issues with and it all works now. Merging.

PR #968, opened 2016-11-02
Optimized dynamic bokeh plots

philippjfr:
Previously we added some optimization that avoids sending updated plot data if the data source is unchanged between frames. However the plotting class still went through the effort of updating the datasource even though it is unchanged, which can in some cases be fairly expensive in itself. This generalizes the previous optimization such that the data is only updated in dynamic mode if it has changed.


jlstevens:
As a generalization of a previous optimization, this seems ok. The core condition seems to be:

```
self.static_source = self.dynamic and (current_id == previous_id)
```

Looking at the code, it seems that this ``self.static_source`` attribute now controls whether the datasource is updated *and* whether the (un)changed plot data is sent (or not). Overall this seems clearer than before (and more efficient) so I will now merge.

PR #969, opened 2016-11-04
Counter mode deprecation

jlstevens:
This PR is work-in-progress and aims to remove the ``DynamicMap`` counter mode now that streams offer a much more principled way of achieving the same thing.

**TODO**:

- [ ]  Update the DynamicMap tutorial.
- [ ] Test this works properly with other notebooks using DynamicMap.
- [ ] Make sure we are clear about ``get_dynamic_item`` and associated semantics.
- [ ] We probably no longer need both ``mode`` and ``call_mode``.
- [ ] Before the assumption was that is ``call_mode='key'`` then the mode was ``'bounded'``. With streams, this is not quite right: bounded should apply if any of the kdims are bounded.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -462,16 +462,12 @@ class DynamicMap(HoloMap):
 
         For open mode where there is an unbounded key dimension, the
         return type can specify a key as well as element as the tuple
-        (key, element). If no key is supplied, a simple counter is used
-        instead.
+        (key, element).
 
         If the callback is a generator, open mode is used and next() is
-        simply called. If the callback is callable and in open mode, the
-        element counter value will be supplied as the single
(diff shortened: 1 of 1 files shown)


jlstevens:
Unit tests need updating and the DynamicMap tutorial also needs to be updated. I've checked the DynamicMap tutorial works up to the point where counter mode is introduced. Seems like we'll have to update the notebook test data for this one...

philippjfr:
Going to close this as it has been superseded by #1238.

jlstevens:
I think this point is still valid:

- [ ]  Make sure we are clear about get_dynamic_item and associated semantics.

philippjfr:
Sure, we can discuss that in #1240, which reimplements ``get_dynamic_item``.

Issue #965, opened 2016-11-01
Dynamic *  doesn't work in open stream mode


jlstevens:
The following doesn't work as expected:

```python
import holoviews as hv
import param
from holoviews import streams
import numpy as np
hv.notebook_extension()
```

```python
class Time(streams.Stream):
    time = param.Number(default=1)
    
time_stream = Time()
tdim = 'time'
dmap1 = hv.DynamicMap(lambda time: hv.Image(np.random.rand(10,10)*time), kdims=[tdim], streams=[time_stream])
dmap2 = hv.DynamicMap(lambda time: hv.Image(-np.random.rand(10,10)*time), kdims=[tdim], streams=[time_stream])
dmap1 * dmap2
```

I get:

```python
/Users/jstevens/Desktop/development/holoviews/holoviews/plotting/mpl/raster.pyc in update_handles(self, key, axis, element, ranges, style)
     79     def update_handles(self, key, axis, element, ranges, style):
     80         im = self.handles['artist']
---> 81         data, style, axis_kwargs = self.get_data(element, ranges, style)
     82         l, r, b, t = style['extent']
     83         im.set_data(data[0])

/Users/jstevens/Desktop/development/holoviews/holoviews/plotting/mpl/raster.pyc in get_data(self, element, ranges, style)
     65         else:
---> 66             l, b, r, t = element.extents
     67         if self.invert_yaxis and type(element) is Raster:
     68             b, t = t, b

ValueError: need more than 0 values to unpack
```

This suggests that ``element`` is an Overlay so ``element.extents`` is creating a new overlay. But if I give the time dimension a range:

```python
tdim = hv.Dimension('time', range=(1,100))
```

It now works as expected. Setting the range shouldn't be necessary to make dynamic overlays work in open mode...

jlstevens:
I encountered another confusing issue. In the example above, if you try to use the opts magic as follows:

```python
%%opts Image {+axiswise}
```

I get a traceback ending in:

```python
/Users/jstevens/Desktop/development/holoviews/holoviews/plotting/plot.pyc in _get_frame(self, key)
    614         elif self.dynamic:
    615             with dimensionless_cache(self.hmap, not self._force or not self.drawn):
--> 616                 key, frame = util.get_dynamic_item(self.hmap, self.dimensions, key)
    617             traverse_setter(self, '_force', False)
    618             if not isinstance(key, tuple): key = (key,)

[...]

PR #970, opened 2016-11-04
Ensure OverlayPlot ranges are computed correctly when +axiswise, +framewise

philippjfr:
Previously the ``compute_ranges`` method would skip computing the ranges for Elements in an OverlayPlot when both +axiswise and +framewise were active.


jlstevens:
The tests have passed and the change makes sense to me. 

I believe this partially addresses the normalization problem mentioned in #965. Merging.

PR #991, opened 2016-11-29
Optimized dask interface groupby implementation

philippjfr:
The ``groupby`` implementation now no longer loads all the columns that are being grouped over. Instead it now uses the categories or unique values in a column for the 1D case and uses the ``itertuple`` method otherwise to accumulate the unique indices without loading the whole column at once.


jbednar:
Sounds good to me!

jlstevens:
Looks good and as this is a new datatype, there are no backwards compatibility implications to worry about. Merging.

Issue #988, opened 2016-11-16
Implement nonzero/bool methods to override len


philippjfr:
Currently when testing any HoloViews object for "truthiness" it uses the ``__len__`` method. In #974 we added the dask interface, which doesn't actually know it's length because it loads the data lazily. Therefore we should implement ``__nonzero__`` (Python 2) and ``__bool__`` (Python 3) methods, which take precedence over the length. Additionally we should confirm that all the code in HoloViews uses the implicit boolean truthiness to check for a zero length element rather than explicitly calling ``len``.

philippjfr:
Implemented in PR referenced above.

PR #992, opened 2016-11-29
Dataset types use __nonzero/bool__ method for truthiness

philippjfr:
As discussed in #988 this PR implements the ``__nonzero__`` (Py2) and ``__bool__`` (Py3) methods on Dataset ensuring, which is used instead of ``__len__``. This ensures that the dask interface can implement the ``__len__`` method correctly, without forcing code throughout holoviews to compute the length on a large out-of-core dask dataframe all the time.

### DIFF
--- a/holoviews/core/data/dask.py
+++ b/holoviews/core/data/dask.py
@@ -229,12 +229,8 @@ class DaskInterface(PandasInterface):
         return columns.data.compute()
 
     @classmethod
-    def length(cls, dataset):
-        """
-        Length of dask dataframe is unknown, always return 1
-        for performance, use shape to compute dataframe shape.
-        """
-        return 1
+    def nonzero(cls, dataset):
+        return True
 
 
 
--- a/holoviews/element/chart.py
+++ b/holoviews/element/chart.py
@@ -309,12 +309,6 @@ class Points(Chart):
 
     _min_dims = 2                      # Minimum number of columns
 
-    def __iter__(self):
-        i = 0
-        while i < len(self):
-            yield tuple(self.data[i, ...])
-            i += 1
-
 
 
 class VectorField(Points):
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -426,6 +426,10 @@ class Dataset(Element):
         """
         return self.interface.length(self)
 
+    def __nonzero__(self):
+        return self.interface.nonzero(self)
+
+    __bool__ = __nonzero__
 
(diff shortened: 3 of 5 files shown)


holoviews/core/data/dask.py
```diff
@@ -229,12 +229,8 @@ def dframe(cls, columns, dimensions):
...
 
     @classmethod
-    def length(cls, dataset):
-        """
-        Length of dask dataframe is unknown, always return 1
-        for performance, use shape to compute dataframe shape.
-        """
-        return 1
+    def nonzero(cls, dataset):
+        return True
```
jbednar: This is a definite improvement!  Hardcoding nonzero is vastly better than hardcoding length.  Even so, is there no way to determine the actual value of nonzero in a way that doesn't load the entire dataset?

jbednar:
Looks great!  Happy to see it merged.

holoviews/element/chart.py
```diff
@@ -309,12 +309,6 @@ class Points(Chart):
 
     _min_dims = 2                      # Minimum number of columns
 
-    def __iter__(self):
-        i = 0
-        while i < len(self):
-            yield tuple(self.data[i, ...])
-            i += 1
-
```
philippjfr: This was old and broken because it assumed the data was always an array. There is an existing issue about adding an iterator method to all the ``Dataset`` interfaces somewhere.

holoviews/core/data/dask.py
philippjfr: I tried various things such as using:

```python
try:
   next(df.iterrows())
except:
   nonzero = False
else:
   nonzero = True
```

But all of these approaches still take a considerable amount of time.

jbednar: Maybe worth asking the dask maintainers if there is a quick method for testing nonzero, then.

philippjfr: I've asked, hopefully they'll have a good solution.

philippjfr: Matt Rocklin suggested using ``.head`` and checking the length of that, not any faster than my solution above though.

philippjfr: Discussed it properly now, and it's fairly clear that there won't be a cheap general solution here. A dask dataframe can be the result of a bunch of chained operations which all have to be evaluated before the length or even a nonzero length can be determined.

jlstevens:
Very happy with this improvement and tests have passed. Merging.

PR #994, opened 2016-11-30
Implement Image interface

philippjfr:
This PR is a WIP in implementing an interface to allow ``Image`` Elements to subclass ``Dataset`` and use the other interfaces, replacing ``GridImage``. There are a few things to be settled so I'm hoping to discuss these in this PR.

1. Our Image types need bounds. When you pass in an array the bounds are required, but when using the xarray and grid interfaces, the bounds should be computed from the data. Currently I deduce the bounds by asserting that all samples must be evenly spaced with a small tolerance (10e-9). Eventually we'll have QuadMesh for non-equal spacing, but that will require some more work. Does this sound reasonable?

2. Dataset does not have an interface to access multiple value arrays at once. This makes it difficult to access RGB and HSV data. We can either accept the overhead of stacking and unstacking these arrays and provide a convenient property on RGB/HSV elements to do this for us, or come up with a general API on the interfaces to get the stacked arrays. There was also the suggestion that other interfaces could support holding stacked value dimension arrays instead of holding them as unstacked arrays per value dimension but that could happen in a later PR.

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -444,27 +444,28 @@ class Dataset(Element):
         aggregated = self.interface.aggregate(self, kdims, function, **kwargs)
         aggregated = self.interface.unpack_scalar(self, aggregated)
 
+        ndims = len(dimensions)
+        min_d, max_d = self.params('kdims').bounds
+        new_type = type(self)
+        if (min_d is not None and ndims < min_d) or (max_d is not None and ndims > max_d):
+            new_type = Dataset
+
         vdims = self.vdims
         if spreadfn:
             error = self.interface.aggregate(self, dimensions, spreadfn)
             spread_name = spreadfn.__name__
             ndims = len(vdims)
-            error = self.clone(error, kdims=kdims)
-            combined = self.clone(aggregated, kdims=kdims)
+            error = self.clone(error, kdims=kdims, new_type=Dataset)
+            combined = self.clone(aggregated, kdims=kdims, new_type=Dataset)
             for i, d in enumerate(vdims):
                 dim = d('_'.join([d.name, spread_name]))
                 dvals = error.dimension_values(d, False, False)
                 combined = combined.add_dimension(dim, ndims+i, dvals, True)
-            return combined
+            return combined.clone(new_type=new_type)
 
-        ndims = len(dimensions)
-        min_d, max_d = self.params('kdims').bounds
         if np.isscalar(aggregated):
             return aggregated
         else:
-            new_type = None
-            if (min_d is not None and ndims < min_d) or (max_d is not None and ndims > max_d):
-                new_type = Dataset
             try:
                 return self.clone(aggregated, kdims=kdims, vdims=vdims,
                                   new_type=new_type)
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -290,12 +290,12 @@ class Dataset(Element):
         if selection_specs and not any(self.matches(sp) for sp in selection_specs):
             return self
 
-        data, kwargs = self.interface.select(self, **selection)
+        data = self.interface.select(self, **selection)
 
         if np.isscalar(data):
             return data
         else:
-            return self.clone(data, **kwargs)
+            return self.clone(data)
 
 
     def reindex(self, kdims=None, vdims=None):
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -167,7 +167,7 @@ class Dataset(Element):
         initialized = Interface.initialize(type(self), data, kdims, vdims,
                                            datatype=kwargs.get('datatype'))
         (data, self.interface, dims, extra_kws) = initialized
-        super(Dataset, self).__init__(data, **dict(extra_kws, **dict(kwargs, **dims)))
+        super(Dataset, self).__init__(data, **dict(kwargs, **dict(dims, **extra_kws)))
         self.interface.validate(self)
 
 
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -193,7 +193,7 @@ class Dataset(Element):
         super(Dataset, self).__setstate__(state)
 
 
-    def closest(self, coords):
+    def closest(self, coords=[], **kwargs):
         """
         Given single or multiple samples along the first key dimension
         will return the closest actual sample coordinates.
@@ -201,6 +201,14 @@ class Dataset(Element):
         if self.ndims > 1:
             raise NotImplementedError("Closest method currently only "
                                       "implemented for 1D Elements")
+
+        if kwargs:
+            dim = self.get_dimension(list(kwargs.keys())[0], strict=True)
+            if len(kwargs) > 1:
+                raise NotImplementedError("Closest method currently only "
+                                          "supports 1D indexes")
+            samples = list(kwargs.values())[0]
+            coords = samples if isinstance(samples, list) else [samples]
 
         xs = self.dimension_values(0)
         if xs.dtype.kind in 'SO':
@@ -364,15 +372,30 @@ class Dataset(Element):
         return data
 
 
-    def sample(self, samples=[], closest=True):
+    def sample(self, samples=[], closest=True, **kwargs):
         """
         Allows sampling of Dataset as an iterator of coordinates
         matching the key dimensions, returning a new object containing
         just the selected samples. Alternatively may supply kwargs
-        to sample a co-ordinate on an object.
+        to sample a coordinate on an object.
         """
-        samples = [util.wrap_tuple(s) for s in samples]
-        if len({len(s) for s in samples}) > 1:
+        if kwargs and samples:
+            raise Exception('Supply explicit list of samples or kwargs, not both.')
+        elif kwargs:
+            sample = [slice(None) for _ in range(self.ndims)]
+            for dim, val in kwargs.items():
+                sample[self.get_dimension_index(dim)] = val
+            samples = [tuple(sample)]
+
+        from ...element import Table
+        if len(samples) == 1:
+            sel = {kd.name: s for kd, s in zip(self.kdims, samples[0])}
+            selection = self.select(**sel)
+            selection = [samples[0]+(selection,)] if np.isscalar(selection) else selection.columns()
+            return self.clone(selection, new_type=Table)
+
+        lens = {len(util.wrap_tuple(s)) for s in samples}
+        if len(lens) > 1:
             raise IndexError('Sample coordinates must all be of the same length.')
 
         if closest:
@@ -380,7 +403,8 @@ class Dataset(Element):
                 samples = self.closest(samples)
             except NotImplementedError:
                 pass
-        return self.clone(self.interface.sample(self, samples), new_type=Dataset)
+        samples = [util.wrap_tuple(s) for s in samples]
+        return self.clone(self.interface.sample(self, samples), new_type=Table)
 
 
     def reduce(self, dimensions=[], function=None, spreadfn=None, **reduce_map):
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -390,13 +390,19 @@ class Dataset(Element):
         from ...element import Table, Curve
         if len(samples) == 1:
             sel = {kd.name: s for kd, s in zip(self.kdims, samples[0])}
+            dims = [kd for kd, v in sel.items() if not np.isscalar(v)]
             selection = self.select(**sel)
-            if self.interface.gridded and self.ndims == 2 and len(sel) == 1:
+            new_type = Table
+            kdims = self.kdims
+            if self.interface.gridded and self.ndims == 2 and len(dims) == 1:
                 new_type = Curve
+                selection = selection.columns()
+                kdims = [self.get_dimension(kd) for kd in dims]
+            elif np.isscalar(selection):
+                selection = [samples[0]+(selection,)]
             else:
-                new_type = Table
-            selection = [samples[0]+(selection,)] if np.isscalar(selection) else selection.columns()
-            return self.clone(selection, new_type=Table)
+                selection = selection.columns()
+            return self.clone(selection, kdims=kdims, new_type=new_type)
 
         lens = {len(util.wrap_tuple(s)) for s in samples}
         if len(lens) > 1:
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -150,6 +150,9 @@ class Dataset(Element):
     # Define a class used to transform Datasets into other Element types
     _conversion_interface = DataConversion
 
+    _vdim_reductions = {}
+    _kdim_reductions = {}
+
     def __init__(self, data, **kwargs):
         if isinstance(data, Element):
             pvals = util.get_param_values(data)
@@ -305,13 +308,16 @@ class Dataset(Element):
         else:
             key_dims = [self.get_dimension(k, strict=True) for k in kdims]
 
+        new_type = None
         if vdims is None:
             val_dims = [d for d in self.vdims if not kdims or d not in kdims]
         else:
             val_dims = [self.get_dimension(v, strict=True) for v in vdims]
+            new_type = self._vdim_reductions.get(len(val_dims), type(self))
 
         data = self.interface.reindex(self, key_dims, val_dims)
-        return self.clone(data, kdims=key_dims, vdims=val_dims)
+        return self.clone(data, kdims=key_dims, vdims=val_dims,
+                          new_type=new_type)
 
 
     def __getitem__(self, slices):
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -422,7 +422,9 @@ class Dataset(Element):
         if np.isscalar(aggregated):
             return aggregated
         else:
-            new_type = None if ndims >= min_d and ndims <= max_d else Dataset
+            new_type = None
+            if (min_d is not None and ndims < min_d) or (max_d is not None and ndims > max_d):
+                new_type = Dataset
             try:
                 return self.clone(aggregated, kdims=kdims, vdims=vdims,
                                   new_type=new_type)
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -196,12 +196,14 @@ class Dataset(Element):
         will return the closest actual sample coordinates.
         """
         if self.ndims > 1:
-            NotImplementedError("Closest method currently only "
-                                "implemented for 1D Elements")
+            raise NotImplementedError("Closest method currently only "
+                                      "implemented for 1D Elements")
 
         xs = self.dimension_values(0)
+        if xs.dtype.kind in 'SO':
+            raise NotImplementedError("Closest only supported for numeric types")
         idxs = [np.argmin(np.abs(xs-coord)) for coord in coords]
-        return [xs[idx] for idx in idxs] if len(coords) > 1 else xs[idxs[0]]
+        return [xs[idx] for idx in idxs]
 
 
     def sort(self, by=[]):
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -55,6 +55,7 @@ except ImportError:
 
 from ..dimension import Dimension
 from ..element import Element
+from ..ndmapping import OrderedDict
 from ..spaces import HoloMap, DynamicMap
 from .. import util
 
@@ -387,6 +388,9 @@ class Dataset(Element):
                 sample[self.get_dimension_index(dim)] = val
             samples = [tuple(sample)]
 
+        # Note: Special handling sampling of gridded 2D data as Curve
+        # may be replaced wih more general handling
+        # see https://github.com/ioam/holoviews/issues/1173
         from ...element import Table, Curve
         if len(samples) == 1:
             sel = {kd.name: s for kd, s in zip(self.kdims, samples[0])}
@@ -396,12 +400,12 @@ class Dataset(Element):
             kdims = self.kdims
             if self.interface.gridded and self.ndims == 2 and len(dims) == 1:
                 new_type = Curve
-                selection = selection.columns()
                 kdims = [self.get_dimension(kd) for kd in dims]
+                selection = tuple(selection.columns(kdims+self.vdims).values())
             elif np.isscalar(selection):
                 selection = [samples[0]+(selection,)]
             else:
-                selection = selection.columns()
+                selection = tuple(selection.columns(kdims+self.vdims).values())
             return self.clone(selection, kdims=kdims, new_type=new_type)
 
         lens = {len(util.wrap_tuple(s)) for s in samples}
@@ -597,7 +601,7 @@ class Dataset(Element):
             dimensions = self.dimensions()
         else:
             dimensions = [self.get_dimension(d, strict=True) for d in dimensions]
-        return {d.name: self.dimension_values(d) for d in dimensions}
+        return OrderedDict([(d.name, self.dimension_values(d)) for d in dimensions])
 
 
     @property
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -387,10 +387,14 @@ class Dataset(Element):
                 sample[self.get_dimension_index(dim)] = val
             samples = [tuple(sample)]
 
-        from ...element import Table
+        from ...element import Table, Curve
         if len(samples) == 1:
             sel = {kd.name: s for kd, s in zip(self.kdims, samples[0])}
             selection = self.select(**sel)
+            if self.interface.gridded and self.ndims == 2 and len(sel) == 1:
+                new_type = Curve
+            else:
+                new_type = Table
             selection = [samples[0]+(selection,)] if np.isscalar(selection) else selection.columns()
             return self.clone(selection, new_type=Table)
 
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -344,7 +344,7 @@ class Dataset(Element):
             value_select = slices[self.ndims]
         elif len(slices) == self.ndims+1 and isinstance(slices[self.ndims],
                                                         (Dimension,str)):
-            raise Exception("%r is not an available value dimension'" % slices[self.ndims])
+            raise Exception("%r is not an available value dimension" % slices[self.ndims])
         else:
             selection = dict(zip(self.dimensions(label=True), slices))
         data = self.select(**selection)
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -189,6 +189,7 @@ class Dataset(Element):
 
         super(Dataset, self).__setstate__(state)
 
+
     def closest(self, coords):
         """
         Given single or multiple samples along the first key dimension
@@ -198,7 +199,6 @@ class Dataset(Element):
             NotImplementedError("Closest method currently only "
                                 "implemented for 1D Elements")
 
-        if not isinstance(coords, list): coords = [coords]
         xs = self.dimension_values(0)
         idxs = [np.argmin(np.abs(xs-coord)) for coord in coords]
         return [xs[idx] for idx in idxs] if len(coords) > 1 else xs[idxs[0]]
@@ -356,13 +356,23 @@ class Dataset(Element):
         return data
 
 
-    def sample(self, samples=[]):
+    def sample(self, samples=[], closest=True):
         """
         Allows sampling of Dataset as an iterator of coordinates
         matching the key dimensions, returning a new object containing
-        just the selected samples.
+        just the selected samples. Alternatively may supply kwargs
+        to sample a co-ordinate on an object.
         """
-        return self.clone(self.interface.sample(self, samples))
+        samples = [util.wrap_tuple(s) for s in samples]
+        if len({len(s) for s in samples}) > 1:
+            raise IndexError('Sample coordinates must all be of the same length.')
+
+        if closest:
+            try:
+                samples = self.closest(samples)
+            except NotImplementedError:
+                pass
+        return self.clone(self.interface.sample(self, samples), new_type=Dataset)
 
 
     def reduce(self, dimensions=[], function=None, spreadfn=None, **reduce_map):
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -99,6 +99,16 @@ class DataConversion(object):
         if vdims is None:
             vdims = self._element.vdims
         if vdims and not isinstance(vdims, list): vdims = [vdims]
+
+        # Checks Element type supports dimensionality
+        type_name = new_type.__name__
+        for dim_type, dims in (('kdims', kdims), ('vdims', vdims)):
+            min_d, max_d = new_type.params(dim_type).bounds
+            if ((min_d is not None and len(dims) < min_d) or
+                (max_d is not None and len(dims) > max_d)):
+                raise ValueError("%s %s must be between length %s and %s." %
+                                 (type_name, dim_type, min_d, max_d))
+
         if groupby is None:
             groupby = [d for d in self._element.kdims if d not in kdims+vdims]
         elif groupby and not isinstance(groupby, list):
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -200,8 +200,12 @@ class Dataset(Element):
 
     def closest(self, coords=[], **kwargs):
         """
-        Given single or multiple samples along the first key dimension
-        will return the closest actual sample coordinates.
+        Given a single coordinate or multiple coordinates as
+        a tuple or list of tuples or keyword arguments matching
+        the dimension closest will find the closest actual x/y
+        coordinates. Different Element types should implement this
+        appropriately depending on the space they represent, if the
+        Element does not support snapping raise NotImplementedError.
         """
         if self.ndims > 1:
             raise NotImplementedError("Closest method currently only "
@@ -460,9 +464,7 @@ class Dataset(Element):
 
         ndims = len(dimensions)
         min_d, max_d = self.params('kdims').bounds
-        new_type = type(self)
-        if (min_d is not None and ndims < min_d) or (max_d is not None and ndims > max_d):
-            new_type = Dataset
+        generic_type = (min_d is not None and ndims < min_d) or (max_d is not None and ndims > max_d)
 
         vdims = self.vdims
         if spreadfn:
@@ -475,7 +477,7 @@ class Dataset(Element):
                 dim = d('_'.join([d.name, spread_name]))
                 dvals = error.dimension_values(d, False, False)
                 combined = combined.add_dimension(dim, ndims+i, dvals, True)
-            return combined.clone(new_type=new_type)
+            return combined.clone(new_type=Dataset if generic_type else type(self))
 
         if np.isscalar(aggregated):
             return aggregated
@@ -486,7 +488,8 @@ class Dataset(Element):
             except:
                 datatype = self.params('datatype').default
                 return self.clone(aggregated, kdims=kdims, vdims=vdims,
-                                  new_type=new_type, datatype=datatype)
+                                  new_type=Dataset if generic_type else None,
+                                  datatype=datatype)
 
 
     def groupby(self, dimensions=[], container_type=HoloMap, group_type=None,
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -373,12 +373,14 @@ class Dataset(Element):
         return data
 
 
-    def sample(self, samples=[], **kwargs):
+    def sample(self, samples=[], closest=True, **kwargs):
         """
         Allows sampling of Dataset as an iterator of coordinates
         matching the key dimensions, returning a new object containing
         just the selected samples. Alternatively may supply kwargs
-        to sample a coordinate on an object.
+        to sample a coordinate on an object. By default it will attempt
+        to snap to the nearest coordinate if the Element supports it,
+        snapping may be disabled with the closest argument.
         """
         if kwargs and samples:
             raise Exception('Supply explicit list of samples or kwargs, not both.')
@@ -416,10 +418,11 @@ class Dataset(Element):
         if len(lens) > 1:
             raise IndexError('Sample coordinates must all be of the same length.')
 
-        try:
-            samples = self.closest(samples)
-        except NotImplementedError:
-            pass
+        if closest:
+            try:
+                samples = self.closest(samples)
+            except NotImplementedError:
+                pass
         samples = [util.wrap_tuple(s) for s in samples]
         return self.clone(self.interface.sample(self, samples), new_type=Table)
 
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -373,7 +373,7 @@ class Dataset(Element):
         return data
 
 
-    def sample(self, samples=[], closest=True, **kwargs):
+    def sample(self, samples=[], **kwargs):
         """
         Allows sampling of Dataset as an iterator of coordinates
         matching the key dimensions, returning a new object containing
@@ -396,27 +396,30 @@ class Dataset(Element):
             sel = {kd.name: s for kd, s in zip(self.kdims, samples[0])}
             dims = [kd for kd, v in sel.items() if not np.isscalar(v)]
             selection = self.select(**sel)
-            new_type = Table
-            kdims = self.kdims
+
+            # If a 1D cross-section of 2D space return Curve
             if self.interface.gridded and self.ndims == 2 and len(dims) == 1:
                 new_type = Curve
                 kdims = [self.get_dimension(kd) for kd in dims]
-                selection = tuple(selection.columns(kdims+self.vdims).values())
-            elif np.isscalar(selection):
+            else:
+                new_type = Table
+                kdims = self.kdims
+
+            if np.isscalar(selection):
                 selection = [samples[0]+(selection,)]
             else:
                 selection = tuple(selection.columns(kdims+self.vdims).values())
+
             return self.clone(selection, kdims=kdims, new_type=new_type)
 
-        lens = {len(util.wrap_tuple(s)) for s in samples}
+        lens = set(len(util.wrap_tuple(s)) for s in samples)
         if len(lens) > 1:
             raise IndexError('Sample coordinates must all be of the same length.')
 
-        if closest:
-            try:
-                samples = self.closest(samples)
-            except NotImplementedError:
-                pass
+        try:
+            samples = self.closest(samples)
+        except NotImplementedError:
+            pass
         samples = [util.wrap_tuple(s) for s in samples]
         return self.clone(self.interface.sample(self, samples), new_type=Table)
 
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -105,7 +105,11 @@ class DataConversion(object):
             groupby = [groupby]
 
         if self._element.interface.gridded:
-            selected = self._element
+            dropped_kdims = [kd for kd in self._element.kdims if kd not in groupby+kdims]
+            if dropped_kdims:
+                selected = self._element.reindex(groupby+kdims, vdims)
+            else:
+                selected = self._element
         else:
             selected = self._element.reindex(groupby+kdims, vdims)
         params = {'kdims': [selected.get_dimension(kd, strict=True) for kd in kdims],
@@ -114,7 +118,7 @@ class DataConversion(object):
         if selected.group != selected.params()['group'].default:
             params['group'] = selected.group
         params.update(kwargs)
-        if len(kdims) == selected.ndims:
+        if len(kdims) == selected.ndims or not groupby:
             element = new_type(selected, **params)
             return element.sort() if sort else element
         group = selected.groupby(groupby, container_type=HoloMap,
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -199,9 +199,9 @@ class GridInterface(DictInterface):
             select = dict(zip(dim_names, unique_key))
             if drop_dim:
                 group_data = dataset.select(**select)
-                group_data = group_data if np.isscalar(group_data) else group_data.columns()
+                group_data, _ = group_data if np.isscalar(group_data) else (group_data.columns(), None)
             else:
-                group_data = cls.select(dataset, **select)
+                group_data, _ = cls.select(dataset, **select)
             if np.isscalar(group_data):
                 group_data = {dataset.vdims[0].name: np.atleast_1d(group_data)}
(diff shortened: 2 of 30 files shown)


jlstevens:
Just to summarize a discussion I had with Philipp on gitter regarding the first question:

> Our Image types need bounds. When you pass in an array the bounds are required, but when using the xarray and grid interfaces, the bounds should be computed from the data. Currently I deduce the bounds by asserting that all samples must be evenly spaced with a small tolerance (10e-9). Eventually we'll have QuadMesh for non-equal spacing, but that will require some more work. Does this sound reasonable?

We agreed that this is a reasonable approach as long as the tolerance is about avoiding machine precision issues only and not some fuzzy ad hoc tolerance to be tweaked by the user. For instance, instead of 10e-9 a suitable value should be determined for the system, e.g using the [sys.float_info](https://docs.python.org/2.7/library/sys.html#sys.float_info).

jlstevens:
Again summarizing our discussion on gitter:

> Dataset does not have an interface to access multiple value arrays at once. This makes it difficult to access RGB and HSV data. We can either accept the overhead of stacking and unstacking these arrays and provide a convenient property on RGB/HSV elements to do this for us, or come up with a general API on the interfaces to get the stacked arrays. There was also the suggestion that other interfaces could support holding stacked value dimension arrays instead of holding them as unstacked arrays per value dimension but that could happen in a later PR.

I agree that a future PR supporting multiple vdims in the data API would be very welcome but for now, we can extend ``dimension_values`` on ``RGB`` elements only. Specifically, we can adapt ``RGB`` to make ``RGB.dimension_values(['R', 'G', 'B'])`` work in this PR and in a later PR we can support stacked output from ``dimension_values`` (for lists of vdims only).

philippjfr:
Now based off #1075 and now slowly reaching API compatibility. Still missing a sample implementation otherwise largely complete. Should also add a lot more unit tests for grid interfaces, which will also cover the new ``ImageInterface``.

Currently I'm also still a bit unclear what to do about Raster itself. It could probably be unified with Image by simply adding integer indices in the constructor.

philippjfr:
Raster is now unified with the image interface. Now I'll have to refactor QuadMesh so it still works independently. Hopefully we'll manage to write interfaces to handle 1D and 2D binned data, e.g. for Histogram and QuadMesh, as part of 1.8.

tests/testellipsis.py
```diff
@@ -87,14 +87,15 @@ class TestEllipsisRaster(ComparisonTestCase):
...
-        self.assertEqual(sliced.data, data)
+        self.assertEqual(sliced.data, np.flipud(data))
 
     def test_raster_ellipsis_slice_value_missing(self):
         data = np.random.rand(10,10)
         try:
             hv.Raster(data)[...,'Non-existent']
         except Exception as e:
-            if str(e) != repr("'z' is the only selectable value dimension"):
+            print(e)
```
philippjfr: Don't forget to remove.

jlstevens:
I know this PR is still WIP but I would like to ask one question...

Glancing through the code, how much of it would be needed if we weren't concerned about backwards compatibility? Could much of this code be eliminated if we decided to break compatibility and clean up what was supported for 2.0?

philippjfr:
>Glancing through the code, how much of it would be needed if we weren't concerned about backwards compatibility? Could much of this code be eliminated if we decided to break compatibility and clean up what was supported for 2.0?

Yes, a lot of it could be eliminated. That said I've started to reconsider what we discussed previously. I now do think ``Image`` should validate that it is regularly sampled, which means we might as well ship the new ImageInterface at least until 2.0, but potentially even beyond then.

holoviews/element/raster.py
```diff
@@ -644,6 +421,8 @@ class RGB(Image):
         If an alpha channel is supplied, the defined alpha_dimension
         is automatically appended to this list.""")
 
+    _vdim_reductions = {1: Image}
```
philippjfr: Remove, general approach for this can be added later.

holoviews/element/raster.py
```diff
@@ -259,6 +260,18 @@ def __init__(self, data, bounds=None, extents=None, xdensity=None, ydensity=None
                 raise ValueError("Input array has shape %r but %d value dimensions defined"
                                  % (self.shape, len(self.vdims)))
 
+
+    def __setstate__(self, state):
+        """
+        Ensures old-style unpickled Image types without an interface
+        use the ImageInterface.
+        """
```
jlstevens: Just one quick comment: might be worth adding the word 'deprecation' so when we grep for it we find it (e.g if we no longer want to support this in 2.0)

philippjfr: Right, in fact we really need to remove the ``__setstate__`` on Dataset as that's for the old terrible NdElement format, which was deprecated in 1.4(?) iirc and might be the only reason this is required at all.

holoviews/core/util.py
```diff
@@ -1184,3 +1184,24 @@ def is_nan(x):
         return np.isnan(x)
     except:
         return False
+
+
+def bound_range(vals, density, TOL=10e-9):
```
philippjfr: To disable the enforcement of a grid we'll just have to replace this function.

philippjfr:
@jlstevens This is now ready for initial review. Raster is still not a Dataset type, as we had discussed I will try to change that in a followup PR. Another annoyance is that type conversions on sampling and aggregating is handled in various ad-hoc ways and should eventually be clearly defined via some general mechanism as suggested in https://github.com/ioam/holoviews/issues/1173. Finally I'll replace the ``bound_range`` function with something that doesn't strictly enforce an evenly sampled grid.

I know this is basically impossible to review but just do your best.

philippjfr:
Before merging I should also do a thorough review of docstrings.

jbednar:
+700 lines; I guess this one cancels out the code lost in the Dynamic Maps PR.  Still, very important functionality...

philippjfr:
>+700 lines; I guess this one cancels out the code lost in the Dynamic Maps PR. Still, very important functionality...

To be fair that's 350 lines of tests, while the deletions in the other PR are dominated by docstrings and tutorials. Also there is a fair amount of duplication going on here and once we can unify Raster and eventually QuadMesh with the interfaces, a few hundred lines will fall away again.

jlstevens:
@jbednar I don't see it as more lines of code but as a class removed (``GridImage``) as well as a way of making data handling more consistent!

jbednar:
Precisely!

holoviews/element/raster.py
philippjfr: Actually it's a bit of pain to replace this, I can file a comment on the issue about this to either generalize this approach in some way or replace it.

jlstevens: Yes, please file an issue. Definitely want to address this but don't want to make this PR any longer either!

philippjfr: There already is an issue about it, so I'll add a comment there, discussing the current state.

jlstevens: Ah yes, I remember the issue. I don't know if you get a reference between issues/PRs from the comments but we can find out: I believe the relevant issue is #1173

jlstevens: Didn't work but the issue was already referenced below.

holoviews/core/data/__init__.py
```diff
@@ -166,7 +171,7 @@ def __init__(self, data, **kwargs):
         initialized = Interface.initialize(type(self), data, kdims, vdims,
                                            datatype=kwargs.get('datatype'))
         (data, self.interface, dims, extra_kws) = initialized
-        super(Dataset, self).__init__(data, **dict(extra_kws, **dict(kwargs, **dims)))
+        super(Dataset, self).__init__(data, **dict(kwargs, **dict(dims, **extra_kws)))
```
jlstevens: I assume this is an unrelated bug-fix regarding argument precedence.

holoviews/core/data/__init__.py
```diff
@@ -188,19 +193,29 @@ def __setstate__(self, state):
 
         super(Dataset, self).__setstate__(state)
 
-    def closest(self, coords):
+
+    def closest(self, coords=[], **kwargs):
         """
         Given single or multiple samples along the first key dimension
         will return the closest actual sample coordinates.
         """
```
jlstevens: docstring should now describe how to use coords.

holoviews/core/data/__init__.py
```diff
@@ -188,19 +193,29 @@ def __setstate__(self, state):
...
         if self.ndims > 1:
-            NotImplementedError("Closest method currently only "
-                                "implemented for 1D Elements")
+            raise NotImplementedError("Closest method currently only "
```
jlstevens: Oops! :-)

holoviews/core/data/__init__.py
philippjfr: I'll have to double check this, might be a leftover from something I tried early on.

holoviews/core/data/__init__.py
```diff
@@ -188,19 +193,29 @@ def __setstate__(self, state):
...
+                                      "implemented for 1D Elements")
+
+        if kwargs:
+            dim = self.get_dimension(list(kwargs.keys())[0], strict=True)
+            if len(kwargs) > 1:
+                raise NotImplementedError("Closest method currently only "
+                                          "supports 1D indexes")
+            samples = list(kwargs.values())[0]
+            coords = samples if isinstance(samples, list) else [samples]
```
jlstevens: In this case is coords a tuple? How often is it used without being a list?

Might be worth always requiring a list of samples for consistency - it all depends on how often a single sample is needed.

philippjfr: >Might be worth always requiring a list of samples for consistency - it all depends on how often a single sample is needed.

I'll make a comment, this is for ``img.sample(x=0)``, which wasn't supported on Datasets before.

holoviews/core/data/__init__.py
```diff
@@ -188,19 +193,29 @@ def __setstate__(self, state):
...
 
-        if not isinstance(coords, list): coords = [coords]
         xs = self.dimension_values(0)
+        if xs.dtype.kind in 'SO':
+            raise NotImplementedError("Closest only supported for numeric types")
         idxs = [np.argmin(np.abs(xs-coord)) for coord in coords]
-        return [xs[idx] for idx in idxs] if len(coords) > 1 else xs[idxs[0]]
+        return [xs[idx] for idx in idxs]
```
jlstevens: Looks like the return format has changed given only a single sample. The new version does seem more consistent.

holoviews/core/data/__init__.py
philippjfr: Wait no, I was confused, I think it is possible to get rid of this. We'll see.

holoviews/core/data/__init__.py
```diff
@@ -330,7 +349,7 @@ def __getitem__(self, slices):
         if isinstance(slices, np.ndarray) and slices.dtype.kind == 'b':
             if not len(slices) == len(self):
                 raise IndexError("Boolean index must match length of sliced object")
-            return self.clone(self.interface.select(self, selection_mask=slices))
+            return self.clone(self.select(selection_mask=slices))
```
jlstevens: I *think* the old version returned data whereas ``self.select`` should return a ``Dataset``. The former seems more intuitive to me but if they both work it doesn't matter much. Is the new version better somehow?

philippjfr: I'll double check.

holoviews/core/data/__init__.py
```diff
@@ -354,13 +373,52 @@ def __getitem__(self, slices):
...
+    def sample(self, samples=[], closest=True, **kwargs):
         """
         Allows sampling of Dataset as an iterator of coordinates
         matching the key dimensions, returning a new object containing
-        just the selected samples.
-        """
-        return self.clone(self.interface.sample(self, samples))
+        just the selected samples. Alternatively may supply kwargs
+        to sample a coordinate on an object.
+        """
```
jlstevens: The closest switch needs to be explain here as well.

holoviews/core/data/__init__.py
```diff
@@ -354,13 +373,52 @@ def __getitem__(self, slices):
...
+        if len(samples) == 1:
+            sel = {kd.name: s for kd, s in zip(self.kdims, samples[0])}
+            dims = [kd for kd, v in sel.items() if not np.isscalar(v)]
+            selection = self.select(**sel)
+            new_type = Table
+            kdims = self.kdims
+            if self.interface.gridded and self.ndims == 2 and len(dims) == 1:
+                new_type = Curve
+                kdims = [self.get_dimension(kd) for kd in dims]
+                selection = tuple(selection.columns(kdims+self.vdims).values())
```
jlstevens: Might be clearer to return here. ie.

```python
return self.clone(selection, kdims=kdims, new_type=Curve)
```

The the other ``self.clone`` can inline ``new_type=Table``. Then there is no reason to declare ``new_type`` as a variable.

philippjfr: Right I think I can clean this up a fair bit.

holoviews/core/data/__init__.py
```diff
@@ -390,24 +448,35 @@ def aggregate(self, dimensions=None, function=None, spreadfn=None, **kwargs):
         aggregated = self.interface.aggregate(self, kdims, function, **kwargs)
         aggregated = self.interface.unpack_scalar(self, aggregated)
 
+        ndims = len(dimensions)
+        min_d, max_d = self.params('kdims').bounds
+        new_type = type(self)
+        if (min_d is not None and ndims < min_d) or (max_d is not None and ndims > max_d):
+            new_type = Dataset
```
philippjfr: Might make sense to make this a ``Table`` instead.

holoviews/core/data/__init__.py
```diff
@@ -354,13 +373,52 @@ def __getitem__(self, slices):
...
+            kdims = self.kdims
+            if self.interface.gridded and self.ndims == 2 and len(dims) == 1:
+                new_type = Curve
+                kdims = [self.get_dimension(kd) for kd in dims]
+                selection = tuple(selection.columns(kdims+self.vdims).values())
+            elif np.isscalar(selection):
+                selection = [samples[0]+(selection,)]
+            else:
+                selection = tuple(selection.columns(kdims+self.vdims).values())
+            return self.clone(selection, kdims=kdims, new_type=new_type)
```
jlstevens: Given the above comment (can return Curve early) then you could do...

```python
            selection = (tuple(selection.columns(kdims+self.vdims).values())
                         if np.isscalar(selection) else  [samples[0]+(selection,)])
```

That might be a little long for your taste so I'll leave it up to you.

holoviews/core/data/__init__.py
```diff
@@ -354,13 +373,52 @@ def __getitem__(self, slices):
...
+
+        lens = {len(util.wrap_tuple(s)) for s in samples}
```
jlstevens: Took me a second to realize this is a set. Not need to change it unless you feel ``set(...)`` is more readable.

holoviews/core/data/__init__.py
```diff
@@ -390,24 +448,35 @@ def aggregate(self, dimensions=None, function=None, spreadfn=None, **kwargs):
         aggregated = self.interface.aggregate(self, kdims, function, **kwargs)
         aggregated = self.interface.unpack_scalar(self, aggregated)
 
+        ndims = len(dimensions)
+        min_d, max_d = self.params('kdims').bounds
+        new_type = type(self)
+        if (min_d is not None and ndims < min_d) or (max_d is not None and ndims > max_d):
```
jlstevens: Would be good to state what you want from the kdim min/max size in a comment...

holoviews/core/data/__init__.py
```diff
@@ -390,24 +448,35 @@ def aggregate(self, dimensions=None, function=None, spreadfn=None, **kwargs):
...
-            error = self.clone(error, kdims=kdims)
-            combined = self.clone(aggregated, kdims=kdims)
+            error = self.clone(error, kdims=kdims, new_type=Dataset)
+            combined = self.clone(aggregated, kdims=kdims, new_type=Dataset)
             for i, d in enumerate(vdims):
                 dim = d('_'.join([d.name, spread_name]))
                 dvals = error.dimension_values(d, False, False)
                 combined = combined.add_dimension(dim, ndims+i, dvals, True)
-            return combined
+            return combined.clone(new_type=new_type)
```
jlstevens: Might be nice to move the definition of ``new_type`` closer to where it is used.

As far I can tell ``new_type = type(self)`` or ``new_type = Dataset`` and the condition shouldn't depend on any of this other code (it just looks at the kdim parameter bounds).

jlstevens: Ah, it also seems to be used further down as well. But it still seems to be between ``type(self)`` and ``Dataset`` so maybe it could be a predicate?

jlstevens: For instance something like..

```python
return self.clone(aggregated, kdims=kdims, vdims=vdims,
                                  new_type=Dataset if generic_type else type(self))
```

(I realize 'generic_type' might not be the best name, maybe 'fallback_type'?)

philippjfr: ``generic_type`` seems fine and defining the generic type (currently Dataset) somewhere should be part of the dimension reductions implementation mentioned in the other issue.

holoviews/core/data/__init__.py
philippjfr: Will do.

holoviews/core/data/__init__.py
```diff
@@ -301,13 +317,16 @@ def reindex(self, kdims=None, vdims=None):
...
         if vdims is None:
             val_dims = [d for d in self.vdims if not kdims or d not in kdims]
         else:
             val_dims = [self.get_dimension(v, strict=True) for v in vdims]
+            new_type = self._vdim_reductions.get(len(val_dims), type(self))
 
         data = self.interface.reindex(self, key_dims, val_dims)
-        return self.clone(data, kdims=key_dims, vdims=val_dims)
+        return self.clone(data, kdims=key_dims, vdims=val_dims,
+                          new_type=new_type)
```
jlstevens:
```python
return self.clone(data, kdims=key_dims, vdims=val_dims,
           new_type=None if vdims is None else self._vdim_reductions.get(len(val_dims), type(self))
```

Or is that too long?

jlstevens: Or better:

```python
return self.clone(data, kdims=key_dims, vdims=val_dims,
           new_type=self._vdim_reductions.get(len(val_dims), type(self) if vdims else None)
```

holoviews/core/data/image.py
```diff
@@ -0,0 +1,250 @@
...
+        kdims = [kd.name if isinstance(kd, Dimension) else kd for kd in kdims]
+        axes = tuple(dataset.ndims-dataset.get_dimension_index(kdim)-1
+                     for kdim in dataset.kdims if kdim not in kdims)
+
+        data = np.atleast_1d(function(dataset.data, axis=axes, **kwargs))
+        if not kdims:
+            if len(dataset.vdims) == 1:
+                return data if np.isscalar(data) else data[0]
+            else:
+                return {vd.name: v for vd, v in zip(dataset.vdims, data)}
```
philippjfr: Double check it shouldn't be: ``vd.name: np.array([v])``

holoviews/core/data/grid.py
```diff
@@ -231,6 +231,8 @@ def key_select_mask(cls, dataset, values, ind):
                 mask &= ind.start <= values
             if ind.stop is not None:
                 mask &= values < ind.stop
+            if mask is True:
+                mask = np.ones(values.shape, dtype=np.bool)
```
jlstevens: Looks like ``mask`` is overloaded here - it seems to be both the boolean predicate figuring out if you want to create a mask array and the mask itself. I think it could be less confusing...

philippjfr: It's an optimization, I'll leave a comment.

holoviews/core/data/interface.py
```diff
@@ -50,22 +50,31 @@ def initialize(cls, eltype, data, kdims, vdims, datatype=None):
...
 
+        if datatype is None:
+            datatype = eltype.datatype
+
         # Process Element data
         if isinstance(data, NdElement):
             kdims = [kdim for kdim in kdims if kdim != 'Index']
-        elif hasattr(data, 'interface') and issubclass(data.interface, Interface):
-            data = data.data
+        elif (hasattr(data, 'interface') and issubclass(data.interface, Interface)):
```
jlstevens: Seems a bit odd - I guess this is because not all HoloViews elements use interfaces (e.g annotations?)

philippjfr: Not just annotations, you can cast a Raster or QuadMesh to a dataset.

holoviews/core/data/interface.py
```diff
@@ -50,22 +50,31 @@ def initialize(cls, eltype, data, kdims, vdims, datatype=None):
...
+            if data.interface.datatype in datatype:
+                data = data.data
+            elif data.interface.gridded:
+                gridded = {kd.name: data.dimension_values(kd.name, expanded=False)
+                           for kd in data.kdims}
+                for vd in data.vdims:
+                    gridded[vd.name] = data.dimension_values(vd, flat=False)
+                data = gridded
```
jlstevens: It looks like ``gridded`` the dictionary format for columns i.e a dictionary indexed by dimensions with array values. I wish there was a better name for this format as the variable ``gridded`` threw me off what it was for a while...

holoviews/core/data/interface.py
jlstevens: Having Raster use interfaces is something that would be nice if possible but not in the scope of this PR. Does QuadMesh not have an interface attribute?

holoviews/core/data/iris.py
```diff
@@ -280,12 +285,18 @@ def select_to_constraint(cls, dataset, selection):
...
+            return slicer
         constraint_kwargs = {}
         for dim, constraint in selection.items():
             if isinstance(constraint, slice):
                 constraint = (constraint.start, constraint.stop)
             if isinstance(constraint, tuple):
-                constraint = iris.util.between(*constraint, rh_inclusive=False)
+                if constraint == (None, None):
+                    continue
+                constraint = get_slicer(*constraint)
```
jlstevens: Was the old way of applying constraints wrong? At any rate, the new version is more readable for me, despite the nested functions.

holoviews/core/spaces.py
```diff
@@ -327,14 +327,15 @@ def sample(self, samples=[], bounds=None, **sample_values):
                 ysamples = [(ly+uy)/2.0 for ly,uy in zip(yedges[:-1], yedges[1:])]
 
                 Y,X = np.meshgrid(ysamples, xsamples)
-                linsamples = zip(X.flat, Y.flat)
+                linsamples = list(zip(X.flat, Y.flat))
```
jlstevens: Guess this isn't tested? Wouldn't there have been Python 3 issues if it had been?

holoviews/core/data/interface.py
philippjfr: No, that's waiting on us supporting bin edges in the interfaces somehow, at which point both Histogram and QuadMesh will be supported. Iris has a concept for them and I don't think it would be to hard to add to the dict interface, I don't know about xarray.

holoviews/core/spaces.py
philippjfr: Possibly it's just being used in a different context now and new tests flagged it up.

holoviews/core/data/iris.py
philippjfr: I don't know what happened here, I think it's an iris bug, but I'm not bothered to investigate tbh. The iris interface can move to geoviews in 2.0.

holoviews/core/util.py
jlstevens: Will be glad to get rid of it. Code looks hairy and it uses the ``np.diff`` approach which isn't particularly efficient. Then there are the tolerance issues...

holoviews/core/data/interface.py
jlstevens: Thanks for clarifying. I assume supporting bin edges won't happen for 1.7 then...

holoviews/element/util.py
```diff
@@ -25,17 +25,6 @@
...
-    """
-    Interface helper function to turn dask Arrays into numpy arrays as
-    necessary. If index_value is True, a value is returned instead of
-    an array holding a single value.
-    """
-    if dask and isinstance(v, dask.array.Array):
-        arr =  v.compute()
-        return arr[()] if index_value else arr
-    else:
-        return v
```
jlstevens: Guess it isn't used. The dask thing was just a prototype so removing it is probably the right thing to do.

philippjfr: We now support dask arrays by proxy through xarray, so there's not really any need.

holoviews/operation/element.py
```diff
@@ -430,11 +431,11 @@ def _process(self, element, key=None):
...
         if type(element) is Raster:
             data = [np.flipud(element.data)]
-        elif isinstance(element, Raster):
-            data = [element.data]
-        
+        elif isinstance(element, Image):
+            data = [np.flipud(element.dimension_values(2, flat=False))]
         elif isinstance(element, QuadMesh):
             data = (element.dimension_values(0, False),
                     element.dimension_values(1, False),
```
jlstevens: Have these changes to operation been tested? We will want operation tests when we get round to cleaning them up.

holoviews/plotting/bokeh/__init__.py
```diff
@@ -54,7 +54,7 @@
 
                 # Rasters
                 Image: RasterPlot,
-                GridImage: ImagePlot,
+                GridImage: RasterPlot,
```
jlstevens: I thought ``GridImage`` was going away...can we not get rid of it here? Or is this for backwards compatibility only?

philippjfr: No, it can be removed, it's just an alias now. Might even make it a function and add a deprecation warning.

jlstevens: > Might even make it a function and add a deprecation warning.

Good idea!

tests/testndmapping.py
```diff
@@ -118,7 +118,8 @@ def test_idxmapping_redim(self):
         data = [((0, 0.5), 'a'), ((1, 0.5), 'b')]
         ndmap = MultiDimensionalMapping(data, kdims=[self.dim1, self.dim2])
         redimmed = ndmap.redim(intdim='Integer')
-        self.assertEqual(redimmed.kdims, [Dimension('Integer'), Dimension('floatdim')])
```
jlstevens: Was this just wrong before? The names indicated types but type wasn't specified. I guess the tests passed as comparison worked with ``type=None``?

holoviews/plotting/mpl/util.py
```diff
@@ -164,3 +165,20 @@ def get_tight_bbox(fig, bbox_extra_artists=[], pad=None):
         bbox_extra = TransformedBbox(_bbox, trans)
         bbox_inches = Bbox.union([bbox_inches, bbox_extra])
     return bbox_inches.padded(pad) if pad else bbox_inches
+
+
+def get_raster_data(image):
+    """
+    Return the array data from any Raster or Image type
+    """
+    if isinstance(image, RGB):
```
jlstevens: Guess this actually works for HSV too...

tests/testndmapping.py
philippjfr: I've implemented proper comparisons for lists so now it actually compares the dimensions using the comparison function defined on ``ComparisonTestCase``.

holoviews/plotting/mpl/util.py
```diff
@@ -164,3 +165,20 @@ def get_tight_bbox(fig, bbox_extra_artists=[], pad=None):
         bbox_extra = TransformedBbox(_bbox, trans)
         bbox_inches = Bbox.union([bbox_inches, bbox_extra])
     return bbox_inches.padded(pad) if pad else bbox_inches
+
+
+def get_raster_data(image):
```
jlstevens: Wouldn't ``get_raster_array`` be a more accurate name?

philippjfr: Sure.

holoviews/core/data/image.py
```diff
@@ -0,0 +1,250 @@
...
+
+        kwargs = {}
+        dimensions = [d.name if isinstance(d, Dimension) else
+                      d for d in kdims + vdims]
+        if isinstance(data, tuple):
+            data = dict(zip(dimensions, data))
+        if isinstance(data, dict):
+            xs, ys = np.asarray(data[kdims[0].name]), np.asarray(data[kdims[1].name])
+            l, r, xdensity, invertx = util.bound_range(xs, None)
+            b, t, ydensity, inverty = util.bound_range(ys, None)
```
jlstevens: Will need replacing (see comment below). I think that at least l,r,b and t can be computed without resorting to ``np.diff``. The densities too I assume.

holoviews/core/data/image.py
```diff
@@ -0,0 +1,250 @@
...
+    @classmethod
+    def validate(cls, dataset):
+        pass
+
+    @classmethod
+    def redim(cls, dataset, dimensions):
+        return dataset.data
+
+    @classmethod
+    def reindex(cls, columns, kdims=None, vdims=None):
```
jlstevens: Does this transpose appropriately if you swap kdims?

holoviews/core/data/image.py
```diff
@@ -0,0 +1,250 @@
...
+            # Raster arrays are stored with different orientation
+            # than expanded column format, reorient before expanding
+            if dataset.data.ndim > 2:
+                data = dataset.data[:, :, dim_idx-dataset.ndims]
+            else:
+                data = dataset.data
+            data = np.flipud(data)
+            return data.T.flatten() if flat else data
+        else:
+            return None
```
jlstevens: Isn't this an error condition?

holoviews/core/data/image.py
```diff
@@ -0,0 +1,250 @@
...
+    def values(cls, dataset, dim, expanded=True, flat=True):
+        """
+        The set of samples available along a particular dimension.
+        """
+        dim_idx = dataset.get_dimension_index(dim)
+        if dim_idx in [0, 1]:
+            l, b, r, t = dataset.bounds.lbrt()
+            dim2, dim1 = dataset.data.shape[:2]
+            d1_half_unit = float(r - l)/dim1/2.
+            d2_half_unit = float(t - b)/dim2/2.
```
jlstevens: Maybe there could be a helper method that takes whatever info you've got and picks an approach to compute half-distances?

holoviews/core/data/image.py
```diff
@@ -0,0 +1,250 @@
...
+        l, b, r, t = dataset.bounds.lbrt()
+        if isinstance(xidx, slice):
+            l = l if xidx.start is None else max(l, xidx.start)
+            r = r if xidx.stop is None else min(r, xidx.stop)
+        if isinstance(yidx, slice):
+            b = b if yidx.start is None else max(b, yidx.start)
+            t = t if yidx.stop is None else min(t, yidx.stop)
+        bounds = BoundingBox(points=((l, b), (r, t)))
+        slc = Slice(bounds, dataset)
+        data = slc.submatrix(dataset.data)
```
jlstevens: Could return one line earlier :-)

holoviews/core/data/image.py
philippjfr: Probably should be (or is already) handled at the Dataset level, I think it's for cdims.

holoviews/core/data/image.py
```diff
@@ -0,0 +1,250 @@
...
+        returning a reduced dimensionality type, which is either
+        a ItemTable, Curve or Scatter. If two dimension samples
+        and a new_xaxis is provided the sample will be the value
+        of the sampled unit indexed by the value in the new_xaxis
+        tuple.
+        """
+        if len(samples[0]) == 1:
+            return dataset.select(**{dataset.kdims[0].name:
+                                     [s[0] for s in samples]}).columns()
+        else:
```
jlstevens: No real need for the else as you are returning from the other branch. I like this method, it is relatively simple. :-)

holoviews/core/data/image.py
```diff
@@ -0,0 +1,250 @@
...
+        Sample the Raster along one or both of its dimensions,
+        returning a reduced dimensionality type, which is either
+        a ItemTable, Curve or Scatter. If two dimension samples
+        and a new_xaxis is provided the sample will be the value
+        of the sampled unit indexed by the value in the new_xaxis
+        tuple.
+        """
+        if len(samples[0]) == 1:
+            return dataset.select(**{dataset.kdims[0].name:
+                                     [s[0] for s in samples]}).columns()
```
philippjfr: Should return a tuple to let it automatically pick an interface.

holoviews/core/data/image.py
```diff
@@ -0,0 +1,250 @@
...
+            group_kwargs['kdims'] = kdims
+        group_kwargs.update(kwargs)
+
+        if len(dimensions) == 1:
+            didx = dataset.get_dimension_index(dimensions[0])
+            coords = dataset.dimension_values(dimensions[0], False)
+            xvals = dataset.dimension_values(abs(didx-1), False)
+            samples = [(i, slice(None)) if didx else (slice(None), i)
+                       for i in range(dataset.data.shape[abs(didx-1)])]
+            if didx:
```
jlstevens: I think didx is an integer index. If that is so, this is really a test about whether it is zero or not - if that is true, it is worth being explicit.

holoviews/core/data/image.py
```diff
@@ -0,0 +1,250 @@
...
+                samples = samples[::-1]
+                data = dataset.data
+            else:
+                data = dataset.data[::-1, :]
```
jlstevens: I can't follow but in one case you invert the samples..and in another you flip the data about an axis? A comment might help...

holoviews/core/data/image.py
```diff
@@ -0,0 +1,250 @@
...
+        group_type = dict if group_type == 'raw' else group_type
+        if issubclass(group_type, Element):
+            group_kwargs.update(util.get_param_values(dataset))
+            group_kwargs['kdims'] = kdims
+        group_kwargs.update(kwargs)
+
+        if len(dimensions) == 1:
+            didx = dataset.get_dimension_index(dimensions[0])
+            coords = dataset.dimension_values(dimensions[0], False)
+            xvals = dataset.dimension_values(abs(didx-1), False)
```
jlstevens: I forget what the ``False`` is about - showing the keyword might help me remember!

holoviews/element/raster.py
```diff
@@ -73,14 +70,29 @@ def __getitem__(self, slices):
...
+        elif not expanded and dim_idx == 1:
+            return np.array(range(self.data.shape[0]))
+        elif dim_idx in [0, 1]:
+            values = np.mgrid[0:self.data.shape[1], 0:self.data.shape[0]][dim_idx]
+            return values.flatten() if flat else values
+        elif dim_idx == 2:
+            arr = self.data.T
+            return arr.flatten() if flat else arr
+        else:
+            return super(Raster, self).dimension_values(dim)
```
jlstevens: Might be worth leaving a comment saying the last branch is assuming you are left with cdims...

holoviews/element/raster.py
```diff
@@ -408,45 +211,157 @@ class Image(SheetCoordinateSystem, Raster):
     access to the data, via the .data attribute.
     """
 
-    bounds = param.ClassSelector(class_=BoundingRegion, default=BoundingBox(), doc="""
+    bounds = param.ClassSelector(class_=BoundingRegion, default=None, doc="""
```
jlstevens: Where does the assumption of unit bounds now come from? I'll edit this comment if I spot it later...

holoviews/element/raster.py
```diff
@@ -408,45 +211,157 @@ class Image(SheetCoordinateSystem, Raster):
...
     group = param.String(default='Image', constant=True)
 
+    kdims = param.List(default=[Dimension('x'), Dimension('y')],
+                       bounds=(2, 2), constant=True, doc="""
+        The label of the x- and y-dimension of the Raster in form
+        of a string or dimension object.""")
+
     vdims = param.List(default=[Dimension('z')],
-                       bounds=(1, 1), doc="""
+                       bounds=(1, None), doc="""
```
jlstevens: I think this should be ``bounds=(1,1)`` to be explicit - better than making an arbitrary choice about which vdim to use (e.g the first one).

holoviews/element/raster.py
```diff
@@ -408,45 +211,157 @@ class Image(SheetCoordinateSystem, Raster):
...
-        if np.isscalar(bounds):
+        extents = extents if extents else (None, None, None, None)
+        if data is None: data = np.array([[0]])
+        Dataset.__init__(self, data, extents=extents, bounds=None, **params)
+
+        if isinstance(data, np.ndarray):
+            dim2, dim1 = data.shape[:2]
+        else:
+            dim1, dim2 = tuple([len(self.dimension_values(kd, expanded=False))
+                                  for kd in self.kdims])
```
jlstevens: Shouldn't the second version always work, even if having the first case is more efficient? Just checking - not saying you should change it.

holoviews/element/raster.py
```diff
@@ -408,45 +211,157 @@ class Image(SheetCoordinateSystem, Raster):
...
+        if self.bounds is not None:
+            bounds = self.bounds
+        elif bounds is None and isinstance(data, np.ndarray):
+            bounds = BoundingBox()
```
jlstevens: Ah this answers my question about the unit bounding box. Editing my earlier comment.

holoviews/element/raster.py
jlstevens: Though why was it moved to the constructor? Isn't this still a sensible default?

holoviews/element/raster.py
philippjfr: No, calling .dimension_values at this point of the constructor would throw an exception, the sheetcoordinatesystem isn't initialized yet.

holoviews/element/raster.py
```diff
@@ -408,45 +211,157 @@ class Image(SheetCoordinateSystem, Raster):
...
+
+        if bounds is None:
+            xvals = self.dimension_values(0, False)
+            l, r, xdensity, _ = util.bound_range(xvals, xdensity)
+            yvals = self.dimension_values(1, False)
+            b, t, ydensity, _ = util.bound_range(yvals, ydensity)
+            bounds = BoundingBox(points=((l, b), (r, t)))
```
jlstevens: Not too bad. I assume we can't just get it from the interface as that would mean accessing special API unique to Image?

holoviews/element/raster.py
```diff
@@ -408,45 +211,157 @@ class Image(SheetCoordinateSystem, Raster):
...
+        """
+        self.__dict__ = state
+        if isinstance(self.data, np.ndarray):
+            self.interface = ImageInterface
+        super(Dataset, self).__setstate__(state)
+
+
+    def aggregate(self, dimensions=None, function=None, spreadfn=None, **kwargs):
+        agg = super(Image, self).aggregate(dimensions, function, spreadfn, **kwargs)
+        return Curve(agg) if isinstance(agg, Dataset) and len(self.vdims) == 1 else agg
```
jlstevens: I thought aggregate would already pick the appropriate element - or is that part of the proposed reductions PR?

holoviews/element/raster.py
```diff
@@ -408,45 +211,157 @@ class Image(SheetCoordinateSystem, Raster):
...
+        indexed = False
+        if not any([isinstance(el, slice) for el in coords]):
+            indexed = True
+            selection = {kd.name: c for kd, c in zip(self.kdims, self.closest(coords))}
+
+        # Compute new bounds
+        shape = self.interface.shape(self)
+        ys, xs = shape[:2]
+        xidx, yidx = coords
+        xdim, ydim = self.kdims
```
jlstevens: Looks like xdim and ydim could be moved (not used immediately here).

Could ``compute_new_bounds`` be a utility? 

Something like ``compute_new_bounds(shape, coords, lbrt, xdensity, ydensity)``

holoviews/element/raster.py
```diff
@@ -408,45 +211,157 @@ class Image(SheetCoordinateSystem, Raster):
...
+
+        # Apply scalar and list indices
+        l, b, r, t = slc.compute_bounds(self).lbrt()
+        if indexed:
+            pass
+        elif not isinstance(xidx, slice):
+            if not isinstance(xidx, (list, set)): xidx = [xidx]
+            if len(xidx) > 1:
+                xdensity = xdensity*(float(len(xidx))/xs)
+            idxs = []
```
jlstevens: As far as I can tell, ``idxs`` is declared twice as an empty list and never used.

holoviews/element/raster.py
```diff
@@ -408,45 +211,157 @@ class Image(SheetCoordinateSystem, Raster):
...
+            if len(yidx) > 1:
+                ydensity = ydensity*(float(len(yidx))/ys)
+            idxs = []
+            bs, ts = [], []
+            for idx in yidx:
+                _, yc = self.closest_cell_center(l, idx)
+                bs.append(yc-yunit/2)
+                ts.append(yc+yunit/2)
+            b, t = np.min(bs), np.max(ts)
+            selection[ydim.name] = slice(b, t)
```
jlstevens: I guess these branches are overriding ``selection`` as necessary?

holoviews/element/raster.py
```diff
@@ -487,135 +401,37 @@ def closest(self, coords=[], **kwargs):
             return [getter(self.closest_cell_center(*el)) for el in coords]
 
 
-    def __getitem__(self, coords):
+    def range(self, dim, data_range=True):
+        idx = self.get_dimension_index(dim)
+        dimension = self.get_dimension(dim)
+        rng = super(Image, self).range(dim, data_range)
```
jlstevens: I would deconstruct straight to ``low, high`` instead of introducing the weird ``rng`` name.

holoviews/element/raster.py
```diff
@@ -487,135 +401,37 @@ def closest(self, coords=[], **kwargs):
...
-            # Raster arrays are stored with different orientation
-            # than expanded column format, reorient before expanding
-            data = np.flipud(self.data).T
-            return data.flatten() if flat else data
-        else:
-            super(Image, self).dimension_values(dim)
-
-
-
-class GridImage(Dataset, Element2D):
```
jlstevens: This is the bit I like the most!

holoviews/element/raster.py
philippjfr: Ah, yes, leftover from when computing the bounds and slicing the data were still combined.

holoviews/element/raster.py
jlstevens: Ok, thanks for clarifying.

holoviews/element/raster.py
philippjfr: Yes, it allows providing a slice and a coordinate.

jlstevens:
I've completed a pass over the PR and made a fair number of comments. It is a big PR and I can't claim to have followed all the detailed logic but at least I've commented on the things that stood out.

Because there is so much here, I would like to merge it ASAP once my queries have been answered so we have a decent amount of time to test it and fix things.

holoviews/core/data/interface.py
jlstevens: Just to say this is a minor comment, I don't mind if you don't address it.

holoviews/core/data/__init__.py
philippjfr: Just checked this is crucial for Image to work correctly, 32 test failures without it.

holoviews/core/data/__init__.py
philippjfr: Seemingly used a few times at least on Image, so I think making the API consistent is reasonable.

jlstevens: > ... so I think making the API consistent is reasonable.

To be clear, that would mean requiring the list format?

philippjfr: No, it would mean to continue supporting both, the idea being that it matches .sample, .reduce, .select signatures.

holoviews/core/data/__init__.py
philippjfr: Refactored it slightly differently.

holoviews/operation/element.py
philippjfr: They are tested via that gif in the one of notebook tests but thinking about it I suppose the operation is symmetric. But apparently I also added some unit tests for these.

holoviews/plotting/bokeh/__init__.py
philippjfr: Unless I commit to releasing a geoviews version right away ``gv.Image`` inheritance would be broken by the function suggestion.

philippjfr: There is another issue removing GridImage, users will have used ``GridImage`` as an option. Is there any way to add Element option aliases, i.e. it would apply options set on ``GridImage`` to ``Image``, I vaguely recall something like that. That may be a good place to put a deprecation warning too.

jlstevens: Please assign that as an issue to me. I think it should be fairly easy to do.

philippjfr: Done here: https://github.com/ioam/holoviews/issues/1252

holoviews/core/data/image.py
philippjfr: It should, I opened an issue: https://github.com/ioam/holoviews/issues/1253

holoviews/element/raster.py
philippjfr: Correct.

holoviews/element/raster.py
philippjfr: Right, part of the reductions PR again.

philippjfr:
Ready for final review and merging once test pass.

philippjfr:
Nevermind still need to make a few changes.

philippjfr:
Ready to merge again, some issues reindexing gridded datasets cropped up but are fixed now.

holoviews/core/util.py
```diff
@@ -1210,3 +1210,19 @@ def is_nan(x):
         return np.isnan(x)
     except:
         return False
+
+
+def bound_range(vals, density):
+    """
+    Computes a bounding range from a number of evenly spaced samples.
+    Will raise an error if samples are not evenly spaced within
+    tolerance.
```
jlstevens: I would just mention which tolerance - namely as reported by ``sys.float_info``

philippjfr: No longer raises an error, so yeah will fix it.

holoviews/element/util.py
```diff
@@ -52,6 +43,54 @@ def compute_edges(edges):
     return np.concatenate([edges, [edges[-1]+width]])
 
 
+def compute_slice_bounds(slices, scs, shape):
```
jlstevens: Nice to have this code split out as a utility.

jlstevens: I'm not sure I would recommend adding tests for this utility right now as I do want to merge ASAP. At least it is much easier to test now!

jlstevens:
Looks good. I made one comment about a docstring - other than that, I'll merge when you tell me it is ready (and the tests are green).

philippjfr:
Already pushed the updated docstring, just waiting on tests now.

jlstevens:
All tests are green! Merging!

PR #1004, opened 2016-12-07
Fixed hover tooltips for Curve and Path types

philippjfr:
Currently dimensions that shouldn't be are appearing in the hover tooltips. Additionally on ``CurvePlot`` the NdOverlay dimension information isn't supplied at all. I've also made various fixes to make color and alpha option work correctly in batched mode.

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -143,16 +143,14 @@ class CurvePlot(ElementPlot):
 
     def _hover_tooltips(self, element):
         if self.batched:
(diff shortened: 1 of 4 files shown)


jbednar:
Looks good to me, and would be very welcome!

philippjfr:
There is a real bokeh notebook test failure because the previous data included wrong hover label. Will have to rebuild. I've also added various unit tests to check the correct hover information is set.

jlstevens:
Looks good to me too and the tests have passed. Ready to merge?

philippjfr:
Yep, go ahead.

PR #1009, opened 2016-12-10
Define Dimension formatters for numpy integer types

philippjfr:
As described in https://github.com/ioam/holoviews/issues/975, widgets do not correctly format integers at the moment. This is easily fixed by defining ``Dimension.type_formatters`` for the numpy integer types.

### DIFF
--- a/holoviews/core/__init__.py
+++ b/holoviews/core/__init__.py
@@ -14,6 +14,12 @@ archive = FileArchive()
 
 # Define default type formatters
 Dimension.type_formatters[int] = "%d"
+Dimension.type_formatters[np.uint16] = '%d'
+Dimension.type_formatters[np.int16] = '%d'
+Dimension.type_formatters[np.uint32] = '%d'
+Dimension.type_formatters[np.int32] = '%d'
+Dimension.type_formatters[np.uint64] = '%d'
+Dimension.type_formatters[np.int64] = '%d'
 Dimension.type_formatters[float] = "%.5g"
 Dimension.type_formatters[np.float32] = "%.5g"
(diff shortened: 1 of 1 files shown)


jlstevens:
Looks good and I a happy to merge once the tests are passing. 

The only suggestion I have is to investigate whether there is an easy way to iterate over all numpy integer types - if so, iterating over them would be better than having to list them all explicitly this way.

philippjfr:
>The only suggestion I have is to investigate whether there is an easy way to iterate over all numpy integer types.

As long as we are treating them all as separate types I prefer to be explicit about it. I do think that rather than specifying it by an actual type it might be more useful to declare that all integer types should respect the formatter. May we should consider whether ``int`` and ``float`` formatters should also apply to all numpy where ``dtype.kind=='i'`` and ``dtype.kind=='f'`` if no more specific formatter has been applied.

philippjfr:
Apparently this has resulted in small changes in the bokeh table tests. Will update.

philippjfr:
@jlstevens Any opinions on my comment above?

jlstevens:
> ...it might be more useful to declare that all integer types should respect the formatter

I think I agree that this makes sense.

> May we should consider whether int and float formatters should also apply to all numpy types where dtype.kind=='i' and dtype.kind=='f' respectively, if no more specific formatter has been set.

Sounds reasonable for scalars...I assume the formatters won't do a good job for arbitrary arrays.

philippjfr:
Let's go with this for now. Rebuilding test data shortly.

jlstevens:
The pr build passed. Merging.

PR #1010, opened 2016-12-10
Fixed bokeh FuncTickFormater

philippjfr:
As described in https://github.com/ioam/holoviews/issues/885 in bokeh 0.12.3 the signature of python FuncTickFormatters was changed in a way that is inconsistent with formatters in holoviews. This PR adds a utility that compiles the python formatter with pyscript directly and then patches the generated JS code so the arg are made available in the namespace rather than being passed to the function. It's not exactly pretty but I don't see a way around it. I've also added some unit tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ install:
   - conda info -a
   - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION scipy numpy freetype nose bokeh pandas jupyter ipython=4.2.0 param pyqt=4 matplotlib=1.5.1 xarray datashader
   - source activate test-environment
-  - conda install -c conda-forge  -c scitools iris sip=4.18 plotly
+  - conda install -c conda-forge  -c scitools iris sip=4.18 plotly flexx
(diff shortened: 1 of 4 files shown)


jlstevens:
I've never been entirely happy with the ``flexx`` dependency but that isn't something newly introduced in this PR. The mix of Python inspection, flexx, Javascript and regular expressions is quite unholy but it at least this scary code has now been confined to its own function.

Just as a shot in the dark, is there any chance AST tricks could be applied to the python function before passing it to ``py2js`` instead of manipulating the javascript output? Not that such an approach would be much better! 

As I know you are aware, this isn't a nice approach and I'm afraid I don't have any better suggestions right now. Having a working tick formatter for bokeh is more important than avoiding one nasty function and on that basis, I'll now merge.

philippjfr:
>The mix of Python inspection, flexx, Javascript and regular expressions is quite unholy but it at least this scary code has now been confined to its own function.

Agreed, definitely not pretty.

>Just as a shot in the dark, is there any chance AST tricks could be applied to the python function before passing it to py2js instead of manipulating the javascript output? Not that such an approach would be much better!

Looked into this briefly but found no nice approach that actually worked, especially across python versions.

PR #1011, opened 2016-12-10
Raise warning when size_index set to non-numeric dimension

philippjfr:
As suggested in https://github.com/ioam/holoviews/issues/982 currently you get an uninformative exception when setting the size_index option to a non-numeric dimension. This raises a warning and adds unit tests to ensure the warning is raised.

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -67,15 +67,22 @@ class PointPlot(ColorbarPlot):
         sdim = element.get_dimension(self.size_index)
         if sdim:
             map_key = 'size_' + sdim.name
-            mapping['size'] = map_key
             if empty:
                 data[map_key] = []
+                mapping['size'] = map_key
             else:
                 ms = style.get('size', np.sqrt(6))**2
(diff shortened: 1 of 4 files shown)


jlstevens:
I think you could shorten this a bit and avoid some repetition with something like this:

```
sizes =  self._apply_size_index(element, self.size_index)
```

This could be implemented in a common base class and would always return ``sizes``, issuing the warning when  the dimension type is inappropriate.

philippjfr:
>This could be implemented in a common base class and would always return sizes, issuing the warning when the dimension type is inappropriate.

I've so far refrained from shared baseclasses for ElementPlots because a fair amount of the API is different. I'd be happy to factor it out into the shared ``compute_sizes`` function that already exists.

jlstevens:
>  I'd be happy to factor it out into the shared compute_sizes function that already exists.

That also makes sense.

philippjfr:
Unfortunately I didn't manage to factor out much, the warning needs the dimension and element name to be useful and I didn't want to add even more arguments to the ``compute_sizes`` function.

jlstevens:
Ok, nevermind then. Merging.

PR #1012, opened 2016-12-10
Temporary fix for RasterGridPlot

philippjfr:
As described in https://github.com/ioam/holoviews/issues/976 RasterGridPlot currently fails because it has no streams attribute. This is because it's a very atypical plot type that does not currently support streams, making streams and comms work with it will be some effort. Once this has been merged we should open an issue to add that support.


philippjfr:
Not sure why the tests didn't fail here but the broken plot in Exploring_Data seems to be working.

philippjfr:
@jlstevens Ready to merge, just a very tiny stub for the time being.

jlstevens:
> Once this has been merged we should open an issue to add that support.

Please do!

PR #1013, opened 2016-12-10
Add show_frame option for bokeh plots

philippjfr:
Currently the ``show_frame`` option that's available in matplotlib is not implemented in bokeh. This suppresses the outline by setting it to zero alpha.

``show_frame=True``

bokeh_plot 23

``show_frame=False``

bokeh_plot 22


jlstevens:
Looks good. Merging.

Issue #727, opened 2016-06-21
Boolean arrays in Bokeh rasters


jlstevens:
Using boolean arrays in `Images` when plotting with Bokeh breaks:

``` python
/Users/jstevens/miniconda/envs/analytics/lib/python2.7/site-packages/bokeh/core/properties.pyc in prepare_value(self, cls, name, value)
    270                     break
    271             else:
--> 272                 raise e
    273         else:
    274             value = self.transform(value)

ValueError: expected a value of type Real, got True of type bool_
```

As this works fine in matplotlib and can be fixed by casting the booleans to a numeric type, I'm filing this as a bug. I suspect this will be an issue for all raster types, not just `Image`.

philippjfr:
The PR fixing this was merged. Closing.

PR #1014, opened 2016-12-10
Added support for boolean arrays in bokeh ImagePlot

philippjfr:
As described in #727, the bokeh image plots do not currently support boolean arrays. This is because the color mapper does not like boolean ``low`` and ``high`` values. However since booleans cannot be serialized easily it will be more efficient to send them as ``int8`` arrays, which will be base64 encoded in an upcoming bokeh PR.


jlstevens:
Looks good and this is an issue that has bitten me several times before. Good to see it fixed!

Ready to merge?

philippjfr:
Ready.

PR #1016, opened 2016-12-11
Fixed saving to html and json in Python 3

philippjfr:
Fixes https://github.com/ioam/holoviews/issues/979 which ended up throwing bugs in ``Renderer.save`` in python3 when a string type was written to file. By defining the MIME type the encoding is applied correctly.

Needs some tests before merging.

### DIFF
--- a/holoviews/plotting/renderer.py
+++ b/holoviews/plotting/renderer.py
@@ -361,12 +361,13 @@ class Renderer(Exporter):
             widget = obj
 
         html = self_or_cls.static_html(widget, fmt, template)
+        encoded = self_or_cls.encode((html, {'mime_type': 'text/html'}))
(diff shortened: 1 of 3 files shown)


jlstevens:
Make sure to test this change with the archive functionality. If I remember right, there is code that needs to use the mime type to figure out how to encode the data (bytes vs strings).

philippjfr:
>If I remember right, there is code that needs to use the mime type to figure out how to encode the data (bytes vs strings).

That's exactly the bit that was failing before I defined these. Will definitely test that.

philippjfr:
Ready once tests pass.

jlstevens:
Looks good. I'll merge now on the understanding that you did test this with ``hv.archive``.

Issue #989, opened 2016-11-16
Add Layout/Grid title support for bokeh


philippjfr:
Currently we do not support an overall title on Grid and Layout based plots in bokeh. This issue https://github.com/bokeh/bokeh/issues/5463 suggests you can add an overall title for a plot by adding a ``Div`` element and laying it out in a column. We should implement this for better consistency with matplotlib.

philippjfr:
Corresponding PR was just merged. Closing.

PR #1017, opened 2016-12-11
Add support for Layout/Grid titles in bokeh

philippjfr:
As I had suggested in #989 it is possible to add a Div as a title to a composite plot. This finally adds consistency in titles between matplotlib and bokeh backends for Layout and Grid plots.

Still needs tests.

### Layout

screen shot 2016-12-11 at 10 24 55 pm

### Grid

screen shot 2016-12-11 at 10 27 19 pm

### DIFF
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -155,6 +155,11 @@ class BokehPlot(DimensionedPlot):
 
 
 class CompositePlot(BokehPlot):
+    """
+    CompositePlot is an abstract baseclass for plot types that draw
+    render multiple axes. It implements methods to add an overall title
+    to such a plot.
+    """
 
     fontsize = param.Parameter(default={'title': '16pt'}, allow_None=True,  doc="""
        Specifies various fontsizes of the displayed text.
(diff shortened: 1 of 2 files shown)


jlstevens:
That looks good! Shame bokeh doesn't let you do this more directly.

Make sure to test this in Chrome/Firefox/Safari and if you find it is working well, I am happy to merge.

philippjfr:

holoviews/plotting/bokeh/plot.py
```diff
@@ -153,7 +154,43 @@ def sync_sources(self):
 
 
 
-class GridPlot(BokehPlot, GenericCompositePlot):
+class CompositePlot(BokehPlot):
+
```
jlstevens: Would be good to have a docstring explaining what this base class is for. Right now, everything ``CompositePlot`` does seems to be only needed by ``LayoutPlot``. Any ideas what ``CompositePlot`` could be used for that isn't ``LayoutPlot``?

philippjfr: GridPlot, just below.

jlstevens:
Looks good, thanks!

PR #1023, opened 2016-12-18
find_file utility handles absolute path

philippjfr:
Ensure the ``find_file`` utility can handle an absolute filename path passed to it. The function ensures that the default widget assets get resolved correctly but the user should be able to override these with an absolute path.  

Concretely you can now supply custom CSS to the widgets to change the layout a bit. Setting:

```
hv.plotting.widgets.SelectionWidget.css = '/Users/philippjfr/Downloads/test.css'
```

can give you something like this:

screen shot 2016-12-18 at 7 41 20 pm

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -758,12 +758,17 @@ def get_spec(obj):
 
 def find_file(folder, filename):
     """
-    Find a file given folder and filename.
+    Find a file given folder and filename. If the filename can be
+    resolved directly returns otherwise walks the supplied folder.
     """
     matches = []
+    if os.path.isabs(filename) and os.path.isfile(filename):
+        return filename
     for root, _, filenames in os.walk(folder):
-        for filename in fnmatch.filter(filenames, filename):
-            matches.append(os.path.join(root, filename))
+        for fn in fnmatch.filter(filenames, filename):
+            matches.append(os.path.join(root, fn))
+    if not matches:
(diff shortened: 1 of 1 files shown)


philippjfr:
Ready after review.

holoviews/core/util.py
```diff
@@ -729,12 +729,17 @@ def get_spec(obj):
...
-    Find a file given folder and filename.
+    Find a file given folder and filename. If the filename can be 
+    resolved directly 
     """
     matches = []
+    if os.path.isabs(filename) and os.path.isfile(filename):
+        return filename
     for root, _, filenames in os.walk(folder):
         for filename in fnmatch.filter(filenames, filename):
             matches.append(os.path.join(root, filename))
```
jlstevens: Might be worth avoid overwriting the ``filename`` argument with the return values of ``fnmatch.filter`` (i.e it might be worth calling it ``fname`` in this loop).

philippjfr: Good catch.

philippjfr:
Addressed your comment, ready to merge after tests pass.

jlstevens:
Just to say that I do think being to change the layout using CSS would be useful. Ideally, we could provide alternate CSS stylesheets and an API to switch them between them that is nicer than setting file paths.

Anyway, that could be a separate feature request and this PR would help make it possible. As the tests have passed, I'll merge now.

PR #1027, opened 2016-12-21
New diagonal_operation-Parameter for the gridmatrix operation

mrksr:
Instead of being restricted to a plot type along the diagonal of a gridmatrix-operation, this pull request also allows the specification of an operation. The main reasoning for this is that it is now possible to change parameters like the number of bins in a diagonal histogram:
```python
data = pd.DataFrame(np.random.normal(size=(250, 3)), columns=['a', 'b', 'c'])
hvset = hv.Dataset(data.reset_index(), kdims=['index'])

hv.operation.gridmatrix(
    hvset,
    diagonal_operation=hv.operation.histogram.instance(num_bins=30, adjoin=False)
)
```

I think in its current state, this PR is not ready however, since there are a few open questions:
- Should `diagonal_type` remain?
- Should there be a `chart_operation` parameter?
- How to formulate the docstrings to be easy to understand?
- Is the normalization relevant for the "operation-histograms"?

When the semantics are decided, I can also add a few tests. I will probably need a few hints however on how to test this.

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -699,9 +699,14 @@ class gridmatrix(param.ParameterizedFunction):
         The Element type used to display bivariate distributions
         of the data.""")
 
-    diagonal_type = param.Parameter(default=Histogram, doc="""
+    diagonal_type = param.Parameter(default=None, doc="""
        The Element type along the diagonal, may be a Histogram or any
-       other plot type which can visualize a univariate distribution.""")
+       other plot type which can visualize a univariate distribution.
+       This parameter overrides diagonal_operation.""")
+
+    diagonal_operation = param.Parameter(default=histogram, doc="""
+       The operation applied along the diagonal, may be a histogram-operation
+       or any other function which returns a viewable element.""")
 
     overlay_dims = param.List(default=[], doc="""
        If a HoloMap is supplied this will allow overlaying one or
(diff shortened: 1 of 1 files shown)


philippjfr:
Sorry I've been so slow in reviewing this, I'll get to it over the weekend.

holoviews/operation/element.py
```diff
@@ -694,17 +699,24 @@ def _process(self, p, element, ranges={}):
...
-                                      adjoin=False)(norm=dict(axiswise=True, framewise=True))
-                else:
+                if p.diagonal_type is not None:
                     values = element.dimension_values(d1)
                     el = p.diagonal_type(values, vdims=[d1])
+                elif isinstance(p.diagonal_operation, histogram):
+                    bin_range = ranges.get(d1.name, element.range(d1))
+                    el = p.diagonal_operation(element,
+                                              dimension=d1.name,
+                                              bin_range=bin_range)
```
philippjfr: I'm afraid this won't behave very well unless axiswise normalization is enabled here.

philippjfr:
Sorry for the long delay, this looks fine, except one small comment. If you don't think you'll have the time I'm happy to merge anyway and apply the change myself.

mrksr:
Your comment should be fixed now.

philippjfr:
Perfect thanks! I'm actually going to clean up the histogram operation a bit to remove the ``adjoin`` nonsense so this can be cleaned up further. For the time being this looks good and I'll merge as soon as tests are passing.

philippjfr:
Unless you feel like writing some simple tests that is?

mrksr:
I'm not sure on how to test this. As I cannot find tests for the other operations, can you point to something similar?

philippjfr:
You're right, it appears operations are not currently tested at all! I'd create a new file called ``testoperation.py`` and then subclass the ``ComparisonTestCase`` class. The ``assertEqual`` on that class knows how to compare HoloViews objects so a few tests that manually construct GridMatrix and compare it to the gridmatrix output should be fine. I'd be happy to add some basic operation tests myself but will probably not get to it until fairly late today.

philippjfr:
Could you rebase the PR, I'll get this merged and add unit tests separately.

mrksr:
Sorry I did not find the time to look into the tests yet. I will do the rebase tomorrow!

mrksr:
@philippjfr I rebased with c563965 in mind.

philippjfr:
Sorry this has stalled so long, I think tests are still failing because framewise normalization is not enabled. I'll fix the tests in a separate PR.

PR #1029, opened 2016-12-27
Implement dynamic relabel, redim and select methods

philippjfr:
As outlined in https://github.com/ioam/holoviews/issues/422 this PR implements some further methods on DynamicMap, ensuring that the ``relabel`` and ``redim`` methods can be used correctly on a DynamicMap. Dynamic methods like this are easily implemented using the ``Dynamic`` utility, which currently lives in ``hv.util``. Since this forces me to use inline imports I'm now considering moving ``Dynamic`` into core.


holoviews/core/dimension.py
```diff
@@ -344,13 +344,15 @@ def relabel(self, label=None, group=None, depth=0):
         Assign a new label and/or group to an existing LabelledData
         object, creating a clone of the object with the new settings.
         """
-        keywords = [('label',label), ('group',group)]
-        obj = self.clone(self.data,
-                         **{k:v for k,v in keywords if v is not None})
+        new_data = self.data
         if (depth > 0) and getattr(obj, '_deep_indexable', False):
-            for k, v in obj.items():
-                obj[k] =  v.relabel(group=group, label=label, depth=depth-1)
```
philippjfr: Reimplemented this because assigning this way is inefficient for an internal operation.

philippjfr:
Would like to get deep-select and ``__getitem__`` into this PR before merge.

philippjfr:
@jlstevens This still needs a lot of unit tests for the ``select`` and ``__getitem__`` methods, but the actual implementation is now ready for review.

holoviews/core/dimension.py
```diff
@@ -767,8 +770,9 @@ def select(self, selection_specs=None, **kwargs):
 
         # Apply selection to self
         if local_kwargs and matches:
-            ndims = (len(self.dimensions()) if any(d in self.vdims for d in kwargs)
-                     else self.ndims)
+            ndims = self.ndims
+            if any(d in self.vdims for d in kwargs):
+                ndims = len(self.kdims+self.vdims)
```
jlstevens: Is this because ``len(self.dimensions())`` includes ``cdims``?

philippjfr: No, because it can include the deep dimensions as well which should be processed below.

holoviews/util.py
```diff
@@ -28,14 +28,17 @@ class Dynamic(param.ParameterizedFunction):
     kwargs = param.Dict(default={}, doc="""
         Keyword arguments passed to the function.""")
 
+    shared_data = param.Boolean(default=False, doc="""
+        Whether the cloned DynamicMap will share the same data.""")
```
jlstevens: As ``.data`` for a ``DynamicMap`` is the callback, it would be good to explain what sharing/not sharing means here. I'm not quite sure what it means to be honest so it might be something we want to document in other places too.

philippjfr: ``.data`` on a DynamicMap is not the callback but the cache. Will update the docstring to reflect that.

holoviews/core/spaces.py
```diff
@@ -655,7 +654,7 @@ def reset(self):
         return self
 
 
-    def _cross_product(self, tuple_key, cache):
+    def _cross_product(self, tuple_key, cache, data_slice):
         """
         Returns a new DynamicMap if the key (tuple form) expresses a
         cross product, otherwise returns None. The cache argument is a
```
jlstevens: Would be good to explain the new ``data_slice`` argument in the docstring.

philippjfr: Agreed, docstrings still need updating. I'd love it if there was some way to inherit method docstrings easily, when you override a method.

holoviews/util.py
jlstevens: Ah yes, thanks! I suppose you might want to copy the cache in some cases (e.g for ``relabel``).

jlstevens:
Instead of the inline imports:

```
   from ..util import Dynamic
```

Maybe it is time we add ``core.dynamic`` which includes ``DynamicMap`` and ``Dynamic``? Or would this just result in different import issues?

holoviews/util.py
philippjfr: Right, it's probably mostly going to be used internally.

philippjfr:
>Maybe it is time we add core.dynamic which includes DynamicMap and Dynamic? Or would this just result in different import issues?

I think HoloMap.__mul__ would have to inline import it, but I think that may be still be preferable, so I'd vote for it.

holoviews/core/spaces.py
```diff
@@ -752,10 +780,33 @@ def __getitem__(self, key):
...
 
+        if data_slice:
+            val = self._dataslice(val, data_slice)
         self._cache(tuple_key, val)
         return val
 
 
+    def select(self, selection_specs=None, **kwargs):
+        selection = super(DynamicMap, self).select(selection_specs, **kwargs)
+        def dynamic_select(obj):
```
jlstevens: Would ``dynamic_select`` have any use on its own in ``util``? It seems to offer a ``select`` that works with both dynamic and non dynamic objects..

holoviews/core/spaces.py
```diff
@@ -708,22 +736,22 @@ def __getitem__(self, key):
...
         (for one of the 'open' modes)
         """
-        tuple_key = util.wrap_tuple_streams(key, self.kdims, self.streams)
+        # Split key dimensions and data slices
+        if key is Ellipsis:
+            return self
+        elif key == ():
+            map_slice, data_slice = (), ()
+        else:
+            map_slice, data_slice = self._split_index(key)
```
jlstevens: Any chance ``_split_index`` could support the empty tuple directly? What happens with an empty tuple currently? It would be nice if this bit could be:

```python
if key is Ellipsis:
    return self
 map_slice, data_slice = self._split_index(key)
```

holoviews/core/spaces.py
```diff
@@ -708,22 +736,22 @@ def __getitem__(self, key):
...
-
-            slices = [el for el in tuple_key if isinstance(el, slice)]
-            if any(el.step for el in slices):
-                raise Exception("Slices cannot have a step argument "
-                                "in DynamicMap bounded mode ")
-            if len(slices) not in [0, len(tuple_key)]:
-                raise Exception("Slices must be used exclusively or not at all")
-            if slices:
-                return  self._slice_bounded(tuple_key)
+            sliced = None
```
jlstevens: Stray line that can be removed?

holoviews/core/spaces.py
philippjfr: ``NdMapping.__getitem__`` does this with empty tuples:

```python
elif indexslice == () and not self.kdims:
     return self.data[()]
```

Won't work here unfortunately, but I guess I could still handle it in ``_split_index``, ``NdMapping`` would just never use that path.

holoviews/core/spaces.py
philippjfr: Yup, thanks.

holoviews/core/dimension.py
jlstevens: Ok.

holoviews/core/spaces.py
```diff
@@ -683,23 +682,52 @@ def _cross_product(self, tuple_key, cache):
...
+                val = self._dataslice(val, data_slice)
             data.append((key, val))
-        return self.clone(data)
+        product = self.clone(data)
+
+        if data_slice:
+            from ..util import Dynamic
+            def dynamic_slice(obj):
+                return obj[data_slice]
+            return Dynamic(product, operation=dynamic_slice, shared_data=True)
```
jlstevens: I forget...is there a good reason why this can't be a lambda?

jlstevens:
> I think HoloMap.__mul__ would have to inline import it, but I think that may be still be preferable, so I'd vote for it.

Ok, go for it!

holoviews/core/spaces.py
philippjfr: No probably could be.

holoviews/core/spaces.py
```diff
@@ -683,23 +682,52 @@ def _cross_product(self, tuple_key, cache):
...
-            cloned.kdims[i].soft_range = (start, stop)
-        return cloned
+            sliced.kdims[i].soft_range = (start, stop)
+        if data_slice:
+            if not isinstance(sliced, DynamicMap):
+                return self._dataslice(sliced, data_slice)
+            else:
+                from ..util import Dynamic
+                def dynamic_slice(obj):
+                    return obj[data_slice]
```
jlstevens: Can also be replaced with a lambda...

jlstevens:
Looks good! 

I've finished making my current set of suggestions for this PR. I do think adding ``core.dynamic`` makes sense to do now as ``core.spaces`` is getting quite long...

philippjfr:
>Ok, go for it!

I was wrong, the circular imports between dynamic.py and operation.py would be pretty bad.

jlstevens:
Shame... any other suggestions for breaking up this file and splitting the dynamic stuff out?

philippjfr:
>Shame... any other suggestions for breaking up this file and splitting the dynamic stuff out?

Would have to think about it. Might open an issue.

jlstevens:
All looking good! Merging.

PR #1030, opened 2016-12-28
Update bokeh ColumnDataSource in one update

philippjfr:
In bokeh 0.12.4 a warning is raised when column lengths do not match therefore all columns on a data source should be updated at once.


jlstevens:
Looks like a simple change. Ready to merge?

philippjfr:
Yes.

PR #1036, opened 2017-01-07
Switched from deprecated bokeh Model.set to Model.update

philippjfr:
Simple API change, bokeh's ``Model.set`` method was a duplicate of the ``Model.update`` method and is now deprecated in >=0.12.4.


jlstevens:
Seems like a straightforward search and replace. If it is tested, I am happy to merge.

philippjfr:
Yes, there's bokeh unit tests and notebook tests covering these changes and the implementation for the two is effectively the same.

PR #1037, opened 2017-01-07
Enabled matplotlib layout fix by default

philippjfr:
It's time to enable the matplotlib layout fix by default. There will be changes to the display data, so that will have to be updated, after careful review.

### DIFF
--- a/holoviews/plotting/mpl/plot.py
+++ b/holoviews/plotting/mpl/plot.py
@@ -698,16 +698,14 @@ class LayoutPlot(GenericLayoutPlot, CompositePlot):
       Specifies the space between horizontally adjacent elements in the grid.
       Default value is set conservatively to avoid overlap of subplots.""")
 
-    vspace = param.Number(default=0.1, doc="""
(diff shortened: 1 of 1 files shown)


jlstevens:
I agree...though how did you come up with the new vspace value? Trial and error?

philippjfr:
>I agree...though how did you come up with the new vspace value? Trial and error?

Yeah, just like the previous value, it should behave more predictably now though. I'm not quite sure why the tests are still passing. Will have to review the test data.

jlstevens:
The tests have now passed. Are the tests updated reviewed making this PR ready to merge?

philippjfr:
No, for some reason the test data hasn't changed and I can't figure out why.

jlstevens:
It might be worth thinking about how this links to #1041 before merging...

jlstevens:
And yes, the fact the tests kept passing is weird!

philippjfr:
I guess just merge it, can always revert it if something goes wrong.

jlstevens:
Ok. Hope it works!

Issue #999, opened 2016-11-30
Document and raise better exceptions about backend loading


relativistic:
When I try to use the %%output cell magic to switch to the bokeh backend, I get the error "Error: list index out of range".

![screen shot 2016-11-30 at 11 10 12 am](https://cloud.githubusercontent.com/assets/371039/20767041/9c180ee2-b6ed-11e6-886b-5f940fb06491.png)

jlstevens:
I believe you just need to load the bokehJS. At the top of the notebook, try:

```
import holoviews as hv
hv.notebook_extension('matplotlib', 'bokeh')
```

Then it should work as long as bokeh is installed.

philippjfr:
Correct, definitely needs a better exception though. There's a few issues like this now that the plotting backends aren't imported by default, so we should catalogue them and make sure to raise better exceptions.

relativistic:
Ah, yes, that fixed it. Thanks!

It might also be a good idea to be explicit this is required "Bokeh Backend" page of the docs, where this is described. I had assumed from the language there that you could use the cell output magic after only loading the matplotlib option in the hv.notebook_extension. Probably obvious to those who are more familiar with the way this kind of thing works, but it wasn't for me.

philippjfr:
Reopening and repurposing this issue.

jlstevens:
Should be easy enough to fix so I'm assigning this as something to fix by 1.7.

philippjfr:
The output magic error message was improved, closing.

PR #1045, opened 2017-01-10
Fixed output magic message when backend unavailable

jlstevens:
This PR addresses #999 regarding the inappropriate error message when trying to load an unavailable backend. Here is the old behavior:

image

And the new behavior:

image

I've tried to make a minimal fix. That said, the output message might suggest how to load the backend with ``notebook_extension``...

### DIFF
--- a/holoviews/ipython/magics.py
+++ b/holoviews/ipython/magics.py
@@ -267,10 +267,6 @@ class OutputMagic(OptionsMagic):
         else:
             raise ValueError("Backend %r does not exist" % value)
 
-
-        raise ValueError(OutputMagic.backend_list, value)
-        # raise ValueError("Backend %r not available. Has it been loaded with the notebook_extension?" % value)
-
     custom_exceptions = {'holomap':missing_dependency_exception,
                          'backend': missing_backend_exception }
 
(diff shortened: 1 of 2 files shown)


jlstevens:
Now with a custom exception message:

image

I think this PR is ready to review/merge. I am happy to improve the exception message if you have other suggestions.

jbednar:
What will happen if you ask for backend='powerpoint', i.e. one that truly doesn't exist?  it would be nice if it were clear whether it's just nonsense/typo, or if there was a missing step.  I understand that might not be possible, of course.

jlstevens:
Good suggestion! I don't think it would be too hard though I don't know if we have a list of all possible backend names somewhere I can use...

holoviews/ipython/magics.py
```diff
@@ -256,11 +256,23 @@ class OutputMagic(OptionsMagic):
...
-    custom_exceptions = {'holomap':missing_dependency_exception}
+    def missing_backend_exception(value, keyword, allowed):
+        if value in OutputMagic.backend_list:
+            raise ValueError("Backend %r not available. Has it been loaded with the notebook_extension?" % value)
+        else:
+            raise ValueError("Backend %r does not exist" % value)
+
+
+        raise ValueError(OutputMagic.backend_list, value)
+        # raise ValueError("Backend %r not available. Has it been loaded with the notebook_extension?" % value)
```
jbednar: Stray comment?

jlstevens:
With the last commit, you have two cases:

image

and

image

I think this PR is now ready for final review/merge.

holoviews/ipython/magics.py
jlstevens: Thanks!

jbednar:
Looks good.

Issue #1021, opened 2016-12-16
archive export labels with slash


vascotenner:
Exporting archives is not possible when Element labels contain '/'. Better escaping is needed.

```python
import holoviews as hv
hv.notebook_extension()
hv.archive.auto()
hv.Curve([1,1], label='./test')
hv.archive.export()
hv.archive.last_export_status()
```

```
The last call to holoviews.archive.export was unsuccessful.

Traceback (most recent call last):
  File "/home/tenner/src/holoviews/holoviews/ipython/archive.py", line 273, in _export_with_html
    info={'notebook':self.notebook_name})
  File "/home/tenner/src/holoviews/holoviews/core/io.py", line 807, in export
    self._directory_archive(export_name, files, root)
  File "/home/tenner/src/holoviews/holoviews/core/io.py", line 747, in _directory_archive
    with open(fpath, 'wb') as f:
IOError: [Errno 2] No such file or directory: '/media/data2/tenner/data/ipython-notebooks/Tests/Test_holoviews_export/Curve-./test.svg'
```

jbednar:
Should also check backslashes, on Windows.

jlstevens:
This has been addressed by the above PR (merged).

PR #1046, opened 2017-01-10
Improved archive escaping of group and label in file/path names

jlstevens:
The PR aims to address issue #1021. I now know where to implement the escaping fix as illustrated in the first commit. We should discuss how to make this prettier/more robust (I don't like the look of the escaped output yet):

image

I suppose that instead of using ``'-sep-'`` I should see if Python offers proper escaping of filenames/paths on Unix/Windows. As I said, this PR is still WIP!

### DIFF
--- a/holoviews/core/io.py
+++ b/holoviews/core/io.py
@@ -31,6 +31,9 @@ from .options import Store
 from .util import unique_iterator, group_sanitizer, label_sanitizer
 
 
+def escape(val, sep=os.sep, replacement='-sep-'):
+    return val.replace(sep, replacement)
+
 class Reference(param.Parameterized):
     """
     A Reference allows access to an object to be deferred until it is
@@ -694,8 +697,8 @@ class FileArchive(Archive):
             hashfn.update(obj_str.encode('utf-8'))
             format_values = {'timestamp': '{timestamp}',
                              'dimensions': dimensions,
-                             'group':   getattr(obj, 'group', 'no-group'),
-                             'label':   getattr(obj, 'label', 'no-label'),
+                             'group':   escape(getattr(obj, 'group', 'no-group')),
+                             'label':   escape(getattr(obj, 'label', 'no-label')),
                              'type':    obj.__class__.__name__,
                              'obj':     obj_str,
                              'SHA':     hashfn.hexdigest()}
@@ -769,7 +772,7 @@ class FileArchive(Archive):
         while (new_name, ext) in existing:
             new_name = basename+'-'+str(counter)
             counter += 1
-        return (new_name, ext)
+        return (escape(new_name), ext)
 
 
     def _truncate_name(self, basename, ext='', tail=10, join='...', maxlen=None):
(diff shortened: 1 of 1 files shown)


jbednar:
Shouldn't the escaping be done using a whitelist (opt in), not a blacklist (opt out)?  I.e., shouldn't it be sanitizing by allowing only certain characters through, rather than specifically removing os.sep?  Maybe I'm confused about the purpose.

jlstevens:
> Shouldn't the escaping be done using a whitelist (opt in), not a blacklist (opt out)

I came to the same conclusion which is why I decided to sanitize in same *style* as the identifier sanitizer but for specific characters that are problematic for filesystems.

Here is the updated example:

image

I think this PR is now ready to review.

jbednar:
Looks fine to me, though I personally would rather have "," instead of "slash" and "backslash".  To each his own.

jlstevens:
If you can suggest a similar replacement for ``:`` and a way to distinguish slash and backslash, I would be happy to change it! Apparently backtick is ASCII character...but I expect it introduces its own problems...

jlstevens:
Perhaps we should not worry about 'bad' filenames, just the specific issue mentioned in #1021 i.e slashes causing problems with paths. Maybe we should just sanitize slash to , and backslash  to \` and leave the filenames alone otherwise - there are a lot of characters that can cause problems in some situations (one site I've seen recommends avoiding all of ``#<$+%>`&*'|{?"=}/:\@!``)

jbednar:
I guess I can't think of any case where one would want to distinguish between slash and backslash, so I'm happy for those to map to the same thing. slash will come up for any mathematical expressions with division, which I think a lot of people will want to plot.  It's hard to see how backslash would be useful in that context, and even harder to see how it would need to be distinguished from slash if it does somehow come up.  So I'd favor being concise, which is very likely to be desirable, over distinguishing between cases I don't think would ever come up.  In the same vein, I'm happy to map ":" to "=", as again I don't think those would need to be distinguished.  But it's up to you; those are just my opinions and gut feelings, and you should go with your own on such ill-defined issues.

jlstevens:
In the end, I've opted to simply replace each of``: ,  \`` and ``/`` with an underscore ``_``.

jlstevens:
The tests have passed so I think this PR is now ready to merge.

holoviews/core/io.py
```diff
@@ -31,6 +31,15 @@
 from .util import unique_iterator, group_sanitizer, label_sanitizer
 
 
+def sanitizer(name, replacements={':':'_', '/':'_', '\\':'_'}):
```
philippjfr: Might be worth sorting the replacements so this is deterministic.

jlstevens: Agreed. I'll use a list of tuples to make sure the order is fixed.

jlstevens:
I think this is now ready to merge once the tests pass. 

Note that I didn't test the last commit directly. It is a simple change and I'm confident it is correct but it never hurts to double check!

philippjfr:
May have to rebase or merge master to get the tests passing.

philippjfr:
The fact that the PR build passed but the push build didn't is confusing me. Seems to indicate that some interaction with master is making this misbehave.

jlstevens:
I've now rebased against latest master. We'll see what happens now!

philippjfr:
Looks good and tests now passing!

Issue #966, opened 2016-11-01
Fix the docs to stop using print(repr(obj))


jlstevens:
Somehow we ended up in the habit of using ``print(repr(obj))`` when ``print obj`` (or ``print(obj)`` for Python 3 compatibility) works just fine.

This is used in a number of our tutorials and it would be good to simplify them to use the sensible version before 1.7.

jlstevens:
I don't think ``print`` works nicely with ``Layouts`` yet:

![image](https://cloud.githubusercontent.com/assets/890576/20150636/cd6aff58-a6ae-11e6-9024-9bec41d0ee06.png)

jlstevens:
Fixed in the PR referenced above (#1048). Closing.

PR #1048, opened 2017-01-10
Aliased __str__ to __repr__

jlstevens:
PR implementing the suggestion in #966. The code change is tiny (aliasing ``__str__`` to ``__repr__``) and most of this PR will be about updating the docs.

Here is the behavior before:

image

And after:

image

Of course I'll make sure to put the parens around print in the notebooks for Python 3 support!

### DIFF
--- a/doc/Tutorials/Columnar_Data.ipynb
+++ b/doc/Tutorials/Columnar_Data.ipynb
@@ -124,9 +124,9 @@
    },
    "outputs": [],
    "source": [
-    "print(repr(hv.Scatter({'x': xs, 'y': ys}) +\n",
-    "           hv.Scatter(np.column_stack([xs, ys])) +\n",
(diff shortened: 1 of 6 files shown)


jlstevens:
I've now updated all the notebooks using ``print(repr(..))`` and I have been careful to make sure the notebook tests kept passing.

Unfortunately, there are two notebooks with fixes that I *do* expect will result in test failures. These involve use of naked ``repr(..)`` which should result in different output  (due to extra quotes compared to ``print``).

jlstevens:
As expected the tests failed. I thought only one cell would fail, but as ``repr`` returns a string whereas ``print`` outputs to standard output, the test data is now shifted. Rebuilding the test data now.

jlstevens:
Now the test data has been updated, the pr build passed and the push build failed (as expected). Ready to review/merge.

Issue #978, opened 2016-11-07
hv.help points to non-existent website


philippjfr:
Example:

```python
hv.help(hv.Curve)
```

>Curve
Online example: https://ioam.github.io/holoviews/Tutorials/Elements.html#Curve
Help for the data object: holoviews.help(Curve) or holoviews.help(<curve_instance>)

jlstevens:
I'm happy to fix this so I've assigned myself to this issue.

philippjfr:
Assigned to 1.7, this is an easy fix and should be implemented asap.

philippjfr:
Now fixed, closing.

PR #1051, opened 2017-01-11
Fixed broken URL in help output

jlstevens:
This PR addresses #978. The code change is trivial but it is likely that the test data will need updating.


jlstevens:
I'm a little surprised that the tests passed...maybe this output is being sent to the pager and therefore isn't captured?

philippjfr:
Yes, I think that's it. One issue I have is that custom Elements will also point to non-existent entries in the elements tutorial. Got any ideas to disable that?

jlstevens:
I agree that is an issue though not a particularly major one. My only suggestion is to suggest the help URL only if the element is one of the ones we offer. Do we have such a list?

Maybe the top level ``__init__.py`` could set a global variable (list of all our elements) in ``core.pprint``. At least this would avoid having to pass this list though help, then Store, then into the pretty printer and all its methods...

philippjfr:
Could probably just iterate over the ``hv.element`` namespace and find all ``Element`` types to get that list.

jlstevens:
I suppose so given that seaborn elements aren't in that tutorial anyway. Should I really do the same for Containers or can I just ignore the possibility of custom Containers? I doubt people will want to build new containers often...

philippjfr:
>Should I really do the same for Containers or can I just ignore the possibility of custom Containers?

Fine to ignore right now I think.

jlstevens:
Added filtering of elements as suggested:

image

Versus a core element:

image

Note this PR isn't quite ready to merge. I might as well fix the issue about showing help for the active backend while I am working on this code.

holoviews/__init__.py
```diff
@@ -25,6 +25,7 @@
 from .operation import ElementOperation, MapOperation, TreeOperation # noqa (API import)
 from .element import *                                               # noqa (API import)
 from . import util # noqa (API import)
+from .element import __all__ as elements_list
```
philippjfr: That doesn't seem right, it should at least filter for objects of type Element.

philippjfr: I know ``hv.element.__all__`` filters for ViewableElement but I think that includes some other types as well.

jlstevens: The problem is that ``__all__`` contains strings (which I want) but doing tighter subclass checks requires a list of those classes. I could iterate over ``locals`` but I find that pretty ugly. What would you suggest?

philippjfr: I'd change line 99 of ``hv.element.__init__`` to:

``return issubclass(obj, Element)``

jlstevens: Done! I was hesitant in case anything was expecting to find ViewableElements...

philippjfr: Appears you were right being hesitant, but it really shouldn't be relying on importing Overlay via ``hv.element``.

jlstevens:
Ok. If the tests pass I think this PR is ready for final review/merge.

philippjfr:
Looks good, merging.

PR #1054, opened 2017-01-12
Fixed python widget key precision to match JS

philippjfr:
So this is pretty strange, as revealed in https://github.com/ioam/holoviews/issues/1052, it turns out the precision the widget keys are set to in Python does not match the precision specified by the widgets. It seems I changed the value as part of a fix for DynamicMap. Will have to investigate what that was about before merging this.

### DIFF
--- a/holoviews/plotting/widgets/__init__.py
+++ b/holoviews/plotting/widgets/__init__.py
@@ -18,7 +18,7 @@ def escape_vals(vals, escape_numerics=True):
     unicode for safety.
     """
     # Ints formatted as floats to disambiguate with counter mode
-    ints, floats = "%.1f", "%.11f"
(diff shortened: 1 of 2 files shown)


philippjfr:
May have to rebuild tests but this is now behaving correctly.

jlstevens:

I hope we will one day be able to test this properly - maybe using selenium? :-)

philippjfr:
Yes, we'll definitely want to add selenium tests at some point. I'd like to add some to test bokeh stream events before releasing 1.7.

jlstevens:
> I'd like to add some to test bokeh stream events before releasing 1.7.

That would be a very good idea. Anyway, tests have passed so I will now merge.

PR #1060, opened 2017-01-14
Add options inheritance on OverlayPlot

philippjfr:
Initial attempt at addressing https://github.com/ioam/holoviews/issues/638, total WIP to see how tests are affected, but it may really be this simple.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1009,7 +1009,6 @@ class BokehMPLRawWrapper(BokehMPLWrapper):
 
 class OverlayPlot(GenericOverlayPlot, LegendPlot):
 
-
     tabs = param.Boolean(default=False, doc="""
         Whether to display overlaid plots in separate panes""")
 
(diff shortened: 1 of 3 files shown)


jlstevens:
Given the tests pass, I think it might be something to enable for 1.7. It is something that we've wanted for a long time and as long as we give the option to restore the old behavior, this might be an easier win than we anticipated!

philippjfr:
Yes, I'm really pleasantly surprised by this, it behaves exactly like I'd expect and after running through a bunch of notebooks my observations are twofold:

* All the notebooks I've run through with this change have not exhibited any change in behavior.
* Almost every notebook I've run through has at least one case where this change would let you simplify the options specification.

However my bandwidth is also limited, so I'd suggest we get as many people playing around with this as possible. As a start, @jlstevens, @jbednar could you both take this PR for a spin on as many notebooks as you can?

philippjfr:
Here's an illustrative example:

```python
%%opts Overlay [xticks=10] Curve [xticks=2 width=600 height=400] Points [show_grid=False bgcolor='WhiteSmoke']
hv.Curve(range(10)) * hv.Curve(np.random.rand(10)) * hv.Points(np.random.rand(100,2)*9)
```

The effective applied options for this example are: ``{'xticks': 10, 'width': 600, 'bgcolor': 'WhiteSmoke', 'show_grid': False, 'height': 400}``

bokeh_plot 30

jlstevens:
It is a fairly small PR so it should be easy to revert if we suddenly find major issues (which seems doubtful). 

For this reason, I think the best way to get it well tested is simply to merge it ASAP. It might still be worth adding a class attribute toggle first though...

jlstevens:
The tests have passed so I will merge now. I'm hopeful this will address one of the biggest annoyances without causing much trouble!

jbednar:
Fabulous!  I think that this will go a long way towards making it intuitive to set and use options.

PR #1062, opened 2017-01-14
Fixed HeatMap shape setting

philippjfr:
The HeatMap shape is set incorrectly atm in matplotlib, causing non-square HeatMaps to be rendered incorrectly.


jlstevens:
Looks good. Merging.

Issue #1049, opened 2017-01-10
Stream events on overlaid DynamicMaps should selectively update layers that reference the stream


jordansamuels:
Based on the recommendation from @philippjfr , I am using a `holoviews.stream.PositionXY`  to overlay `hv.Text` onto a `DynamicMap`.   This works well in theory; but with wide plots with several layers, the feedback is very slow.  Here is an example:

```python
import holoviews as hv
import numpy as np
import pandas as pd
from holoviews.operation.datashader import datashade, aggregate, shade
from holoviews.streams import PositionXY

hv.notebook_extension('bokeh')

%opts DynamicMap [width=1800, height=450, tools=['xwheel_zoom']] Layout [shared_datasource=False]

n=100000
def f(n):
    return np.cumsum(np.random.randn(n))
df = pd.DataFrame({'t': range(n), 'y1': f(n), 'y2': f(n), 'y3' : .10 * f(n)} )

def p(y):
    return hv.Curve(df, kdims=['t'], vdims=[y])

ts = datashade(p('y1'))
ch = hv.DynamicMap(lambda x, y: hv.HLine(y) * hv.VLine(x) * hv.Text(x+0.15, y, '%.3f' % x + ", %.3f" % y),
              kdims=[], streams=[PositionXY()], source=ts)
%opts RGB [width=1800, height=450]
(ts * datashade(p('y2')) * ch + ts).cols(1)
```

I'm using bokeh 0.12.3 and holoviews latest code as of 2425fadfd80b10e65b3c7f87cb00a102932e6dce .

philippjfr:
Thanks @jordansamuels, this highlights an important issue with our current implementation. What is happening here is that ``ts * datashade(p('y2')) * ch`` is combined into one DynamicMap, which calls each of the individual layers. This means the ``datashade`` DynamicMap gets called every time a ``PositionXY`` event is sent. This will massively slow down what should be a very quick plot update. What should happen instead is that the outer DynamicMap only calls the individual layers if they are also attached to that ``PositionXY`` event. I'll have to think about how best to implement this, but it's an important issue that needs to be addressed before release. @jlstevens Any thoughts?

jlstevens:
I agree this raises an important issue. I'll have to think a bit more about before suggesting anything though...

jlstevens:
I've thought a bit about this a little and I think there are the following approaches:

[...]

PR #1063, opened 2017-01-15
Add memoization to dynamic Callable class

philippjfr:
In order to address #1049 the Callable class now memoizes the last returned value from a DynamicMap, based on the args, kwargs and nested stream values.

Needs docstrings and tests. @jordansamuels this seems to address your issue. Small example for testing:

```python
import holoviews as hv
import numpy as np
import pandas as pd
from holoviews.streams import PositionXY

%opts RGB [width=800, height=450] VLine HLine (line_color='black' line_width=0.5)

n=100000
def f(n):
    return np.cumsum(np.random.randn(n))
df = pd.DataFrame({'t': range(n), 'y1': f(n), 'y2': f(n), 'y3' : .10 * f(n)} )

def p(y):
    return hv.Curve(df, kdims=['t'], vdims=[y])

ts = datashade(p('y1'))
ch = hv.DynamicMap(lambda x, y: hv.HLine(y) * hv.VLine(x) * hv.Text(x+0.15, y, '%.3f' % x + ", %.3f" % y),
              kdims=[], streams=[PositionXY()])
(ts * datashade(p('y2')) * ch + ts).cols(1)
```

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -530,7 +530,6 @@ class DynamicMap(HoloMap):
 
         self.call_mode = self._validate_mode()
         self.mode = 'bounded' if self.call_mode == 'key' else 'open'
-        self._dimensionless_cache = False
 
 
     def _initial_key(self):
@@ -778,7 +777,7 @@ class DynamicMap(HoloMap):
         try:
             dimensionless = util.dimensionless_contents(get_nested_streams(self),
                                                         self.kdims, no_duplicates=False)
-            if (dimensionless and not self._dimensionless_cache):
+            if dimensionless:
                 raise KeyError('Using dimensionless streams disables DynamicMap cache')
             cache = super(DynamicMap,self).__getitem__(key)
             # Return selected cache items in a new DynamicMap
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -406,8 +406,8 @@ class Callable(param.Parameterized):
     makes it possible to traverse the graph of operations applied
     to a DynamicMap. Additionally a Callable will memoize the last
     returned value based on the arguments to the function and the
-    state of all streams on its inputs, avoiding calling the function
-    unncessarily.
+    state of all streams on its inputs, to avoid calling the function
+    unnecessarily.
     """
 
     callable_function = param.Callable(default=lambda x: x, doc="""
@@ -421,8 +421,8 @@ class Callable(param.Parameterized):
         self._memoized = {}
 
     def __call__(self, *args, **kwargs):
-        inputs = [inp for inp in self.inputs if isinstance(inp, DynamicMap)]
-        streams = [s for inp in inputs for s in get_nested_streams(inp)]
+        inputs = [i for i in self.inputs if isinstance(i, DynamicMap)]
+        streams = [s for i in inputs for s in get_nested_streams(i)]
         values = tuple(tuple(sorted(s.contents.items())) for s in streams)
         key = args + tuple(sorted(kwargs.items())) + values
 
@@ -517,7 +517,7 @@ class DynamicMap(HoloMap):
        """)
 
     def __init__(self, callback, initial_items=None, **params):
-        if not isinstance(callback, Callable) and not isinstance(callback, types.GeneratorType):
+        if not isinstance(callback, (Callable, types.GeneratorType)):
             callback = Callable(callable_function=callback)
         super(DynamicMap, self).__init__(initial_items, callback=callback, **params)
 
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -404,7 +404,10 @@ class Callable(param.Parameterized):
     allowing their inputs (and in future outputs) to be defined.
     This makes it possible to wrap DynamicMaps with streams and
     makes it possible to traverse the graph of operations applied
-    to a DynamicMap.
+    to a DynamicMap. Additionally a Callable will memoize the last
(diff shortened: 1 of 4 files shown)


jbednar:
Wow!  And the total amount of code even went down.  Fabulous!

philippjfr:
@jlstevens Requested review, the code is fairly straightforward and I've added at least one test to ensure the memoization behaves correctly. There are already a number of tests to check memoization works during plotting stopping the callback from being called repeatedly.

holoviews/core/spaces.py
```diff
@@ -404,7 +404,10 @@ class Callable(param.Parameterized):
     allowing their inputs (and in future outputs) to be defined.
     This makes it possible to wrap DynamicMaps with streams and
     makes it possible to traverse the graph of operations applied
-    to a DynamicMap.
+    to a DynamicMap. Additionally a Callable will memoize the last
+    returned value based on the arguments to the function and the
+    state of all streams on its inputs, avoiding calling the function
```
jlstevens: I think either 'avoiding calls to the function' or 'to avoid calling the function ...' would be read better.

holoviews/core/spaces.py
```diff
@@ -404,7 +404,10 @@ class Callable(param.Parameterized):
...
+    unncessarily.
     """
```
jlstevens: Typo.

holoviews/core/traversal.py
```diff
@@ -128,29 +128,3 @@ def hierarchical(keys):
                 store1[v2].append(v1)
         hierarchies.append(store2 if hierarchy else {})
     return hierarchies
-
-
-class dimensionless_cache(object):
-    """
```
jlstevens: Glad to get rid of this!

holoviews/core/spaces.py
```diff
@@ -500,6 +517,8 @@ class DynamicMap(HoloMap):
        """)
 
     def __init__(self, callback, initial_items=None, **params):
+        if not isinstance(callback, Callable) and not isinstance(callback, types.GeneratorType):
+            callback = Callable(callable_function=callback)
```
jlstevens: This means all callbacks will now be wrapped in ``Callable``. This is a change from the previous behavior, and probably a good one (more consistent, at least with ``__mull__``) though we should now document this. It would also be good to have an example of where a user might want to supply ``Callable`` themselves....

holoviews/core/spaces.py
```diff
@@ -413,8 +416,22 @@ class Callable(param.Parameterized):
     inputs = param.List(default=[], doc="""
          The list of inputs the callable function is wrapping.""")
 
+    def __init__(self, **params):
+        super(Callable, self).__init__(**params)
+        self._memoized = {}
+
     def __call__(self, *args, **kwargs):
-        return self.callable_function(*args, **kwargs)
+        inputs = [inp for inp in self.inputs if isinstance(inp, DynamicMap)]
```
jlstevens: Very minor point - I would call the loop variable ``i`` given that ``inp`` is a bit weird and you probably don't want to clash with the ``input`` built-in. Alternatively, this is one instance of a pure ``filter`` so you could consider using that (it would be shorter).

holoviews/core/spaces.py
```diff
@@ -413,8 +416,22 @@ class Callable(param.Parameterized):
...
+        streams = [s for inp in inputs for s in get_nested_streams(inp)]
+        values = tuple(tuple(sorted(s.contents.items())) for s in streams)
```
jlstevens: This means it is memoizing on the stream parameters in ascending alphanumeric order. I think that is fine (and makes sense!) but it is worth noting that this is a particular policy for how streams parameters are ordered into a tuple key. This is important to stay consistent if we ever need it somewhere else and I am now wondering if it might be worth having a utility to codify the idea....up to you.

philippjfr: It really doesn't matter much, it just needs to be consistently ordered. If we're using it somewhere else it can use the same scheme or another scheme without having any effect here.

jlstevens:
Looks good! 

I'm very happy this turned out to be far more straightforward than I anticipated. Overall only seven lines were added and the new approach is far more intuitive to me than ``dimensionless_cache`` ever was.

I think my only worry is that is that, from what I understand, ``Callback`` will memoise at every level so ``dmap1 * dmap2 * dmap3`` is memoising 3 + 2 + 1 i.e 6 times (3 leaves, two binary muls, one triple mul). The main worry then is unnecessary/excessive memory usage.

holoviews/core/spaces.py
jlstevens: True - my point is that we should try to be consistent anyway!

philippjfr:
>I think my only worry is that is that, from what I understand, Callback will memoise at every level so dmap1 * dmap2 * dmap3 is memoising 3 + 2 + 1 i.e 6 times (3 leaves, two binary muls, one triple mul). The main worry then is unnecessary/excessive memory usage.

That's accurate although each level already caches these values in the DynamicMap cache so I don't think there will be much being cached additionally. It's also worth noting that each level simply wraps the previous level, so it's not holding copies of the data, so the overhead should simply be in the parameterized objects that are created, which should be fairly cheap.

holoviews/core/spaces.py
jlstevens: For instance, we might want an example of the user declaring a ``Callable`` with their own set of ``inputs``.

jlstevens:
Tests are passing so I will merge now.

PR #1065, opened 2017-01-16
Deprecate DFrame conversion methods

philippjfr:
While I'm not ready to drop these interfaces entirely (I use some of the types like ``Distribution`` and ``Regression`` frequently),  the conversion methods on the pandas/seaborn interface types have long been marked with a deprecation warning so I think we can safely remove them for v1.7.


philippjfr:
Ready to merge as soon as tests pass, I am assuming these are untested but may be wrong.

philippjfr:
PR build failure was the a transient failure, ready to merge whenever. Also yay, first time we broke the 77% coverage mark I believe.

jlstevens:
Looks good. I  brought back the [changelog wiki page](https://github.com/ioam/holoviews/wiki/Changelog) so we don't forget to mention this on release.

Merging.

PR #1066, opened 2017-01-16
Renamed conversion interface mdims argument to groupby

philippjfr:
As suggested in https://github.com/ioam/holoviews/issues/1061 renaming mdims to groupby in the conversion interface ``.to`` method is a lot clearer. This PR renames the kwarg while retaining support for the mdims argument along with a deprecation warning.

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -68,25 +68,37 @@ class DataConversion(object):
     def __init__(self, element):
         self._element = element
 
-    def __call__(self, new_type, kdims=None, vdims=None, mdims=None,
+    def __call__(self, new_type, kdims=None, vdims=None, groupby=None,
                  sort=False, **kwargs):
         """
         Generic conversion method for Column based types. Supply the
-        Columns based type to convert to and optionally the
-        key dimensions (kdims), value dimensions (vdims) and HoloMap
-        key dimensions (mdims). Converted Columns can be automatically
-        sorted via the sort option and kwargs can be passed through.
+        Columns based type to convert to and optionally the key
+        dimensions (kdims), value dimensions (vdims) and the
+        dimensions.  to group over. Converted Columns can be
+        automatically sorted via the sort option and kwargs can be
+        passed through.
         """
+        if 'mdims' in kwargs:
+            if groupby:
+                raise ValueError('Cannot supply both mdims and groupby')
+            else:
+                self._element.warning("mdims keyword has been renamed to "
+                                      "groupby and will be deprecated in "
+                                      "version 1.8.")
+                groupby = kwargs['mdims']
+
         if kdims is None:
             kdims = self._element.kdims
         elif kdims and not isinstance(kdims, list): kdims = [kdims]
         if vdims is None:
             vdims = self._element.vdims
         if vdims and not isinstance(vdims, list): vdims = [vdims]
-        if mdims is None:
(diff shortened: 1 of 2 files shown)


holoviews/core/data/__init__.py
```diff
@@ -68,25 +68,37 @@ class DataConversion(object):
...
-        """
+        Columns based type to convert to and optionally the key
+        dimensions (kdims), value dimensions (vdims) and the
+        dimensions.  to group over. Converted Columns can be
+        automatically sorted via the sort option and kwargs can be
+        passed through.
+        """
+        if 'mdims' in kwargs:
+            if groupby:
+                raise ValuError('Cannot supply both mdims and groupby')
```
jbednar: ValueError?

philippjfr: Thanks.

holoviews/core/data/__init__.py
```diff
@@ -68,25 +68,37 @@ class DataConversion(object):
...
+        can be automatically sorted via the sort option and kwargs can
+        bepassed through.
         """
+        if 'mdims' in kwargs:
+            if groupby:
+                raise ValueError('Cannot supply both mdims and groupby')
+            else:
+                self._element.warning("mdims keyword has been renamed to "
+                                      "groupby and will be deprecated in "
+                                      "version 1.8.")
```
jbednar: "will be deprecated" is ambiguous; isn't it now deprecated already?  Plus it seems like we shouldn't promise a version 1.8 we might not have (in favor of 2.0). How about:

``mdims keyword has been renamed to groupby; the name mdims is deprecated and will be removed after version 1.7.``

philippjfr: Agreed.

jbednar:
Looks great; I think that will be much clearer.  Also seems to add support for passing a single dimension, not just a list of dimensions.

philippjfr:
>Looks great; I think that will be much clearer. Also seems to add support for passing a single dimension, not just a list of dimensions.

Yes, should have listed that explicitly, both the kdims and vdims arguments support this so it's been annoying me persistently that you had to pass a list here.

philippjfr:
Passing and ready to merge.

holoviews/core/data/__init__.py
```diff
@@ -68,25 +68,38 @@ class DataConversion(object):
...
+        bepassed through.
         """
+        if 'mdims' in kwargs:
+            if groupby:
+                raise ValueError('Cannot supply both mdims and groupby')
+            else:
+                self._element.warning("mdims keyword has been renamed "
+                                      "to groupby; the name mdims is "
+                                      "deprecated and will be removed "
+                                      "after version 1.7.")
```
jlstevens: I never know whether it should be ``'mdims' keyword ...`` or ``mdims keyword...``. As we are referring to the name of the keyword, I think I would include the quotes. Anyway, very minor issue and I don't mind if you prefer to leave it as it is.

jlstevens:
Made one minor comment about the deprecation message but otherwise, I am happy to merge now.

holoviews/core/data/__init__.py
jbednar: I agree the quote marks (or even backquote marks) would be the correct way, but figured we were being expedient.

philippjfr:
Comment addressed, ready to merge when tests pass.

jlstevens:
Looks good. Merging!

PR #1068, opened 2017-01-17
Fixed setting of matplotlib axis labels

philippjfr:
As part of changing the options inheritance I made a small change to matplotlib axes settings which caused a number of issues. This reverts that change.


philippjfr:
Okay, I'm highly confused by this. This change definitely changes the test data, yet everything is still passing.

philippjfr:
Not quite accurate this change actually makes the reference data correct again, so it should be passing. However master should definitely be failing right now.

jlstevens:
I'll merge now and hope that helps us figure out what is going on with testing...

PR #1069, opened 2017-01-17
Fixes for axis options handling in bokeh

philippjfr:
Two small bugs here, the ``top-bare`` and ``right-bare`` options for xaxis and yaxis respectively were being ignored and secondly elements with only one dimension like Spikes do not correctly processed their axis settings.


jlstevens:
Looks good. Merging.

Issue #1067, opened 2017-01-17
Dynamic Callback memoization breaks for non-hashable stream values


philippjfr:
In #1063 we added memoization for stream values by generating a key from the callback args, kwargs and all stream values. This is fragile because certain types like lists cannot be hashed. We can either cast to hashable types or simply use the string repr of the key.

jlstevens:
I'm not super comfortable using ``repr`` for this purpose although it *should* work. Probably the best thing is to implement the following behavior:

1. Assume the key is hashable and try to use it.
2. Fallback to the repr because something in the key is unhashable.
3. Skip memoisation (maybe the repr can fail).

In no circumstance should there be an error - the worst case should be that you lose the advantage of memoization because the repr of the unhashable object is unstable.

The other solution is to limit the types of allowed stream parameter - but that is also tricky. If you specify a ``param.List``, the elements of the list could be anything. It probably isn't worth restricting what streams can do for the sake of a performance optimization.

The only other thing we could do is warn when falling back to ``repr``. The idea would be to recommend that the user uses hashable stream values (e.g tuples instead of lists). Then again, if you want to use dictionaries as the stream value, there is no easy hashable alternative (a tuple of tuples?). If only Python had a [frozendict](https://www.python.org/dev/peps/pep-0416/)....

philippjfr:
I'm not a huge fan of having it fall back to a different behavior. I do think it should a single approach, although I do recognize your worries about stability of the repr so I'm wondering whether we should just traverse the key and replace unhashable objects:

* list -> tuple
* dict -> tuple(sorted(dict.items()))
* set -> frozenset

What else would be left? I believe most other types should hash correctly or am I wrong?

jbednar:
I would think that solid handling for lists and dicts is very important, and I don't see any reason why converting them to tuples or tuples of tuples, respectively, would be an issue, apart from needing to do so recursively.

jlstevens:
We would at least want to support ``OrderedDict`` as well as dict. I prefer the idea of recursive replacement with hashables but I am still wondering whether our suggested replacements cover everything. Do we ever use defaultdicts with ``param.Dict``?

philippjfr:
Both OrderedDict and defaultdict are dict subclasses so are covered by dict.

jlstevens:
I wouldn't have bet that cyordereddict also being a subclass of dict (being cython and all) but thankfully it is.

jlstevens:
[...]

PR #1074, opened 2017-01-18
Generalized hashing of keys for memoization

jlstevens:
Addresses #1067. 

This approach isn't perfect but it will allow a lot more keys to be memoized than before (sets, lists, dictionaries). The main limitation is that composite values as dictionary keys are not allowed by the JSON spec.

Currently pandas Dataframes/Series and numpy arrays are allowed as mutables. In general you will want to support mutable types explicitly as using the ``id`` does not reflect the contents of the object.

This PR isn't ready yet - I want to add a bunch of unit tests.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -32,17 +32,22 @@ except ImportError:
 class HashGenerator(json.JSONEncoder):
     """
     Extends JSONEncoder to generate a hashable string for as many types
-    of key as possible. These keys are supplied by widgets and Stream
-    parameters and may be nested. If an unrecognized object type is
-    supplied, the objects id is used - this may be an issue when dealing
-    with mutable objects. For this reason, it is important to support
-    common mutable types such as pandas Dataframes and numpy arrays
-    (supported).
+    of object as possible including nested objects and objects that are
+    not normally hashable. The purpose of this class is to generate
+    unique strings that once hashed are suitable for memoization.
+
+    By default JSONEncoder supports booleans, numbers, strings, lists,
+    tuples and dictionaries. In order to support other types such as
+    sets, datetime objects and mutable objects such as pandas Dataframes
+    or numpy arrays, HashGenerator has to convert these types to
+    datastructures that can normally be represented as JSON.
+
+    Support for other object types may need to be introduced in
+    future. By default, unrecognized object types are represented by
+    their id.
 
     One limitation of this approach is that dictionaries with composite
     keys (e.g tuples) are not supported due to the JSON spec.
-
-    Used by keyhash to help memoize DynamicMap output.
     """
     string_hashable = (dt.datetime,)
     repr_hashable = ()
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -70,13 +70,13 @@ class HashableJSON(json.JSONEncoder):
 
 
 
-def deephash(key):
+def deephash(obj):
     """
-    Given a key, return a hash using HashableJSON. This hash is not
+    Given an object, return a hash using HashableJSON. This hash is not
     architecture, Python version or platform independent.
     """
     try:
-        json_str = json.dumps(key, cls=HashableJSON, sort_keys=True)
+        json_str = json.dumps(obj, cls=HashableJSON, sort_keys=True)
         return hash(json_str)
     except:
         return None
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -76,8 +76,7 @@ def deephash(obj):
     architecture, Python version or platform independent.
     """
     try:
-        json_str = json.dumps(obj, cls=HashableJSON, sort_keys=True)
-        return hash(json_str)
+        return hash(json.dumps(obj, cls=HashableJSON, sort_keys=True))
     except:
         return None
 
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -60,7 +60,7 @@ class HashableJSON(json.JSONEncoder):
         elif isinstance(obj, np.ndarray):
             return obj.tolist()
         if pd and isinstance(obj, (pd.Series, pd.DataFrame)):
-            return sorted(list(obj.to_dict().items()))
+            return repr(sorted(list(obj.to_dict().items())))
         elif isinstance(obj, self.string_hashable):
             return str(obj)
         elif isinstance(obj, self.repr_hashable):
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -70,7 +70,7 @@ class HashableJSON(json.JSONEncoder):
 
 
 
-def keyhash(key, as_string=False):
+def deephash(key, as_string=False):
     """
     Given a key, return a hash using HashableJSON. This hash is not
     architecture, Python version or platform independent.
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -34,7 +34,9 @@ class HashableJSON(json.JSONEncoder):
     Extends JSONEncoder to generate a hashable string for as many types
     of object as possible including nested objects and objects that are
     not normally hashable. The purpose of this class is to generate
-    unique strings that once hashed are suitable for memoization.
+    unique strings that once hashed are suitable for use in memoization
+    and other cases where deep equality must be tested without storing
+    the entire object.
 
     By default JSONEncoder supports booleans, numbers, strings, lists,
     tuples and dictionaries. In order to support other types such as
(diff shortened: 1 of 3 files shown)


jlstevens:
Just some quick examples/performance statistics:

```
%%timeit
   ...: keyhash([datetime.datetime(1,2,3), set([1,2,3]), pd.DataFrame({'a':[1,2],'b':[3,4]}), np.array([1,2,3]), np.int64(34)])
   ...:
1000 loops, best of 3: 659 µs per loop
```

```
%%timeit
   ...: keyhash([1,2,3, (4,5)])
   ...:
10000 loops, best of 3: 23.5 µs per loop
```

Unsurprisingly, big keys hash slower than small keys (I expect most keys to be pretty small). Note that ``datetime`` objects are also supported.

philippjfr:
Looks good to me. I doubt we'll catch all the cases we'll need straight away so it's good you can add ``repr_hashable`` and ``str_hashable`` types. Happy to merge when you've added some unit tests.

holoviews/core/util.py
```diff
@@ -24,6 +26,57 @@
...
+
+class HashGenerator(json.JSONEncoder):
+    """
+    Extends JSONEncoder to generate a hashable string for as many types
+    of key as possible. These keys are supplied by widgets and Stream
+    parameters and may be nested. If an unrecognized object type is
+    supplied, the objects id is used - this may be an issue when dealing
+    with mutable objects. For this reason, it is important to support
+    common mutable types such as pandas Dataframes and numpy arrays
+    (supported).
```
jbednar: The bit discussing mutable objects is confusing -- it calls out Pandas and Numpy, but doesn't mention dicts, lists, sets, etc. -- surely those are both mutable and supported as well?

holoviews/core/util.py
```diff
@@ -24,6 +26,57 @@
...
     dd = None
 
+
+
+
+class HashGenerator(json.JSONEncoder):
+    """
+    Extends JSONEncoder to generate a hashable string for as many types
+    of key as possible. These keys are supplied by widgets and Stream
+    parameters and may be nested. If an unrecognized object type is
```
jbednar: I don't think the docstring for this class should mention widgets or Streams in any way other than something explicitly labeled as an example.  It could mention memoization, though, so that we know the goal of this object is to generate an id that can tell whether two objects differ, not to provide an encoding of the object.  With that in mind, this class seems misnamed -- I don't see anything about hashing in it.  Isn't it a TextualRepresentationGenerator (if you don't want to play up the JSON aspects)?

jbednar: And I'm confused about whether it's truly a HashGenerator -- the docstring says it will generate a hashable string, not a hash, and the code indicates that sometimes it's one, sometimes it's the other.  Which is it?

holoviews/core/util.py
jlstevens: The distinction is between what is supported directly by JSON (dicts, lists) and things that we need to add support for it ``HashGenerator``. Currently support is added for sets, datetime, numpy arrays and pandas dataframes/series. More things may need to be explicitly supported in future.

holoviews/core/util.py
jlstevens: ``HashableStringGenerator`` is a pretty ugly name but probably more accurate.

jlstevens: > I don't think the docstring for this class should mention widgets or Streams in any way other than something explicitly labeled as an example. It could mention memoization, though, so that we know the goal of this object is to generate an id that can tell whether two objects differ, not to provide an encoding of the object.

Happy to update the docstring as suggested.

jbednar: Please do!  Thanks.

holoviews/core/util.py
jbednar: Ok, please reword to convey that.

holoviews/core/util.py
jbednar: HashableString?

holoviews/core/util.py
```diff
@@ -24,6 +26,57 @@
...
+        elif isinstance(obj, self.repr_hashables):
+            return repr(obj)
+        try:
+            return hash(obj)
+        except:
+            return id(obj)
+
+
+
+def keyhash(key, as_string=False):
```
jbednar: Or is this hashable_string?

In any case, I'm not sure in what sense the argument is a "key".

holoviews/core/util.py
jlstevens: Hmm, all strings are hashable already. It is about taking an arbitrary thing and generating a string from it. I agree that is a less ugly name though... @philippjfr Any suggestions?

holoviews/core/util.py
jlstevens: > Or is this ``hashable_string``?

No. I'll get rid of the ``as_string`` argument as I only needed it temporarily for debugging. The output is an integer, as returned by the Python ``hash`` function.

I am happy to rename it. In the context it is used (memoisation) the argument is what we have always called a 'key'. I'll just call it ``obj`` here... but then this function then needs a new name as ``hash`` clashes with the built-in. Again, any suggestions?

holoviews/core/util.py
jbednar: StringForHashing?  Not great.

jlstevens: ``HashableJSON`` might be better - the output is a JSON string with some customised support for additional types.

jlstevens: ``HashableJSONEncoder`` if we want to be pedantic...

holoviews/core/util.py
jbednar: Sure, you'll use it as a key, but that doesn't mean the function has anything to do with keys.  obj is a good name for the argument; that's what it is (an arbitrary object).  For the function name, maybe objhash (to contrast it with the regular hash(), which doesn't work for arbitrary objects)?

holoviews/core/util.py
jbednar: HashableJSON seems fine to me.

holoviews/core/util.py
jlstevens: > Sure, you'll use it as a key, but that doesn't mean the function has anything to do with keys.

Agreed! Not that keen on ``objhash`` as a name though...

jlstevens:
Implemented suggestions and added unit tests. Ready for final review/merge once tests pass.

holoviews/core/util.py
```diff
@@ -29,20 +29,25 @@
...
     Extends JSONEncoder to generate a hashable string for as many types
-    of key as possible. These keys are supplied by widgets and Stream
-    parameters and may be nested. If an unrecognized object type is
-    supplied, the objects id is used - this may be an issue when dealing
-    with mutable objects. For this reason, it is important to support
-    common mutable types such as pandas Dataframes and numpy arrays
-    (supported).
+    of object as possible including nested objects and objects that are
+    not normally hashable. The purpose of this class is to generate
+    unique strings that once hashed are suitable for memoization.
```
jbednar: Strictly speaking, it's not the hashed strings that are memoized, it's the actual results (and only implicitly even those, in this case). The hashed strings are *useful* for memoization, but they themselves are not memoized.

So maybe "The purpose of this class is to generate unique strings that once hashed are suitable for use in memoization and other cases where deep equality must be tested without storing the entire object".

holoviews/core/util.py
```diff
@@ -29,20 +29,25 @@
...
+
+    By default JSONEncoder supports booleans, numbers, strings, lists,
+    tuples and dictionaries. In order to support other types such as
+    sets, datetime objects and mutable objects such as pandas Dataframes
```
jbednar: and custom mutable

jbednar:
Not sure why it's failing, but it looks good to me apart from the minor docstring fixes suggested.

jlstevens:
Python 3 has an additional limitation - the ``sort_keys`` option in ``json.dumps`` fails if it encounters dictionaries with heterogenous keys. This is a very clear bug as the final JSON keys are always strings which *can* be sorted but unfortunately this issue is [unlikely to be fixed](https://bugs.python.org/issue25457).

If we are happy to accept these limitations, the PR is ready to merge as the test have passed. Even with its caveats/holes, this approach greatly expands what can be memoized (before a simple list wouldn't allow memoization).

jbednar:
Looks good to me.

philippjfr:
Annoying this turned out to be such a pain. We'll just have to see if we run into any issues in practice. Currently the types of objects that will appear in a stream are fairly limited so I don't foresee any immediate issues. Will merge.

Issue #843, opened 2016-09-01
Dimension preferences


jlstevens:
Since the very beginning of the project, there has been some tension as to what `Dimension` objects are for. There have been two major pressures:
- Keeping dimension objects as lightweight as possible, used to describe key features _about_ the data.
- Using dimension objects to implement features and improve usability. For instance, we've wanted a way to set a default position for sliders and `Dimension` is the obvious place to put that. Other things related to usability are `soft_range` and the formatters.

This situation has made things difficult as `Dimension` is a core component of HoloViews that we want to keep as simple as possible on one hand while also wanting to support new features on the other. Things are further complicated by questions as to whether two dimension objects are 'equal' and therefore interchangeable.

After discussing this problem with @philippjfr and @jbednar, I think we've found a good compromise based on this insight: some of what `Dimension` contains is indeed core metadata _about_ the data and the rest of the options are aimed at usability that are issues of  _preference_ (which can be subjective).

The core metadata should be a small list of things that define the most important information about the dimension and the preferences can be a much larger, open-ended set of options aimed at usability. For this reason, we should group these two types of metadata separately.
- Core metadata: `name`, `unit`, `range` (often constrained by the unit definition, e.g (0, 360) for angle in degrees and (0, None) for a temperature in Kelvin). 
- Preferences: `soft_range` (the typical _useful_ range as determined by the domain or user preference), `default_key` and the various formatters.

Based on this, the proposal is that `Dimension` has only the core metadata as its own parameters as well as a `preference` parameter that holds an object containing all the preferences. This makes it clear as to what the most important information is regarding a dimension while allowing flexibility in terms of the various preferences.

We still need to discuss the exact form this preference object takes and we should think of finding a syntactically convenient way of defining them.

jbednar:
Sounds good!  Deciding what to do with existing preferences like soft_range is also  important, as a a balance between compatibility and clean semantics.

jlstevens:
Here is one syntax for this I think might be quite nice:

```
K = hv.Dimension('Kelvin', range=(0,None)).prefs(soft_range=(200,300), default=220)
```

jbednar:
Would the prefs object accept any Parameterized, or only the one sample one that we define (with soft_range, default, etc. already enumerated as Parameters)?

jlstevens:
Of course, we also need to be careful around the issue of compatibility. Here is what we could do:

[...]

PR #1083, opened 2017-01-26
Add Dimension.label replacing global alias state

philippjfr:
Addresses #1001 and supersedes #1075.

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -475,7 +475,7 @@ class Dataset(Element):
         kdims = replace_dimensions(self.kdims, dimensions)
         vdims = replace_dimensions(self.vdims, dimensions)
         zipped_dims = zip(self.kdims+self.vdims, kdims+vdims)
-        renames = {pk.alias: nk for pk, nk in zipped_dims if pk != nk}
+        renames = {pk.key: nk for pk, nk in zipped_dims if pk != nk}
         data = self.data
         if renames:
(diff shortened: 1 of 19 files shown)


philippjfr:
This is now passing without any major issues. This needs a lot more unit tests still and @jlstevens promised to write a short tutorial but we're both very happy with how this will work out.

jbednar:
Looks good to me.  Some examples of usage would be good.

philippjfr:
Yes, those should be part of the tutorial. I'd actually like to suggest to broaden the scope of that tutorial to discuss Dimensions in general, with a significant section devoted to ``name`` and ``label`` but also covering the ``redim`` method and the difference between ``range`` and ``soft_range``.

Another thing I've done in this PR is using ``strict`` dimension lookups in user facing APIs, which should result in much better tracebacks when referencing non-existing dimensions in many cases.

jlstevens:
Thanks @philippjfr!

I'll shortly write up a notebook tutorial describing aliases, ``redim`` and the distinction between``range`` ``soft_range`` and add it to this PR.

jlstevens:
I know have part of a Dimensions tutorial but I would like to work on #843 before continuing work on it. I am happy with this PR and would like to merge it now. Are you ok with me adding the Dimensions notebook to a different PR?

philippjfr:
Sure, let's see how it goes. Better to get it in early to find issues.

PR #1086, opened 2017-01-27
Fix for datetime support on bokeh BoxPlot

philippjfr:
Bokeh does not currently handle datetimes on their boxwhisker chart implementation, in the meantime this PR simply casts the datetimes to strings, replacing any colons with semi-colons because those are also not supported. I'll file an issue about it on the bokeh repo, but since the charts API may get deprecated it may not get addressed.

### DIFF
--- a/holoviews/core/__init__.py
+++ b/holoviews/core/__init__.py
@@ -25,6 +25,11 @@ Dimension.type_formatters[np.float32] = "%.5g"
 Dimension.type_formatters[np.float64] = "%.5g"
 Dimension.type_formatters[np.datetime64] = '%Y-%m-%d %H:%M:%S'
 
+try:
+    import pandas as pd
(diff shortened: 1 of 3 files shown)


jlstevens:
I would like to hear what the Bokeh devs have to say before merging - this issue of not supporting colons is very weird!

philippjfr:
Here's the issue on it: https://github.com/bokeh/bokeh/issues/516, currently marked 0.12.5. In the meantime I need this for a project, so I'd suggest we merge and I'll open an issue to follow up on, which can hopefully be merged before 1.7 is released.

jlstevens:
> ... so I'd suggest we merge and I'll open an issue to follow up on, which can hopefully be merged before 1.7 is released.

Sounds good.

PR #1087, opened 2017-01-28
SideHistogram supports coloring weighted average

philippjfr:
The adjoined histograms support coloring by a dimension on the main plot, but they also used to support coloring weighted histograms, which take the average in bins along the x- and y-axis. Coloring by the mean weighted value produces plots like this (matplotlib already handles this):

```python
x,y = np.mgrid[-50:51, -50:51] * 0.1

img = hv.Image(np.sin(x**2+y**2), bounds=(-1,-1,1,1))

hmap = hv.HoloMap({phase: img.clone(np.sin(0.1*x**2+0.1*y**2+phase))
                   for phase in np.linspace(0, np.pi*2, 6)}, kdims=['Phase'])
hmap.hist(num_bins=100, dimension=['x', 'y'], weight_dimension='z', mean_weighted=True)
```

weighted_avg_hist

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -300,7 +300,8 @@ class SideHistogramPlot(ColorbarPlot, HistogramPlot):
             data = dict(top=element.values, left=element.edges[:-1],
                         right=element.edges[1:])
 
-        dim = element.get_dimension(0)
+        color_dims = self.adjoined.traverse(lambda x: x.handles.get('color_dim'))
+        dim = color_dims[0] if color_dims else None
         cmapper = self._get_colormapper(dim, element, {}, {})
         if cmapper:
             data[dim.name] = [] if empty else element.dimension_values(dim)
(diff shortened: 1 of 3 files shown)


philippjfr:
Ready to merge, missing test data doesn't matter as no tests have changed.

jlstevens:
Looks good! 

I was expecting to see a much longer and more complicated diff given the description. Given that this is simply exposing functionality from the backend, I am happy to merge this PR.

That said, does this work in Bokeh if matplotlib isn't available? I'm still a little confused as the diff mostly seems to address color mapping issues and I don't yet see where the ``weight_dimension='z', mean_weighted=True`` arguments get supported (in a backend agnostic way).

philippjfr:
>That said, does this work in Bokeh if matplotlib isn't available? I'm still a little confused: the diff mostly seems to address color mapping issues and I don't yet see where the weight_dimension='z', mean_weighted=True arguments get supported (in a backend agnostic way).

The histogram operation supports these already. It's just that bokeh didn't grab the correct color mapper in this case.

jlstevens:
> The histogram operation supports these already. It's just that bokeh didn't grab the correct color mapper in this case.

Thanks for clarifying. Merging.

PR #1092, opened 2017-01-31
Improve hover tool behavior on Curve types

philippjfr:
Fixes https://github.com/ioam/holoviews/issues/986. Unfortunately in batched mode the hover tool will show all the values in the curve array rather than the specific coordinate you are hovering over so I've disabled displaying curve dimensions when in batched mode.

```python
hv.NdOverlay({i: hv.Curve((range(4), (4-i,i*2, i+3, i))) for i in range(10)}, kdims=['z'])
```

Batched:

screen shot 2017-01-31 at 3 20 38 am

Non-batched:

screen shot 2017-01-31 at 3 19 21 am

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -158,9 +158,7 @@ class CurvePlot(ElementPlot):
             dims = list(self.hmap.last.kdims)
             line_policy = 'prev'
         else:
-            dims = [(d.pprint_label, d.pprint_value(v))
-                    for d, v in self.overlay_dims.items()]
-            dims += element.dimensions()
+            dims = list(self.overlay_dims.keys())+element.dimensions()
             line_policy = 'nearest'
         return dims, dict(line_policy=line_policy)
 
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -7,7 +7,7 @@ try:
 except:
     Bar, BokehBoxPlot = None, None
 from bokeh.models import (Circle, GlyphRenderer, ColumnDataSource,
-                          Range1d, CustomJS, FactorRange)
+                          Range1d, CustomJS, FactorRange, HoverTool)
 from bokeh.models.tools import BoxSelectTool
 
 from ...element import Raster, Points, Polygons, Spikes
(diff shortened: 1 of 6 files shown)


jbednar:
Why is `x` repeated for the non-batched case?

I'm not sure how to parse 'Element: 4'; is that the 4th curve from the bottom?  Could you maybe assign names or values to the curves in the example, so that we can see that the name/value of that curve will be shown (hopefully true?)?

Also, apart from the name/value of the curve, presumably what we really want to see is the values of all other dimensions not mapped to the screen, at that point of that curve.  I guess that's not possible?  Presumably that's shown in the non-batched case?  Would be good to clarify that, and to put that in the docs.

jbednar:
Ok, that looks good.  Presumably the second ``x`` should be suppressed when it's the same as x and y, for the non-batched case?

philippjfr:
>Why is x repeated for the non-batched case?

I've fixed that and renamed the dimension the Curves are indexed by to 'z' in the example.

>Also, apart from the name/value of the curve, presumably what we really want to see is the values of all other dimensions not mapped to the screen, at that point of that curve.

In the non-batched case that will work all dimensions associated which each curve sample will be shown. In the batched case we will unfortunately only see the NdOverlay dimensions and values.

jbednar:
Sounds good, thanks!  Worth explaining anywhere we explain batched=True.  Not sure why one can't do a lookup of whatever data we want to show, knowing the coordinates, but I guess that would turn this into something needing streams, whereas you are talking about Bokeh's native JS-based hover support?

philippjfr:
>I guess that would turn this into something needing streams, whereas you are talking about Bokeh's native JS-based hover support?

Right, indeed I don't really know how to modify the hover labels even using dynamic callbacks yet. This is purely about default ``hover`` behavior.

philippjfr:

jlstevens:
Unfortunately, the new ``test_curve_overlay_hover`` test seems to be failing on Travis...

philippjfr:
Ah, both a python3 issue and outdated test, will push shortly.

philippjfr:
This was an actual bug in python3, hover was broken in some cases.

philippjfr:
Good to have unit tests.

philippjfr:
Ready to merge again.

jlstevens:
Looks good. Merging.

PR #1093, opened 2017-01-31
Operation fixes

philippjfr:
A number of small fixes for existing operations:

1) Disable the ``group`` being inherited from composite types as would sometimes occur when applying operations to an NdOverlay resulting in Element being labeled "NdOverlay".

2) Hide warning in ``datashade`` operation when shading a categorical aggregate.

3) Removed the adjoin parameter to ``histogram`` operation, it continually annoys me because the ``hist`` method has to handle the adjoining itself anyway and having the operation return an adjoined object is a) weird and b) actually breaks if you're trying to apply the operation to a HoloMap or other kind of composite object.

### DIFF
--- a/holoviews/core/element.py
+++ b/holoviews/core/element.py
@@ -39,8 +39,7 @@ class Element(ViewableElement, Composable, Overlayable):
         hists = []
         for d in dimension[::-1]:
             hist = histogram(self, num_bins=num_bins, bin_range=bin_range,
-                             adjoin=False, individually=individually,
-                             dimension=d, **kwargs)
+                             individually=individually, dimension=d, **kwargs)
             hists.append(hist)
         if adjoin:
             layout = self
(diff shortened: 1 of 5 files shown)


philippjfr:
Ready to review.

jlstevens:
Looks good though I am worried about backwards compatibility implications for point 3. I agree with your rationale for removing the adjoin option but is anyone using it?

How about checking if ``adjoin`` is in the kwargs and issuing a useful exception message with some advice? Maybe something like 'Adjoin option deprecated: use the << operator to explicitly adjoin output histogram'?

philippjfr:
>How about checking if adjoin is in the kwargs and issuing a useful exception message with some advice? Maybe something like 'Adjoin option deprecated: use the << operator to explicitly adjoin output histogram'?

I suspect people were sufficiently discouraged from using. I'm happy to add the exception but in fixing it I realized the operation itself was fairly useless in all kinds of ways, auto-ranging was disabled for instance so without supplying an explicit ``bin_range`` or None you'd get a histogram ranging from 0 to 1.

jlstevens:
Ok. In that case, I am happy to merge...although the pr build does seem to have failed right now...

philippjfr:
One of the many transients we're struggling with again, about 1/4 builds fail.

PR #1094, opened 2017-01-31
Allow PositionX/Y/XY streams to work with categorical axes

philippjfr:
Changes the parameter types of these stream types to allow strings from categorical axes, allowing them to work with HeatMap and other categorical plots.

### DIFF
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -5,6 +5,7 @@ server-side or in Javascript in the Jupyter notebook (client-side).
 """
 
 import param
+from numbers import Number
 from collections import defaultdict
 from .core import util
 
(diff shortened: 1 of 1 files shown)


jbednar:
Looks good.  Is numbers.Number better or worse than (int, float)?

jlstevens:
Although the use of ``ClassSelector`` makes the definition of ``PositionXY`` look less trivial than before (which I liked), people *using* this stream class won't notice any difference relative to the old version (unless they look at the docs) and the new version will now work in more cases.

Therefore, I think this approach is better in terms of usability which outweighs the value of having a nice, trivial stream definition in the code. Merging.

PR #1097, opened 2017-02-01
Add interpolate_curve operation and Curve interpolation plot option

philippjfr:
Recently I prototyped a ``Steps`` Element which would plot changes in the y-values of a Curve as discrete steps. It was pointed out this really just represents different interpolation behavior for ``Curve`` plots. I think this is a good suggestion, and I'd like to start using it for a project I'm working on. I've added ``step_curve`` as a separate operation so you can also plot steps when datashading simply by applying it to the Curve before datashading is applied. Here's what it looks like:

```python
%%opts Overlay [legend_position='right' width=600 toolbar='above']
arr = np.random.rand(10)
hv.Overlay([hv.Curve(arr, label=intp)(plot=dict(interpolation=intp))
            for intp in ['linear', 'steps-pre', 'steps-mid', 'steps-post']])
```

screen shot 2017-02-01 at 9 11 19 pm

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -605,6 +605,59 @@ class decimate(ElementOperation):
         return sliced
 
 
+
+
+class interpolate_curve(ElementOperation):
+    """
+    Resamples a Curve using the defined interpolation method, e.g.
+    to represent changes in y-values as steps.
+    """
+
+    interpolation = param.ObjectSelector(objects=['steps-pre', 'steps-mid',
+                                                  'steps-post', 'linear'],
+                                         default='steps-mid', doc="""
+       Controls the transition point of the step along the x-axis.""")
+
+    @classmethod
+    def pts_to_prestep(cls, x, y):
+        steps = np.zeros((2, 2 * len(x) - 1))
+        steps[0, 0::2] = x
+        steps[0, 1::2] = steps[0, 0:-2:2]
+        steps[1:, 0::2] = y
+        steps[1:, 1::2] = steps[1:, 2::2]
+        return steps
+
+    @classmethod
+    def pts_to_midstep(cls, x, y):
+        steps = np.zeros((2, 2 * len(x)))
+        x = np.asanyarray(x)
+        steps[0, 1:-1:2] = steps[0, 2::2] = (x[:-1] + x[1:]) / 2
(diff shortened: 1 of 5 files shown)


philippjfr:
@jordansamuels I've now implemented support for stepping on Curves.

philippjfr:
Bit perplexed by the persistent python2 test failure. Importing from import ``hv.operation`` in plotting somehow changed some axis ranges in two bokeh chart plots (Bars and BoxWhisker)?

jlstevens:
I think an interpolation plot option would be very reasonable but I think the implementation should be generalized further.

I would like to see ``CurvePlot`` pass the ``interpolation`` parameter to a general ``interpolate_curve`` function (or similar) that happens to support the various step options. Right now this operation would support 'linear', 'steps-mid', 'steps-pre' and 'steps-post' but you could imagine it being extended to support additional types of interpolation in future.

holoviews/operation/element.py
```diff
@@ -605,6 +605,53 @@ def _process(self, element, key=None):
...
+    steps = np.zeros((2, 2 * len(x)))
+    x = np.asanyarray(x)
+    steps[0, 1:-1:2] = steps[0, 2::2] = (x[:-1] + x[1:]) / 2
+    steps[0, 0], steps[0, -1] = x[0], x[-1]
+    steps[1:, 0::2] = y
+    steps[1:, 1::2] = steps[1:, 0::2]
+    return steps
+
+
+class step_curve(ElementOperation):
```
jlstevens: I would consider generalizing this to an ``interpolate_curve``  that supports the same interpolation options as ``CurvePlot`` - currently this would just mean adding a 'linear' option that does nothing. I would also consider making the functions above into classmethods.

philippjfr:
Updated, ready to merge when tests pass.

philippjfr:
After rebasing with master this now seems to work okay, ready to merge.

jlstevens:
I guess we could have bezier/splice interpolation in future by sampling a spline into a ``Curve``  - although this would make the sampling rate into a free parameter (though I suspect plotting libraries often do this behind the scenes). I wonder if we could have a ``Spline`` element as a subclass of ``Curve`` but I would need to think about it more...

Merging.

Issue #909, opened 2016-10-06
Support transposing layouts?


jbednar:
Right now, layouts are filled with plots in row order, i.e. left to right and then top to bottom:

![image](https://cloud.githubusercontent.com/assets/1695496/19167404/2f7874b2-8bd1-11e6-8b3a-c52fe45c111b.png)

Sometimes it makes much more sense to order the plots in column order, i.e. top to bottom and then left to right, as in this case, to align related plots.  It is possible to reorder the plots in the layout manually:

``` python
def lay_out_in_columns(_list, cols, spacer=hv.Empty()):
    """Given a list, reorder it into the number of columns specified, filling missing items with the given spacer."""
    rows = int(np.ceil(len(_list)*1.0/cols))
    padded = _list + [spacer] * (rows*cols-len(_list))
    return list(np.array(padded).reshape(cols,rows).T.flat)
```

![image](https://cloud.githubusercontent.com/assets/1695496/19167464/6e005614-8bd1-11e6-97a7-304e8d29561a.png)

This works, but it's awkward, and if subfigure labels weren't suppressed here they would still be in row order, which would be confusing.  So it would be great if there were a plot option to transpose the layout after it has been assigned subfigure labels.

basnijholt:
Seems related to https://github.com/ioam/holoviews/issues/130.

jlstevens:
@basnijholt Thanks for linking to issue #130!

For ``GridSpace``, using the ``reindex`` method worked as that object has its own dimensions. The difference with Layouts is that they can contain heterogeneous mixtures of different element types so I don't believe the same approach will work here.

Certainly a worthwhile comparison to make and definitely something we want to document better.

jbednar:
The layout is essentially just an ordered list, and thus doesn't have any inherent dimensions to transpose directly.  It's only when it's laid out for display that it gets put into a row, column format.  So in the code above I'm reordering this list so that the eventual result is in column order, but really this needs to be something that the layout engine decides as it puts into a columnar shape.

philippjfr:
The ``transpose`` plot option has now been added for all backends, closing.

PR #1100, opened 2017-02-05
Add plot option to transpose layouts

philippjfr:
As requested in #909 this adds a plot option to transpose layouts. It was trivial to implement and provides useful functionality.

### DIFF
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -1030,6 +1030,9 @@ class GenericLayoutPlot(GenericCompositePlot):
     displays the elements in a cartesian grid in scanline order.
     """
 
+    transpose = param.Boolean(default=False, doc="""
+        Whether to transpose the layout when plotting""")
+
     def __init__(self, layout, **params):
         if not isinstance(layout, (NdLayout, Layout)):
             raise ValueError("GenericLayoutPlot only accepts Layout objects.")
@@ -1038,6 +1041,6 @@ class GenericLayoutPlot(GenericCompositePlot):
 
         super(GenericLayoutPlot, self).__init__(layout, **params)
         self.subplots = {}
-        self.rows, self.cols = layout.shape
+        self.rows, self.cols = layout.shape[::-1] if self.transpose else layout.shape
         self.coords = list(product(range(self.rows),
                                    range(self.cols)))
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -1031,7 +1031,9 @@ class GenericLayoutPlot(GenericCompositePlot):
     """
 
     transpose = param.Boolean(default=False, doc="""
-        Whether to transpose the layout when plotting""")
+        Whether to transpose the layout when plotting. Switches
+        from row-based left-to-right and top-to-bottom scanline order
+        to column-based top-to-bottom and left-to-right order.""")
 
     def __init__(self, layout, **params):
         if not isinstance(layout, (NdLayout, Layout)):
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -361,7 +361,7 @@ class LayoutPlot(CompositePlot, GenericLayoutPlot):
         layout_subplots, layouts, paths = {}, {}, {}
         for r, c in self.coords:
             # Get view at layout position and wrap in AdjointLayout
-            key, view = layout_items.get((r, c), (None, None))
+            key, view = layout_items.get((c, r) if self.transpose else (r, c), (None, None))
             view = view if isinstance(view, AdjointLayout) else AdjointLayout([view])
             layouts[(r, c)] = view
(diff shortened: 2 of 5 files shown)


jbednar:
Thanks so much!

holoviews/plotting/plot.py
```diff
@@ -1030,6 +1030,9 @@ class GenericLayoutPlot(GenericCompositePlot):
     displays the elements in a cartesian grid in scanline order.
     """
 
+    transpose = param.Boolean(default=False, doc="""
+        Whether to transpose the layout when plotting""")
```
jlstevens: Would be good to say the normal behavior is scanline order (left to right, top to bottom) and that transposing makes it work top to bottom and left to right.

jlstevens:
Other than a suggestion for improving the docstring, this looks good and I'm happy to merge.

holoviews/plotting/plot.py
philippjfr: Sounds good, I'll make sure to add some tests too.

jbednar:
I can't tell from the code -- will this make the subfigure labels transpose as well?  Presumably we'd want them to, because if the ordering in the layout has any meaning, the layout's order will now be top to bottom instead of left to right.

philippjfr:
>I can't tell from the code -- will this make the subfigure labels transpose as well? Presumably we'd want them to, because if the ordering in the layout has any meaning, the layout's order will now be top to bottom instead of left to right.

True, currently not the case, it simply labels as normal now.

jbednar:
> >I can't tell from the code -- will this make the subfigure labels transpose as well? Presumably we'd want them to, because if the ordering in the layout has any meaning, the layout's order will now be top to bottom instead of left to right.
> 
> True, currently not the case, it simply labels as normal now.

It's a start, then, but changing the label order seems dangerous to put off until later, because if we ever do implement that, a previously labeled transposed figure will no longer match its caption.

philippjfr:
Now handles the subfigure labels too. Ready to merge when tests pass.

jlstevens:
All tests are green. Merging.

jbednar:
Sorry I didn't get a chance to test this earlier, but I think this is a problem:  the `cols()` method now seems to determine the number of rows, which is very unintuitive:

image

Shouldn't `cols()` still determine the number of columns?  Being able to determine the number of rows is nice, but would seem to require having a method other than `cols()`.

philippjfr:
>Shouldn't cols() still determine the number of columns? Being able to determine the number of rows is nice, but would seem to require having a method other than cols().

It could do, but it's not really a transpose in that case, and the implementation would be a lot uglier.

jbednar:
Drat.  If you look at that example notebook as updated with the new colormaps function above, you'll see that the transpose option doesn't really work, then -- the idea was to have a fixed number of columns, but to start to fill them top to bottom, not left to right.  As it is now, sometimes there are two columns, sometimes 50, which doesn't work from a layout perspective because the width of a web page is fixed:

image

I.e., a fixed number of columns is about the only thing that makes sense on a webpage that scrolls vertically; cramming in all the plots left to right doesn't seem like it would normally be useful in practice.

jbednar:
It can be faked by having setting the cols() to a fraction of the length of the layout, but that's very fragile.

philippjfr:
>It can be faked by having setting the cols() to a fraction of the length of the layout, but that's very fragile.

True, it depends on the layout not changing length. I'll investigate what the alternative looks like.

jlstevens:
I'm wondering if the ``(Generic)LayoutPlot`` could use a bit of modulo arithmetic on the number of elements to convert the ``cols`` value appropriately (if transposed).

jbednar:
This definition of cols seems to work fine:

```
def colormaps(named_cms,cols=3,array=xs):
    images = [hv.Image(array, group=name)(style=dict(cmap=cm)) 
              for name,cm in named_cms]
    return hv.Layout(images)(plot=dict(transpose=True)).display('all').cols(int(np.ceil(len(images)*1.0/cols)))
```

So is it possible to stick that when the columns value is used, so that it uses that version instead of the cols number directly, if it's transposed?

jlstevens:
> So is it possible to stick that when the columns value is used, so that it uses that version instead of the cols number directly, if it's transposed?

Right. That is what I was suggesting...and now you've shown a concrete implementation, I can't see why the plotting class can't use similar logic.

philippjfr:
>I can't see why the plotting class can't use similar logic.

It probably can easily enough, I'll look into it.

philippjfr:
One annoyance is that we have the current ``Layout`` logic encoded in ``Layout.shape`` and ``NdLayout.shape``. I'd vote for either deprecating those or actually allow transposing at the object level.

jlstevens:
I don't think ``Layout.shape`` is public API so I am happy to see it changed. Changing ``cols`` is a different matter though but I am open to any backwards-compatible approach.

PR #1103, opened 2017-02-05
Merge dynamic dimension values

philippjfr:
Merges the values from multiple composed DynamicMaps addressing https://github.com/ioam/holoviews/issues/920.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -611,7 +611,16 @@ def python2sort(x,key=None):
 def merge_dimensions(dimensions_list):
     """
     Merges lists of fully or partially overlapping dimensions by
-    combining their values.
+    merging their values.
+
+    >>> from holoviews import Dimension
+    >>> dim_list = [[Dimension('A', values=[1, 2, 3]), Dimension('B')],
+    ...             [Dimension('A', values=[2, 3, 4])]]
+    >>> dimensions = merge_dimensions(dim_list)
+    >>> dimensions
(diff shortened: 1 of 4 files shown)


holoviews/core/util.py
```diff
@@ -608,6 +608,23 @@ def python2sort(x,key=None):
     return itertools.chain.from_iterable(sorted(group, key=key) for group in groups)
 
 
+def merge_dimensions(dimensions_list):
+    """
+    Merges lists of fully or partially overlapping dimensions by
+    combining their values.
+    """
```
jlstevens: The code is clear enough, but it might be nice to add a doctest example here (if we are using doctests).

jlstevens:
Looks like my comment in the issue regarding a ``KeyError`` was confused. Anyway, merging dimensions with ``merge_dimensions`` makes sense - my only question, is there anywhere else we should be using it?

philippjfr:
>my only question, is there anywhere else we should be using it?

I think this uses it everywhere it matters, i.e. when merging ``HoloMaps/DynamicMaps`` in an Overlay and when computing the dimensions during plotting.

jlstevens:
Thanks for adding the doctest!

philippjfr:
Some syntax error due to the test being multi-line.

philippjfr:
Ready to merge when it passes.

jlstevens:
Tests have passed. Merging.

PR #1112, opened 2017-02-08
Set up AWS CLI configuration file

philippjfr:
Testing whether external branch PR can upload test data.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ language: python
 
 sudo: false
 
+env:
+  global:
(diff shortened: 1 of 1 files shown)


philippjfr:
@jlstevens After writing the environment variables to an AWS config file I'm now seeing a new error:

>upload failed: ./test_data.zip to s3://preview.holoviews.org/6558/test_data_py3.zip An error occurred (InvalidAccessKeyId) when calling the CreateMultipartUpload operation: The AWS Access Key Id you provided does not exist in our records.

Maybe the environment variables aren't made available in an external PR build? Would make some sense because otherwise anyone could submit a PR which ``echo``s the supposedly secret variables on travis. My guess is we should switch to the approach of encrypting the variable.

philippjfr:
Superceded by #1117

PR #1115, opened 2017-02-09
Fixed python3 'lod' bug

philippjfr:
Fix for https://github.com/ioam/holoviews/issues/1005. Still don't know what actually caught the bug as 'lod' should always be available in the defaults but this will fix the immediate issue.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -364,7 +364,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         if self.border is not None:
             for p in ['left', 'right', 'top', 'bottom']:
                 plot_props['min_border_'+p] = self.border
-        lod = dict(self.defaults()['lod'], **self.lod)
+        lod = dict(self.defaults().get('lod', {}), **self.lod)
         for lod_prop, v in lod.items():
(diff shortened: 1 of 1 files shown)


philippjfr:

jlstevens:
I remember suggesting this as a temporary fix and I am happy to merge. I'm just wondering what we can do to try and address the underlying issue?

philippjfr:
>I'm just wondering what we can do to try and address the underlying issue?

Add a bunch of print statements to ``Parameterized.defaults`` and then spawn 50 python three processes instantiating a bokeh ElementPlot and call ``defaults`` on it would be my only suggestion.

jlstevens:
Sounds like a real pain! 

It isn't something we need to do now but maybe we should still have an issue to remind us that it is still an issue? What bad behavior can we expect when the ``lod`` is missing after this fix?

philippjfr:
>What bad behavior can we expect when the lod is missing after this fix?

Unless someone sets an explicit ``lod`` configuration as a plot option it will make zero difference. If they do then it will fall back to the ``lod`` defaults bokeh sets up.

jlstevens:
Ok. Doesn't sound too bad, but it is probably worth filing an issue just to register that we are aware of this issue.

Merging.

PR #1124, opened 2017-02-11
Added tutorial about working with renderers and plots

philippjfr:
Initial draft of a Tutorial about working with renderers and plots. Needs review. Here's a rendered version: https://anaconda.org/philippjfr/plots_and_renderers/notebook

### DIFF
--- a/doc/Tutorials/Plots_and_Renderers.ipynb
+++ b/doc/Tutorials/Plots_and_Renderers.ipynb
@@ -1,0 +1,624 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": true
+   },
+   "outputs": [],
+   "source": [
+    "import holoviews as hv\n",
+    "import numpy as np"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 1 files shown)


philippjfr:
I'd like to get this integrated soon, probably as a user guide since it's already very API focused. @jbednar correctly pointed out that it was a bit of mess and a bit repetitive (largely because it contains copied and pasted bits from 3 different versions of similar documents). Suggestions on how to make quick improvements to this very welcome. The main reason I'd like to see this merged is so I can reference it from the user guides about bokeh server deployment, to avoid having to introduce the Renderer concepts there.

jlstevens:
I would like to make a pass through this notebook but I do understand you want it merge ASAP. Here is what I suggest:

1. Don't add this to ``Tutorials``, add it to ``examples/userguide``.
2. Write the user guide reference material in it.
3. When I make a pass through it, I'll have to make sure to update the user guide accordingly which means I'll wait till you have your user guide merged.

Maybe we can have a PR after your first user guide version where I do a pass through this notebook, you can review it and then we can update the user guide accordingly.

jlstevens:
@philippjfr Is this material now in the user guide?

philippjfr:
Yes, it is. Still needs polish but we can close this.

PR #1125, opened 2017-02-11
Added datashader dynspread operation

philippjfr:
As the title says this adds a dynspread operation which implements dynamic pixel spreading on RGB types. It works on all Image, RGB and GridImage types.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -16,12 +16,13 @@ from datashader.pandas import pandas_pipeline
 from datashader.dask import dask_pipeline
 from datashape.dispatch import dispatch
 from datashape import discover as dsdiscover
+import datashader.transfer_functions as tf
 
 from ..core import (ElementOperation, Element, Dimension, NdOverlay,
                     Overlay, CompositeOverlay, Dataset)
 from ..core.data import ArrayInterface, PandasInterface, DaskInterface
 from ..core.util import get_param_values, basestring
-from ..element import GridImage, Path, Curve, Contours, RGB
+from ..element import GridImage, Image, Path, Curve, Contours, RGB
 from ..streams import RangeXY
 
 DF_INTERFACES = [PandasInterface, DaskInterface]
@@ -216,7 +217,6 @@ class aggregate(ElementOperation):
 
 
 
-
 class shade(ElementOperation):
     """
     shade applies a normalization function followed by colormapping to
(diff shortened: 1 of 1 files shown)


holoviews/operation/datashader.py
```diff
@@ -332,3 +332,70 @@ def _process(self, element, key=None):
...
+
+
+class dynspread(ElementOperation):
+    """
+    Spreading expands each pixel in an Image based Element a certain
+    number of pixels on all sides according to a given shape, merging
+    pixels using a specified compositing operator. This can be useful
+    to make sparse plots more visible. Dynamic spreading determines
+    how many pixels to spread based on a density heuristic.
+    """
```
jbednar: Maybe point to this section of the datashader manual, to pass the buck for having to explain it in any more detail? (Not that the datashader site necessarily does, just that if someone wants more explanation, that's the place it should be provided, not in holoviews.)

holoviews/operation/datashader.py
```diff
@@ -332,3 +332,70 @@ def _process(self, element, key=None):
...
+                                        'saturate', 'add'],
+                               doc="""
+        The name of the compositing operator to use when 
+        combining pixels.       
+    """)
+
+    max_px = param.Integer(default=2, doc="""
+        Maximum number of pixels to spread on all sides.
+    """)
+
```
jbednar: Is this the HoloViews docstring style?  It's up to you, but personally I much prefer to have it tighter:

```
max_px = param.Integer(default=2, doc="""
         Maximum number of pixels to spread on all sides.""")
```

philippjfr: You're right, but we generally indent the docstring by four spaces.

jbednar:
Looks good, thanks!

holoviews/operation/datashader.py
jbednar: Oops; that's just from having a proportional font in Github's editing box.  Fixed; I too always use 4.

holoviews/operation/datashader.py
```diff
@@ -332,3 +332,67 @@ def _process(self, element, key=None):
...
+    max_px = param.Integer(default=2, doc="""
+        Maximum number of pixels to spread on all sides.""")
+    
+    shape = param.ObjectSelector(default='circle', objects=['circle', 'square'],
+                                 doc="""
+        The shape to spread by. Options are 'circle' [default] or 'square'.""")
+
+    threshold = param.Number(default=0.8, doc="""
+        A tuning parameter in the range [0, 1]. Indicates the minimum
+        value for the density heuristic.""")
```
jbednar: The important thing to convey here is the directionality -- does a higher value mean more spreading, or less?

philippjfr: I just copied the datashader docstring so you presumably update it there as well.

jbednar: Right; the original isn't clear either. I don't have it running here in a handy way to test it, but if you do, please check which way it works, update the docstring, and then I'll merge and update the original docstring.

holoviews/operation/datashader.py
```diff
@@ -332,3 +332,67 @@ def _process(self, element, key=None):
...
+    """
+
+    how = param.ObjectSelector(default='source',
+                               objects=['source', 'over',
+                                        'saturate', 'add'], doc="""
+        The name of the compositing operator to use when combining
+        pixels.""")
+
+    max_px = param.Integer(default=2, doc="""
+        Maximum number of pixels to spread on all sides.""")
```
jbednar: Is there a reason this default value does not match datashader's default value of 3? I think that could be confusing.

philippjfr: Agreed, I copied from one of my examples.

holoviews/operation/datashader.py
jbednar: How about:

```
threshold = param.Number(default=0.8, bounds=(0,1), doc="""
    When spreading, determines how far to spread.
    Spreading starts at 1 pixel, and stops when the number
    of non-empty pixels adjacent to other non-empty pixels
    reaches this threshold.  Higher values give more 
    spreading.""")
```

jbednar: I've updated the docstring in the datashader source to say the above, now.

philippjfr: Great will make the fix shortly.

PR #1130, opened 2017-02-12
Fix GridInterface multi-dimensional groupby

philippjfr:
As @basnijholt originally reported there was an issue for multi-dimensional groupbys in the GridInterface. This PR fixes the bug and adds a unit test.

### DIFF
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -189,11 +189,11 @@ class GridInterface(DictInterface):
         group_kwargs.update(kwargs)
 
         # Find all the keys along supplied dimensions
-        keys = [dataset.data[d.name] for d in dimensions]
+        keys = [dataset.dimension_values(d.name, False) for d in dimensions]
 
         # Iterate over the unique entries applying selection masks
         grouped_data = []
-        for unique_key in zip(*util.cartesian_product(keys)):
+        for unique_key in zip(*(arr.flat for arr in util.cartesian_product(keys))):
             group_data = cls.select(dataset, **dict(zip(dim_names, unique_key)))
             if np.isscalar(group_data):
                 group_data = {dataset.vdims[0].name: np.atleast_1d(group_data)}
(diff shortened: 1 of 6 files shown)


holoviews/core/data/grid.py
```diff
@@ -189,11 +189,11 @@ def groupby(cls, dataset, dim_names, container_type, group_type, **kwargs):
         group_kwargs.update(kwargs)
 
         # Find all the keys along supplied dimensions
-        keys = [dataset.data[d.name] for d in dimensions]
+        keys = [dataset.dimension_values(d.name, False) for d in dimensions]
 
         # Iterate over the unique entries applying selection masks
         grouped_data = []
-        for unique_key in zip(*util.cartesian_product(keys)):
+        for unique_key in zip(*(arr.flat for arr in util.cartesian_product(keys))):
```
jlstevens: The definition of ``cartesian_product(arrays):`` is simply  ``np.broadcast_arrays(*np.ix_(*arrays))``. I'm wondering if it might be useful behavior to always output an iterator over flattened/flat arrays...

jlstevens: I should mention that I am assuming that we are almost always doing cartesian products over keys which we want to iterate over. I haven't checked where else ``cartesian_product`` maybe used... e.g elsewhere in the data API...

jlstevens:
Made one comment about ``cartesian_product``. Otherwise looks good and I am happy to merge once the tests pass.

holoviews/core/data/grid.py
philippjfr: Right, they're actually always flattened, I'll make that fix.

philippjfr: Ah, I remember now the issue is the difference between ``flat`` and ``flatten``. Often ``flat`` is enough other times you need to flatten. I'll give the function a ``copy`` kwarg.

philippjfr: Hmm, actually there is one usage where it's not flat. Should I also have a ``flat`` kwarg?

jlstevens: Yes, a ``flat`` kwarg would be good and a ``copy`` kwarg also makes sense (should probably default to False).

philippjfr:
Ready to merge.

PR #1131, opened 2017-02-13
Fixed ErrorBarPlot mpl2.0 compat and made it more general

philippjfr:
ErrorBars were one of the only failures in our compatibility with mpl 2.0, the difference is that errorbar caps are disabled by default and that the return signature differs. This adds that compatibility and also made the plot correctly animate ErrorBar plots with varying numbers of error bars and adds support for inverting axes on the plot allowing it to be adjoined correctly.

### DIFF
--- a/holoviews/plotting/mpl/chart.py
+++ b/holoviews/plotting/mpl/chart.py
@@ -1,12 +1,17 @@
 from __future__ import unicode_literals
 
 from itertools import product
+from distutils.version import LooseVersion
 
 import numpy as np
+import matplotlib as mpl
 from matplotlib import cm
 from matplotlib import pyplot as plt
 from matplotlib.collections import LineCollection
+from matplotlib.path import Path as MPLPath
 from matplotlib.dates import date2num, DateFormatter
+
+mpl_version = LooseVersion(mpl.__version__)
 
 import param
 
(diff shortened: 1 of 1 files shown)


holoviews/plotting/mpl/chart.py
```diff
@@ -101,7 +107,14 @@ class ErrorPlot(ChartPlot):
     _plot_methods = dict(single='errorbar')
 
     def init_artists(self, ax, plot_data, plot_kwargs):
-        _, (bottoms, tops), verts = ax.errorbar(*plot_data, **plot_kwargs)
+        handles = ax.errorbar(*plot_data, **plot_kwargs)
+        bottoms, tops = None, None
+        if mpl_version >= str('2.0'):
```
jlstevens: So this is the only bit that changes specifically for matplotlib 2.0+?

philippjfr: No the update code also had to change because the caps aren't instantiated by default.

philippjfr:
Tests appear to be failing because this display output 46 in Elements has changed. Here is what it looks like:

image

What that has to do with errorbars I do not know.

jlstevens:
Wow..that is a pretty strange side-effect! Definitely no error bars there!

philippjfr:
So I'm super confused now, the test above is meant to look like this:

image

Why would annotations disappear after fixing a completely different plot?

philippjfr:
Figured it out, this is why ``*`` imports are bad. The import of ``Path`` from matplotlib is getting imported into ``mpl/__init__.py`` which ends up overriding our registration ``hv.Path``, so paths never end up getting plotted.

jlstevens:
> The import of Path from matplotlib is getting imported into mpl/__init__.py which ends up overriding our registration of hv.Path, so paths never end up getting plotted.

Ha! Yes, let's not use glob imports.

Anyway, now that is resolved, this PR looks fine and the tests have passed without needing to be updated. Merging.

PR #1132, opened 2017-02-13
Update matplotlib titles rather than redrawing

philippjfr:
Fixes https://github.com/ioam/holoviews/issues/1080, but may also cause display test failures because titles aren't drawn multiple times.


jlstevens:
The tests have passed so maybe it can be merged? 

That said, best to check that the tests passed because this case isn't in the tutorials and not because the tests are broken again!

philippjfr:
Tests are definitely running. Ready to merge.

jlstevens:
Tests have passed. Merging.

PR #1133, opened 2017-02-13
Define length on Annotation Elements

philippjfr:
Defining a length is part of the core API and I have run into failures which I cannot reproduce right now because the annotations do not support it. Setting length 1 is consistent with how their ``dimension_values`` methods work, which return length 1 arrays.

Edit: I threw a bug fix for a small bug on the Arrow Annotation

### DIFF
--- a/holoviews/core/element.py
+++ b/holoviews/core/element.py
@@ -193,15 +193,6 @@ class Element(ViewableElement, Composable, Overlayable):
         return np.column_stack(columns)
 
 
-    def __len__(self):
-        """
-        Elements must define a length, by default will simply be
-        computed from dimension values, subclasses may override this
-        for efficiency and correctness.
-        """
-        return len(self.dimension_values(0))
-
-
 
 class Tabular(Element):
     """
--- a/holoviews/core/element.py
+++ b/holoviews/core/element.py
@@ -184,6 +184,15 @@ class Element(ViewableElement, Composable, Overlayable):
         return np.column_stack(columns)
 
 
+    def __len__(self):
+        """
+        Elements must define a length, by default will simply be
+        computed from dimension values, subclasses may override this
+        for efficiency and correctness.
+        """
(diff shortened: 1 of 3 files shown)


jlstevens:
A length of 1 seems reasonable. If ``__len__`` is now required, why not raise a ``NotImplementedError`` in the appropriate base class (maybe ``Element``?)

philippjfr:
I just added a general ``__len__`` implementation for Element which is based on the length of ``dimension_values`` along the first dimension. This should work fine because ``dimension_values`` **HAS** to be implemented and the length of the element is always defined as the length of its ``dimension_values``. Subclasses can override for more efficient implementations.

jlstevens:

philippjfr:
Hmm, so this is problematic, turns out something is actively looking for the ``__len__`` attribute and enabling checks in that case. Since ``Element`` doesn't actually have any values but is used all over the tests this now fails. Making it ``NotImplemented`` would also fail, really not sure what the best solution is.

jlstevens:
> Hmm, so this is problematic, turns out something is actively looking for the __len__ attribute

My initial reaction is this could be due to testing of the boolean value of elements or pretty printing...

philippjfr:
@jlstevens I think this is the correct fix now, I'll still try to reproduce the original bug so I can add a unit test.

philippjfr:
I also just found out that cloning of ``Text`` and ``Arrow`` types was broken, so any deep selection, redim, relabel etc. was breaking. I've now addressed this by reordering the ``.data`` of these types to match the signature and implementing ``.clone`` to handle that.

jlstevens:
Other than one transient error, tests are passing. Ready to merge?

philippjfr:
Yes, I think it's ready now.

jlstevens:
Tests have all passed now. Merging.

PR #1137, opened 2017-02-15
Added support for categorical colormapping in bokeh backend

philippjfr:
Adds support for categorical colormapping on bokeh ``ColorbarPlot`` and implements it specifically for Points. We already have this support in matplotlib and bokeh now makes it easy with a CategoricalColorMapper. This is also already achievable with an NdOverlay of Points but this allows selections to work on a single datastructure which is a lot easier to deal with for streams.

```python
%%opts Scatter [color_index=2 toolbar='above' legend_position='right' width=600] (cmap='Set1' size=10)
hv.Scatter([('Week %d' % (i%10), np.random.rand(), chr(65+np.random.randint(5)))
           for i in range(100)], kdims=['Date'], vdims=['r2', 'Group'])
```

screen shot 2017-02-15 at 12 33 09 am

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -893,7 +893,20 @@ class ColorbarPlot(ElementPlot):
         low, high = ranges.get(dim.name, element.range(dim.name))
         palette = mplcmap_to_palette(style.pop('cmap', 'viridis'), ncolors)
         colors = {k: rgba_tuple(v) for k, v in self.clipping_colors.items()}
+        colormapper, opts = self._get_cmapper_opts(low, high, factors, colors)
 
+        if 'color_mapper' in self.handles:
+            cmapper = self.handles['color_mapper']
+            cmapper.palette = palette
+            cmapper.update(**opts)
+        else:
+            cmapper = colormapper(palette=palette, **opts)
+            self.handles['color_mapper'] = cmapper
+            self.handles['color_dim'] = dim
+        return cmapper
+
+
+    def _get_cmapper_opts(self, low, high, factors, colors):
         if factors is None:
             colormapper = LogColorMapper if self.logz else LinearColorMapper
             if isinstance(low, (bool, np.bool_)): low = int(low)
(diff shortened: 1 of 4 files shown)


philippjfr:

holoviews/plotting/bokeh/element.py
```diff
@@ -887,25 +887,34 @@ def _get_colormapper(self, dim, element, ranges, style):
...
+            if isinstance(high, (bool, np.bool_)): high = int(high)
+            opts = {'low': low, 'high': high}
+            color_opts = [('NaN', 'nan_color'), ('max', 'high_color'), ('min', 'low_color')]
+            opts.update({opt: colors[name] for name, opt in color_opts if name in colors})
+        else:
+            colormapper = CategoricalColorMapper
+            opts = dict(factors=factors)
+            if 'NaN' in colors:
+                opts['nan_color'] = colors['NaN']
+
```
jlstevens: I am wondering if a chunk of this should be a utility, e.g [these lines](https://github.com/ioam/holoviews/blob/77f14160c249b0fcced16ffe31f7d2e00c942e7b/holoviews/plotting/bokeh/element.py#L891-L910). Both ``self.clipping_colors`` and ``self.logz`` seem to be sensible argument for a general color mapping utility.

I'll leave it up to you but to me it seems like a fair chunk of code that doesn't have to be tied up in a plotting class.

philippjfr: Yes, some of that can definitely be made into a utility.

holoviews/plotting/bokeh/element.py
```diff
@@ -942,16 +951,33 @@ class LegendPlot(ElementPlot):
...
     legend_cols = param.Integer(default=False, doc="""
        Whether to lay out the legend as columns.""")
 
-
     legend_specs = {'right': dict(pos='right', loc=(5, -40)),
                     'left': dict(pos='left', loc=(0, -40)),
                     'top': dict(pos='above', loc=(120, 5)),
                     'bottom': dict(pos='below', loc=(60, 0))}
 
+    def _process_legend(self, plot=None):
```
jlstevens: I see this was moved from ``PointPlot`` and I expect it is more general here. That said, does this have anything to do with colormapping or is it just an unrelated improvement?

jlstevens:
Once my two comments are addressed, I'm happy to merge.

holoviews/plotting/bokeh/element.py
philippjfr: No, it was moved from ``BarPlot``. Basically it handles providing control over automatically created legends, which is a general concept. Since categorical colormappers can automatically create a legend it's useful to move this method up to LegendPlot and let ``PointPlot`` make use of it.

holoviews/plotting/bokeh/element.py
philippjfr: Wasn't really worth it factoring much of this out.

philippjfr:
Factored out a tiny little function but the rest of it is just very closely tied to the signatures of the bokeh colormappers and won't benefit much from being separate utilities. This should be ready to merge now.

philippjfr:
Ended up factoring it out into a separate method. Ready to merge now.

jlstevens:
A separate method also helps break up the block of code.

Looks good, merging.

PR #1140, opened 2017-02-20
Added better support for computing bokeh plot size

philippjfr:
The current ``BokehRenderer.get_size`` method only supports very simple plots that are not part of a layout grid, or contain toolbars and tables. This PR computes the size of a bokeh plot more accurately for all these cases, which can be very useful when wanting to embed the plot in HTML and know the size ahead of time.

### DIFF
--- a/holoviews/plotting/bokeh/renderer.py
+++ b/holoviews/plotting/bokeh/renderer.py
@@ -1,18 +1,23 @@
 import uuid
 
+import numpy as np
+import param
+from param.parameterized import bothmethod
+
+from bokeh.charts import Chart
+from bokeh.embed import notebook_div
+from bokeh.io import load_notebook
+from bokeh.model import _ModelInDocument as add_to_document
+from bokeh.models import (Row, Column, Plot, Model, ToolbarBox,
+                          WidgetBox, Div, DataTable, Tabs)
+from bokeh.plotting import Figure
+from bokeh.resources import CDN, INLINE
+
 from ...core import Store, HoloMap
 from ..renderer import Renderer, MIME_TYPES
 from .widgets import BokehScrubberWidget, BokehSelectionWidget
 from .util import compute_static_patch, serialize_json
 
-import param
-from param.parameterized import bothmethod
-
-from bokeh.embed import notebook_div
-from bokeh.io import load_notebook
-from bokeh.resources import CDN, INLINE
-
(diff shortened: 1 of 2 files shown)


philippjfr:
Ready for review and merge.

holoviews/plotting/bokeh/renderer.py
```diff
@@ -114,7 +119,21 @@ def get_size(self_or_cls, plot):
...
+            if not plot.children: return 0, 0
+            horizontal = isinstance(plot, Row) or (isinstance(plot, ToolbarBox) and plot.toolbar_location not in ['right', 'left'])
+            widths, heights = zip(*[self_or_cls.get_size(child) for child in plot.children])
+            w_agg, h_agg = (np.sum, np.max) if horizontal else (np.max, np.sum)
+            width, height = w_agg(widths), h_agg(heights)
+        elif isinstance(plot, (Chart, Figure)):
+            width, height = plot.plot_width, plot.plot_height
+        elif isinstance(plot, (Plot, Div, DataTable)):
+            width, height = plot.width, plot.height
+        if width is None: width = 0
```
jlstevens: Isn't this an error condition? What are you going to do with a plot that is either zero width, zero height or both?

jlstevens:
I made one comment (although I can't say I follow all the detailed size computation logic!) but generally it looks good. 

The main thing I checked was that ``get_size`` is defined as a ``@bothmethod`` which means it is easy to get the size of bokeh plot as long as you can import the ``BokehRenderer`` class.

Happy to merge once you address my comment.

holoviews/plotting/bokeh/renderer.py
philippjfr: Certain models do not declare a width/height and rather than guess at it I'd take the hit and just say it's zero. The only thing that I've found that's actually affected by this is the ``Div`` model used to add titles to grid plots.

philippjfr: Perhaps I'll just add an explicit branch for the Div model,  and make a comment about this.

jlstevens: > Perhaps I'll just add an explicit branch for the Div model, and make a comment about this.

Sounds like a good idea.

philippjfr:
@jlstevens Ready to merge.

jlstevens:
Looks good. Merging.

PR #1143, opened 2017-02-21
Use correct dpi in MPLRenderer.get_size

philippjfr:
Following on from #1140 it turns out the matplotlib equivalent to get the plot size in pixels is also incorrect using the wrong dpi.


jlstevens:
I don't get why ``plot.state.dpi`` isn't always the right dpi value to use....why wouldn't it match ``self_or_cls.dpi`` if output by the same renderer?

philippjfr:
Because the dpi is controlled by the renderer and the plotting classes don't fiddle with the default dpi value on the figure.

jlstevens:
Ok. Is it ready to merge then?

philippjfr:
Yes.

PR #1147, opened 2017-02-23
Deprecate NdElement

philippjfr:
This is an experiment.

Edit: A failed experiment by the looks of it, scarily quite a few of the objects in our tutorials end up as NdElements by default.


jlstevens:
That is scary....``ndelement`` was at the end of the ``datatypes`` list!

philippjfr:
I think it's mostly this data type signature that's not handled by other interfaces:

``{1: 1, 2: 2, 3: 3}`` and ``{(1, 'A'): 1, (2, 'B'): 2, (3, 'C'): 3}``

jlstevens:
Though are both examples of dictionaries, so shouldn't ``DictInterface`` be handling them?

I'm guessing the latter is a way of specifying two key dimensions, one of which is integer and one which is string (and integers for the vdim).

philippjfr:
>Though are both examples of dictionaries, so shouldn't DictInterface be handling them?

No, the dict interface expects dimension names as keys not values.

jlstevens:
That makes sense..and those examples get the names from the kdim/vdim declarations. Tricky.

PR #1150, opened 2017-02-25
Add support for GridSpace axes in bokeh

philippjfr:
Creates outer dimension axes for GridSpaces, like we support in matplotlib:

screen shot 2017-02-25 at 2 21 52 pm

### DIFF
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -202,6 +202,19 @@ class GridPlot(CompositePlot, GenericCompositePlot):
     object.
     """
 
+    axis_offset = param.Integer(default=50, doc="""
+        Number of pixels to adjust row and column widths and height by
+        to compensate for shared axes.""")
+
+    fontsize = param.Parameter(default={'title': '16pt'},
+                               allow_None=True,  doc="""
+       Specifies various fontsizes of the displayed text.
+
+       Finer control is available by supplying a dictionary where any
+       unmentioned keys reverts to the default sizes, e.g:
+
+          {'title': '15pt'}""")
+
     shared_xaxis = param.Boolean(default=False, doc="""
         If enabled the x-axes of the GridSpace will be drawn from the
         objects inside the Grid rather than the GridSpace dimensions.""")
@@ -261,16 +274,17 @@ class GridPlot(CompositePlot, GenericCompositePlot):
                 vtype = None
 
             # Create axes
+            offset = self.axis_offset
             kwargs = {}
             if c == 0 and r != 0:
                 kwargs['xaxis'] = None
-                kwargs['width'] = self.plot_size+50
+                kwargs['width'] = self.plot_size+offset
             if c != 0 and r == 0:
                 kwargs['yaxis'] = None
-                kwargs['height'] = self.plot_size+50
+                kwargs['height'] = self.plot_size+offset
             if c == 0 and r == 0:
-                kwargs['width'] = self.plot_size+50
-                kwargs['height'] = self.plot_size+50
+                kwargs['width'] = self.plot_size+offset
+                kwargs['height'] = self.plot_size+offset
             if r != 0 and c != 0:
                 kwargs['xaxis'] = None
                 kwargs['yaxis'] = None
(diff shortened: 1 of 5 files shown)


philippjfr:
Small change to test data due to a small change in how grid plots are padded.

philippjfr:
Merge this one last, after all the latest once since it will change the test data and then all other PRs will need rebasing.

jlstevens:
All the other HoloViews PRs are now merged so feel free to rebase now and update the test data as appropriate.

holoviews/plotting/bokeh/plot.py
```diff
@@ -235,24 +263,30 @@ def _create_subplots(self, layout, ranges):
             # Create axes
             kwargs = {}
             if c == 0 and r != 0:
-                kwargs['xaxis'] = 'bottom-bare'
-                kwargs['width'] = 150
+                kwargs['xaxis'] = None
+                kwargs['width'] = self.plot_size+50
             if c != 0 and r == 0:
```
jlstevens: Looks like ``50`` is a magic number...

philippjfr: True, it's the offset used to account for the axis width. Could parameterize it I suppose.

holoviews/plotting/bokeh/util.py
```diff
@@ -146,18 +148,64 @@ def layout_padding(plots):
...
+    nchars = np.max([len(f) for f in factors])
+    ranges = FactorRange(factors=factors)
+    ranges2 = Range1d(start=0, end=1)
+    axis_label = dim_axis_label(dim)
+
+    rotation = np.radians(rotation)
+    if axis == 'x':
+        align = 'center'
+        # Adjust height to compensate for label rotation
+        height = int(50 + np.abs(np.sin(rotation)) * (nchars*8))
```
jlstevens: I assume this is the same magic number ``50``. Even if it doesn't have to be a parameter, it should at least be a class attribute...

tests/testrenderclass.py
```diff
@@ -205,7 +205,7 @@ def test_get_size_grid_plot(self):
         grid = GridSpace({(i, j): self.image1 for i in range(3) for j in range(3)})
         plot = self.renderer.get_plot(grid)
         w, h = self.renderer.get_size(plot)
-        self.assertEqual((w, h), (360, 360))
+        self.assertEqual((w, h), (418, 410))
```
jlstevens: Interesting that it used to be perfectly square, but isn't anymore. Generally, square plots seems easier when considering layouts etc...

holoviews/plotting/bokeh/util.py
philippjfr: Sure, might be good to parameterize it actually, if you have really long axis labels the layout solver could blow up I suppose and this would let you have some control.

tests/testrenderclass.py
philippjfr: >Interesting that it used to be perfectly square, but isn't anymore.

Taking into account the fact that a vertical axis is wider than a horizontal one. Might be worth considering using this calculation more generally because I've found I can compute axes widths pretty accurately.

philippjfr: (Probably not a good idea when a JS formatter is responsible for computing ticks).

holoviews/plotting/bokeh/util.py
```diff
@@ -125,7 +128,7 @@ def mpl_to_bokeh(properties):
     return new_properties
 
 
-def layout_padding(plots):
+def layout_padding(plots, renderer):
     """
     Temporary workaround to allow empty plots in a
     row of a bokeh GridPlot type. Should be removed
```
jlstevens: The PR reference in this docstring https://github.com/bokeh/bokeh/issues/2891 looks like it was resolved to your satisfaction in June. This suggests either this function should be removed or the docstring updated.

holoviews/plotting/bokeh/util.py
jlstevens: A class attribute seems like a step in the right direction for now. I wouldn't make it a parameter till we have a concrete, real world example of it being needed.

holoviews/plotting/bokeh/util.py
philippjfr: Yes, will update the docstring (and perhaps reopen the issue). It's not actually fixed.

holoviews/plotting/bokeh/util.py
jlstevens: What about the 8 in ``nchars*8``? Where is the 8 from?

holoviews/plotting/bokeh/plot.py
jbednar: Would be much better to query it than parameterize it.  Can it be calculated as outer dimension minus inner dimension, now that the inner dimensions are available in Bokeh JS?

philippjfr: >Can it be calculated as outer dimension minus inner dimension, now that the inner dimensions are available in Bokeh JS?

No bokehJS is too late to set this, if it's wrong the layout solver breaks and you never get an inner width.

holoviews/plotting/bokeh/util.py
philippjfr: Cheating a bit, it's the fontsize, which I actually don't let you change atm. Could allow changing it and scaling this size accordingly.

jbednar: Sounds very fragile.

philippjfr: I'm actually surprised how well it works, but it will definitely get more brittle if I add support for fontsizes.

philippjfr: It also only works because I'm explicitly formatting the ticks as strings, which makes sense in this case, but it is not a general approach because it will break for ticks computed on the frontend.

philippjfr:
While the approach isn't ideal it works robustly for a wide range of font sizes and grid sizes I've tried and GridSpaces are really useful particularly with the linked zooming that bokeh supports. Here's what the example from the containers tutorial looks like now rendered with bokeh: 

screen shot 2017-02-28 at 2 19 45 am

jbednar:
It's great to have this. Should there be an issue at the bokeh repo to gave an API to query anything you currently have hardcoded?

philippjfr:
>Should there be an issue at the bokeh repo to gave an API to query anything you currently have hardcoded?

I don't think there's much bokeh can do as long as the layout solver is run client side and once that has run it's too late because the solver may have failed completely.

jbednar:
So are these hints, then, not hardcoding? I.e., will the current then figure out the actual values?

philippjfr:
>So are these hints, then, not hardcoding? I.e., will the current then figure out the actual values?

Not it will respect those values but has to draw everything inside those bounds, if things get squeezed too tight it blows up and doesn't render.

philippjfr:
@jlstevens Ready to merge.

jlstevens:
Looks good! Merging.

PR #1152, opened 2017-02-25
Fixed issue with no colorbar dimension and unicode label

philippjfr:
Tiny little fix that allows a colorbar to be drawn for data with no corresponding dimension (e.g. aggregated counts) and fixes a usage of ``str(dimension)`` which is not unicode safe.


jlstevens:
Looks good but seems to have affected the test data which will need to be checked/updated.

jlstevens:
Tests are now passing. Merging.

PR #1155, opened 2017-02-26
Drop usages of len(Element)

philippjfr:
We've decided to avoid using len(Element) wherever possible.


jlstevens:
Happy to merge - I'm just wondering if these were new instances of ``len`` that are being removed again or whether these were missed in an earlier PR when we first discouraged using ``len``...

philippjfr:
I think we just missed these, I've had this fix sitting in my stashes a while so possibly I just forgot to commit them at the time.

jlstevens:
Ok, looks good!

PR #1156, opened 2017-02-26
Fix seaborn Regression plot

philippjfr:
Fixes bug I introduced for some inexplicable reason.


jlstevens:
Looks good and thanks for the additional unit test. Merging.

PR #1158, opened 2017-02-27
Fixed reraising of plotting exceptions

philippjfr:
This has been plaguing me for a long time. In python2 I **always** get these exceptions when there's an issue with the options. I think this is due to the nested try/excepts. Reraising an exception in this way differs slightly between py2 and py3 hence extra conditional. 

```
TypeError: exceptions must be old-style classes or derived from BaseException, not NoneType
```

### DIFF
--- a/holoviews/ipython/display_hooks.py
+++ b/holoviews/ipython/display_hooks.py
@@ -136,9 +136,12 @@ def display_hook(fn):
             return "<b>{name}</b>{msg}<br>{message}".format(msg=msg, **info)
 
         except Exception as e:
+            t, v, tb = sys.exc_info()
             try: option_state(element, state=optstate)
-            except: pass
-            raise
+            except Exception: pass
+            if sys.version_info[0] < 3:
+                raise (t, v, tb)
+            raise v.with_traceback(tb)
     return wrapped
 
 


philippjfr:
Appears there's some persistent py2 failure now, will have to look into it.

jlstevens:
I've always found reports of this error very mysterious - not only does the error message make no sense (all our classes derive from object, we don't have *any* old style classes!) but I've never been able to reproduce it myself.

Using ``raise`` to re-raise an exception is official, documented Python behavior so I don't know why it would cause issues. I am happy to merge a fix as long as the right traceback is displayed - ideally, it would just be a re-raised exception but that is what it is doing now and for some reason it has issues!

philippjfr:
>Using raise to re-raise an exception is official, documented Python behavior so I don't know why it would cause issues

It is if you're simply reraising, if you have another try/except in the same block you have to do what I did in the PR.

jlstevens:
I am happy to merge if you have checked the correct traceback is shown (which was a previous issue with this code that was fixed) and once the tests pass.

jbednar:
This problem has also bedeviled me, and I'm very glad there is a fix!

philippjfr:
I've tested it on py2 and py3. The only thing that would be nice is to maybe shorten the traceback a bit, or rather show just the exception message for ``ValueErrors`` coming from param validation.

jlstevens:
> The only thing that would be nice is to maybe shorten the traceback a bit, or rather show just the exception message for ValueErrors coming from param validation.

That can be implemented in a separate PR.Merging.

PR #1162, opened 2017-02-27
Consistent handing of axis label and tick fontsize

philippjfr:
Also adds support for setting xticks and yticks size independently the same way xlabel and ylabel can be set independently. The bokeh backend wasn't respecting the axis label and tick font sizes previously.


jlstevens:
Looks good - seems like a sensible extension. ``'xticks'`` and ``'yticks'`` are to ``'ticks'`` what ``xlabel`` and ``ylabel`` are to ``labels``. The pluralization is different but in this case it makes sense (``xtick`` sounds like it is for a single tick label which would be wrong).

Merging.

PR #1163, opened 2017-02-27
Matplotlib boxplot bug

philippjfr:
BoxPlot raises an error when put in a Layout currently, small fix to ensure it doesn't. Also snuck in a style parameter for HistogramPlot.

### DIFF
--- a/holoviews/plotting/mpl/chart.py
+++ b/holoviews/plotting/mpl/chart.py
@@ -1059,8 +1059,8 @@ class BoxPlot(ChartPlot):
             data.append(group[group.vdims[0]])
             labels.append(label)
         style['labels'] = labels
-        style.pop('zorder', None)
-        style.pop('label', None)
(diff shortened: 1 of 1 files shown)


holoviews/plotting/mpl/chart.py
```diff
@@ -1058,8 +1059,8 @@ def get_data(self, element, ranges, style):
             data.append(group[group.vdims[0]])
             labels.append(label)
         style['labels'] = labels
-        style.pop('zorder')
-        style.pop('label')
+        style.pop('zorder', None)
+        style.pop('label', None)
```
jlstevens: ``None`` is the default if unset? If you just want to remove these keys we might want to switch to:

```
style = {k:v for k,v in style.items() if k not in ['zorder', 'label']
```

I might prefer this to using pop, especially as you aren't using the popped value.

jlstevens:
Looks good. Merging.

PR #1164, opened 2017-02-27
Correctly import top-level util module

philippjfr:
Fixes https://github.com/ioam/holoviews/issues/1134


jlstevens:
Glad this is fixed, but can you explain it to me?

philippjfr:
Honestly, I can't. Apparently a module has to exported with an explicit import and since we have ``from __future__ import absolute_imports`` enabled and:

```python
import .util
```

isn't valid, this was the only thing left to try and it seems to work.

PR #1170, opened 2017-03-02
Fix Dynamic utility operation parameter updates

philippjfr:
The ``Dynamic`` utility allows applying an operation dynamically. When applying an operation with a stream in this way currently the stream is updated with all the operations parameters. This is problematic for two reasons, it overrides defaults you've set on your stream and it currently triggers the stream to update which can cause bad loops.

This PR ensures that the stream only inherits parameter values which are not already defined on the stream and ensures that simply applying the dynamic operation does not trigger stream subscribers to be notified.

### DIFF
--- a/holoviews/util.py
+++ b/holoviews/util.py
@@ -50,8 +50,10 @@ class Dynamic(param.ParameterizedFunction):
                 elif not isinstance(stream, Stream):
                     raise ValueError('Stream must only contain Stream '
                                      'classes or instances')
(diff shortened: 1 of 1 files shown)


jbednar:
Looks like the right thing to do, to me.

jlstevens:
Looks good to me!

philippjfr:
Ready to merge.

jlstevens:
Merging.

PR #1174, opened 2017-03-05
Optimization for heatmap aggregation with pandas

philippjfr:
A 5-10x speedup for HeatMap aggregation when using a pandas/dask interface.

```python
data = [(i, j, np.random.rand()) for i in range(500) for j in range(500)]

%%timeit
hv.HeatMap(data)
```

Before:

1 loop, best of 3: 8.61 s per loop

After:

1 loop, best of 3: 800 ms per loop

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -516,17 +516,15 @@ class Dataset(Element):
         return self.interface.dimension_type(self, dim_obj)
 
 
-    def dframe(self, dimensions=None, copy=True):
+    def dframe(self, dimensions=None):
         """
         Returns the data in the form of a DataFrame. Supplying a list
         of dimensions filters the dataframe. If the data is already
-        a DataFrame copy=False may be supplied to avoid making a copy.
+        a DataFrame a copy is returned.
         """
-        if pd is None:
-            raise Exception("Cannot return data as dataframe, pandas is not available")
-        elif dimensions:
+        if dimensions:
             dimensions = [self.get_dimension(d, strict=True).name for d in dimensions]
-        return self.interface.dframe(self, dimensions, copy)
+        return self.interface.dframe(self, dimensions)
 
 
     def columns(self, dimensions=None):
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -17,13 +17,11 @@ from .grid import GridInterface
 from .ndelement import NdElementInterface
 
 datatypes = ['array', 'dictionary', 'grid', 'ndelement']
-DF_INTERFACES = []
 
 try:
     import pandas as pd # noqa (Availability import)
     from .pandas import PandasInterface
     datatypes = ['array', 'dataframe', 'dictionary', 'grid', 'ndelement']
-    DF_INTERFACES.append(PandasInterface)
     DFColumns = PandasInterface
 except ImportError:
     pass
@@ -51,7 +49,6 @@ except ImportError:
 try:
     from .dask import DaskInterface
     datatypes.append('dask')
-    DF_INTERFACES.append(DaskInterface)
 except ImportError:
     pass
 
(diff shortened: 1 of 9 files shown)


jbednar:
Looks good to me.

jlstevens:
I find ``DF_INTERFACES`` a bit ugly but otherwise it looks good.

philippjfr:
Anything naming you would prefer?

jlstevens:
Could the interfaces not declare the sort of APIs they support?

philippjfr:
>Could the interfaces not declare the sort of APIs they support?

This is bypassing the API, because pandas/dask have optimized implementations for this operation.

jlstevens:
To clarify: shouldn't the interfaces declare the third party APIs (dataframe-like, array-like) that the *data* supports? This is also the API assumed by the interface class itself.

jlstevens:
Something like:

```
if 'Dataframe' in reindexed.interface.interface_type:
  ...
```

philippjfr:
>shouldn't the interfaces declare the third party APIs (dataframe-like, array-like) that the data supports?

Sure they could, although "dataframe-like" isn't a particular solid guarantee on how similar and extensive the API is.

jlstevens:
I think the same can be argued for the current approach. For instance, maybe only dask and pandas should claim to support the same interface type. I think such a mechanism is cleaner than building a list based on successful/failing imports.

jlstevens:
In addition, you might want to always have ``'dask'`` ahead of ```'dataframe'`` in ``datatypes`` if the former can be considered a more highly optimized version of the latter. If dask isn't installed, it won't be used.

philippjfr:
>In addition, you might want to always have 'dask' ahead of ```'dataframe'indatatypes`` if the former can be considered a more highly optimized version of the latter.

That's not really a safe assumption, by default dask will simple use a single partition and may therefore actually be slower than pandas and their lazy nature is probably a bit surprising/confusing for a user who doesn't know anything about them.

jlstevens:
Ok sure, that suggestion was orthogonal to the original one about declaring the type of data interface anyhow.

philippjfr:
Okay, I did two things in the end, I got rid of ``DF_INTERFACES`` and added a ``copy`` keyword argument to the ``Dataset.dframe`` method, which lets you avoid making copies if the data is already a dataframe.

holoviews/core/data/__init__.py
```diff
@@ -516,13 +516,17 @@ def get_dimension_type(self, dim):
         return self.interface.dimension_type(self, dim_obj)
 
 
-    def dframe(self, dimensions=None):
+    def dframe(self, dimensions=None, copy=True):
```
jlstevens: I'm not sure copy argument makes much sense if the element isn't already using a dataframe based interface - for other interfaces, don't you always have to create a new dataframe - which would be the same as ``copy`` being fixed to ``True``?

holoviews/element/util.py
```diff
@@ -134,7 +139,13 @@ def _aggregate_dataset(self, obj, xcoords, ycoords):
         dtype = 'dataframe' if pd else 'dictionary'
         dense_data = Dataset(data, kdims=obj.kdims, vdims=obj.vdims, datatype=[dtype])
         concat_data = obj.interface.concatenate([dense_data, obj], datatype=[dtype])
-        agg = concat_data.reindex([xdim, ydim], vdims).aggregate([xdim, ydim], reduce_fn)
+        reindexed = concat_data.reindex([xdim, ydim], vdims)
+        if pd:
```
jlstevens: Why not use ``reindexed.interface.dframe(dimensions=None, copy=False)`` instead of exposing the copy keyword argument at the element level? For ``copy=False`` to work, you are already assuming a dataframe type interface is being used...

jlstevens: I suppose the other thing you could do is complain if ``copy=False`` is passed to the ``dframe`` method of any interface that isn't based on dataframes.

holoviews/core/data/__init__.py
philippjfr: That's true, it's more like ``avoid_copy``, but I think providing a consistent API to get a hold of a dataframe with the minimal amount of overhead is useful.

holoviews/element/util.py
philippjfr: >For copy=False to work, you are already assuming a dataframe type interface is being used...

Because then I need conditional branches for the "is already dataframe" and "convert to dataframe" paths again. I guess I agree ``copy`` is confusing because you might assume you can mutate the dataframe and have an effect on the original element if you don't make a copy, when the real point of it is to avoid making pointless copies.

holoviews/core/data/__init__.py
philippjfr: That said, I'd also be fine having a utility for it instead.

holoviews/element/util.py
jlstevens: Would there be any harm with the dataframe interfaces just avoiding pointless copies automatically? Then it doesn't have to be something the user needs to ever think about...

philippjfr: In my usage of dframe I often create it and then assign to it so that would be a bit of pain.

jlstevens:
I feel the new approach using a utility is much nicer, thanks!

Tests are passing now. Merging.

Issue #1176, opened 2017-03-05
Some kind of global variable naming issue regarding `datashader` in holoviews


PeterDSteinberg:
Try:
```
from holoviews.operation.datashader import datashade
import datashader
```

It will fail on the second command.

Reverse it and still fails on second command:

```
import datashader
from holoviews.operation.datashader import datashade
```
Suggests some confusion regarding the use of identifier `datashader` within `holoviews.operation`.

This problem exists in Jupyter notebook, but if you go into Ipython console you can do the imports in either order.

philippjfr:
Thanks for reopening this here. Could you confirm the Jupyter, IPython and Python versions you are using? I've tried to reproduce with py2 and py3 without luck.

PeterDSteinberg:
Hi @philippjfr 

Python: `2.7.13`
Here's the Jupyter details:
```
jupyter                   1.0.0                    py27_3
jupyter_client            5.0.0                    py27_0
jupyter_console           5.1.0                    py27_0
jupyter_core              4.3.0                    py27_0
```

Ipython:

```
ipykernel                 4.5.2                    py27_0
ipython                   5.3.0                    py27_0
ipython_genutils          0.1.0                    py27_0
ipywidgets                5.2.2                    py27_1
```

I arrived at my current environment by doing:
```
conda create --name ds27 python=2.7
source activate ds27
conda install -c ioam/label/dev holoviews
conda install datashader
cd $C/datashader
python setup.py develop
conda install -c ioam paramnb
conda install -c scitools/label/dev -c conda-forge iris cartopy
conda install xarray
conda install -c conda-forge -c ioam holoviews>=1.6.2 geoviews
cd $C/holoviews && python setup.py install
```

philippjfr:
Perfect thanks, I'll have a stab.

philippjfr:
@PeterDSteinberg I still haven't been able to reproduce this issue. Are you still seeing this?

philippjfr:
I'm going to close this, since I suspect it's caused by this line:

>conda install -c conda-forge -c ioam holoviews>=1.6.2 geoviews

as it is installing holoviews 1.6.2, which does not include the datashader operations. It should therefore automatically be fixed once 1.7 is out.

PR #1177, opened 2017-03-05
fix datashade import from .operation

PeterDSteinberg:
This is a one-line fix to allow `from holoviews.operations import datashade` .  See also issue #1176 that mentioned the import error.

### DIFF
--- a/holoviews/operation/__init__.py
+++ b/holoviews/operation/__init__.py
@@ -3,6 +3,7 @@ from ..core.options import Compositor
 
 from .element import *      # noqa (API import)
 from ..core import Overlay  # noqa (API import)
+from .datashader import datashade
 
 def public(obj):
     if not isinstance(obj, type): return False


philippjfr:
Ah, I've been misreading the import the whole time. The issue here is that HoloViews or other operations do not depend on datashader and the other heavy-weight dependencies it requires. Importing the module here will make those a hard dependency. I do agree it should be simpler to import. @jlstevens @jbednar What would you think of conditionally importing them into the operation module if the dependencies are met?

jlstevens:
> What would you think of conditionally importing them into the operation module if the dependencies are met?

That seems reasonable to me. That said, it might be better to explicitly import from the datashader module instead of sometimes having an import that fails (i.e if the dependencies aren't met)? Do we know what is going on with the ``datashader`` variable in #1176? If not, maybe we should just fix that issue....

jbednar:
I'd want to avoid doing anything weird with imports, and don't understand what's going wrong in this case.

PeterDSteinberg:
I was under the understanding that absolute imports like `from holoviews.operation.datashader import datashade` rather than dotted relative imports were preferred and I have done most projects here with absolute imports.  Is there a particular reason why we do relative here?

philippjfr:
>I was under the understanding that absolute imports like from holoviews.operation.datashader import datashade rather than dotted relative imports were preferred and I have done most projects here with absolute imports.

Depends on where this preference comes from, the original anti-recommendation against relative imports preceded support for dotted relative imports, here's Guido's comment on the topic:

>> The issue is implementing a PEP with nice support for relative
> imports, and then documenting that it should never be used.

>Isn't this mostly historical? Until the new relative-import syntax was
implemented there were various problems with relative imports. The
short-term solution was to recommend not using them. The long-term
solution was to implement an unambiguous syntax. Now it is time to
withdraw the anti-recommendation. Of course, without going overboard
-- I still find them an acquired taste; but they have their place.

That said, I don't have any strong preferences so the relative import style used in holoviews is more of a historical artifact at this point. That said we should at minimum be importing ``from __future__ import absolute_import`` everywhere.

philippjfr:
I'm going to close this because it has either been fixed or was the result of accidentally installing 1.6.2 due to faulty install instructions. If this does recur we'll have to find a fix that doesn't rely on always importing the datashader module.

PR #1180, opened 2017-03-07
Fixed top-level util import issue

philippjfr:
As pointed out in https://github.com/ioam/holoviews/issues/1134#event-979913858 it turns out my previous fix wasn't quite sufficient, the wildcard import from ``hv.core.spaces`` was overriding ``hv.util`` with ``hv.core.util``. Ideally we should get rid of all wildcard imports to avoid issues like this.


PeterDSteinberg:
I agree on getting rid of wildcard imports, or limiting their usage and using them only when `__all__` has been defined.

philippjfr:
Right, I think we can easily replace the remaining wildcard import in the top-level ``__init__``, the only one I would leave is for the ``element`` module, which does define ``__all__`` and would be too much of a pain to list out explicitly.

philippjfr:
Just checked and all other imports define ``__all__``. Ready to merge.

jlstevens:
Great!

PR #1187, opened 2017-03-10
Fixed handling of shared_axes

philippjfr:
The bokeh backend supports a shared axes option which allows axes with the same axis labels to be linked. The current approach taken here is not quite correct because the plot doesn't remember that an axis is shared so it will simply override extents set by a different plot. Secondly setting ``x/yaxis='bare'`` will currently disable ``shared_axes`` because it removes the axis label.

While this behavior is a lot more consistent, we might want to consider disabling it by default in some cases (e.g. on a simply ``Layout``) because it can interfere with axiswise normalization.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -181,6 +181,9 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         self.callbacks = self._construct_callbacks()
         self.static_source = False
 
+        # Whether axes are shared between plots
+        self._shared = {'x': False, 'y': False}
+
 
     def _construct_callbacks(self):
         """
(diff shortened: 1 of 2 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -260,6 +263,28 @@ def _get_hover_data(self, data, element, empty=False):
...
+                if plot.xaxis[0].axis_label == xlabel:
+                    plot_ranges['x_range'] = plot.x_range
+                if plot.xaxis[0].axis_label == ylabel:
+                    plot_ranges['y_range'] = plot.x_range
+            if hasattr(plot, 'yaxis'):
+                if plot.yaxis[0].axis_label == ylabel:
+                    plot_ranges['y_range'] = plot.y_range
+                if plot.yaxis[0].axis_label == xlabel:
+                    plot_ranges['x_range'] = plot.y_range
+        return plot_ranges
```
jlstevens: I'm a little confused here...there are only two keys 'x_range' and 'y_range' so if you iterate over a bunch of plots can't you end up clobbering these keys with different values? Won't you end up with whatever was found at the end of the loop?

philippjfr: That is correct, I could have two separate loops and break when either is found first, but in practice if ``shared_axes`` are enabled all the plots will share the same axes anyway so it doesn't much matter whether I end up using the axis from the first or last plot, because they are all the same anyway.

jlstevens: Ok, good that you've thought about this. It is also good to have this pulled out as a method.

jlstevens:
Looks good and thanks for adding the unit tests.

Merging.

PR #1191, opened 2017-03-11
Avoid dependence on NdElement and add deprecation warning

philippjfr:
Seemingly straightforward, needs further testing though.

Ended up reviving the NdElement deprecation PR here.

### DIFF
--- a/holoviews/core/data/dictionary.py
+++ b/holoviews/core/data/dictionary.py
@@ -56,6 +56,11 @@ class DictInterface(Interface):
             data = {k: data[:,i] for i,k in enumerate(dimensions)}
         elif isinstance(data, list) and np.isscalar(data[0]):
             data = {dimensions[0]: np.arange(len(data)), dimensions[1]: data}
+        elif (isinstance(data, list) and isinstance(data[0], tuple) and len(data[0]) == 2
+              and any(isinstance(v, tuple) for v in data[0])):
+            dict_data = zip(*((util.wrap_tuple(k)+util.wrap_tuple(v))
+                              for k, v in data))
+            data = {k: np.array(v) for k, v in zip(dimensions, dict_data)}
         # Ensure that interface does not consume data of other types
         # with an iterator interface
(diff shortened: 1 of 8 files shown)


jlstevens:
> Ended up reviving the NdElement deprecation PR here.

If you are trying to deprecate ``NdElement`` entirely, then I fully support that! The title of this PR will need to be updated though...

philippjfr:
>If you are trying to deprecate NdElement entirely, then I fully support that! The title of this PR will need to be updated though...

That is the idea yes, the problem is pickle compatibility which would be a real pain.

jlstevens:
> the problem is pickle compatibility which would be a real pain.

Agreed. For this reason if you do get pickle compatibility working, make sure to add a deprecation warning so that when the time comes, we can strip out the (inevitably!) horrible pickling hacks.

philippjfr:
As outlined in https://github.com/ioam/holoviews/issues/1146, this PR now stops NdElement being used by default under any circumstances, with an additional deprecation warning issued when it an NdElement is instantiated for an unforeseen reason. In v2.0 NdElements will be removed and I'd even support dropping pickle support.

jlstevens:
>  In v2.0 NdElements will be removed and I'd even support dropping pickle support.

Agreed. Version 2.0 is also there to allow us to clean up things like this and make a break with the 1.x series. We should consider backporting fixes from 2.0 into 1.7.x and if we find there is demand for continued 1.0 support, we should consider backporting key features from 2.0 into a 1.8 release (after 2.0).

philippjfr:
Ready to merge.

jlstevens:
The title should still be updated as I think it is now more general than what is suggested.

I assume I'm not seeing any ``__getstate__`` or ``__setstate__`` magic because the class is still there and if unpickled will warn? I'm not entirely sure that is true as I can't remember if pickle works via the constructor or not...

jlstevens:
In discussion we decided that there would be plenty of warnings when working with ``NdElement`` (due to clone) even if the unpickling step itself doesn't warn.

Merging.

PR #1192, opened 2017-03-11
Implemented general xarray aggregate/reduce

philippjfr:
Turns out xarray has a reduce method making single- or multi-dimensional aggregate and reduce trivial. Thanks @schlady.


jlstevens:
Very nice! Merging.

schlady:
Thank you. Works like a charm 👍

PR #1194, opened 2017-03-11
Gridmatrix diagonal types

philippjfr:
Allowed different types along the diagonal of a gridmatrix. Also lifted the restriction on duplicate dimensions on a Dimensioned object.  While I can sort of see why we did that, plotting a variable against itself can be a valid thing to do, e.g. to see the distribution of a variable as shown along the diagonal here (which is now also a holoviews-contrib gallery example):

http://bokeh.pydata.org/en/latest/docs/gallery/iris_splom.html

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -758,12 +758,12 @@ class gridmatrix(param.ParameterizedFunction):
         for d1, d2 in permuted_dims:
             if d1 == d2:
                 if p.diagonal_type is not None:
-                    if p.diagonal_type._1d:
-                        values = element.dimension_values(d1)
-                        el = p.diagonal_type(values, vdims=[d1])
-                    else:
+                    if p.diagonal_type._auto_indexable_1d:
                         el = p.diagonal_type(el_data, kdims=[d1], vdims=[d2],
                                              datatype=['dataframe', 'dictionary'])
+                    else:
+                        values = element.dimension_values(d1)
+                        el = p.diagonal_type(values, vdims=[d1])
                 elif p.diagonal_operation is histogram or isinstance(p.diagonal_operation, histogram):
                     bin_range = ranges.get(d1.name, element.range(d1))
                     opts = dict(axiswise=True, framewise=True)
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -758,8 +758,12 @@ class gridmatrix(param.ParameterizedFunction):
         for d1, d2 in permuted_dims:
             if d1 == d2:
                 if p.diagonal_type is not None:
-                    values = element.dimension_values(d1)
-                    el = p.diagonal_type(values, vdims=[d1])
+                    if p.diagonal_type._1d:
+                        values = element.dimension_values(d1)
+                        el = p.diagonal_type(values, vdims=[d1])
+                    else:
+                        el = p.diagonal_type(el_data, kdims=[d1], vdims=[d2],
+                                             datatype=['dataframe', 'dictionary'])
                 elif p.diagonal_operation is histogram or isinstance(p.diagonal_operation, histogram):
                     bin_range = ranges.get(d1.name, element.range(d1))
                     opts = dict(axiswise=True, framewise=True)
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -144,7 +144,7 @@ class Dataset(Element):
 
     # In the 1D case the interfaces should not automatically add x-values
     # to supplied data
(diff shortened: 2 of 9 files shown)


holoviews/operation/element.py
```diff
@@ -758,8 +758,12 @@ def _process(self, p, element, ranges={}):
         for d1, d2 in permuted_dims:
             if d1 == d2:
                 if p.diagonal_type is not None:
-                    values = element.dimension_values(d1)
-                    el = p.diagonal_type(values, vdims=[d1])
+                    if p.diagonal_type._1d:
```
jlstevens: What is this ``_1d`` attribute on elements?

philippjfr: It makes the distinction between Elements that will automatically fill in the x-coordinates and ones that simply render a distribution, e.g. Scatter/Curve vs Distribution/BoxWhisker.

jlstevens: I'm not sure I like this approach. When was it added and how much is it used?

As far as I can tell this is a class attribute and I'm coming to the conclusion there are lots of properties (such as this one) we want to query at the level of the element type.  I would rather not have an undocumented API consisting of all sorts of random class attributes all specified on the element classes.

Instead I'm thinking there could be a set of utilities (either in ``util`` or even a new file) that is all about specifying the properties of Elements in a public way. These could be a bunch of predicate functions (with docstrings to explain what the properties mean). E.g the above might be something like:

```python
if util.is_distribution1D(p.diagonal_type):
...
```

If nothing else, the issue here is that I had no idea what ``_1d`` indicates from the name!

philippjfr: That does not seem extensible. If I want to declare a new Element type, which has one of these properties how would I tell ``is_distribution1D`` about that? As far as I'm aware this is the only class attribute like this and is defined like this on ``Dataset``:

```
    # In the 1D case the interfaces should not automatically add x-values
    # to supplied data
    _1d = False
```

jlstevens: > That does not seem extensible. If I want to declare a new Element type, which has one of these properties how would I tell is_distribution1D about that?

Well, nothing says you'll remember to override ``_1d`` when you define your new class! It is good to hear it is one of the only things like this - maybe it just needs a better name then?

philippjfr: Any suggestions on naming, ``_add_index`` would come close, but really it's ``_add_integer_index_when_only_1d_array_supplied``.

jlstevens: Naming this one is hard!

Also, I was wondering if the suggestion for utilities to get these properties might help with figuring out appropriate output elements after reducing the number of dimensions? These utilities could look at something like ``element._properties`` defined at the class level which could be a dictionary (with a key ``integer_index_1d`` - or something similar in this case)?

This helps the naming issue a bit because if you have a function, at least there is a docstring to consult describing what it is, instead of trying to find a sufficiently descriptive attribute name.

Not something we should do now though. From what I understand about it - how about ``_integer_indexable_1d``? Still not great though...

philippjfr: ``element._properties`` sounds like a reasonable suggestion. For the time being ``_integer_indexable_1d`` isn't really correct though, it only adds an integer index if you don't supply x-values yourself, ``_add_1d_index`` maybe?

jlstevens: ``_augmentable_1d_index``? or ``_allows_implicit_1d_index``? or ``_auto_1d_indexable``?

Tricky!

philippjfr: >``_augmentable_1d_index``

Don't really know what that means.

> ``_allows_implicit_1d_index``

Allows doesn't seem right, it's adding an index if you don't provide it.

> ``_auto_1d_indexable``

This comes closest out of your suggestions.

jlstevens: I thought that might be the best one. How about ``_auto_indexable_1d``?

philippjfr: Sounds good, I'll go with that.

philippjfr:
I'm going to open a separate issue about the add index flag because currently it's being (ab)used in two ways, 1) to find out whether to add a 1D integer index if no x-values have been supplied and 2) whether an Element can represent a distribution of values without an index (i.e. no kdims). Ideally I'd just use the default number of kdims to determine 2) but it seems Spikes currently does not correctly follow that pattern.

jlstevens:
Please do file an issue. The ``_properties`` suggestion would let you decouple these concepts as necessary.

jlstevens:
Now the tests are passing, I'll merge.

PR #1195, opened 2017-03-11
Fixed bug handling rgba tuples in bokeh

philippjfr:
Bokeh accepts an rgba tuple not a list.


jlstevens:
Thanks for the fix. Merging.

PR #1196, opened 2017-03-12
Added bokeh VectorField plot

philippjfr:
This completes compatibility for all chart plots, that only leaves the ``Arrow`` annotation and 3D Elements.

### DIFF
--- a/doc/Tutorials/Bokeh_Elements.ipynb
+++ b/doc/Tutorials/Bokeh_Elements.ipynb
@@ -38,7 +38,7 @@
     "  <dt><a href=\"#BoxWhisker\"><code>BoxWhisker</code></a></dt><dd>Distributions of data varying by 0-N key dimensions.<font color='green'>&#x2713;</font></dd>\n",
     "  <dt><a href=\"#Scatter\"><code>Scatter</code></a></dt><dd>Discontinuous collection of points indexed over a single dimension. <font color='green'>&#x2713;</font></dd>\n",
     "  <dt><a href=\"#Points\"><code>Points</code></a></dt><dd>Discontinuous collection of points indexed over two dimensions. <font color='green'>&#x2713;</font></dd>\n",
-    "  <dt><a href=\"#VectorField\"><code>VectorField</code></a></dt><dd>Cyclic variable (and optional auxiliary data) distributed over two-dimensional space. <font color='red'>&#x2717;</font></dd>\n",
+    "  <dt><a href=\"#VectorField\"><code>VectorField</code></a></dt><dd>Cyclic variable (and optional auxiliary data) distributed over two-dimensional space. <font color='green'>&#x2713;</font></dd>\n",
     "  <dt><a href=\"#Spikes\"><code>Spikes</code></a></dt><dd>A collection of horizontal or vertical lines at various locations with fixed height (1D) or variable height (2D). <font color='green'>&#x2713;</font></dd>\n",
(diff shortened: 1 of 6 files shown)


jlstevens:
Fantastic!

Would it be possible to update the website so I can have a look at what the new bokeh plot looks like? I'm still seeing the old version:

image

I can merge before then if you need it.

philippjfr:
You could trigger a rebuild, but for simplicities sake, here's a side-by-side:

screen shot 2017-03-12 at 1 27 56 pm

screen shot 2017-03-12 at 1 28 42 pm

I am cheating a little bit because I've not found a way to add arrows to the vectors, so really it's showing orientation rather than direction. Better than nothing though, unless you think that's a dealbreaker.

jlstevens:
Personally, I think the direction (and not just the angle) is a pretty important part of the semantics of 'vector'. I'm now undecided, @jbednar you have lots of experience with both orientation/vector fields..what do you think?

jbednar:
I think it's better to have this than not to, so I'm happy to see it merged.  But yes, I think having some way to figure out which end is which is a dealbreaker for the majority of uses of this plot.  Can you file an issue with Bokeh about how to do the arrows, both in this case and for the Arrow annotation?  I'm not sure what the underlying problem is.

philippjfr:
> both in this case and for the Arrow annotation?

I'll try some ideas I have on this implementation, but bokeh's Arrow annotation is perfectly well defined, the issue is that our ``Arrow`` Element is very limited in what it can do and should be changed, see https://github.com/ioam/holoviews/issues/34.

jlstevens:
> ... is that our Arrow Element is very limited in what it can do

This is mainly because the holoviews ``Arrow`` was only designed as an annotation element i.e to point on things on a plot with a label. I don't think the ``Arrow`` element should be used to define vector fields, but it would be ok if the bokeh ``VectorFieldPlot`` class uses bokeh arrow annotations.

philippjfr:
>... is that our Arrow Element is very limited in what it can do

Sorry wasn't clear, what I meant is that our arrow element only has an x, y location and a direction but what we really need is for it to specify the start and end points of the arrow so you can have arbitrary directions. Nothing to do with vector fields.

philippjfr:
Now with arrowheads:

screen shot 2017-03-13 at 12 02 46 am

jlstevens:
Great!

Looks like it was a bit more effort to get the arrows positioned properly (using 'segment' instead of 'ray') but I do think supporting arrows properly is worth it.

philippjfr:
Ready to merge.

jlstevens:
Tests are green. I'm glad I can merge it with proper arrow support - thanks!

jbednar:
Excellent!  Thanks for doing this!

philippjfr:
Now ``Arrow`` is really bugging me though:

screen shot 2017-03-13 at 8 08 19 pm

jbednar:
It does stand out there, doesn't it? :-)

jbednar:
Probably ought to move Chart3D to the end of that list, btw; it's not the first thing people need and it's demotivating to see something unsupported as the first thing in a list.

philippjfr:
Agreed, although I just updated the image, I'd left out all the charts.

jbednar:
Right; I figured that out and updated my own comment to match. :-)

philippjfr:
>3D elements are definitely their own separate thing, not being easily overlaid like the rest are, so it makes sense in many ways to have them at the end.

3D Elements are overlayable with other 3D Elements, and at least in matplotlib even 2D Elements, where the z-coordinate defaults to 0.

jlstevens:
> Probably ought to move Chart3D to the end of that list

I concur.

jbednar:
Being able to overlay 2D Elements is cool; that should be mentioned at the start of the Charts3D section.

PR #1197, opened 2017-03-12
Handle document creation directly

philippjfr:
In the latest bokeh dev builds the ``_ModelInDocument`` context manager was moved and changed. Rather than relying on it, we simply create a new Document per plot and detach the Document from all existing models.

### DIFF
--- a/holoviews/plotting/bokeh/renderer.py
+++ b/holoviews/plotting/bokeh/renderer.py
@@ -5,9 +5,9 @@ import param
 from param.parameterized import bothmethod
 
 from bokeh.charts import Chart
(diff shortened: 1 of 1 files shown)


jlstevens:
Seems fine to be explicit and I am assuming this has been tested with latest bokeh dev and previous versions.

Both versions access underscore attributes: ``doc_handler._doc`` in the old version and now `` m._document``. On that basis we are still relying on internal bokeh API but this version should at least work with the next bokeh release.

Merging.

jbednar:
Please file an issue on Bokeh asking that there be a defended API entry point to wrap anything we are relying on.

PR #1198, opened 2017-03-12
Building docs using nbsphinx

philippjfr:
Switched to using nbsphinx to build our docs.

### DIFF
--- a/doc/Examples/Earthquake_Visualization.ipynb
+++ b/doc/Examples/Earthquake_Visualization.ipynb
@@ -350,11 +350,6 @@
   }
  ],
  "metadata": {
-  "kernelspec": {
-   "display_name": "Python 2",
-   "language": "python",
-   "name": "python2"
-  },
   "language_info": {
    "codemirror_mode": {
     "name": "ipython",
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -43,7 +43,9 @@ html_favicon = '_static/favicon.ico'
 # List of patterns, relative to source directory, that match files and
 # directories to ignore when looking for source files.
 exclude_patterns = ['_build', 'test_data', 'reference_data', 'nbpublisher',
-                    'builder']
+                    'builder', '**.ipynb_checkpoints', 'Examples/*.ipynb']
+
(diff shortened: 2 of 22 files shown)


doc/Examples/Earthquake_Visualization.ipynb
```diff
@@ -350,11 +350,6 @@
   }
  ],
  "metadata": {
-  "kernelspec": {
-   "display_name": "Python 2",
-   "language": "python",
-   "name": "python2"
-  },
```
jlstevens: Are all these because nbsphinx needs Python 3?

doc/conf.py
```diff
@@ -22,9 +20,9 @@
 # built documents.
 #
 # The short X.Y version.
-version = setup_args['version']
+version = "1.7"
 # The full version, including alpha/beta/rc tags.
-release = setup_args['version']
+release = "1.7dev7"
```
jlstevens: What is wrong with ``setup_args['version']``? My new process for making dev releases included keeping ``setup.py`` up-to-date. This approach will require updating on every dev-release...

philippjfr: I've had to switch to python3 to build it, which does not allow imports like this. I'm no even quite sure how it's currently working since the top level has no ``__init__.py``

doc/Examples/Earthquake_Visualization.ipynb
philippjfr: Basically yes, lately nb_conda has also been adding specific environment names which I've had to scrub out. It looks like the latest notebook format also adds ``deletable`` and ``editable`` metadata to each cell, so I'll probably have a PR to add those. This is all a giant pain and I'm going to have to write a notebook parsing script that strips all this stuff out.

doc/conf.py
jlstevens: Odd that it works on Python 2 but not Python 3. Seems we will have to find a way to get it working on both as having another place to set the version string is just going to be a pain unless it is done from setup.py.

philippjfr:
Closing for now, while nbsphinx is nice but is not as flexible as a directive.

PR #1202, opened 2017-03-13
Guard against pdf output in IPython display hooks

philippjfr:
Stops the display hooks from rendering to pdf when plotting in the notebook, however exported plots will still be pdf, when ``fig='pdf'`` declared. Fixes https://github.com/ioam/holoviews/issues/519.

### DIFF
--- a/holoviews/ipython/display_hooks.py
+++ b/holoviews/ipython/display_hooks.py
@@ -49,7 +49,12 @@ def render(obj, **kwargs):
         return render_anim(obj)
 
     backend = Store.current_backend
-    return Store.renderers[backend].html(obj, **kwargs)
+    renderer = Store.renderers[backend]
+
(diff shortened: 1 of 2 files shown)


holoviews/ipython/display_hooks.py
```diff
@@ -49,7 +49,11 @@ def render(obj, **kwargs):
         return render_anim(obj)
 
     backend = Store.current_backend
-    return Store.renderers[backend].html(obj, **kwargs)
+    renderer = Store.renderers[backend]
```
jlstevens: Might be worth adding a comment here about why we want to do this.

jlstevens:
Looks good. I made one comment but otherwise is it ready to merge?

philippjfr:

jlstevens:
Merging.

PR #1203, opened 2017-03-13
Dropped 'ls' as matplotlib 'linestyle' alias

philippjfr:
It is not consistently supported as shown in: https://github.com/ioam/holoviews/issues/551


jlstevens:
Annoying as it is an inconsistency in matplotlib and I am sure some people would expect it to work. Oh well, I agree with this change in the circumstances.

jlstevens:
Tests are passing. Merging.

PR #1206, opened 2017-03-13
Fixed batched plot style cycles/palettes

philippjfr:
Color cycles on batched plots were computed incorrectly resulting in colors to be repeated and not matching with other plots.


jlstevens:
Sounds like a bug we would notice pretty quickly! Since when have color cycles been behaving badly with batch mode?

philippjfr:
>Since when have color cycles been behaving badly with batch mode?

Always. It would have been less noticeable if only a few items were changing each time or if there are a lot of them.

jlstevens:
If you rebase against master, the tests should run properly again. Then if it is ready, I can merge.

PR #1209, opened 2017-03-15
Matplotlib aspect fixes

philippjfr:
Fixes https://github.com/ioam/holoviews/issues/199, https://github.com/ioam/holoviews/issues/996 and https://github.com/ioam/holoviews/issues/1144. Plots in Layouts are no longer forced to square, semilog/loglog plots correctly apply aspects, and axis ranges are set after ticks.

### DIFF
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -200,7 +200,8 @@ class ElementPlot(GenericElementPlot, MPLPlot):
                 self._set_axis_limits(axis, element, subplots, ranges)
 
             # Apply aspects
-            self._set_aspect(axis, self.aspect)
+            if self.aspect is not None and self.projection != 'polar':
+                self._set_aspect(axis, self.aspect)
 
         if not subplots and not self.drawn:
             self._finalize_artist(key)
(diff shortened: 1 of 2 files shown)


jlstevens:
Good to see three issues sorted with so little code! I guess the test data will now need to be updated...

philippjfr:
Maybe, don't know yet, made a typo.

philippjfr:
Continuous Coordinates
------------------------

Before:

image

Now:

image

Elements
---------

Before:

image

Now:

image


Sampling Data
---------------

Before:

image


Now:

image

I also still need to make sure that polar plots are forced to be square.

jlstevens:
I think the original reason we tried to force everything to be square is that it made layouts easier.

philippjfr:
>I think the original reason we tried to force everything to be square is that it made layouts easier.

Absolutely and without the layout fix that I added these examples would have really odd spacing I'm sure.

philippjfr:
Ready to merge whenever, rebuilt the test data and PR build is passing.

jbednar:
Excellent; thanks so much for fixing this previously very surprising behavior.  That said, even though these are surely reasonable to consider as bug fixes, they will affect the look and feel of existing notebooks, so they should be mentioned under compatibility changes.  Basically, if someone was relying on Layout to make things square, they will now have to force them to be square, and it will be helpful to mention how to do that in the release notes.

jlstevens:
Sorry if I missed it but that is a good point - any easy way to restore the old behavior if necessary?

philippjfr:
Not really no, except setting ``hv.plotting.mpl.RasterPlot = 'square'``, which will also affect plots not in Layouts.

jlstevens:
Ok, then as Jim said it will need to be mentioned in the release notes.

The pr build is passing, is it ready to merge?

philippjfr:
Yes, ready.

jlstevens:
Great. Merging.

PR #1211, opened 2017-03-16
Fixed bugs in mpl colorbar arrows and Image.range

philippjfr:
In fixing a bug in the colorbars (https://github.com/ioam/holoviews/issues/1109) I found a bug in handling partially bounded Dimension.range settings, e.g.:

```
hv.Image(np.array([[0, 1], [2, 3]]).redim(z=dict(range=(None, 2))).range('z')
```

Would report ``(0, 3)`` when all our other Elements respect explicitly set ranges. I'll probably open another PR to add an argument to the range method to ignore the ``soft_range`` and ``range`` on the Dimension. Knowing the actual data range is perhaps more useful and we should consider returning that by default and only returning ``Dimension.range`` when requested, which all the plotting code could do.

### DIFF
--- a/holoviews/element/raster.py
+++ b/holoviews/element/raster.py
@@ -519,7 +519,7 @@ class Image(SheetCoordinateSystem, Raster):
     def range(self, dim, data_range=True):
         dim_idx = dim if isinstance(dim, int) else self.get_dimension_index(dim)
         dim = self.get_dimension(dim_idx)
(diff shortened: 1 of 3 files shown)


jlstevens:
Tests have passed. Ready to merge?

philippjfr:
Yes, ready to merge, I'll open the related issue now.

Issue #1042, opened 2017-01-09
Holomap.overlay and default dimension


vascotenner:
When no dimensions are specified, a HoloMap takes the dimension 'Default'. However, it is not possible anymore to overlay this dimension.

```python
hv.HoloMap({'B':hv.Curve(([5.1,17.4], (0,0))), 'A':hv.Curve(([17.4,30], (0,0)))}, ).overlay()
```

```
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-9-85ea22825678> in <module>()
----> 1 hv.HoloMap({'B':hv.Curve(([5.1,17.4], (0,0))), 'A':hv.Curve(([17.4,30], (0,0)))}).overlay()

/home/tenner/src/holoviews/holoviews/core/spaces.pyc in overlay(self, dimensions, **kwargs)
     34         if len(dimensions) == self.ndims:
     35             with item_check(False):
---> 36                 return NdOverlay(self, **kwargs).reindex(dimensions)
     37         else:
     38             dims = [d for d in self.kdims if d not in dimensions]

/home/tenner/src/holoviews/holoviews/core/ndmapping.pyc in reindex(self, kdims, force)
    374             kdims = [d for d in old_kdims
    375                      if not len(set(self.dimension_values(d))) == 1]
--> 376         indices = [self.get_dimension_index(el) for el in kdims]
    377 
    378         keys = [tuple(k[i] for i in indices) for k in self.data.keys()]

/home/tenner/src/holoviews/holoviews/core/dimension.pyc in get_dimension_index(self, dim)
    706         except ValueError:
    707             raise Exception("Dimension %s not found in %s." %
--> 708                             (dim, self.__class__.__name__))
    709 
    710 

Exception: Dimension Default not found in NdOverlay.
```

jlstevens:
Thanks for reporting this!

I have mixed feelings about the 'Default' dimension. The idea is to make it easier to use a HoloMap as a dictionary. Generally, I would not rely on it...but I am not really suggesting deprecating it either...

vascotenner:
Quite often, i do not care about the dimension name (or do not want to think about it). As short term solution, I call this dimension 'a', instead of giving it a real name. Hence, I prefer to have a default dimension.

jlstevens:
Glad to know it is of use!

User feedback on such things helps me decide whether I feel they are helpful or a hinderance.

philippjfr:
[...]

PR #1217, opened 2017-03-17
Made sorting on NdMapping optional

philippjfr:
Let's users provide an explicitly sorted NdMapping. Still need to add a warning when a regular dict and sort=False are supplied since sort order is nondeterministic in that case.

~~Edit: Widgets are currently assuming sorted keys and NdElement seems to have been sorting in some cases.~~

Addresses: https://github.com/ioam/holoviews/issues/1216

Fixes: https://github.com/ioam/holoviews/issues/1042

### DIFF
--- a/holoviews/core/ndmapping.py
+++ b/holoviews/core/ndmapping.py
@@ -109,6 +109,10 @@ class MultiDimensionalMapping(Dimensioned):
         self._cached_categorical = any(d.values for d in self.kdims)
 
         if initial_items is None: initial_items = []
+        if type(initial_items) is dict and not sort:
+            raise ValueError('If sort=False the data must define a fixed '
+                             'ordering, please supply a list of items or '
+                             'an OrderedDict, not a regular dictionary.')
         if isinstance(initial_items, tuple):
             self._add_item(initial_items[0], initial_items[1])
         elif not self._check_items:
--- a/holoviews/core/ndmapping.py
+++ b/holoviews/core/ndmapping.py
@@ -49,11 +49,11 @@ class sorted_context(object):
         self.enabled = enabled
 
     def __enter__(self):
-        self._enabled = MultiDimensionalMapping._sorted
-        MultiDimensionalMapping._sorted = self.enabled
+        self._enabled = MultiDimensionalMapping.sort
+        MultiDimensionalMapping.sort = self.enabled
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        MultiDimensionalMapping._sorted = self._enabled
+        MultiDimensionalMapping.sort = self._enabled
 
 
 
@@ -89,9 +89,11 @@ class MultiDimensionalMapping(Dimensioned):
 
     vdims = param.List(default=[], bounds=(0, 0), constant=True)
 
+    sort = param.Boolean(default=True, doc="""
+        Whether the items should be sorted in the constructor.""")
+
     data_type = None          # Optional type checking of elements
     _deep_indexable = False
-    _sorted = True
     _check_items = True
 
     def __init__(self, initial_items=None, **params):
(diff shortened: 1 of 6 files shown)


philippjfr:
Ready to review/merge when tests pass.

holoviews/core/ndmapping.py
```diff
@@ -49,11 +49,11 @@ def __init__(self, enabled):
         self.enabled = enabled
 
     def __enter__(self):
```
jlstevens: The docstring of this context_manager should be updated. As now ``sort=False`` is valid, it should just say it disables sorting regardless of whether the NdMapping has ``sort=True`` or ``sort=False``. 

I also think the line 'Should only be used if values are guaranteed to be sorted before or after the operation is performed.' should just say something else - maybe just that the initial ordering (whatever it is) should be preserved?

holoviews/core/ndmapping.py
```diff
@@ -89,20 +89,17 @@ class MultiDimensionalMapping(Dimensioned):
...
-            map_type = type(initial_items)
-            own_params = self.params()
-            new_params = dict(initial_items.get_param_values(onlychanged=True))
-            if new_params.get('group') == map_type.__name__:
-                new_params.pop('group')
-            params = dict({name: value for name, value in new_params.items()
-                           if name in own_params}, **params)
+        if isinstance(initial_items, MultiDimensionalMapping):
+            params = dict(util.get_param_values(initial_items),
+                          **dict({'sort': self.sort}, **params))
```
jlstevens: Nice to see the simplified constructor!

holoviews/core/ndmapping.py
```diff
@@ -169,11 +170,7 @@ def _add_item(self, dim_vals, data, sort=True, update=True):
 
         for dim, val in valid_vals:
             vals = self._cached_index_values[dim.name]
-            if vals == 'initial': self._cached_index_values[dim.name] = []
-            if not self._instantiated and self.get_dimension(dim).values == 'initial':
-                if val not in vals:
-                    self._cached_index_values[dim.name].append(val)
-            elif vals and val is not None and val not in vals:
```
jlstevens: Was this chunk of logic made redundant by some other change or was it always redundant?

holoviews/core/ndmapping.py
```diff
@@ -283,10 +279,9 @@ def groupby(self, dimensions, container_type=None, group_type=None, **kwargs):
         container_type = container_type if container_type else type(self)
         group_type = group_type if group_type else type(self)
         dimensions = [self.get_dimension(d, strict=True) for d in dimensions]
-        sort = not self._sorted
         with item_check(False):
             return util.ndmapping_groupby(self, dimensions, container_type,
-                                          group_type, sort=sort, **kwargs)
+                                          group_type, sort=True, **kwargs)
```
jlstevens: Probably worth mentioning either in a comment - or maybe even the docstring - that sorting will always be applied.

holoviews/core/ndmapping.py
philippjfr: That's where the bug reported in #1042 was, this was written very early on before we had utilities for this kind of thing.

jlstevens:
Looks good. Most of my comments relate to documentation issues...

I agree a ``ValueError`` is better than a warning when ``sort=False`` and an unsortable dictionary is supplied. It is probably best to completely disallow the nondeterministic case rather than simply warning.

holoviews/core/ndmapping.py
philippjfr: This was my early hacky approach to define a fixed order, I think it has been broken for a long time.

holoviews/core/ndmapping.py
philippjfr: Will do.

philippjfr: Added to the docstring above.

holoviews/core/ndmapping.py
philippjfr: Docstring updated.

jlstevens:
Tests are now fixed (pr build is passing at least) and you've made the docstring changes I suggested. Merging!

PR #1219, opened 2017-03-20
Handle bug in gridded groupby when dropping dimensions

philippjfr:
In certain scenarios dropping dimensions in a gridded groupby would not behave correctly. This fixes the bug and adds some tests.

### DIFF
--- a/holoviews/core/data/iris.py
+++ b/holoviews/core/data/iris.py
@@ -117,7 +117,8 @@ class CubeInterface(GridInterface):
 
     @classmethod
     def validate(cls, dataset):
-        pass
+        if len(dataset.vdims) > 1:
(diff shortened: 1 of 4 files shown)


philippjfr:
Ready.

holoviews/core/data/iris.py
```diff
@@ -117,7 +117,8 @@ def init(cls, eltype, data, kdims, vdims):
 
     @classmethod
     def validate(cls, dataset):
-        pass
+        if len(dataset.vdims) > 1:
```
jlstevens: Always good to have some extra validation!

jlstevens:
Looks good and extra tests are always worthwhile. Merging.

PR #1220, opened 2017-03-21
Bokeh selection colors

philippjfr:
As requested in https://github.com/ioam/holoviews/issues/1139 this adds full control over selection and unselection glyph styles letting you apply custom colors and alphas. Additionally this also adds control of a new feature in bokeh 0.12.5, which is muting glyphs with an interactive legend, this styling can also be controlled via style option but default to a ``alpha=0.2``. This means all our bokeh legends are now clickable.

- [x] Update Bokeh_Backend Tutorial unselection glyph styling examples.
- [x] Tests to check updating all the different glyphs is working correctly.

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -112,7 +112,7 @@ class PointPlot(LegendPlot, ColorbarPlot):
             nvals = len(data[k][-1])
             if 'color' not in elmapping:
                 val = styles[zorder].get('color')
-                elmapping['color'] = 'color'
+                elmapping['color'] = {'field': 'color'}
                 if isinstance(val, tuple):
                     val = rgb2hex(val)
                 data['color'].append([val]*nvals)
@@ -148,7 +148,7 @@ class VectorFieldPlot(ColorbarPlot):
        Whether the lengths will be rescaled to take into account the
        smallest non-zero distance between two vectors.""")
 
-    style_opts = ['color'] + line_properties
+    style_opts = line_properties
     _plot_methods = dict(single='segment')
 
     def _get_lengths(self, element, ranges):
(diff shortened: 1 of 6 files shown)


jlstevens:
Just curious but what is a ``muted_alpha``?

jbednar:
The amount of alpha to use when something is muted?

jlstevens:
> The amount of alpha to use when something is muted?

Ah, now I get it - maybe``deselected_alpha`` would be a more obvious name, but it is what it is.

philippjfr:
>Ah, now I get it - maybedeselected_alpha would be a more obvious name, but it is what it is.

Right before this I had only added support for ``unselected_alpha/color`` because that had seemed like a more sensible name, but it turned out that was really limiting so now I'm just letting the user set ``selection_color/alpha``, ``unselection_color/alpha`` and ``muted_color/alpha``, which is what bokeh calls the different concepts and provides full control.

jbednar:
Sounds good.  Great stuff!

holoviews/plotting/bokeh/chart.py
```diff
@@ -45,7 +46,7 @@ class PointPlot(LegendPlot, ColorbarPlot):
       Function applied to size values before applying scaling,
       to remove values lower than zero.""")
 
-    style_opts = (['cmap', 'palette', 'marker', 'size', 's', 'alpha', 'color',
+    style_opts = (['cmap', 'palette', 'marker', 'size', 's', 'color',
```
jlstevens: I feel ``style_opts`` should be a set - something for 2.0 perhaps? I am assuming 'alpha' was being duplicated...

holoviews/plotting/bokeh/chart.py
```diff
@@ -170,7 +148,7 @@ class VectorFieldPlot(ColorbarPlot):
        Whether the lengths will be rescaled to take into account the
        smallest non-zero distance between two vectors.""")
 
-    style_opts = ['color'] + line_properties
+    style_opts = line_properties
```
jlstevens: Good to see this simplified.

jlstevens:
Looks good! I'll merge once the tests pass.

jlstevens:
Tests are passing. Merging.

PR #1223, opened 2017-03-23
Allow casting from GridImage to QuadMesh

philippjfr:
Useful now and will continue working just fine in the future when the image interface has been added. Let's you do something like this to get hover information from a datashaded plot:

```python
%%opts QuadMesh (fill_alpha=0) [tools=['hover']]
points = hv.Points(np.random.multivariate_normal((0,0), [[0.1, 0.1], [0.1, 1.0]], (1000000,)))
datashade(points, width=400, height=400) * hv.util.Dynamic(aggregate(points, width=100, height=100),
                                                           operation=lambda x: hv.QuadMesh(x))
```

hover

### DIFF
--- a/holoviews/element/raster.py
+++ b/holoviews/element/raster.py
@@ -243,8 +243,13 @@ class QuadMesh(Raster):
     def depth(self): return 1
 
     def _process_data(self, data):
-        data = tuple(np.array(el) for el in data)
-        x, y, zarray = data
+        if isinstance(data, GridImage):
+            x = data.dimension_values(0, expanded=False)
+            y = data.dimension_values(1, expanded=False)
+            zarray = data.dimension_values(2, flat=False)
+        else:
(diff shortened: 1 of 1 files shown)


jbednar:
Looks good to me and seems safe; happy to merge when tests pass.

jlstevens:
Looks good to me as well. Merging.

stephanecollot:
Hi,
That looks great!
Unfortunately I didn't succeed to make it work in my notebook. Could you show us the imports that are loading 'datashade' and 'aggregate'.
Also, it is possible to make it work for multiple timeseries.

Thanks for your help

jbednar:
You can see how to use it in https://anaconda.org/jbednar/holoviews_datashader , which shows all required imports, including how to work with multiple plots.  This all requires ``conda install -c ioam holoviews``.

PR #1225, opened 2017-03-23
Apply matplotlib rc parameters correctly throughout

philippjfr:
I've long struggled with the correct way to apply rcparams in matplotlib, I now think a decorator which is applied during axis/figure creation, for all ``initialize_plot`` and ``update_frame`` calls and during rendering is the best approach.

Fixes https://github.com/ioam/holoviews/issues/1222

### DIFF
--- a/holoviews/plotting/mpl/annotation.py
+++ b/holoviews/plotting/mpl/annotation.py
@@ -4,6 +4,7 @@ from matplotlib import patches as patches
 from ...core.util import match_spec
 from ...core.options import abbreviated_exception
 from .element import ElementPlot
+from .plot import mpl_rc_context
 
 
 class AnnotationPlot(ElementPlot):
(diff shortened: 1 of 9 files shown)


philippjfr:
Ready for review.

jlstevens:
Seems like a sensible approach and I am happy with it as long as it works (preferably better than before).

My only comment is that I think ``mpl_context`` should be called ``mpl_rc_context`` to be explicit.

jlstevens:
I suppose my only other comment is that it might be nice to test - if it is easy enough to set up such tests.

philippjfr:
Ready now, pretty sure the coverage report is false, when I click through to it, it doesn't actually tell me which files now have reduced coverage.

jlstevens:
> Ready now, pretty sure the coverage report is false

Yeah, a -1.4% coverage change is surprising.

Anyway, looks good and I'll now merge.

PR #1227, opened 2017-03-24
Implemented a dynamic hist method

philippjfr:
One of the last methods to implement to make the DynamicMap fully lazy as outlined in https://github.com/ioam/holoviews/issues/422.


jlstevens:
Looks good and the tests are passing (not that this is currently tested). It might be worth adding tests later but I am happy to merge right now.

PR #1228, opened 2017-03-24
Allow ElementOperation to process NdLayout types

philippjfr:
Partial fix for https://github.com/ioam/holoviews/issues/1107, while we haven't addressed the issue about (Nd)Overlays, this is a straightforward addition that let's ElementOperations process NdLayouts just as it already handles GridSpaces.

### DIFF
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -11,7 +11,7 @@ except:
     pass
 
 from .dimension import ViewableElement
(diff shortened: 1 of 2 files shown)


jlstevens:
Looks good and I see now reason why ``NdLayout`` should be excluded from operations.

I do think it might be worth adding a simple ``NdLayout`` operation in the unit tests to make sure the plumbing works as expected.

philippjfr:
Ready to merge.

Issue #1186, opened 2017-03-09
Allow banner and logo to be suppressed?


jbednar:
Right now, holoviews acts like Bokeh and prints a banner and logo when the notebook_extension is enabled:

![image](https://cloud.githubusercontent.com/assets/1695496/23765306/674d9318-04c6-11e7-9356-ba5d24019e9d.png)

This is helpful for debugging (if it isn't there, things aren't working!), but (a) things almost always work properly, making it not helpful for me in practice, and (b) it disrupts the flow of the page when I'm converting notebooks into web pages for documentation or examples.

Can there be a simple, concise way to turn off the logo and header?  ``hv.notebook_extension(quiet=True)`` would work as a syntax, but that's actually still a bit distracting, because then every notebook will make people think about what "quiet" means, rather than what this particular notebook is about.  I don't have any less-conspicuous way to suggest, though.

jbednar:
I guess I personally would rather see the message *gone* by default, with an option needing to be provided to enable it, as in ``hv.notebook_extension(banner=True)``.

jlstevens:
I agree with adding an option to suppress visual output when loading an extension.

> I guess I personally would rather see the message gone by default, with an option needing to be provided to enable it, as in hv.notebook_extension(banner=True).

The original reason we wanted the banner by default is to show that the cell contains invisible Javascript output - if you delete the cell things will break (HoloViewsJS is gone). I think this is still try so I would prefer ``banner=True`` by default with ``banner=False`` when it is needed (e.g for dashboards).

jbednar:
I don't understand the concern about deleting the cell -- if you delete the cell, then the notebook won't be runnable anyway.  So when would that be an issue in practice?  Try counting how many times you have seen that banner message (here or in Bokeh), compared to how many times it was relevant to an action you needed to take.  In my case, those counts are about 10 million and 0, respectively, which makes me want to make it go away.

jlstevens:
> if you delete the cell, then the notebook won't be runnable anyway

It won't be *re-runnable* but deleting such cells can break the currently running notebook without a kernel restart. These cells are special due to their hidden contents, and not just their code.

> Try counting how many times you have seen that banner message (here or in Bokeh), compared to how many times it was relevant to an action you needed to take or a decision you needed to make. In my case, those counts are about 10 million and 0, respectively, which makes me want to make it go away.

[...]

PR #1231, opened 2017-03-28
Replaced extension logo parameter with banner toggle

jlstevens:
This simple PR replaces the ``logo`` option of the notebook extension (which only toggled the logo) with a ``banner`` switch with can suppress both the logo and the loading message. Addresses #1186.

We'll need to mention the replacement of the ``logo`` parameter in the CHANGELOG but I don't think it is a big deal - the new behavior simply extends the old behavior and I don't think the ``logo`` parameter was used much anyway.

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -117,7 +117,8 @@ class notebook_extension(param.ParameterizedFunction):
 
     css = param.String(default='', doc="Optional CSS rule set to apply to the notebook.")
 
-    logo = param.Boolean(default=True, doc="Toggles display of HoloViews logo")
+    banner = param.Boolean(default=True, doc="""
+        Toggles display of HoloViews logo and loading message.""")
 
     inline = param.Boolean(default=True, doc="""Whether to inline JS and CSS resources,
         if disabled resources are loaded from CDN if one is available.""")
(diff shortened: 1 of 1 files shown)


jbednar:
Thanks.  Will this still print a blank line, or will it avoid printing at all (preferred)?

jlstevens:
> Thanks. Will this still print a blank line, or will it avoid printing at all (preferred)?

I noticed a tiny it more whitespace than normal (i.e more than when you just enter ``None`` into a cell) but didn't manage to remove it, even after simplifying the template to just JS. Here is what I mean:

image

I assume this odd effect happens when there is JS code 'displayed' even though nothing should be visible (including that slightly larger white gap).

jbednar:
I'm not so worried about what happens if the `notebook_extension()` call is the last line in the cell, but what happens if I then put some useful things with real output in the same cell (e.g. "2.0") -- will the formatting then be messed up?

philippjfr:
>I'm not so worried about what happens if the notebook_extension() call is the last line in the cell, but what happens if I then put some useful things with real output at the end of that same cell (e.g. "2.0") -- will the formatting then be messed up?

No, it'll behave as now and just put that bit of whitespace above your figure.

philippjfr:
Ready to merge?

jlstevens:
Yup, I think so.

jbednar:
Thanks!

Issue #1224, opened 2017-03-23
Refactor stream renaming and stream preprocessors


jlstevens:
Currently ``Stream`` classes have a ``preprocessors`` argument that allow for renaming and other transformations of stream parameters. This was an initial concept we introduced, primarily because we realized that stream names might clash and wanted to introduce a mechanism that could resolve these clashes.

Now we have more experience using the streams system, it no longer seems the appropriate level at which to introduce preprocessors, which we do know can  be a useful concept in practice e.g we've found one example needing a preprocessor relating to datashader.

As the callback that generates elements lives on ``DynamicMap``, this now seems to be the place to apply such transformations. The proposal here is two-fold: Streams should handle clashes only and preprocessors on ``DynamicMap`` should handle transformations, once there are no name clashes.

## Resolving name clashes

Streams should now lose the ``preprocessors`` argument and gain a ``rename`` method instead:

```python
coords = PositionXY().rename(x='long', y='lat')
```

As this should be possible via the constructor, this should be equivalent to:

```python
coords = PositionXY(rename={'x':'long', 'y':'lat'})
```

Note that I think ``rename`` should give you a new stream instance which is *identical* except for the renamed values in the streams ``contents`` property.

As there can be no clashes between stream parameters on a single stream, so you only really need to use the ``rename`` in the ``streams`` argument to DynamicMap:

```python
DynamicMap(my_callback, streams=[Tap(), DoubleTap().rename(x='dtap_x', y='dtap_y')])
```

or equivalently

```python
DynamicMap(my_callback, streams=[Tap(), DoubleTap(rename=dict(x='dtap_x', y='dtap_y'))])
```

This resolves the fact that both ``Tap`` and ``DoubleTap`` have ``x`` and ``y`` stream parameters by default.

Note that ``DynamicMap`` can check for clashes in the constructor - I suspect we do this already. I am also wondering if ``map`` might be better as it is shorter than ``rename`` but that might be confusing as we have ``map`` methods elsewhere.  Note that ``remap`` is also shorter than ``rename``...

## Preprocessors

Now that stream parameters will be disambiguated, we can have a single preprocessor at the ``DynamicMap`` level that simply maps a dictionary of kwargs to a new dictionary of kwargs. We decided to have a single preprocessor instead of a list as:

1. This is simpler.
2. You can easily chain preprocessors yourself e.g

```python
preprocessor=lambda kws: foo(bar(baz(kws)))
```

[...]

PR #1232, opened 2017-03-29
Preprocessor refactor

jlstevens:
This PR addressed #1224 by greatly simplifying how stream parameters can be renamed and how stream preprocessing is implemented (when necessary).

The ``rename`` suggestion from that issue has been implemented exactly as suggested but after discussion with @philippjfr and @jbednar, it was decided that we don't really need explicit preprocessor hooks. Instead, the streams that require it can implement the ``preprocess`` method (which occurs before stream renaming) so transform the stream values as necessary.

I would still like a better name than ``rename`` and although I personally like ``remap`` more, I know this isn't the majority opinion. Consider

```python
PositionXY(rename={'x':'x1', 'y':'y1'}, x=2)
```

and 

```python
PositionXY(remap={'x':'x1', 'y':'y1'}, x=2)
```

I feel the latter conveys the idea that ``x`` will be called ``x1`` on the output better. Any other suggestions?

Action items:

- [x] Remove old preprocessor system
- [x] Offer new simplified means of preprocessing streams
- [x]  Implement ``rename`` constructor argument and method
- [x] Update ``PlotSize`` stream with a ``preprocess`` method
- [x] Add unit tests
- [x] Update tutorials **Edit: postponed till a later PR**

### DIFF
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -10,17 +10,6 @@ from collections import defaultdict
 from .core import util
 
 
-class Preprocessor(param.Parameterized):
-    """
-    A Preprocessor is a callable that takes a dictionary as an argument
-    and returns a dictionary. Where possible, Preprocessors should have
-    valid reprs that can be evaluated.
-    """
-
-    def __call__(self, params):
-        return params
-
-
 class Stream(param.Parameterized):
     """
     A Stream is simply a parameterized object with parameters that
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -127,28 +127,24 @@ class Stream(param.Parameterized):
         self.registry[id(source)].append(self)
 
 
-    def preprocess(self, param_values):
+    def transform(self):
         """
         Method that can be overwritten by subclasses to process the
-        parameter values before the renaming step. Allows transformation
-        of parameter values as a function of other parameter values.
+        parameter values before renaming is applied. Returns a
+        dictionary of transformed parameters.
         """
-        return param_values
+        return {}
 
     @property
     def contents(self):
         filtered = {k:v for k,v in self.get_param_values() if k!= 'name' }
-        processed = self.preprocess(filtered)
-        return {self._rename.get(k,k):v for (k,v) in processed.items()}
+        return {self._rename.get(k,k):v for (k,v) in filtered.items()}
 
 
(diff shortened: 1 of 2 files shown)


jlstevens:
After some discussion, we decided to rename the ``preprocess`` method to ``transform`` and have it set the transformed parameters in the ``update`` method. The ``PlotSize`` stream now implements scaling based on this approach.

jlstevens:
I will update the tutorial material once I've finished going through ``DynamicMap``. The tests should pass and this PR is ready for review.

holoviews/streams.py
```diff
@@ -115,7 +61,7 @@ def trigger(cls, streams):
             stream.deactivate()
 
 
-    def __init__(self, preprocessors=[], source=None, subscribers=[],
+    def __init__(self, rename={}, source=None, subscribers=[],
```
philippjfr: Did none of the docstrings mention preprocessors? Either way docstrings should now mention the rename argument.

holoviews/streams.py
```diff
@@ -143,6 +89,23 @@ def __init__(self, preprocessors=[], source=None, subscribers=[],
...
+        param_names = [k for k in self.params().keys() if k != 'name']
+        for k,v in mapping.items():
+            if k not in param_names:
+                raise KeyError('Cannot rename %r as it is not a stream parameter' % k)
+            if v in param_names:
+                raise KeyError('Cannot rename to %r as it clashes with a '
+                               'stream parameter of the same name' % v)
+        return mapping
+
+    def rename(self, **mapping):
```
philippjfr: Add docstring.

philippjfr:
Looks good, just update/add some docstrings.

holoviews/streams.py
jlstevens: Fixed in 65131a2

holoviews/streams.py
jlstevens: Fixed in 65131a2

jlstevens:
Updated as suggested. Tests are passing for Python 2.7 and the Python 3 tests should pass soon,

jlstevens:
The pr build just passed on the commit before I added the docstrings. I think it is ready to merge.

philippjfr:
Merging.

PR #1235, opened 2017-03-29
Simplified how Event subscribers are handled

jlstevens:
This PR offers ``clear`` and ``add_subscriber`` methods to ``Event`` and does away with ``_hidden_subscribers``. The ``subscribers`` attribute is now in fact a read-only property which means the correct way to add subscribers is with the ``add_subscribers`` method.

Outstanding items to address:

- [x] Clarify whether ``DynamicMap.event`` and ``Stream.update`` methods support names before or after renaming.
- [x] Add more unit tests
- [x] Add docstrings

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -598,9 +598,14 @@ class DynamicMap(HoloMap):
         (renamed as appropriate) to be updated in an event.
         """
         stream_params = set(util.stream_parameters(self.streams))
+        for k in stream_params - set(kwargs.keys()):
+            raise KeyError('Key %r does not correspond to any stream parameter')
+
         updated_streams = []
         for stream in self.streams:
-            rkwargs = util.rename_stream_kwargs(stream, kwargs, reverse=True)
+            applicable_kws = {k:v for k,v in kwargs.items()
+                              if k in set(stream.contents.keys())}
+            rkwargs = util.rename_stream_kwargs(stream, applicable_kws, reverse=True)
             stream.update(**dict(rkwargs, trigger=False))
             updated_streams.append(stream)
 
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -594,16 +594,15 @@ class DynamicMap(HoloMap):
 
     def event(self, trigger=True, **kwargs):
         """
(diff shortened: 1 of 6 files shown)


jlstevens:
@philippjfr I think it is ready for review - I'm expecting the tests to pass.

philippjfr:
Addresses https://github.com/ioam/holoviews/issues/1168

holoviews/core/spaces.py
```diff
@@ -594,16 +594,15 @@ def _validate_key(self, key):
...
         """
         stream_params = set(util.stream_parameters(self.streams))
         updated_streams = []
         for stream in self.streams:
-            overlap = set(stream.params().keys()) & stream_params & set(kwargs.keys())
-            if overlap:
-                stream.update(**dict({k:kwargs[k] for k in overlap}, trigger=False))
-                updated_streams.append(stream)
+            rkwargs = util.rename_stream_kwargs(stream, kwargs, reverse=True)
+            stream.update(**dict(rkwargs, trigger=False))
```
philippjfr: Maybe I'm not following the logic correctly in the utilities but won't this try to send the supplied kwargs to all the streams even though they may not apply? That's what the overlap bit was for I assume.

jlstevens: Thanks, that has helped me realize what we need:

* An initial loop over ``set(kwargs.keys()) - set(all_streams)`` to raise an exception for anything in that difference set: these are kwargs that cannot be applicable to any stream. Previously invalid kwargs were silently ignored and I think this is bad behavior unless you have reasons to say otherwise.

* ``util.rename_stream_kwargs(stream, {k:v for k,v in kwargs.items() if k in set(stream.contents.keys())}, reverse=True)`` to filter the kwargs to the applicable ones.

Do you agree that this is the right approach?

philippjfr: Yes, that seems right.

jlstevens: Fixed in 79b7de1

philippjfr:
Looks good now, will merge when tests pass.

jlstevens:
Tests now passing except for one transient error.

philippjfr:
Will merge, since the PR build has passed.

PR #1239, opened 2017-04-03
Remove bokeh matplotlib compatibility

philippjfr:
As suggested in https://github.com/ioam/holoviews/issues/1218 this removes the BokehMPLWrapper plot class and support for Distribution, Bivariate, and TimeSeries Elements. I was initially hesitant to remove this entirely in 1.7 but with the amount of warnings these currently generate and other restrictions on these plot types, I think just removing them is the correct thing to do. Hopefully we can readd Distribution and Bivariate support in 2.0 as part of the statistical elements module (https://github.com/ioam/holoviews/issues/1200).

### DIFF
--- a/doc/Tutorials/Bokeh_Backend.ipynb
+++ b/doc/Tutorials/Bokeh_Backend.ipynb
@@ -311,35 +311,8 @@
   {
    "cell_type": "markdown",
    "metadata": {
-    "focus": false,
-    "id": "85e1ed20-1e12-4194-822e-c2fe0810ce41"
-   },
-   "source": [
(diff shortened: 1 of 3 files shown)


jlstevens:
Looks good.

I share your worries about backwards compatibility but ``BokehMPLWrapper`` is not meant to be - you can tell from the name that if nothing else, you are linking two dependencies that ought to be separate.

As we are removing functionality, I do think we definitely ought to get the corresponding statistical elements into 2.0. This is something we should mention in the CHANGELOG to try and re-assure anyone using these things that this functionality will be restored in a better way soon.

Merging.

PR #1240, opened 2017-04-03
Implemented DynamicMap.map method

philippjfr:
Implements the final ``DynamicMap`` method which was not yet lazy as outlined in https://github.com/ioam/holoviews/issues/422.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -817,6 +817,22 @@ class DynamicMap(HoloMap):
         self.data[key] = val
 
 
+    def map(self, map_fn, specs=None, clone=True):
+        """
+        Recursively replaces elements using a map function when the
+        specification applies. Extends regular map with functionality
+        to dynamically apply functions.
+        """
+        deep_mapped = super(DynamicMap, self).map(map_fn, specs, clone)
+        if isinstance(deep_mapped, type(self)):
+            from ..util import Dynamic
+            def apply_map(obj):
+                return obj.map(map_fn, specs, clone)
+            dmap = Dynamic(deep_mapped, shared_data=True, operation=apply_map)
+            return dmap
+        return deep_mapped
+
+
     def relabel(self, label=None, group=None, depth=1):
         """
         Assign a new label and/or group to an existing LabelledData
--- a/tests/testdynamic.py
+++ b/tests/testdynamic.py
@@ -1,7 +1,7 @@
 from collections import deque
 
 import numpy as np
-from holoviews import Dimension, NdLayout, GridSpace
+from holoviews import Dimension, NdLayout, GridSpace, Layout
 from holoviews.core.spaces import DynamicMap, HoloMap, Callable
 from holoviews.element import Image, Scatter, Curve, Text
 from holoviews.streams import PositionXY, PositionX, PositionY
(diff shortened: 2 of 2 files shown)


philippjfr:
Ready to review. ``get_dynamic_item`` is still not ideal, but I much prefer not using ``.map`` for this, since I think .map is very useful functionality for ``DynamicMap``, since it flexibly let's you apply operations and function to specific objects returned by the ``DynamicMap``.

holoviews/core/util.py
```diff
@@ -1163,22 +1163,26 @@ def arglexsort(arrays):
...
-        map_obj.traverse(lambda x: x[()], ['DynamicMap'])
-        return key, map_obj.map(lambda x: x.last, ['DynamicMap'])
-    elif isinstance(key, tuple):
-        dims = {d.name: k for d, k in zip(dimensions, key)
-                if d in map_obj.kdims}
-        key = tuple(dims.get(d.name) for d in map_obj.kdims)
-        el = map_obj.select(['DynamicMap', 'HoloMap'], **dims)
+    Traverses the supplied object selecting the requested key on
+    all HoloMap and DynamicMap objects.
+    """
```
jlstevens: It would be nice to have a doctest style example here (even if it isn't a working one). Just be sure that it doesn't get picked up as one and then fail!

I understand the problem with a working doctest comes down to circular imports.

holoviews/core/util.py
```diff
@@ -1163,22 +1163,26 @@ def arglexsort(arrays):
...
+            el = map_obj.select(map_specs, **dims)
+    elif map_obj._deep_indexable and not any(map_obj.matches(spec) for spec in overlay_specs):
+        el = map_obj.clone(shared_data=False)
+        for k, v in map_obj.data.items():
+            _, deep_obj = get_dynamic_item(v, dimensions, key)
+            el[k] = deep_obj
     else:
-        el = None
+        el = map_obj
     return key, el
```
jlstevens: The reason why key is also returned should be explained in the docstring. Why isn't this the input key?

holoviews/core/util.py
jlstevens: Is ``map_obj`` mutated, cloned or is the result potentially a little bit of both?

holoviews/core/util.py
```diff
@@ -1163,22 +1163,26 @@ def arglexsort(arrays):
...
+    overlay_specs = ('Overlay', 'NdOverlay')
+    if any(map_obj.matches(spec) for spec in map_specs):
+        if key == ():
+            el = map_obj[key]
+        else:
+            dims = {d.name: k for d, k in zip(dimensions, key)
+                    if d in map_obj.kdims}
+            key = tuple(dims.get(d.name) for d in map_obj.kdims)
+            el = map_obj.select(map_specs, **dims)
+    elif map_obj._deep_indexable and not any(map_obj.matches(spec) for spec in overlay_specs):
```
jlstevens: I don't understand this condition whereas the other ones do make sense to me. Maybe a comment to explain what this branch is for?

philippjfr: Hmm, there might be a better way to express it, it's recurses into all container objects excluding (Nd)Overlay types.

jlstevens: Ah, I see. I guess this is because you can apply ``select`` at the overlay level and you don't need to apply it to the elements here.

holoviews/core/spaces.py
```diff
@@ -759,6 +759,27 @@ def _cache(self, key, val):
...
+        """
+        if specs and not isinstance(specs, list): specs = [specs]
+        applies = specs is None or any(self.matches(spec) for spec in specs)
+
+        if applies:
+            deep_mapped = map_fn(self)
+        else:
+            deep_mapped = self.clone() if clone else self
+        for k, v in deep_mapped.data.items():
+            deep_mapped[k] = v.map(map_fn, specs, clone)
```
jlstevens: How do we know it is safe to use ``deep_mapped.data.items()``? 

If it is ``self`` or ``self.clone()`` it should be fine but what if ``map_fn`` returns something where this isn't valid?

philippjfr: It seems in the non-dynamic version we first map the items and then apply it to itself if it applies, so I'll do that here.

holoviews/core/spaces.py
```diff
@@ -759,6 +759,27 @@ def _cache(self, key, val):
...
+
+        from ..util import Dynamic
+        def dynamic_map(obj):
```
jlstevens: I would just call this ``map_operation`` as ``dynamic_map`` could be confusing!

philippjfr:
Managed to simplify this a lot.

holoviews/core/spaces.py
```diff
@@ -817,6 +817,22 @@ def _cache(self, key, val):
         self.data[key] = val
 
 
+    def map(self, map_fn, specs=None, clone=True):
+        """
+        Recursively replaces elements using a map function when the
+        specification applies. Extends regular map with functionality
+        to dynamically apply functions.
+        """
```
jlstevens: Wow! That really was a massive simplification. I like it!

I just hope it works as expected and doesn't become complicated again. Perhaps a few more unit tests would help?

philippjfr: Added one more test to check the clone=False condition, I generally prefer not adding multiple unit tests that just test end up testing the same code paths and Dimensioned.map is already tested.

jlstevens: Ok. As long as you feel ``map`` is tested on sufficiently complex examples somewhere, I don't mind.

philippjfr:
Ready to merge.

jlstevens:
Great! Tests are passing so I'll merge as requested.

PR #1241, opened 2017-04-03
Handle batched styles consistently

philippjfr:
The handling of batched style options was handled differently on all the different plots. As I outlined in https://github.com/ioam/holoviews/issues/795 this should be handled consistently and this PR makes that happen.

### DIFF
--- a/holoviews/plotting/bokeh/util.py
+++ b/holoviews/plotting/bokeh/util.py
@@ -547,3 +547,40 @@ def get_tab_title(key, frame, overlay):
         title = ' | '.join([d.pprint_value_string(k) for d, k in
                             zip(overlay.kdims, key)])
     return title
+
+
+def expand_batched_style(style, opts, data, mapping):
+    """
+    Expands styles applied to a batched plot by iterating over the
+    supplied list of style options and any options found in the supplied
+    style dictionary to the ColumnDataSource data and mapping.
+    """
+    nvals = len(list(data.values()[0])[-1])
+    for opt in opts:
+        if 'color' in opt:
+            alias = 'color'
+        elif 'alpha' in opt:
+            alias = 'alpha'
+        else:
+            alias = None
+        if opt not in style:
+            continue
+        elif opt == alias:
+            if alias in mapping:
+                continue
+            elif 'line_'+alias in mapping:
+                if 'fill_'+alias not in opts:
+                    continue
+                opt = 'fill_'+alias
+                val = style[alias]
+            elif 'fill_'+alias in mapping:
+                opt = 'line_'+alias
+                val = style[alias]
+            else:
+                val = style[alias]
+        else:
+            val = style[opt]
+        mapping[opt] = {'field': opt}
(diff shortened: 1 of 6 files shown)


jlstevens:
Sounds like a good idea. Do you think it is possible to set up tests to try and enforce consistency between batched and non-batched?

philippjfr:
>Do you think it is possible to set up tests to try and enforce consistency between batched and non-batched?

Yes, I'll add some of those now. That said there's certain cases that don't work yet, which I'll list here once I'm done.

philippjfr:
As far as I can tell there's only one style option that doesn't work in batched mode, which is ``marker``. Definitely annoying but a huge improvement because we actually only handled colors correctly before.

philippjfr:
While the lack of support for ``marker`` is annoying and is something I will file with bokeh, every single other option now works correctly in batched mode including ``selection`` and ``nonselection`` colors, which I'm pretty happy about. Also added some tests, ready to review/merge.

holoviews/plotting/bokeh/util.py
```diff
@@ -547,3 +547,46 @@ def get_tab_title(key, frame, overlay):
...
+
+def expand_batched_style(style, opts, data, mapping, nvals=None, multiple=False):
+    """
+    Computes styles applied to a batched plot by iterating over the
+    supplied list of style options and expanding any options found in
+    the supplied style dictionary to the ColumnDataSource data and
+    mapping. Supplying nvals adds the style entry multiple times, added
+    as an array by default. When multiple is active multiple scalar values
+    will be added.
+    """
```
jlstevens: Is there a simple illustrative example you could put in the docstring? I have trouble imagining what it does from this description alone...

philippjfr: Yeah, it's fairly complex I'll try to come up with some examples, since it supports three modes of expanding styles corresponding to Path, Points, and Curve types.

jlstevens: Actually, in that case it would just be good to have a nice set of unit tests, starting simple and getting more complicated, covering all three modes. This is one of those cases where I feel unit tests can *almost* substitute docstrings...

philippjfr: There are unit tests, you can comment on their clarity though.

philippjfr: They are more integration tests, not testing the utility but what actually ends up on the datasource.

jlstevens: > They are more integration tests, not testing the utility but what actually ends up on the datasource.

Sure. These are definitely useful and implicitly tests ``expand_batched_style``. 

What I'm saying is that explicit  unit tests for ``expand_batched_style`` wouldn't hurt and would also help give some concrete examples of what it does.

philippjfr:
I simplified the utility a fair bit and added some explicit unit tests. Ready to merge when tests passing.

jlstevens:
Thanks for the new unit tests and I'm glad you managed to simplify the utility. Tests are passing.

Merging.

Issue #1188, opened 2017-03-10
Support .collate for DynamicMaps?


jbednar:
The example on the holoviews.org home page shows how to use `.collate()` to turn the illegal "HoloMap of Layouts" construction into the displayable "Layout of HoloMaps", which is very convenient in practice.  However, `.collate()` does not seem to be supported for DynamicMaps.  

Here's an example of what I'd want to use that for:

```python
import numpy as np
import holoviews as hv
import imagen as ig
import imagen.transferfn as tf

from imagen.image import FileImage
from imagen.transferfn.sheet_tf import Convolve

hv.notebook_extension()

%opts Image (cmap="gray") {+axiswise}

FileImage.filename = "/Users/jbednar/imagen/doc/images/ellen_arthur.pgm"

kdims = [hv.Dimension('size',                   range=(0.1, 2)),
         hv.Dimension('positive_size',          range=(0.1, 2)),
         hv.Dimension('negative_size',          range=(0.3, 4)),
         hv.Dimension('positive_aspect_ratio',  range=(1.0,10)),
         hv.Dimension('negative_aspect_ratio',  range=(1.0,10))]
```

Right now, I can easily view one static image along with a slider-controlled dynamic map, but I have to pick a single HoloViews object to return from the callback:

```python
def filtered_images(size,positive_size,negative_size,positive_aspect_ratio,negative_aspect_ratio):
    convolver = Convolve(kernel_pattern=ig.DifferenceOfGaussians(**locals()))
    return FileImage(output_fns=[convolver])[:]

FileImage()[:] + hv.DynamicMap(filtered_images, kdims=kdims)
```

![image](https://cloud.githubusercontent.com/assets/1695496/23803067/efcb9c30-057a-11e7-8389-959e7318bc87.png)

```python
def filtered_images(size,positive_size,negative_size,positive_aspect_ratio,negative_aspect_ratio):
    convolver = Convolve(kernel_pattern=ig.DifferenceOfGaussians(**locals()))
    return convolver.kernel_pattern[:]

FileImage()[:] + hv.DynamicMap(filtered_images, kdims=kdims)
```

![image](https://cloud.githubusercontent.com/assets/1695496/23803077/fe95d668-057a-11e7-8b42-0619cee6a0d9.png)

But what I want is to make one single callback that returns a Layout of both the two dynamic plots above:

```python
def filtered_images(size,positive_size,negative_size,positive_aspect_ratio,negative_aspect_ratio):
[...]

PR #1243, opened 2017-04-04
Dynamic collation

philippjfr:
Implements dynamic collation as requested and described in #1188. The user can declare a stream on a ``DynamicMap`` returning a ``(Nd)Layout``, and then specify which item in the Layout should be used as the source for the stream in the collate call by either supplying a list of streams for each item or a dictionary where the key corresponds to the index of the item in the Layout. Here's a small example:

```python
%%opts Layout [shared_axes=False]

def test(x_range, y_range):
    return hv.Image(np.random.rand(10,10)) + hv.Bounds((x_range[0], y_range[0], x_range[1], y_range[1]))

stream = hv.streams.RangeXY(x_range=(0,-.5), y_range=(0,-0.5))
dmap = hv.DynamicMap(test, streams=[stream], kdims=[])

dmap.collate(streams={0: [stream]})
# OR dmap.collate(streams=[[stream]])

```

Here the RangeXY stream will be attached to the Image and the Bounds in the second subplot will reflect the zoom ranges of the first plot. Leaving out the ``streams`` declaration on the collate call will mean that the ``Stream`` gets ranges from both subplots.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -138,7 +138,7 @@ class HoloMap(UniformNdMapping, Overlayable):
             except KeyError:
                 pass
             return Overlay(layers)
-        callback = Callable(callable_function=dynamic_mul, inputs=[self, other])
+        callback = Callable(dynamic_mul, inputs=[self, other])
         if map_obj:
             return map_obj.clone(callback=callback, shared_data=False,
                                  kdims=dimensions, streams=[])
@@ -205,8 +205,7 @@ class HoloMap(UniformNdMapping, Overlayable):
                 def dynamic_mul(*args, **kwargs):
                     element = self[args]
                     return element * other
-                callback = Callable(callable_function=dynamic_mul,
-                                    inputs=[self, other])
+                callback = Callable(dynamic_mul, inputs=[self, other])
                 return self.clone(shared_data=False, callback=callback,
                                   streams=[])
             items = [(k, v * other) for (k, v) in self.data.items()]
@@ -403,26 +402,33 @@ class Callable(param.Parameterized):
     state of all streams on its inputs, to avoid calling the function
     unnecessarily.
 
-    A Callable may also specify a stream_mapping which allows
-    specifying which objects to attached linked streams to on
-    callbacks which return composite objects like (Nd)Layout and
-    GridSpace objects. The mapping should map between an integer index
-    or a type[.group][.label] specification and lists of streams
-    matching the object.
+    A Callable may also specify a stream_mapping which specifies the
+    objects that are associated with interactive (i.e linked) streams
+    when composite objects such as Layouts are returned from the
+    callback. This is required for building interactive, linked
+    visualizations (for the backends that support them) when returning
+    Layouts, NdLayouts or GridSpace objects.
+
+    The mapping should map from an appropriate key to a list of
+    streams associated with the selected object. The appropriate key
+    may be a type[.group][.label] specification for Layouts, an
+    integer index or a suitable NdLayout/GridSpace key. For more
+    information see the DynamicMap tutorial at holoviews.org.
     """
 
-    callable_function = param.Callable(default=lambda x: x, doc="""
+    callable = param.Callable(default=None, doc="""
          The callable function being wrapped.""")
 
     inputs = param.List(default=[], doc="""
          The list of inputs the callable function is wrapping.""")
 
-    def __init__(self, callable_function=None, stream_mapping={}, **params):
-        if callable_function is not None:
-            params['callable_function'] = callable_function
-        super(Callable, self).__init__(**params)
+    stream_mapping = param.Dict(default={}, doc="""
+         Defines how streams should be mapped to objects returned by
+         the Callable, e.g. when it returns a Layout.""")
+
+    def __init__(self, callable, **params):
+        super(Callable, self).__init__(callable=callable, **params)
         self._memoized = {}
-        self.stream_mapping = stream_mapping
 
     def __call__(self, *args, **kwargs):
         inputs = [i for i in self.inputs if isinstance(i, DynamicMap)]
@@ -430,11 +436,10 @@ class Callable(param.Parameterized):
         values = tuple(tuple(sorted(s.contents.items())) for s in streams)
         key = args + tuple(sorted(kwargs.items())) + values
 
-
         hashed_key = util.deephash(key)
         ret = self._memoized.get(hashed_key, None)
         if hashed_key and ret is None:
-            ret = self.callable_function(*args, **kwargs)
+            ret = self.callable(*args, **kwargs)
             self._memoized = {hashed_key : ret}
         return ret
 
@@ -492,7 +497,7 @@ class DynamicMap(HoloMap):
 
     def __init__(self, callback, initial_items=None, **params):
         if not isinstance(callback, Callable):
-            callback = Callable(callable_function=callback)
+            callback = Callable(callback)
         super(DynamicMap, self).__init__(initial_items, callback=callback, **params)
 
         # Set source to self if not already specified
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -498,11 +498,17 @@ class DynamicMap(HoloMap):
         values on the key dimensions.
         """
         key = []
+        undefined = []
         for kdim in self.kdims:
             if kdim.values:
                 key.append(kdim.values[0])
             elif kdim.range:
                 key.append(kdim.range[0])
+            else:
+                undefined.append(kdim)
+        if undefined:
+            raise KeyError('dimensions do not specify a range or values, '
+                           'cannot supply initial key' % ', '.join(undefined))
         return tuple(key)
 
 
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -811,101 +811,72 @@ class DynamicMap(HoloMap):
 
     def collate(self):
         """
-        Collation allows collapsing DynamicMaps with invalid nesting
+        Collation allows reorganizing DynamicMaps with invalid nesting
         hierarchies. This is particularly useful when defining
         DynamicMaps returning an (Nd)Layout or GridSpace
-        type. Collating will split the DynamicMap into of individual
-        DynamicMaps. Note that the composite object has to be of
-        consistent length and types for this to work
-        correctly. Associating streams with specific viewables in the
-        returned container declare a stream_mapping on the DynamicMap
-        Callable during instantiation.
+        types. Collating will split the DynamicMap into individual
+        DynamicMaps for each item in the container. Note that the
+        composite object has to be of consistent length and types for
+        this to work correctly.
         """
         # Initialize
         if self.last is not None:
             pass
         else:
-            self[self._initial_key()]
+            self.clone()[self._initial_key()]
 
-        if isinstance(self.last, HoloMap):
-            # Get nested kdims and streams
-            streams = list(self.streams)
-            if isinstance(self.last, DynamicMap):
-                dimensions = [d(values=self.last.dimension_values(d.name))
-                              for d in self.last.kdims]
-                streams += self.last.streams
-                stream_kwargs = set()
-                for stream in streams:
-                    contents = set(stream.contents())
-                    if stream_kwargs & contents:
-                        raise KeyError('Cannot collate DynamicMaps with clashing '
-                                       'stream parameters.')
-            else:
-                dimensions = self.last.kdims
-            kdims = self.kdims+dimensions
+        if not isinstance(self.last, (Layout, NdLayout, GridSpace)):
+            return self
 
-            # Define callback
-            def collation_cb(*args, **kwargs):
-                return self[args[:self.ndims]][args[self.ndims:]]
-            callback = Callable(collation_cb, inputs=[self])
+        container = self.last.clone(shared_data=False)
 
-            return self.clone(shared_data=False, callback=callback,
-                              kdims=kdims, streams=streams)
-        elif isinstance(self.last, (Layout, NdLayout, GridSpace)):
-            # Expand Layout/NdLayout
-            from ..util import Dynamic
-            new_item = self.last.clone(shared_data=False)
-
-            # Get stream mapping from callback
-            remapped_streams = []
-            streams = self.callback.stream_mapping
-            for i, (k, v) in enumerate(self.last.data.items()):
-                vstreams = streams.get(i, [])
-                if not vstreams:
-                    if isinstance(self.last, Layout):
-                        for l in range(len(k)):
-                            path = '.'.join(k[:l])
-                            if path in streams:
-                                vstreams = streams[path]
-                                break
-                    else:
-                        vstreams = streams.get(k, [])
-                if any(s in remapped_streams for s in vstreams):
-                    raise ValueError(
-                        "The stream_mapping supplied on the Callable "
-                        "is ambiguous please supply more specific Layout "
-                        "path specs.")
-                remapped_streams += vstreams
-
-                # Define collation callback
-                def collation_cb(*args, **kwargs):
-                    return self[args][kwargs['collation_key']]
-                callback = Callable(partial(collation_cb, collation_key=k),
-                                    inputs=[self])
-                vdmap = self.clone(callback=callback, shared_data=False,
-                                   streams=vstreams)
-
-                # Remap source of streams
-                for stream in vstreams:
-                    if stream.source is self:
-                        stream.source = vdmap
-                new_item[k] = vdmap
-
-            unmapped_streams = [repr(stream) for stream in self.streams
-                                if (stream.source is self) and
-                                (stream not in remapped_streams)
-                                and stream.linked]
-            if unmapped_streams:
+        # Get stream mapping from callback
+        remapped_streams = []
+        streams = self.callback.stream_mapping
+        for i, (k, v) in enumerate(self.last.data.items()):
+            vstreams = streams.get(i, [])
+            if not vstreams:
+                if isinstance(self.last, Layout):
+                    for l in range(len(k)):
+                        path = '.'.join(k[:l])
+                        if path in streams:
+                            vstreams = streams[path]
+                            break
+                else:
+                    vstreams = streams.get(k, [])
+            if any(s in remapped_streams for s in vstreams):
                 raise ValueError(
-                   'The following streams are set to be automatically '
-                   'linked to a plot, but no stream_mapping specifying '
-                   'which item in the (Nd)Layout to link it to was found:\n%s'
-                    % ', '.join(unmapped_streams)
-                )
-            return new_item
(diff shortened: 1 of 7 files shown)


jbednar:
Thanks so much!  This will be really useful functionality.  

Integer indexing is easy enough to specify, but it seems fragile and also difficult to understand when skimming (the reader won't immediately see what "0" might mean). I had to re-read your description several times before I was fairly confident I understood how it was meant to work.  Is there a way to specify the item by group.label-style paths as well, to make it more explicit what is connected to what?  In this case there's no explicit label for any of the plots, but maybe if there were, it would be easier for people to see which bit is connected to what.

holoviews/core/spaces.py
```diff
@@ -788,6 +796,73 @@ def dynamic_redim(obj):
...
+        """
+        Collation allows collapsing DynamicMaps with invalid nesting
+        hierarchies. This is particularly useful when defining
+        DynamicMaps returning an (Nd)Layout. Collating will split the
+        DynamicMap into an (Nd)Layout of individual DynamicMaps. Note
+        that the Layout should be of consistent length and types for
+        this to work correctly. In order to attach a stream as a source
+        for a particular object in the Layout you may supply either
+        a dictionary or list of lists of streams corresponding to each
+        Element in the Layout.
```
jbednar: Layouts don't just contain Elements, do they?  E.g. an Overlay isn't an Element, is it, yet can be contained in a Layout?

philippjfr: Yes, need to be clear about terminology here. Not quite clear what the correct term is anymore.

philippjfr:
>Is there a way to specify the item by group.label-style paths as well, to make it more explicit what is connected to what?

Sure, I can add that, personally that seems more painful to specify but there's no reason why it shouldn't be allowed.

philippjfr:
@jbednar Would you expect only the full ``type[.group][.label]`` spec to match or should it allow partial matching?

holoviews/core/spaces.py
jbednar: Viewable?

jbednar: View?  Can't remember.

jbednar:
I'd expect whatever normally works elsewhere to work in this case.  Options are probably the most common place people encounter such paths, and they do support partial matching.  Moreover, I do think partial matching would be useful here (e.g. if I have an Image next to a Curve, it would be very clear to say I want this stream to be attached to the Image plot, not the Curve plot.  Plus getting a partial match is a lot easier than getting a complete match, given that labels are often auto-generated.  So yes, I would like partial matching, but since you're the one doing it, you'd be the one to make the call. :-)  Thanks.

holoviews/core/spaces.py
jlstevens: I think ``Viewable`` may be right. We rarely use that term though but I don't think I mind it here...

jlstevens:
I think ``collate`` to support ``DynamicMaps`` that output layouts is a great idea!

That said, I am strongly in favor of supplying the association between streams and elements up front so that you can just call ``collate()`` (without arguments) and have it work.

For this reason, I would prefer something like:

```
def test(x_range, y_range):
    return hv.Image(np.random.rand(10,10)) + hv.Bounds((x_range[0], y_range[0], x_range[1], y_range[1]))

stream = hv.streams.RangeXY(x_range=(0,-.5), y_range=(0,-0.5))
dmap = hv.DynamicMap(Callable(test, stream_mapping=spec), streams=[stream], kdims=[])
```

Where ``stream_mapping`` is a similar type of specification as you were implementing collate. Then if the ``Callable`` returns a Layout and has this specification define, ``collate`` can be called automatically. I really think it is best to declare the stream associations up-front when defining the ``DynamicMap`` (and not afterwards).

jbednar:
Sounds good to me.

philippjfr:
Ready to review/merge.

jlstevens:
Reviewing it now. One initial comment though..

Do you have what you consider a 'nice' (simple) example to show us what the code looks like, with the corresponding output? 

I know the unit tests has lots of examples but it would be good to have one you would like to highlight. It would then be something I would consider incorporating into the updated DynamicMap tutorials when I get round to working on those docs.

holoviews/core/spaces.py
```diff
@@ -402,6 +402,13 @@ class Callable(param.Parameterized):
...
     state of all streams on its inputs, to avoid calling the function
     unnecessarily.
+
+    A Callable may also specify a stream_mapping which allows
+    specifying which objects to attached linked streams to on
+    callbacks which return composite objects like (Nd)Layout and
+    GridSpace objects. The mapping should map between an integer index
+    or a type[.group][.label] specification and lists of streams
+    matching the object.
     """
```
jlstevens: Some of this is very hard to parse. Here is my attempt at making it easier to read:

> A Callable may also specify a stream_mapping which specifies the objects that are associated with interactive (i.e linked) streams when composite objects such as Layouts are returned from the callback. This is required for building interactive, linked visualizations (for the backends that support them) when returning Layouts, NdLayouts or GridSpace objects from the callback.

> The mapping should map from an appropriate key to a list of streams associated with the selected object. The appropriate key may be a type[.group][.label] specification for Layouts,  an integer index or a suitable NdLayout/GridSpace key. For more information see the DynamicMap tutorial at holoviews.org.

(Assuming it would be the DynamicMap tutorial)

holoviews/core/spaces.py
```diff
@@ -402,6 +402,13 @@ class Callable(param.Parameterized):
...
 
     callable_function = param.Callable(default=lambda x: x, doc="""
```
jlstevens: This doesn't have to happen in this PR but I'm strongly considering having this renamed to simply ``callable``. It would be a parameter/attribute and so wouldn't shadow the ``callable`` builtin.

holoviews/core/spaces.py
```diff
@@ -410,9 +417,12 @@ class Callable(param.Parameterized):
     inputs = param.List(default=[], doc="""
          The list of inputs the callable function is wrapping.""")
 
-    def __init__(self, **params):
+    def __init__(self, callable_function=None, stream_mapping={}, **params):
+        if callable_function is not None:
+            params['callable_function'] = callable_function
```
jlstevens: Any reason ``stream_mapping`` can't just be a parameter, allowing us to go back to the ``def __init__(self, **params)`` constructor?

In addition, how can ``callable_function`` be None? Isn't it required to define a valid ``Callable``?

holoviews/core/spaces.py
```diff
@@ -496,11 +506,17 @@ def _initial_key(self):
...
         for kdim in self.kdims:
             if kdim.values:
                 key.append(kdim.values[0])
             elif kdim.range:
                 key.append(kdim.range[0])
+            else:
+                undefined.append(kdim)
+        if undefined:
+            raise KeyError('dimensions do not specify a range or values, '
+                           'cannot supply initial key' % ', '.join(undefined))
```
jlstevens: No harm having this check here but I was thinking that ``DynamicMap`` should also be doing similar (if not identical) validation. I definitely want to be checking that kdims have the appropriate range settings there.

holoviews/core/spaces.py
```diff
@@ -788,6 +804,105 @@ def dynamic_redim(obj):
...
+        Collation allows collapsing DynamicMaps with invalid nesting
+        hierarchies. This is particularly useful when defining
+        DynamicMaps returning an (Nd)Layout or GridSpace
+        type. Collating will split the DynamicMap into of individual
+        DynamicMaps. Note that the composite object has to be of
+        consistent length and types for this to work
+        correctly. Associating streams with specific viewables in the
+        returned container declare a stream_mapping on the DynamicMap
+        Callable during instantiation.
+        """
```
jlstevens: > Collation allows collapsing DynamicMaps ...

'restructuring'? I'm not sure it is necessarily 'collapsing'...

> Collating will split the DynamicMap into of individual DynamicMaps.

Probably mostly true but what if the callback returns something invalid lower down in the hierarchy that (Nd)Layout or GridSpace? If collate were entirely general it would fix other weird things like NdOverlays of Overlays (actually what happens right now if you try this?).

> 'Note that the composite object ...' I would add the word 'returned' as in 'Note that the returned composite object ...'

> Associating streams with specific viewables in the returned container declare a stream_mapping on the DynamicMap Callable during instantiation.

Probably information that is more useful as a comment for developers and users. Users should be guided by the appropriate warning earlier on if they need to use ``stream_mapping`` so collate probably doesn't need to mention it (it should all just work!).

holoviews/plotting/renderer.py
```diff
@@ -159,16 +159,12 @@ def get_plot(self_or_cls, obj, renderer=None):
...
             obj = collate(obj)
-
-        # Initialize DynamicMaps with first data item
-        dmaps = obj.traverse(lambda x: x, specs=[DynamicMap])
-        for dmap in dmaps:
-            if dmap.sampled:
-                # Skip initialization until plotting code
-                continue
-            dmap[dmap._initial_key()]
+            initialize_dynamic(obj)
```
jlstevens: Makes sense as a utility.

holoviews/core/spaces.py
philippjfr: >Any reason stream_mapping can't just be a parameter, allowing us to go back to the def __init__(self, **params) constructor?

Agreed, not sure why I didn't make it one.

>In addition, how can callable_function be None? Isn't it required to define a valid Callable?

The parameter currently defines a no-op, which this leaves unchanged, but thinking about it, I don't think ``callable`` should just be None and become a required first argument.

holoviews/plotting/util.py
```diff
@@ -70,6 +72,19 @@ def collate(obj):
...
 
 
+def initialize_dynamic(obj):
+    """
+    Initializes all DynamicMap objects contained by the object
+    """
+    dmaps = obj.traverse(lambda x: x, specs=[DynamicMap])
+    for dmap in dmaps:
+        if dmap.sampled:
+            # Skip initialization until plotting code
```
jlstevens: Does this mean plotting does some of its own initialization? Maybe this utility could have a ``samples`` argument (or similar) to handle that case as well i.e if ``samples`` is None, this function stays the same as it is now, otherwise handles what is needed for sampled DynamicMaps.

holoviews/core/spaces.py
philippjfr: I think we already assert that when you're not in sampled mode.

holoviews/streams.py
```diff
@@ -149,7 +150,9 @@ def source(self):
     @source.setter
     def source(self, source):
         if self._source:
-            raise Exception('source has already been defined on stream.')
+            source_list = self.registry[id(self._source)]
+            if self in source_list:
+                source_list.remove(self)
```
jlstevens: Looks like it didn't clear itself up before...surely that causes some bugs?

holoviews/core/spaces.py
jlstevens: > I don't think callable should just be None and become a required first argument.

Agreed

holoviews/core/spaces.py
```diff
@@ -788,6 +804,105 @@ def dynamic_redim(obj):
...
+        # Initialize
+        if self.last is not None:
+            pass
+        else:
+            self[self._initial_key()]
```
jlstevens: Docstring should mention that ``collate`` will initialize the dynamicmap if it hasn't been already. Alternatively, shouldn't ``collate`` just complain if the dynamicmap is uninitialized and the renderer can initialize?

philippjfr: Hmm, then using it explicitly is a bit awkward.

holoviews/core/spaces.py
```diff
@@ -788,6 +804,105 @@ def dynamic_redim(obj):
...
+                raise ValueError(
+                   'The following streams are set to be automatically '
+                   'linked to a plot, but no stream_mapping specifying '
+                   'which item in the (Nd)Layout to link it to was found:\n%s'
+                    % ', '.join(unmapped_streams)
+                )
+            return new_item
+        else:
+            self.warning('DynamicMap does not need to be collated.')
+            return dmap
```
jlstevens: Why warn? 

It isn't inefficient to call collate pointlessly in such a case so just return the DynamicMap as it is already ok.

philippjfr: Sure.

holoviews/core/spaces.py
jlstevens: How about an ``initialize`` argument defaults to ``True``? I don't object to this as it is a simple boolean that disables behavior that is normally helpful...

holoviews/core/spaces.py
```diff
@@ -788,6 +804,105 @@ def dynamic_redim(obj):
...
+            def collation_cb(*args, **kwargs):
+                return self[args[:self.ndims]][args[self.ndims:]]
+            callback = Callable(collation_cb, inputs=[self])
+
+            return self.clone(shared_data=False, callback=callback,
+                              kdims=kdims, streams=streams)
+        elif isinstance(self.last, (Layout, NdLayout, GridSpace)):
+            # Expand Layout/NdLayout
+            from ..util import Dynamic
+            new_item = self.last.clone(shared_data=False)
```
jlstevens: Based on the ``isinstance`` check, ``container`` might be a better name for this variable than ``new_item``...

holoviews/core/spaces.py
```diff
@@ -788,6 +804,105 @@ def dynamic_redim(obj):
...
+                                vstreams = streams[path]
+                                break
+                    else:
+                        vstreams = streams.get(k, [])
+                if any(s in remapped_streams for s in vstreams):
+                    raise ValueError(
+                        "The stream_mapping supplied on the Callable "
+                        "is ambiguous please supply more specific Layout "
+                        "path specs.")
+                remapped_streams += vstreams
```
jlstevens: Maybe this could be a ``remap_streams`` utility?

As far as I can tell the required signature would be ``remap_streams(stream_mapping, object)`` where ``object`` here is ``self.last``.

holoviews/plotting/util.py
philippjfr: I don't think there is any explicit initialization in the plotting code it just works because ``select`` is called requesting the first sample defined by other HoloMap(s).

holoviews/core/spaces.py
philippjfr: I can see lines 858-867 becoming a nice self-contained utility, up to you if you think it's worth it.

holoviews/core/spaces.py
```diff
@@ -788,6 +804,105 @@ def dynamic_redim(obj):
...
+
+                # Define collation callback
+                def collation_cb(*args, **kwargs):
+                    return self[args][kwargs['collation_key']]
+                callback = Callable(partial(collation_cb, collation_key=k),
+                                    inputs=[self])
```
jlstevens: A couple of question:

Why use a partial? Can't ``collation_cb`` use ``k`` directly, forming a closure? Or is the issue that ``k`` is changing in a loop?

To make sure I understand, ``collation_key`` is really about selecting a portion of the structure returned by the callback? In which case ``selection_key`` might be a more intuitive name...

holoviews/core/spaces.py
philippjfr: Hmm, and raise an ``Exception`` when False? Doesn't seem very useful.

holoviews/plotting/util.py
jlstevens: Ok, that makes sense. No need to change this utility then.

holoviews/core/spaces.py
philippjfr: Sounds good.

holoviews/core/spaces.py
jlstevens: I'm not saying this option would be used a lot (in fact I'm counting on the opposite). I just think we can offer the option to disable some slightly surprising behavior. Won't calling ``self[self._initial_key()]`` change the cache? It must do as you are then using ``last``.

Maybe you should clone self and then use ``self[self._initial_key()]`` which would then not result in a side-effect. Then, if I understand correctly, the use of ``.last`` is just to get an example of the sort of thing the callback is returning...

holoviews/streams.py
philippjfr: Overriding an existing ``source`` raised an Exception before, but I think it's fine as it's not something you'd do without knowing what you're doing anyway.

holoviews/core/spaces.py
jlstevens: I think it is worth it if it can be given a sensible description in the docstring that makes sense and if you would like to write a few unit tests to both test it and show how it is used.

holoviews/core/spaces.py
```diff
@@ -788,6 +804,105 @@ def dynamic_redim(obj):
...
+                vdmap = self.clone(callback=callback, shared_data=False,
+                                   streams=vstreams)
+
+                # Remap source of streams
+                for stream in vstreams:
+                    if stream.source is self:
+                        stream.source = vdmap
```
jlstevens: Is this needed? Won't ``clone`` go through the ``DynamicMap`` constructor which assigns the stream sources to ``self``?

philippjfr: Not if it's already set to the original DynamicMap.

jlstevens: Ah, you are only doing this if ``stream.source is self`` (instance collate is called on). Not sure why yet...

holoviews/streams.py
jlstevens: Ok, that makes sense.

holoviews/core/spaces.py
philippjfr: >Why use a partial? Can't collation_cb use k directly, forming a closure? Or is the issue that k is changing in a loop?

Yes.

>To make sure I understand, collation_key is really about selecting a portion of the structure returned by the callback? In which case selection_key might be a more intuitive name...

Sure.

jlstevens:
I've made a full pass and posted my immediate comments. Generally looks good and my suggestions are mostly for clarification, not for any major refactoring.

holoviews/core/spaces.py
philippjfr: Making a clone as to not pollute the cache seems fine, an option that gives you an exception when you use it doesn't seem sensible.

jlstevens: Agreed. The reason I suggest the clone is so we don't need to worry about adding an option.

holoviews/core/spaces.py
philippjfr: This is the what does the actual remapping. You're remapping from the original DynamicMap returning the Layout to the DynamicMap created by collate, which returns a specific item.

jlstevens: Got it. And you need the full ``vstreams`` even if they don't have source as ``self`` as you should supply all streams no matter their source.

philippjfr:
Made all suggested fixes, except for creating another utility, which I don't think is worth it.

jlstevens:
Great! Tests are passing now.

Merging.

PR #1245, opened 2017-04-06
Tightened semantics of Dimension objects

jlstevens:
This PR aims to address issue #843 and clean-up the definition of Dimension objects generally.

It is still work in progress and I've started this PR at the first commit just so that I can see the results of the notebook tests as I progress.

**Edit**: I have copied over the action items from the issue:

- [x] Renaming the first argument from ``name`` to ``spec``.
- [x] Remove the deprecated 'initial' option from the values parameter.
- [x] Declaring ``label`` as a parameter
- [x] Allowing ``label`` to be passed as a kwarg and warning when ambiguous.
- [x] Fixing the ``__call__`` method, making it work with ``(name, label)`` tuples
- [x] Aliased ``__call__`` to ``clone``.
- [x] Implemented the ``spec`` property
- [x] Update class docstring  
- [x] Updated equality semantics
- [x] Updated comparison semantics
- [x] Add more unit tests.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -231,7 +231,8 @@ class Dimension(param.Parameterized):
         settings = dict(self.get_param_values(onlychanged=True), **overrides)
         spec = spec if (spec is not None) else (overrides.get('name', self.name),
                                                 overrides.get('label', self.label))
-        return self.__class__(spec, **{k:v for k,v in settings.items() if k != 'name'})
+        return self.__class__(spec, **{k:v for k,v in settings.items()
+                                       if k not in ['name', 'label']})
 
     def __hash__(self):
         """
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -228,8 +228,16 @@ class Dimension(param.Parameterized):
         except for the supplied, explicit overrides
         """
         settings = dict(self.get_param_values(onlychanged=True), **overrides)
-        spec = spec if (spec is not None) else (overrides.get('name', self.name),
-                                                overrides.get('label', self.label))
+
+        if spec is None:
+            spec = (self.name, overrides.get('label', self.label))
+        if 'label' in overrides and isinstance(spec, basestring) :
+            spec = (spec, overrides['label'])
+        elif 'label' in overrides and isinstance(spec, tuple) :
+            self.warning('Using label as supplied by keyword ({!r}), ignoring '
+                             'tuple value {!r}'.format(overrides['label'], spec[1]))
+            spec = (spec[0],  overrides['label'])
+
         return self.__class__(spec, **{k:v for k,v in settings.items()
                                        if k not in ['name', 'label']})
 
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -169,6 +169,12 @@ class Dimension(param.Parameterized):
         super(Dimension, self).__init__(**all_params)
 
 
+    @property
+    def spec(self):
+        "Returns the corresponding tuple specification"
+        return (self.name, self.label)
+
+
     def pprint(self):
         spec = self.name if self.name == self.label else (self.name, self.label)
         excluded = ['name'] if self.name == self.label else ['name', 'label']
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -91,6 +91,11 @@ class Dimension(param.Parameterized):
        'weight'. Valid Python identifiers make good names, because they
        can be used conveniently as a keyword in many contexts.""")
 
+    label = param.String(doc="""
+        Unrestricted label used to describe the dimension. A label
+        should succinctly describe the dimension and may contain any
+        characters, including Unicode and LaTeX expression.""")
+
     cyclic = param.Boolean(default=False, doc="""
         Whether the range of this feature is cyclic such that the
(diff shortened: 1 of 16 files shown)


holoviews/core/dimension.py
```diff
@@ -151,10 +151,11 @@ def __init__(self, name, **params):
             all_params['name'] = name
         self.label = label
 
-        if not isinstance(params.get('values', None), basestring):
-            all_params['values'] = sorted(list(unique_array(params.get('values', []))))
-        elif params['values'] != 'initial':
-            raise Exception("Values argument can only be set with the string 'initial'.")
+        values = params.get('values', [])
+        if isinstance(values, basestring) and values == 'initial':
+            self.warning("The 'initial' string for dimension values is no longer supported.")
```
philippjfr: Unless I'm confused set ``values`` to an empty list here, otherwise param will throw an exception after you've already warned.

philippjfr: Actually no, you'll get ``['a', 'i', 'i', 'i', 'l', 'n', 't']``.

jlstevens: Fixed in 9d57a58 (I pushed forced an amended commit)

philippjfr:
Looks good to me and I'm happy with the pprint.

jlstevens:
Great! 

I'm happy the tests have passed - I was worried there might be double specification of ``name`` somewhere. I'll now update the equality semantics and see if the tests still pass...

I am also considering making ``__str__`` return only ``self.name`` so equality with strings makes sense. This means using ``pprint_label`` - which I would also consider aliasing to a nicer name, ``pretty``, so it would be ``dim.pretty``.

jlstevens:
Rebasing against current master to get the push builds passing (pr builds are green).

jlstevens:
The pr build has now passed for Python 3 and 3. I will now try to eliminate old uses of ``str(dim)`` and make sure ``pprint_label`` is used instead.

jlstevens:
Rebasing again due to recent notebook test data updates.

holoviews/core/dimension.py
```diff
@@ -64,7 +63,8 @@ def replace_dimensions(dimensions, overrides):
         elif isinstance(override, Dimension):
             replaced.append(override)
         elif isinstance(override, dict):
-            replaced.append(d(**override))
+            replaced.append(d(override.get('name',None),
```
philippjfr: Should be using ``.clone`` internally now?

jlstevens: Yes, ```__call__`` is now an alias to clone but I do want to recommend ``clone`` from now on. Thanks for point it out and I'll fix it shortly.

holoviews/core/dimension.py
```diff
@@ -116,57 +154,120 @@ class Dimension(param.Parameterized):
...
+            existing_params = dict(spec.get_param_values())
+        elif (spec, params.get('unit', None)) in self.presets.keys():
+            preset = self.presets[(str(spec), str(params['unit']))]
             existing_params = dict(preset.get_param_values())
-        elif name in self.presets.keys():
-            existing_params = dict(self.presets[str(name)].get_param_values())
-        elif (name,) in self.presets.keys():
-            existing_params = dict(self.presets[(str(name),)].get_param_values())
+        elif spec in self.presets.keys():
+            existing_params = dict(self.presets[str(spec)].get_param_values())
```
philippjfr: This doesn't seem right, you check ``spec`` is in presets but then do the lookup with ``str(spec)``? Same below, also is ``.keys()`` needed in the conditional?

jlstevens:
Tests have passed and I think this PR is now ready to review.

holoviews/core/dimension.py
```diff
@@ -116,57 +154,120 @@ class Dimension(param.Parameterized):
...
+    def clone(self, spec=None, **overrides):
         """
         Derive a new Dimension that inherits existing parameters
         except for the supplied, explicit overrides
         """
         settings = dict(self.get_param_values(onlychanged=True), **overrides)
-        if name is not None: settings['name'] = name
-        return self.__class__(**settings)
+        spec = spec if (spec is not None) else (overrides.get('name', self.name),
+                                                overrides.get('label', self.label))
```
philippjfr: Don't want to check if name or label are supplied along with a spec?

holoviews/core/dimension.py
jlstevens: Fixed in 99ef6a8

holoviews/core/dimension.py
jlstevens: Addressed in 	703c01d

holoviews/core/dimension.py
jlstevens: Does it matter? If there is a spec, it gets passed through as the first argument, shouldn't matter if it is just a string or just a tuple.

philippjfr: Hmm, since you're filtering out name and label from kwargs below, those will simply be silently be dropped.

jlstevens: Ah got you. Fixed in 6d50a8a7.

jlstevens:
@philippjfr  I've addressed your comments and the tests are passing. Any other suggestions for dimension improvements?

philippjfr:
No, should I merge?

jlstevens:
If you are happy with the changes, please do!

PR #1247, opened 2017-04-07
Bokeh widgets editable

philippjfr:
Adds an editable option, which is False by default to improve the look of the widgets, but because that disables support for enabling text input on sliders, can be set to True if needed.

### DIFF
--- a/holoviews/plotting/bokeh/widgets.py
+++ b/holoviews/plotting/bokeh/widgets.py
@@ -8,7 +8,7 @@ import param
 import numpy as np
 from bokeh.io import _CommsHandle
 from bokeh.util.notebook import get_comms
-from bokeh.models.widgets import Select, Slider, AutocompleteInput, TextInput
+from bokeh.models.widgets import Select, Slider, AutocompleteInput, TextInput, Div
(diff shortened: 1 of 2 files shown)


philippjfr:
Ready to review/merge.

holoviews/plotting/bokeh/widgets.py
```diff
@@ -27,14 +27,8 @@ class BokehServerWidgets(param.Parameterized):
...
-    basejs = param.String(default=None, precedence=-1, doc="""
-        Defines the local CSS file to be loaded for this widget.""")
-
-    extensionjs = param.String(default=None, precedence=-1, doc="""
-        Optional javascript extension file for a particular backend.""")
-
-    css = param.String(default=None, precedence=-1, doc="""
-        Defines the local CSS file to be loaded for this widget.""")
+    editable = param.Boolean(default=False, doc="""
+        Whether the slider text fields should be editable.""")
```
jlstevens: Should mention the  the reason it is disabled by default. It sounds useful, so there is a reason we want it off by default...

jlstevens:
Fairly short PR and other than updating the docstring, it looks good.

jlstevens:
Tests passing. Merging.

PR #1250, opened 2017-04-07
Fixed bokeh hover tool specification

philippjfr:
Correctly escaping the column source name seems to be required when the visible label contains unicode characters. This fixes another small bug mentioned in https://github.com/ioam/holoviews/issues/1248.


jlstevens:
Looks good. Merging.

PR #1255, opened 2017-04-08
Implemented correct Raster.range

philippjfr:
Fix for https://github.com/ioam/holoviews/issues/1020, it wasn't accessing the masked array directly and hence was returning inconsistent results with Image, which does.

```python
%%opts Raster Image [colorbar=True]
tmp = np.random.rand(10,10)-0.5
tmp = np.ma.masked_where(tmp<=0, tmp)

hv.Image(tmp)+hv.Raster(tmp)
```

image

As we can see the masked values are no longer included in the range but are still indicated via the colorbar and the behavior is now consistent between Raster and Image.


jlstevens:
Looks good. Any tests we can add to make sure we don't lose the new (desired) behavior?

Otherwise happy to merge.

philippjfr:
Ready, merge when tests pass.

jlstevens:
Tests are passing. Merging.

PR #1259, opened 2017-04-10
Added logging output support to unittests

jlstevens:
This PR allows us to finally check param's logging output. We have often wanted to check the output of ``param.warning`` but haven't had the infrastructure to do so. This is important as we want to improve exceptions and warnings throughout HoloViews.

This PR includes an example of how the new ``LoggingComparisonTestCase`` is used, involving a recent unit test where I wanted to check the param.warning output. 

I'll be using this new system in a number of new unit tests I am planning but I don't intend to go back and update all the existing tests right now as it would take some thought to figure out where logging output should be checked. That could be work for a future PR and could be something to do before HoloViews 2.0.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,4 +1,8 @@
 import os, sys
+import param
+import logging
+
+from holoviews.element.comparison import ComparisonTestCase
 
 try:
     # Standardize backend due to random inconsistencies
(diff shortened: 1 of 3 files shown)


jlstevens:
@philippjfr I want this to be a quick PR and I do think it is ready to review/merge once the tests go green.

philippjfr:
Looks good.

Issue #1189, opened 2017-03-10
Support kwargs-based callbacks for DynamicMap?


jbednar:
Currently, DynamicMap requires that its callback use positional arguments, and it provides the values along each kdim in the order of the kdims, as in:

```python
import numpy as np
import holoviews as hv
import imagen as ig
import imagen.transferfn as tf
from imagen.image import FileImage
from imagen.transferfn.sheet_tf import Convolve
hv.notebook_extension()
%opts Image (cmap="gray") {+axiswise}

FileImage.filename = "/Users/jbednar/imagen/doc/images/ellen_arthur.pgm"

kdims = [hv.Dimension('size',                   range=(0.1, 2)),
         hv.Dimension('positive_size',          range=(0.1, 2)),
         hv.Dimension('negative_size',          range=(0.3, 4)),
         hv.Dimension('positive_aspect_ratio',  range=(1.0,10)),
         hv.Dimension('negative_aspect_ratio',  range=(1.0,10))]

def filtered_images(size,positive_size,negative_size,positive_aspect_ratio,negative_aspect_ratio):
    convolver = Convolve(kernel_pattern=ig.DifferenceOfGaussians(**locals()))
    return FileImage(output_fns=[convolver])[:]

FileImage()[:] + hv.DynamicMap(filtered_images, kdims=kdims)
```

Positional arguments work well for one or two arguments, but once there are a lot of them (and the above code shows only a small subset of the arguments that *could* be kdims in this example), keyword arguments can be much safer.  E.g. in the code above, if I add a kdim to that list, I need to be sure to add the appropriate argument to the `filtered_images` argument list, in precisely the right spot, or else everything will get jumbled.  If all I'm doing is passing the arguments down to some other object, as I am here, it's redundant and fragile to redeclare them in this way.

Can we make DynamicMap support calling its callback using keyword arguments instead, either by adding a parameter to it ``kwarg_callback=True`` or by inspecting the callback to determine whether it uses positional or keyword arguments?

jlstevens:
I think this would be a very reasonable thing to support. I'm just wondering if we should consider using inspection to do the right thing automatically...

jlstevens:
Addressed by the PR referenced above. Closing.

PR #1260, opened 2017-04-10
Automatically handle different signatures in Callable

jlstevens:
This PR aims to address #1189. The goal is to make ``Callable`` very flexible in what it accepts when it dispatches args and kwargs to the callback. This should mean that DynamicMap can accept almost any sensible callback.

## Action items

- [x] Make Callable flexible in how it dispatches arguments
- [x] Validate argspec against kdims and streams in DynamicMap 
- [x] Add more unit tests

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -460,6 +460,22 @@ class Callable(param.Parameterized):
         if memoize and hashed_key in self._memoized:
             return self._memoized[hashed_key]
 
+        if self.argspec.varargs is not None:
+            # Missing information on positional argument names, cannot promote to keywords
+            pass
+        elif len(args) != 0: # Turn positional arguments into keyword arguments
+            pos_kwargs = {k:v for k,v in zip(self.argspec.args, args)}
+            ignored = range(len(self.argspec.args),len(args))
+            if len(ignored):
+                self.warning('Ignoring extra positional argument %s'
+                             % ', '.join('%s' % i for i in ignored))
+            clashes = set(pos_kwargs.keys()) & set(kwargs.keys())
+            if clashes:
+                self.warning('Positional arguments %r overriden by keywords'
+                             % list(clashes))
+            args, kwargs = (), dict(pos_kwargs, **kwargs)
+
+
         ret = self.callable(*args, **kwargs)
         if hashed_key is not None:
             self._memoized = {hashed_key : ret}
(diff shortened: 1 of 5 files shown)


holoviews/core/util.py
```diff
@@ -99,6 +101,33 @@ def deephash(obj):
...
 
 
+
+def argspec(callable_obj):
+    """
+    Returns an ArgSpec object for functions, staticmethods, instance
+    methods, classmethods and partials.
+
+    Note that the args list for instance and class methods are those as
+    seen by the user. In other words, the first argument with is
```
jbednar: s/with/which?

jbednar:
Thanks for doing this!

holoviews/core/util.py
jlstevens: Thanks. Fixed in 51a10166

holoviews/core/spaces.py
```diff
@@ -625,7 +648,14 @@ def _execute_callback(self, *args):
         flattened = [(k,v) for kws in kwarg_items for (k,v) in kws
                      if k not in kdims]
         with dynamicmap_memoization(self.callback, self.streams):
-            retval = self.callback(*args, **dict(flattened))
+
+            if self._posarg_keys:
+                kwargs = dict(flattened, **dict(zip(self._posarg_keys, args)))
+                args = ()
+            else:
+                kwargs = dict(flattened)
```
philippjfr: No reason to have this inside the context manager as far as I can tell.

jlstevens: Agreed. I just did this because I wanted to be close to the actual call when I wrote it.

jlstevens: Addressed in 50022dff

jlstevens:
I think this is ready to merge once the tests pass. As ever, the new functionality needs to be documented in the tutorials, but that is outside the scope of this PR.

philippjfr:
>As ever, the new functionality needs to be documented in the tutorials, but that is outside the scope of this PR.

Agreed, but in the future after we've released 2.0 and overhauled our docs, I'd say that every feature PR should update be documented as a pre-requisite for being merged.

jlstevens:
That would be a good policy for the future and we should codify it at some point (e.g when we decide on a PR template).

Right now it doesn't make sense as the tutorials relating to DynamicMap need a complete overhaul.

philippjfr:
Looks good, those unit tests cover all the cases I can think of so I'm happy to merge.

PR #442, opened 2016-02-04
Set default mpl BoxWhisker linewidths

philippjfr:
As straightforward as the title says, will change a few display tests though.

PR #1262, opened 2017-04-10
Raised NotImplementedError for unsupported DynamicMap methods

philippjfr:
As outlined in #442 the reindex, add_dimension, and drop_dimension methods on DynamicMap are simply not valid because the callback expects specific arguments which cannot be changed. Since these are inherited from ``MultiDimensionalMapping`` which is really far removed from DynamicMap it'd be a nightmare to to generate sensible baseclasses to ensure these methods don't appear at all. Therefore I'm just raising NotImplementedErrors letting the user know that they'll have to cast their DynamicMap with a cache to a HoloMap before calling the one of those methods.

Edit: Reindex does make sense as long as you're only reordering dimensions, so I've now implemented it.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -1053,9 +1053,27 @@ class DynamicMap(HoloMap):
         else:
             return hist
 
+
     def reindex(self, kdims=[], force=False):
-        raise NotImplementedError('Cannot reindex a DynamicMap, '
-                                  'cast to a HoloMap first.')
+        """
+        Reindexing a DynamicMap allows reordering the dimensions but
+        not dropping an individual dimension. The force argument which
+        usually allows dropping non-constant dimensions is therefore
+        ignored and only for API consistency.
+        """
+        kdims = [self.get_dimension(kd, strict=True) for kd in kdims]
+        dropped = [kd for kd in self.kdims if kd not in kdims]
+        if dropped:
+            raise ValueError("DynamicMap does not allow dropping dimensions, "
+                             "reindex may only be used to reorder dimensions.")
+        reindexed = super(DynamicMap, self).reindex(kdims, force)
+        def reindex(*args, **kwargs):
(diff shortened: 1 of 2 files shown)


holoviews/core/spaces.py
```diff
@@ -1054,6 +1054,35 @@ def dynamic_hist(obj):
...
+        """
+        kdims = [self.get_dimension(kd, strict=True) for kd in kdims]
+        dropped = [kd for kd in self.kdims if kd not in kdims]
+        if dropped:
+            raise ValueError("DynamicMap does not allow dropping dimensions, "
+                             "reindex may only be used to reorder dimensions.")
+        reindexed = super(DynamicMap, self).reindex(kdims, force)
+        def reindex(*args, **kwargs):
+            keymap = {kd.name: arg for kd, arg in zip(self.kdims, args)}
+            keymap.update(kwargs)
```
philippjfr: Note: This bit is written with all kwarg callbacks in mind.

jbednar: Sounds good.

jlstevens: You mean you are assuming ``Callable`` can be invoked with a bunch of kwargs and nothing else?

That is fine unless the callback contains ``*args`` in which case ``Callable`` needs positional arguments it passes through directly to the callback. Once my PR is merged you can easily inspect the argspec to see if ``varargs`` is set, in which case you could raise another exception.

philippjfr: >That is fine unless the callback contains *args in which case Callable needs positional arguments it passes through directly to the callback. Once my PR is merged you can easily inspect the argspec to see if varargs is set, in which case you could raise another exception.

No need, this handles both cases already.

philippjfr:
Ready to merge.

jlstevens:
Looks good and it is nice to have reindex available: it could help people improve their indexing and can also be used to change the order of widgets.

Tests are green so I'll merge now.

Issue #1261, opened 2017-04-10
Support datetime formatting in bokeh hover tools


thoth291:
Here is what I have right now (installed all the packages for Python 3.6.1 a minute ago):
![image](https://cloud.githubusercontent.com/assets/2971670/24867348/848fb850-1dd2-11e7-9c0f-d8171b0b7e7a.png)

Please note that axis is rendered properly - it's just a hover tool who is missing the right rendering.

P.S. It's pretty typical `last mile` issue with `holoviews`: when you get your plot defined properly up to 95% but the last 5% makes you drop it and move to native bokeh, matplotlib struggling.

Thanks in Advance!

philippjfr:
I've now handled this correctly in #1263. That said there are various mechanisms to get around ``last mile`` issues, the simplest of which is to use ``finalize_hooks`` to modify the plot objects directly, e.g. to address this issue something like this would work:

```python
def fix_hover_date(plot, element):
    if 'hover' not in plot.handles:
        return
    xdim = element.kdims[0]
    hover = plot.handles['hover']
    source = plot.handles['source']
    hover.tooltips[0] = ('x', '@{x_dt}')
    source.data['x_dt'] = [xdim.pprint_value(v) for v in source.data['x']]

%%opts Curve [tools=['hover'] finalize_hooks=[fix_hover_date]]
hv.Curve((list(pd.date_range('2016-01-01', '2016-01-31')), range(31)))
```

philippjfr:
This has now been merged.

bacalfa:
I'm still having this issue.

![image](https://user-images.githubusercontent.com/1297074/37407218-84916298-276f-11e8-8197-a5bf20dd6418.png)

Here's the `dtypes` result:

```
Entry Time            datetime64[ns]
Frequency of Calls             int64
dtype: object
```

Package versions:
bokeh=0.12.14=py36_0
holoviews=1.9.4=py36_0

bacalfa:
This is weird... a small example worked fine:

![image](https://user-images.githubusercontent.com/1297074/37437465-cc9c7056-27c3-11e8-8865-1d99fe227351.png)

I have to figure out what's happening in the other case. Unfortunately, I can't share it. Any guesses?

bacalfa:
It seems it has to do with changing column/dimension names (?). Here's what's happening.

The following doesn't work as expected (dates aren't formatted properly in hover tool text):
```
df_resample= {}
df_resample_curves = []
for d in dlist:
[...]

PR #1263, opened 2017-04-10
Applied formatting to bokeh datetime hover tool labels

philippjfr:
As requested in #1261 datetimes in the hover tool should be formatted correctly. This has annoyed me myself a fair bit. Currently the only approach to make this work is to send the formatted strings as a column in the ColumnDataSource as outlined here, in future we can hopefully use JS datetime formatting to support this. I do think this is an important feature (it's visible in a number of my demos like the StockExplorer) and should make it into 1.7.

See https://github.com/bokeh/bokeh/issues/1239 for status of a more convenient/efficient mechanism for this.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -261,6 +261,8 @@ class ElementPlot(BokehPlot, GenericElementPlot):
             dim = util.dimension_sanitizer(d.name)
             if dim not in data:
                 data[dim] = element.dimension_values(d)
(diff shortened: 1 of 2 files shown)


jlstevens:
Shame it is such a pain but I agree it needs to be fixed. 

File an issue if you think there may be cleaner way to handle this in future. In the meantime I'll merge this PR given that the tests are passing.

philippjfr:
>File an issue if you think there may be cleaner way to handle this in future. In the meantime I'll merge this PR given that the tests are passing.

Not until bokeh do something about it.

PR #1266, opened 2017-04-10
Filter spectral cmaps from palette to avoid warning

philippjfr:
Didn't catch this one because it was py3 only. Just need to filter these, afaik they are dropped from ``plt.cm.datad`` in matplotlib master.


jlstevens:
Looks good. Merging.

PR #1267, opened 2017-04-11
Simplified DynamicMap reindex as it is handled automatically

philippjfr:
The new Callable inspection in _execute_callback now automatically remaps key dimensions onto the callback so wrapping the callable is no longer needed.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -1096,13 +1096,7 @@ class DynamicMap(HoloMap):
         if dropped:
             raise ValueError("DynamicMap does not allow dropping dimensions, "
                              "reindex may only be used to reorder dimensions.")
-        reindexed = super(DynamicMap, self).reindex(kdims, force)
-        def reindex(*args, **kwargs):
-            keymap = {kd.name: arg for kd, arg in zip(self.kdims, args)}
-            keymap.update(kwargs)
-            args = tuple(keymap[kd.name] for kd in kdims)
-            return reindexed[args]
-        return reindexed.clone(callback=Callable(reindex, inputs=[self]))
+        return super(DynamicMap, self).reindex(kdims, force)
 
 
     def drop_dimension(self, dimensions):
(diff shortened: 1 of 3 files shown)


philippjfr:
I also added made sure that any clone of a DynamicMap references the original DynamicMap in its ``Callable.inputs``, ensuring that the original source is inherited.

holoviews/core/spaces.py
```diff
@@ -667,9 +680,19 @@ def clone(self, data=None, shared_data=True, new_type=None, *args, **overrides):
...
-                                                   shared_data, new_type,
-                                                   *(data,) + args, **overrides)
+        clone = super(UniformNdMapping, self).clone(overrides.pop('callback', self.callback),
+                                                    shared_data, new_type,
+                                                    *(data,) + args, **overrides)
+
+        # Ensure the clone references this object to ensure
+        # stream sources are inherited
+        if clone.callback is self.callback:
+            clone.callback = self.callback.clone()
```
jlstevens: I don't quite get this line. Why would ``self.callback`` have a clone method? Isn't that a ``Callable`` around the callback?

jlstevens:
Other than one bit I don't understand it looks fine and I'll merge once the tests pass.

holoviews/core/spaces.py
philippjfr: Yes, see above I gave it a clone method. Can do it manually here I suppose and remove it again. Up to you.

philippjfr:
Needs a bunch of tests too, so not quite ready yet.

holoviews/core/spaces.py
jlstevens: It is at the top of the diff and I must have just skimmed over ``clone`` without registering it. Thanks for clarifying!

philippjfr:
Let's get this merged now since tests are failing on master holding up other PRs. I'll add tests later.

jlstevens:
Agreed. Merging now.

PR #1268, opened 2017-04-11
Defined option_state context manager

jlstevens:
This PR defines an ``option_state`` context manager in ``display_hooks.py``. It fixes the tracebacks and I think this is a better approach than what we had before.

What I don't know is whether it fixes the exception about not inheriting from None or something other than object. I don't quite remember what the reported message was as I've never been able to reproduce that particular issue myself!

### DIFF
--- a/holoviews/ipython/display_hooks.py
+++ b/holoviews/ipython/display_hooks.py
@@ -107,11 +107,15 @@ def dynamic_optstate(element, state=None):
 @contextmanager
 def option_state(element):
     optstate = dynamic_optstate(element)
+    raised_exception = False
     try:
         yield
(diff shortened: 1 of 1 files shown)


jlstevens:
Philipp gave me a snippet to reproduce the old-issue about None as exceptions and the last commit seems to finally fix it as well, while displaying the correct traceback.

If the tests pass (fingers crossed!) then I am hoping we can merge and say goodbye to this particular issue forever.

philippjfr:
Seems like the PR wasn't rebased but the PR build passed so I'll merge. Fingers crossed this is definitely the correct fix.

PR #1271, opened 2017-04-11
Ensure old paths aren't dropped when combining Layouts

philippjfr:
Fixes bug encountered when adding another plot to a collated Layout. The main issue seems to be that it ignored custom paths which do not match the auto-computed paths generated by computing a path from the group and label.


jlstevens:
Ok, the change here is simple but the layout code can be tricky. I hope the tests all pass!

jlstevens:
Test have passed. Best to merge early to help us catch any other issues.

PR #1274, opened 2017-04-11
Fixed comm creation and stream subscriber attachment

philippjfr:
Previously the top_level plot attached itself as a subscriber to all the streams found on the plot. Apart from being inefficient because any change to a stream would trigger replotting of the whole Layout, this could cause issues with one plot triggering range updates on another even though the second plot hadn't changed at all. 

In this PR each subplot is given access to the comm which means each subplot can be responsible for updating themselves.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -27,7 +27,7 @@ from ...core import util
 from ...element import RGB
 from ...streams import Stream, RangeXY, RangeX, RangeY
 from ..plot import GenericElementPlot, GenericOverlayPlot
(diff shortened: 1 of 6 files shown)


jlstevens:
That description makes sense as does the solution. There isn't too much changed code either.

That said, I think the only way we can be sure that everything is working properly is by merging ASAP and building lots of stream examples.

jlstevens:
Test have passed. Want to add some unit tests (might be tricky) or should I just merge?

philippjfr:
There are some unit tests, which actually caught a bug in my initial attempts at this.

PR #1278, opened 2017-04-12
Fixes for grid interface aggregation support with spreadfn

philippjfr:
Fixes small bug and adds test for aggregation with spreadfn.


jlstevens:
Looks good and the tests have passed. Merging.

Issue #1077, opened 2017-01-24
Links on http://holoviews.org/Tutorials/Transforming_Data.html are broken


jordansamuels:
http://holoviews.org/Tutorials/Transforming_Data.html has two broken links, for http://holoviews.org/Tutorials/Sampling_Data and http://holoviews.org/Tutorials/Exploring_Data .

jordansamuels:
Note that http://holoviews.org/Tutorials/Transforming_Data.html is linked twice from http://holoviews.org/features.html .

jlstevens:
Thanks for reporting this!

I'll assign myself to fix it.

philippjfr:
Might be worth discussing what the Transforming Data tutorial should actually contain or whether we should just remove it entirely.

jlstevens:
Addressed in the PR referenced above.

jordansamuels:
Thanks guys!

PR #1280, opened 2017-04-12
Remove links to transforming data page

jlstevens:
This PR addresses #1077. We opted to remove links to the transforming data page as it was incomplete and not in the notebook format. In addition the links to it were mostly mentioning slicing so the Sampling Data tutorial seems more appropriate anyway.

I don't think there is much to discuss and unless I get some feedback, I'll just trigger a docs build and merge.


jlstevens:
I've rebuilt the docs on [build.holoviews.org](http://build.holoviews.org) and fixed various links in the README. I think this PR is ready to merge.

philippjfr:
Thanks looks good.

PR #1281, opened 2017-04-12
Handle empty aggregation in datashader operations

philippjfr:
As the title says, this handles empty aggregates gracefully in the datashader operations and ensures the both the interfaces and plotting backends display handle the resulting aggregates correctly.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -130,7 +130,7 @@ class aggregate(ElementOperation):
         paths = []
         kdims = obj.kdims
         vdims = obj.vdims
-        x, y = obj.dimensions(label=True)[:2]
+        dims = obj.dimensions(label=True)[:2]
         if isinstance(obj, Path):
             glyph = 'line'
             for p in obj.data:
@@ -139,17 +139,28 @@ class aggregate(ElementOperation):
                     df[obj.vdims[0].name] = p.level
                 paths.append(df)
         elif isinstance(obj, CompositeOverlay):
+            element = None
             for key, el in obj.data.items():
                 x, y, element, glyph = cls.get_agg_data(el)
+                dims = (x, y)
                 df = PandasInterface.as_dframe(element)
                 if isinstance(obj, NdOverlay):
                     df = df.assign(**dict(zip(obj.dimensions('key', True), key)))
                 paths.append(df)
-            kdims += element.kdims
-            vdims = element.vdims
+            if element is None:
+                dims = None
+            else:
+                kdims += element.kdims
+                vdims = element.vdims
         elif isinstance(obj, Element):
             glyph = 'line' if isinstance(obj, Curve) else 'points'
             paths.append(PandasInterface.as_dframe(obj))
+
+        if dims is None or len(dims) != 2:
+            return None, None, None, None
+        else:
+            x, y = dims
+
         if len(paths) > 1:
             if glyph == 'line':
                 path = paths[0][:1]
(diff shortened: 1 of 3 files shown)


philippjfr:
Ready to review.

jlstevens:
Reviewing now. Planning on adding some unit  tests?

philippjfr:
>Reviewing now. Planning on adding some unit tests?

I suppose, there are currently no datashader unit tests.

holoviews/operation/datashader.py
```diff
@@ -139,17 +139,28 @@ def get_agg_data(cls, obj, category=None):
...
+                dims = None
+            else:
+                kdims += element.kdims
+                vdims = element.vdims
         elif isinstance(obj, Element):
             glyph = 'line' if isinstance(obj, Curve) else 'points'
             paths.append(PandasInterface.as_dframe(obj))
+
+        if dims is None or len(dims) != 2:
+            return None, None, None, None
```
jlstevens: Purely syntactic preference, but I would wrap the return tuple in parentheses i.e ``(None, None, None, None)``

holoviews/operation/datashader.py
```diff
@@ -168,6 +179,13 @@ def get_agg_data(cls, obj, category=None):
             df = paths[0]
         if category and df[category].dtype.name != 'category':
             df[category] = df[category].astype('category')
+
+        for d in (x, y):
+            if df[d].dtype.kind == 'M':
+                param.warning('Casting %s dimension data to integer '
+                              'datashader cannot process datetime data ')
```
jlstevens: How interpretable would the rest be after such datetime to int casting? I suppose it might work out but maybe it doesn't really make sense?

philippjfr: You can add a datetime formatter to your axis and it will work. I think it's fine with the warning.

holoviews/operation/datashader.py
```diff
@@ -176,6 +194,15 @@ def _process(self, element, key=None):
         category = agg_fn.column if isinstance(agg_fn, ds.count_cat) else None
         x, y, data, glyph = self.get_agg_data(element, category)
 
+        if x is None or y is None:
+            x0, x1 = self.p.x_range or (-0.5, 0.5)
+            y0, y1 = self.p.y_range or (-0.5, 0.5)
```
jlstevens: Seems like unit range is the default. Guess that is fine as long as this is sensible default behavior when the data is missing.

holoviews/operation/datashader.py
```diff
@@ -307,7 +334,12 @@ def _process(self, element, key=None):
 
         with warnings.catch_warnings():
             warnings.filterwarnings('ignore', r'invalid value encountered in true_divide')
-            img = tf.shade(array, **shade_opts)
+            if np.isnan(array.data).all():
```
jlstevens: Seems like it might be a little inefficient to compute this predicate on large arrays but I think it is okay for now. No need to optimize anything just yet.

holoviews/operation/datashader.py
```diff
@@ -176,6 +194,15 @@ def _process(self, element, key=None):
         category = agg_fn.column if isinstance(agg_fn, ds.count_cat) else None
         x, y, data, glyph = self.get_agg_data(element, category)
 
+        if x is None or y is None:
+            x0, x1 = self.p.x_range or (-0.5, 0.5)
+            y0, y1 = self.p.y_range or (-0.5, 0.5)
+            xc = np.linspace(x0, x1, self.p.width)
+            yc = np.linspace(y0, y1, self.p.height)
+            xarray = xr.DataArray(np.full((self.p.height, self.p.width), np.NaN, dtype=np.float32),
+                                  dims=['y', 'x'], coords={'x': xc, 'y': yc})
```
jlstevens: Isn't this all ``np.NaNs``? If so you could set a ``is_all_nans`` switch and use it later...

holoviews/operation/datashader.py
philippjfr: I worried about that too, but 100ms for a 10000x10000 array (which is considerably larger than we'll ever use), is okay.

jlstevens:
> I suppose, there are currently no datashader unit tests.

Ok, maybe file an issue about that then (referencing this PR) and we can address it later.

holoviews/operation/datashader.py
philippjfr: Not unless you want to add a is_all_nan switches to ``Image`` Elements. It's two distinct operations.

holoviews/operation/datashader.py
jlstevens: Sure.

jlstevens:
Made a few comments but otherwise I'm happy to merge.

philippjfr:
>Made a few comments but otherwise I'm happy to merge.

Let's just merge, I'll open an issue about unit tests for datashader operations.

jlstevens:
Looks good. Merging.

jbednar:
Looks good, thanks.  If you are working around behavior in datashader that you think should be fixed (e.g. if it should be raising a more sensible exception in some of these cases) then please file an issue there.

PR #1283, opened 2017-04-12
Added BokehRenderer.app method to create bokeh apps in scripts and notebooks

philippjfr:
Adds convenience method to create bokeh Apps from any holoviews object. In a script it simply returns the ``doc`` but when used with notebook option creates an ``Application`` instance which is displayed and returned.


jlstevens:
Great!

All it needs is an example script in ``plotting/bokeh/examples/apps/apps`` and maybe a script and notebook in ``holoviews-contrib``.

philippjfr:
>All it needs is an example script in plotting/bokeh/examples/apps/apps

I'll just change the existing examples to use this.

jlstevens:
> I'll just change the existing examples to use this.

Makes sense.

philippjfr:
Updated the examples, ready to merge once tests pass.

holoviews/plotting/bokeh/__init__.py
```diff
@@ -33,6 +33,9 @@
 
 Store.renderers['bokeh'] = BokehRenderer.instance()
 
+if len(Store.renderers) == 1:
```
jlstevens: Why not check that the one available renderer is 'bokeh'?

philippjfr: Guess I could but defining that renderer is literally the line above so it has to be bokeh.

philippjfr: Is your suggestion this?

``if list(Store.renderers.keys()) == ['bokeh']``

jlstevens: I suppose that's fine.

jlstevens:
Thanks for fixing the ``examples/apps/app`` nesting. I notice ``player.py`` was renamed but not changed..was that your intention?

holoviews/plotting/bokeh/__init__.py
jlstevens: You last post appeared before my reply. Either way is fine.

philippjfr: Yes, ``player`` does not use ``BokehRenderer.app`` it's an example of creating a custom bokeh layout with custom widgets.

philippjfr:
Yes, player does not use BokehRenderer.app it's an example of creating a custom bokeh layout with custom widgets.

holoviews/plotting/bokeh/__init__.py
jlstevens: In that case, I would mention that in a script-level docstring at the top of the file. Same for the crossfilter app. It only needs to be one line explaining what these examples are supposed to show.

jlstevens:
Looks good. Merging.

PR #1285, opened 2017-04-12
Disable caching of renderer parameters on OutputMagic 

philippjfr:
Additionally ensures that the order of backends supplied to the notebook_extension is respected. Fixes https://github.com/ioam/holoviews/issues/947.

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -206,6 +206,11 @@ class notebook_extension(param.ParameterizedFunction):
         if css:
             display(HTML(css))
 
+        for r in [r for r in resources if r != 'holoviews']:
+            Store.renderers[r].load_nb(inline=p.inline)
+        if selected_backend is not None:
+            Store.current_backend = selected_backend
+
         resources = list(resources)
         if len(resources) == 0: return
 
@@ -217,9 +222,6 @@ class notebook_extension(param.ParameterizedFunction):
         message = '' if not p.banner else '%s successfully loaded in this cell.' % loaded
         load_hvjs(logo=p.banner, JS=('holoviews' in resources), message = message)
 
-        for r in [r for r in resources if r != 'holoviews']:
-            Store.renderers[r].load_nb(inline=p.inline)
-        Store.current_backend = selected_backend
 
 
     def _get_resources(self, args, params):
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -146,15 +146,22 @@ class notebook_extension(param.ParameterizedFunction):
                  'plotly': 'plotly'}
 
     def __call__(self, *args, **params):
-        imports = [(name, b) for name, b in self._backends.items()
-                   if name in args or params.get(name, False)]
-        if not imports or 'matplotlib' not in Store.renderers:
-            imports = imports + [('matplotlib', 'mpl')]
+        # Get requested backends
+        imports = [(arg, self._backends[arg]) for arg in args
+                   if arg in self._backends]
+        for p, val in sorted(params.items()):
+            if p in self._backends:
+                imports.append((p, self._backends[p]))
+        if not imports:
+            imports = [('matplotlib', 'mpl')]
 
         args = list(args)
+        selected_backend = None
         for backend, imp in imports:
             try:
                 __import__('holoviews.plotting.%s' % imp)
+                if selected_backend is None:
+                    selected_backend = backend
             except ImportError:
                 if backend in args:
                     args.pop(args.index(backend))
@@ -187,7 +194,7 @@ class notebook_extension(param.ParameterizedFunction):
             ip = get_ipython() if ip is None else ip # noqa (get_ipython)
             param_ext.load_ipython_extension(ip, verbose=False)
             load_magics(ip)
-            OutputMagic.initialize(list( self._backends.keys()))
+            OutputMagic.initialize([backend for backend, _ in imports])
             set_display_hooks(ip)
             notebook_extension._loaded = True
 
@@ -209,11 +216,10 @@ class notebook_extension(param.ParameterizedFunction):
 
         message = '' if not p.banner else '%s successfully loaded in this cell.' % loaded
         load_hvjs(logo=p.banner, JS=('holoviews' in resources), message = message)
+
         for r in [r for r in resources if r != 'holoviews']:
             Store.renderers[r].load_nb(inline=p.inline)
-
-        if resources[-1] != 'holoviews':
-            get_ipython().magic(u"output backend=%r" % resources[-1]) # noqa (get_ipython))
+        Store.current_backend = selected_backend
 
 
     def _get_resources(self, args, params):
--- a/holoviews/ipython/magics.py
+++ b/holoviews/ipython/magics.py
@@ -97,7 +97,6 @@ class OptionsMagic(Magics):
                         info = (keyword,value)+allowed
                         raise ValueError("Value %r for key %r not between %s and %s" % info)
                 options[keyword] = value
-
         return cls._validate(options, items, linemagic)
 
     @classmethod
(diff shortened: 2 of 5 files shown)


holoviews/plotting/renderer.py
```diff
@@ -84,7 +84,7 @@ class Renderer(Exporter):
         The full, lowercase name of the rendering backend or third
         part plotting package used e.g 'matplotlib' or 'cairo'.""")
 
-    dpi=param.Integer(None, allow_None=True, doc="""
+    dpi=param.Integer(72, doc="""
```
jlstevens: Do we want to be setting a specific dpi in the Renderer baseclass? That would suggest all backends will want to use this dpi...

holoviews/ipython/__init__.py
```diff
@@ -146,15 +146,22 @@ class notebook_extension(param.ParameterizedFunction):
...
+        if not imports:
+            imports = [('matplotlib', 'mpl')]
 
         args = list(args)
+        selected_backend = None
         for backend, imp in imports:
             try:
                 __import__('holoviews.plotting.%s' % imp)
+                if selected_backend is None:
+                    selected_backend = backend
```
jlstevens: Good to see it is now obeying the specified order, with the first thing listed activated.

holoviews/plotting/renderer.py
philippjfr: Good point, I'll move it to MPLRenderer.

holoviews/ipython/__init__.py
```diff
@@ -209,11 +216,10 @@ def __call__(self, *args, **params):
 
         message = '' if not p.banner else '%s successfully loaded in this cell.' % loaded
         load_hvjs(logo=p.banner, JS=('holoviews' in resources), message = message)
+
         for r in [r for r in resources if r != 'holoviews']:
             Store.renderers[r].load_nb(inline=p.inline)
-
-        if resources[-1] != 'holoviews':
-            get_ipython().magic(u"output backend=%r" % resources[-1]) # noqa (get_ipython))
```
jlstevens: Good to get rid of this..it was hackish.

holoviews/ipython/magics.py
```diff
@@ -231,21 +230,28 @@ class OutputMagic(OptionsMagic):
...
+                            ('size'        , None),
+                            ('dpi'         , None),
                             ('charwidth'   , 80),
                             ('filename'    , None),
                             ('info'        , False),
-                            ('css'         , {})])
+                            ('css'         , None)])
+
+    # Defines the options the OutputMagic remembers the remainder
+    # is simply state on the backend specific Renderer
```
jlstevens: 'OutputMagic remembers the remainder is simply state'? Either punctuation or words are missing...

jlstevens: Maybe..

```
# Defines the options the OutputMagic remembers. All other options 
# are held by the backend specific Renderer.
```

Which is what I think you meant...

holoviews/ipython/magics.py
```diff
@@ -364,74 +375,73 @@ def update_options(cls, options, items):
...
     def initialize(cls, backend_list):
         cls.backend_list = backend_list
-        backend = cls.options.get('backend', cls.defaults['backend'])
+        backend = cls.options.get('backend', Store.current_backend)
         if backend in Store.renderers:
-            cls.options = dict(cls.defaults)
-            cls._set_render_options(cls.defaults)
-            cls.set_backend(backend)
+            cls.options = dict({k: cls.defaults[k] for k in cls.remembered})
+            cls._set_render_options({}, backend)
```
jlstevens: Why do you need to call ``_set_render_options`` if there are no options? I suppose there is state being initialize there?

philippjfr: I'll have a look at what I can do.

jlstevens: It was more of a question than a request. But if you see a way to improve it further, great!

holoviews/ipython/magics.py
```diff
@@ -364,74 +394,83 @@ def update_options(cls, options, items):
         Switch default options and backend if new backend
         is supplied in items.
         """
-        backend = items.get('backend', '')
+        # Get previous backend
         prev_backend = Store.current_backend
-        renderer = Store.renderers[Store.current_backend]
+        renderer = Store.renderers[prev_backend]
```
jlstevens: I would move these three line:

```python
prev_backend = Store.current_backend
renderer = Store.renderers[prev_backend]
prev_backend += ':%s' % renderer.mode
```

Just after the block starting with the ``# Get new backend`` comment and before ``# Update allowed formats``.

Seeing a method starting with 'prev_backend = Store.current_backend' just seems weird but once you've seen the code about getting the new backend it makes more sense. As far as I can tell, these variables don't seem to be referenced before that point.

holoviews/ipython/magics.py
```diff
@@ -364,74 +394,83 @@ def update_options(cls, options, items):
...
         prev_backend += ':%s' % renderer.mode
 
-        available = backend in Store.renderers.keys()
-        if (not backend) or (not available) or backend == prev_backend:
+        # Get new backend
+        backend = items.get('backend', Store.current_backend)
+        split = backend.split(':')
+        core_backend, mode = split if len(split)==2 else (split[0], 'default')
```
jlstevens: It would be nice if we had a word that meant 'backend plus mode' so that 'backend' is the thing without the mode.

jlstevens: Then we could get rid of 'core_backend' as a name (which I don't like). Not sure what terminology would make more sense. Maybe backend+mode is a ``backend_spec``?

holoviews/ipython/magics.py
```diff
@@ -337,25 +344,48 @@ def output(self, line, cell=None):
...
-            self.update_options(options, {'backend': restore_copy['backend']})
-            OutputMagic.options = restore_copy
-            self._set_render_options(restore_copy)
+            # After cell magic restore previous options and restore
+            # temporarily selected renderer
+            OutputMagic.options = prev_restore
+            self._set_render_options(render_params, backend)
+            if backend.split(':')[0] != prev_backend:
+                self.set_backend(prev_backend)
+                self._set_render_options(prev_params, prev_backend+':'+prev_mode)
```
jlstevens: I'm tempted to say something about context managers. Maybe for another day and another PR...

holoviews/ipython/magics.py
philippjfr: Hmm, so you just want to switch ``Get new backend`` and ``Get previous backend`` blocks around?

holoviews/ipython/magics.py
```diff
@@ -337,25 +344,48 @@ def output(self, line, cell=None):
...
             return
 
-        restore_copy = OrderedDict(OutputMagic.options.items())
+        # Make backup of previous options
+        prev_backend = Store.current_backend
+        prev_renderer = Store.renderers[prev_backend]
+        prev_mode = prev_renderer.mode
+        prev_params = {k: v for k, v in prev_renderer.get_param_values()
+                       if k in self.render_params}
+        prev_restore = dict(OutputMagic.options)
```
jlstevens: Maybe ``prev`` is confusing as it suggests the backend is changing (which it might not). Maybe ``initial_backend``? Implement this if you like the suggestion, it isn't something worth holding up the PR for.

holoviews/ipython/magics.py
philippjfr: Sure.

holoviews/ipython/magics.py
jlstevens: I think it is a bit more readable that way around. If there is an issue swapping them around, don't bother...

jlstevens:
I've made a few comments. It mostly looks fine and I'm happy to merge once you've seen my suggestions and decided whether to use them or not.

jlstevens:

PR #1286, opened 2017-04-12
Allowed styling bokeh hover glyphs

philippjfr:
I introduced all the machinery for this in https://github.com/ioam/holoviews/pull/1220 but forgot to enable it for hover glyphs. Here's an example of applying styling to the glyph for a QuadMesh Element:

hover


jlstevens:
Looks good and seems like a nice thing to enable. Merging.

PR #1288, opened 2017-04-13
Ensure max_cycles persists on style for non-batched plots

philippjfr:
Fixes a bug where in non-batched mode the max_cycles on styles would be reset causing Palettes to work incorrectly. Going to try adding some tests for this.


philippjfr:
Ready to merge.

jlstevens:
Looks good. Merging.

PR #1290, opened 2017-04-13
Improved handling of errors in notebook_extension

philippjfr:
Better exception handling when loading backends using the notebook_extension.

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -162,22 +162,29 @@ class notebook_extension(param.ParameterizedFunction):
                 __import__('holoviews.plotting.%s' % imp)
                 if selected_backend is None:
                     selected_backend = backend
-            except ImportError:
+            except Exception as e:
                 if backend in args:
                     args.pop(args.index(backend))
                 if backend in params:
                     params.pop(backend)
-                self.warning("HoloViews %s backend could not be imported, "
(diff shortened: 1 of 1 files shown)


jlstevens:
Looks good but could you mention exactly what the old behavior was and how the new behavior improve it?

In addition, I assume this doesn't fix the issue with the output magic where switching to matplotlib (when only bokeh has been loaded) doesn't issue an appropriate warning. Or does it? If not, it is something I wouldn't mind being included in this PR.

jlstevens:
Ok, I do see that it now handles ImportErrors differently from other exceptions and I agree that's useful. Does this PR do anything else?

philippjfr:
>Looks good but could you mention exactly what the old behavior was and how the new behavior improve it?

Yes, sorry lost my comment somewhere. Basically this does two things that probably mostly affect developers.

1) If a backend fails to import due to an Exception that's not an ImportError you get a warning with the original exception, usually this will happen when you've introduced a bug **somewhere** in the code and it would otherwise give you an uninformative error.

2) If the matplotlib backend fails to import it now no longer throws an exception, (note this is note the same as making bokeh independent of matplotlib in general).

3) If no backend imports there's now an exception.

jbednar:
That all sounds fabulous!

jlstevens:
Looks good and I think this is very useful. Merging.

Issue #1252, opened 2017-04-07
Smooth GridImage deprecation


philippjfr:
After #994 is applied ``GridImage`` will simply be an alias for ``Image``. Unfortunately in the meantime many users will have started to apply options to the ``GridImage`` name, we should make sure those options still apply but perhaps warn in those cases.

PR #1292, opened 2017-04-14
Added support for GridImage deprecation in OptsSpec parser

jlstevens:
This PR addresses issue #1252 by converting 'GridImage' to 'Image' in the opts parser. Here is an example of what it does:

image

### DIFF
--- a/holoviews/element/raster.py
+++ b/holoviews/element/raster.py
@@ -393,7 +393,10 @@ class Image(Dataset, Raster, SheetCoordinateSystem):
         return self.sheet2matrixidx(*coord)
 
 
-GridImage = Image
+class GridImage(Image):
+    def __init__(self, *args, **kwargs):
+        self.warning('GridImage is now deprecated. Please use Image element instead.')
+        super(GridImage, self).__init__(*args, **kwargs)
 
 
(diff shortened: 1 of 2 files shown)


jlstevens:
@philippjfr In raster.py we have:

```python
GridImage = Image
```

Would it be worth wrapping ``GridImage`` as a subclass of ``Image``, just to issue a deprecation warning in the constructor?

philippjfr:
>Would it be worth wrapping GridImage as a subclass of Image, just to issue a deprecation warning in the constructor?

Not a bad idea.

jlstevens:
Now when you try to build a ``GridImage``:

image

jlstevens:
@philippjfr Ready to review.

I think this PR should make things easier for anyone who might be using GridImage. You could try very fancy things (e.g remapping the option tree) but I think these simple changes should do the trick.

philippjfr:
>This PR addresses issue #1252 by converting 'GridImage' to 'Image' in the opts parser.

This does miss any explicitly assigning to the option tree which I've used in a few places. But since anything else is bound to be complex I'm fine with this since we also have the warning. Will merge when tests pass.

Issue #1026, opened 2016-12-21
Large layouts do not display


vascotenner:
A larger layout is not displayed when created with hv.Layout.

# Works:
```a[0]+a[1]+a[2]+a[3]```

# Does not work:
```hv.Layout(a)```

But returns
```
:Layout
   .Overlay.Beam     :Overlay
      .Image.Beam :Image   [x,y]   (z)
      .Level.Beam :Overlay
         .Beam :NdOverlay   [Levels]
            :Contours   [x,y]   (z)
   .Diff.Differences :Image   [x,y]   (z)
   .Table.I          :Table   [Name]   (radius,error)
   .Overlay.I        :Overlay
      .Curve.I  :Curve   [y]   (z)
      .Curve.II :Curve   [y]   (z)
```

# Current solution:
```hv.Layout(a).display('all')```

# Proposition
always display large layouts

**Comment by jbednar** [comment]
Is there instead an option that determines how large a layout can be and still be displayed, which could be set to infinity by a user that always wants to see large layouts?  That said, four items doesn't *sound* like a large layout.

**Comment by vascotenner** [comment]
It should at least give a warning that it does not display because it is to large + explanation how to show it.

Furthermore, the behaviour of +, Layout, hmap.layout should be identical.

**Comment by jbednar** [comment]
I agree on all counts!

**Comment by jlstevens** [comment]
I also agree.

I think issuing a warning and ensuring consistent behavior regardless of how the Layout is constructed is something we could fix for 1.7.

**Comment by philippjfr** [comment]
I say we throw out the ``max_branches`` option on the OutputMagic, then there's only one limit to worry about which is ``max_frames``, which is far easier to reason about. We could consider adding a max subplots option instead but I think that should have nothing to do with a Layout in particular. Therefore my preference would be to put a deprecation warning in ``Layout.display`` and then remove the conditional in the ``layout_display`` hook. It would just be something like this:

```python
if isinstance(layout, Layout) and len(layout.data) * nframes > max_frames:
       max_frame_warning(max_frames)
       return '<tt>'+ sanitize_HTML(layout) + '</tt>'
```

philippjfr:
Now implemented, we no longer check the number of branches on the Layout now and the ``Layout.display`` function is deprecated.

PR #1293, opened 2017-04-14
Deprecated max_branches output magic option

jlstevens:
This PR addresses #1026 by deprecating the ``display`` method on Layout and the ``max_branches`` option of the output magic. All Layouts should now be displayed consistently.

The only potential improvement might be to use param.warning instead of a print statement. I'm not sure this is necessary though as the output magic isn't normal user API and isn't parameterized. In this case a simple print statement seems fine instead of using the global param logger which isn't ideal in any case.


philippjfr:
Looks good, will merge when tests pass (and I'm back from lunch).

PR #1296, opened 2017-04-14
Explanation of rich display approach in HoloViews

jbednar:
Notebook starting from scratch explaining how HoloViews provides rich display for complex data, just as Python already does for simple data.


jlstevens:
Thanks!

I'll have another pass through this just to make sure I am happy with everything but it already looks good. The main difficulty will be generating and adding the test data and linking to it from the docs.

I do think it is worth adding for the 1.7 release.

jbednar:
I don't think I've tested it with py3, btw.

jlstevens:
@jbednar Could you add a line right at the start of the tutorial saying that this notebook supports our SciPy paper (and linking to it).

I think that will make it clear that this is material for people who want to dig a bit deeper into what HoloViews is really about i.e people who want to think a bit more deeply about it as opposed to a simplified introduction.

jlstevens:
@jbednar If you are busy I am happy to merge right away and have that change implemented in a separate PR. I would like to merge sooner rather than later...

PR #1297, opened 2017-04-16
Fixed dynamic stream sources assignment in plotting code

philippjfr:
Previously the ``get_sources`` function was very wrong in the way it reconstructed the individual layers of an Overlay from a DynamicMap. This resulted in issues in correctly detecting stream sources while plotting. This PR readds the OverlayCallable class which lets an improved ``get_sources`` function correctly infer stream sources.

Will add a lot of unit tests for the ``get_sources`` function before merging.

### DIFF
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -150,7 +150,7 @@ class ElementOperation(Operation):
         elif dynamic:
             from ..util import Dynamic
             processed = Dynamic(element, streams=self.p.streams,
-                                link_inputs=self.p.link.inputs,
+                                link_inputs=self.p.link_inputs,
                                 operation=self, kwargs=params)
         elif isinstance(element, ViewableElement):
             processed = self._process(element)
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -104,8 +104,12 @@ class ElementOperation(Operation):
        component is Normalization.keys. """)
 
     link_inputs = param.Boolean(default=False, doc="""
-         Whether a plot created from the output of this operation
-         should be linked to the inputs of this operation.""")
+         If the operation is dynamic, whether or not linked streams
+         should be transferred from the operation inputs for backends
+         that support linked streams. For example if an operation is
+         applied to a DynamicMap with an RangeXY determines if the
+         output of the operation will update the stream on the input
+         with current axis ranges.""")
 
     streams = param.List(default=[], doc="""
         List of streams that are applied if dynamic=True, allowing
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -104,12 +104,14 @@ class ElementOperation(Operation):
        component is Normalization.keys. """)
 
     link_inputs = param.Boolean(default=False, doc="""
-         If the operation is dynamic, whether or not linked streams
-         should be transferred from the operation inputs for backends
-         that support linked streams. For example if an operation is
-         applied to a DynamicMap with an RangeXY determines if the
-         output of the operation will update the stream on the input
-         with current axis ranges.""")
+       If the operation is dynamic, whether or not linked streams
+       should be transferred from the operation inputs for backends
+       that support linked streams.
+
+       For example if an operation is applied to a DynamicMap with an
+       RangeXY, this switch determines whether the corresponding
+       visualization should update this stream with range changes
+       originating from the newly generated axes.""")
 
     streams = param.List(default=[], doc="""
         List of streams that are applied if dynamic=True, allowing
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -104,7 +104,8 @@ class ElementOperation(Operation):
        component is Normalization.keys. """)
 
     link_inputs = param.Boolean(default=False, doc="""
-       Whether to link to any stream sources on the inputs.""")
+         Whether a plot created from the output of this operation
+         should be linked to the inputs of this operation.""")
 
     streams = param.List(default=[], doc="""
         List of streams that are applied if dynamic=True, allowing
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -103,6 +103,9 @@ class ElementOperation(Operation):
        first component is a Normalization.ranges list and the second
        component is Normalization.keys. """)
 
+    link_inputs = param.Boolean(default=False, doc="""
+       Whether to link to any stream sources on the inputs.""")
+
     streams = param.List(default=[], doc="""
         List of streams that are applied if dynamic=True, allowing
         for dynamic interaction with the plot.""")
@@ -139,8 +142,8 @@ class ElementOperation(Operation):
             processed = element.clone(grid_data)
         elif dynamic:
             from ..util import Dynamic
-            streams = getattr(self.p, 'streams', [])
-            processed = Dynamic(element, streams=streams,
+            processed = Dynamic(element, streams=self.p.streams,
+                                link_inputs=self.p.link.inputs,
                                 operation=self, kwargs=params)
         elif isinstance(element, ViewableElement):
             processed = self._process(element)
--- a/holoviews/core/overlay.py
+++ b/holoviews/core/overlay.py
@@ -135,7 +135,15 @@ class Overlay(Layout, CompositeOverlay):
 
 
     def __mul__(self, other):
-        if not isinstance(other, ViewableElement):
+        if type(other).__name__ == 'DynamicMap':
+            from .spaces import OverlayCallable
+            def dynamic_mul(*args, **kwargs):
+                element = other[args]
+                return self * element
+            callback = OverlayCallable(dynamic_mul, inputs=[self, other])
+            return other.clone(shared_data=False, callback=callback,
+                               streams=[])
+        elif not isinstance(other, ViewableElement):
             raise NotImplementedError
         return Overlay.from_values([self, other])
 
--- a/holoviews/core/overlay.py
+++ b/holoviews/core/overlay.py
@@ -29,7 +29,7 @@ class Overlayable(object):
                 element = other[args]
                 return self * element
             callback = Callable(dynamic_mul, inputs=[self, other])
-            callback._overlay = True
+            callback._is_overlay = True
             return other.clone(shared_data=False, callback=callback,
                                streams=[])
         if isinstance(other, UniformNdMapping) and not isinstance(other, CompositeOverlay):
@@ -142,7 +142,7 @@ class Overlay(Layout, CompositeOverlay):
                 element = other[args]
                 return self * element
             callback = Callable(dynamic_mul, inputs=[self, other])
-            callback._overlay = True
+            callback._is_overlay = True
             return other.clone(shared_data=False, callback=callback,
                                streams=[])
         elif not isinstance(other, ViewableElement):
--- a/holoviews/core/overlay.py
+++ b/holoviews/core/overlay.py
@@ -24,11 +24,11 @@ class Overlayable(object):
 
     def __mul__(self, other):
         if type(other).__name__ == 'DynamicMap':
-            from .spaces import Callable
+            from .spaces import OverlayCallable
             def dynamic_mul(*args, **kwargs):
                 element = other[args]
                 return self * element
-            callback = Callable(dynamic_mul, inputs=[self, other])
+            callback = OverlayCallable(dynamic_mul, inputs=[self, other])
             return other.clone(shared_data=False, callback=callback,
                                streams=[])
         if isinstance(other, UniformNdMapping) and not isinstance(other, CompositeOverlay):
@@ -38,7 +38,6 @@ class Overlayable(object):
             return NotImplemented
 
         return Overlay.from_values([self, other])
-
 
 
 
--- a/holoviews/core/overlay.py
+++ b/holoviews/core/overlay.py
@@ -24,11 +24,12 @@ class Overlayable(object):
 
     def __mul__(self, other):
         if type(other).__name__ == 'DynamicMap':
-            from .spaces import OverlayCallable
+            from .spaces import Callable
             def dynamic_mul(*args, **kwargs):
                 element = other[args]
                 return self * element
-            callback = OverlayCallable(dynamic_mul, inputs=[self, other])
+            callback = Callable(dynamic_mul, inputs=[self, other])
+            callback._overlay = True
             return other.clone(shared_data=False, callback=callback,
                                streams=[])
         if isinstance(other, UniformNdMapping) and not isinstance(other, CompositeOverlay):
@@ -136,11 +137,12 @@ class Overlay(Layout, CompositeOverlay):
 
     def __mul__(self, other):
         if type(other).__name__ == 'DynamicMap':
-            from .spaces import OverlayCallable
+            from .spaces import Callable
             def dynamic_mul(*args, **kwargs):
                 element = other[args]
                 return self * element
-            callback = OverlayCallable(dynamic_mul, inputs=[self, other])
+            callback = Callable(dynamic_mul, inputs=[self, other])
+            callback._overlay = True
             return other.clone(shared_data=False, callback=callback,
                                streams=[])
         elif not isinstance(other, ViewableElement):
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -120,27 +120,11 @@ class HoloMap(UniformNdMapping, Overlayable):
             key_map = {d.name: k for d, k in zip(dimensions, key)}
             layers = []
             try:
-                if isinstance(self, DynamicMap):
-                    if self.kdims:
-                        self_el = self.select(**key_map)
-                    else:
-                        self_el = self[()]
-                    if self_el is not None:
-                        layers.append(self_el)
-                else:
-                    layers.append(self[key])
+                self_el = self.select(**key_map) if self.kdims else self[()]
             except KeyError:
                 pass
             try:
-                if isinstance(other, DynamicMap):
-                    if other.kdims:
-                        other_el = other.select(**key_map)
-                    else:
-                        other_el = other[()]
-                    if other_el is not None:
-                        layers.append(other_el)
-                else:
-                    layers.append(other[key])
+                other_el = other.select(**key_map) if other.kdims else other[()]
             except KeyError:
                 pass
             return Overlay(layers)
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -4,7 +4,6 @@ from numbers import Number
 from itertools import groupby
 from functools import partial
 from contextlib import contextmanager
-from collections import defaultdict
 
 import numpy as np
 import param
@@ -100,6 +99,7 @@ class HoloMap(UniformNdMapping, Overlayable):
         return [tuple(zip([d.name for d in self.kdims], [k] if self.ndims == 1 else k))
                 for k in self.keys()]
 
+
     def _dynamic_mul(self, dimensions, other, keys):
         """
         Implements dynamic version of overlaying operation overlaying
@@ -117,11 +117,14 @@ class HoloMap(UniformNdMapping, Overlayable):
         map_obj = self if isinstance(self, DynamicMap) else other
 
         def dynamic_mul(*key, **kwargs):
+            key_map = {d.name: k for d, k in zip(dimensions, key)}
             layers = []
             try:
                 if isinstance(self, DynamicMap):
-                    safe_key = () if not self.kdims else key
-                    _, self_el = util.get_dynamic_item(self, dimensions, safe_key)
+                    if self.kdims:
+                        self_el = self.select(**key_map)
+                    else:
+                        self_el = self[()]
                     if self_el is not None:
                         layers.append(self_el)
                 else:
@@ -130,8 +133,10 @@ class HoloMap(UniformNdMapping, Overlayable):
                 pass
             try:
                 if isinstance(other, DynamicMap):
-                    safe_key = () if not other.kdims else key
-                    _, other_el = util.get_dynamic_item(other, dimensions, safe_key)
+                    if other.kdims:
+                        other_el = other.select(**key_map)
+                    else:
+                        other_el = other[()]
                     if other_el is not None:
                         layers.append(other_el)
                 else:
@@ -139,7 +144,8 @@ class HoloMap(UniformNdMapping, Overlayable):
             except KeyError:
                 pass
             return Overlay(layers)
-        callback = OverlayCallable(dynamic_mul, inputs=[self, other])
+        callback = Callable(dynamic_mul, inputs=[self, other])
+        callback._overlay = True
         if map_obj:
             return map_obj.clone(callback=callback, shared_data=False,
                                  kdims=dimensions, streams=[])
@@ -206,7 +212,8 @@ class HoloMap(UniformNdMapping, Overlayable):
                 def dynamic_mul(*args, **kwargs):
                     element = self[args]
                     return element * other
-                callback = OverlayCallable(dynamic_mul, inputs=[self, other])
+                callback = Callable(dynamic_mul, inputs=[self, other])
+                callback._overlay = True
                 return self.clone(shared_data=False, callback=callback,
                                   streams=[])
             items = [(k, v * other) for (k, v) in self.data.items()]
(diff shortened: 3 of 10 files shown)


philippjfr:
This is ready for review now.

holoviews/util.py
```diff
@@ -80,7 +80,8 @@ def dynamic_operation(*key, **kwargs):
         else:
             def dynamic_operation(*key, **kwargs):
                 self.p.kwargs.update(kwargs)
-                _, el = util.get_dynamic_item(map_obj, map_obj.kdims, key)
+                safe_key = () if not map_obj.kdims else key
+                el = map_obj[key]
```
jlstevens: You could inline ``map_obj[key]`` into ``self._process``...

holoviews/core/spaces.py
```diff
@@ -509,6 +516,47 @@ def get_nested_streams(dmap):
     return list(set(layer_streams))
 
 
+
+def get_stream_sources(obj, branch=0):
```
jlstevens: Any particular reason for the name change from ``get_sources``? Or you just think this is a better name?

philippjfr: Better name, especially in the context of bokeh where I regularly use ``source`` to refer to ColumnDataSources.

holoviews/core/spaces.py
```diff
@@ -686,11 +734,14 @@ def clone(self, data=None, shared_data=True, new_type=None, *args, **overrides):
...
         # stream sources are inherited
-        if clone.callback is self.callback:
-            clone.callback = self.callback.clone()
-        if self not in clone.callback.inputs:
+        if isinstance(clone.callback, OverlayCallable):
+            from ..util import Dynamic
+            return Dynamic(clone, shared_data=shared_data)
+        elif clone.callback is self.callback:
+            clone.callback = clone.callback.clone()
+        if not any(inp is self for inputs in get_stream_sources(clone).values() for inp in inputs):
```
jlstevens: How about ``if all(inp is not self ...``?

Why do you only do this if self isn't in the inputs?

holoviews/core/spaces.py
```diff
@@ -495,6 +495,13 @@ def __call__(self, *args, **kwargs):
         return ret
 
 
+class OverlayCallable(Callable):
+    """
+    A Callable subclass specifically meant to indicate that the Callable
+    represents an overlay operation between two objects.
+    """
```
jlstevens: I'm not convinced it is worth introducing a subclass just for isinstance checks. Why not some parameter/attribute on ``Callable``?

This is another classname to know about and it isn't clear from this docstring why callables that involve overlay operations should be special.

jlstevens: Ok, the diff didn't make it clear that this class has methods too. But why can't those methods be on ``Callable``?

holoviews/core/spaces.py
```diff
@@ -509,6 +516,47 @@ def get_nested_streams(dmap):
     return list(set(layer_streams))
 
 
+
+def get_stream_sources(obj, branch=0):
+    """
+    Traverses Callable graph to resolve sources on DynamicMap objects,
+    returning a dictionary of sources indexed by the Overlay layer. The
+    branch variables is used for internal record-keeping while recursing
+    through the graph.
```
jlstevens: I'm still not clear what this means or does. Maybe an example would help?

jlstevens:
I've had multiple passes through the code and I still have no idea what problem this PR is solving. Could you explain what the problem was?

holoviews/core/spaces.py
philippjfr: It doesn't have methods. I'd be okay making it a parameter on a Callable, but that would make it more likely for users use it incorrectly, which is worse than simply not declaring it at all because it has the potential to mess up the other streams on an Overlay.

holoviews/core/spaces.py
jlstevens: As this is a function, any chance this could move to util?

holoviews/plotting/bokeh/element.py
```diff
@@ -190,9 +190,18 @@ def _construct_callbacks(self):
...
+        else:
+            zorders = [self.zorder]
+
+        if isinstance(self, OverlayPlot) and not self.batched:
+            sources = []
         if not self.static or isinstance(self.hmap, DynamicMap):
-            sources = [(i, o) for i, o in get_sources(self.hmap)
-                       if i in [None, self.zorder]]
+            sources = [(i, o) for i, inputs in self.stream_sources.items()
+                       for o in inputs if i in zorders]
```
jlstevens: What does being a source have to do with a z-ordering? The latter is a visualization concept so what does that have to with being a source or not?

holoviews/core/spaces.py
jlstevens: Is this a class users should ever make use of directly? I would have no idea whether to use it or not based on the current docstring.

philippjfr:
>Could you explain what the problem was?

When you use dynamic_mul or simply return an Overlay in a DynamicMap, it has to traverse the graph specified by the dynamicmap and assign each node to a particular layer in the returned Overlay. Each subplot created by an OverlayPlot has an index of the layer it is plotting (the zorder), by computing a mapping between the layer index (i.e. zorder) and the sources associated with that layer we can figure out what callbacks a particular plot should create. The previous approach was simply wrong in all kinds of ways.

holoviews/core/spaces.py
philippjfr: Util cannot import from any files in core.

holoviews/plotting/bokeh/element.py
philippjfr: See below

holoviews/core/spaces.py
philippjfr: So I don't add self to the callback inputs multiple times.

philippjfr: I'll go back to a ``self not in ...`` approach for clarity.

jlstevens: This line seems to be the only reason ``get_stream_sources`` can't be to ``plotting.util``.

holoviews/core/spaces.py
jlstevens: Why not make this an underscore attribute on ``Callable``? Then users have no reason to touch it or know about it?

holoviews/core/overlay.py
```diff
@@ -136,7 +135,15 @@ def __add__(self, other):
...
-        if not isinstance(other, ViewableElement):
+        if type(other).__name__ == 'DynamicMap':
+            from .spaces import OverlayCallable
+            def dynamic_mul(*args, **kwargs):
+                element = other[args]
+                return self * element
+            callback = OverlayCallable(dynamic_mul, inputs=[self, other])
+            return other.clone(shared_data=False, callback=callback,
+                               streams=[])
+        elif not isinstance(other, ViewableElement):
```
jlstevens: We really need to refactor ``__mul__`` (dynamic and non-dynamic) into an operation for 2.0. It was bad enough before considering dynamic and now we have `` _dynamic_mul`` methods and ``dynamic_mul`` inner functions all over the place.

holoviews/core/spaces.py
jlstevens: I would recommend ``plotting.util`` but that doesn't seem possible either.

holoviews/core/spaces.py
jlstevens: Sounds reasonable.

What I don't like is that nothing about this mentions streams! It is really ``get_potential_stream_sources`` but I'm thinking the name should really reflect that is is about splitting objects in overlays into a dictionary.

jlstevens: I'm thinking this function should really be called something like ``overlayable_zorder``.

holoviews/core/spaces.py
jlstevens: I think you don't need ``get_stream_sources`` here as you aren't using z-ordering - you just use ``.values``. If we can make this bit not use `` get_stream_sources`` then that function could move to ``plotting.util``.

philippjfr: Agreed, I can make a much simpler function which simply traverses the object, ignoring the zordering.

holoviews/core/spaces.py
philippjfr: Sure sounds reasonable.

philippjfr:
@jlstevens This is ready for review again, any suggestions for improving docstrings welcome. The concept of ``link_inputs`` will have to be explained in the Linked Streams tutorial.

holoviews/util.py
jlstevens: Suggestion still stands.

holoviews/plotting/util.py
```diff
@@ -72,6 +73,65 @@ def collate(obj):
...
 
 
+def isoverlay_fn(obj):
+    """
+    Determines whether object is a DynamicMap returning (Nd)Overlay types.
+    """
+    return isinstance(obj, DynamicMap) and (isinstance(obj.last, CompositeOverlay))
+
+
+def linked_zorders(obj, path=[]):
```
jlstevens: Although *much* better than before, I'm still not sure this name is right. The ``linked`` part of the name is about how it is used (candidate sources for streams) *not* what it does which is better described by the first part of the docstring which talks about 'objects'.

How about ``object_zorders``?

philippjfr: ``overlayable_zorders``?

holoviews/core/operation.py
```diff
@@ -103,6 +103,10 @@ class ElementOperation(Operation):
        first component is a Normalization.ranges list and the second
        component is Normalization.keys. """)
 
+    link_inputs = param.Boolean(default=False, doc="""
+         Whether a plot created from the output of this operation
+         should be linked to the inputs of this operation.""")
+
```
jlstevens: Needs a *lot* more explanation as almost none of the required concepts exist at this level:

1. An operation doesn't create plots, it generates a data-structure.
2. Specifically, linking is about streams which relates to DynamicMap so this should say something about when the operation is dynamic.
3. Linked streams only exist for specific backends.

As a very rough suggestion, it should be something along the lines of:

'If the operation is dynamic, whether or not linked streams should be transferred from the operation inputs for backends that support linked streams. For example ... '

holoviews/core/spaces.py
```diff
@@ -429,6 +439,10 @@ class Callable(param.Parameterized):
          The list of inputs the callable function is wrapping. Used
          to allow deep access to streams in chained Callables.""")
 
+    link_inputs = param.Boolean(default=True, doc="""
+         Whether a plot created from this Callable should be linked
+         to the declared inputs.""")
```
jlstevens: Similar issue with the ``link_inputs`` parameter on operations - it needs to explain a lot more but in terms of ``Callable`` concepts this time.

holoviews/plotting/util.py
jlstevens: Sure.

holoviews/core/spaces.py
```diff
@@ -413,7 +420,10 @@ class Callable(param.Parameterized):
     when composite objects such as Layouts are returned from the
     callback. This is required for building interactive, linked
     visualizations (for the backends that support them) when returning
-    Layouts, NdLayouts or GridSpace objects.
+    Layouts, NdLayouts or GridSpace objects. When chaining multiple
+    DynamicMaps into a pipeline the link_inputs parameter declares
+    whether a plot created from the Callable owner is linked to objects
+    referenced in the inputs.
```
jlstevens: Maybe 'visualization' instead of plot.

How about something like...

'When chaining multiple DynamicMaps into a pipeline, the link_inputs parameter declares whether the visualization generated from this Callable will inherit the linked streams. This parameter is used as a hint by the applicable backend'

holoviews/core/spaces.py
```diff
@@ -441,6 +455,8 @@ class Callable(param.Parameterized):
     def __init__(self, callable, **params):
         super(Callable, self).__init__(callable=callable, **params)
         self._memoized = {}
+        self._overlay = False
```
jlstevens: Maybe ``self._is_overlay`` to make it clearer it is a predicate as opposed to a command?

holoviews/operation/datashader.py
```diff
@@ -264,6 +264,10 @@ class shade(ElementOperation):
         and any valid transfer function that accepts data, mask, nbins
         arguments.""")
 
+    link_inputs = param.Boolean(default=True, doc="""
+         Whether a plot created from the output of this operation
+         should be linked to the inputs of this operation.""")
```
jlstevens: Docstring should now be about what this means for the ``shade`` operation. Something like:

'By default, the link_inputs parameter is set to True so that when applying shade ...'

holoviews/operation/datashader.py
```diff
@@ -395,6 +399,10 @@ class dynspread(ElementOperation):
         Higher values give more spreading, up to the max_px
         allowed.""")
 
+    link_inputs = param.Boolean(default=True, doc="""
+         Whether a plot created from the output of this operation
+         should be linked to the inputs of this operation.""")
+
```
jlstevens: Same issue as for the shade operation.

holoviews/core/spaces.py
```diff
@@ -117,11 +117,14 @@ def _dynamic_mul(self, dimensions, other, keys):
...
+            key_map = {d.name: k for d, k in zip(dimensions, key)}
             layers = []
             try:
                 if isinstance(self, DynamicMap):
-                    safe_key = () if not self.kdims else key
-                    _, self_el = util.get_dynamic_item(self, dimensions, safe_key)
+                    if self.kdims:
+                        self_el = self.select(**key_map)
+                    else:
+                        self_el = self[()]
```
jlstevens: Might as well be:

```python
self_el = self.select(**key_map) if self.kdims else self[()]
```

holoviews/core/operation.py
```diff
@@ -103,6 +103,14 @@ class ElementOperation(Operation):
        first component is a Normalization.ranges list and the second
        component is Normalization.keys. """)
 
+    link_inputs = param.Boolean(default=False, doc="""
+         If the operation is dynamic, whether or not linked streams
+         should be transferred from the operation inputs for backends
+         that support linked streams. For example if an operation is
+         applied to a DynamicMap with an RangeXY determines if the
+         output of the operation will update the stream on the input
+         with current axis ranges.""")
```
jlstevens: Much better!

Minor edits:

> For example if an operation is applied to a DynamicMap with an RangeXY, *this switch* determines
> whether the *corresponding visualization*  should update this stream with range changes originating
> from the newly generated axes.

holoviews/core/spaces.py
```diff
@@ -413,7 +404,11 @@ class Callable(param.Parameterized):
     when composite objects such as Layouts are returned from the
     callback. This is required for building interactive, linked
     visualizations (for the backends that support them) when returning
-    Layouts, NdLayouts or GridSpace objects.
+    Layouts, NdLayouts or GridSpace objects. When chaining multiple
+    DynamicMaps into a pipeline, the link_inputs parameter declares
+    whether the visualization generated from this Callable will
+    inherit the linked streams. This parameter is used as a hint by
+    the applicable backend.
```
jlstevens: Think this read a bit better:

> the link_inputs parameter declares  whether the visualization generated *using this* Callable

holoviews/core/spaces.py
```diff
@@ -429,6 +424,16 @@ class Callable(param.Parameterized):
...
 
+    link_inputs = param.Boolean(default=True, doc="""
+         If the Callable wraps around other DynamicMaps in its inputs,
+         determines if linked streams attached to the inputs areb
+         transferred to the objects returned by the Callable.
+
+         For example if the Callable wraps a DynamicMap with an
+         RangeXY stream determines if the output of the Callable will
+         update the stream on the input with current axis ranges for
+         backends that support linked streams.""")
```
jlstevens: I think you can use my earlier suggestion here as well without changing it much.

jlstevens:
A general comment about the docstring. For instance you have:

> '... determines if linked streams ...'

I would replace all these instance of the word 'if' (when relating to the switch) with the word 'whether'. That might just be my personal preference though...

holoviews/plotting/util.py
```diff
@@ -72,6 +73,67 @@ def collate(obj):
...
+def isoverlay_fn(obj):
+    """
+    Determines whether object is a DynamicMap returning (Nd)Overlay types.
+    """
+    return isinstance(obj, DynamicMap) and (isinstance(obj.last, CompositeOverlay))
+
+
+def compute_overlayable_zorders(obj, path=[]):
+    """
+    Traverses the Callables on DynamicMap to determine which objects
```
jlstevens: Isn't the traversal of DynamicMap one of the things it needs to do as opposed to the only thing it needs to do?

How about:

> Traverses an overlayable  composite container to determine which objects are associated with specific (Nd)Overlay layers by z-order, taking DynamicMap Callables into account.  Returns a mapping between the zorders of each layer and a corresponding lists of objects.

holoviews/plotting/util.py
```diff
@@ -72,6 +73,67 @@ def collate(obj):
...
+    in the graph are associated with specific (Nd)Overlay layers.
+    Returns a mapping between the zorders of each layer and lists of
+    objects associated with each.
+
+    Used to determine which subplots should be linked with Stream
```
jlstevens: Maybe 'which *overlaid* subplots'?

holoviews/plotting/util.py
```diff
@@ -72,6 +73,67 @@ def collate(obj):
...
+    callbacks.
+    """
+    path = path+[obj]
+    zorder_map = defaultdict(list)
+
+    # Process non-dynamic layers
+    if not isinstance(obj, DynamicMap):
+        if isinstance(obj, CompositeOverlay):
+            for i, o in enumerate(obj):
```
jlstevens: Might be more readable renaming the ``i`` local variable to ``z``.

holoviews/plotting/util.py
```diff
@@ -72,6 +73,67 @@ def collate(obj):
...
+                zorder_map[i] = [o, obj]
+        else:
+            if obj not in zorder_map[0]:
+                zorder_map[0].append(obj)
+        return zorder_map
+
+    isoverlay = isinstance(obj.last, CompositeOverlay)
```
jlstevens: As the previous conditional returned for non-DynamicMaps this must be a DynamicMap now and there is no guarantee that there is anything in the cache for ``.last``. As long as you have anticipated that ``.last`` may be None - in which case ``isoverlay`` is False - then this should be ok...

holoviews/plotting/util.py
```diff
@@ -72,6 +73,67 @@ def collate(obj):
...
+    isdynoverlay = obj.callback._is_overlay
+    found = any(isinstance(p, DynamicMap) and p.callback._overlay for p in path)
```
jlstevens: I think the definition of ``found`` can be moved just above where it is used.

holoviews/plotting/util.py
```diff
@@ -72,6 +73,67 @@ def collate(obj):
...
+    if obj not in zorder_map[0] and not isoverlay:
+        zorder_map[0].append(obj)
```
jlstevens: I think this would execute if ``.last`` is None. Maybe this is ok because you are assuming ``initial_key`` is called already?

holoviews/plotting/util.py
philippjfr: The relevant nodes are evaluated by this point.

holoviews/plotting/util.py
philippjfr: Yes, this is in plotting, the graph has to have been evaluated by this point, even if certain (unimportant) intermediate nodes haven't been.

holoviews/plotting/util.py
jlstevens: Ok, just to be sure I would assert that ``.last`` is not ``None`` to make sure ``initial_key`` was called.

philippjfr: No, not all paths have to have been evaluated, that would break everything.

holoviews/plotting/util.py
```diff
@@ -72,6 +73,67 @@ def collate(obj):
...
+
+    # Process the inputs of the DynamicMap callback
+    dmap_inputs = obj.callback.inputs if obj.callback.link_inputs else []
+    for i, inp in enumerate(dmap_inputs):
+        if any(not (isoverlay_fn(p) or p.last is None) for p in path) and isoverlay_fn(inp):
+            # Skips branches of graph that collapse Overlay layers
+            # to avoid adding layers that have been reduced or removed
+            continue
```
jlstevens: What happens if you have a DynamicMap that takes in an overlay of 3 things and returns an overlay of 2 things?

holoviews/plotting/util.py
jlstevens: Right, I agree.

holoviews/plotting/util.py
```diff
@@ -82,55 +82,56 @@ def isoverlay_fn(obj):
...
+    for z, inp in enumerate(dmap_inputs):
         if any(not (isoverlay_fn(p) or p.last is None) for p in path) and isoverlay_fn(inp):
             # Skips branches of graph that collapse Overlay layers
             # to avoid adding layers that have been reduced or removed
             continue
 
         # Recurse into DynamicMap.callback.inputs and update zorder_map
-        i = i if isdynoverlay else 0
+        z = z if isdynoverlay else 0
         dinputs = compute_overlayable_zorders(inp, path=path)
```
jlstevens: Maybe call this ``zinputs``?

holoviews/plotting/util.py
```diff
@@ -82,55 +82,56 @@ def isoverlay_fn(obj):
...
         offset = max(zorder_map.keys())
         for k, v in dinputs.items():
```
jlstevens: Instead of ``k`` maybe this should be called ``zinp`` or something?

holoviews/plotting/util.py
```diff
@@ -82,55 +82,56 @@ def isoverlay_fn(obj):
...
     # Process the inputs of the DynamicMap callback
     dmap_inputs = obj.callback.inputs if obj.callback.link_inputs else []
-    for i, inp in enumerate(dmap_inputs):
+    for z, inp in enumerate(dmap_inputs):
         if any(not (isoverlay_fn(p) or p.last is None) for p in path) and isoverlay_fn(inp):
             # Skips branches of graph that collapse Overlay layers
             # to avoid adding layers that have been reduced or removed
             continue
 
         # Recurse into DynamicMap.callback.inputs and update zorder_map
```
jlstevens: So this would *not* happen for a DynamicMap callback with inputs that is user defined i.e ``isdynoverlay`` is ``False`` because ``_is_overlay`` wouldn't be set to True by the user? Not saying this is wrong, just making sure this is what we want...

philippjfr: >So this would not happen for a DynamicMap callback with inputs that is user defined i.e isdynoverlay is False because _is_overlay wouldn't be set to True by the user? 

That's correct, the branch below handles that by iterating over the items in the user defined DynamicMap, which will ensure the zorder is maintained but won't attempt to figure out linked inputs.

jlstevens: I see. As we discussed, this should be ok as the user can set things up explicitly if they want to transfer sources around (which means they shouldn't ever need to use ``_is_overlay``.

Seems fine.

holoviews/plotting/util.py
jlstevens: Now handled.

holoviews/plotting/util.py
```diff
@@ -116,13 +132,18 @@ def compute_overlayable_zorders(obj, path=[]):
             # Skips branches of graph that collapse Overlay layers
             # to avoid adding layers that have been reduced or removed
             continue
+        elif depth is not None and depth < overlay_depth(inp):
+            # Skips branch of graph where the number of elements in an
+            # overlay has been reduced
+            continue
```
jlstevens: Does this mean linking from inputs is disabled for this case?

jlstevens:
PR is now looking a lot better than when I first saw it!

The code and concepts involved are fairly tricky but I think it has now been greatly clarified and having lots of unit tests does reassure me. I've asked one more question in a comment but otherwise I am happy to merge once the tests pass.

holoviews/plotting/util.py
philippjfr: Yes, it can't figure out which parts of the input operation should continue to be linked. A user may be able to supply that information but at that point its easier for them to declare their own Callable and list the inputs they need explicitly.

jlstevens: Thanks for clarifying!

philippjfr: For example in this unit test they could do this:

```python
combined = (area_redim*curve_redim*area2).map(lambda x: x.clone(x.items()[:2]), Overlay)
combined.callback.inputs += [area_redim, curve_redim]
```

jlstevens: That's not bad at and better than trying to complicate things for what I expect is a fairly rare case anyway.

jlstevens:
Tests are passing. 

Merging now so we can test it as much as possible before the release.

PR #1300, opened 2017-04-17
Added custom rcParam context manager

philippjfr:
For some reason the context manager to set rcParams provided by matplotlib is not working correctly, I've implemented a tiny one of our own here which fixes the bug mentioned in https://github.com/ioam/holoviews/issues/1299.

### DIFF
--- a/holoviews/plotting/mpl/plot.py
+++ b/holoviews/plotting/mpl/plot.py
@@ -1,6 +1,7 @@
 from __future__ import division
 
 from itertools import chain
+from contextlib import contextmanager
 
 import numpy as np
 import matplotlib as mpl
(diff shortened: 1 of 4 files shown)


holoviews/plotting/mpl/plot.py
```diff
@@ -19,12 +20,27 @@
 from .util import compute_ratios, fix_aspect
 
 
+@contextmanager
+def rc_context(rcparams):
```
jlstevens: To avoid confusion maybe this should be ``_rc_context`` if we expect to use the decorator and not this context manager directly. Perhaps you could even inline this context manager inside ``mpl_rc_context``?

philippjfr: Inlining seems inefficient but I'll happily rename.

jlstevens:
Looks good apart from one comment. It is suspicious that the context manager supplied by matplotlib doesn't seem to work right...

basnijholt:
This fix seems to work for me 👍

holoviews/plotting/mpl/plot.py
jlstevens: I don't think it is at all inefficient to inline - decorators are only applied once at definition time.

jlstevens:
Looks good. Merging.

PR #1301, opened 2017-04-17
Do not attach stream subscribers to batched ElementPlot

philippjfr:
Batched plots create both an OverlayPlot and an ElementPlot instance which currently both attach themselves as stream subscribers, however the correct thing to do is for only the OverlayPlot to attach itself so that the batched plot is updated correctly.


jlstevens:
Are there any unit tests that could make sure the correct things are attacked as subscribers when batched?

jlstevens:
New unit test looks good. I'll merge once the tests go green.

jlstevens:
Tests are green. Merging.

PR #1302, opened 2017-04-17
Unified redim implementation with convenience methods

jlstevens:
This PR unifies redim and makes it much more convenient to use, which is especially helpful for DynamicMap declarations. You can now do:

image

Instead of the more verbose equivalent we have been using till now:

image

Even for this simplest possible example with only a single kdim, the new style is shorter.

All existing use of redim should act in exactly the same way. What is new are all the redim auxiliary methods that make it easier to pick a particular Dimension parameter to change.

## Tasks

- [x] Unify redim, removing existing redim methods
- [x] Implement auxiliary redim methods
- [x] Write more redim unit test

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -9,7 +9,6 @@ except ImportError:
 import numpy as np
 import param
 
-from ..dimension import replace_dimensions
 from .interface import Interface
 from .array import ArrayInterface
 from .dictionary import DictInterface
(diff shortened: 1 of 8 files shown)


jlstevens:
I've added 10 unit tests of the auxiliary redim method. They are very simple and the core redim method is well tests already.

@philippjfr  I think this PR is ready for review. I hope to get it merged ASAP as I have some nice improvements to DynamicMap usage that builds on top of this.

philippjfr:
Looks good, I would have considered subclasses that override the ``__call__`` but this is fine too. Happy to merge when tests pass.

jlstevens:
>  I would have considered subclasses that override the ``__call__``

Agreed. I would definitely split it up into subclasses if the current redim ``__call__`` method gets any more complicated than it is now i.e if we find ourselves wanting new functionality or needing a new mode.

philippjfr:
Okay sounds good, merging.

PR #1303, opened 2017-04-17
Handle additional linked stream condition

philippjfr:
Followup for #1297 to handle one more condition where a DynamicMap which returns an (Nd)Overlay itself has linked streams.


jlstevens:

PR #1304, opened 2017-04-17
Reverted change to Dynamic utility

philippjfr:
In #1297 I seem to have introduced a bug to the Dynamic utility, reverting here.


jlstevens:
I don't mind reverting as long as you know what went wrong and why it wouldn't be worth trying to fix the new approach...

philippjfr:
Closing in favour #1306.

PR #1306, opened 2017-04-17
Avoid side-effects in datashader operation

philippjfr:
The kdims and vdims declared here are modified later messing up modifying the kdims and vdims on the original object.


jlstevens:
Happy to merge right away, although I might as well wait for the tests to complete.

jlstevens:
The pr tests passed and the failure in the push build is just a transient. Merging.

PR #1308, opened 2017-04-18
Fixed unreferenced variable in decimate operation

philippjfr:
Tiny one line fix.


jlstevens:
Merging as the one test failure is just a transient.

PR #1310, opened 2017-04-18
Small fix for Polygons/Contours colormapping

philippjfr:
Small bug when colormapping a Polygons/Contours object. The unit test already checked the colormapper was created but didn't check it was actually used.


jlstevens:
Looks good. Merging.

PR #1313, opened 2017-04-18
Ensure batched plots respect parameters set on (Nd)Overlay

philippjfr:
In batched mode the OverlayPlot exists only as an intermediate layer and is not responsible for creating the axes or the actual plot object. This means this is all delegated to the batched ElementPlot, but this also means that any options set on the OverlayPlot are ignored, which is inconsistent and confusing. Since the options this applies to already propagate upwards from ElementPlots to the OverlayPlot we can use the same mechanism to pass those options back down to the batched ElementPlot, ensuring that they are respected.

This is particularly import for things like grids where options should propagate downwards from the gridplot to all the subplots, e.g. here's an example from a GridMatrix plot with batched subplots

**Before:**

screen shot 2017-04-18 at 5 03 48 pm

**After**:

screen shot 2017-04-18 at 5 02 59 pm

### DIFF
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -850,6 +850,9 @@ class GenericOverlayPlot(GenericElementPlot):
             (not self.show_legend or len(ordering) > self.legend_limit)):
             self.batched = True
             keys, vmaps = [()], [self.hmap]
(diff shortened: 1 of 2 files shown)


jlstevens:

PR #1315, opened 2017-04-18
Ensure streams are transferred when applying Dynamic

philippjfr:
Currently the Dynamic utility does not inherit the streams of the DynamicMap it is wrapping. Since it is used for internal operations like redim, map, select, etc. this is confusing because streams you specified just disappear. This PR ensures that internal operations always inherit **ALL** the streams, while Dynamic also ensures that dimensioned streams (i.e. streams that have contents which correspond to a Dynamic key dimension) are always inherited. This will also have the added bonus that validation will ensure that you you don't accidentally override a dimensioned stream.

* [x] Ensure internal operations/methods transfer all streams
* [x] Ensure that Dynamic always transfers dimensioned streams
* [x] Improve exception message when overriding dimensioned stream.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -125,9 +125,10 @@ class redim(object):
             return redimmed
 
         from ..util import Dynamic
-        def dynamic_redim(obj):
+        def dynamic_redim(obj, **dynkwargs):
             return obj.redim(specs, **dimensions)
-        return Dynamic(redimmed, shared_data=True, operation=dynamic_redim)
+        return Dynamic(redimmed, shared_data=True, streams=parent.streams,
+                       operation=dynamic_redim)
 
 
     def _redim(self, name, specs, **dims):
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -119,7 +119,7 @@ class HoloMap(UniformNdMapping, Overlayable):
         if isinstance(self, DynamicMap) and isinstance(other, DynamicMap):
             self_streams = util.dimensioned_streams(self)
             other_streams = util.dimensioned_streams(other)
-            streams = self_streams+other_streams
+            streams = list(util.unique_iterator(self_streams+other_streams))
         else:
             streams = map_obj.streams
 
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -114,7 +114,14 @@ class HoloMap(UniformNdMapping, Overlayable):
             dimensions = [d(values=grouped[d.name]) for d in dimensions]
             map_obj = None
 
+        # Combine streams
         map_obj = self if isinstance(self, DynamicMap) else other
+        if isinstance(self, DynamicMap) and isinstance(other, DynamicMap):
+            self_streams = util.dimensioned_streams(self.streams)
+            other_streams = util.dimensioned_streams(other.streams)
+            streams = self_streams+other_streams
+        else:
+            streams = map_obj.streams
 
         def dynamic_mul(*key, **kwargs):
             key_map = {d.name: k for d, k in zip(dimensions, key)}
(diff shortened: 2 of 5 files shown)


holoviews/core/dimension.py
```diff
@@ -125,9 +125,10 @@ def __call__(self, specs=None, **dimensions):
             return redimmed
 
         from ..util import Dynamic
-        def dynamic_redim(obj):
+        def dynamic_redim(obj, **kwargs):
             return obj.redim(specs, **dimensions)
-        return Dynamic(redimmed, shared_data=True, operation=dynamic_redim)
+        return Dynamic(redimmed, shared_data=True, streams=parent.streams,
+                       operation=dynamic_redim)
```
jlstevens: ``**dynkwargs`` for consistency with the rest?

holoviews/core/spaces.py
```diff
@@ -781,8 +781,8 @@ def _slice_bounded(self, tuple_key, data_slice):
                 if len(self):
                     slices = [slice(None) for _ in range(self.ndims)] + list(data_slice)
                     sliced = super(DynamicMap, sliced).__getitem__(tuple(slices))
-                return Dynamic(sliced, operation=lambda obj: obj[data_slice],
-                               shared_data=True)
+                return Dynamic(sliced, operation=lambda obj, **kwargs: obj[data_slice],
+                               streams=self.streams, shared_data=True)
```
jlstevens: And the two above ``**kwargs`` -> ``**dynkwargs``

holoviews/core/dimension.py
philippjfr: Sure, used dynkwargs because kwargs was sometimes in the namespace.

holoviews/util.py
```diff
@@ -47,26 +47,30 @@ class Dynamic(param.ParameterizedFunction):
...
-                if inspect.isclass(stream) and issubclass(stream, Stream):
-                    stream = stream()
-                elif not isinstance(stream, Stream):
-                    raise ValueError('Stream must only contain Stream '
-                                     'classes or instances')
+            dmap = self._make_dynamic(map_obj, callback, streams)
+        return dmap
+
+
+    def _get_streams(self, map_obj):
```
jlstevens: Might be worth having a docstring here to explain the policy...

holoviews/util.py
```diff
@@ -47,26 +47,30 @@ class Dynamic(param.ParameterizedFunction):
...
+        streams = util.dimensioned_streams(map_obj) if isinstance(map_obj, DynamicMap) else []
```
jlstevens: ``dimstreams``? Calling it ``streams`` is confusing with ``self.p.streams``...

holoviews/util.py
```diff
@@ -47,26 +47,30 @@ class Dynamic(param.ParameterizedFunction):
...
+        for stream in self.p.streams:
+            if inspect.isclass(stream) and issubclass(stream, Stream):
+                stream = stream()
```
jlstevens: I remember this bit - we are instantiating streams from the classes but I've forgotten why...

holoviews/util.py
jlstevens: Ah nevermind. I see why you did this.

holoviews/util.py
philippjfr: Because ElementOperations can supply their own streams, which are instantiated when applied.

holoviews/util.py
philippjfr: This is merging the ``dimstreams`` and ``self.p.streams`` so I think it's correct.

holoviews/util.py
jlstevens: Why as classes though and not instances?

jlstevens:
Seems fine although I asked a few questions in the comments.

Do you still want to improve the exception messages and maybe add a few unit tests before merging?

holoviews/util.py
philippjfr: Added one.

jlstevens:
Tests are passing. Merging.

Issue #1311, opened 2017-04-18
Remove sanitize_HTML from display_hooks


jlstevens:
This is a very simple request: ``sanitize_HTML`` is used in the display hooks to return an HTML version of an objects repr. This isn't really the right thing to do - instead, the display hook should return None and let the repr be automatically generated by the proper notebook stdout machinery.

This is a 5 minute PR to implement. The only concern is that this easy fix might mean updating notebook tests.

jlstevens:
Fixed in #1311

PR #1316, opened 2017-04-18
Removed sanitize_HTML from display hooks

jlstevens:
Simple PR to address #1311. The changes are trivial but the notebook display tests may need updating.


jlstevens:
Looks like the tests passed on the first go!

Thinking about it, this is probably because we are using ``print(obj)`` in the notebooks and the HTML reprs would only happen on error conditions which we avoid.

Ready for review/merge.

philippjfr:
Makes sense, merging.

Issue #1237, opened 2017-03-30
Displaying a plot calls _execute_callback multiple times


jlstevens:
While implementing #1232, I noticed that simply displaying a DynamicMap with streams results in ``DynamicMap._execute_callback`` being called multiple times:

![image](https://cloud.githubusercontent.com/assets/890576/24501324/ed15943e-1540-11e7-8c74-194f313b03e1.png)

This is similar to a previous issue we had in the past and it is only called once if streams are not involved. The new memoization system insulates the callback from this issue but I do still consider this a bug, though it may not need to have high priority because memoization is working properly.

I'm assigning this to Philipp as I'm fairly sure it is something in the plotting system that cause this -
 note the screenshot above is using the matplotlib backend.

jlstevens:
This issue was addressed in PR #1328. Closing.

PR #1319, opened 2017-04-19
Restore generator and argument-free callable support

jlstevens:
This PR is clearly docs related: instead of apologizing for dropping support for generators, I thought we might as well support them again. ;-p

Essentially, I realized that you could have stream objects without stream parameters, allowing support for callbacks accepting no arguments:

image

And generators:

image

Very little code needed to be changed so I think this is a decent prototype of how generators can be supported with our new simplified DynamicMaps.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -1213,6 +1213,13 @@ class DynamicMap(HoloMap):
         raise NotImplementedError('Cannot add dimensions to a DynamicMap, '
                                   'cast to a HoloMap first.')
 
+    def next(self):
+        if self.callback.noargs:
+            return self[()]
+        else:
+            raise Exception('The next method can only be used for DynamicMaps using'
+                            'generators (or callables without arguments)')
+
     # For Python 2 and 3 compatibility
     __next__ = next
 
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -779,8 +779,9 @@ class DynamicMap(HoloMap):
         # Ensure the clone references this object to ensure
         # stream sources are inherited
         if clone.callback is self.callback:
-            clone.callback = clone.callback.clone(inputs=[self],
-                                                  link_inputs=True)
+            with util.disable_constant(clone):
+                clone.callback = clone.callback.clone(inputs=[self],
+                                                      link_inputs=True)
         return clone
 
 
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -582,7 +582,7 @@ class DynamicMap(HoloMap):
         The key dimensions of a DynamicMap map to the arguments of the
         callback. This mapping can be by position or by name.""")
 
-    callback = param.ClassSelector(class_=Callable, doc="""
+    callback = param.ClassSelector(class_=Callable, constant=True, doc="""
         The callable used to generate the elements. The arguments to the
         callable includes any number of declared key dimensions as well
         as any number of stream parameters defined on the input streams.
@@ -590,7 +590,7 @@ class DynamicMap(HoloMap):
         If the callable is an instance of Callable it will be used
         directly, otherwise it will be automatically wrapped in one.""")
 
-    streams = param.List(default=[], doc="""
+    streams = param.List(default=[], constant=True, doc="""
        List of Stream instances to associate with the DynamicMap. The
        set of parameter values across these streams will be supplied as
        keyword arguments to the callback when the events are received,
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -620,15 +620,14 @@ class DynamicMap(HoloMap):
                    'are not Stream instances: {objs}')
             raise TypeError(msg.format(objs = ', '.join('%r' % el for el in invalid)))
 
-        if isinstance(self.callback, Generator):
+
+        noargs = ArgSpec(args=[], varargs=None, keywords=None, defaults=None)
+        if self.callback.argspec == noargs:
+            prefix = 'DynamicMaps using generators (or callables without arguments)'
             if self.kdims:
-                raise Exception('DynamicMaps using generators can only be declared '
-                                'without key dimensions')
+                raise Exception(prefix + ' must be declared without key dimensions')
             if not (len(self.streams) == 1 and isinstance(self.streams[0], Next)):
-                raise Exception('DynamicMaps using Generators must be declared '
-                                'with a single Next() stream')
-            if util.stream_parameters(self.streams):
-                raise Exception('Generators can only be used with empty streams')
+                raise Exception(prefix +' must be declared using streams=[Next()]')
 
         self._posarg_keys = util.validate_dynamic_argspec(self.callback.argspec,
                                                           self.kdims,
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -523,8 +523,9 @@ class Generator(Callable):
     arguments and never memoize.
     """
 
-    callable = param.Parameter(default=None, constant=True, doc="""
-         The generator function being wrapped.""")
+    callable = param.ClassSelector(default=None, class_ = types.GeneratorType,
+                                   constant=True, doc="""
+         The generator that is wrapped by this Generator.""")
 
     @property
     def argspec(self):
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -15,7 +15,7 @@ from .layout import Layout, AdjointLayout, NdLayout
 from .ndmapping import UniformNdMapping, NdMapping, item_check
 from .overlay import Overlay, CompositeOverlay, NdOverlay, Overlayable
 from .options import Store, StoreOptions
-from ..streams import Stream
+from ..streams import Stream, Next
 
 class HoloMap(UniformNdMapping, Overlayable):
     """
@@ -622,9 +622,11 @@ class DynamicMap(HoloMap):
 
         if isinstance(self.callback, Generator):
             if self.kdims:
-                raise Exception('Generators can only be used without key dimensions')
-            if self.streams == []:
-                self.warning('Empty streams required to trigger generator')
+                raise Exception('DynamicMaps using generators can only be declared '
+                                'without key dimensions')
+            if not (len(self.streams) == 1 and isinstance(self.streams[0], Next)):
+                raise Exception('DynamicMaps using Generators must be declared '
+                                'with a single Next() stream')
             if util.stream_parameters(self.streams):
                 raise Exception('Generators can only be used with empty streams')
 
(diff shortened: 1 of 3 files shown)


jlstevens:
By the way, I can even get rid of the ``Next`` stream. It's definition is empty so you might as well use ``Stream`` itself.

jlstevens:
Actually, I don't think this PR should be merged until this bug is fixed: #1237.

It would be nice to have this for 1.7 but I don't know how tricky that issue is to rectify.

jlstevens:
It looks like issue #1328 will address  #1237 as well as simplify some nasty plotting code! That means this PR can be considered soon.

As I don't want this PR to break anything I'll make a comment every single change with my reasoning why it is ok.

holoviews/core/spaces.py
```diff
@@ -478,6 +478,8 @@ def clone(self, callable=None, **overrides):
 
 
     def __call__(self, *args, **kwargs):
+        # Nothing to do for callbacks that accept no arguments
+        if not args and not kwargs: return self.callable()
```
jlstevens: If the inner callable takes no arguments at all, there is nothing to do but call it and return. Memoization makes no sense as there is only an empty tuple to memoize on - the output is constant for all time so might as well use an element. There can be no streams to worry about at this level as there are no args or kwargs either.

holoviews/core/spaces.py
```diff
@@ -513,6 +515,25 @@ def __call__(self, *args, **kwargs):
...
+    arguments and never memoize.
+    """
+
+    callable = param.Parameter(default=None, constant=True, doc="""
+         The generator function being wrapped.""")
+
+    @property
+    def argspec(self):
+        from inspect import ArgSpec
+        return ArgSpec(args=[], varargs=None, keywords=None, defaults=None)
```
jlstevens: Same (empty) signature as for functions with no arguments. Not really used but worth keeping the API consistent. I'll move the inline import to the top of the file.

holoviews/core/spaces.py
```diff
@@ -513,6 +515,25 @@ def __call__(self, *args, **kwargs):
...
 
+
+class Generator(Callable):
+    """
+    Generators are considered a special case of Callable that accept no
+    arguments and never memoize.
+    """
+
+    callable = param.Parameter(default=None, constant=True, doc="""
+         The generator function being wrapped.""")
```
jlstevens: I don't believe param has a ``Generator`` parameter which is why I am using ``param.Parameter``. Maybe I could use a class selector over generator types?

holoviews/core/spaces.py
```diff
@@ -513,6 +515,25 @@ def __call__(self, *args, **kwargs):
...
+
+    @property
+    def argspec(self):
+        from inspect import ArgSpec
+        return ArgSpec(args=[], varargs=None, keywords=None, defaults=None)
+
+    def __call__(self):
+        return next(self.callable)
```
jlstevens: The ``next`` built-in takes no arguments so as before, there is nothing to do.

Actually that isn't *quite* right. In the Python help:

```
next(...)
    next(iterator[, default])
```

Where ``default`` is output instead of raising StopIteration when raised. I don't think I've ever seen ``default`` used and always raising ``StopIteration`` on exhaustion is what I expect.

holoviews/core/spaces.py
```diff
@@ -580,7 +601,10 @@ class DynamicMap(HoloMap):
        the cache is full.""")
 
     def __init__(self, callback, initial_items=None, **params):
-        if not isinstance(callback, Callable):
+
+        if isinstance(callback, types.GeneratorType):
+            callback = Generator(callback)
+        elif not isinstance(callback, Callable):
             callback = Callable(callback)
```
jlstevens: Trivial conditional switch between ``Callable`` and ``Generator`` - a specific type of ``Callable``.

holoviews/core/spaces.py
```diff
@@ -590,6 +614,15 @@ def __init__(self, callback, initial_items=None, **params):
 
         super(DynamicMap, self).__init__(initial_items, callback=callback, **params)
         invalid = [s for s in self.streams if not isinstance(s, Stream)]
+
+        if isinstance(self.callback, Generator):
+            if self.kdims:
+                raise Exception('Generators can only be used without key dimensions')
```
jlstevens: The ``kdims`` would have nowhere to go.

holoviews/core/spaces.py
jlstevens: Note that:

1. There is an *empty* stream expected on the DynamicMap but no *stream parameters*.
2. The user is responsible for initializing the generator with appropriate state.

holoviews/core/spaces.py
```diff
@@ -590,6 +614,15 @@ def __init__(self, callback, initial_items=None, **params):
...
+            if self.streams == []:
+                self.warning('Empty streams required to trigger generator')
```
jlstevens: We still need an *empty* stream without *any* stream parameters to trigger the generator to trigger ``next`` via ``dmap.event``. Without it, nothing can change after first display.

jlstevens: I haven't explicitly forced it to be only a single *empty* stream instance but I could. Surely lots of empty stream instance are equivalent (0+0+.. = 0).

holoviews/core/spaces.py
```diff
@@ -590,6 +614,15 @@ def __init__(self, callback, initial_items=None, **params):
...
+            if util.stream_parameters(self.streams):
+                raise Exception('Generators can only be used with empty streams')
```
jlstevens: Making sure there are no stream parameters across all the supplied streams i.e they are all empty.

holoviews/core/spaces.py
```diff
@@ -654,6 +687,7 @@ def _validate_key(self, key):
         Make sure the supplied key values are within the bounds
         specified by the corresponding dimension range and soft_range.
         """
+        if key == () and len(self.kdims) == 0: return ()
```
jlstevens: No key and no kdims means there is nothing to do. Had to skip ``util.wrap_tuple`` as it wraps the empty tuple in another tuple.

holoviews/core/spaces.py
```diff
@@ -856,7 +890,8 @@ def __getitem__(self, key):
         try:
             dimensionless = util.dimensionless_contents(get_nested_streams(self),
                                                         self.kdims, no_duplicates=False)
-            if dimensionless:
+            empty = util.stream_parameters(self.streams) == [] and self.kdims==[]
```
jlstevens: No streams or kdims implies it is empty i.e if valid, a callable without arguments or a generator.

holoviews/core/spaces.py
```diff
@@ -856,7 +890,8 @@ def __getitem__(self, key):
...
+            if dimensionless or empty:
```
jlstevens: Never cache for these two types of DynamicMap.

holoviews/streams.py
```diff
@@ -267,6 +267,14 @@ def __str__(self):
         return repr(self)
 
 
+class Next(Stream):
+    """
+    Example of an empty stream that can be used to trigger generators or
+    callbacks that take no arguments.
+    """
+    pass
```
jlstevens: Not strictly necessary as you could use ``Stream``. Another possible name could be ``Trigger``.

jlstevens:
Gone through and added all my explanations. Do you agree with them?

I think the fact that argument-free callables and generators cannot have streams means everything should nest ok - I don't expect any bad consequences involving ``inputs`` ... and I'm not sure there can be any.

The only other nagging question is what happens when a generator is exhausted and StopIteration is raised.

jlstevens:
I could also implement ``__next__`` and ``next`` methods on ``DynamicMap`` that only work if ``isinstance(self, Generator)``. They would just call ``self.event()``.

I *could* automatically add a ``Stream``/``Next``/``Trigger`` empty stream to streams if I detect a generator but I'm not sure if this is necessarily a good idea. It would make documenting generator support in DynamicMap easier though.

holoviews/core/spaces.py
philippjfr: Makes sense.

holoviews/core/spaces.py
philippjfr: Yes a ClassSelector might be good if ``types.GeneratorType`` works.

holoviews/core/spaces.py
philippjfr: Can you use it with a ``Counter`` stream?

holoviews/core/spaces.py
philippjfr: Makes sense.

holoviews/core/spaces.py
jlstevens: Just to be completely clear: There can be no stream *parameters* to worry about at this level. This becomes obvious in my comments below.

holoviews/core/spaces.py
philippjfr: >Had to skip util.wrap_tuple as it wraps the empty tuple in another tuple.

It shouldn't be doing that, it should be checking if the input is already a tuple and skip in that case.

holoviews/core/spaces.py
jlstevens: No as the ``Counter`` stream has a single stream parameter called ``counter``. That would become an argument so wouldn't apply to argument-free functions/generators.

You *could* imagine Streams without parameters but attributes. E.g a type of ``Counter`` that tracks how many times it is triggered. Not sure why we would want that though.

holoviews/core/spaces.py
jlstevens: Ok, I'll look into that.

jlstevens:
I'm now going to rebase to grab the fixes in  #1237 which is now merged.

jbednar:
Do you mean the fixes in #1328 ?  #1237 is still open.  Can #1237 now be closed?

jlstevens:
@jbednar That is correct - I meant PR #1328 and I'll close #1237 now.

jlstevens:
Just a note, generators aren't considered callable so ``Generator.callable`` is a bit of a misnomer. It isn't worth changing to a different name.

If we could think of a good name that covers both callables *and* generators, I would consider a search and replace. @jbednar @philippjfr Any thoughts?

jlstevens:
I was considering automatically adding a single ``Next()`` stream for generators. In the end, I just decided to enforce this condition: we almost dropped generator support entirely and in the end I think it is better to be consistent than to try and implement 'helpful' behavior for a use case we think will be rarely used.

jlstevens:
Taking the logic that consistency is more important than trying to be helpful, I've enforced the use of the ``Next()`` stream for anything without arguments (generators or callables). If the user is going to do something unusual, it is probably best not to hide the extra bit of machinery from them.

jlstevens:
I've now implemented everything I want for this PR. Ready for review when the tests pass.

jlstevens:
@philippjfr Travis seems to be particular slow to run test builds today. Probably worth reviewing without waiting for the tests.

philippjfr:
Great to be able support this again. Looks good to me now.

jlstevens:
Thanks!

I'm also happy we were able to support this in a much simpler framework without worrying about 'modes'.

PR #1320, opened 2017-04-19
Consistent event method on DynamicMap and Stream

jlstevens:
This PR just makes triggering events consistent between ``DynamicMap`` and ``Stream``: use the ``event`` method which always triggers.

Ideally the default of ``Stream.update`` would be ``trigger=False`` but that might have some backwards compatibility implications. Namely, anyone using ``stream.update`` as a callback would have to change it to ``stream.event`` to get the same behavior.


jlstevens:
Talking to @jbednar and @philippjfr we decided that we can remove the ``trigger`` from ``update`` entirely (i.e triggering is always False).

jlstevens:
@philippjfr Ready for review once the tests pass.

Now we just need to make sure to update ``stream.update`` to ``stream.event`` as appropriate.

philippjfr:
Looks good, happy to merge when tests pass.

PR #1323, opened 2017-04-19
Do not change mpl interactive mode in notebook

### LINKED ISSUE #1314: holoviews.notebook_extension() overrides %matplotlib inline
screen shot 2017-04-18 at 20 44 55

```
%matplotlib inline 
import holoviews
import matplotlib.pyplot as plt
holoviews.notebook_extension()
```

philippjfr:
When I implemented support for using the matplotlib backend outside a notebook context I added some code which fiddles with the interactive mode of pyplot, which interacts badly with the inline mode in the notebook. This PR ensures that when the notebook_extension loads the matplotlib backend a flag is set that ensures interactive mode is not disabled.

Fixes #1314

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -152,6 +152,7 @@ class notebook_extension(param.ParameterizedFunction):
             if p in self._backends:
                 imports.append((p, self._backends[p]))
         if not imports:
+            args = ['matplotlib']
             imports = [('matplotlib', 'mpl')]
 
         args = list(args)
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -221,6 +221,7 @@ class notebook_extension(param.ParameterizedFunction):
         resources = list(resources)
         if len(resources) == 0: return
 
+        Renderer.load_nb()
         for r in [r for r in resources if r != 'holoviews']:
             Store.renderers[r].load_nb(inline=p.inline)
 
--- a/holoviews/plotting/mpl/renderer.py
+++ b/holoviews/plotting/mpl/renderer.py
@@ -93,9 +93,6 @@ class MPLRenderer(Renderer):
              'nbagg':   (NbAggJupyterComm, None),
              'mpld3':   (JupyterComm, mpld3_msg_handler)}
 
-    # Whether in a notebook context
-    _notebook = False
-
     def __call__(self, obj, fmt='auto'):
         """
         Render the supplied HoloViews component or MPLPlot instance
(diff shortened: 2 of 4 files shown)


jbednar:
Seems like an important fix.  We should presumably have a "Working with matplotlib" tutorial showing what the possible workflows are, including %inline, so that we can preserve this as allowed and expected behavior.

holoviews/plotting/mpl/renderer.py
```diff
@@ -318,3 +321,4 @@ def load_nb(cls, inline=True):
         """
```
jlstevens: As the bokeh renderer also has a ``load_nb`` classmethod (and notebooks are explicitly mentioned in that method name), I would consider making ``_notebook`` into a constant ``notebook_context`` class parameter. The docstring can say it is set to True if the ``load_nb`` classmethod has been called.

holoviews/ipython/__init__.py
```diff
@@ -152,6 +152,7 @@ def __call__(self, *args, **params):
             if p in self._backends:
                 imports.append((p, self._backends[p]))
         if not imports:
+            args = ['matplotlib']
```
jlstevens: Was missing this out a bug? I assume this is what makes matplotlib the default if you do ``hv.notebook_extension()``.

jlstevens:
Looks good. 

I've made one comment and asked one question but the approach seems fine.

holoviews/ipython/__init__.py
philippjfr: Yes, this was a bug, matplotlib would be imported but ``MPLRenderer.load_nb`` was never called.

holoviews/plotting/mpl/renderer.py
philippjfr: I assumed you meant a class attribute since that makes sense to me. I've now moved that handling onto the  Renderer baseclass, so that all subclasses share the attribute.

jlstevens: An attribute is fine but I was thinking of a parameter. Maybe a property makes the most sense as it is read-only by default?

philippjfr: A parameter doesn't make much sense to me because it means it can be changed at the instance level which doesn't seem quite correct. I think it's fine as is tbh because I'd still have to make the property settable somehow.

jlstevens: I agree that a property would involve an underscore attribute then the property itself (public API). I don't mind really, a public attribute is fine too.

jlstevens:
Tests have passed. Merging.

PR #1324, opened 2017-04-19
Send numeric bokeh Table data unformatted

philippjfr:
Currently we are formatting all the ``Table`` cells with the ``Dimension.pprint_value`` method before they are sent to display in bokeh. As pointed out in https://github.com/ioam/holoviews/issues/1317, this means that floats and integers are treated as strings, which means they do not sort correctly. Additionally sending floats and integers is more efficient because they can make use of base64 serialization.


jlstevens:
Looks like the Bokeh tests will need updating - not too surprising!

jlstevens:
Tests are passing. Merging.

PR #1326, opened 2017-04-19
Link inputs fixes

philippjfr:
The main issue fixed here is that linked_inputs was being inherited when making a clone, when a clone should always link_inputs. I've also ensured that the decimate operation links inputs and fixed a small bug while hooking up the stream sources.

Fixes: https://github.com/ioam/holoviews/issues/1325


jbednar:
Sounds great, thanks!

jlstevens:

PR #1328, opened 2017-04-20
Ensure Plots correctly cache lookups in DynamicMaps

philippjfr:
Ensures the plots respect the _force option telling them not to request a new frame on the DynamicMap, both during initialization and for subsequent updates. Relies on the fact that the initialization pass of creating Plots does not care which exact Element is being plotted, since it is subsequently updated again anyway.

### DIFF
--- a/holoviews/core/traversal.py
+++ b/holoviews/core/traversal.py
@@ -60,7 +60,22 @@ def unique_dimkeys(obj, default_dim='Frame'):
         dims = merge_dimensions(dim_groups)
         all_dims = sorted(dims, key=lambda x: dim_groups[0].index(x))
     else:
-        all_dims = [default_dim]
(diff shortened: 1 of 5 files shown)


philippjfr:
I'm really happy to have finally simplified this logic, I thoroughly hated ``get_dynamic_item`` and I've not replaced it completely. We now also explicitly throw an exception when trying to display a composite plot containing multiple HoloMaps with non-overlapping dimensions. Another happy side effect is that DynamicMaps with non-overlapping dimensions now work:

nonoverlap

jlstevens:
Looks great and I am also really happy to see this simplified.

The tests have passed now. Merge?

philippjfr:
Yes, ready now.

jlstevens:
The push builds have passed and the failures in the pr builds are just transient display test errors. Merging.

PR #1329, opened 2017-04-20
Avoid overriding %matplotlib inline in notebook

philippjfr:
We were previously overriding the default backend because the all defaults were loaded from the rc file. The notebook_extension now also doesn't override the currently set backend if it is already valid for the notebook.


jlstevens:
Looks good. Happy to merge when the tests go green.

Feel free to add any unit tests if you want...

jlstevens:
Tests are green. Merging.

PR #1330, opened 2017-04-20
Silenced expected warning from the rolling_outlier_std operation

jbednar:
The `rolling_outlier_std` operation first computes a rolling window, using a calculation that leaves NaN values at the beginning and end of the series (depending on the size of the window).  `np.abs()` was raising a warning ``invalid value encountered in absolute``, but this warning is just distracting because invalid values are expected in this case, so this PR ensures that the warning is silenced.


jlstevens:
Looks good to me!

jlstevens:
Tests are green. Merging.

PR #1333, opened 2017-04-20
Added support for batching plot updates

philippjfr:
Adds precedences to Stream subscribers, which allow updates to individual subplots to be batched. Each subplot now adds itself as a subscriber along with the top-level plot. When a subplot is updated it marks itself with an ``_update`` flag so that the top-level plot can then dispatch the necessary updates once the subplot updates have completed.

* [x] Ensure GridPlot/LayoutPlot titles are updated
* [x] Ensure that batched OverlayPlots aren't getting updated multiple times.

### DIFF
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -14,7 +14,7 @@ from ...core.util import basestring, wrap_tuple, unique_iterator
 from ...element import Histogram
 from ..plot import (DimensionedPlot, GenericCompositePlot, GenericLayoutPlot,
                     GenericElementPlot)
-from ..util import get_dynamic_mode
+from ..util import get_dynamic_mode, attach_streams
 from .renderer import BokehRenderer
 from .util import (bokeh_version, layout_padding, pad_plots,
                    filter_toolboxes, make_axis, update_shared_sources)
@@ -292,10 +292,11 @@ class GridPlot(CompositePlot, GenericCompositePlot):
                                        ranges=ranges, keys=keys, **params)
         self.cols, self.rows = layout.shape
         self.subplots, self.layout = self._create_subplots(layout, ranges)
-        top_level = keys is None
-        if top_level:
+        if self.top_level:
             self.comm = self.init_comm()
             self.traverse(lambda x: setattr(x, 'comm', self.comm))
+            self.traverse(lambda x: attach_streams(self, x.hmap, 1),
+                          [GenericElementPlot])
 
 
     def _create_subplots(self, layout, ranges):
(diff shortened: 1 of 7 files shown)


holoviews/streams.py
```diff
@@ -162,15 +163,15 @@ def reset(self):
...
+    def add_subscriber(self, subscriber, precedence=0):
         """
         Register a callable subscriber to this stream which will be
         invoked either when event is called or when this stream is
         passed to the trigger classmethod.
         """
         if not callable(subscriber):
             raise TypeError('Subscriber must be a callable.')
-        self._subscribers.append(subscriber)
+        self._subscribers.append((precedence, subscriber))
```
jlstevens: Wouldn't ``(subscriber, precedence)`` be a more natural order?

philippjfr: Doesn't sort nicely though.

holoviews/streams.py
```diff
@@ -90,9 +90,10 @@ def trigger(cls, streams):
             param.main.warning('Parameter name clashes for keys: %r' % clashes)
 
         # Currently building a simple set of subscribers
-        groups = [stream.subscribers for stream in streams]
-        subscribers = util.unique_iterator([s for subscribers in groups
-                                            for s in subscribers])
+        groups = [stream._subscribers for stream in streams]
+        sorted_groups = sorted([s for subscribers in groups
+                                for s in subscribers])
```
jlstevens: Ah, you might need a key function here if you switch the order.

holoviews/streams.py
jlstevens: Right, you would need to use a key function in the sort. I realized this before I saw your comment...

holoviews/plotting/bokeh/plot.py
```diff
@@ -292,10 +292,11 @@ def __init__(self, layout, ranges=None, layout_num=1, keys=None, **params):
                                        ranges=ranges, keys=keys, **params)
         self.cols, self.rows = layout.shape
         self.subplots, self.layout = self._create_subplots(layout, ranges)
-        top_level = keys is None
-        if top_level:
+        if self.top_level:
             self.comm = self.init_comm()
             self.traverse(lambda x: setattr(x, 'comm', self.comm))
+            self.traverse(lambda x: attach_streams(self, x.hmap, 1),
```
jlstevens: So a precedence of 1.0 for 'top level'?

philippjfr: Right, I just need two levels, any value higher than 1 will be fine.

jlstevens:
Looks good!

I'm leave it up to you if you rather store as ``(precedence, subscriber)`` which is slightly less natural on its own but is correctly sortable. I would also rename ``_update`` to ``_updated``.

Otherwise it looks straightforward and I think it is a usability improvement. We might still want ``Stream.batch_event`` and ``Stream.batch_update`` to trigger across all the streams of an object but that is separate from this PR.

philippjfr:
This should be ready, there's already a bunch of plotting unit tests which failed on my first attempt and that should pass now.

jlstevens:
I'll merge when the tests go green.

PR #1334, opened 2017-04-20
Disabled batched plot optimization

philippjfr:
This optimization was a nice idea as it would avoid sending ColumnDataSource columns which all contain the same value and it might be possible to get it to work but it doesn't deal with empty elements well currently and I don't want to fiddle more before release.


jlstevens:
Ok. Just make sure to file an issue to re-enable and fix after the release.

jlstevens:
Tests have passed. Merging.

PR #1335, opened 2017-04-20
Flip dynspread output correctly

philippjfr:
The output is currently flipped along the y-axis.


jbednar:
I just tested it, and it works now. Thanks!  Seems ok to merge since this operation remains untested?

jlstevens:
The push build is passing. Merging.

PR #1337, opened 2017-04-20
Fix for static_source optimization

philippjfr:
Since glyph styles depend on the data especially on batched plots the static_source optimization on bokeh plots can result in the wrong colors being applied. This fix means the data will be processed but not sent.


jlstevens:
Ready to merge when the tests pass?

jlstevens:
Looks good. Merging.

PR #1342, opened 2017-04-21
Expose edgecolors for 3D surface plots

basnijholt:
Because the default style in mpl 2.0 was changed from 'k' to 'w' I exposed the `edgecolors` style option to change it back if wanted.


philippjfr:
Seems like a sensible addition that should have been there in the first place. Will merge when tests pass.

PR #1343, opened 2017-04-21
Handle empty subplots and overlay layers

philippjfr:
This PR adds some additional validation to the plotting code to check for empty objects, which are then skipped rather than causing errors somewhere downstream in the code.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -594,6 +594,8 @@ class LabelledData(param.Parameterized):
         # Assumes composite objects are iterables
         if self._deep_indexable:
             for el in self:
+                if el is None:
+                    continue
                 accumulator += el.traverse(fn, specs, full_breadth)
                 if not full_breadth: break
         return accumulator
--- a/holoviews/core/traversal.py
+++ b/holoviews/core/traversal.py
@@ -91,7 +91,7 @@ def unique_dimkeys(obj, default_dim='Frame'):
 
     with item_check(False):
         sorted_keys = NdMapping({key: None for key in unique_keys},
-                                kdims=all_dims_emptied).data.keys()
+                                kdims=all_dims).data.keys()
     return all_dims, list(sorted_keys)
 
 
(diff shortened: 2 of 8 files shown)


jlstevens:
Looks like it might be ready? The tests are passing.

If you could add another unit test for this, that would be great but if it is too tricky to bother with then nevermind.

philippjfr:
That last commit is a bit misnamed, bokeh now handles hv.Empty plots correctly and both backends treat empty objects as Empty except they also issue a warning.

jbednar:
Glad to see Empty being handled more uniformly!  Just curious -- why is having an Empty generating a warning?  Isn't that a valid thing for someone to add to help get the layout they want?

jlstevens:
Tests are passing. If you are happy with it, I am ready to merge.

philippjfr:
Not yet no, further fixes coming.

jlstevens:
> why is having an Empty generating a warning? Isn't that a valid thing for someone to add to help get the layout they want?

I was about the ask the same thing!

philippjfr:
Sorry that was confusing and relied on the distinction between empty objects and Empty objects to parse. I just meant objects that have no length e.g. empty NdOverlays and HoloMaps will show a warning.

jlstevens:
Ok. That is what I hoped you meant.

philippjfr:
It turns out we weren't handling setting the visible flag on missing frames correctly, fixed that and am currently adding some unit tests.

philippjfr:
When a HoloMap or DynamicMap key does not exist the corresponding glyphs are now hidden correctly. I also added some unit tests. Ready to merge once I've rebuilt the bokeh test data, which now includes the ``visible`` flag on the GlyphRenderer.

philippjfr:
No eference data update needed apparently, just a py3 bug in one of my new unit tests. Ready to merge when tests pass.

jlstevens:
Tests are green. Merging.

PR #1344, opened 2017-04-21
Added RollingBase for consistency between rolling and rolling_outlier_std

jbednar:
With a shared base class, it's simpler to set parameters that apply to both these operations.  This change also makes rolling_outlier_std slightly more powerful, because users can now control more of the options involved. 

Also made docstring fixes and fixed a bug where rolling was meant to have special behavior depending on the window_type but was checking the window length instead.

### DIFF
--- a/holoviews/operation/timeseries.py
+++ b/holoviews/operation/timeseries.py
@@ -7,40 +7,48 @@ from ..core.data import PandasInterface
 from ..element import Scatter
 
 
-class rolling(ElementOperation):
+class RollingBase(param.Parameterized):
     """
-    Applies a function over a rolling window.
+    Parameters shared between `rolling` and `rolling_outlier_std`.
     """
 
     center = param.Boolean(default=True, doc="""
         Whether to set the x-coordinate at the center or right edge
         of the window.""")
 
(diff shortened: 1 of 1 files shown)


holoviews/operation/timeseries.py
```diff
@@ -104,9 +113,9 @@ def _process_layer(self, element, key=None):
         ys = element.dimension_values(1)
 
         # Calculate the variation in the distribution of the residual
-        avg = pd.Series(ys).rolling(window, center=True).mean()
+        avg = pd.Series(ys).rolling(**self._roll_kwargs()).mean()
         residual = ys - avg
-        std = pd.Series(residual).rolling(window, center=True).std()
+        std = pd.Series(residual).rolling(**self._roll_kwargs()).std()
```
philippjfr: This will blow up when a window_type is specified, not sure what we can do about that.

philippjfr:
Looks good now, will merge once it passes.

holoviews/operation/timeseries.py
jbednar: Good point; apparently .std() is only defined for a rectangular window, so I've moved the window type specification into `rolling` itself.

PR #1348, opened 2017-04-22
Fixed JS logic to check for comm_manager

philippjfr:
Currently when you reload a notebook with a comm in it, displaying the plot will fail because at time of execution a kernel might not yet exist and the logic to check for it is not quite correct.


philippjfr:
Ready to merge.

PR #1349, opened 2017-04-22
Small fixes for dynamic select

philippjfr:
There was two small issues with select, dynamic overlaying was not specifying which objects it wants to be selecting on and secondly Datasets weren't validating selections applied to them.

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -316,7 +316,10 @@ class Dataset(Element):
         supplied, which will ensure the selection is only applied if the
         specs match the selected object.
         """
-        if selection_specs and not any(self.matches(sp) for sp in selection_specs):
+        selection = {dim: sel for dim, sel in selection.items()
+                     if dim in self.dimensions()+['selection_mask']}
+        if (selection_specs and not any(self.matches(sp) for sp in selection_specs)
+            or not selection):
             return self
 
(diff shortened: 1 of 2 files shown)


holoviews/core/data/__init__.py
```diff
@@ -316,7 +316,10 @@ def select(self, selection_specs=None, **selection):
         supplied, which will ensure the selection is only applied if the
         specs match the selected object.
         """
-        if selection_specs and not any(self.matches(sp) for sp in selection_specs):
+        selection = {dim: sel for dim, sel in selection.items()
+                     if dim in self.dimensions()+['selection_mask']}
```
jlstevens: What is the special string 'selection_mask'? I grepped the code base and it doesn't currently appear anywhere...

jlstevens:
Looks good and the tests are passing. Happy to merge if you say it is ready.

holoviews/core/data/__init__.py
philippjfr: It's used a bunch in the interfaces and allows applying a boolean mask in ``__getitem__``.

philippjfr:
Ready to merge.

holoviews/core/data/__init__.py
jlstevens: Ah, it just doesn't seem to appear elsewhere in the codebase as a string like this...

PR #1350, opened 2017-04-22
Fix Callback comm caching and empty events

philippjfr:
This PR fixes two nasty bugs, the first is in caching comms which previously did not work and a new comm was created for each and every event. The second is PlotSize events sending zero sizes which in some cases could cause kernel crashes either due to datashader or somewhere in the plotting code choking on the zero sized array.


jlstevens:
Looks good. 

Glad we caught these bugs before releasing! Constantly creating comms would have been a nasty thing to leave in...

Merge when the tests go green?

philippjfr:
Yes, ready to merge. This was quite a pain to track down.

PR #1351, opened 2017-04-22
Miscellaneous fixes

jlstevens:
This PR contains an assortment of fixes for issues I encountered while updating the documentation. I will update the descriptions of what I've fixed as I go:

* Added some unit tests for swapping positional kdim arguments when listed by name.
* Unbounded DynamicMaps no longer display regardless of cache size. This is more consistent with the old behavior and less confusing.
* Applying redim on values now filters the DynamicMap cache accordingly. Some nasty errors occurred without this.
* Validated HLines and VLine constructors by declaring appropriate parameters.
* Fixed clone on annotations accordingly.
* Generalized the x,y parameters of annotations to support time types.
* Improved Callable names and improved exception handling in Callables.
* Enforced that OperationCallables have a supplied operation.

### DIFF
--- a/holoviews/element/annotation.py
+++ b/holoviews/element/annotation.py
@@ -186,8 +186,6 @@ class Arrow(Annotation):
             self.data = (x, y, text, direction, points, arrowstyle)
 
 
-        return self.__class__(*args, **settings)
-
     def dimension_values(self, dimension, expanded=True, flat=True):
         index = self.get_dimension_index(dimension)
         if index == 0:
--- a/holoviews/element/annotation.py
+++ b/holoviews/element/annotation.py
@@ -1,7 +1,8 @@
+from numbers import Number
 import numpy as np
-
 import param
 
+from ..core.util import datetime_types
 from ..core import Dimension, Element2D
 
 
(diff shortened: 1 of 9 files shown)


holoviews/element/annotation.py
```diff
@@ -165,14 +186,6 @@ def __setstate__(self, d):
...
 
-    # Note: This version of clone is identical in Text and path.BaseShape
-    # Consider implementing a mix-in class if it is needed again.
-    def clone(self, *args, **overrides):
-        if len(args) == 1 and isinstance(args[0], tuple):
-            args = args[0]
-        arg_names = ['x', 'y', 'text', 'direction', 'points', 'arrowstyle']
-        settings = {k: v for k, v in dict(self.get_param_values(), **overrides).items()
-                    if k not in arg_names[:len(args)]}
         return self.__class__(*args, **settings)
```
philippjfr: Maybe I'm confused by the diff but shouldn't this last line be removed as well?

jlstevens: I was wondering where that line went! Fixed in 9c7a1ea9

jlstevens:
In the end I just commented out the unit test relying on ioam/param#149 and replaced it with a check for any sort of exception. It should be uncommented once that fix is released.

I'm done with my current set of fixes and I think this PR is ready for review when the tests go green.

philippjfr:
Could you add hv.core.util.basestring to the Text x, y values. This is valid:

```
hv.Text('A', 1, 'A') * hv.Points([('A', 1)])
```

screen shot 2017-04-23 at 1 44 30 am

I've even used it in one of the examples in Bokeh_Backend.

jlstevens:
Implemented your suggestion in 5cca1a9b

philippjfr:
Looks good to me, merging.

PR #1352, opened 2017-04-23
Fixes for the PlotSize stream

philippjfr:
I accidentally left the PlotSize stream in my previous PR disabled, I also set the default width/height to None so the user specified datashade operation width/height end up overriding it on initialization.


jlstevens:
Looks fine. Merge?

philippjfr:
Yes, ready.

PR #1353, opened 2017-04-23
Stream clear method now supports policies by precedence

jlstevens:
This PR improves the clear method to make it easier for users to use their own subscribers and clear them without clobbering the subscribers installed by HoloViews. Currently the default policy is 'all' but I might consider making it 'user' instead.


jlstevens:
@philippjfr Ready for review/merge.

philippjfr:
Looks good, merging.

PR #1355, opened 2017-04-23
Added support for constructing empty Dataset types

philippjfr:
All Dataset interfaces except the deprecated NdElement interface now support constructing an empty dataset of the right shape when given an empty list. If you can think of another specification that should also mean empty that would be trivial to add. I think this capability is important when working with streams because often your selection, streamed data or whatever else is empty to begin with and constructing an empty Element can often be a pain.

### DIFF
--- a/holoviews/core/data/array.py
+++ b/holoviews/core/data/array.py
@@ -45,6 +45,8 @@ class ArrayInterface(Interface):
                 data = np.column_stack(data)
             else:
                 raise ValueError('ArrayInterface expects data to be of uniform shape.')
+        elif isinstance(data, list) and data == []:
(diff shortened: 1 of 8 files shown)


jlstevens:
Looks good. 

I would like unit tests for all possible empty constructors though (i.e across data types).

philippjfr:
>I would like unit tests for all possible empty constructors though (i.e across data types).

That's the beauty of Dataset tests, you get a unit test per data type for every one you write.

jlstevens:
Great! Happy to merge if you think it is ready and the tests go green.

philippjfr:
Ready to merge.

PR #1356, opened 2017-04-24
Added Streams.define to easily define custom streams

jlstevens:
I'm very excited about this PR as I think it is a huge improvement in ease-of-use for users as I suspect custom streams will be very common.

Users will no longer need to start defining class in notebooks (which is ugly!) when they require a custom stream. Compare this:

image

to this:

image

This uses ``type`` to dynamically create a streams class and figures out an appropriate parameter type and sets the value as the default. For a more complete parameter definition, you can do:

image

Some programmers might object to this dynamic class creation as being too magical (though really, the implementation of ``define`` is trivial) and I personally wouldn't recommend people use this in library code - they should define a normal ``Stream`` subclass in the usual way.

For notebook use where you want to be concise and avoid code and for anyone not familiar with Python OOP, I think this will be highly appreciated.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -15,7 +15,7 @@ from .layout import Layout, AdjointLayout, NdLayout
 from .ndmapping import UniformNdMapping, NdMapping, item_check
 from .overlay import Overlay, CompositeOverlay, NdOverlay, Overlayable
 from .options import Store, StoreOptions
-from ..streams import Stream, Next
+from ..streams import Stream
 
 class HoloMap(UniformNdMapping, Overlayable):
     """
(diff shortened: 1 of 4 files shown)


jlstevens:
Note that this would allow us to delete almost all our non linked-stream definitions: ``Next``, ``Counter``, ``X``, ``Y`` and ``XY``. We would no longer have to anticipate what common stream parameters people might want as this will make it trivial for them to define whatever they want as they need it.

jbednar:
Sounds cool, but can you provide a slightly more complex example that actually uses the fact that it's a stream or dynamic?  This example doesn't seem to achieve anything over what the original object does; what does the custom stream achieve (simple as it now is to specify)?

jlstevens:
This PR is used by the tutorial update PR referenced above. I think it is now ready to merge.

philippjfr:
Looks good and we all agree it's useful. Going to merge.

PR #1365, opened 2017-04-25
Added initial draft of Linked_Streams tutorial

philippjfr:
Can be revised later, main addition I want to make is adding links to multiple examples for each stream type.


jlstevens:
I'll merge it right away. 

I do think it needs more polish but that is true of all our new tutorials. It will be helpful to have the new notebooks on master.

PR #1366, opened 2017-04-25
Updated submodule reference to fix Comms error in docbuilding

philippjfr:


jlstevens:
It is just a submodule reference update. Merging.

PR #1368, opened 2017-04-25
Added hv.examples to copy notebooks locally

jlstevens:
There may be a better approach but this stuff is very tricky. I'm fairly sure this approach will work:

1 - Set ``pypi_version(True) `` only when building for PyPI.
2 - Get conda to copy the notebooks to a path in the same way [datashader does it](https://github.com/bokeh/datashader/blob/master/conda.recipe/build.sh).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -35,7 +35,7 @@ install:
       conda install python=3.4.3;
     fi
   - rm -r $SITE_PACKAGES/numpy*
-  - python setup.py install
+  - python setup.py develop
   - pip install git+git://github.com/ioam/param.git
   - pip install path.py
   - pip install coveralls awscli deepdiff
--- a/holoviews/__init__.py
+++ b/holoviews/__init__.py
@@ -86,21 +86,3 @@ def help(obj, visualization=True, ansi=True, backend=None,
         print((msg if visualization is False else '') + info)
     else:
         pydoc.help(obj)
-
-
-def examples(path='holoviews-examples', verbose=False, force=False):
-    """
-    Copies the notebooks to the supplied path.
-    """
-    filepath = os.path.abspath(os.path.dirname(__file__))
-    example_dir = os.path.join(filepath, './examples')
-    if not os.path.exists(example_dir):
-        example_dir = os.path.join(filepath, '../examples')
-    if os.path.exists(path):
-        if not force:
-            print('%s directory already exists, either delete it or set the force flag' % path)
-            return
-        shutil.rmtree(path)
-    ignore = shutil.ignore_patterns('.ipynb_checkpoints','*.pyc','*~')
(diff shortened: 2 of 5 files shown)


holoviews/__init__.py
```diff
@@ -86,3 +87,23 @@ def help(obj, visualization=True, ansi=True, backend=None,
...
+
+
+def examples(path='holoviews-examples', verbose=False):
+    """
+    Copies the notebooks to the supplied path.
+    """
+    import os, glob
+    from shutil import copytree, ignore_patterns
+
+    candidates = [os.path.join(__path__[0], '../doc/Tutorials/'),   # local doc available
```
jbednar: "local doc available" means git?  If so maybe say so.

setup.py
```diff
@@ -83,26 +83,29 @@ def check_pseudo_package(path):
             raise Exception("Please make sure pseudo-package %s is populated." % path)
 
 
-if __name__=="__main__":
+def pypi_version(pypi=False):
+    if not pypi: return
+    # Make sure to create these directories and populate them before upload
```
jbednar: Not sure what that comment means; surely there is no manual step involved?

jlstevens: Right now there is a manual step. Populating the pseudo packages is also a manual step.

I'm all in favor of replacing the manual steps with something else... as long as it works!

holoviews/__init__.py
jlstevens: Yes, I mean git. I'll update shortly.

setup.py
jbednar: Right.  I don't see any reason to have  a manual step!

philippjfr:
Any idea what these errors are about? Somehow option IDs are now wrong: https://travis-ci.org/ioam/holoviews/jobs/225631022

jlstevens:
@philippjfr Thanks for doing this and I do think it is the right approach.

As I don't think our notebooks are quite ready to ship in the current state, I'm assigning this to the 1.8 milestone.

jlstevens:
I've reassigned this to 1.7.1 as the notebooks and examples should be presentable at that point.

philippjfr:
Will have to update this once we have all the assets and examples in place.

.travis.yml
```diff
@@ -35,7 +35,7 @@ install:
       conda install python=3.4.3;
     fi
   - rm -r $SITE_PACKAGES/numpy*
-  - python setup.py install
```
jlstevens: Why did this need to change?

philippjfr: So it skips installing the examples mainly. If you use ``develop`` then the examples are already in a place where it will find them.

jlstevens: Makes sense, thanks.

jlstevens:
There seems to be an issue with having to run ``python setup.py sdist`` twice but otherwise this seems to be working. Merging.

Issue #1373, opened 2017-04-26
Suggestion: ignore options if not used by backend


basnijholt:
Right now the following happens:

<img width="1092" alt="screen shot 2017-04-26 at 12 58 50" src="https://cloud.githubusercontent.com/assets/6897215/25431265/23403c5e-2a80-11e7-9681-cb899025db3b.png">

Sometimes I change `hv.notebook_extension('matplotlib')` to `hv.notebook_extension('bokeh')` temporarily to zoom. 

Right now I need to comment out all the options that are incompatible.

It would be nice to still see the plots and just print a warning.

philippjfr:
Just use ``hv.notebook_extension('matplotlib', 'bokeh')``, it will pick the first backend in the list but load all of them and validation applies to all **loaded** backends.

jlstevens:
@basnijholt Are there any particular reasons why you wouldn't want to load both bokeh and matplotlib at the top, even if you only use one at a time?

Philipp's suggestion is correct and it should work once both backends are loaded. This issue might be something we can refer to if other people bump into this while we update the docs.

@philippjfr Have we documented this and if we haven't where should we?

philippjfr:
>@philippjfr Have we documented this and if we haven't where should we?

No I don't think this has been properly documented, not sure where would be best.

jbednar:
A release page would be a great idea.  I do think we can do a little bit just on the release notes, to fill in the "backwards compatibility" section with practical implications.

That said, I don't know why the plot isn't being shown even when there are incompatible options; why is that not just a warning?

jlstevens:
> That said, I don't know why the plot isn't being shown even when there are incompatible options; why is that not just a warning?

It is probably too late to make it into a warning now - the behavior is similar to the validation we had previously before we turned it off.

The issue might be that the current message isn't quite clear enough. In @basnijholt screenshot above you have: 'Possible keywords in the currently active 'bokeh' backend are: ....

How can this be rephrased to make the correct fix easier to discover? The message is technically correct but it would have been better if it could have pointed to ``hv.notebook_extension('matplotlib', 'bokeh')`` as the solution.

Something along the lines of:

> Invalid keyword for the currently active 'bokeh' backend, please make sure that all the required backends are loaded in the notebook extension. Possible keywords in the loaded backends are: ....

The other thing I want to fix in the screenshot above is the ``repr`` - it is showing the raw repr and not being formatted correctly. I'll open an issue about that.

[...]

PR #1375, opened 2017-04-26
Added strict class attribute to OptsMagic

jlstevens:
This PR addresses the issue raised in #1373 

With ``strict=True``:

image

And ``strict=False`` (default):

image

I used ``sys.stderr`` because the magic classes are not parameterized and the root param logger adds very little of value and would make this type of warning a lot uglier than necessary.

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -353,7 +353,6 @@ class Options(param.Parameterized):
         for kwarg in sorted(kwargs.keys()):
             if allowed_keywords and kwarg not in allowed_keywords:
                 if self.skip_invalid:
-                    kwargs.pop(kwarg)
                     invalid_kws.append(kwarg)
                 else:
                     raise OptionError(kwarg, allowed_keywords)
--- a/holoviews/ipython/magics.py
+++ b/holoviews/ipython/magics.py
@@ -664,16 +664,21 @@ class OptsMagic(Magics):
             loaded=' in loaded backends {0} and {1!r}'.format(backend_list,
                                                             loaded_backends[-1])
 
+        suggestion = ("If you believe this keyword is correct, please make sure "
+                      "the backend hasbeen imported or loaded with the "
+                      "notebook_extension.")
+
         group = '{0} option'.format(err.group_name) if err.group_name else 'keyword'
         msg=('Unexpected {group} {kw} {target}{loaded}.<br><br>'
              '{similarity} keywords in the currently active '
-             '{current_backend} backend are: {matches}')
+             '{current_backend} backend are: {matches}<br><br>{suggestion}')
         return msg.format(kw="'%s'" % err.invalid_keyword,
                           target=target,
                           group=group,
                           loaded=loaded, similarity=similarity,
                           current_backend=repr(Store.current_backend),
-                          matches=matches)
+                          matches=matches,
+                          suggestion=suggestion)
 
     @classmethod
     def register_custom_spec(cls, spec):
--- a/holoviews/ipython/magics.py
+++ b/holoviews/ipython/magics.py
@@ -665,7 +665,7 @@ class OptsMagic(Magics):
                                                             loaded_backends[-1])
 
         suggestion = ("If you believe this keyword is correct, please make sure "
-                      "the backend hasbeen imported or loaded with the "
+                      "the backend has been imported or loaded with the "
                       "notebook_extension.")
 
         group = '{0} option'.format(err.group_name) if err.group_name else 'keyword'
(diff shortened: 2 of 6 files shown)


jbednar:
Looks good to me, thanks!

basnijholt:
This is exactly what I would liked 👍

philippjfr:
You said you didn't like the red warning, do you want to just use ``IPython.display`` to render the HTML warning or are you okay with it now?

jlstevens:
I have a pretty negative association with warnings and that red color - mostly due to ipywidgets, numpy warnings you can't do anything about etc. I do think it looks ugly but maybe that is a good thing: it means users should fix incorrect options so we never have to see them!

I can use this PR as a opportunity to update the message. What would be a good way to help make sure the user has loaded all the backends they want?

jlstevens:
I'm now filtering plot options at the ``Plot.lookup_options`` level. I'm now wondering if Options should ever be dropping their kwargs (the parameter is ``skip_invalid`` not ``drop_invalid``).

I'm going to push a commit that stops options from popping invalid kwargs.  If the tests pass we can think more carefully about whether it is the right thing to do (I suspect so).

jlstevens:
@philippjfr Ok the tests passed with my last commit. So..

1. Are you happy with these changes? All filtering is now done by ``Plot.lookup_options`` instead of some of it, with some filtering by options.
2. Are we happy with the current message? Maybe a line at the end to suggest loading more backends? 'If you believe this keyword is correct, please make sure the backend is loaded using notebook_extension.' Something like that?

Maybe @jbednar has a better suggestion for the latter?

philippjfr:
>Are you happy with these changes? All filtering is now done by Plot.lookup_options instead of some of it, with some filtering by options.

I'm going to have to go through the code to double check options are never accessed directly elsewhere.

>Are we happy with the current message? Maybe a line at the end to suggest loading more backends? 'If you believe this keyword is correct, please make sure the backend is loaded using notebook_extension.' Something like that?

This warning will also appear outside a notebook context right? So I suppose it should be:

>If you believe this keyword is correct, please make sure the backend has been imported or loaded with the notebook_extension.

jlstevens:
> This warning will also appear outside a notebook context right?

I'm not sure it will. 

The ``OptsMagic`` generates this message and only the ``OptsMagic`` currently uses the ``StoreOptions.validate_spec`` method to do this kind of validation.

philippjfr:
>The OptsMagic generates this message and only the OptsMagic currently uses the StoreOptions.validate_spec method to do this kind of validation.

So what happens when using the ``__call__``, no validation at all?

jlstevens:
Right, ``__call__`` simply uses:

```python
StoreOptions.set_options(deep_clone, options, **kwargs)
```

The validation you'll get are the option errors in ``skip_invalid`` is False and the ``allowed_keywords`` are set - which given what I found out above, might not be working.

jlstevens:
Note that even if we added the same sort of validation to ``__call__`` (which would be a valid feature request, just not for this particular PR) then the message would be generated at that point. 

As the current validation only happens in the ``OptsMagic``, the message we are currently discussing will only apply to notebooks.

jbednar:
Seems like that message could appear in other contexts if we add explicit Opts parsing via python calls to the parser, so Philipp's version seems less confusing to me.

jlstevens:
I would agree if it were ``StoreOptions.validate_spec`` that generates the message. Instead it generates an ``OptionError`` which is then formatted and output appropriately at the point where it is caught. 

Right now this is in the opts magic with the message discussed in this PR and in future it could be in the ``__call__`` method where it could output a similar message but one that is agnostic to the notebook. I don't mind using Philipp's suggestion, it is just that I don't think the ' backend has been imported' bit is relevant currently.

jlstevens:
I've used Philipp's suggestion as there is no harm in including the extra bit which may be good advice if we add validation to ``__call__`` later. Here is what it looks like:

image

The wrapping in the warning is a bit annoying but that is probably something that varies from browser to browser and monitor to monitor.

philippjfr:
Could you have a look at ``MPLRenderer.plot_options`` and ``BokehRenderer.plot_options``? I think they need extra handling now.

philippjfr:
It's possible they can just use ``Plot.lookup_options`` as well.

jlstevens:
> It's possible they can just use Plot.lookup_options as well.

That probably makes the most sense.

jlstevens:
Using ``Plot.lookup_options`` from the renderers seems fine with nosetests at least. Let's see whether it is also ok with the notebook tests.

philippjfr:
Looks like it will pass, merging.

Issue #1369, opened 2017-04-25
Options tests teardown not working properly


jlstevens:
In PR #1368 we found that copying tests into the main package directory (so that the tests run twice) causes option failures. This means the setup/teardown of option tests isn't quite right.

This isn't high priority as we only intend to run the tests once but it does indicate a bug.

jlstevens:
This issue was fixed in the PR referenced above (#1377). Closing.

PR #1377, opened 2017-04-26
Fixed issue with option test teardown

jlstevens:
Fixes this bad interaction in the tests and it might fix part/all of the issue in #1369:

image

@philippjfr I only cleared the custom option trees where I needed to in order to fix the tests. Should I do it in all the option testing classes to be safe?

### DIFF
--- a/tests/testmagics.py
+++ b/tests/testmagics.py
@@ -16,6 +16,7 @@ class ExtensionTestCase(IPTestCase):
         self.ip.run_line_magic("load_ext", "holoviews.ipython")
 
     def tearDown(self):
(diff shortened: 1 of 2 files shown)


jlstevens:
With the last commit fixing the teardown of magic tests, I can confirm that #1369 is resolved. There are still a few failures when I duplicate tests before that commit but they all pass afterwards.

Ready to review/merge.

jbednar:
Looks good to me!

philippjfr:
Looks good to me too, can merge once tests pass but we probably need to do this everywhere we apply custom options.

PR #1384, opened 2017-04-28
Polishing and improving the tutorials added in 1.7

jlstevens:
This PR aims to improve our new tutorials before 1.7.1 is released, hopefully getting the improved versions of the tutorials live before then. The tutorials to improve are: ``DynamicMap, Streams`` and ``Linked_Streams`` and they are expected to be read in that order.

I think any of us can use this PR to make the improvements we feel are necessary. Obviously anything that might be controversial should be discussed first!

My first commit improves the listing of available linked stream classes.


doc/Tutorials/Linked_Streams.ipynb
```diff
@@ -16,24 +16,24 @@
...
+    "Currently only the bokeh backend for HoloViews supports the linked streams system but the principles used should extend to any backend can define callbacks that fire when a user zooms or pans or interacts with a plot.\n",
     "\n",
     "<center><div class=\"alert alert-info\" role=\"alert\">To use and visualize <b>DynamicMap</b> or <b>Stream</b> objects you need to be running a live Jupyter server.<br>This tutorial assumes that it will be run in a live notebook environment.<br>\n",
     "When viewed statically, DynamicMaps will only show the first available Element,<br></div></center>\n",
     "\n",
     "## Available Linked Streams\n",
     "\n",
-    "There a huge number of ways one might want to interact with a plot. The HoloViews streams module aims to offer many of the most common interactions you might want to specify while making it easy to extend it with custom linked Streams. \n",
+    "There are a huge number of ways one might want to interact with a plot. The HoloViews streams module aims to expose many of the most common interactions you might want want to engage in, while also supporting extensibility via ustom linked Streams. \n",
```
philippjfr: "engage in" seems awkward, maybe "utilize" or "employ"

Also "ustom linked Streams"

jlstevens: 'employ' sounds good and I'll fix the typo now.

jlstevens:
I've updated the notebook format version because the diffs were getting too miserable - all because of a *minor* version bump in the format.

jlstevens:
I am skipping over the 'Unlinking streams' section and going through the rest of the tutorial. I think that section needs a better example to illustrate why you might want to unlink.

jlstevens:
> However if it is a ``DynamicMap`` we have to make a copy with the ``Dynamic`` utility and declare ``link_inputs=False``.

@philippjfr  Maybe ``clone`` should support a ``link_inputs`` argument everywhere such that it does nothing for normal elements but works appropriately for ``DynamicMap``?

jlstevens:
I've updated the last section:

* Transient events are not covered in the basic Streams tutorial. I think introducing them in Linked Streams makes sense.
* I updated the code in the example, using ``if None not in [x,y]:`` instead of ``if x and y:``. This is both more explicit (and makes more sense with the text) and more correct as ``x=0`` and ``y=0`` are perfectly valid (but falsey) stream parameter values.

jlstevens:
I'm done making edits to the linked streams tutorial for now. There are two things I would like to see:

1. A better example for unlinking streams.
2. A better conclusion, maybe ending with a nice example.
3. Possibly an example showing selection?

I'm happy to review the tutorial and have another pass once these things are added.

jlstevens:
I've now replaced the crosshairs with a better example of dimensioned streams:

image

There are two things to mention though:

* The bit about casting to HoloMap needs dimension defaults to be added (to be more interesting, otherwise the area has no width)
* This could then motivate the linked streams tutorial, if @philippjfr would like to include the linked version:

```python
xs = np.linspace(-np.pi, np.pi, 40)

def section(x,t=1):
    print x
    x = 0 if x is None else x
    fn = lambda xs, t: np.exp(np.sin(xs+np.pi/t))
    split = np.where(xs>x)[0][0]
    return (hv.Area((xs[:split], fn(xs[:split], t)))
            * hv.Curve((xs[split:], fn(xs[split:], t))) * hv.VLine(x))

hv.DynamicMap(section, kdims=['t'], 
              streams=[Stream.define('time', t=1)(), streams.PointerX()])
```

@philippjfr It might be nice to make such a link between the tutorials. Do you have a use for this example in the linked streams tutorial?

jlstevens:
I will rebase now so I can make edits to the new operation tutorial if necessary.

jlstevens:
From commit 290fcd57 onwards I will be assuming that ``ElementOperation`` will be renamed to ``Operation`` i.e 1421 will be merged.

jlstevens:
In addition I am assuming issue #1386 will be addressed having removed all mention of ``key`` argument in ``_process``.

jlstevens:
I'm done making edits on the Operations tutorial for now. I am very happy with the examples it uses and I think the main issue now is merging the required PRs I mention above.

jlstevens:
In my last commit, I add a new section about what to do when you don't have an operation that maps over elements: I recommend always using ``param.ParameterizedFunction`` and documented appropriately.

jlstevens:
Note that I added ``np.clip(limit,-3,3)`` to the new linked streams 'integral' example. This behaves better but may be unnecessary if we decide to limit linked pointer position values to the axis bounds - which I think we should.

@philippjfr @jbednar @ceball This PR is ready now, all feedback on these three new tutorials (Streams, Linked Streams, Operations) is welcome!

jlstevens:
As one final note, adding the following code (thanks @philippjfr !) is the same as applying ``np.clip``:

```python
clipjs = "if (x_range.type.endsWith('Range1d')) { if (cb_obj.x < x_range.start) { data['x'] = x_range.start } else if (cb_obj.x > x_range.end) { data['x'] = x_range.end }}"
hv.plotting.bokeh.callbacks.PointerXYCallback.code = clipjs
hv.plotting.bokeh.callbacks.PointerXYCallback.models.append('x_range')
```

We can decide whether to make this the default (and tidy up the code) in a separate PR.

doc/Tutorials/Linked_Streams.ipynb
```diff
@@ -138,131 +151,130 @@
...
+    "pointer = streams.PointerXY(source=img)\n",
+    "pointer_dmap = hv.DynamicMap(lambda x, y: hv.Points([(x, y)]), streams=[pointer])"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Now try hovering over the ``Image`` above again and watch the cross-sections update as you move the cursor."
+    "Now if we display a ``Layout`` consisting of the ``Image`` acting as the source together with the ``DynamicMap``, the point shown on the right tracks the cursor position when hovering voer the image on the left:"
```
philippjfr: Type: hovering **over**

philippjfr:
Looks great, all good improvements. Merging.

Issue #1379, opened 2017-04-27
Documentation plan


philippjfr:
A major part of improving our docs is to add various examples. The way many libraries structure their examples is in a top-level examples directory (see [bokeh](https://github.com/bokeh/bokeh/tree/master/examples), and [matplotlib](https://github.com/matplotlib/matplotlib/tree/master/examples)). These examples then usually get built into a gallery (see [bokeh](http://bokeh.pydata.org/en/latest/docs/gallery.html), [matplotlib](http://matplotlib.org/gallery.html), and [cartopy](http://scitools.org.uk/cartopy/docs/latest/gallery.html)).

Adding an examples directory also encourages adding examples when new features are added. I'd even suggest that in future new features should be accompanied by small example notebooks or scripts. We are going to be splitting out the different plotting backends but I'd still strongly argue examples for officially supported backends should live on the core repo, where they are all in one place.

We've also had various definitions of examples in the past, what I think it should mean in this context is small self-contained notebooks with at most one or two examples, which are focused on the code, not on telling a story or explaining deeper concepts. That contrasts with quickstart guides, tutorials and the "examples" that are linked to from [holoviews.org/Examples](http://holoviews.org/Examples/index.html), which are really case studies. My suggestion for the different types of notebooks:

1. Tutorials - Long, detailed notebooks explaining particular concepts in detail, living in ``doc/Tutorials``. New tutorials should be added to ``holoviews-contrib`` and can move to the main repo once polished.
2. Quickstart guides - Shorter notebooks getting the user started on using a particular feature without providing extensive background. Again should start out in ``holoviews-contrib`` but once we have a few I'd suggest creating a User Guide (see [bokeh](http://bokeh.pydata.org/en/latest/docs/user_guide.html)) that provides a quick introduction to holoviews.
3. Examples - These are what this issue is about, they are short and self-contained and generally should just go straight into the main repo since they don't need detailed explanation.
4. Case studies - These are what's currently on [holoviews.org/Examples](http://holoviews.org/Examples/index.html), and basically show how to apply holoviews to a particular domain or problem. I believe these should all live in holoviews-contrib providing a wide-ranging collection of user examples. Keeping them all in one place this way will encourage us to test and update them for each release.

If we agree on these different formats and where they live we should settle the structure of the examples, my suggestion is that each example should be implemented for all backends that support the features used in the example. Then each example links to the equivalent versions for other backends. Each example should contain the following information:

[...]

PR #1385, opened 2017-04-28
Added various linked Stream examples

philippjfr:
A number of relatively simple linked stream examples with synthetic data. It might be nice to also have more complex examples which use real data. Feedback on structure of the examples welcome. Also worth considering whether I should add gifs to these notebooks so you can see what they're meant to do when viewed statically.


jlstevens:
When I saw the commit message 'Added gifs for streams examples' I thought you added the gif files to the PR - which I would have objected to.

Links to GIFs are ok though. :-)

jlstevens:
I've tried out all the examples and I like them!

``HeatMap_Tap`` has an issue with ``df`` being undefined and I think we still need to discuss exactly where these examples go. We need to figure out how to structure the elements gallery, these examples, what happens to contrib etc and also make sure it is a long term plan (i.e after the plotting backends have been split out, how we will support these examples with a live server eventually etc).

jlstevens:
We decided on an organization here in issue #1379. That said, I would like to see that other material added to examples first at which point I would be happy to include these examples too.

philippjfr:
Sounds good, there's still a few selection examples I wanted to add and we all agreed that this one was a good demo to show what datashader really does:

datashade_reveal

jlstevens:
> HeatMap_Tap has an issue with df being undefined 

I ran through this notebook again and this still seems to be an issue..

jlstevens:
I've rebased this PR and gone through the examples. They all work properly except for HeatMap_Tap which is missing the DataFrame ``df``. I believe we plan to add some real data to assets but I don't think that should be included just yet. Once the code is update, I'll add the compressed CSV file locally to test, at which point I'm happy to merge.

Maybe we should start adding asset files to a PR?

jlstevens:
I've now updated the 'HeatMap Tap' example and I think I'm ready to merge. 

Here is the disease.csv.gz file we will want to add to ``examples/assets`` later: [disease.csv.gz](https://github.com/ioam/holoviews/files/1069297/disease.csv.gz)

PR #1387, opened 2017-04-29
Reverse map renamed stream parameters in dynamic operations

philippjfr:
Streams can inherit parameters from dynamic operations when their values are None. However this currently breaks for renamed parameters because the operation and stream parameters do not match in those cases. This PR ensures that renamed stream parameters are reverse mapped fixing the bug.

### DIFF
--- a/holoviews/util.py
+++ b/holoviews/util.py
@@ -74,7 +74,8 @@ class Dynamic(param.ParameterizedFunction):
                 updates = {k: self.p.operation.p.get(k) for k, v in stream.contents.items()
                            if v is None and k in self.p.operation.p}
                 if updates:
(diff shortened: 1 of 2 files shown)


philippjfr:
Ready to review/merge.

jbednar:
The title of this issue wins today's Most Difficult to Parse award! Only the most deeply versed in HoloViews lingo will be able to distinguish it from a random string of words.  Congratulations! Well done! :-)

jlstevens:
@jbednar Don't worry, I had to ask Philipp what it was all about too!

Anyway, now I understand it, I am happy to merge. The main issue here is (as we already know!) we need a *lot* more documentation.

Merging.

PR #1399, opened 2017-05-03
Fixed bug in sorting on array interface

philippjfr:
It appears that sorting on the ArrayInterface was broken in certain cases, this creates a recarray view to perform the sorting on, which should work more robustly. I'll add a couple of tests before this is ready to merge.

### DIFF
--- a/holoviews/core/data/array.py
+++ b/holoviews/core/data/array.py
@@ -102,9 +102,10 @@ class ArrayInterface(Interface):
         if len(by) == 1:
             sorting = cls.values(dataset, by[0]).argsort()
         else:
-            dtypes = (dataset.data.dtype,)*dataset.data.shape[1]
-            sort_fields = tuple('f%s' % dataset.get_dimension_index(d) for d in by)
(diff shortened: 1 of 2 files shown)


philippjfr:
Ready to review. Note that the grid interfaces don't support sorting in general and neither does dask, hence the "Not supported" unit tests.

jlstevens:
Using a record array like that seems a bit like a big hammer but I can see why it might be more reliable. Anyway, looks good and the tests are passing. Merging.

philippjfr:
>Using a record array like that seems a bit like a big hammer but I can see why it might be more reliable.

Since it's just a view it should be fairly lightweight, and looking back at it I think it was using a record array without explicit naming before anyway. Seems to be the only way to sort by column.

jlstevens:
For lexical sort? Right, I've used record arrays for that before...

philippjfr:
>For lexical sort? Right, I've used record arrays for that before...

Right.

PR #1400, opened 2017-05-03
Fixed py3 bug mapping bokeh Table values

philippjfr:
Table values get mapped to strings using the d.pprint_value function but in py3 ``map`` returns a ``map`` object which has to be expanded to a list before it can be passed as a column to the table.


jlstevens:
Makes sense...been a while since I've seen the ``map`` built-in in actual code!

philippjfr:
Ready to merge.

PR #1401, opened 2017-05-03
Fixed bug in HeatMap aggregation when no global ordering found

philippjfr:
When generating a HeatMap the aggregation code attempts to figure out a global ordering for the x-axis and y-axis categorical factors by considering the partial orderings of the y-coordinates for each x-coordinate using topological sort. Finding a global ordering in this way can fail in which case the entire HeatMap fails to display. If no global ordering can be found this now simply sorts the y-coordinates ensuring something is displayed.

### DIFF
--- a/holoviews/element/util.py
+++ b/holoviews/element/util.py
@@ -155,7 +155,8 @@ class categorical_aggregate2d(ElementOperation):
         if sort or one_to_one(orderings, ycoords):
             ycoords = np.sort(ycoords)
         elif not is_cyclic(orderings):
-            ycoords = list(itertools.chain(*sort_topologically(orderings)))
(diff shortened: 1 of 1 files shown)


philippjfr:
This should really be merged before 1.8. I'll try to put together a unit test for it tomorrow.

jlstevens:
Seems reasonable to have a fallback. I'm happy to merge once the unit test(s) are added and go green.

philippjfr:
@jlstevens I'm having trouble reproducing the error now, but I know I've run into it a bunch of times on real datasets. Also possible this was fixed in another way but having a fallback won't hurt anyway. Would you mind merging?

jlstevens:
Looks good. Merging.

PR #1405, opened 2017-05-04
Fixed bokeh BarPlot implementation

philippjfr:
We've been discussing the Bars element quite a lot recently and that its behavior was confusing. Having had to use Bars extensively I now realize the reason for that is that the current implementation doesn't really make any sense and is completely inconsistent with what the matplotlib version does.

While we will deprecate this bokeh charts plot implementation in favor of our own very soon, I still think this is worth fixing for the time being. In future we have plans to allow grouping and stacking of Bars by overlaying them. However as we move away from a bokeh charts based interface the behavior it has now is probably reasonable for a transitional period while we solve issues surrounding categorical axes in the matplotlib backend and issues surrounding nested coordinate systems in bokeh. The matplotlib plot should still be simplified to disallow simultaneous grouping and stacking, resulting in a horrendously complicated plotting class.

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -10,6 +10,7 @@ from bokeh.models import ( GlyphRenderer, ColumnDataSource,
                           Range1d, CustomJS, HoverTool)
 from bokeh.models.tools import BoxSelectTool
 
+from ...core.dimension import Dimension
 from ...core.util import max_range, basestring, dimension_sanitizer
 from ...core.options import abbreviated_exception
 from ...core.spaces import DynamicMap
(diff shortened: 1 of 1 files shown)


jlstevens:
I'm all in favor of simplifying the barplot implementations. Happy to merge if you think it is ready and the tests pass.

jlstevens:
The main worry is probably backwards compatibility...

philippjfr:
>The main worry is probably backwards compatibility...

I've made sure that it still behaves as before for the simple case and for the stacking case. Any other combinations were simply wrong and buggy before (probably part of the reason you complained about it being unintuitive before).

philippjfr:
I've opened a PR to summarize my conclusions about Bars.

jlstevens:
Not sure there are any new unit tests to write. If you are done, I'm happy to merge as the tests are green.

philippjfr:
>Not sure there are any new unit tests to write. If you are done, I'm happy to merge as the tests are green.

Bars are tested as part of the Bokeh_Elements tutorial.

Issue #1410, opened 2017-05-05
Pandas 0.20.1 compatibility


philippjfr:
Pandas 0.20.1 has been released with some [backward incompatible changes](http://pandas.pydata.org/pandas-docs/stable/release.html). Based on a quick scan of the changelog the deprecation of ``.ix`` stands out to me as we use it in ``.select``. This should be a trivial fix, but I'll have to do some testing to see if there's more issues.

ea42gh:
Issue with holoviews/pandas/dask
--------------------------------------------------
----> 6 import holoviews as hv
/usr/lib/python3.5/site-packages/holoviews/core/__init__.py in <module>()
----> 2 from .data import *            # noqa (API import)
/usr/lib/python3.5/site-packages/holoviews/core/data/__init__.py in <module>()
---> 11 from ..dimension import redim
/usr/lib/python3.5/site-packages/holoviews/core/dimension.py in <module>()
---> 14 from ..core.util import (basestring, sanitize_identifier,
     15                          group_sanitizer, label_sanitizer, max_range,
     16                          find_range, dimension_sanitizer, OrderedDict,
/usr/lib/python3.5/site-packages/holoviews/core/util.py in <module>()
---> 32     import dask.dataframe as dd
/usr/lib/python3.5/site-packages/dask/dataframe/__init__.py in <module>()
      1 from __future__ import print_function, division, absolute_import
----> 3 from .core import (DataFrame, Series, Index, _Frame, map_partitions,
      4                    repartition, to_delayed)
/usr/lib/python3.5/site-packages/dask/dataframe/core.py in <module>()
---> 40 pd.computation.expressions.set_use_numexpr(False)
AttributeError: module 'pandas' has no attribute 'computation'

ea42gh:
The problem was resolve by upgrading dask to version 0.14.3

ea42gh:
holoviews/core/util.py:27: FutureWarning: pandas.tslib is deprecated

jlstevens:
@philippjfr Is this purely a pandas/dask issue that we need to be aware of but otherwise has nothing to do with our end?

philippjfr:
>Is this purely a pandas/dask issue that we need to be aware of but otherwise has nothing to do with our end?

That dask ``pandas.computation`` error is the remainder I've just opened a PR for.

philippjfr:
All the required fixes were merged. Closing.

PR #1412, opened 2017-05-06
Fixes for pandas 0.20(.1) compatibility

philippjfr:
Addresses #1410 avoiding imports from ``pandas.tslib`` and replacing a usage of ``df.ix`` which have both been deprecated.

### DIFF
--- a/holoviews/core/__init__.py
+++ b/holoviews/core/__init__.py
@@ -27,7 +27,7 @@ Dimension.type_formatters[np.datetime64] = '%Y-%m-%d %H:%M:%S'
 
 try:
     import pandas as pd
-    Dimension.type_formatters[pd.tslib.Timestamp] = "%Y-%m-%d %H:%M:%S"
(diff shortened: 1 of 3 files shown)


philippjfr:
Ready to merge.

jlstevens:
Looks good.

Before I merge, I do want to ask how far back (in pandas versions) these changes will work with. If we now require the pandas version to be more recent than some release, we should put that in setup.py and meta.yml.

philippjfr:
>Before I merge, I do want to ask how far back (in pandas versions) these changes will work with.

I've gone back to 0.16.1 which was released in May 2015 without issues. My guess is ``pd.Timestamp`` has been a top-level import since it was introduced and ``.iloc`` would always have worked.

jlstevens:
Sounds good. 

A few people might still be using a version of pandas older than 2 years but it won't be a big worry. Merging.

PR #1413, opened 2017-05-06
Various fixes for bokeh static_source optimization

philippjfr:
In https://github.com/ioam/holoviews/issues/1396 it was suggested that the ``static_source`` optimization in bokeh should only kick in if both the element and its data are static.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -14,11 +14,11 @@ import param
 from ..core.util import (basestring, sanitize_identifier,
                          group_sanitizer, label_sanitizer, max_range,
                          find_range, dimension_sanitizer, OrderedDict,
-                         bytes_to_unicode, unicode, dt64_to_dt, unique_array)
+                         bytes_to_unicode, unicode, dt64_to_dt, unique_array,
+                         builtins)
 from .options import Store, StoreOptions
 from .pprint import PrettyPrinter
 
-obj_id = id
 # Alias parameter support for pickle loading
 
 ALIASES = {'key_dimensions': 'kdims', 'value_dimensions': 'vdims',
@@ -489,7 +489,7 @@ class LabelledData(param.Parameterized):
         """
         self.data = data
         self.id = id
-        self._plot_id = plot_id or obj_id(self)
+        self._plot_id = plot_id or builtins.id(self)
         if isinstance(params.get('label',None), tuple):
             (alias, long_name) = params['label']
             label_sanitizer.add_aliases(**{alias:long_name})
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -18,6 +18,7 @@ from ..core.util import (basestring, sanitize_identifier,
 from .options import Store, StoreOptions
 from .pprint import PrettyPrinter
 
+obj_id = id
 # Alias parameter support for pickle loading
 
 ALIASES = {'key_dimensions': 'kdims', 'value_dimensions': 'vdims',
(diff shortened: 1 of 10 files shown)


jlstevens:
Looks good.

I can't really see how to test this right now so I'll merge once the tests pass.

philippjfr:
So this approach does not work since it just disables the optimization entirely. The issue is twofold a) when a custom style is applied the element id changes, b) something in the plotting pipeline applies a dynamic method to the data, which again ends up changing the element id. Therefore the plotting code never sees the same element id twice and the optimization does nothing. Will have to think about this.

jlstevens:
Problem b) seems like the real issue to me as I don't think disabling the optimization for a) means it will always be disabled.

holoviews/core/dimension.py
```diff
@@ -488,6 +489,7 @@ def __init__(self, data, id=None, **params):
         """
         self.data = data
         self.id = id
+        self._plot_id = plot_id or obj_id(self)
```
jlstevens: Just use ``__builtins__.id`` to avoid the name shadowing issue.

holoviews/core/dimension.py
```diff
@@ -488,6 +489,7 @@ def __init__(self, data, id=None, **params):
         """
         self.data = data
         self.id = id
```
jlstevens: Do we ever actually set the ``id`` via the constructor?

jlstevens: Ah, probably for clone. Nevermind my question!

holoviews/plotting/bokeh/element.py
```diff
@@ -804,15 +804,16 @@ def update_frame(self, key, ranges=None, plot=None, element=None, empty=False):
         # Cache frame object id to skip updating data if unchanged
         previous_id = self.handles.get('previous_id', None)
         if self.batched:
-            current_id = sum(element.traverse(lambda x: id(x.data), [Element]))
+            current_id = sum(element.traverse(lambda x: x._plot_id, [Element]))
         else:
-            current_id = id(element.data)
+            current_id = element._plot_id
```
jlstevens: Is it no longer taking into account the ``id`` of the data? It seems to do that in tabular...

holoviews/plotting/bokeh/plot.py
```diff
@@ -147,7 +147,7 @@ def sync_sources(self):
         from the same object.
         """
         get_sources = lambda x: (id(x.current_frame.data), x)
-        filter_fn = lambda x: (x.shared_datasource and x.current_frame and
+        filter_fn = lambda x: (x.shared_datasource and x.current_frame is not None and
```
jlstevens: Just to confirm: is this a a fix unrelated to the optimization itself?

holoviews/plotting/bokeh/tabular.py
```diff
@@ -73,11 +73,12 @@ def current_handles(self):
         if self.static and not self.dynamic:
             return handles
 
+
+        element = self.current_frame
         previous_id = self.handles.get('previous_id', None)
-        current_id = id(self.current_frame.data) if self.current_frame else None
+        current_id = None if self.current_frame is None else id(element)+id(element.data)
```
jlstevens: Summing ``ids`` is not a safe test for uniqueness as many pairs of ids could sum to the same value (even if it is unlikely!). Note that collisions might not even be *that* rare as the ``id`` doesn't return a hash - it is more like  to a memory address.

To check the identity of both the element and the data I would use the tuple ``(id(element), id(element.data))`` and not the sum.

holoviews/core/dimension.py
philippjfr: Yes, clone does all the time.

holoviews/plotting/bokeh/plot.py
philippjfr: Seems so, not sure why I pushed it here, but it is an important fix.

jlstevens:
Looks good with the latest fixes. One last import issue to fix and then I'll merge once the tests go green. That said, are there any meaningful unit tests we should set up to test the optimization?

jlstevens:
Thanks for adding the unit tests! Merging.

Issue #1403, opened 2017-05-03
Bokeh marker orientation API has changed


philippjfr:
API for setting the marker orientation seems to have changed this no longer works:

```
hv.Points(([1,2],[3,4]))(style=dict(marker='^'))
```

>**AttributeError** [Call holoviews.ipython.show_traceback() for details]
unexpected attribute 'orientation' to Triangle, possible attributes are angle, angle_units, fill_alpha, fill_color, js_event_callbacks, js_property_callbacks, line_alpha, line_cap, line_color, line_dash, line_dash_offset, line_join, line_width, name, size, subscribed_events, tags, x or y

Presumably needs to be updated to use ``angle``.

jlstevens:
The PR referenced above that fixes this has been merged. Closing.

PR #1415, opened 2017-05-06
Fix for change in bokeh marker orientation API

philippjfr:
Fix for #1403, the API to change the bokeh marker orientation changed a while back and the aliases defining certain marker types was never updated.

### DIFF
--- a/holoviews/plotting/bokeh/util.py
+++ b/holoviews/plotting/bokeh/util.py
@@ -33,14 +33,14 @@ from ..util import dim_axis_label, rgb2hex
 # Conversion between matplotlib and bokeh markers
 markers = {'s': {'marker': 'square'},
            'd': {'marker': 'diamond'},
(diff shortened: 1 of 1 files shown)


jlstevens:
>  ... the API to change the bokeh marker orientation changed a while back

Do you know when exactly? Probably doesn't matter too much as we require the latest bokeh version for stream support anyhow.

As this is a utility, I imagine that this could be tested with a unit test or two e.g make to make sure the conversion is accepted without raising an error.

philippjfr:
I can't work out when it was ever called ``orientation`` tbh, the angle property has been there for 2 years at least.

jlstevens:
So it never worked? 

Sounds like a good reason to add a unit test (if possible) before I merge!

jlstevens:
I think I'll just merge.

PR #1416, opened 2017-05-07
Bokeh bar plot

philippjfr:
In https://github.com/ioam/holoviews/issues/1407 I outlined the way forward for the Bar plot implementations. Since bokeh charts will be split out imminently and I'll be using Bars in at least one project I thought it would be worth replacing the bokeh charts based implementation immediately. This new implementation retains all the functionality of the charts based implementation, while adding more features and improvements:

* Better control over legends
* Better support for updating barplots as part of HoloMaps/DynamicMaps
* **Much** better performance (at least 10x faster for more complex charts) because this implementation uses one vectorized glyph rather than a bunch of individual glyphs like the chart based implementation did.
* Support for colormapping bars
* Better support for overlaying bars
* Support for horizontal bars via the ``invert_axes`` plot option
* Support for log mapping bar heights

[Here](https://anaconda.org/philippjfr/bars/notebook) is the notebook I used to develop the plotting class with an extensive list of examples.

### DIFF
--- a/holoviews/plotting/bokeh/__init__.py
+++ b/holoviews/plotting/bokeh/__init__.py
@@ -110,6 +110,7 @@ options.Points = Options('style', color=Cycle(), size=point_size, cmap='hot')
 options.Histogram = Options('style', line_color='black', fill_color=Cycle())
 options.ErrorBars = Options('style', color='black')
 options.Spread = Options('style', color=Cycle(), alpha=0.6, line_color='black')
+options.Bars = Options('style', color=Cycle(), line_color='black', width=0.8)
 
 options.Spikes = Options('style', color='black')
 options.Area = Options('style', color=Cycle(), line_color='black')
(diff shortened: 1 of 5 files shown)


philippjfr:
I also just added support for batching BarPlots, this means that you can stack and group NdOverlays of Bars, which means you can use the simpler Bars format with only a single key dimension. In other words these two versions are equivalent:

```python
%%opts Bars [stack_index='group]
hv.Bars((range(10)*5, sorted(map(chr, range(65,70))*10), np.random.rand(50)), kdims=['x', 'group'])
```

```python
%%opts NdOverlay [legend_limit=0] Bars [stack_index='group' show_legend=True]
hv.NdOverlay({chr(65+i): hv.Bars(np.random.rand(10)) for i in range(5)}, kdims=['group'])
```

bokeh_plot 44

philippjfr:
Ready to merge once tests are passing.

jlstevens:
The improvements list is quite impressive and we are explicitly moving away from the bokeh charts API so this work was definitely necessary.

The tests are passing and I am happy to merge. My only comment here is regarding unit tests, do you you think existing unit tests are sufficient or can we add more to test this new implementation?

jlstevens:
I think the issue regarding additional testing can be deferred by #1422 (assigned to version 1.8). Merging.

PR #1418, opened 2017-05-08
Drop HLine/VLine zorder to annotation level

philippjfr:
HLine/VLine now no longer go outside the space inside the axes.


jlstevens:
Nice simple fix that gets rid of a pretty visible annoyance!

I'll merge once the tests go green.

jlstevens:
The tests are now passing. Merging.

Issue #1428, opened 2017-05-10
Offer a non-blocking way of triggering events


jlstevens:
This idea has been prototyped, first as a threaded stream and then as a general utility for periodically invoking a callback without blocking.

After some discussion, we can to the conclusion that a method on ``DynamicMap`` might offer the most convenient API as it doesn't require extra imports. Something like this might work quite well:

```python
dmap.periodic_event(0.001, 1000) # Periodically fires the event method without arguments
dmap.periodic_event(0.001, 1000, lambda i: {'counter':i})
```

The first argument is the interval in seconds and the second argument is a count.

The main issue to solve is that we want something that works well when deployed as a bokeh app. One idea is to make the public ``periodic_event`` use a utility for the implementation that can be set to one appropriate for bokeh apps when necessary. For instance, this could be set by an underscore method e.g ``DynamicMap._set_periodic_utility(...)`` by the BokehRenderer.

Once we have prototyped the necessary utility that works with ``bokeh serve``, I think this could end up being quite useful as blocking loops in notebooks don't work with linked streams and are generally a pain.

jlstevens:
This issue has been addressed in PR #1429. The default behavior is blocking (which can be easily disabled) as the non-threaded blocking behavior can have unexpected results if variables involved in the thread are set while the thread is running.

PR #1429, opened 2017-05-10
Non-blocking periodic_events method on DynamicMap

jlstevens:
This PR aims to address #1428 and replaces the Periodic stream prototype we were testing. It should be possible for the bokeh renderer to install a bokeh appropriate utility when running as a bokeh app.

## Action items

- [x] Add general periodic utility to core (simple threaded approach)
- [x] Add periodic_events method to DynamicMap.  
- [x] All count of None to run indefinitely
- [x] Bokeh server support (i.e when running as a bokeh app)
- [x] Update tutorials to mention this and use it
- [x] Unit tests

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -600,6 +600,8 @@ class DynamicMap(HoloMap):
     # Declare that callback is a positional parameter (used in clone)
     __pos_params = ['callback']
 
+    _periodic_util = util.periodic # Utility used by periodic_events method
+
     kdims = param.List(default=[], constant=True, doc="""
         The key dimensions of a DynamicMap map to the arguments of the
         callback. This mapping can be by position or by name.""")
@@ -754,6 +756,25 @@ class DynamicMap(HoloMap):
 
         Stream.trigger(self.streams)
 
+
+    def periodic_events(self, period, count, param_fn=None):
+        """
+        Run a non-blocking loop that updates the stream parameters using
+        the event method. Runs count times with the specified period.
+
+        If param_fn is not specified, the event method is called without
+        arguments. If it is specified, it must be a callable accepting a
+        single argument (the iteration count) that returns a dictionary
+        of the new stream values to be passed to the event method.
+
+        The returned object allows the loop to be stopped at any time
+        using the stop() method.
+        """
+        def inner(i):
+            kwargs = {} if param_fn is None else param_fn(i)
+            self.event(**kwargs)
+
+        return self._periodic_util(inner, period, count)
 
     def _style(self, retval):
         """
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -566,7 +566,7 @@ def get_nested_dmaps(dmap):
     dmaps = [dmap]
     for o in dmap.callback.inputs:
         if isinstance(o, DynamicMap):
-            dmaps.extend(get_nested_streams(o))
+            dmaps.extend(get_nested_dmaps(o))
     return list(set(dmaps))
 
 
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -760,7 +760,8 @@ class DynamicMap(HoloMap):
     def periodic_events(self, period, count, param_fn=None):
         """
         Run a non-blocking loop that updates the stream parameters using
-        the event method. Runs count times with the specified period.
+        the event method. Runs count times with the specified period. If
+        count is None, runs indefinitely.
 
         If param_fn is not specified, the event method is called without
         arguments. If it is specified, it must be a callable accepting a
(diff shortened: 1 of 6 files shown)


jlstevens:
@philippjfr 

I don't have anything more to do on this PR right now. If you could figure out how to get this working when running as a bokeh app, that would be great!

I won't be touching this again (until you tell me to) so don't worry about merge conflicts and just push your changes here.

jlstevens:
Maybe a ``stop()`` method on ``DynamicMap`` would be nice instead of having to grab a handle on something that has ``stop()``. ``DynamicMap`` would have to hold onto these thread like objects internally so it could stop all of them that might still be running.

jlstevens:
In the last commit, I implemented a slightly different approach - a periodic object with a stop method. This means you can do something lik this:

```python
dmap.periodic(0.001, 1000000)
dmap.periodic.stop() # To stop whenever you want 
```

In addition to making stopping periodic callbacks more convenient, this approach allows  ``_periodic_util`` attribute to be set on the ``periodic`` class. This can be set to something different from the default e.g in a bokeh app context. In addition the instances of this utility are available via the ``instances`` attribute which, by default, is used to stop anything that is running but can also be accessed by the renderer if necessary.

@philippjfr I think this approach should fit better with the bokeh app support. Over to you!

jlstevens:
I've just had another thought: maybe it would be conceptually simpler to only have one instance.

If we did that, then when you run ``__call__``, it stops that instance (if it is already running). As the event method can update all the stream arguments at once, I'm not sure why we would want to have multiple periodic runner instances.

holoviews/core/spaces.py
```diff
@@ -559,18 +559,23 @@ def __call__(self):
...
     callback.
     """
-    layer_streams = list(dmap.streams)
-    if not isinstance(dmap.callback, Callable):
-        return list(set(layer_streams))
-    for o in dmap.callback.inputs:
-        if isinstance(o, DynamicMap):
-            layer_streams += get_nested_streams(o)
-    return list(set(layer_streams))
+    return list({s for dmap in get_nested_dmaps(dmap) for s in dmap.streams})
```
jlstevens: Makes sense to refactor it this way.

holoviews/plotting/bokeh/renderer.py
```diff
@@ -122,9 +123,11 @@ def server_doc(self, plot, doc=None):
         if doc is None:
             doc = curdoc()
         if isinstance(plot, BokehServerWidgets):
-            plot.plot.document = doc
-        else:
-            plot.document = doc
+            plot = plot.plot
+        plot.document = doc
```
jlstevens: Maybe add a comment as a reminder of what ``attach_periodic`` does...

holoviews/plotting/bokeh/util.py
```diff
@@ -664,3 +665,58 @@ def categorize_array(array, dim):
...
+        self.count = 0
+        self.document.remove_periodic_callback(self.callback)
+
+    @property
+    def completed(self):
+        class completed(object):
+            @classmethod
+            def is_set(cls):
+                return self.count == 0
+        return completed
```
jlstevens: I'll update the API of the other periodic utility so it doesn't need to be this complex. I'll also simplify this.

jlstevens:
@philippjfr  All looks good!

I'll update the stuff around ``completed`` and ``is_set`` to simplify it, but otherwise we just need to add unit tests and update the docs. Maybe one of the existing bokeh app examples should use it...if not maybe add another one?

Any way you can set up bokeh specific unit tests for that utility? I'll add some unit tests for the core utility...

holoviews/core/spaces.py
```diff
@@ -589,6 +596,53 @@ def dynamicmap_memoization(callable_obj, streams):
...
+        instance.start()
+        self.instance= instance
+
+    def stop(self):
+        "Stop the periodic process."
+        self.instance.stop()
+
+    def __str__(self):
+        return "<holoviews.core.spaces.periodic method>"
+
```
jbednar: Maybe add a comment explaining what this representation is meant to achieve?

jlstevens: If you have another idea of what the repr should be, I would be happy to implement it! What sort of comment are you thinking of?

jlstevens:
@philippjfr Ready for review once the tests go green.

holoviews/core/spaces.py
jbednar: Just something to indicate why the default representation was not suitable.

jlstevens: It isn't strictly necessary, but I think it is a bit nicer than the default which is something like:

```
<holoviews.core.spaces.periodic at 0x10886e850>
```

Instead of a random memory address, I say that it is a method which reflects how this utility is used (i.e it lives on ``DynamicMap``).

Happy to remove it if you think the default is better.

jbednar: No need to remove it, just put a docstring saying what you say here (that this repr provides additional information that should be more useful).

jlstevens:
@philippjfr The pr tests have passed and I don't think the push build ones will unless I rebase.

philippjfr:
Looks good, merging.

PR #1431, opened 2017-05-11
Pinned pandas version on travis to 0.19.2

philippjfr:
Tests are failing because pandas 0.20.1 is incompatible with dask 0.13. Might also consider pinning pandas to 0.19.2 instead.


philippjfr:
Going to merge this myself so we can have working tests again.

jlstevens:
> Going to merge this myself so we can have working tests again.

Yes, that's fine...there was no need to wait for me!

Issue #1426, opened 2017-05-09
Miscellaneous linked streams improvements


jlstevens:
Right now the ``Tap`` stream only processes single taps while ``DoubleTap`` process double taps. For consistent, it should be ``SingleTap`` and ``DoubleTap`` and then ``Tap`` could process *all* taps, which I think might generally be more useful than ``SingleTap``.

jlstevens:
In addition I think a ``DragMotion`` stream that uses the pan event (but does not actually pan) would be very useful. It shouldn't be called ``Drag`` as nothing (i.e a glyph or other visual element) is necessarily being dragged.

jlstevens:
While on the topic of improving the defined linked streams, #1384 suggested that PositionXY linked streams could be limited to the axis limits (with a prototype).

As this issue now contains a number of different suggestions, I'm updating the title.

jlstevens:
This has been addressed in PR #1436 referenced above.

PR #1436, opened 2017-05-14
Miscellaneous stream improvements

jlstevens:
This PR aims to implement the suggestions in issue #1426. 

The first commit distinguishes ``SingleTap`` from ``Tap`` and ``DoubleTap`` where ``Tap`` is supposed to capture all tap events.

@bryevdv Unfortunately I can't quite implement the intended semantics for ``Tap`` as a double tap event should resolve into two taps. As far as I can tell, the way hammer is used in bokeh does not allow for this. 

Is this correct? Would it be reasonable to add a bokeh event that reflects all taps, whether single or part of a double tap?

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -3,7 +3,7 @@ from bokeh.models import CustomJS
 
 from ...core import OrderedDict
 from ...streams import (Stream, PointerXY, RangeXY, Selection1D, RangeX,
-                        RangeY, PointerX, PointerY, Bounds, Tap,
+                        RangeY, PointerX, PointerY, Bounds, Tap, SingleTap,
                         DoubleTap, MouseEnter, MouseLeave, PlotSize)
 from ...streams import PositionX, PositionY, PositionXY # Deprecated: remove in 2.0
 from ..comms import JupyterCommJS
@@ -541,6 +541,17 @@ class PointerYCallback(PointerXYCallback):
 class TapCallback(PointerXYCallback):
     """
     Returns the mouse x/y-position on tap event.
+
+    Note: As of bokeh 0.12.5, there is no way to distinguish the
+    individual tap events within a doubletap event.
+    """
+
+    on_events = ['tap', 'doubletap']
+
+
+class SingleTapCallback(PointerXYCallback):
+    """
+    Returns the mouse x/y-position on tap event.
     """
 
     on_events = ['tap']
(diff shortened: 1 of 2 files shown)


jlstevens:
Here is an example (including all the code!) of a little visualization using the new ``Draw`` stream:

draw

jlstevens:
Note that even after having updated the skip conditions, odd things happen when using ``%%opts Path [toolbar=None]`` in the example shown above. Needs investigating.

jbednar:
Looks super useful, thanks!  Is the three-pronged "if" statement needed because a Curve cannot be empty?  Avoiding tricky logic like that seems like a good reason to allow empty Curves...

philippjfr:
>Is the three-pronged "if" statement needed because a Curve cannot be empty?

I think in this case it's to start a new curve every time a new drag even is started. That said in most cases we do support empty Elements.

jlstevens:
> I think in this case it's to start a new curve every time a new drag even is started.

That's right! If you can suggest a way to shorten the example, I would be very happy to try it out.

jlstevens:
Anyway, in my last commit I added the JS to clip the x,y values to the axis ranges for ``PointerXY``. The JS is pretty horrible so I want to clean it up before this PR is merged. Otherwise, I have implemented all the proposal I made in the issue.

philippjfr:
>The JS is pretty horrible so I want to clean it up before this PR is merged.

It's not pretty but I'm not sure how it can be improved.

jlstevens:
> It's not pretty but I'm not sure how it can be improved.

If nothing else, I can improve the formatting. After that, I think this PR is ready to merge though it might be worth investigating the odd behavior when ``toolbar=None``...

jlstevens:
To make it easier to test, here is the text of the gif example above:

```

from holoviews.streams import Draw
vals, curves = [], {}

def callback(x,y, stroke_count):
    global vals, curves
    curve = hv.Curve(vals)
    if len(curves)==0:
        curves[stroke_count] = hv.Curve([(x,y)])
    elif stroke_count != (len(curves)-1):
        curves[stroke_count]= curve
        vals = []
    else:
        vals.append((x,y))
        curves[stroke_count]= curve

    return hv.Path([curve.data for curve in curves.values()])

hv.DynamicMap(callback, streams=[Draw(x=0,y=0)])
```

jlstevens:
@philippjfr Other than the odd behavior with ``toolbar=None`` (which I don't know what to make of) I think this PR is ready for review.

philippjfr:
Should I go ahead and merge? Toolbar option is fixed in https://github.com/ioam/holoviews/pull/1442.

jlstevens:
> Should I go ahead and merge?

Given that the toolbar issue is fixed, yes please!

PR #1438, opened 2017-05-15
Filter style opts for plots not declaring style_opts

philippjfr:
The logic here was slightly wrong in that plots can declare that they have no valid ``style_opts`` and filtering should still be applied.


jlstevens:
Looks good! Merge?

philippjfr:
Yes, ready to merge.

PR #1439, opened 2017-05-15
Attach linked streams in batched mode

philippjfr:
In batched mode linked streams weren't being attached, previously this would have been handled by the OverlayPlot wrapping the batched plot. I think this fix won't cause any issues but I'll run through examples now to double check.


jlstevens:
Happy to merge once you've tested it a bit as the change itself is very simple.

philippjfr:
Added a unit test and ran through my examples with encountering issues.

jlstevens:
Tests have now passed. Merging.

PR #1440, opened 2017-05-15
Fix for initializing batched plot with empty element

philippjfr:
Since we now support empty elements the previous approach in this code to get a non-empty element breaks if you're trying to initialize a batched plot with an empty element.


philippjfr:
@jlstevens Ready to merge.

jlstevens:
Looks good. Merging.

PR #1442, opened 2017-05-15
Disable bokeh toolbar when set to None

philippjfr:
Seems to have been an indentation issue, but I've also ensured the toolbar option is propagated to the OverlayPlot. Layouts and Grids probably needs their own way of disabling and placing toolbars, but that's a separate issue.


jlstevens:
Ready to merge? 

I know it isn't likely to be tests properly but I'll still wait till at least some of the tests turn green.

philippjfr:
Yes, ready to merge.

PR #1444, opened 2017-05-16
Improved BokehRenderer.app API

philippjfr:
The BokehRenderer.app method now has three ways of working:

1) By default it just returns a document with plot attached to it, this can be used in a script to deploy it with ``bokeh serve``.

2) When using the ``show`` option it will create an Application instance, if the notebook extension has been loaded it will start a server and display it inline.

3) When using the ``show`` option when the notebook extension has not been loaded or if the ``new_window`` option is supplied it will create the app and open it in a new window. This may be used from an IPython terminal or within the notebook.

Suggestions for improving the naming of the ``show`` and ``new_window`` arguments or other suggestions for improving the API welcome.

### DIFF
--- a/holoviews/plotting/bokeh/renderer.py
+++ b/holoviews/plotting/bokeh/renderer.py
@@ -84,10 +84,24 @@ class BokehRenderer(Renderer):
         elif fmt == 'json':
             return self.diff(plot), info
 
+
     @bothmethod
-    def get_widget(self_or_cls, plot, widget_type, **kwargs):
+    def get_plot(self_or_cls, obj, doc=None, renderer=None):
+        """
+        Given a HoloViews Viewable return a corresponding plot instance.
+        Allows supplying a document attach the plot to, useful when
+        combining the bokeh model with another plot.
+        """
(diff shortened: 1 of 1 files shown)


jlstevens:
Both``show`` and ``new_window`` seem fine to me. I think the original suggestion was for a ``plot`` utility but a ``get_plot`` bothmethod on the renderer also seems fine. The only naming suggestion I have is to maybe call it ``plot_instance`` to make it clear it isn't the plotting class...

philippjfr:
>I think the original suggestion was for a plot utility but a get_plot bothmethod on the renderer also seems fine. 

That's a separate suggestion I haven't implemented yet, ``renderer.get_plot`` is a general method on the baseclass which does return the matplotlib plotting class.

jlstevens:
In that case the names ``plot_class`` and ``plot_instance`` would make that clearer.

jlstevens:
Ok,  I got confused as ``get_plot`` is not a new method  (I got mislead by the diff). The new API and names seem reasonable so I'll merge now.

PR #1446, opened 2017-05-16
Element extents fixes

philippjfr:
Certain Elements have special semantics for dimension ranges, e.g. on ErrorBars/Spread the y-range is determined by the combination of y-values and y-errors. This is handled correctly but it currently completely ignores explicitly set ranges, which should always override auto-computed ranges. Additionally ``Histogram`` uses extents to compute the bin ranges when it should simply use ``.range`` like all other Elements.

Fixes:

https://github.com/ioam/holoviews/issues/1432
https://github.com/ioam/holoviews/issues/1433
https://github.com/ioam/holoviews/issues/1445

### DIFF
--- a/holoviews/element/chart.py
+++ b/holoviews/element/chart.py
@@ -245,18 +245,16 @@ class Histogram(Element2D):
 
 
     def range(self, dimension, data_range=True):
-        drange = super(Histogram, self).range(dimension, data_range)
-        dimidx = self.get_dimension_index(dimension)
-        dim    = self.get_dimension(dimension)
-        if dimidx == 0 and data_range:
+        if self.get_dimension_index(dimension) == 0 and data_range:
+            dim = self.get_dimension(dimension)
             lower, upper = np.min(self.edges), np.max(self.edges)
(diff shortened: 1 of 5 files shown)


jlstevens:
Sounds good and respecting the Dimension range argument is far more important than respecting ``extents`` (which will go away eventually).

Happy to merge once the tests pass though I would prefer to see some unit tests first.

philippjfr:
>Sounds good and respecting the Dimension range argument is far more important than respecting extents (which will go away eventually).

I'm not actually changing anything about respecting ``extents`` here, I'm just making sure that ``Histogram`` no longer uses extents internally to signal the bin range, and that the plots take into account explicitly set ``Dimension.range`` values. ``extents`` still override everything else if set.

jlstevens:
Thanks for clarifying. Keeping the behavior the same (wrt extents) while reducing dependence on it internally will be helpful in future.

philippjfr:
Ready once tests pass.

jlstevens:
Tests have passed. Merging.

Issue #1427, opened 2017-05-09
Document, rename or remove the Aliases utility


jlstevens:
I think this utility is still useful, even if it might now be misnamed. Example usage:

```
>>> import holoviews as hv
>>> al = hv.util.Aliases(name1='Long label', name2='Long label 2')
>>> al.name1
('name1', 'Long label')
```

It can be useful for defining all your dimension tuples at the top of the notebook instead of repeating such tuples. I believe it can also accept fully-fledged dimension objects.

jlstevens:
@jbednar I really would like a solution to this and I do feel such a utility would be really helpful, especially for more elaborate examples and bigger projects.

Implementing something here is easy, the problem is deciding what to implement. I'll leave this issue on the 1.8 milestone in case you have any new suggestions otherwise I will remove that milestone.

philippjfr:
I'm still -1 on having the alias object, but won't object if you insist, but I would like to see this and the corresponding PR closed on this release as this has dragged out a long time now.

philippjfr:
The issue does correctly suggest that if we're not going to improve aliases we should delete the current implementation, which is what I'd vote for.

PR #1448, opened 2017-05-16
Alias utility update

jlstevens:
This PR aims to address #1427. I realize we don't yet have a solution everybody is happy with but I think this is worth updating while the utility still exists.

Here is an example of it in use:

image

For this example it doesn't do much but if you have a long notebook that reuses dimensions/groups/labels constantly, I think this utility really pays off, making everything more succinct and organized. It means you can specify the definitions once at the top, making the rest of the notebook more succinct.

Once we do find a good solution to defining dimensions and sanitized groups/labels without awkward repetition we should document it and probably remove the dimension preset system.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -347,27 +347,6 @@ def capitalize_unicode_name(s):
     return s[:index] + tail
 
 
-class Aliases(object):
-    """
-    Helper class useful for defining a set of alias tuples on a single object.
-
-    For instance, when defining a group or label with an alias, instead
-    of setting tuples in the constructor, you could use
-    ``aliases.water`` if you first define:
-
-    >>> aliases = Aliases(water='H_2O', glucose='C_6H_{12}O_6')
-    >>> aliases.water
-    ('water', 'H_2O')
-
-    This may be used to conveniently define aliases for groups, labels
-    or dimension names.
-    """
-    def __init__(self, **kwargs):
-        for k,v in kwargs.items():
-            setattr(self, k, (k,v))
-
-
-
 class sanitize_identifier_fn(param.ParameterizedFunction):
     """
     Sanitizes group/label values for use in AttrTree attribute
--- a/holoviews/util.py
+++ b/holoviews/util.py
@@ -2,7 +2,7 @@ import inspect
 
 import param
 
-from .core import DynamicMap, HoloMap, ViewableElement
+from .core import Dimension, DynamicMap, HoloMap, ViewableElement
 from .core.operation import Operation
 from .core.operation import OperationCallable
 from .core.spaces import Callable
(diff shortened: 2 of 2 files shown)


jlstevens:
In the end, I decided to rename ``Aliases`` to ``Dims`` and focus on dimension aliases only:

image

Unless I add something for tuple aliases, this means we offer nothing to help with sanitized group/label aliases.

jbednar:
Would something like this work as a more general approach, not specific to Dimensions or groups or labels, but just to things that you might want to look up easily?

```
class Aliases(object):
"""
Helper class useful for defining aliases, which are 
Python identifiers that expand either to a named
object or to an arbitrary Python string.

Sample usage:

``
d = Aliases(hv.Dimension('female_weight', unit='kg', range=(0,None)),
            male_weight='Body weight of male participants',
            mh=hv.Dimension('male_height', unit='kg', range=(0,None)))
Curve(data, kdims=[d.female_weight], vdims=[d.mh])
``

Here `female_weight` is passed as a positional argument, and is thus
addressed by its `name` attribute, while the others are addressed
by the specified keyword.
"""
def __init__(self, *pargs, **kwargs):
    for o in pargs:
        setattr(self, o.name, o)
    for k,v in kwargs.items():
        setattr(self, k, v)
```

jlstevens:
@jbednar Interesting suggestion!

Pros:

1. Conceptually simpler
2. Would work with any parameterized object (even if the names aren't always nice)

Cons:

Doesn't help map ``keyword='val'`` to the tuple ``('keyword', 'val')``

For instance, in your example:

```python
Dimension(d.male_weight)
```

Is not supplying the name (``'male_weight'``) to the name parameter but supplies the label instead (``''Body weight of male participants''``). To define a name and label, you would have to do:


```python
Dimension('male_weight', label=d.male_weight) # OR
Dimension(('male_weight',d.male_weight))
```

Which isn't so helpful : now the alias is only used as a handle on the label string and you still have to repeat typing 'male_weight' twice.

Anyway, the idea of using ``*args`` instead of a list had occurred to me (and I think it is an improvement) and I hadn't realized this can be used for anything with a ``.name`` (making that part more general that I had initially thought).

To adapt your code, the more version (imho) is this:

```python
def __init__(self, *pargs, **kwargs):
    for o in pargs:
        setattr(self, o.name, o)
    for k,v in kwargs.items():
        setattr(self, k, (k,v))
```

Unfortunately, it is a little harder to explain in the docstring!

Here is another way you might try to generalize it though it isn't very sensible unless we can refine the idea some more:

```python
# Example 'mappers' where this could be a class parameter
mapper = lambda k,v: v                        # Same output as Jim's proposal above
mapper = lambda k,v: hv.Dimension(k, label=v) # Now v has to be a string to make a dim
mapper =  lambda k,v: (k,v)                   # Tuple usable for dims or sanitization

def __init__(self, *pargs, **kwargs):
    for o in pargs:
        setattr(self, o.name, o)
    for k,v in kwargs.items():
        setattr(self, k, self.mapper(k,v) )
```

I'm thinking it would behave like what you suggested but there is a way to specify a callable to output the appropriate value given the supplied keyword and value. One issue with this is that if you don't want to restrict the possible kwargs, these 'mapper's would have to be supplied at the class level (which might not be so bad really).

Hope that is clear! I still feel something like this is needed, but it does still need more thought...

jbednar:
> For instance, in your example:
> Dimension(d.male_weight)

I don't see that code in my example, so I'm not sure what you mean. In any case, I'm not sure exactly why you want to map to a tuple instead of just the longer string, so please do explain that.

philippjfr:
I'm closing this, it's been stale for over a year and both I and @jbednar agreed that we would like to see it closed.

PR #1452, opened 2017-05-17
Make the naming consistent

yuvallanger:
The other option is to rename everything in this tutorial into "waves" instead of "sine"

### DIFF
--- a/doc/Tutorials/Dynamic_Map.ipynb
+++ b/doc/Tutorials/Dynamic_Map.ipynb
@@ -72,7 +72,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


jlstevens:
Thanks for doing this!

> The other option is to rename everything in this tutorial into "waves" instead of "sine"

What do you prefer? To be honest, I think ``waves`` might be a little nicer as the fact it is generated from a sine function is irrelevant. What matters is there is a pattern (of some sort) you want to look at.

yuvallanger:
I myself would also go with `waves`, but there is already `sine_array` in
another tutorial file. Maybe also change it in the other tutorial?

jlstevens:
Good point.

@philippjfr @jbednar What do you recommend?

jbednar:
I prefer just `waves`.  Not sure if the tutorials need to be consistent with each other, but `waves` is probably better for both.

philippjfr:
Sounds good to me.

jlstevens:
@yuvallanger Are you planning to apply the same renaming to the other tutorial you mentioned?

If so great! Otherwise, I'm happy to merge as it is.

yuvallanger:
Let's just merge it as it is. The other tutorial mentions sines and
cosines, so it isn't simply wavey things.

jlstevens:
Ok great! And @yuvallanger thanks again for your PR!

Issue #1393, opened 2017-05-02
Do not raise exceptions when importing notebook extension


jlstevens:
Currently we have:

```python
try:
    import IPython  # noqa (API import)
    from .ipython import notebook_extension
except ImportError as e:
    class notebook_extension(param.ParameterizedFunction):
        def __call__(self, *args, **opts):
            raise Exception("IPython notebook not available")
    if str(e) != 'No module named IPython':
        raise e
```

The last two lines caused problems with Imagen as the exception message wasn't as expected (it was an ipykernel import error not an IPython import error). We should probably just ignore all exceptions when importing the notebook extension regardless of the exception message.

jlstevens:
Addressed in #1454. Closing.

PR #1454, opened 2017-05-18
Ignoring all ImportErrors when setting notebook_extension stub

jlstevens:
Trivial PR to address #1393. @philippjfr I think it can be merged right away.


philippjfr:
Looks good, merging.

Issue #1190, opened 2017-03-10
Utility for constructing Dimensions from Parameters?


jbednar:
An `hv.Dimension` is quite like a `param.Parameter`, in that both establish a named dimension with bounds, and they are often used for the same purpose, e.g. to determine the type of widget and the numerical range covered by that widget (such as a slider).   In the code below, the declared `kdims` list is a copy of data already stored and declared on the underlying `param.Parameterized` object `ig.DifferenceOfGaussians`, and it would reduce redundancy and increase generality if those values could simply be looked up on the Parameterized object.  

So, maybe we could have a utility that will accept a Parameterized object and then return a list of Dimensions for use as kdims, possibly with specified filtering (whitelist and blacklist) on the parameter names?  Coupled with #1189, this approach could get rid of all the duplication between the DynamicMap and the DifferenceOfGaussians object, making it possible to use the same code with any such object to give widgets for whatever Parameters it found there.

```python
import numpy as np
import holoviews as hv
import imagen as ig
import imagen.transferfn as tf
from imagen.image import FileImage
from imagen.transferfn.sheet_tf import Convolve
hv.notebook_extension()
%opts Image (cmap="gray") {+axiswise}

FileImage.filename = "/Users/jbednar/imagen/doc/images/ellen_arthur.pgm"

kdims = [hv.Dimension('size',                   range=(0.1, 2)),
         hv.Dimension('positive_size',          range=(0.1, 2)),
         hv.Dimension('negative_size',          range=(0.3, 4)),
         hv.Dimension('positive_aspect_ratio',  range=(1.0,10)),
         hv.Dimension('negative_aspect_ratio',  range=(1.0,10))]

def filtered_images(size,positive_size,negative_size,positive_aspect_ratio,negative_aspect_ratio):
    convolver = Convolve(kernel_pattern=ig.DifferenceOfGaussians(**locals()))
    return FileImage(output_fns=[convolver])[:]

FileImage()[:] + hv.DynamicMap(filtered_images, kdims=kdims)
```

jlstevens:
Seems like a very plausible utility, either to just declare kdims from parameters or to declare the entire dynamic map...

jbednar:
Filtering by precedence would also be useful, so that by default hidden parameters aren't included, but the precedence value can be increased or decreased to get more or fewer widgets.  That way there can be filtering that works across a set of Parameterized objects without knowing anything about their internals.

jlstevens:
[...]

PR #1456, opened 2017-05-18
Simple utility to turn parameters into dimensions

jlstevens:
This PR aims to address #1190. 

Other than finding a better name for the utility, everything is straightforward (see the tests for an example of how to use it). Ready for review/merge once the tests are green.

### DIFF
--- a/holoviews/util.py
+++ b/holoviews/util.py
@@ -141,6 +141,9 @@ class ParamDimList(param.ParameterizedFunction):
     label = param.Boolean(default=True, doc="""
         Whether to use the parameter documentation as the dimension label""")
 
+    precedence_threshold = param.Number(default=0, doc="""
+        The threshold below which parameters are ignored.""")
+
     def __call__(self, parameterized, **kwargs):
 
         types = {param.String:str, param.Integer:int}
@@ -149,7 +152,9 @@ class ParamDimList(param.ParameterizedFunction):
         params = parameterized.params()
         for name, param_obj in params.items():
             if name in p.exclude:  continue
-            elif param_obj.precedence  and param_obj.precedence < 0: continue
+            elif param_obj.precedence:
+                if param_obj.precedence < p.precedence_threshold:
+                    continue
             dim_kws = {}
             if param_obj.doc and p.label:
                 dim_kws['label'] = param_obj.doc
--- a/holoviews/util.py
+++ b/holoviews/util.py
@@ -2,7 +2,7 @@ import inspect
 
 import param
 
-from .core import DynamicMap, HoloMap, ViewableElement
+from .core import Dimension, DynamicMap, HoloMap, ViewableElement
 from .core.operation import Operation
 from .core.util import Aliases  # noqa (API import)
 from .core.operation import OperationCallable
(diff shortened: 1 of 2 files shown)


holoviews/util.py
```diff
@@ -127,3 +127,42 @@ def _make_dynamic(self, hmap, dynamic_fn, streams):
...
+    def __call__(self, parameterized, **kwargs):
+
+        types = {param.String:str, param.Integer:int}
+        p = param.ParamOverrides(self, kwargs)
+        dim_list = []
+        params = parameterized.params()
+        for name, param_obj in params.items():
+            if name in p.exclude:  continue
+            elif param_obj.precedence  and param_obj.precedence < 0: continue
+            dim_kws = {}
```
jbednar: Can the precedence threshold be a parameter (defaulting to 0)?

jbednar:
Looks good to me, thanks!!

holoviews/util.py
jlstevens: Sure. Addressed in f125e19d.

philippjfr:
I question a little how useful this is in practice, it seems to cover a fairly limited number of parameters. At minimum ObjectSelector should also be handled, although I'm not sure yet how well arbitrary objects will work as Dimension.values, which is something we should probably handle in some way.

holoviews/util.py
```diff
@@ -127,3 +127,47 @@ def _make_dynamic(self, hmap, dynamic_fn, streams):
...
+            if param_obj.doc and p.label:
+                dim_kws['label'] = param_obj.doc
+            if isinstance(param_obj, param.Number):
+                if param_obj.bounds != None:
+                    dim_kws['range'] = param_obj.bounds
+                if param_obj.softbounds != None:
+                    dim_kws['soft_range'] = param_obj.softbounds
+            elif isinstance(param_obj, param.List):
+                dim_kws['values'] = param_obj.default
```
philippjfr: This doesn't make sense to me. A ``param.List`` expects the whole list as input but this treats it basically as an ObjectSelector.

jlstevens:
I don't have any strong feelings about this utility - I see this PR as a skeleton for what (I hope!) will remain a fairly simple utility. Feel free to think about how you might want to use this in practice and push whatever improvements you think make sense.

philippjfr:
In its current form it doesn't seem to handle some of the basic cases, in particular ObjectSelector and in some cases e.g. ``param.List`` arguably doesn't make sense. The main question I have is how we expect this to be used. Dimensions will never be able to cover all the cases that parameters can handle.

So in my opinion this should be skipping cases that won't work as Dimensions such as ListSelector/List/Tuple/NumericTuple etc. (they will all break when used as a HoloMap/DynamicMap because they are not valid keys). Because of this this utility will only ever cover a narrow range of cases. If you and @jbednar still think that's useful I'd be happy to merge it, but not without doing that validation.

jlstevens:
> I question a little how useful this is in practice...

I also wonder about that but I don't object to having a *small* utility that is useful for simple cases. I don't see how to promote all the various parameter types to dimension but I do think many of the most common parameter case can be covered.

jlstevens:
I posted my comment above at the same time as you posted yours. I think we essentially agree and it is up to @jbednar to decide...

philippjfr:
I'm going to close this PR, we can revisit this later if we think there's a general approach that's not fundamentally limited.

jlstevens:
Fine by me!

Issue #1383, opened 2017-04-28
Dimension ordering is not preserved when defined


jxrossel:
Hi,

According to the docs (http://holoviews.org/Tutorials/Elements.html >> Bars), 

> To preserve the initial ordering specify the Dimension with values set to 'initial', or you can supply an explicit list of valid dimension keys

In `dimension.Dimension.values`, the doc says:

> Optional specification of the allowed value set for the dimension that may also be used to retain a categorical  ordering

But the code reads:
```
        values = params.get('values', [])
        if isinstance(values, basestring) and values == 'initial':
            self.warning("The 'initial' string for dimension values is no longer supported.")
            values = []
        all_params['values'] = sorted(list(unique_array(values)))
```

So that (1) 'initial' cannot be used, (2) ordering is not preserved (thanks to the call to `sorted`)

philippjfr:
Thanks for catching this, we need to update the tutorial, ``initial`` is no longer supported in general. Also @jlstevens do you know why we're sorting the ``values`` here?

jlstevens:
I think this was something @basnijholt requested for ordered sliders.

jlstevens:
I think making the values unique is sensible, not sure about the sorting now we removed ``'initial'``...

jlstevens:
Whatever the fix is, I'm sure it is simple. Philipp and I just need to make sure the semantics are clear and that we aren't breaking backwards compatibility.

jxrossel:
Well, I guess that if a user specifies the dimension values, they should be left as given (or possibly made unique, but that could be the user's responsability as well)

jlstevens:
I think automatically making them unique is almost always appropriate and I don't *think* anything is relying on the sorting (which I agree we should remove). Any thoughts @philippjfr ?

philippjfr:
>I think automatically making them unique is almost always appropriate and I don't think anything is relying on the sorting (which I agree we should remove). Any thoughts @philippjfr ?

I think the original issue was related to partial overlaps, i.e. where one DynamicMap defines some list of values and another defines a different partially overlapping list of values. In that case sorting is usually the most sensible thing to do, but I don't think it's the best policy in general.

jlstevens:
Ok, sounds like we should just disable the sorting. I'm happy to do that.

jlstevens:
Addressed in the PR referenced above (#1457) - the values are now unique but not sorted automatically. Closing.

PR #1457, opened 2017-05-18
Removed automatic sorting of dimension values

jlstevens:
Very simple PR to address issue #1383: ``Dimension`` no longer sorts the supplied values. This change *should* be ok - the notebook tests will help us decide!


jlstevens:
The tests are passing so this change seems fine after all...

jlstevens:
@philippjfr I talked to @basnijholt and I think we can go ahead with this change. If you agree, please merge.

Issue #1419, opened 2017-05-08
Simplify core operation classes


jlstevens:
In PR #1391 we finally started documented operations. While doing through this tutorial in #1384, we realized that we hardly *ever* use anything other than ``ElementOperation``.

``MapOperation`` has never been used to our knowledge and its definition is so light a user might as well just write their own ``param.ParameterizedFunction``. This just leaves ``TreeOperation`` which has only been used by the holoviews devs in a few places (in particular, one operation in featuremapper is a tree operation).

My recommendation is as follows:

* Remove ``MapOperation`` entirely
* Fold the current ``Operation`` baseclass into ``ElementOperation``
* Rename ``ElementOperation`` to ``Operation`` (a better name for users).
* Leave a ``class ElementOperation(Operation):`` definition just to issue a deprecation warning.
* Remove ``TreeOperation``, factoring any general and useful functionality into a utility.

This plan maintains backwards compatibility as we never used or documented ``MapOperation`` and ``TreeOperation`` is very unlikely to be used by users as it is needed for rare, more complex tasks (and was also undocumented).

Having a single ``Operation`` base class would simplify things a lot!

jlstevens:
Everything suggested in this issue has been addressed by the PR referenced above (#1421) except for removing ``TreeOperation``.

@philippjfr Happy for me to just remove it from HoloViews and add it to featuremapper?

philippjfr:
Sure, as long as you update the imports in topographica as well.

jlstevens:
The final bit (removing ``TreeOperation``) was merged in the PR above. Closing.

PR #1458, opened 2017-05-18
Removed TreeOperation

jlstevens:
Finishes off the last suggestion made in in issue #1419. 

Featuremapper still has TreeOperations so ioam/featuremapper#14 should be merged first. I've already pushed a commit to Topographica to make sure it can use the featuremapper version when the HoloViews version disappears.


jlstevens:
I think both this PR and the one on featuremapper are now ready to merge.

PR #1467, opened 2017-05-21
Avoid xarray len implementation loading data

philippjfr:
XArray supports lazy loading of data but the current xarray interface uses the actual array data to determine the length of the data, which is highly inefficient. This PR instead uses the product of the coordinate array lengths which are always in memory to determine the overall length.


jlstevens:
Makes sense, happy to merge when the tests go green.

jlstevens:
Tests are green. Merging.

PR #1469, opened 2017-05-21
Fix for BokehRenderer.server_doc for plots with widgets

philippjfr:
This fixes a bug in the ``BokehRenderer.server_doc`` method, which resulted in the bokeh widgets not being added to the Document and therefore not appearing.

### DIFF
--- a/holoviews/plotting/bokeh/renderer.py
+++ b/holoviews/plotting/bokeh/renderer.py
@@ -161,13 +161,13 @@ class BokehRenderer(Renderer):
         """
         if doc is None:
             doc = curdoc()
+        root = plot.state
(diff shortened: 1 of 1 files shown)


jlstevens:
Looks good. Any thoughts on how we will test the bokeh app functionality?

philippjfr:
>Looks good. Any thoughts on how we will test the bokeh app functionality?

I'll have a look at how bokeh tests server app functionality but I think we should be able to do some actual testing by starting up a server and perhaps connecting to it with a separate client.

philippjfr:
I think we should do this separately though. Can we merge this for now?

jlstevens:
> Can we merge this for now?

Sure!

PR #1476, opened 2017-05-22
Add example notebooks for Elements

philippjfr:
As one of the main tasks for the new documentation we want to split out the example notebooks. I started converting individual examples and in doing so found that many examples in the existing Element tutorials are either outdated or downright wrong so I started updating them. I got through a fair few of them and made bokeh specific additions. It probably would have been easier to start porting the existing materials and then adapting them once they are added and only then add backend specific additions for the different backend specific versions.

Given that I've already started updating them I'd suggest the following plan, that @jlstevens can help me with. I'd suggest you work through the Elements I've already added and add matplotlib versions for each, making improvements to both versions if you think they are necessary and replacing/removing bits that rely on bokeh features in the matplotlib version (e.g. hover tool -> show_values in the HeatMap example). In the meantime I'll work on bokeh and matplotlib versions of the outstanding elements, which are: 

* [x] Contours
* [x] QuadMesh
* [x] Paths
* [x] Histogram
* [x] HLine/VLine
* [x] Arrow
* [x] Text
* [x] Spline
* [x] Box
* [x] Ellipse
* [x] Bounds
* [ ] Surface (mpl only)
* [ ] Scatter3D (mpl only)
* [ ] Trisurface (mpl only)


philippjfr:
One thing I want to add later is to automatically parse all the other examples and then insert a final cell to each example notebooks which links to all the examples in which the element is used.

jlstevens:
Sounds like a plan!

jlstevens:
Note that the ``Scatter`` histogram wasn't working (should be fixed in #1501) and the ``yticks`` option isn't currently working for ``Spikes.ipynb`` (both in the bokeh backend).

jlstevens:
@philippjfr I've now gone through and updated all the notebooks in the original commit as well as ``Contours`` and ``QuadMesh``. I also believe you have some additions you want to make to ``VectorField``.

From now one, I'm happy to edit/review the element notebooks as you push them. Alternatively, I can start authoring some myself in which case we need to decide how to split the remaining elements between us.

jlstevens:
How about you do  ``Paths`` and ``Histogram``, I'll add all the annotations and we leave the 3D ones till after we've added all the matplotlib versions?

philippjfr:
>How about you do Paths and Histogram, I'll add all the annotations and we leave the 3D ones till after we've added all the matplotlib versions?

Sounds like a plan, I've got Path and Histogram drafts locally already.

jlstevens:
I've copied all the bokeh notebooks and I am working through them to update them.  So far I've updated: ``Area, Bounds, Box, BoxWhisker, Contours`` and ``Curve``.

## Notes:

* Contours colormaps should work but hasn't been tested yet (PR not merged).
* Bars does not work the same way in matplotlib as bokeh:``ValueError: Parameter 'group_index' must be an integer.`` Skipping it for now.
* Curves needs a width option in bokeh (funny aspects) which matplotlib doesn't need.

jlstevens:
Contours colormapping and ticks PRs now merged.

jlstevens:
This PR contains a ton of work we should probably merge sooner rather than later. I can file an issue about the things we discovered while running through all this. @philippjfr What do you think?

philippjfr:
Sounds good, let's make sure to squash merge this.

PR #1477, opened 2017-05-22
Added initial demo notebooks

philippjfr:
Adds a number of demos which will become part of the gallery. Where possible the demos are implemented across backends and link to each other (plotly only has 3d examples).


jlstevens:
Fantastic! 37 demo notebooks is a great contribution. Unfortunately, they will also be a challenge to test to make sure they keep working.

I propose merging sooner than later.

philippjfr:
>Unfortunately, they will also be a challenge to test to make sure they keep working.

Would definitely be nice if we could test these, I'd suggest we launch them in a separate build though so they run in parallel.

philippjfr:
That said we're going to be running into issues with the size of our test data if we add this many examples. Perhaps we should just convert them to scripts and test that they run without exceptions.

philippjfr:
Alternatively how about selectively skipping data tests since they are usually the bulk of the size of the test data and display tests will catch most differences.

jlstevens:
> Alternatively how about selectively skipping data tests

That is one option...but then we are at the mercy of display test stability.

philippjfr:
We should finally try to update to matplotlib 2.0 and see if that improves stability.

jlstevens:
I'll go ahead and merge.

PR #1484, opened 2017-05-25
Bokeh colormapping fixes

philippjfr:
Two colormapping fixes the first fixes a bug in assigning colors to Bars in the right order, the second allows passing a list of colors to bokeh instead of a matplotlib colormap which should always have been allowed.

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -808,6 +808,8 @@ class BarPlot(ColorbarPlot, LegendPlot):
         for name, val in mapping.items():
             if isinstance(val, basestring):
                 mapping[name] = dimension_sanitizer(mapping[name])
+            elif isinstance(val, dict) and 'field' in val:
+                val['field'] = dimension_sanitizer(val['field'])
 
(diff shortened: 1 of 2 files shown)


philippjfr:
This PR now adds support for supplying explicit lists of colors as cmaps and selecting some of the inbuilt bokeh palettes as strings. In doing is I discovered that using bokeh palettes is not sufficient, they contain at most 11 colors and do not interpolate between the colors which is not very nice for continuous values. This means for the time being we still rely on the colormaps shipped with matplotlib. I see a number of solutions:

1. Add code to interpolate between colors and generate larger colormaps on the fly, the base colors for each colormap would still have to come from somewhere though.
2. Require colorcet and only offer the colormaps it includes unless matplotlib is installed.
3. Ship a few colormaps ourselves.

jlstevens:
Ready to merge? I would consider putting your question about colormaps into an issue and assigning it to the 1.8 milestone.

philippjfr:
> I would consider putting your question about colormaps into an issue and assigning it to the 1.8 milestone.

Sounds like a plan. Yes this can be merged.

PR #1486, opened 2017-05-25
Made BokehRenderer.server_doc method more general

philippjfr:
The BokehRenderer.server_doc method was mostly used internally to create and attach a plot to a bokeh Document and return it. Apart from fixing a bug attaching the periodic utility, this PR allows the method to be called directly with a HoloViews object.

### DIFF
--- a/holoviews/plotting/bokeh/renderer.py
+++ b/holoviews/plotting/bokeh/renderer.py
@@ -120,7 +120,7 @@ class BokehRenderer(Renderer):
     def app(self_or_cls, plot, show=False, new_window=False):
         """
         Creates a bokeh app from a HoloViews object or plot. By
-        default simply uses attaches plot to bokeh's curdoc and
+        default simply attaches the plot to bokeh's curdoc and
         returns the Document, if show option is supplied creates
         an Application instance and displays it either in a browser
         window or inline if notebook extension has been loaded.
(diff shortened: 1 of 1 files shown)


holoviews/plotting/bokeh/renderer.py
```diff
@@ -155,18 +155,23 @@ def show_callback():
         return server
 
 
-    def server_doc(self, plot, doc=None):
+    @bothmethod
+    def server_doc(self_or_cls, plot, doc=None):
```
jlstevens: Isn't it really ``plot_or_obj``? I would consider renaming ``plot`` to ``obj`` as a plot is a type of object...

philippjfr:
Ready to merge.

holoviews/plotting/bokeh/renderer.py
jlstevens: Do you not agree with this suggestion then?

philippjfr: I implemented it.

jlstevens: Ah I see it. Thanks!

philippjfr: Suppose I should make the other methods consistent though.

jlstevens:
Looks good. Merging.

PR #1487, opened 2017-05-25
Ensured notebook_extension aborts outside notebook context

philippjfr:
This PR ensures that when running the notebook extension outside an IPython context it skips the notebook initialization steps. This allows for a HoloViews notebook which loads the extension to be deployed as a bokeh app.

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -185,12 +185,20 @@ class notebook_extension(param.ParameterizedFunction):
                 OutputMagic.allowed['fig'] = list_formats('fig', backend)
                 OutputMagic.allowed['holomap'] = list_formats('holomap', backend)
 
+        if selected_backend is None:
+            raise ImportError('None of the backends could be imported')
+
+        # Abort if IPython not found
+        try:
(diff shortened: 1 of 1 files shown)


holoviews/ipython/__init__.py
```diff
@@ -198,7 +198,10 @@ def __call__(self, *args, **params):
             display(HTML('<b>Warning</b>: %s' % msg))
 
         if notebook_extension._loaded == False:
-            ip = get_ipython() if ip is None else ip # noqa (get_ipython)
+            try:
+                ip = get_ipython() if ip is None else ip # noqa (get_ipython)
+            except:
+                return
```
jlstevens: Shouldn't this bit be run even if outside of an IPython context?:

```
         if selected_backend is None:
             raise ImportError('None of the backends could be imported')
         else:
             Store.current_backend = selected_backend
 ```

philippjfr: Good point.

jlstevens:
I think the ``ip`` parameter is unnecessary and should be removed. Otherwise, looks good!

jlstevens:

jlstevens:
Looks good and tests are green. Merging.

PR #1491, opened 2017-05-28
Magic preprocessors for conversion to vanilla Python

jlstevens:
This is very much a work-in-progress prototype. I will update this description if we decide to go ahead with this approach and after I have pushed forced a cleaner commit history.


jlstevens:
@jbednar @philippjfr Tests are passing so we can now discuss the main change - the refactor of the output magic to allow a new ``hv.util.output`` utility to work the same way in pure Python.

The refactor moves almost all the logic to an ``OutputOptions`` class so the actual magic class is tiny. An instance of this class is then kept on the ``Store`` so it can be used by the output magic and the utility.

I think this is the right thing to do. My major misgiving here is this is *not* code we like (or are at all proud of!) and I am worried to see this horrible code escape the ``ipython`` module where it was safely contained until now.

holoviews/ipython/__init__.py
```diff
@@ -179,18 +179,21 @@ def __call__(self, *args, **params):
                                                                               fig='svg')
                         holoviews.archive.exporters = [svg_exporter] +\
                                                       holoviews.archive.exporters
-                OutputMagic.allowed['backend'] = list_backends()
-                OutputMagic.allowed['fig'] = list_formats('fig', backend)
-                OutputMagic.allowed['holomap'] = list_formats('holomap', backend)
+
+                Store.output_options = OutputOptions()
```
philippjfr: Don't quite get this, ``Store.output_options`` is an instance but all the actual configuration is set on the class? Can it all be class based?

jlstevens: Maybe although there is this:

```
    def __init__(self, *args, **kwargs):
        self.shell = kwargs.pop('shell', None)
        super(OutputOptions, self).__init__(*args, **kwargs)
        self.output.__func__.__doc__ = self._generate_docstring()
```

I can try setting ``shell`` as a class attribute and removing this, though it will lose the generated docstring.

philippjfr: When would a user every look at the ``OutputOptions``? Seems like internal API to me.

jlstevens:
Switching everything to classmethods seems to have worked, though it does mean the param generated docstring on the magic has been lost.

philippjfr:
>Are you using the PNG export option from the latest Bokeh?

Yes, see https://github.com/ioam/holoviews/pull/1493. You can use it in the notebook too but it's slow so not too useful for interactive use.

jbednar:
Ok, great!  Looks like this support arrived just in time for what we needed it for!

holoviews/core/dimension.py
```diff
@@ -1077,13 +1077,25 @@ def __unicode__(self):
         return unicode(PrettyPrinter.pprint(self))
 
 
-
     def __call__(self, options=None, **kwargs):
+        self.warning('Use opts method instead')
```
philippjfr: I'm wary about merging this right straight away. This will overwhelm users with warnings and **a lot** of our docs and examples  are still using ``__call__``. We need to come up with a **very** soft deprecation cycle for this. I'd even say that we should have one release where the only thing nudging users towards using ``.opts`` is consistent usage of ``.opts`` throughout our docs. Anything else is just too disruptive and annoying imo. @jbednar thoughts?

jlstevens: This was mostly an experiment and having actually seen the results, I agree.

> I'd even say that we should have one release where the only thing nudging users towards using .opts is consistent usage of .opts throughout our docs.

That is a start though can we add this warning for the release cycle following that? I'll remove this warning from this PR.

jlstevens: We can do the following things:

1. Document using opts only (no argument there!)
2. Mention that people should please stop using ``__call__`` in the release notes where we mention the addition of opts.
3. When we do enable the warning, you should be able to disable it via the config system we really want and need (set deprecation settings, set styles, other things like that!)

jlstevens:
Note: We need to allow the opts string syntax (i.e using the parser) everywhere. Including where options can be set for saving/outputting data from the renderers.

jbednar:
Do we have a way that the user could explicitly enable a warning? If so we could suggest that they do so in the release notes, but otherwise not warn for now.

jlstevens:
> Do we have a way that the user could explicitly enable a warning?

Right that is another alternative. This is what I mean about wanting a 'config' system - to set things like this.

jlstevens:
@philippjfr I'm expecting the tests to go green and I think this PR should now be reviewed/merged.

It isn't 100% polished or complete but it is probably best to merge sooner rather than later, to help with debugging and to allows us to continue with the docs.

### Outstanding work

Here are the things introduced in this PR that need finishing off/fixing. These can be associated with issues as necessary:

- [ ] ``examples/README.rst`` and ``examples/examples.py`` are incomplete. I intend for ``examples.py`` to generate an index and allow grepping of notebooks referred to in the user guide.
- [ ] There is a FIXME in one of the notebook preprocessor tests relating to multiple expressions in the line line of a cell.
- [ ] The parsing of the opts string in the ``opts`` method is very rough right now and needs to be made more robust.
- [ ] The holoviews command/entry point is also quite rough and needs to be improved e.g decent commandline help using argparse etc.
- [ ] I think the idea of ``hv.config`` and a ``load_backend``` utility (used by the notebook_extension) should be added to the new utils subpackage in separate PRs.

Although there are a few loose bits and pieces, the bulk of what this PR offers is the ``hv.util.opts`` and ``hv.util.output`` helpers that can be used from pure Python instead of the magics as well as a collection of preprocessors that make use of these utils to convert notebooks to simple .py scripts. One of the main things this will enable is launching notebooks as bokeh apps with minimal (if any) changes.

holoviews/__init__.py
```diff
@@ -12,7 +12,7 @@
                             commit="$Format:%h$", reponame='holoviews')
 
 from .core import archive                                # noqa (API import)
-from .core.dimension import OrderedDict, Dimension       # noqa (API import)
+from .core.dimension import OrderedDict, Dimension, Dimensioned       # noqa (API import)
```
philippjfr: How come we need ``Dimensioned`` in the top-level namespace?

holoviews/core/dimension.py
```diff
@@ -1077,13 +1077,24 @@ def __unicode__(self):
...
         all ids are reset.
         """
+        from ..util.parser import OptsSpec
+        if isinstance(options, basestring):
+            try:
+                options = OptsSpec.parse(options)
+            except:
+                options = OptsSpec.parse(
+                    '{clsname} {options}'.format(clsname=self.__class__.__name__,
+                                                 options=options))
```
philippjfr: Seems okay for now but it might be better to list an explicit exception.

holoviews/util/__init__.py
```diff
@@ -2,13 +2,42 @@
...
+    else:
+        return StoreOptions.set_options(obj, OptsSpec.parse(options))
+
+
+def output(line=None, obj=None, **options):
+    """
+    Cell magic version acts as a no-op.
+    """
+    if obj is not None:
+        return obj
```
philippjfr: Did we want to allow this in a later PR? I remember discussing that all it has to do is temporarily set the renderer options.

holoviews/core/dimension.py
jlstevens: Sure. This falls under my bullet points for something to improve. The exception is part of it, but the whole ``try/except`` here isn't a very robust approach. I can specify an exception now but this whole thing needs improving (in another PR imho).

philippjfr:
Looks good, as you laid out in your summary there's some more work to be done here but the PR looks to be in a mergeable state. Happy to merge when you tell me.

holoviews/util/__init__.py
jlstevens: Could you outline how you can imagine that working? Even if I set the options state temporarily for an expression, it will have no effect unless that expression has some sort of side-effect, namely output. The cell magic resets everything after display and this concept of 'display' doesn't exist outside the notebook.

If you do have ideas about this, I would like to hear them as I would like these utilities to correspond to the magics as closely as possible...

holoviews/core/dimension.py
jlstevens: Specified ``SyntaxError`` in a3250417

holoviews/__init__.py
jlstevens: Fixed in f5f12c2f

holoviews/util/__init__.py
philippjfr: Good point probably doesn't make sense afterall.

jlstevens:
@philippjfr Let's merge once the tests are green.

holoviews/util/__init__.py
jlstevens: I added a suggestion above regarding a potential ``show`` call that could be injected by an optional preprocessor. Something to think about (maybe it isn't worth it?) and definitely not for this PR.

philippjfr:
Great, merging.

jlstevens:
I've added another item to the list above regarding validation. We need to make sure we address all the things in that list.

PR #1492, opened 2017-05-28
Handled bkcharts compatibility

philippjfr:
In bokeh 0.12.6 bokeh charts will move into their own repository. For the meantime we still depend on their box and whisker plot, aiming to replace it with our own implementation by holoviews 2.0. Since bokeh 0.12.6 will still require bkcharts this will buy us a bit more time. I did some extra refactoring creating a compute_plot_size utility so that ``hv.plotting.bokeh.renderer`` doesn't have to import various bokeh models and handle the conditional ``Chart`` import.

### DIFF
--- a/holoviews/plotting/bokeh/__init__.py
+++ b/holoviews/plotting/bokeh/__init__.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import
+
 import numpy as np
 
 from ...core import (Store, Overlay, NdOverlay, Layout, AdjointLayout,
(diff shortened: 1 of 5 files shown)


jlstevens:
Splitting out ``bkcharts.py`` now (with the aim of removing it later one) makes sense as does splitting out a ``compute_plot_size`` utility. Happy to merge if you think it is ready.

philippjfr:
Ready to merge.

PR #1493, opened 2017-05-29
Added support for static bokeh png rendering

philippjfr:
The bokeh 0.12.6 release will add static png rendering support and this PR exposes this support at the bokeh renderer level. Because rendering to png is so slow this is not particularly useful for working interactively but will be very useful for generating output which can be embedded in any document, including the thumbnails for our gallery.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ install:
   - conda info -a
   - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION scipy=0.18.1 numpy freetype nose bokeh=0.12.5 pandas=0.19.2 jupyter ipython=4.2.0 param pyqt=4 matplotlib=1.5.1 xarray datashader dask=0.13
   - source activate test-environment
-  - conda install -c conda-forge  -c scitools iris sip=4.18 plotly flexx
+  - conda install -c conda-forge  iris sip=4.18 plotly flexx
   - if [[ "$TRAVIS_PYTHON_VERSION" == "3.4" ]]; then
       conda install python=3.4.3;
     fi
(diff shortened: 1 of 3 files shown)


jlstevens:
Happy to merge this simple addition. My only worry is we need to keep testing it to make sure it continues to work as the Bokeh export API evolves. Might be worth having a unit test that just checks that a PNG file is generated (without trying to check the contents).

philippjfr:
>Might be worth having a unit test that just checks that a PNG file is generated (without trying to check the contents).

Sounds reasonable.

jlstevens:
I suppose getting this unit tests working means updating the bokeh version on Travis. Alternatively, we could annotate such tests like we do with 'optional' to skip them for now...

philippjfr:
>Alternatively, we could annotate such tests like we do with 'optional' to skip them for now...

It is marked optional, just like the other bokeh renderer tests.

jlstevens:
> It is marked optional, just like the other bokeh renderer tests.

We need a new annotation as Travis runs *all* tests (doesn't use ``-a '!optional'``) right now. I'm proposing an annotation that means 'don't run on travis yet' or 'will be run on travis once ready and this annotation is removed'.

philippjfr:
>I'm proposing an annotation that means 'don't run on travis yet' or 'will be run on travis once ready and this annotation is removed'.

Seems easier to just add a SkipTest that checks the bokeh version for now.

jlstevens:
> Seems easier to just add a SkipTest that checks the bokeh version for now.

Good idea.

jlstevens:
There is no need to ever remove the version check for the ``SkipTest`` either. When you say 'for now', I am happy with 'forever'! :-)

jlstevens:
The pr builds are green so I'll merge once the push build is green as well.

PR #1494, opened 2017-05-29
Fix dynamic sort

philippjfr:
The Dynamic utility converts HoloMaps to DynamicMaps by setting the HoloMap key dimension values as Dimension.values on the ``DynamicMap``. Currently they are unordered, this PR ensures they retain their original order.


jlstevens:
Fix makes sense to me and it is good you added a unit test. Happy to merge once the tests go green.

PR #1495, opened 2017-05-29
Fixed batched Spike overlay extents bug

philippjfr:
Fixes https://github.com/ioam/holoviews/issues/1480, the SpikePlot was iterating over the ``current_frame`` in batched mode for no reason causing issues when the plot was overlaid and the ``current_frame`` isn't initialized.


jlstevens:
Looks good. Merging.

PR #1501, opened 2017-05-31
Only colormap adjoined Histogram when dimensions match

philippjfr:
Fixes dimension lookup error with this example:

```python
%%opts Scatter [color_index=2 size_index=3 scaling_factor=50]
np.random.seed(10)
data = np.random.rand(100,4)

scatter = hv.Scatter(data, vdims=['y', 'z', 'size'])
scatter + scatter[0.3:0.7, 0.3:0.7].hist()
```


philippjfr:
Ready to merge, the Points/Scatter examples added to https://github.com/ioam/holoviews/pull/1476 should cover this case.

jlstevens:
Thanks for fixing this! Tests are green now, merging.

PR #1503, opened 2017-05-31
Added support for supplying fixed ticks with labels

philippjfr:
Finally adds ticking support with fixed labels to the bokeh backend, e.g.:

```python
%%opts Curve [xticks=[(0, 'Foo'), (1, 'Bar'), (2, 'Baz')]]
hv.Curve([1,2,3])
```

bokeh_plot 60

Will be available when 0.12.6 is released, both the code and the tests are appropriately guarded to work only with bokeh version > 0.12.5.


jlstevens:
Great! How about we warn and suggest upgrading bokeh if this ticker format is detected?

jlstevens:
Looks good. Merging.

PR #1508, opened 2017-06-01
Added span parameter for datashade operation

philippjfr:
Not sure ``span`` is the best name for this parameter but I just used the datashader argument name for now. Maybe ``clims`` or ``color_range`` would be better?

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 from collections import Callable, Iterable
+from distutils.version import LooseVersion
 import warnings
 
 import param
@@ -337,6 +338,11 @@ class shade(Operation):
         and any valid transfer function that accepts data, mask, nbins
         arguments.""")
 
+    clims = param.NumericTuple(default=None, length=2, doc="""
+        Min and max data values to use for colormap interpolation, when
+        wishing to override autoranging.
+        """)
+
     link_inputs = param.Boolean(default=True, doc="""
          By default, the link_inputs parameter is set to True so that
          when applying shade, backends that support linked streams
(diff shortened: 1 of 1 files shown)


jbednar:
I definitely don't like the name `span`.  Is `clims` already in use for this purpose in HoloViews?  If not it's more like a value_range.

jlstevens:
I don't think ``clims`` would clash in this case and it is clearer/more intuitive than ``span``. Seems like the issue is that this isn't what this argument is called in datashader.

philippjfr:
>Seems like the issue is that this isn't what this argument is called in datashader.

We already changed a bunch of the argument names because they're not very intuitive.

jbednar:
The datashader API needs some tidying up before I'm ready to call it 1.0, so don't worry about what it's called in datashader.  I think HV should be supplying whatever dimension range it normally uses for non-datashader plots, though it could be convenient to provide an argument to `shade` that makes it simple to override that value when needed.

jbednar:
@philippjfr, can you please rename `span` to whatever the best match to HoloViews usage is, and then make it respect the dimension's range if not overridden here?

philippjfr:
>@philippjfr, can you please rename span to whatever the best match to HoloViews usage is, and then make it respect the dimension's range if not overridden here?

Was going to go with ``value_range`` and already have those changes applied locally. Will push in a minute.

philippjfr:
Might go back to ``clims`` since it's shorter.

jlstevens:
I also think ``clims`` is better:  this operation outputs and ``RGB`` so there is color mapping and from what I understand this parameter is really about controlling colors.

jbednar:
clims is fine by me.  It's an edge case; the actual parameter has nothing to do with colors, and applies to the value rather than to any colors, but it has no effect on anything but colors, in the end.

jlstevens:
Looks good to me and the tests are green. Merging.

PR #1509, opened 2017-06-01
Ellipse parameter now supports major/minor axis lengths

jlstevens:
Implements a suggestion by @ea42gh that is backwards compatible with the old behavior. Needs testing and final discussion on whether this is useful before I add unit tests.

**Edit:** I'll also update the new element notebook for ellipse if we decide this is useful.

### DIFF
--- a/examples/elements/bokeh/Box.ipynb
+++ b/examples/elements/bokeh/Box.ipynb
@@ -30,7 +30,30 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "A ``Box`` is an annotation that takes a center x-position, a center y-position and a width:"
+    "A ``Box`` is an annotation that takes a center x-position, a center y-position and a size:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "%%opts Box (line_width=5 color='red') Image (cmap='gray')\n",
+    "data = np.sin(np.mgrid[0:100,0:100][1]/10.0)\n",
+    "data[range(40, 60) + range(40,50), range(20, 40)+ range(70,80)] = -3\n",
+    "hv.Image(data) * hv.Box(-0.2, 0, 0.2 ) * hv.Box(-0, 0, (0.4,0.9) )"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In addition to these arguments, it supports an optional ``aspect ratio``:\n",
+    "\n",
+    "By default, the size argument results in a square such as the small square shown above. Alternatively, the size can be given as the tuple ``(width, height)`` resulting in a rectangle. If you only supply a size value, you can still specify a rectangle by specifying an optional aspect value. In addition, you can also set the orientation (in radians, rotating anticlockwise):"
    ]
   },
   {
@@ -44,26 +67,7 @@
     "%%opts Box (line_width=5 color='purple') Image (cmap='gray')\n",
     "data = np.sin(np.mgrid[0:100,0:100][1]/10.0)\n",
     "data[range(30, 70), range(30, 70)] = -3\n",
-    "hv.Image(data) * hv.Box(-0, 0, 0.5 )"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
-    "In addition to these arguments, it supports an optional ``aspect ratio``:"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {
-    "collapsed": false
-   },
-   "outputs": [],
-   "source": [
-    "%%opts Box (line_width=5 color='purple') Image (cmap='gray')\n",
-    "hv.Image(data) * hv.Box(-0, 0, 0.25, aspect=3)"
+    "hv.Image(data) * hv.Box(-0, 0, 0.25, aspect=3, orientation=-np.pi/4)"
    ]
   }
  ],
--- a/examples/elements/bokeh/Box.ipynb
+++ b/examples/elements/bokeh/Box.ipynb
@@ -43,8 +43,9 @@
    "source": [
     "%%opts Box (line_width=5 color='red') Image (cmap='gray')\n",
     "data = np.sin(np.mgrid[0:100,0:100][1]/10.0)\n",
-    "data[range(40, 60) + range(40,50), range(20, 40)+ range(70,80)] = -3\n",
-    "hv.Image(data) * hv.Box(-0.2, 0, 0.2 ) * hv.Box(-0, 0, (0.4,0.9) )"
+    "data[np.arange(40, 60), np.arange(20, 40)] = -1\n",
+    "data[np.arange(40, 50), np.arange(70, 80)] = -3    \n",
+    "hv.Image(data) * hv.Box(-0.2, 0, 0.25 ) * hv.Box(-0, 0, (0.4,0.9) )"
    ]
   },
   {
@@ -66,7 +67,7 @@
    "source": [
     "%%opts Box (line_width=5 color='purple') Image (cmap='gray')\n",
     "data = np.sin(np.mgrid[0:100,0:100][1]/10.0)\n",
-    "data[range(30, 70), range(30, 70)] = -3\n",
+    "data[np.arange(30, 70), np.arange(30, 70)] = -3\n",
     "hv.Image(data) * hv.Box(-0, 0, 0.25, aspect=3, orientation=-np.pi/4)"
    ]
   }
--- a/holoviews/element/path.py
+++ b/holoviews/element/path.py
@@ -218,7 +218,7 @@ class Ellipse(BaseShape):
 
     orientation = param.Number(default=0, doc="""
        Orientation in the Cartesian coordinate system, the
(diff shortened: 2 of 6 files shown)


holoviews/element/path.py
```diff
@@ -191,24 +203,52 @@ class Ellipse(BaseShape):
...
+
+        if 'height' in params:
+            spec = params['height']
+
+        if isinstance(spec, tuple):
+            (major, minor) = spec
+            if major < minor:
+                raise ValueError('Minor axis length must be less than major axis length')
+            params['height'] = major
+        else:
```
jbednar: Is it necessary to enforce this restriction?  It's confusing if someone sets them the wrong way around, but e.g. if there are widgets, it can be tough to ensure that it doesn't get flipped in that way.  You could even consider labelling them axis1 and axis2 for that reason, and *suggesting* that people use axis1 for the major axis.

jlstevens: Sure!  @ea42gh also said something similar. Renaming to ``axis1`` and ``axis2`` would mean I would be happy to remove this restriction. If it is named 'major' and 'minor' then I really feel the latter should be smaller.

jbednar: You could also simplify it and just use `height` and `width`, so that there would be no need to deprecate.

jlstevens: I had suggested this on gitter - the counter-argument was that these aren't really the correct names when discussing ellipses.

jbednar: Naming parameters for geometrical objects is always difficult because the word that people use depends on things that aren't up to that parameter to decide.  E.g. "height" and "width" are tricky for *anything* with an orientation, since rotating by 90 degrees will make the words seem backwards.  But there's really no way around that, and the alternative of using names with no meaning at all (axis1) isn't any better, because then there's no way for people to guess which dimension will be controlled by which parameter. Given that there is already `height`, I'd suggest just keeping it and adding `width`; why not do the simplest thing, in a case where there is no truly great answer?

jbednar: And of course even "Ellipse" is the wrong word if the height and width are equal, which is the default -- there's no way to win!  So let's give up and do the simplest non-horrible thing.

jlstevens: On balance I probably agree and just adding ``width`` was my first instinct. Just having support for ``width`` and ``height`` as the two axes should be a huge improvement over having to use ``aspect``. Even if that isn't quite the right terminology for this geometric shape, those concepts are general.

jbednar: Sounds good.

ea42gh: That definitely would work well. The unnecessary restriction a<b would just force user code to guard against it, so I would definitely argue against it.

holoviews/element/path.py
```diff
@@ -193,22 +205,36 @@ class Ellipse(BaseShape):
 
     height = param.Number(default=1, doc="The height of the ellipse.")
 
-    aspect= param.Number(default=1.0, doc="The aspect ratio of the ellipse.")
+    width = param.Number(default=1, doc="The width of the ellipse.")
+
+    aspect= param.Number(default=1.0, doc="""
+       Final divisor only applied to the final height.""")
```
jbednar: The code seems to apply it to the width, not the height?

jlstevens: Well spotted! Fixed in 3c254475.

jbednar: Well, but in the case where `aspect` is useful, the width is initially set from the provided `height`.  So it's still confusing.

jlstevens: I probably wouldn't recommend using ``aspect`` anymore tbh and would consider deprecating it. I am only leaving it in right now for backwards compatibility. Should I issue a deprecation warning when it is used?

jbednar: I think aspect is very useful.  In some situations one wants to control height and width explicitly (e.g. to enclose a certain area), and in others one wants to control the shape and size separately (with aspect setting the shape, and height setting the size).  Neither case is more valid than the other.

jlstevens: > Well, but in the case where aspect is useful, the width is initially set from the provided height. So it's still confusing.

How would you like it to behave? I can't change how it works currently as that would break backwards compatibility but we could change the behavior for HoloViews 2.0.

jbednar: Oh, I'm not arguing for the code to change, only for the docstring to reflect what the code does.

holoviews/element/path.py
```diff
@@ -148,7 +149,11 @@ def clone(self, *args, **overrides):
         settings = dict(self.get_param_values(), **overrides)
         if not args:
             settings['plot_id'] = self._plot_id
-        return self.__class__(*args, **settings)
+
+        pos_args = getattr(self, '_' + type(self).__name__ + '__pos_params', [])
```
philippjfr: Is this necessary? Why would ``self.__pos_params`` work?

jlstevens: Why *wouldn't* it work? Name mangling due to the double underscore basically.

jlstevens: Oh, instead of ``getattr``? In this case it might work...

jlstevens: Nope just checked, name mangling is still an issue.

philippjfr: Might not work actually, for that reason a single underscore may have been cleaner, but there's a bunch of them now, so up to you.

jlstevens: I believe single underscore would have messed up ``Layouts`` and ``AttrTrees``. It was done like this for a reason but I can't quite remember all the details.

jlstevens: I partially remember - ``clone`` has to use ``getattr`` as not all elements need the positional parameters declared (most don't). Calling ``getattr`` will create a path on ``AttrTrees`` resulting in an annoying entry which doesn't happen for double underscored entries. You have to be careful with ``getattr`` when ``AttrTrees`` are involved...

holoviews/element/path.py
```diff
@@ -162,26 +167,64 @@ class Box(BaseShape):
...
-        The aspect ratio of the box if supplied, otherwise an aspect
-        of 1.0 is used.""")
+    orientation = param.Number(default=0, doc="""
+       Orientation in the Cartesian coordinate system, the
+       counterclockwise angle in radians between the first axis and the
+       horizontal.""")
+
+    aspect= param.Number(default=1.0, doc="""
+       Optional multiplier applied to the box length to compute the
+       width in cases where only the length value is set.""")
```
jbednar: length -> height ?

jlstevens:
@philippjfr @jbednar I think I am now happy with the API and behavior of these shapes. If you agree I'll add some unit tests, update the element notebooks and we can merge!

holoviews/element/path.py
```diff
@@ -162,26 +167,64 @@ class Box(BaseShape):
...
+    the given orientation.
+
+    The simplest (default) Ellipse is a circle, specified using:
+
+    Ellipse(x,y, diameter)
+
+    A circle is a degenerate ellipse where the width and height are
+    equal. To specify these explicitly, you can use:
+
+    Ellipse(x,y, (height, width))
```
jbednar: (width, height)

holoviews/element/path.py
jlstevens: The first argument corresponds to the width - the problem is that saying you compute the width from the width and aspect is confusing. I wanted the equivalent of 'diameter' for squashing a circle with the aspect. Box 'size' maybe?

holoviews/element/path.py
jlstevens: Thanks. Fixed in 30beb5c1.

holoviews/element/path.py
jlstevens: Tried using 'size' instead in 30beb5c1

philippjfr:
Looks good to me as well. Happy to merge after adding some tests.

holoviews/element/path.py
jbednar: Looks clearer, thanks.

jlstevens:
I've updated the element notebooks. As soon as I've added a few unit tests I think this PR will be ready1

examples/elements/bokeh/Box.ipynb
```diff
@@ -41,17 +41,19 @@
    },
    "outputs": [],
    "source": [
-    "%%opts Box (line_width=5 color='purple') Image (cmap='gray')\n",
+    "%%opts Box (line_width=5 color='red') Image (cmap='gray')\n",
     "data = np.sin(np.mgrid[0:100,0:100][1]/10.0)\n",
-    "data[range(30, 70), range(30, 70)] = -3\n",
-    "hv.Image(data) * hv.Box(-0, 0, 0.5 )"
+    "data[range(40, 60) + range(40,50), range(20, 40)+ range(70,80)] = -3\n",
```
philippjfr: Adding ranges will fail in py3.

jlstevens: Good point. What would you recommend for a py2/py3 solution?

philippjfr: Not sure tbh, np.arange but then you have to concatenate, ``list`` is also ugly.

jlstevens:
@philippjfr Ready to merge when the tests go green.

examples/elements/bokeh/Box.ipynb
jlstevens: Fixed in 9079666

philippjfr:
Looks good. Merging.

PR #1511, opened 2017-06-01
Switched Lissajous example to use Path instead of Curve

jlstevens:
Tiny fix to use ``Path`` instead of ``Curve`` for the first example in the streams tutorial as I think ``Path`` is more appropriate.


jlstevens:
@philippjfr It is a super simple change and I think it is ready to merge.

philippjfr:
Looks good.

PR #1512, opened 2017-06-02
Allowed using xarray with dask arrays

philippjfr:
XArray supports wrapping around dask arrays. Generally this doesn't cause any issues but the ``dimension_values`` method should always return an in-memory NumPy array so this PR is sufficient for us to have full support for dask arrays via xarray.

### DIFF
--- a/holoviews/core/data/xarray.py
+++ b/holoviews/core/data/xarray.py
@@ -5,6 +5,11 @@ import types
 import numpy as np
 import xarray as xr
 
+try:
+    import dask
+except ImportError:
+    dask = None
+
 from .. import util
 from ..dimension import Dimension
 from ..ndmapping import NdMapping, item_check, sorted_context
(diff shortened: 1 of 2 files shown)


jlstevens:
Looks good! 

I had a prototype line like this once before but supporting dask via xarray makes sense. My main comment is that a few unit tests would be nice (marked optional and dask would need to be available on travis).

philippjfr:
I'm now running all existing Dataset tests against an xarray dataset containing dask arrays. Those 60 unit tests should cover it pretty well.

jlstevens:
I'll merge when the tests go green.

philippjfr:
Hold off on merging, found some issues.

philippjfr:
Okay, now fixed. Ready to merge once tests pass.

jlstevens:
Tests have passed. Merging.

PR #1513, opened 2017-06-02
Allow aggregate to rasterize to target Image gridding

philippjfr:
This PR allows the aggregate datashader operation to generate rasterize data onto the same grid as an existing Image. This is very useful when working with columnar and rasterized datasets and want to merge them into the same space, allowing simple arithmetic and more complex machine learning pipelines to be performed on the data.

Technically this is a complete regridding solution even for Images but hopefully we can decide what to do about a more efficient ``gridtools`` based implementation (which I already prototyped).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,9 +27,10 @@ install:
   - conda update -q conda
   # Useful for debugging any issues with conda
   - conda info -a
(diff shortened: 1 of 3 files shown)


philippjfr:
This PR now also fixes an issue arising due to the difference of coordinate systems in datashader and holoviews. Datashader coordinates are aligned at the bottom, left edge of each bin, while holoviews coordinates are placed at the bin centers. I'm now appropriately offsetting the datashader coords and added some (optional) unit tests to check this.

jlstevens:
Looks good. Merging.

PR #1514, opened 2017-06-03
Added basic MultiInterface for Path types

philippjfr:
This is my initial attempt at making a general interface for Path types as discussed extensively in https://github.com/ioam/holoviews/issues/416. This PR should keep complete backward compatibility paving the way for a later PR that expands the capabilities of this approach. This PR will also allow me to write an initial geopandas interface for paths in geoviews.

### DIFF
--- a/holoviews/core/data/multipath.py
+++ b/holoviews/core/data/multipath.py
@@ -25,21 +25,15 @@ class MultiInterface(Interface):
         pass
 
     @classmethod
-    def get_interface(cls, data):
-        for st in cls.subtypes:
-            intfc = Interface.interfaces.get(st, None)
-            if intfc and isinstance(data, intfc.types):
-                return intfc
-
-    @classmethod
     def dimension_type(cls, dataset, dim):
-        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes)
+        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes, vdims=[])
         return ds.interface.dimension_type(ds, dim)
 
     @classmethod
     def range(cls, dataset, dim):
         ranges = []
-        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes)
+        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes,
+                           vdims=[])
         for d in dataset.data:
             ds.data = d
             ranges.append(ds.interface.range(ds, dim))
@@ -60,7 +54,8 @@ class MultiInterface(Interface):
     @classmethod
     def shape(cls, dataset):
         rows, cols = 0, 0
-        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes)
+        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes,
+                           vdims=[])
         for d in dataset.data:
             ds.data = d
             r, cols = ds.interface.shape(ds)
@@ -70,7 +65,8 @@ class MultiInterface(Interface):
     @classmethod
     def length(cls, dataset):
         length = 0
-        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes)
+        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes,
+                           vdims=[])
         for d in dataset.data:
             ds.data = d
             length += ds.interface.length(ds)
@@ -83,7 +79,8 @@ class MultiInterface(Interface):
     @classmethod
     def redim(cls, dataset, dimensions):
         new_data = []
-        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes)
+        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes,
+                           vdims=[])
         for d in dataset.data:
             ds.data = d
             new_data.append(ds.interface.redim(ds, dimensions))
@@ -92,7 +89,8 @@ class MultiInterface(Interface):
     @classmethod
     def values(cls, dataset, dimension, expanded, flat):
         values = []
-        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes)
+        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes,
+                           vdims=[])
         for d in dataset.data:
             ds.data = d
             values.append(ds.interface.values(ds, dimension))
@@ -103,7 +101,7 @@ class MultiInterface(Interface):
     def split(cls, dataset, start, end):
         objs = []
         for d in dataset.data[start: end]:
-            objs.append(dataset.clone(d, datatype=cls.subtypes))
+            objs.append(dataset.clone(d, datatype=cls.subtypes, vdims=[]))
         return objs
 
 
--- a/holoviews/core/data/multipath.py
+++ b/holoviews/core/data/multipath.py
@@ -1,0 +1,110 @@
+import numpy as np
+
+from ..util import max_range
+from .interface import Interface
+
+class MultiInterface(Interface):
+
+    types = (list,)
+
+    datatype = 'multi'
+
+    subtypes = ['dataframe', 'dictionary', 'array']
+
+    @classmethod
+    def init(cls, eltype, data, kdims, vdims):
+        new_data = []
+        for d in data:
+            d, interface, dims, extra_kws = Interface.initialize(eltype, d, kdims, vdims,
+                                                                 datatype=cls.subtypes)
+            new_data.append(d)
+        return new_data, dims, extra_kws
+
+    @classmethod
+    def validate(cls, dataset):
+        pass
+
+    @classmethod
+    def get_interface(cls, data):
+        for st in cls.subtypes:
+            intfc = Interface.interfaces.get(st, None)
+            if intfc and isinstance(data, intfc.types):
+                return intfc
+
+    @classmethod
+    def dimension_type(cls, dataset, dim):
+        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes)
+        return ds.interface.dimension_type(ds, dim)
+
+    @classmethod
+    def range(cls, dataset, dim):
+        ranges = []
+        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes)
+        for d in dataset.data:
+            ds.data = d
+            ranges.append(ds.interface.range(ds, dim))
+        return max_range(ranges)
+
+    @classmethod
+    def aggregate(cls, columns, dimensions, function, **kwargs):
+        raise NotImplementedError
+
+    @classmethod
+    def groupby(cls, columns, dimensions, container_type, group_type, **kwargs):
+        raise NotImplementedError
+
+    @classmethod
+    def sample(cls, columns, samples=[]):
+        raise NotImplementedError
+
+    @classmethod
+    def shape(cls, dataset):
+        rows, cols = 0, 0
+        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes)
+        for d in dataset.data:
+            ds.data = d
+            r, cols = ds.interface.shape(ds)
+            rows += r
+        return rows+len(dataset.data)-1, cols
+
+    @classmethod
+    def length(cls, dataset):
+        length = 0
+        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes)
+        for d in dataset.data:
+            ds.data = d
+            length += ds.interface.length(ds)
+        return length+len(dataset.data)-1
+
+    @classmethod
+    def nonzero(cls, dataset):
+        return bool(cls.length(dataset))
+
+    @classmethod
+    def redim(cls, dataset, dimensions):
+        new_data = []
+        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes)
+        for d in dataset.data:
+            ds.data = d
+            new_data.append(ds.interface.redim(ds, dimensions))
+        return new_data
+
+    @classmethod
+    def values(cls, dataset, dimension, expanded, flat):
+        values = []
+        ds = dataset.clone(dataset.data[0], datatype=cls.subtypes)
+        for d in dataset.data:
+            ds.data = d
+            values.append(ds.interface.values(ds, dimension))
+            values.append([np.NaN])
+        return np.concatenate(values[:-1]) if values else []
+
+    @classmethod
+    def split(cls, dataset, start, end):
+        objs = []
+        for d in dataset.data[start: end]:
+            objs.append(dataset.clone(d, datatype=cls.subtypes))
+        return objs
+
+
+Interface.register(MultiInterface)
--- a/holoviews/core/data/multipath.py
+++ b/holoviews/core/data/multipath.py
@@ -31,8 +31,8 @@ class MultiInterface(Interface):
     def template(cls, dataset):
         from . import Dataset
         vdims = dataset.vdims if getattr(dataset, 'level', None) is None else []
-        return dataset.clone(dataset.data[0], datatype=cls.subtypes,
-                             vdims=vdims, new_type=Dataset)
+        return Dataset(dataset.data[0], datatype=cls.subtypes,
+                       kdims=dataset.kdims, vdims=vdims)
 
 
     @classmethod
@@ -126,6 +126,9 @@ class MultiInterface(Interface):
 
     @classmethod
     def split(cls, dataset, start, end):
+        from ...element.path import BaseShape
+        if isinstance(dataset, BaseShape):
+            return [dataset]
         objs = []
         for d in dataset.data[start: end]:
             objs.append(dataset.clone(d, datatype=cls.subtypes))
--- a/holoviews/core/data/multipath.py
+++ b/holoviews/core/data/multipath.py
@@ -14,6 +14,8 @@ class MultiInterface(Interface):
     @classmethod
     def init(cls, eltype, data, kdims, vdims):
         new_data = []
+        dims = {'kdims': eltype.kdims, 'vdims': eltype.vdims}
+        extra_kws = {}
         for d in data:
             d, interface, dims, extra_kws = Interface.initialize(eltype, d, kdims, vdims,
                                                                  datatype=cls.subtypes)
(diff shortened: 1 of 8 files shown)


jlstevens:
I'm very happy to see this PR after many hours of discussion. I think our new plan is very nice it is conceptually simple, useable and backwards compatible!

jlstevens:
I'll summarize my understanding of the new proposal, if only for my benefit so I remember when I come back to this later:

1. ``MultiInterface`` will support a constructor that is simply a list of other supported chart constructors.
2. This includes the dictionary format ``{'x':xarr, 'y':arr, 'z':3}`` which allows memory-efficient specification of iso-contours (constant values per subpath).
3. We will make a semantic distinction between isocontours (such as the dictionary above, or an array where the values are all the same) and ``Paths`` where the subpaths have varying values along their length: the former go into the ``kdims`` and the latter in the ``vidms`` . This means ``Contours`` (which are iso-contours) have the levels in the ``kdims``.
4. This scheme will allow value based slicing in the following ways:

```python
# Implicit default key dimension, slicing subpaths based on their position in the list.
# Philipp: Would we have kdims=['Default'] kdim by default?
Path([{'x':xarr, 'y':yarr}, {'x':xarr, 'y':yarr}])[:,:,1:2]  

# Slicing iso-contours by value
Path([{'x':xarr, 'y':yarr, 'z':2}, 
   {'x':xarr, 'y':yarr, 'z':3}])[:,:,2:3]

# Slicing 'valued' contours by value
Path([{'x':xarr, 'y':yarr, 'z':zarr},  # zarr is not constant 
   {'x':xarr, 'y':yarr, 'z':zarr}])[:,:,2:3]

# Set selection for categorical, non-numeric (i.e labelled) isocontours
Path([{'x':xarr, 'y':yarr, 'z':'Everest'}, 
   {'x':xarr, 'y':yarr, 'z':'Ben Nevis'}])[:,:,{'Everest'}]
```

Of course, you can have multiple isocontours with the same value, in which case selecting on that value will give you multiple subpaths. This is where the ``split`` method comes in:
5. ``Path`` will support a ``split`` method to output a list of ``Path`` elements *not* using the ``MultiInterface`` i.e each output is a subpath that is a normal chart that could be cast to a ``Curve`` (for example).
6. Later on we will support this dictionary format ``{('x','y'):data, 'z':3}`` so people aren't forced to split up their data to specify iso-contours.

That summarizes my overview of the plan. There are lots of technical details to talk about but I think I've outlined what is happening from the user's perspective.

philippjfr:
Thanks for the summary, it captures everything we discussed afaik. Note that as part of this PR we will only really address 1. and 5., followup PRs will add all the new functionality. This PR exists solely as a step towards that while maintaining complete backward compatibility with the old code. The only change to the old code is that if no ``level`` is supplied, no value dimension is created, which means the test data does have to be updated. The level argument will eventually be deprecated but will be supported instead of the proposals outlined in 2., 3. and 4.

This PR is now getting very close to passing so I'll start adding some new unit tests and docstrings.

philippjfr:
@jlstevens Ready for review.

holoviews/element/path.py
```diff
@@ -73,20 +94,10 @@ def collapse_data(cls, data_list, function=None, kdims=None, **kwargs):
...
-    def dimension_values(self, dimension):
-        dim_idx = self.get_dimension_index(dimension)
-        if dim_idx >= len(self.dimensions()):
-            return super(Path, self).dimension_values(dimension)
-        values = []
-        for contour in self.data:
-            values.append(contour[:, dim_idx])
-        return np.concatenate(values) if values else []
-
+    def split(self, start=None, end=None):
```
philippjfr: Need to add a docstring here.

holoviews/core/data/multipath.py
```diff
@@ -0,0 +1,160 @@
...
+    """
+    MultiInterface allows wrapping around a list of tabular datasets
+    including dataframes, the columnar dictionary format or 2D tabular
+    NumPy arrays. Using the split method the list of tabular data can
+    be split into individual datasets.
+    """
+
+    types = ()
+
+    datatype = 'multi'
```
jlstevens: Wouldn't mind a longer, more descriptive name but can't think of a good suggestion right now.

jlstevens: If it really is only for tabular, how about 'multitable' or 'multitabular'?

holoviews/core/data/multipath.py
```diff
@@ -0,0 +1,160 @@
...
+
+    subtypes = ['dataframe', 'dictionary', 'array', 'dask']
```
jlstevens: Really this is 'subdatatypes' but 'subtypes' is fine.

holoviews/core/data/multipath.py
```diff
@@ -0,0 +1,160 @@
...
+        dims = {'kdims': eltype.kdims, 'vdims': eltype.vdims}
+        extra_kws = {}
+        for d in data:
+            d, interface, dims, extra_kws = Interface.initialize(eltype, d, kdims, vdims,
+                                                                 datatype=cls.subtypes)
+            new_data.append(d)
+        return new_data, dims, extra_kws
+
+    @classmethod
+    def validate(cls, dataset):
```
jlstevens: Will this ever be anything other than a no-op?

philippjfr: Yes, will probably add some validation to ensure that all the entries in ``.data`` are the same type and then validate the sub-elements in some way, mixing types is just too weird imo.

holoviews/core/data/multipath.py
```diff
@@ -0,0 +1,160 @@
...
+    subtypes = ['dataframe', 'dictionary', 'array', 'dask']
+
+    @classmethod
+    def init(cls, eltype, data, kdims, vdims):
+        new_data = []
+        dims = {'kdims': eltype.kdims, 'vdims': eltype.vdims}
+        extra_kws = {}
+        for d in data:
+            d, interface, dims, extra_kws = Interface.initialize(eltype, d, kdims, vdims,
+                                                                 datatype=cls.subtypes)
```
jlstevens: A couple of comments:

1. ``for d in data`` could iterate over some giant iterable (e.g an image array) if the wrong type of data is passed in. Shouldn't we do a little bit of type checking here? My basic assumption is that the input is a list.
2. Out of ``d, interface, dims, extra_kws``, the ``interface`` value is never used so that one could definitely be replaced with ``_``. This brings me to ....
3.  Both ``dims`` and ``extra_kws`` are being overwritten in this loop which means the final return of ``init`` corresponds to the values from the last interface initialized - if ``data`` isn't empty. Otherwise it  uses the ``dims`` you declare before the loop and ``{}`` for ``extra_kws``. I assume the interfaces can change in the loop but perhaps there is some sort of validation that can be done here to ensure consistency? Making the return of ``MultiInterface.init`` correspond to the return of the last interface used may work but strikes me as a little odd...

holoviews/core/data/multipath.py
jlstevens: That might be a reply to part 3. of the comment I made below...

holoviews/core/data/multipath.py
```diff
@@ -0,0 +1,160 @@
...
+    @classmethod
+    def validate(cls, dataset):
+        pass
+
+    @classmethod
+    def template(cls, dataset):
+        """
+        Returns a Dataset template used as a wrapper around the data
+        contained within the multi-interface dataset.
+        """
```
jlstevens: This seems to be an implementation detail specific to ``MultiInterface``. Do you see any reasons ``template`` should be exposed as part of the public interface API?

I also don't get why it is a template: the docstring suggests it could be called ``_metadataset`` or ``_multidataset``.

jlstevens: It seems to use ``dataset.data[0]`` so maybe it is a 'template' in the sense it is using the first data element as an 'exemplar'?

philippjfr: >This seems to be an implementation detail specific to MultiInterface. Do you see any reasons template should be exposed as part of the public interface API?

I don't consider any method on the interfaces as public API really and there are a number of these methods. I don't mind making it an underscored method though.

>I also don't get why it is a template: the docstring suggests it could be called _metadataset or _multidataset.

Not really, although template might not be the best name for it. It's used to wrap each individual element in the list of data entries and operate on it and for efficiency purposes is reused over and over again assigning the data entry to it as appropriate.

jlstevens: Seems to me that a full semantic name would be something like ``_inner_dataset_exemplar`` or ``_inner_dataset_shape`` (for a generalized nothing of 'shape') or ``_inner_dataset_template``. Is that roughly right?

holoviews/core/data/multipath.py
```diff
@@ -0,0 +1,160 @@
...
+        from . import Dataset
+        vdims = dataset.vdims if getattr(dataset, 'level', None) is None else []
+        return Dataset(dataset.data[0], datatype=cls.subtypes,
+                       kdims=dataset.kdims, vdims=vdims)
+
+    @classmethod
+    def dimension_type(cls, dataset, dim):
+        if not dataset.data:
+            return float
```
jlstevens: I don't quite get what it means when the list of inputs is empty and why it is of type float...

jlstevens: Same comment when ``(None, None)`` is the ``range`` below...

holoviews/core/data/multipath.py
```diff
@@ -0,0 +1,160 @@
...
+
+    @classmethod
+    def range(cls, dataset, dim):
+        if not dataset.data:
+            return (None, None)
+        ranges = []
+        ds = cls.template(dataset)
+
+        # Backward compatibility for level
+        level = getattr(dataset, 'level', None)
```
jlstevens: IIRC 'level' is only a contour thing. If that is true I would right something like:

``# Backwards compatibility for 'level' vdim of Contours``

holoviews/core/data/multipath.py
philippjfr: Have to think about why I did the float thing, but what would you expect the range to be?

holoviews/core/data/multipath.py
philippjfr: Contours and Polygons.

holoviews/core/data/multipath.py
```diff
@@ -0,0 +1,160 @@
...
+        if level is not None and dim is dataset.vdims[0]:
+            return (level, level)
+
+        for d in dataset.data:
+            ds.data = d
+            ranges.append(ds.interface.range(ds, dim))
+        return max_range(ranges)
+
+    @classmethod
+    def select(cls, dataset, selection_mask=None, **selection):
```
jlstevens: Needs a docstring. From what I can tell, a ``select`` on a ``MultiInterface`` simply applies the ``select`` to all the components and returns the result in a list.

philippjfr: Right, although it's not exposed at the ``Path`` level because we didn't allow slicing paths in the past. Will have to decide what to do here in future, in geopandas interface doing it properly will be trivial but handling polygon slicing correctly here would be a pain.

holoviews/core/data/multipath.py
```diff
@@ -0,0 +1,160 @@
...
+        if not dataset.data:
+            return (0, len(dataset.dimensions()))
+
+        rows, cols = 0, 0
+        ds = cls.template(dataset)
+        for d in dataset.data:
+            ds.data = d
+            r, cols = ds.interface.shape(ds)
+            rows += r
+        return rows+len(dataset.data)-1, cols
```
jlstevens: I'm not quite sure how the shapes from all the component interfaces are being folded into a single 2-tuple and whether that makes sense. In isolation, I think what ``MultiInterface`` really has is ``shapes`` (plural) that would return a list of shapes for the component interfaces.

holoviews/core/data/multipath.py
philippjfr: Yes, seems right, as you say it's an implementation detail so can have an obscure but descriptive name, will go with ``_inner_dataset_template`` for now.

holoviews/core/data/multipath.py
```diff
@@ -0,0 +1,160 @@
...
+    @classmethod
+    def length(cls, dataset):
+        if not dataset.data:
+            return 0
+        length = 0
+        ds = cls.template(dataset)
+        for d in dataset.data:
+            ds.data = d
+            length += ds.interface.length(ds)
+        return length+len(dataset.data)-1
```
jlstevens: Similar comment to the one above. I would naively expect either ``lengths`` (plural) or ``length`` to just be ``len(dataset.data)``.

holoviews/core/data/multipath.py
```diff
@@ -0,0 +1,160 @@
...
+            return dataset.data
+        new_data = []
+        ds = cls.template(dataset)
+        for d in dataset.data:
+            ds.data = d
+            new_data.append(ds.interface.redim(ds, dimensions))
+        return new_data
+
+    @classmethod
+    def values(cls, dataset, dimension, expanded, flat):
```
jlstevens: Needs a docstring to explain how all the data in the components is unpacked and returned as a numpy array.

holoviews/core/data/multipath.py
philippjfr: It's part of the core Dataset API and has to exist, and there's currently no use for shapes.

holoviews/core/data/multipath.py
```diff
@@ -0,0 +1,160 @@
...
+        return np.concatenate(values[:-1] if expanded else values) if values else []
+
+    @classmethod
+    def split(cls, dataset, start, end):
+        """
+        Splits a multi-interface Dataset into regular Datasets using
+        regular tabular interfaces.
+        """
+        from ...element.path import BaseShape
+        if isinstance(dataset, BaseShape):
```
jlstevens: What is this special handling about?

holoviews/core/data/multipath.py
philippjfr: Backward incompatible and not semantically correct, the length should return the same length as the array returned by ``values``.

holoviews/element/path.py
```diff
@@ -98,21 +109,31 @@ class Contours(Path):
     level = param.Number(default=None, doc="""
         Optional level associated with the set of Contours.""")
 
-    vdims = param.List(default=[Dimension('Level')], doc="""
+    vdims = param.List(default=[], doc="""
         Contours optionally accept a value dimension, corresponding
-        to the supplied values.""", bounds=(1,1))
+        to the supplied values.""")
```
jlstevens: Isn't this a backwards compatibility breaking change?

holoviews/core/data/multipath.py
philippjfr: Sounds good. Additionally this can be optimized a fair bit by pre-allocating the array instead of concatenating. That should also make datashading of large paths much faster.

holoviews/element/path.py
```diff
@@ -98,21 +109,31 @@ class Contours(Path):
...
 
     group = param.String(default='Contours', constant=True)
 
+    _level_vdim = Dimension('Level') # For backward compatibility
```
jlstevens: Ok, I assume this is being injected back into vdims for backwards compatibility as needed. You can ignore the comment immediately above.

holoviews/core/data/multipath.py
jlstevens: I ended up using the word 'component' for the bits of the multi-interface in my comments below. If you like this, we could standardize on it so this would be ``_component_dataset_template`` (long, I know!)

holoviews/core/data/multipath.py
jlstevens: I suppose ``range`` being ``(None,None)`` is the only sensible option. So it is just the float thing that baffles me now...

holoviews/core/data/multipath.py
jlstevens: Sure. I would then update the comment accordingly.

holoviews/core/data/multipath.py
jlstevens: I think we need to decide on the multi-interface API at the user level in a separate PR. This includes slicing, selecting subsets of paths etc.

holoviews/core/data/multipath.py
jlstevens: Right, I know ``shapes`` doesn't exist anywhere.

My questions are 1. whether this ever gets seen at the user level 2. how it is you mapped what is really multiple shapes (of the components) to a single shape.

holoviews/core/data/multipath.py
jlstevens: Pre-allocating memory sounds good. The docstring here could maybe make it clear how the format returned by``values`` relates to the ``shape``.

holoviews/element/path.py
jlstevens: Ignore. I saw the backwards compatibility handling code later on.

jlstevens:
I've made a number of comments but overall it looks good!

The biggest thing that is missing right now is an explanation of the scheme used to make the multiple components appear as one entity (mostly centering around ``values`` but also ``shape``). I'm also wondering about the validation: I like the ideal of supporting a mix of tabular input formats due to the increased flexibility but doing that properly (and raising sensible exceptions early) seems like it might be a pain.

holoviews/core/data/multipath.py
philippjfr: Turns out np.concatenate is already much faster than pre-allocating arrays.

holoviews/core/data/multipath.py
philippjfr: Now documented on the ``MultiPath`` interface and ``Path``.

philippjfr:
@jlstevens Ready for a second round of reviews and then let's merge it.

philippjfr:
Will have to regenerate the data due to a change in the help output in Options, but otherwise this seems ready.

jlstevens:
Great! 

I'll review as soon as the tests are green - I'm not expecting to request much in the second review, and certainly nothing that would affect the tests.

jlstevens:
Very happy to merge this at last! 

There are bound to be issues but it is best to merge and test (and write the user API) now. Travis tests are passing. Merging.

PR #1517, opened 2017-06-05
Defined hv.extension and hv.renderer utilities

jlstevens:
Defines ``hv.renderer`` which helps load rendering backends outside of a notebook context.


holoviews/util/settings.py
```diff
@@ -361,7 +361,7 @@ def update_options(cls, options, items):
     @classmethod
     def initialize(cls, backend_list):
         cls.backend_list = backend_list
-        backend = cls.options.get('backend', Store.current_backend)
+        backend = Store.current_backend
```
jlstevens: @philippjfr I don't quite know why I had to make this change. I'll try to summarize:

* In the magic tests, ``%unload_ext`` is called which meant that ``initialize`` was called multiple times because that unsets ``_loaded`` on the notebook extension. If the ``options`` dictionary had a value, this meant state was bleeding across tests and breaking.
* I can't tell what changed: I believe ``initialize`` would have been called multiple times in the tests before this PR too.

That said, why isn't this an appropriate fix? When you initialize, it should reset the state even if initialize was called before. Why use a value from ``cls.options`` when the purpose here is to reset it to the defaults (line below?).

holoviews/ipython/__init__.py
```diff
@@ -254,6 +191,24 @@ def _get_resources(self, args, params):
             resources = ['holoviews'] + resources
         return resources
 
+    @classmethod
+    def load_hvjs(cls, logo=False, JS=True, message='HoloViewsJS successfully loaded.'):
```
jlstevens: @philippjfr Is this ok? Anything else using this (i.e users)?

jlstevens:
@philippjfr I think this PR is ready for review once the tests are green.

holoviews/ipython/__init__.py
philippjfr: What am I meant to check here? Looks like you just moved it slightly.

jlstevens: Just wondering if anything else might have been using it. Why wasn't it a classmethod to begin with?

holoviews/util/settings.py
philippjfr: I can't be sure tbh, this code was always fairly brittle and I had to reorganize it a few times so methods may just have been badly named. I'd strongly suggest you make a notebook and then switch back and forth between backends with output a few times, rerun the notebook_extension, set fig formats then switch backends and so on. I don't think tests are currently a sufficient indication that things are still working.

jlstevens: Sure. I can do that to check. That said, those switches shouldn't be going through ``initialize`` each time anyway.

jlstevens: > I'd strongly suggest you make a notebook and then switch back and forth between backends with output a few times, rerun the notebook_extension, set fig formats then switch backends and so on.

I've just done that (included rerunning ``notebook_extension``) and it seems to be working correctly.

philippjfr: Okay happy to merge once test pass.

philippjfr:
Looks good to me.

PR #1524, opened 2017-06-06
Fixed bug in canonicalizing xarray data

philippjfr:
There is a bug in how xarray arrays are canonicalized which occurs when there is scalar dimension in the dataset which is not referenced in the kdims. This fixes that bug and also fixes another bug that ignores explicitly supplied kdims when constructing an Element from a xr.DataArray.

Fixes bug https://github.com/ioam/geoviews/issues/72


jlstevens:

PR #1529, opened 2017-06-12
Bokeh SplinePlot handles multiple cubic splines

philippjfr:
The bokeh bezier glyph supports cubic splines, by reading the matplotlib based spline codes we can easily split a set of cubic splines and draw them correctly with bokeh. This PR adds supports for drawing multiple cubic splines in this way. Splines of length 1 are simply ignored, while any other splines will raise warning. Briefly I thought I could handle quadratic and linear splines as well but that does not seem to be the case. Also adds support for instantiating and plotting empty splines with ``hv.Spline(([], []))``.

Here's an example:

bokeh_plot 66

### DIFF
--- a/holoviews/plotting/bokeh/annotation.py
+++ b/holoviews/plotting/bokeh/annotation.py
@@ -88,13 +88,26 @@ class SplinePlot(ElementPlot):
 
     def get_data(self, element, ranges=None, empty=False):
         data_attrs = ['x0', 'y0', 'x1', 'y1',
-                      'cx0', 'cx1', 'cy0', 'cy1']
-        if empty:
-            data = {attr: [] for attr in data_attrs}
-        else:
-            verts = np.array(element.data[0])
-            xs, ys = verts[:, 0], verts[:, 1]
-            data = dict(x0=[xs[0]], y0=[ys[0]], x1=[xs[-1]], y1=[ys[-1]],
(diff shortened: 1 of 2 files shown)


holoviews/plotting/bokeh/annotation.py
```diff
@@ -88,13 +88,26 @@ class SplinePlot(ElementPlot):
...
+                      'cx0', 'cy0', 'cx1', 'cy1']
+        verts = np.array(element.data[0])
+        inds = np.where(np.array(element.data[1])==1)[0]
+        data = {da: [] for da in data_attrs}
+        skipped = False
+        for vs in np.split(verts, inds[1:]):
+            if len(vs) != 4:
+                skipped = len(vs) > 1
+                continue
+            xs, ys = vs[:, 0], vs[:, 1]
```
jlstevens: Maybe zip the string keys and values and loop over them to append?

jlstevens:
Looks good! Not sure what unit tests could be added and I made one comment to try to avoid all the lines using ``append`` but otherwise I am happy to merge.

philippjfr:
Simplified the SplinePlot implementation. We could expand on the Spline Element notebook with the hope that they will be tested eventually.

jbednar:
Looks good!

jlstevens:
Looks good! I'll merge once the tests are green.

jlstevens:
Tests have passed. Merging.

PR #1532, opened 2017-06-12
Suppress bokeh validation warnings

philippjfr:
Bokeh has recently added various validation warnings which are helpful for developers but in most of our usage cases warn about deliberate decisions by us, such as padding a bokeh layout with empty plots. This PR suppresses those warnings.

### DIFF
--- a/holoviews/plotting/bokeh/renderer.py
+++ b/holoviews/plotting/bokeh/renderer.py
@@ -1,10 +1,11 @@
 from io import BytesIO
+import logging
 
 import numpy as np
 import param
 from param.parameterized import bothmethod
 
-
+import bokeh.core
 from bokeh.application.handlers import FunctionHandler
 from bokeh.application import Application
 from bokeh.document import Document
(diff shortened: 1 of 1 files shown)


bryevdv:
FYI would appreciate (eventually) a PR to bokeh itself to allow configuration of this e.g. silencing all or some selected validation messages.

philippjfr:
>FYI would appreciate (eventually) a PR to bokeh itself to allow configuration of this e.g. silencing all or some selected validation messages.

I'll definitely look into that. Any suggestions on what this might look like? A registry of warnings maybe that you can modify inplace?

bryevdv:
yes, probably a hidden module variable that holds messages to filter or suppress and a couple of public functions to manipulate the filter data. I guess there is a question of also whether to validate and not report, or simply not validate at all.

philippjfr:
Yes, let's merge for now, I'll just open an issue on holoviews for now reminding me to find a way to suppress the warnings more selectively probably by making the suggested changes in bokeh.

PR #1533, opened 2017-06-13
Escaping braces in opts preprocessor

jlstevens:
Tiny PR where I fix an escaping issue in the opts preprocessor (e.g when normalization options are used) as well as an outdated exception message I spotted in ``Renderer``.


jlstevens:
@philippjfr This is a tiny PR so I would go ahead and merge as I don't plan to do anything else with it.

philippjfr:
Sounds good, this is untested anyway so I will just merge.

PR #1535, opened 2017-06-13
Added Square limit example notebook

jlstevens:
In addition to adding this notebook to ``examples/topic`` and ``examples/demos``, we decided to standardize all non element notebooks to underscore separated lowercase filenames.


jlstevens:
@philippjfr I tried to use the empty spline in the matplotlib demo but I couldn't get it to work. I was having troubles with shapes and mixed datatypes which meant the empty Spline was more difficult to work with than the ``nan`` version.

philippjfr:
Okay, will have to revisit that. What were the empty splines good for anyway?

jlstevens:
> What were the empty splines good for anyway?

As a base case for recursion - the definition involves combining splines together including empty ones.

jlstevens:
@philippjfr I've now added all the notebooks I want and would ready to merge except for two issues:

1. The empty spline issue mentioned already (not a deal breaker)
2. The matplotlib version is red and the bokeh version is blue. This will be glaringly obvious when switching between bokeh/matplotlib in the gallery.

philippjfr:
>The empty spline issue mentioned already (not a deal breaker)

Okay, well the Spline constructor should handle the empty case and convert the empty lists to a tuple of ``(np.zeros((0, 2)), np.zeros((0,))``. You can file an issue maybe.

>The matplotlib version is red and the bokeh version is blue. This will be glaringly obvious when switching between bokeh/matplotlib in the gallery.

Let's just add this to the style change issue/PR.

philippjfr:
Looks good, merging then rebuilding the build website.

PR #1541, opened 2017-06-14
Allow redim.label if no custom label set

philippjfr:
As the title says this allows using redim.relabel to set a custom label as long as the current label is the same as the name, i.e. no custom label has been supplied previously.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -161,7 +161,7 @@ class redim(object):
     def label(self, specs=None, **values):
         for k, v in values.items():
             dim = self.parent.get_dimension(k)
-            if dim.name != dim.label and dim.name != v:
+            if dim and dim.name != dim.label and dim.name != v:
                 raise ValueError('Cannot override an existing Dimension label')
         return self._redim('label', specs, **values)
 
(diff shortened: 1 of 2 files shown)


holoviews/core/dimension.py
```diff
@@ -158,6 +158,13 @@ def value_format(self, specs=None, **values):
     def range(self, specs=None, **values):
         return self._redim('range', specs, **values)
 
+    def label(self, specs=None, **values):
+        for k, v in values.items():
+            dim = self.parent.get_dimension(k)
```
jlstevens: Seems like ``None`` is returned if there is no match for ``dim``? In which case a suitable exception/warning should be given if ``dim==None``.

philippjfr: ``redim`` doesn't validate Dimensions so you can apply it nested datastructures. That said I'd possibly be open to tightening that up when explicit ``specs`` are supplied.

jlstevens: Ah right, thanks for the reminder. It is annoying but I can't see a way around that issue right now...

jlstevens:
Looks good! Happy to merge when the tests pass.

jlstevens:
I see a display transient in the 2.7 pr build but otherwise the tests are passing. Merging.

PR #1542, opened 2017-06-15
Minor fixes for unnecessary warnings/exceptions

jlstevens:
This PR allows reuse of ``redim.label`` if the supplied label is unchanged and prevents a warning about label overrides between the keyword/tuple format if the label is in fact the same.


jlstevens:
The push build has passed and this is a very simple PR. I think it can be merged right away.

philippjfr:
Thanks for the fixes. Merging.

PR #1543, opened 2017-06-15
Small fixes for computing spreadfn aggregates

philippjfr:
Fixed a small bug when aggregating with a spreadfn. In writing unit tests for this I also noticed that pandas uses ddof=1 (delta degrees of freedom) while the numpy default is ddof=0. I've adjusted the code to catch those cases and change the pandas version to ddof=0.


philippjfr:
Looks like I need to rebuild test data, the Columnar_Data examples of using the spread function were using pandas and are therefore affected by the ``ddof`` change.

jlstevens:
Sounds like a bad bug - glad it is fixed and I'll merge once the tests are passing again.

jlstevens:
Tests are green. Merging.

Issue #1411, opened 2017-05-05
Output with bokeh exports faulty html file


jendas1:
Hi,
I have tried outputing html file with jupyter magic:
```python
%%output filename="holoviews"
x = [1, 2, 3, 4, 5]
y = [6, 7, 2, 4, 5]
hv.Curve((x,y),label="Temp")
```
but the resulted file displays nothing.
After further investigation I was able to see an output by adding sources to bokeh.js and bokeh.css from cdn.
The faulty output html: https://pastebin.com/1eCSYaUn

philippjfr:
Fixed by the PR referenced above. Closing.

PR #1547, opened 2017-06-15
Bokeh renderer now saves necessary HTML assets to file

jlstevens:
I did the simplest thing that I could think of to address issue #1411. I'm happy to hear any suggestions on how to improve it and I just need to test on Python 2: I get nervous when I see anything relating to unicode!

```python
%%output filename="output"
x = [1, 2, 3, 4, 5]
y = [6, 7, 2, 4, 5]
hv.Curve((x,y),label="Temp")
```

Now generates a self-contained HTML file:

image

### DIFF
--- a/holoviews/plotting/bokeh/renderer.py
+++ b/holoviews/plotting/bokeh/renderer.py
@@ -93,6 +93,12 @@ class BokehRenderer(Renderer):
         elif fmt == 'json':
             return self.diff(plot), info
 
+    @bothmethod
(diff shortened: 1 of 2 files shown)


philippjfr:
As long as you tested this on py2 and py3 and perhaps use a plot with a unicode character in the title I'm happy to merge once tests pass.

jlstevens:
This worked with py2 and py3:

philippjfr:
Great, will merge once tests pass.

PR #1548, opened 2017-06-15
Output files generated with hv.output now respect all settings

jlstevens:
My initial approach was too simplistic and didn't use all the supplied keywords. For instance this now correctly outputs an SVG file:

image

In addition I added keyword validation to make sure only the appropriate keywords are supplied.


jlstevens:
@philippjfr Tests have passed and ready for review. There isn't much too it so I think it can just be merged.

philippjfr:
Looks good, merging.

PR #1545, opened 2017-06-15
Fix issues with empty aggregates and Bars

philippjfr:
Fixes issues aggregating empty Datasets and displaying empty Bars.

PR #1549, opened 2017-06-15
Fixed newly introduced bar width bug

philippjfr:
Introduced a bug in #1545, the width variable is used again later and shouldn't be overridden.


jlstevens:

PR #1550, opened 2017-06-15
hv.output can now warn if an object is supplied without a filename

jlstevens:
Simple PR that allows the ``hv.output`` utility to warn about object supplied without an export filename. The ``filename_warning`` parameter can be set at the class level or supplied as a keyword:

image


jlstevens:
@philippjfr Tests pass and I think it is ready to review/merge.

PR #1555, opened 2017-06-16
Made count None by default in DynamicMap.periodic

jlstevens:
Simple PR that allows me to use:

```
dmap.periodic(0.01, timeout=60)
```

Without the awkward:

```
dmap.periodic(0.01, count=None, timeout=60)
```

Only the first argument (the period) is mandatory so if you supply nothing else it will run forever. This is fine for the blocking version (now the default) and bokeh server and this combination is only disabled for the threaded version where killing threads can be a problem.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -611,7 +611,7 @@ class periodic(object):
         self.dmap = dmap
         self.instance = None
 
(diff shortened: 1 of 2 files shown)


philippjfr:
This makes sense to me now, happy to merge when tests pass.

PR #1556, opened 2017-06-16
Added boids example to topics/simulation/boids.ipynb

jlstevens:
PR title says it all. Here is what it looks like:

boids

I decided to strip out my attempts at interactive boids for now as that needs more work.


jlstevens:
@philippjfr Ready to merge. It will be good to have a second thing under ``example/topics``

philippjfr:
Great, looks good. Not clear to me where we will keep and build the topics yet but I'm happy to merge.

PR #1557, opened 2017-06-16
Fix to square limit example so it works on Python 2

jlstevens:
PR contents described by the title. Problem was integer division in Python 2.


jlstevens:
@philippjfr The one test failure is a transient. Ready to merge.

philippjfr:
Okay, no need to wait on tests for these anyway as they're untested for now. Merging.

PR #1559, opened 2017-06-17
Updated the website search to search holoviews rather than bokeh

philippjfr:
As title says, the new theme relied on a custom Google search which was still pointing at bokeh. I've now set up a custom search for holoviews.

### DIFF
--- a/doc/holoviews_theme/search.html
+++ b/doc/holoviews_theme/search.html
@@ -21,19 +21,17 @@
   </form>
 
   <div id="search-results">
-
-    <script>
-      (function() {
-        var cx = '007519673294607119974:n9uana_s26m';
-        var gcse = document.createElement('script');
-        gcse.type = 'text/javascript';
-        gcse.async = true;
-        gcse.src = (document.location.protocol == 'https:' ? 'https:' : 'http:') +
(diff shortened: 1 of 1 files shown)


jlstevens:
Great! Should I just merge?

philippjfr:
Yes, please do. We may want to switch back to the inbuilt search but that is limited in all kinds of ways (e.g. it only processes the first search term, anything else is ignored). The problem with Google or any free custom search is they are likely to include ads, Google has even discontinued their paid custom search engine forcing every new user to use the free, ad-supported version. Let's merge because anything is better than having it search the bokeh site, but keep looking for better alternatives.

jlstevens:
Feel free to file an issue about improving search. I don't consider it a high priority right now as long as we have some sort of working search functionality.

philippjfr:
Maybe I'll file an issue the first time it tries to sell me something I don't like. Although I can't say I ever use the search.

jlstevens:
> Although I can't say I ever use the search.

Same. I suspect it is most likely to get fixed as a consequence of a user complaint than one of us noticing it. For that reason I think it is worth filing an issue now even if we don't get around to addressing it any time soon.

PR #1561, opened 2017-06-17
Added support for Empty adjoint plots in bokeh

philippjfr:
This was already handled correctly by matplotlib but https://github.com/ioam/holoviews/issues/1478 highlighted that bokeh did not handle padding for empty adjoints correctly. Also added a unit test, and here is an example:

```python
hv.AdjointLayout([hv.Curve([0,1,1,2,3]), hv.Empty() ,hv.Curve([0,1,1,0,1])(plot=dict(height=100))])
```

screen shot 2017-06-17 at 2 10 35 pm

### DIFF
--- a/holoviews/core/layout.py
+++ b/holoviews/core/layout.py
@@ -31,7 +31,7 @@ class Composable(object):
 
 
     def __lshift__(self, other):
-        if isinstance(other, (ViewableElement, NdMapping)):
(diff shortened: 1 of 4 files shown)


jlstevens:
Looks good. Happy to merge if it is ready and the tests turn green.

PR #1563, opened 2017-06-17
Label bokeh tabs with subplot title if possible

philippjfr:
As outlined in https://github.com/ioam/holoviews/issues/1425, this PR uses the subplots ``_format_title`` method to compute tab titles. This is more consistent affords better control over the title using the ``title_format`` plot option and by setting just a ``label``. The tab titles do not contain information about the current frame, i.e. HoloMap/DynamicMap key dimension information. If no label, group or custom title_format is set it will revert back to using the Layout/Overlay path to compute the title.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1302,7 +1302,9 @@ class OverlayPlot(GenericOverlayPlot, LegendPlot):
             if self.batched:
                 self.handles['plot'] = child
             if self.tabs:
-                title = get_tab_title(key, frame, self.hmap.last)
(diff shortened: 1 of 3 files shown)


jlstevens:
What if the title is very long? Does it have any rules to truncate?

philippjfr:
No, but it didn't before either and at least now you can control it with the ``title_format`` if you have to. I could have a hard limit I suppose but that would be fairly arbitrary.

jlstevens:
Ok, that seems fine then and I would rather not add a parameter for truncating. The push build passed as did the Python 3.5 pr build so I think it is safe to merge.

PR #1564, opened 2017-06-17
Correctly associate hover tool instances with plot renderer

philippjfr:
When passing an explicit HoverTool instance to an ElementPlot it should associate that tool with only that subplot. This PR fixes the issue described in https://github.com/ioam/holoviews/issues/1449 and adds a unit test.


jlstevens:
Looks good. Will merge when tests pass.

PR #1566, opened 2017-06-17
Fixed bug plotting empty bokeh SpikesPlot

philippjfr:
As outlined in https://github.com/ioam/holoviews/issues/1497, the bokeh SpikesPlot breaks when the Spikes Element is empty at the moment. This ensures it works correctly and adds a unit test.


jlstevens:
Looks good. Merging.

PR #1568, opened 2017-06-17
Switched bokeh SideHistogram to use ybox_select for colormapping

philippjfr:
Slightly better UI to select the colormapping range on an adjoined Histogram.


jlstevens:
I agree 'ybox_select' is more intuitive. I'll merge when the tests are green.

philippjfr:
Have to rebuild tests on this one.

jlstevens:
The pr build is now green. Merging.

PR #1570, opened 2017-06-17
Importing unicode_literals into histogram example notebook

jlstevens:
PR title says it all - thumbnailing has to be run with Python 2 and this should allow the thumbnail to be generated for this example.

### DIFF
--- a/examples/demos/bokeh/histogram_example.ipynb
+++ b/examples/demos/bokeh/histogram_example.ipynb
@@ -19,6 +19,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
+    "from __future__ import unicode_literals\n",
     "import numpy as np\n",
     "import scipy\n",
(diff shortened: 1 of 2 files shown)


jlstevens:
As it turns out this may not actually help and the ``from __future__`` import needs to be in the thumbnailer script. Problem is I can't reproduce the unicode error seen in doc building so I can't test whether that would help either.

@philippjfr  Happy to see this PR closed and I am happy to merge a PR with this import in the thumbnailer script in ``nbpublisher`` if it helps fix the issue with the [thumbnail for this notebook](http://build.holoviews.org/gallery/demos/bokeh/histogram_example.html#bokeh-gallery-histogram-example).

philippjfr:
I'll close and see if I can fix the issue myself in the thumbnailer. Thanks for submitting the fix though.

PR #1571, opened 2017-06-17
Added verhulst_mandelbrot demo notebook for bokeh and matplotlib

jlstevens:
Adds one of my favorite visualizations from an old notebook I had:

Bokeh:

</img>

Matplotlib:

</img>

There is enough material that I could write a topic notebook to expand on this later on. My main worry with this example is that Bokeh seems to struggle with all the green points a lot more than matplotlib does (which admittedly never needs any re-rendering).


philippjfr:
Not a huge issue and happy to merge without it, but is there maybe a nicer color combination? Not a huge fan of the current combination of the red, green and blue.

jlstevens:
I'm open to whatever suggestions you have. What do you think would work better?

jlstevens:
Matplotlib handles gen=200 with little issue while Bokeh chokes and dies on all those points (had to set ``gen=110``).

Now enabled ``logz=True`` which is prettier.

</img>

To make this work with matplotlib I had to add 1 to the iteration so the lowest value is one, not zero. This is a hack...

jlstevens:
@jbednar Bokeh may struggle but not if you use datashader!

25 million points:

bifurcation

Seems like a great synthetic example to me...

jlstevens:
Without the mapping function and with ``x_sampling`` set:

bifurcation

jbednar:
Nice. So is datashader re-rendering on a zoom, or merely resampling?

jlstevens:
The issue with logz has now been resolved in matplotlib and I've removed the hack that avoided zero values. Ready to merge.

philippjfr:
Looks good, merging.

PR #1572, opened 2017-06-18
Box spec is now matches Ellipse

jlstevens:
This is a simple fix to ``Box`` so it uses the (width, height) spec, matching ``Ellipse``. I think what happened is I fixed the flip in ``Ellipse`` but forgot to make the same fix in ``Box``.


jlstevens:
@philippjfr Tests are passing. Ready to merge.

philippjfr:
Any idea what's up with the coverage drop?

philippjfr:
I'll merge and assume coveralls just got confused.

PR #1574, opened 2017-06-19
hv.renderer tries to import backend if not available

philippjfr:
This avoids having to import the backend when deploying an app. The ``hv.renderer`` function now simply attempts to import a backend if it's not already available and errors if the import fails.

### DIFF
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -163,13 +163,13 @@ def renderer(name):
     Helper utility to access the active renderer for a given extension.
     """
     try:
+        if name not in Store.renderers:
+            extension(name)
         return Store.renderers[name]
-    except KeyError:
-        msg = ('Could not find a {name!r} renderer in list of available '
-               'renderers: {available}. Please make sure the appropriate extension '
-               'has been loaded with hv.extension().')
-        raise KeyError(msg.format(name=name,
-                                  available=', '.join(repr(k) for k in Store.renderers)))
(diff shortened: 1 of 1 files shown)


holoviews/util/__init__.py
```diff
@@ -163,13 +163,16 @@ def renderer(name):
...
+            __import__('holoviews.plotting.%s' % imp)
         return Store.renderers[name]
-    except KeyError:
+    except ImportError:
         msg = ('Could not find a {name!r} renderer in list of available '
                'renderers: {available}. Please make sure the appropriate extension '
                'has been loaded with hv.extension().')
-        raise KeyError(msg.format(name=name,
-                                  available=', '.join(repr(k) for k in Store.renderers)))
+        available = ', '.join(repr(k) for k in Store.renderers))
```
jlstevens: There is now a syntax error at the end with ``))`` instead of just ``)``

jlstevens:
Why not use ``hv.util.extension`` to do the imports? I would rather keep the backend import machinery in one place.

jlstevens:
Looks good and happy to merge if you've tested the latest version.

philippjfr:
py2 build and py3 build passing when looking at push and PR builds. Ready to merge.

jlstevens:
I restarted the failing builds and they are running now. Shouldn't take too long...

jlstevens:
Tests are green. Merging.

PR #1581, opened 2017-06-19
Moved and addded bokeh server apps to examples

philippjfr:
Revised existing bokeh apps and moved them to ``examples/apps/bokeh`` and added a few new apps. These apps can definitely still be polished but this is a start.


jlstevens:
> These apps can definitely still be polished but this is a start.

Agreed. Nothing here is tested so I'll just merge.

PR #1582, opened 2017-06-19
Added 3D Element notebooks and made small fixes

philippjfr:
Still need to add Scatter3D.


jlstevens:
Looks good to me and there is no need to wait for the tests as these new notebooks aren't tested.

PR #1585, opened 2017-06-19
Cleaned up demo notebooks

philippjfr:
Removed titles from demo notebooks so sphinx can index them correctly. Also cleared a couple notebooks which still had embedded output and renamed two notebooks.


jlstevens:
Looks good. Happy to merge.

jlstevens:
I'll merge right away as this stuff is not tested at the moment.

Issue #1490, opened 2017-05-27
Give plot instances rich display in the display hooks


jlstevens:
I believe you can now do something like:

```python
plot = BokehRenderer.get_plot(obj)
```

To get a plot instance - note, I would like this renamed to something like ``plotobj`` instead of ``get_plot`` before 1.8!

If ``plot`` were registered with the display hooks, you could just view it as normal in the notebook:

```python
plot = BokehRenderer.get_plot(obj)
plot
```

The advantage of this is that with a handle on the plotting object you can use the ``Plot`` API, e.g to call ``.refresh()``. This would be useful for advanced use-cases where you might want to control plotting updates directly.

philippjfr:
Just tested it and this works:

```python
from holoviews.plotting import Plot
def plot_display(plot):
    return plot.renderer.html(plot)

html_formatter.for_type(Plot, plot_display)
```

jlstevens:
@philippjfr Want me to just add this to the display hooks so I can close this issue?

PR #1588, opened 2017-06-20
Registered Plot instances with the display hooks

jlstevens:
Addresses #1490 so that this works (also works with matplotlib):

image

My main worry about this is it would make it easy to confuse plot instances with HoloViews elements...


jlstevens:
@philippjfr Tests are passing. Ready for review/merge.

philippjfr:
Looks good.

PR #1589, opened 2017-06-20
Added DynamicMap.opts method to apply opts to returned object

philippjfr:
When applying opts to a DynamicMap using the ``.opts`` method the options sometimes aren't applied correctly. By mapping the applied opts on the DynamicMap the opts are applied directly to the returned Elements. I think the issue may be that the ``id`` doesn't get copied or applied correctly, which means that the ``DynamicMap._style`` method doesn't apply the options correctly.


jlstevens:
Tests are now passing. Merging.

PR #1592, opened 2017-06-21
Compatibility for png export in bokeh 0.12.7

philippjfr:
Minor fix for API change for png export in bokeh 0.12.7devX.

### DIFF
--- a/holoviews/plotting/bokeh/renderer.py
+++ b/holoviews/plotting/bokeh/renderer.py
@@ -82,7 +82,10 @@ class BokehRenderer(Renderer):
             if bokeh_version < '0.12.6':
                 raise RuntimeError('Bokeh png export only supported by versions >=0.12.6.')
             from bokeh.io import _get_screenshot_as_png
-            img = _get_screenshot_as_png(plot.state)
+            if bokeh_version > '0.12.6':
+                img = _get_screenshot_as_png(plot.state, None)
+            else:
+                img = _get_screenshot_as_png(plot.state)
             imgByteArr = BytesIO()
             img.save(imgByteArr, format='PNG')
             return imgByteArr.getvalue(), info


jlstevens:
Looks good. You could maybe set up a unit test to check for any output when requesting a PNG from Bokeh. But then we would probably need to update Travis with the new dependencies.

I'll leave that up to you but I am happy to merge.

philippjfr:
>But then we would probably need to update Travis with the new dependencies.

Right that would probably require rebuilding the tests.

jlstevens:
Right. Might be worth doing later and if you agree just file an issue as a reminder. Merging.

philippjfr:
>In that case it might be worth doing later and if you agree just file an issue as a reminder. Merging.

At a later point we could even consider having some bokeh display tests which use png export.

jlstevens:
> At a later point we could even consider having some bokeh display tests which use png export.

Agreed. Probably worth filing an issue. Not sure if the output is byte-by-byte consistent but we would certainly find out!

holoviews/plotting/bokeh/renderer.py
```diff
@@ -82,7 +82,10 @@ def __call__(self, obj, fmt=None, doc=None):
             if bokeh_version < '0.12.6':
                 raise RuntimeError('Bokeh png export only supported by versions >=0.12.6.')
             from bokeh.io import _get_screenshot_as_png
-            img = _get_screenshot_as_png(plot.state)
+            if bokeh_version > '0.12.6':
+                img = _get_screenshot_as_png(plot.state, None)
+            else:
+                img = _get_screenshot_as_png(plot.state)
```
bryevdv: Well this sucks, I wish you'd mentioned you were using a private API, perhaps we could have made a public one with better guarantees

philippjfr: Yes, should have mentioned it, it's my fault. I was going to bring it up but we needed something that would work for this dev release and I only noticed last night.

PR #1593, opened 2017-06-21
Fixed typo in dask dataframe interface

philippjfr:
Tiny typo in the dask interface. Fixes the issue mentioned here: https://stackoverflow.com/questions/44666455/plotting-tabular-data-with-holoviews


jlstevens:
Probably not tests so I am happy to merge right away if you request it (instead of waiting for tests).

jlstevens:
I think there is one transient error in the Python 3 pr build. The py2 pr build was fine though and the push build is green. Merging.

PR #1599, opened 2017-06-22
Fix ndloc on descending coordinates

philippjfr:
Currently ndloc does not work when the coordinates are descending rather than ascending, this fix ensures ndloc only ever sees the ascending coordinates. I've also made sure that ``select`` returns the same datatype if possible. Will add unit tests shortly.


jlstevens:
@philippjfr Looks good. Ready to merge?

philippjfr:
Yes, ready.

PR #1603, opened 2017-06-24
Fixes for bound_range utility

philippjfr:
Since the introduction of ``ndloc`` the precise numerical precision of Image coordinates is not a real issue anymore. This PR drops the rounding that we previously applied to the density based on the system float precision. This means that Image can handle very small values, which are currently rounded to zero. I've also filtered some warnings.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -1481,6 +1481,11 @@ def bound_range(vals, density):
     if not density:
         with warnings.catch_warnings():
             warnings.filterwarnings('ignore', r'invalid value encountered in double_scalars')
-            density = 1./((high-low)/(len(vals)-1))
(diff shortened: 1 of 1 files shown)


philippjfr:
Looks like there's a test failure here:

>AssertionError: 2.9999999999999996 != 3

Suppose this is expected but we need to decide whether nicely rounded density values are more important than handling very small value ranges, which a user reported issues with.

jlstevens:
> Suppose this is expected but we need to decide whether nicely rounded density values are more important than handling very small value ranges, which a user reported issues with.

Any way to detect tiny ranges in which case rounding should be disabled?

jlstevens:
Fix makes sense to me: if after rounding you get zero, use the full precision. if this works correctly, I am happy to merge.

philippjfr:
Great, ready to merge then.

PR #1605, opened 2017-06-25
Define Palettes and Cycles from backend specific colormaps/palettes

philippjfr:
Since we no longer import the matplotlib backend the list of Palettes are only populated if that particular backend is imported. This PR does several things:

1) A Cycle can be instantiated with a key or a list of values without having to specify the parameter name, i.e. before you had to do ``Cycle(values=['red', 'green', 'blue'])`` or ``Cycle(key='default_colors')`` now you can do ``Cycle(['red', 'green', 'blue'])`` or ``Cycle('default_colors')``.
2) We now define a number of default Cycles for both matplotlib and bokeh, since there is at least some overlap we can always rely on things like ``Set1-3``, ``Dark2`` and a few others to be there.
3) We also define default Palettes in bokeh.

The behavior of ``Palette`` has always been a bit odd and we should probably discuss/address that for 2.0, since it samples the provided function like a colormap and therefore does not cycle. I think that distinction still makes sense as cycling is what Cycle is for but we need to document this at some point. For the time being this is 1) completely backward compatible, 2) allows using categorical colormaps as Cycles and 3) ensures that at least a subset of categorical colormaps/palettes are available for both matplotlib and bokeh.

### DIFF
--- a/holoviews/plotting/bokeh/__init__.py
+++ b/holoviews/plotting/bokeh/__init__.py
@@ -1,6 +1,7 @@
 from __future__ import absolute_import
 
 import numpy as np
+from bokeh.palettes import all_palettes
 
 from ...core import (Store, Overlay, NdOverlay, Layout, AdjointLayout,
                      GridSpace, GridMatrix, NdLayout, config)
@@ -9,7 +10,7 @@ from ...element import (Curve, Points, Scatter, Image, Raster, Path,
                         Box, Bounds, Ellipse, Polygons, BoxWhisker,
                         ErrorBars, Text, HLine, VLine, Spline, Spikes,
                         Table, ItemTable, Area, HSV, QuadMesh, VectorField)
-from ...core.options import Options, Cycle
+from ...core.options import Options, Cycle, Palette
 
 try:
     from ...interface import DFrame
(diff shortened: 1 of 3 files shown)


holoviews/plotting/bokeh/__init__.py
```diff
@@ -106,6 +107,17 @@
 Cycle.default_cycles['default_colors'] =  ['#30a2da', '#fc4f30', '#e5ae38',
                                            '#6d904f', '#8b8b8b']
 
+# Define Palettes and Cycles from bokeh palettees
```
jlstevens: Typo: palettees

holoviews/plotting/bokeh/__init__.py
```diff
@@ -106,6 +107,17 @@
...
+def colormap_generator(palette):
```
jlstevens: Any reason not to just return a lambda?

jlstevens:
@philippjfr Other than the two comments I made, looks good.

I would like to merge this ASAP as this would help me for some of the examples I am currently working on...

jlstevens:
> I think that distinction still makes sense as cycling is what Cycle is for but we need to document this at some point.

Agreed. I think it might be worth filing an issue to discuss palettes but I don't think that discussion should hold up this PR.

holoviews/plotting/bokeh/__init__.py
philippjfr: Palettees?

philippjfr: Ah I'm blind.

jlstevens:
Looks good. Merge when the tests pass?

philippjfr:
Yes, sounds good, here's the set of categorical Cycles that are available for both matplotlib and bokeh:

```
{u'Accent',
 u'Dark2',
 u'Paired',
 u'Pastel1',
 u'Pastel2',
 u'Set1',
 u'Set2',
 u'Set3'}
```

jlstevens:
Wonder where we should record that information about what is consistent and available...

jlstevens:
Maybe colormaps/palettes should have a user guide?

philippjfr:
>Maybe colormaps/palettes should have a user guide?

Seems sensible, or fold it into a more general user guide about plotting in addition to the backend specific guides we already have.

jlstevens:
Tests are green. Merging.

PR #1607, opened 2017-06-26
Add .idea to .gitignore for PyCharm users

jordansamuels:


jlstevens:
Happy to merge right away.

PR #1609, opened 2017-06-26
Command utility improvements

jlstevens:
This PR improves the usability of the new ``holoviews`` command.

### DIFF
--- a/holoviews/util/command.py
+++ b/holoviews/util/command.py
@@ -1,3 +1,4 @@
+#! /usr/bin/env python
 """
 python -m holoviews.util.command Conversion_Example.ipynb
 OR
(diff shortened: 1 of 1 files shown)


jlstevens:
@philippjfr Ready for review.

philippjfr:
Looks good to me. I'll merge once its green (even though it's untested).

holoviews/util/command.py
```diff
@@ -1,16 +1,28 @@
+#! /usr/bin/env python
```
jbednar: Is that the recommended most cross-platform approach?  Presumably that won't work on Windows, but just "python" wouldn't work alone?

jlstevens: This is only for testing really. In setup.py this script is setup as an entry point which means it will just be available, regardless of this line.

jbednar: Got it.

PR #1610, opened 2017-06-26
Updated style with new show_frame settings 

jlstevens:
Talking to @philippjfr we decided that only raster types, path types and vectorfields should have frames.


philippjfr:
What about Scatter and Points? They share a plotting class so setting it on the class won't be sufficient to give Points a frame and Scatter none.

jlstevens:
> What about Scatter and Points? 

I think that is the only one that needs to be set with the options system and not at the class level. I'll push that fix now.

jlstevens:
@philippjfr Looks like the tests will go green. Ready for final review/merge.

philippjfr:
Looks good will merge shortly.

PR #1616, opened 2017-06-26
Renamed bokeh HeatmapPlot to HeatMapPlot

jlstevens:
This makes the naming consistent with Matplotlib.


jlstevens:
@philippjfr Tests just passed and this should be merged ASAP as there is currently an undefined name error...

PR #1620, opened 2017-06-27
Improved the first two user guides

jlstevens:
This PR is to get the first two user guides presentable.


jlstevens:
@philippjfr I think this is ready to merge. It will never be perfect but I think it is a big improvement!

philippjfr:
Looks like a good start, merging.

PR #1627, opened 2017-06-27
User guide fix links

jlstevens:
Fixed links in the 5th and 6th user guides.


jlstevens:
@philippjfr Think it can be merged right away.

PR #1632, opened 2017-06-28
Cleaned up user guides and converted to bokeh

philippjfr:
As the title says mostly includes fixes for links in user guides but also converts plots to bokeh and uses gifs to illustrate custom interactivity guide.


doc/user_guide/Custom_Interactivity.rst
```diff
@@ -2,4 +2,5 @@ Custom Interactivity
 ____________________
 
 .. notebook:: holoviews ../../examples/user_guide/12-Custom_Interactivity.ipynb
+    :skip_execute: True
```
jlstevens: Are you sure you don't want *any* output (e.g print output)? You can use the skip_output option if you add a comment line 'When run live, this cell's output should match the behavior of the GIF below' into the code cells that should be skipped.

jlstevens:
Made one comment and maybe it would be nice to clear out kernel specs before merging? I guess they aren't too big and I can strip them out of all the notebooks just before we package...

doc/user_guide/Custom_Interactivity.rst
philippjfr: I've added the print output to the notebook for now. Could you explain ``skip_output`` again? What specific line does it look for again?

jlstevens: > I've added the print output to the notebook for now.

That is fine for now but isn't a long term solution.

```
1+1
# When run live, this cell's output should match the behavior of the GIF below
hv.DynamicMap(....)
```
[GIF in markdown here]

The comment can be anywhere in the cell you want to skip. The rst:

```rst
.. notebook:: holoviews ../../examples/user_guide/12-Custom_Interactivity.ipynb
    :skip_output: When run live, this cell's output should match the behavior of the GIF below
```

See the Live Data rst for the getting started guide...

jlstevens: Don't think you did this...happy to merge but we shouldn't forget about it!

jlstevens:
Looks good. I can add blurbs about the topics later. Merging.

PR #1645, opened 2017-06-29
Allowed using timedeltas in widgets

philippjfr:
Tiny fix to allow timedeltas to appear in the widgets. Tiny and safe fix that I hope can make it into 1.8.


jbednar:
Seems safe to me.

jlstevens:
Seems find to me too. The test failure is just a transient display test issue. Merging.

PR #1646, opened 2017-06-29
Added multi-res favicon

jbednar:
Not sure how to test it without deploying it, but I think this version will be more visible...


jlstevens:
Let's try it!

PR #1647, opened 2017-06-29
Updated extension logo

jlstevens:
Here are some of the ones I tried, currently using the bottom right one:

options

Using a size of 64x64 (used to be 25x25 which I think is too small).


jlstevens:
My favorite ones are on the far right.

jlstevens:
I think I would pick the one on the top right.

jlstevens:
Here is what it looks like right now:

image

jlstevens:
Out final choice:

image


@jbednar @philippjfr Ready to merge.

PR #1652, opened 2017-06-29
Fixed bug in renderer size scaling for bokeh

philippjfr:
Since the plot sizes are defined "inside out" in bokeh (i.e. each component defines its own size) the size scaling should be applied on bokeh's ElementPlot unlike matplotlib where it can be applied on the renderer. Before this fix it was applied twice, which was obviously wrong.


jlstevens:
Will display tests need updating?

philippjfr:
I hope not!

jlstevens:
Tests are now green. Merging.

PR #1653, opened 2017-06-29
Tweaks to user guides

jlstevens:
Main change is the reworked 'customizing plot' user guide.


philippjfr:
Looks good. Like most of the user guide we'll have to edit and revise later.

PR #1658, opened 2017-06-30
Start matplotlib guide

jordansamuels:
* follow intro style of bokeh guide
* copy some info from Exporting guide
* add new animated GIF example

### DIFF
--- a/examples/user_guide/Plotting_with_Matplotlib.ipynb
+++ b/examples/user_guide/Plotting_with_Matplotlib.ipynb
@@ -1,0 +1,188 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Plotting with Matplotlib"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The default plotting backend for HoloViews is based on [Matplotlib](http://matplotlib.org).  There is also a [Bokeh](http://bokeh.pydata.org) backend - see the [Plotting with Bokeh](Plotting_with_Bokeh.ipynb) user guide.\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


jbednar:
Thanks!  Bokeh just added SVG and PNG exporting, which is supported by HoloViews, so maybe it's best if this notebook just points to the relevant bit of Exporting when discussing output like that? Exporting is no longer matplotlib-only.

jbednar:
What I would think would be useful to add to this notebook is a discussion of how the matplotlib backend works, in comparison to the Bokeh or plotly backends, and what the pros and cons are.  Basically, for static output the matplotlib backend renders each the plot (or each plot, for a HoloMap) to a PNG, and then HoloViews generates JavaScript that provides widgets to select between these pre-rendered PNGs.  This process is very different from the Bokeh or Plotly backends, which generate JavaScript/HTML directly that contains specific glyphs corresponding to the data in the plot.

The matplotlib backend approach has the advantage that the output size is largely independent of the amount of data involved, and that the resulting output requires less JavaScript and is overall simpler once rendered, which also makes it more future-proof as a saved plot with widgets.

The bokeh/plotly approach has the advantage that these backends provide interactivity *within* the plot, even in a static HTML/JS output, which opens up a wide range of possible user-visible functionality such as zooming in to see detail, dynamically re-rendering plots on zooming with Datashader, and so on .  At least in principle, for small numbers of datapoints spread across many frames of a HoloMap, the Bokeh/Plotly approach could have much smaller file sizes, but I haven't done benchmarking to see where the crossover point would be.

jbednar:
You could also mention that the matplotlib and plotly backends provide 3D plotting with HoloViews, unlike the Bokeh backend.

philippjfr:
I agree with all your comments @jbednar. That said this seems like a reasonable start and I'd be open to merging it and improving it later.

jlstevens:
I agree with Philipp and I'll merge now. I've recorded Jim's suggestions on the newly created wiki page for this user guide [here](https://github.com/ioam/holoviews/wiki/Plotting-with-Matplotlib).

PR #1663, opened 2017-06-30
Added simpler .hist example

jbednar:
The Histogram page in the Reference Gallery didn't have a simple example of using '.hist()', which is typically much more convenient than the complex example shown.  Added this simple example:

image

Works fine for Bokeh, but in mpl there is a strange colormap:

image

The last example on that page (not really modified in this PR)  is also aligned and sized very strangely for matplotlib, though it works fine for bokeh:

image

### DIFF
--- a/examples/reference/elements/bokeh/Histogram.ipynb
+++ b/examples/reference/elements/bokeh/Histogram.ipynb
@@ -67,7 +67,25 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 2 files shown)


jlstevens:
That is a worthwhile example to add. Best we wait to see which of those matplotlib bugs can be fixed before merging.

philippjfr:
Will see, dealing with adjoints more generally in matplotlib has been a longstanding issue (see https://github.com/ioam/holoviews/issues/1033).

jlstevens:
@philippjfr Should we merge despite the weird matplotlib output?

philippjfr:
I'll see if I can deal with the wrong colormapping at least. Doubt I'll be able to fix the sizing issue easily.

philippjfr:
@jbednar Would you mind updating this PR by setting an explicit range for the x and y dimensions? For now matplotlib won't normalize the points and histogram together so you end up with different scales.

jbednar:
It sounds like you know what's needed; can you just push what you suggest to the PR?

philippjfr:
Done. @jlstevens ready to merge when tests pass.

jlstevens:

Issue #1669, opened 2017-07-02
Spikes get unexpectedly colormapped


jlstevens:
For instance:

![image](https://user-images.githubusercontent.com/890576/27766690-b81e16e6-5ed0-11e7-85f7-67ae3cabeaba.png)

I believe this is a simple parameter change to the color indexing but I would like it fixed in time for the SciPy tutorial...

jlstevens:
Also note that the colormap is viridis, not fire!

jlstevens:
Fixed in the PR referenced above (#1671). Closing.

PR #1671, opened 2017-07-02
Set SpikesPlot color index to None by default

jlstevens:
Before:

</img>

After:

</img>

I think this is a bug fix, not a style change as before a vdim was being mapped to both height *and* color when it should only be mapped to height.


jlstevens:
Test data will need update. Also should mention this PR addresses #1669.

jlstevens:
I'll update the colormap to 'fire' before updating the display data though...

jlstevens:
@philippjfr pr builds are finally green. Ready to merge.

PR #1672, opened 2017-07-03
Renamed Bounds to BoundsXY for consistency

jlstevens:
PR title says everything other than the fact that ``Bounds`` still exists as an alias (with a deprecation warning).


jlstevens:
@philippjfr Ready to merge if you are happy with this PR.

philippjfr:
Looks good, would be good if we could decide on what timeframe deprecations should happen and perhaps leave searchable comments marking deprecations with the release when we expect the feature to be removed. Will just merge this though.

PR #1683, opened 2017-07-06
Choose better magic values for GridSpace plot label offsets

philippjfr:
The GridSpace plot has to appropriately expand the sizes of the axes when the label length changes. This code is never going to be perfectly robust but it appears the magic values I chose initially weren't very good and resulted in errors for even fairly short labels, as seen here:

https://github.com/ioam/holoviews/issues/1675

Tested the new values with GridSpaces of varying size, varying fontsize with tick labels ranging from 1 to 100 characters.

### DIFF
--- a/holoviews/plotting/bokeh/util.py
+++ b/holoviews/plotting/bokeh/util.py
@@ -208,7 +208,7 @@ def font_size_to_pixels(size):
 
 
 def make_axis(axis, size, factors, dim, flip=False, rotation=0,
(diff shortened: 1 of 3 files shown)


jlstevens:
Bit confused. Both pr builds failed but both push builds passed?

philippjfr:
@jlstevens Passing now.

jlstevens:
Looks good. Merging.

PR #1685, opened 2017-07-06
LabelledData.map drops None values returned by map function

philippjfr:
Discussion about why this fix is needed in https://github.com/ioam/holoviews/issues/1681


jlstevens:
Tests have passed and I am happy to merge. That said, if map is discussed anywhere in the user guide atm this particular behavior should be mentioned...

philippjfr:
I think we meant to mention it in the nesting data user guide but haven't done so far.

jlstevens:
Please open a docs issue as a reminder to do that.

PR #1686, opened 2017-07-06
Make NdMapping slice lower bound always be inclusive 

philippjfr:
Fixes a longstanding bug when slicing NdMapping types. When slicing an NdMapping without an upper bound on the slice the lower bound was exclusive for some reason, which is inconsistent with our documentation and what happens when upper bound is supplied.


jlstevens:
If the tests pass, I am happy to merge. This fix definitely deserves an entry in the release notes/changelog!

jlstevens:
Tests are green. Merging.

PR #1690, opened 2017-07-06
User guide 01 and 02 fixups

ceball:
First pass through 01 and 02.

Note that I have changed some example code, and I have potentially changed the meanings of some sentences. But no reorganization.


jlstevens:
Skimming the diffs, this all looks good. I think I'll probably have a look using nbdime diff to make sure all the meanings are still correct.

jlstevens:
Thanks for doing this! Changes look good.

Issue #1679, opened 2017-07-04
log scale fails for Bokeh (works for matplotlib)


wjaskowski:
```python
%%opts Curve [logy=True]
hv.extension('bokeh')
def sine_curve(phase, freq):
    xvals = [0.1* i for i in range(100)]
    return hv.Curve((xvals, [np.sin(phase+freq*x) for x in xvals]))

phases      = [0, 1, 2]
frequencies = [0.5, 1.0, 1.5]
curves = {(p,f):sine_curve(p,f) for p in phases for f in frequencies}
#del curves[0, 1.0]
hmap = hv.HoloMap(curves, kdims=['phase', 'frequency'])
hmap.layout(dimensions=['phase', 'frequency'])
```
produces a set of empty plots (frames only)

philippjfr:
This is because bokeh breaks when you supply values below zero for the range. It's not clear to me what we can do about this since we'd have to make an arbitrary choice for the y-range lower bound. For now you can us ``hmap.layout(dimensions=['phase', 'frequency']).redim.range(y=(0.01, 1))`` or similar to display just positive values.

jlstevens:
Maybe we should display our arbitrary lower bound as a warning?

philippjfr:
The PR above addressed this, closing.

PR #1691, opened 2017-07-07
Handled bokeh log axes with axis lower bound <=0

philippjfr:
Addresses https://github.com/ioam/holoviews/issues/1679. If the range lower bound is <= 0, it uses a lower bound of 0.01 unless the upper bound is <0.01, in which case it backs up to a value one decimal point lower than the upper value.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -557,12 +557,12 @@ class ElementPlot(BokehPlot, GenericElementPlot):
             xfactors, yfactors = self._get_factors(element)
         framewise = self.framewise
         if not self.drawn or (not self.model_changed(x_range) and framewise):
-            self._update_range(x_range, l, r, xfactors, self.invert_xaxis, self._shared['x'])
+            self._update_range(x_range, l, r, xfactors, self.invert_xaxis, self._shared['x'], self.logx)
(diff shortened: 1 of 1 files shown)


jlstevens:
Should it warn? Isn't 0.01 an arbitrary value?

jlstevens:
Nevermind, I see you have a warning in there.

jlstevens:
@philippjfr Ready to merge?

philippjfr:
Yes.

TheoMathurin:
> Addresses #1679. If the range lower bound is <= 0, it uses a lower bound of 0.01 unless the upper bound is <0.01, in which case it backs up to a value one decimal point lower than the upper value.

Looking at the [current file](https://github.com/holoviz/holoviews/blob/0defcef994d6dd6d2054f75a0e332d02d121f8b0/holoviews/plotting/bokeh/element.py#L935), if I'm not mistaken it seems like the lower bound is set to 0.01 _if_ (and not _unless_) the upper bound is <0.01: `low = 0.01 if high < 0.01 else 10**(np.log10(high)-2)`

TheoMathurin:
Should I file an issue on this matter?

If it's just about changing the "lesser than" sign, I can make the corresponding PR as well.

PR #1692, opened 2017-07-07
Fixed link to reference gallery on hv.org.

ceball:


jlstevens:
Thanks. I think this one is worth updating manually to make the fix live.

PR #1697, opened 2017-07-07
Set default datashader cmap to fire

philippjfr:
Simple change, the default defined by datashader isn't great.


jlstevens:
Doubt this is tested right now, happy to merge.

philippjfr:
Yes, only testing aggregate atm, so just merge.

Issue #1684, opened 2017-07-06
Code Example: SRI Immunization Model


ghost:
When running the sri.model.animate(21) function, I get an index error. Any help is appreciated.

<img width="1242" alt="screenshot 2017-07-05 18 53 26" src="https://user-images.githubusercontent.com/5890399/27892228-478cf934-61b3-11e7-94fd-3532bfd197df.png">

jlstevens:
@senayyitbarek Thanks for reporting this!

I've started a PR (referenced above) to make fixes. The main things I had to do was use:

```python
            path_data = [list(el[()]) for el in paths]
            # Create network path and node Elements
            network_paths = hv.Path(path_data, key_dimensions=spatial_dims)
            network_nodes = hv.Points(np.hstack([points, state_array]),
                                      key_dimensions=spatial_dims,
                                      value_dimensions=['State'])
            
            # Create overlay and accumulate in network HoloMap
            network_hmap[i] = (network_paths * network_nodes * nlabel).relabel(group='Network', label='SRI')
            self.step()

        # Create Overlay of Curves
        curves = hv.NdOverlay({label: hv.Curve(zip(range(steps), sird[:, i]),
                                               kdims=['Time'], vdims=['Count'])
```

In the animate loop.

The main changes are: the data format supplied to ``Path`` was invalid and ``key_dimensions`` and ``value_dimensions`` in the ``NdOverlay`` had to be changed to ``kdims`` and ``vdims``.

philippjfr:
This has been fixed a while ago and further discussion about topics notebooks should happen in: https://github.com/ioam/holoviews/issues/1638

PR #1699, opened 2017-07-07
Fixes to the SRI model notebook

jlstevens:
Addresses #1684. The fix that worried me most is that ``key_dimensions`` didn't work when ``kdims`` did.


philippjfr:
I'm just going to merge this. Can do further cleanup in another PR.

PR #1705, opened 2017-07-10
fix typos "{X,Y}Bounds" -> "Bounds{X,Y}"

jordansamuels:


jbednar:
Looks like an obvious and correct fix; not sure why there is a Py3 test error at https://travis-ci.org/ioam/holoviews/jobs/252146034#L1608 .

jordansamuels:
Hmm, now coverage is failing in addition to continuous.  Are we discovering build issues?  In any case, this is super low priority, so we can wait.

philippjfr:
@jordansamuels Annoyingly we have some transient errors in our builds. Usually restarting it will resolve those, which I've just done.

jlstevens:
@jordansamuels Thanks for these fixes!

PR #1711, opened 2017-07-10
Table finalize hook fix and example

jordansamuels:
One solution to https://github.com/ioam/holoviews/issues/1708


jordansamuels:
(My additional commits were in confusion as I forgot it's another PR that's failing the Travis build.)

jlstevens:
Thanks for this!

We started adding little feature examples just like this one in holoviews-contrib and we haven't yet found a new home for them (see [this example](https://github.com/ioam/holoviews-contrib/blob/master/notebooks/examples/arrow_plotting_hook.ipynb) which is very similar).

Discussing this with Philipp, we think this can be the first example in ``/examples/reference/features/bokeh``. Could you please initialize those directories and move your example there?

jordansamuels:
@jlstevens yes, will reorganize and add the example soon.  Thanks.

jlstevens:
@jordansamuels Thanks!

PR looks good merging.

PR #1724, opened 2017-07-14
Define default group on Dataset

philippjfr:
Simple oversight which results in some bad behaviors when inheriting the groups for some operations.


jlstevens:
Looks fine. Merging.

PR #1725, opened 2017-07-14
Added a groupby argument to the histogram operation

philippjfr:
As suggested in https://github.com/ioam/holoviews/issues/1716 we should have an easier way to generate a grouped/faceted histogram. The current approach is as follows:

```
hv.Dataset(autompg).groupby('cyl').hist(dimension='mpg', adjoin=False).overlay()
```

For simplicities sake I suggest adding this:

```
hv.Dataset(autompg).hist(dimension='mpg', groupby='cyl', adjoin=False)
```

OR

```
histogram(hv.Dataset(autompg), dimension='mpg', groupby='cyl')
```

bokeh_plot 90

Going to add gallery demos before this is ready to merge.

### DIFF
--- a/examples/gallery/demos/bokeh/autompg_histogram.ipynb
+++ b/examples/gallery/demos/bokeh/autompg_histogram.ipynb
@@ -1,0 +1,66 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Most examples work across multiple plotting backends, this example is also available for:\n",
+    "\n",
+    "* [Matplotlib - autompg_histogram](../matplotlib/autompg_histogram.ipynb)"
+   ]
+  },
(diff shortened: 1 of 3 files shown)


jlstevens:
Is a nice exception raised if you try to groupby dimensions that are not available/make no sense.

philippjfr:
Just the usual exception raised when a dimension is not found:

>KeyError: u'Dimension cylinder not found'

jlstevens:
Maybe ``groupby`` should be explicitly listed as an argument of ``.hist`` so you can see this argument exists when inspecting the signature (instead of just seeing ``**kwargs``).

philippjfr:
Let's merge this now, you've already opened an issue to make the .hist signatures more consistent.

jlstevens:
Sounds good.

PR #1730, opened 2017-07-15
Fix spelling for strings and comments

jbcrail:


philippjfr:
Thanks so much @jbcrail, restarting the travis build now since there was a transient test failure.

PR #1732, opened 2017-07-15
Ensure dask.array is imported

philippjfr:
Simple fix to ensure dask.array is actually imported, fixing issue described in

https://github.com/ioam/holoviews/issues/1710


jlstevens:
Looks good. I'll merge when the tests are green.

Issue #1738, opened 2017-07-20
NdLayout opts undoes cols


jordansamuels:
It appears that calling `.opts()` on an `NdLayout` will undo the effects of `.cols()`.  E.g.
```
c = hv.Curve(np.arange(3))
hv.NdLayout([(n,c) for n in range(5)]).cols(2).opts(plot=dict(title_format="> 2 columns"))
```

results in

![image](https://user-images.githubusercontent.com/610909/28437096-01b4252e-6d5f-11e7-87ac-e635b10a6131.png)

The does not appear to be a problem with `Layout`.

jlstevens:
I suspect ``NdLayout`` needs a clone method that preserve ``cols``, the way ``Layout`` does it:

```
    def clone(self, *args, **overrides):
        """
        Clone method for Layout matches Dimensioned.clone except the
        display mode is also propagated.
        """
        clone = super(Layout, self).clone(*args, **overrides)
        clone._max_cols = self._max_cols
        clone.id = self.id
        return clone
```

Not sure why the ``id`` isn't propagated by the ``super`` call here but that is separate from what is important for this issue, namely transferring the value of``_max_cols``.

philippjfr:
Now fixed, closing.

PR #1739, opened 2017-07-20
Fixed clone on NdLayout to propagate cols properly

jlstevens:
PR title explains the change. My only worry is why does the ``id`` need to be copied across, doesn't super handle that?

### DIFF
--- a/holoviews/core/layout.py
+++ b/holoviews/core/layout.py
@@ -339,6 +339,18 @@ class NdLayout(UniformNdMapping):
         return self.clone(last_items)
 
 
+    def clone(self, *args, **overrides):
+        """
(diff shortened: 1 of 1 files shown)


jlstevens:
I want to figure out whether the propagation of ``id`` is really needed before merging.

jlstevens:
Note, this PR addresses #1738.

philippjfr:
>I want to figure out whether the propagation of id is really needed before merging.

It definitely is ``Layout.clone`` does the same thing. A clone of an Element should resets any custom options while a Layout should not. I'm happy to merge if you think it's ready.

jlstevens:
>  A clone of an Element should resets any custom options while a Layout should not. 

Ok. That makes sense, in which case this is ready to merge.

Issue #1745, opened 2017-07-22
hv.Table(df).to.image throws warning and error


stsievert:
``` python
hv.Table(df).to.image(kdims=['samples', 'sparsity'], vdims='error')
```

throws an warning "WARNING:root:GridImage: GridImage is now deprecated. Please use Image element instead." and an error with a traceback

``` traceback
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-8b2b98813c82> in <module>()
----> 1 data.to.image(kdims=['samples', 'sparsity'], vdims='error')
/Users/scott/anaconda/lib/python3.6/site-packages/holoviews/element/__init__.py in image(self, kdims, vdims, groupby, **kwargs)
---> 58         return self(GridImage, kdims, vdims, groupby, **kwargs)
/Users/scott/anaconda/lib/python3.6/site-packages/holoviews/core/data/__init__.py in __call__(self, new_type, kdims, vdims, groupby, sort, **kwargs)
--> 134                                  group_type=new_type, **params)
/Users/scott/anaconda/lib/python3.6/site-packages/holoviews/core/data/__init__.py in groupby(self, dimensions, container_type, group_type, dynamic, **kwargs)
--> 561                                       group_type, **kwargs)
/Users/scott/anaconda/lib/python3.6/site-packages/holoviews/core/data/pandas.py in groupby(cls, columns, dimensions, container_type, group_type, **kwargs)
--> 128                 columns.data.groupby(group_by, sort=False)]
/Users/scott/anaconda/lib/python3.6/site-packages/holoviews/core/data/pandas.py in <listcomp>(.0)
--> 127         data = [(k, group_type(v, **group_kwargs)) for k, v in
/Users/scott/anaconda/lib/python3.6/site-packages/holoviews/element/raster.py in __init__(self, *args, **kwargs)
--> 443         super(GridImage, self).__init__(*args, **kwargs)
/Users/scott/anaconda/lib/python3.6/site-packages/holoviews/element/raster.py in __init__(self, data, bounds, extents, xdensity, ydensity, **params)
--> 241         dim2, dim1 = self.interface.shape(self, gridded=True)[:2]
TypeError: shape() got an unexpected keyword argument 'gridded'
```

The below works in the same example above:

``` python
hv.Table(df).to.heatmap(kdims=['samples', 'sparsity'], vdims='error')
```

I'm on holoviews version: 1.8.1.

jlstevens:
[...]

PR #1746, opened 2017-07-23
Switched from deprecated GridImage to Image in conversion interface

jlstevens:
Simple PR to address some of the issues in #1745. Fixes to the image conversion interface should probably go into this PR too.


philippjfr:
Should add a condition to check if the input Dataset is gridded and throw an exception in that case.

jlstevens:
Would that involve checking the ``gridded`` class attribute of the interface? I've not looked how to access that from ``DataConversion`` yet...

philippjfr:
Will just merge now, can add gridded checks in the appropriate places later.

PR #1763, opened 2017-07-27
Add opts in doc 2

jordansamuels:
Cleaner version of https://github.com/ioam/holoviews/pull/1760, which inadvertently picked up https://github.com/ioam/holoviews/pull/1759 during a rebase.


jlstevens:
Having skimmed the diff, all the changes look correct to me. As the tests are green, I think it is better to merge sooner than later. I'm sure there will be other places that ``__call__`` need to be replaced but having this merged will make those other instances easier to find.

@jordansamuels Thanks again for this PR! Merging.

PR #1773, opened 2017-08-01
Added datashader regridding operation

philippjfr:
As suggested and outlined in https://github.com/ioam/holoviews/issues/1552 this PR adds a ``regrid`` operation that allows dynamically downsampling and upsampling HoloViews Image, RGB and HSV types to a specified x_range, y_range, width and height. It closely mirrors the ``aggregate`` operation but operates on Images instead. I introduced a common baseclass for the two operations and have started using it in various projects. It will require some further cleanup, decisions on naming and documentation but is now fully functional.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1,4 +1,4 @@
-from __future__ import absolute_import
+from __future__ import absolute_import, division
 
 from collections import Callable, Iterable
 from distutils.version import LooseVersion
@@ -30,7 +30,10 @@ from ..streams import RangeXY, PlotSize
 from ..plotting.util import fire
 
 
-class resample_operation(Operation):
+class ResamplingOperation(Operation):
+    """
+    Abstract baseclass for resampling operations
+    """
 
     dynamic = param.Boolean(default=True, doc="""
        Enables dynamic processing by default.""")
@@ -123,7 +126,7 @@ class resample_operation(Operation):
 
 
 
-class aggregate(resample_operation):
+class aggregate(ResamplingOperation):
     """
     aggregate implements 2D binning for any valid HoloViews Element
     type using datashader. I.e., this operation turns a HoloViews
@@ -337,7 +340,7 @@ class aggregate(resample_operation):
 
 
 
-class regrid(resample_operation):
+class regrid(ResamplingOperation):
     """
     regrid allows resampling a HoloViews Image type using specified
     up- and downsampling functions defined using the aggregator and
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -56,8 +56,9 @@ def dataset_pipeline(dataset, schema, canvas, glyph, summary):
         agg = dask_pipeline(dataset.data, schema, canvas,
                             glyph, summary)
 
-    agg = agg.rename({'x_axis': kdims[0].name,
-                      'y_axis': kdims[1].name})
+    if 'x_axis' in agg and 'y_axis' in agg:
+        agg = agg.rename({'x_axis': kdims[0].name,
+                          'y_axis': kdims[1].name})
     return agg
 
 
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -20,11 +20,13 @@ from datashape import discover as dsdiscover
 
 from ..core import (Operation, Element, Dimension, NdOverlay,
                     CompositeOverlay, Dataset)
-from ..core.data import PandasInterface, DaskInterface
+from ..core.data import PandasInterface, DaskInterface, XArrayInterface
+from ..core.sheetcoords import BoundingBox
 from ..core.util import get_param_values, basestring
 from ..element import Image, Path, Curve, Contours, RGB
 from ..streams import RangeXY, PlotSize
 from ..plotting.util import fire
+
 
 @dispatch(Element)
 def discover(dataset):
@@ -59,32 +61,8 @@ def dataset_pipeline(dataset, schema, canvas, glyph, summary):
     return agg
 
 
-class aggregate(Operation):
-    """
-    aggregate implements 2D binning for any valid HoloViews Element
-    type using datashader. I.e., this operation turns a HoloViews
-    Element or overlay of Elements into an hv.Image or an overlay of
-    hv.Images by rasterizing it, which provides a fixed-sized
-    representation independent of the original dataset size.
 
-    By default it will simply count the number of values in each bin
-    but other aggregators can be supplied implementing mean, max, min
-    and other reduction operations.
-
-    The bins of the aggregate are defined by the width and height and
-    the x_range and y_range. If x_sampling or y_sampling are supplied
-    the operation will ensure that a bin is no smaller than the minimum
-    sampling distance by reducing the width and height when the zoomed
-    in beyond the minimum sampling distance.
-
-    By default, the PlotSize stream is applied when this operation
-    is used dynamically, which means that the height and width
-    will automatically be set to match the inner dimensions of 
-    the linked plot.
-    """
-
-    aggregator = param.ClassSelector(class_=ds.reductions.Reduction,
-                                     default=ds.count())
+class resample_operation(Operation):
 
     dynamic = param.Boolean(default=True, doc="""
        Enables dynamic processing by default.""")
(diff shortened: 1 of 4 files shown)


philippjfr:
@jlstevens @jbednar I've now finished the code changes to this PR leaving decisions on naming and docstrings. I've included various changes to the ``aggregate`` implementation for compatibility with the most recent changes in datashader. All the changes should be backward compatible. I've also added unit tests for regridding.

The main naming decisions to make are about the ``aggregator`` and ``interpolation`` parameters which correspond to the ``downsample`` and ``upsample`` methods in gridtools respectively.

philippjfr:
@jlstevens Any idea how I can make the raster regridding unit tests dependent on a particular version of datashader? They're failing now because it's still pulling datashader 0.5.

jlstevens:
You could try doing a version check in the test class ``setUp`` method and raising ``SkipTest`` if it isn't a version it should be testing against. Not entirely sure that would work but seems worth trying...

holoviews/operation/datashader.py
```diff
@@ -123,6 +73,77 @@ class aggregate(Operation):
                                         doc="""
         The type of the returned Elements, must be a 2D Dataset type.""")
 
+    link_inputs = param.Boolean(default=True, doc="""
+         By default, the link_inputs parameter is set to True so that
+         when applying shade, backends that support linked streams
+         update RangeXY streams on the inputs of the shade operation.""")
+
```
jbednar: Maybe expand to say why one might want to set it to False?  E.g. if you reuse objects in different cells of a notebook and find them inappropriately linked?

holoviews/operation/datashader.py
```diff
@@ -123,6 +73,77 @@ class aggregate(Operation):
...
+    By default it will simply count the number of values in each bin
+    but other aggregators can be supplied implementing mean, max, min
+    and other reduction operations.
+
+    The bins of the aggregate are defined by the width and height and
+    the x_range and y_range. If x_sampling or y_sampling are supplied
+    the operation will ensure that a bin is no smaller than the minimum
+    sampling distance by reducing the width and height when the zoomed
+    in beyond the minimum sampling distance.
```
jbednar: when zoomed

holoviews/operation/datashader.py
```diff
@@ -319,19 +312,94 @@ def _process(self, element, key=None):
...
+            eldata = agg if ds_version > '0.5.0' else (xs, ys, agg.data)
+            return self.p.element_type(eldata, **params)
         else:
             layers = {}
             for c in agg.coords[column].data:
                 cagg = agg.sel(**{column: c})
-                layers[c] = self.p.element_type((xs, ys, cagg.data), **params)
+                eldata = cagg if ds_version > '0.5.0' else (xs, ys, cagg.data)
+                layers[c] = self.p.element_type(eldata, **params)
             return NdOverlay(layers, kdims=[data.get_dimension(column)])
```
jbednar: Hopefully once there is a new ds release we can remove this bit and simply require ds>=0.6.0.

holoviews/operation/datashader.py
```diff
@@ -319,19 +312,94 @@ def _process(self, element, key=None):
...
+        Aggregation method.
+        """)
+
+    interpolation = param.ObjectSelector(default='nearest',
+        objects=['linear', 'nearest'], doc="""
+        Interpolation method""")
+
+    upsample = param.Boolean(default=True, doc="""
+        Whether to allow upsampling if the source array is smaller than
+        the requested array.""")
```
jbednar: Should this be False by default since interpolation is nearest by default, and thus the result will be approximately the same whether or not upsampling is done?  With linear interpolation the results will be different, but one could say in the docstring for the interpolation parameter that one might want upsample=False for interpolation methods that smooth the data, unlike nearest.

jbednar:
> The main naming decisions to make are about the aggregator and interpolation parameters which correspond to the downsample and upsample methods in gridtools respectively.

I'm fine with those names.

holoviews/operation/datashader.py
```diff
@@ -20,75 +22,23 @@
...
-    aggregator = param.ClassSelector(class_=ds.reductions.Reduction,
-                                     default=ds.count())
+class resample_operation(Operation):
 
     dynamic = param.Boolean(default=True, doc="""
        Enables dynamic processing by default.""")
 
+    expand = param.Boolean(default=True, doc="""
+       Whether the x_range and y_range should be allowed to expand
+       beyond the extent of the data.""")
```
philippjfr: Yes, that is aggregate. I agree that making it ``False`` might be useful as that's usually what you want for visualization purposes, ``True`` is mostly needed when you want to match grid sizes for computation.

jbednar: Ok, False then; HV should be optimized for viz, I think.

philippjfr: Actually there's a good reason not to enable it for ``aggregate``. Computing the actual ``range`` is expensive when you have to iterate over all the datapoints and it doesn't actually gain you anything except making sure that pixels aren't wasted on regions where there is no data. I could set it to ``True`` for regrid only, but it might be better to be consistent.

holoviews/operation/datashader.py
philippjfr: Will do.

holoviews/operation/datashader.py
philippjfr: Yes, I'd agree with that, upsampling is rarely needed I think, again only when you're trying to match a higher resolution target gridding (in which case you should probably downsample the higher resolution grid rather than upsampling the lower resolution one).

holoviews/operation/datashader.py
jbednar: I think it should be True for regrid only; gridded data has very clear bounds and it's probably more confusing if we don't respect those, and definitely slower.

philippjfr: >and definitely slower.

Not necessarily true, it just changes whether the pixels are crammed into the bounds of the image or whether they're wasted on the area beyond the original image bounds.

holoviews/operation/datashader.py
```diff
@@ -133,7 +139,7 @@ class aggregate(resample_operation):
     the x_range and y_range. If x_sampling or y_sampling are supplied
     the operation will ensure that a bin is no smaller than the minimum
     sampling distance by reducing the width and height when the zoomed
-    in beyond the minimum sampling distance.
+    beyond the minimum sampling distance.
```
jbednar: when zoomed in beyond

jbednar:
Ready to merge?

jlstevens:
I'll be able to review this PR shortly. Don't merge till then!

holoviews/operation/datashader.py
```diff
@@ -20,75 +22,29 @@
...
-
-    By default, the PlotSize stream is applied when this operation
-    is used dynamically, which means that the height and width
-    will automatically be set to match the inner dimensions of 
-    the linked plot.
-    """
-
-    aggregator = param.ClassSelector(class_=ds.reductions.Reduction,
-                                     default=ds.count())
+class resample_operation(Operation):
```
jlstevens: Why is it called ``resample_operation`` and not just ``resample``?

holoviews/operation/datashader.py
jlstevens: Agree with Jim's suggestion here. Doesn't look like it has been updated yet...

holoviews/operation/datashader.py
```diff
@@ -123,6 +79,77 @@ class aggregate(Operation):
...
+            width, height = self.p.width, self.p.height
+        (xstart, xend), (ystart, yend) = x_range, y_range
+
+        # Compute highest allowed sampling density
+        xspan = xend - xstart
+        yspan = yend - ystart
+        if self.p.x_sampling:
+            width = int(min([(xspan/self.p.x_sampling), width]))
+        if self.p.y_sampling:
+            height = int(min([(yspan/self.p.y_sampling), height]))
```
jlstevens: I think you need to be careful here - I don't see why ``xspan`` and ``self.p.x_sampling`` can't both be integers on Python 2 causing potential integer division issues.

holoviews/operation/datashader.py
jlstevens: Is this an abstract base class or a usable one?

philippjfr: Abstract, doesn't have a ``_process``

holoviews/operation/datashader.py
philippjfr: Sure will add the ``__future__`` import.

holoviews/operation/datashader.py
```diff
@@ -319,19 +318,109 @@ def _process(self, element, key=None):
...
+                coords['layer'] = range(len(arrays))
+                dims = ['layer'] + dims
+            else:
+                array = arrays[0]
+            xarr = xr.DataArray(array, coords=coord_dict, dims=dims)
+
+        # Disable upsampling if requested
+        (xstart, xend), (ystart, yend) = (x_range, y_range)
+        xspan, yspan = (xend-xstart), (yend-ystart)
+        if not self.p.upsample and self.p.target is None:
```
jlstevens: Shouldn't there at least be a warning to say upsampling has been disabled?

holoviews/operation/datashader.py
jlstevens: My issue here is that the name does not sound like an abstract class (assuming it is). I think it can have a simple name and should simply be hidden from imports.

holoviews/operation/datashader.py
philippjfr: Huh why? It's a parameter.

holoviews/operation/datashader.py
philippjfr: What's your suggestion?

jlstevens:
Looks good other than a few minor comments I made. Happy to see this PR merged once those things are addressed.

holoviews/operation/datashader.py
jbednar: ResampleOperation?

holoviews/operation/datashader.py
jlstevens: I think we settled for ``ResamplingOperation``.

holoviews/operation/datashader.py
jbednar: Seems like the parameter should be respected as set, with no warning.

holoviews/operation/datashader.py
jbednar: Looks like it's been updated below; maybe it was supposed to be done here?

holoviews/operation/datashader.py
jlstevens: In that case the docstring needs improvement:

"Whether to allow upsampling if the source array is smaller than the requested array."

What is the behavior if ``upsample=False``? Padding?

holoviews/operation/datashader.py
```diff
@@ -20,75 +22,29 @@
...
+       beyond the extent of the data.  Setting this value to True is
+       useful for the case where you want to ensure a certain size of
+       output grid, e.g. if you are doing masking or other arithmetic
+       on the grids.  A value of False ensures that the grid is only
+       just as large as it needs to be to contain the data, which will
+       be faster and use less memory if the resulting aggregate is
+       being overlaid on a much larger background.""")
+
     height = param.Integer(default=400, doc="""
        The height of the aggregated image in pixels.""")
```
jlstevens: Isn't this parameter used by ``regrid``? I'm not sure it is clear to talk about an 'aggregated' image in that case. Maybe just talk about the 'output' image?

holoviews/operation/datashader.py
jlstevens: Ok, Philipp's now explained it to me. The behavior makes sense though I think there could be a bit more clarification in the docstrings.

holoviews/operation/datashader.py
philippjfr: True, although in aggregation is still the correct term at least when downsampling. Will fix it anyway.

holoviews/operation/datashader.py
philippjfr: Yes, meant to add it everywhere.

jlstevens:
Looks good and the unit tests have passed. Merging.

jakirkham:
So GitHub makes it look like this is in `v1.8.3` and `v1.8.4`, but went to `import` it and it wasn't. Then I realized it had been removed from those releases ( https://github.com/ioam/holoviews/pull/1884 ). Maybe GitHub is confused. Anyways when is this planned to land in a release?

philippjfr:
>Anyways when is this planned to land in a release?

Today :-)

jakirkham:
Well isn't it my lucky day. 😄

Issue #1769, opened 2017-07-31
format changed when converting into HTML


Jerryzhuht:
Whenever I use `hv.extension('bokeh')` in Jupyter notebook and downloaded as HTML, there is a huge gap under the cell with that code, and the content blow are no longer inside the box.

In notebook:
![1](https://user-images.githubusercontent.com/30041733/28769476-43252c3a-760e-11e7-8634-a2dee54431e3.png)

When use: File -- Download as HTML:
![2](https://user-images.githubusercontent.com/30041733/28769498-6160ad8c-760e-11e7-9b15-457c7dd4c7f2.png)

jlstevens:
Oh wow, I just reproduced this and thank you for reporting this problem!

This must be down to the additional logos added in 1.8.1 and probably down to how the divs are structured there (which looks fine in normal notebooks but is wrong after nbconvert). This is a priority fix for 1.8.2.

Jerryzhuht:
Yeah.. the divs messed the HTML all up..

philippjfr:
Note the divs are also messing up the rendering of notebooks exported to anaconda.org, e.g. have a look [at this notebook](https://anaconda.org/philippjfr/graphs/notebook). Unlike older notebooks it scrolls.

Jerryzhuht:
BTW, are we going to release the new patch recently, like this month?

jlstevens:
> Unlike older notebooks it scrolls.

I noticed that but didn't make the link to the logo.

> BTW, are we going to release the new patch recently, like this month?

Yes, this is a good reason to have 1.8.2 very soon, I estimate we can release that early next week.

Jerryzhuht:
> Yes, this is a good reason to have 1.8.2 very soon, I estimate we can release that early next week.
That would be great +1:

jlstevens:
The fix for this issue has now been merged. Closing.

PR #1778, opened 2017-08-02
HTML fix in logo for export

jlstevens:
PR to address #1769. The current commit (removing inner divs) fixes the export. The problem is that the logos now look like this:

image

As opposed to the previous arrangement that was:

image

With all three logos, it now looks like:

image

Which I don't quite like as much but either way, fixing the export is far more important than the arrangement of the logos!

### DIFF
--- a/holoviews/ipython/load_notebook.html
+++ b/holoviews/ipython/load_notebook.html
@@ -4,7 +4,6 @@
 
 {% if logo %}
 <div>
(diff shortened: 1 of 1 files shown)


jbednar:
I like the new one much better!

philippjfr:
I'm happy with it too, I'll go ahead and merge.

Issue #1734, opened 2017-07-17
Holoviews does not install reliably with pip


jordansamuels:
It appears that holoviews does not properly install with pip in a clean environment, and that
the last release that installed properly is 1.7.0 .  See further down on clean reproduction.

I've tested this on Centos 7 / Python 3.6.0.  Installing 1.8.0 or 1.8.1 with or without `[all]`
results in an error message about params which is similar to the one supplied in a comment to
https://github.com/ioam/holoviews/issues/1659 . Sometimes the error is about numpy instead of params.

These commands fail:
* `pip install holoviews`
* `pip install holoviews[all]`
* `pip install holoviews==1.8.0`
* `pip install holoviews[all]==1.8.0`

These commands succeed:
* `pip install holoviews==1.7.0`
* `pip install holoviews[all]==1.7.0`

## Clean reproduction
Because issue is difficult to test and reproduce reliably, I have reproduced it using
[Docker](http://www.docker.com). I get the same behavior whether I'm using conda or pyenv.
Steps to reproduce cleanly:

1. [Install Docker](https://docs.docker.com/engine/installation/)
2. Use this [Dockerfile](https://gist.github.com/jordansamuels/f8ec860c7b49034e9be338cb20c58337) for conda, or this [Dockerfile](https://gist.github.com/jordansamuels/27ef6440a7df0256b0f6c437db531a79) for pyenv.  Put them in a `conda` or `pyenv` directory and `cd` to it.
3. `docker build -t repro .` 
4. `docker run -t repro pip install holoviews` or another similar command from above.

jordansamuels:
Sample failure:
```
% docker run -ti hv-issue-pyenv  pip install 'holoviews[all]'
Collecting holoviews[all]
  Downloading holoviews-1.8.1.tar.gz (3.1MB)
    100% |################################| 3.1MB 64kB/s
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-f6dq5rx3/holoviews/setup.py", line 131, in <module>
        package_assets(example_path)
      File "/tmp/pip-build-f6dq5rx3/holoviews/setup.py", line 117, in package_assets
        import holoviews
      File "/tmp/pip-build-f6dq5rx3/holoviews/holoviews/__init__.py", line 5, in <module>
        import numpy as np # noqa (API import)
    ModuleNotFoundError: No module named 'numpy'
```

[...]

Issue #1659, opened 2017-06-30
holoviews --install-examples doesn't work with pip


jlstevens:
The examples directory is in the archive directory uploaded to pypi but doesn't seem to install anywhere when pip installing.

jlstevens:
This is a closely related issue to #1643.

jlstevens:
Another pypi/pip issue: numpy and param aren't being fetched automatically for some reason...

jlstevens:
At least not if you do ``pip install holoviews``. At least ``pip install holoviews[all]`` (the recommended thing) works correctly.

ipfans:
v1.8.1 also not working with pip install:

```
$ pip install holoviews
Collecting holoviews
  Using cached holoviews-1.8.1.tar.gz
    Complete output from command python setup.py egg_info:
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "/private/var/folders/g9/m3cmg0m10cl80xt362wpsld00000gn/T/pip-build-7crrhg0l/holoviews/setup.py", line 131, in <module>
        package_assets(example_path)
      File "/private/var/folders/g9/m3cmg0m10cl80xt362wpsld00000gn/T/pip-build-7crrhg0l/holoviews/setup.py", line 117, in package_assets
        import holoviews
      File "/private/var/folders/g9/m3cmg0m10cl80xt362wpsld00000gn/T/pip-build-7crrhg0l/holoviews/holoviews/__init__.py", line 10, in <module>
        import param
    ModuleNotFoundError: No module named 'param
```

I think ```pip install holoviews[all]``` should be clarify in the docs.

jlstevens:
Right now``holoviews[all]`` or ``holoviews[recommended]`` is the best thing for pip installs and it is mentioned in the docs [here](http://holoviews.org/user_guide/Installing_and_Configuring.html). 

The pip installer will correctly pull in param and numpy in 1.8.2 so you won't see the stated error anymore though ``pip install holoviews`` still won't be pulling in bokeh/matplotlib.

ipfans:
sorry，I missed that informatiin. Thx for your reply.

tdsmith:
Adding extras like `[all]` should not actually help; if it appeared to help, it may have been because you had successfully installed holoview using pip in the past, and pip had cached the built holoview wheel. Installing from a cached wheel does not run setup.py and so does not produce the bug.

tdsmith:
[...]

PR #1782, opened 2017-08-03
Fixes to setup.py for pip install

jlstevens:
PR to try and address #1734 and #1659.

### DIFF
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -14,29 +14,6 @@ from .settings import OutputSettings, list_formats, list_backends
 
 Store.output_settings = OutputSettings
 
-
-def examples(path='holoviews-examples', verbose=False, force=False, root=__file__):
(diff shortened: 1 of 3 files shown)


jlstevens:
Testing in a clean env using:

```
pip install https://github.com/ioam/holoviews/archive/setup_fixes.zip
```

Suggests that  #1734 is fixed (no more initial numpy error). I'm not yet sure what to do about #1659 though...

jlstevens:
It looks like the issue with the ``examples`` function was the cause of the problems in #1659 as well. @philippjfr If you believe it is fixed now, I think it is ready to review/merge.

philippjfr:
Thanks for fixing this, I know it was a pain.

jlstevens:
I'm glad duplicating the ``example`` function was sufficient! I was expecting it to get a lot worse!

PR #1792, opened 2017-08-06
Periodic utility now stops when StopIteration is raised

jlstevens:
What is hopefully a simple fix to stop the periodic utility automatically when ``StopIteration`` is raised e.g when using a generator in a ``DynamicMap``.

I understand why you might want to skip ``KeyError`` in ``get_plot_frame`` but I don't get why it is blocking all exceptions and simply printing them. For now I am just catching ``StopIteration`` specifically and re-raising those exceptions. 

@philippjfr Why can't exceptions other than ``KeyError`` be left alone and allowed to bubble up?

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -164,7 +164,10 @@ class periodic(Thread):
             else:
                 self._completed.wait(self.period)
             self.counter += 1
-            self.callback(self.counter)
+            try:
+                self.callback(self.counter)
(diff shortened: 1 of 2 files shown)


jlstevens:
The tests pass but that might not tell us much...

philippjfr:
This seems safe to me, I'd also like to investigate removing that catch-all exception handler, but not in this PR.

PR #1798, opened 2017-08-09
Added support for live streaming with a DataStream

philippjfr:
Initial prototype for a special stream type that allows live streaming of data. Mostly opened for discussion.

A simple example looks like this:

curve_stream


philippjfr:
Closing this, I think we should instead build on the streaming dataframe provided by [streamz](https://github.com/mrocklin/streamz).

PR #1804, opened 2017-08-10
Replaced deprecated VBox and HBox imports

philippjfr:
VBox and HBox were deprecated a while back and have been removed in 0.12.7dev so I've just replaced them Row/Column which is what they were aliasing.


jlstevens:
Looks good. Ready to merge?

philippjfr:
Yes, ready.

PR #1806, opened 2017-08-14
Correctly register empty stream sources

philippjfr:
Fixes small bug registering stream source for empty element.


jlstevens:
Looks good. Ready to merge?

Issue #1797, opened 2017-08-08
HTML format


Jerryzhuht:
Hi. Under the new release, there's still a minor problem of the HTML formatting. When rendering from Jupyter to HTML, there's an extra ending tag `</div>` in the tag of container which contains `hv.extension('Bokeh')`. The outcome would be the rest of content cannot be inside the container.
 Just a minor one.
![capture](https://user-images.githubusercontent.com/30041733/29066549-174dac0e-7c62-11e7-9384-f1007956a468.PNG)

jlstevens:
Are you using version 1.8.2 which was released Friday last week?

Here is what I am currently seeing:

![image](https://user-images.githubusercontent.com/890576/29074448-3a232d50-7c47-11e7-8994-25a52acc8187.png)

I recently removed the divs to fix HTML export (see #1769) so I think if you upgrade using:

```
conda install -c conda-forge holoviews
```

it should work properly. Note, you can also pip install if you wish.

Jerryzhuht:
@jlstevens Thanks for your reply.
I already updated to the latest 1.8.2. It looks nice in Jupyter notebook, but when using File -- Download as --HTML(.html), the problems shows up in the html file(text outside container).
I believe using nbconvert gives the same outcome.

jlstevens:
Nightmare! I thought I had fixed this. :-(

jlstevens:
Ah, I just verified this is only an issue when ``logo=False`` so it isn't as bad as I first thought! Thanks for reporting this and it shouldn't be too hard to fix in the end...

Jerryzhuht:
Exactly! Thanks.

philippjfr:
This was fixed and has now made it into the latest release. Closing.

PR #1809, opened 2017-08-18
Fixed jinja templating when logo=False

jlstevens:
Minor fix to address #1797.

One annoyance I don't know how to fix yet is the vertical whitespace above the logo. Here are the divs responsible but I don't know what is generating them:

image

@philippjfr If you don't have any ideas how to fix that whitespace, this PR is ready to merge.


philippjfr:
The whitespace comes from loading bokehJS in the extension so I don't think we can easily avoid it.  Will merge once the PR build passes.

philippjfr:
For reference, you can reproduce the whitespace with:

```
from bokeh.io import load_notebook, curdoc, show as bkshow
load_notebook(hide_banner=True)
```

Not sure there's anything bokeh can do about it though.

PR #1814, opened 2017-08-21
Define bokeh slider start values

philippjfr:
In bokeh 0.12.6 the slider start value would simply default to the supplied value, in bokeh 0.12.7 this causes an error so this fix is required for compatibility.


jlstevens:
Looks good. Merging.

PR #1829, opened 2017-08-28
Graph support

philippjfr:
Based on the discussion @jlstevens and I had during SciPy sprints this PR adds a Graph Element to HoloViews. The outline below represents the discussion surrounding the API we had:

# Terminology

Graph: Element that *may* or may not visualize itself.
Nodes: Element associating indices to positions and other information. Self-visualizing.
NodePaths: dataframe of connectivity information. Self-visualizing.

### Graph - Composite Element

- Can be abstract
- .data represents the abstract graph (i.e. start and end pairs of node indices)
- Can be visualized by a bokeh layout algorithm as a plot option.
- Can be visualized with straight line connections if given Nodes.
- Can be visualized with arbitrary paths if given NodePaths.

*data*: start and end columns (where start and end are the node indices)
*kdims*: [start, end]
*vdims*: [edge label and other metadata]

### Nodes

Defines concrete position of each node using three columns/kdims, x/y position plus node index/label. Any value dimensions will be exposed as hover labels.

*data*: x,y positions + node index + anything else
*kdims*: [x, y, index]
*vdims*: [anything else (used as hover labels)]

### NodePaths

Subclass of ``Path`` using multi-interface representing the concrete paths

*data*: [ dfs/array, ... ]
*kdims*: [x, y]
df/array: [xs,ys] Defines the path for each edge defined in abstract Graph

## Graph Constructors

Graph(edges)  - Abstract graph (cannot be displayed).
Graph((edges, nodes)) - Defines concrete node positions 
Graph((edges, nodes, nodepaths)) - Define concrete node positions and concrete paths for edges connecting paths

## Graph attributes

* .data - exposes just the abstract graph (i.e. start and end connections)
* .nodes - exposes concrete nodes
* .nodepaths - Computed automatically from .nodes (direct connections) or returns explicitly supplied paths.

# ToDo

* [x] Add matplotlib plotting code
* [x] Add user guide
* [x] Provide control over inspect and hover policy in bokeh
* [x] Add better support for hover information
* [x] Add gallery examples
* [x] Add tests


jbednar:
Looks like a great plan to me.

jlstevens:
@philippjfr Ping me when you think it is in a good state for me to test it out.

philippjfr:
@jbednar @jlstevens This is ready for a review round. You should probably also play with it a bit so you get a feel for it. I'll mostly be focusing on tests at this point.

examples/user_guide/Network_Graphs.ipynb
```diff
@@ -0,0 +1,335 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Hover and selection policies\n",
+    "\n",
+    "Thanks to Bokeh we can reveal more about the graph by hovering over the nodes and edges. The ``Graph`` element provides an ``inspection_policy`` and a ``selection_policy``, which define whether hovering and selection highlight edges associated with the selected node or nodes associated with the selected edge, these policies can be toggled by setting the policy to ``'nodes'`` and ``'edges'``."
+   ]
```
jbednar: Why do we have to choose?  I can see why someone might want to restrict to one or the other, but can't we have the default be that if hit testing says i'm over a node, do the node hovering, and if not do the edge hovering?  If that's possible, then people will almost never need to mess with this option; it will just usually do what they want.

examples/user_guide/Network_Graphs.ipynb
```diff
@@ -0,0 +1,335 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%opts Graph [tools=['hover']]\n",
+    "bezier_graph.opts(plot=dict(inspection_policy='edges'))"
+   ]
```
jbednar: Can we have hover info about the edge? I don't see an example of that.

examples/user_guide/Network_Graphs.ipynb
```diff
@@ -0,0 +1,335 @@
...
+    "G = nx.karate_club_graph()\n",
+    "hv.Graph.from_networkx(G, nx.layout.circular_layout).redim.range(**padding)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Animating graphs"
+   ]
```
jbednar: Animating graphs sounds like a great thing to include, but I don't see anything in this section about that.  Can you add something that actually does animate a graph, e.g. while dragging a slider?

examples/user_guide/Network_Graphs.ipynb
```diff
@@ -0,0 +1,335 @@
...
+   "outputs": [],
+   "source": [
+    "%%opts Graph [color_index='circle']\n",
+    "%%opts Graph (node_size=10 edge_line_width=1)\n",
+    "colors = ['#000000']+hv.Cycle('Category20').values\n",
+    "edges_df = pd.read_csv('../assets/fb_edges.csv')\n",
+    "fb_nodes = hv.Nodes(pd.read_csv('../assets/fb_nodes.csv')).sort()\n",
+    "fb_graph = hv.Graph((edges_df, fb_nodes), label='Facebook Circles')\n",
+    "fb_graph = fb_graph.redim.range(x=(-0.05, 1.05), y=(-0.05, 1.05)).opts(style=dict(cmap=colors))\n",
+    "fb_graph"
```
jbednar: Here the hover information seems like it should be "circle: 50", not "circle: circle50".  I know that means the original data needs updating, but I do think that's important to avoid being confusing.

jbednar: Or it could be something like "group: circle50"...

jbednar: Also, seems like the hover should have something (even if we have to make it up) to indicate the specific identity of the node.  I imagine these are anonymized, but presumably we need something like "identity: person2090" to show that we can use this information to learn not just what category it's in, but who it is (the node identity).

examples/user_guide/Network_Graphs.ipynb
```diff
@@ -0,0 +1,335 @@
...
+   ]
+  },
```
jbednar: I think we need to work on the visibility of the hovering.  Right now, it's clear that nodes and edges are being selected, because they change color, but as soon as the mouse stops moving it is very difficult to see the difference between selected and unselected nodes and especially edges:
image

image

Here the selected node has a slightly different outline, which is very difficult to see, and the selected edges turn red, which is only visible because these happen to be on white; when they are in a complex nest it's super hard to make them out.

Can we make all the non-selected nodes and edges turn dim (Hunt's suggestion)?  Or make the selected ones be much thicker than the rest, so that they actually stand out?

examples/user_guide/Network_Graphs.ipynb
```diff
@@ -0,0 +1,335 @@
...
+    "bundled = bundle_graph(fb_graph)\n",
+    "bundled"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Datashading graphs"
+   ]
```
jbednar: This plot isn't yet compelling.  For one thing, the colorized edges don't really show up, which is partly because colors of fine lines are just never very visible, and partly because they are using the fire colormap, which is an inappropriate choice against a white background.  Here the strongest bundles are colored yellow, which is very hard to see against white, thereby deemphasizing precisely the bits that should be most emphasized.  I really think we need to get rid of the default "fire" colormap for datashader in HoloViews; it's almost never the right choice for the default white backgrounds.  

In any case, this example isn't yet very compelling, perhaps because the data volume isn't high enough to need datashader in the first place.  Maybe show the unbundled graph too, where the color shading can help make sense of the tangled thicket around the main clump of nodes?

examples/user_guide/Network_Graphs.ipynb
```diff
@@ -0,0 +1,335 @@
...
+   "metadata": {},
+   "source": [
+    "### Applying selections"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Alternatively we can select the nodes and edges by an attribute that resides on either. In this case we will select the nodes and edges for a particular circle and then overlay just the selected part of the graph on the datashaded plot. In this way a smaller subgraph can be highlighted and the larger graph can be datashaded."
```
jbednar: This description sounds great, but I really couldn't make sense of the example; it's super hard to see what's datashaded and what isn't and what to make of it.  Again I think that's probably because this dataset isn't sufficiently large for there to be a difference between what bokeh alone can do and what datashader can.  Whatever it is, something needs to change for this to be a useful example.

examples/user_guide/Network_Graphs.ipynb
philippjfr: That's an issue to raise with bokeh.

examples/user_guide/Network_Graphs.ipynb
philippjfr: Not currently, another issue to raise with bokeh.

examples/user_guide/Network_Graphs.ipynb
philippjfr: If you pulled the latest version you should have the node index as well at least, is that sufficient?

examples/user_guide/Network_Graphs.ipynb
philippjfr: I'll look into it.

examples/user_guide/Network_Graphs.ipynb
jbednar: Ok, just pulled again and got the index info.  Yes, that's helpful, but in general people won't *only* want an index, they'll want some sort of human-meaningful information.  But that's presumably available if it's in a column, so we can just say that in words ("Hovering over the node will reveal the index of that node, along with any other columns of information about that node that are available")?

BTW, the index info isn't always coming out ok:

image

examples/user_guide/Network_Graphs.ipynb
jbednar: And these are issues you haven't already raised?

examples/user_guide/Network_Graphs.ipynb
philippjfr: That's pretty damn weird.

examples/user_guide/Network_Graphs.ipynb
philippjfr: Just checked, this is another thing to file with bokeh, the nonselection style options are not exposed at the moment.

examples/user_guide/Network_Graphs.ipynb
philippjfr: I swear there was an issue about this in bokeh but I can't find it now.

jbednar: I can raise one, but it's probably better if you do it, because you're down there using the bokeh implementation and can be very precise about what's there now and what you'd like to see.

examples/user_guide/Network_Graphs.ipynb
philippjfr: No, we still don't have a way of efficiently storing edge information in HoloViews (although that is work I'll be doing for 1.9), so I haven't gotten around to this yet and won't in this PR.

jbednar: Ok. Please raise a separate issue for that and reference it in this PR, but then leave it open.

examples/user_guide/Network_Graphs.ipynb
```diff
@@ -0,0 +1,335 @@
...
+    "\n",
+    "The ``Graph`` ``Element`` differs from other elements in HoloViews in that it consists of multiple sub-elements. The data of the ``Graph`` element itself are the abstract edges between the nodes, on its own this abstract graph cannot be visualized. In order to visualize it we need to give each node in the ``Graph`` a concrete ``x`` and ``y`` position in form of the ``Nodes``. The abstract edges and concrete node positions are sufficient to render the ``Graph`` by drawing straight-line edges between the nodes. In order to supply explicit edge paths we can also declare ``EdgePaths``, providing explicit coordinates for each edge to follow.\n",
+    "\n",
+    "To summarize a ``Graph`` consists of three different components:\n",
+    "\n",
+    "* The ``Graph`` itself holds the abstract edges stored as a table of node indices.\n",
+    "* The ``Nodes`` hold the concrete ``x`` and ``y`` positions of each node along with a node ``index``. The ``Nodes`` may also define any number of value dimensions, which can be revealed when hovering over the nodes or to color the nodes by.\n",
```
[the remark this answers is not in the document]
philippjfr: Thanks for the reminder, I still have to do some work around this. You will be able to pass in node information without positions in the constructor and it will add the positions automatically. Constructing an abstract ``Nodes`` object without positions doesn't make sense to me though.

examples/user_guide/Network_Graphs.ipynb
philippjfr: Sure, I can add a different example.

examples/user_guide/Network_Graphs.ipynb
jbednar: The ??? isn't the only odd bit, it's odd that it's trying to show *all* the node ids at all.  Maybe that's useful sometimes, but I'd guess probably not, since in general that can be a huge list and it's unclear what any of those other indexes would be (and unclear which is the one node I'm hovering over).  Or maybe those are edge indexes.  Confusing!

examples/user_guide/Network_Graphs.ipynb
jbednar: Ok, please do file those issues, and you can tag me if you need someone other than you to express a user-level opinion.  But none of those things should hold up this PR.

examples/user_guide/Network_Graphs.ipynb
philippjfr: Since this is a user guide I don't think a larger dataset is workable. I agree about changing the color scheme, but do you think a larger dataset is required? I'd like to avoid switching to synthetic data.

examples/user_guide/Network_Graphs.ipynb
philippjfr: I can understand that visually it's not hugely compelling and a change in color schemes should help here, but is it really not clear conceptually? It's simply datashading everything and then overlaying part of the graph based on some criterion.

jbednar: The concept is clear to me, but it's not clear visually -- I can't tell which bits are overlaid and which bits are datashaded.  If I zoom quickly I can see that both are present, but I really can't visually get any idea which is which.

examples/user_guide/Network_Graphs.ipynb
jbednar: Well, try the non-bundled case; the effect of datashader's coloring should be clear there.

examples/user_guide/Network_Graphs.ipynb
jbednar: This one also could use the padding that helps the ones above.  I guess auto-padding is only for 2.0?

examples/user_guide/Network_Graphs.ipynb
philippjfr: I was wrong, can probably get this working.

philippjfr: I realize now what the issue was, there is no way of styling something that's not being hovered over. I can only customize non-selected nodes/edges, e.g. when using the tap or box_select tools.

philippjfr: Here's what that looks like:

bokeh_plot

examples/user_guide/Network_Graphs.ipynb
philippjfr: >Or maybe those are edge indexes.

I suspect that's it, in which case I can definitely fix it.

examples/user_guide/Network_Graphs.ipynb
philippjfr: The other issue is that bundling shifts the start and end points of the edges slightly (and sometimes significantly) so they don't line up with the nodes anymore. I'd expect bundling to leave the start and end points unaffected.

jlstevens:
> Also, there's no example of mapping edge "weight" onto anything visible ...

I agree, I think mapping a weight to an edge thickness/color is a very common case.

examples/user_guide/Network_Graphs.ipynb
jbednar: @jbcrail ? I would also have expected bundling to leave nodes *exactly* where they are.

examples/user_guide/Network_Graphs.ipynb
jbednar: Nice!

jbednar:
BTW, is there any way to turn off the x,y locations of graph nodes by default for hovering? The location is usually not informative at all, given that the layout is from some algorithm instead of the data.

philippjfr:
>I agree, I think mapping a weight to an edge thickness/color is a very common case.

Yes, I agree, not implemented yet. Any suggestion on plot options to control this? Ideally we'd have addressed that issue ``size/color/alpha_index`` options by now and there was a general way to do this mapping. I have some further ideas about that, which I'll have to write up. Not sure where that leaves us in the meantime. 

>The location is usually not informative at all, given that the layout is from some algorithm instead of the data.

I'll implement that.

jlstevens:
One thing that I think is missing from this PR are element notebooks documenting ``Graph``, ``Nodes`` and ``EdgePaths`` with the latter two obviously linking to ``Graph``. One goal of this design was to have each element be meaningful with a visual representation in isolation and the element notebooks would be a good way to show that.

philippjfr:
Yes, element notebooks and gallery examples are still missing.

jlstevens:
There is a fair bit of documentation work that needs doing. For this reason, I wouldn't mind merging this PR to get the code in place with a separate PR improving the user guide in the ways Jim has suggested as well as adding the various other notebooks.

philippjfr:
I'd prefer to just do it in this PR tbh. Also have to add tests still.

jlstevens:
I agree we want tests! At least enough to turn coveralls green again. :-)

examples/user_guide/Network_Graphs.ipynb
philippjfr: Unfortunately that's probably the case.

examples/user_guide/Network_Graphs.ipynb
philippjfr: Nevermind, the ``inspection_policy`` determines whether you can inspect nodes or edges and you can just add the edge information to the abstract graph edges, so this works now.

examples/user_guide/Network_Graphs.ipynb
philippjfr: Fixed.

examples/user_guide/Network_Graphs.ipynb
philippjfr: It just looks very messy unfortunately, I've switched to 'linear' normalization, which means that the paths are mostly black. If we can get the nodes to line up I think this is reasonable.

examples/user_guide/Network_Graphs.ipynb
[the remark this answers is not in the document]
philippjfr: Done and example added.

examples/user_guide/Network_Graphs.ipynb
jbcrail: @jbednar I would expect that also. A quick look at the code indicates at least two areas where the start and end points are resampled and modified.

@philippjfr Do you have an example? If so, could you put that into a datashader issue?

philippjfr: @jbcrail Yes, I'll work up an example using the datashader API.

philippjfr:
@jbednar @jlstevens I've now added Element reference and gallery examples for graphs, both are very similar to sections of the user guide and the element notebooks link to the user guide for further detail.

examples/user_guide/Network_Graphs.ipynb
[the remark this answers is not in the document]
jbednar: Ok, sounds good.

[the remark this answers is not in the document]
jbednar: Just to clarify, is assigning positions to the nodes lazy, in that no positions will be assigned if no one ever tries to visualize it?  If so, that sounds good.

jbednar:
Ok, looks great!  Can you please look over our comments above, pull out anything that you haven't been able to deal with yet into an open issue, and then we can merge this?

examples/user_guide/Network_Graphs.ipynb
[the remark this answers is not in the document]
jlstevens: > Just to clarify, is assigning positions to the nodes lazy, in that no positions will be assigned if no one ever tries to visualize it?

That sounded good at first but then I realized I don't want to have to reason about a datastructure based on whether it has been looked at or not. In addition, I don't think there would be a reasonable mechanism for such conditional behavior.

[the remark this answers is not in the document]
jbednar: Hmm.  Automatically assigning node positions for no reasons, for a data structure that could be quite large, also seems unreasonable to me. Not sure what you mean by "conditional", here.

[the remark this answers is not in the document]
philippjfr: Computing a circular layout is not particularly expensive and I don't really see the point of creating a completely abstract graph Element. The way I see it you will instantiate a Graph element when you want to either datashade it or display it directly and both require concrete positions. You also certainly don't want to use HoloViews pickles as a storage format for your **huge** graph. Finally, we don't have the concept for a conditionally displayable element anywhere else, so overall I think it's the most sensible behavior, unless you have a concrete use case for actually creating a Graph with raw edges.

philippjfr:
My to do list is basically:

* Once we have implemented the ``op`` aesthetic mapping suggestion we should add an example that maps egde weights to line_width or alpha.
* Make a pure datashader example to show how datashader bundling shifts the nodes.

examples/user_guide/Network_Graphs.ipynb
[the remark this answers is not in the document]
jbednar: Computing a circular layout isn't expensive in time, but it is expensive in storage, at least for a sparse graph -- each node then gets two extra columns of data.  I'm still not sure what either of you mean by "conditional"; conditional on what?  I was suggesting that the element always be displayable, just that this extra data not be computed until precisely the point where it first needs to be displayed.

The use case I imagine for that is to declare a huge Graph but then use HoloViews machinery to select only a subset of it to make a much smaller Graph that you actually display. I'm not worried about pickling it, just about making the handy tools in HoloViews available without forcing the Graph to be something displayable, as that's a separate consideration.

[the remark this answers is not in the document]
philippjfr: >I was suggesting that the element always be displayable, just that this extra data not be computed until precisely the point where it first needs to be displayed.

Okay sure, I can make it a property which computes nodes the first time they are accessed, which will basically happen when you call any method.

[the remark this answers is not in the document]
jbednar: @jlstevens, is that ok?  If you veto it, I won't object; up to you.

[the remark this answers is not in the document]
jlstevens: If the idea of the property is simply to defer assigning positions in a way that you can't tell the positions weren't there all along (except by tracking memory usage) then that seems ok to me. I won't veto it and I'll let Philipp decide although it does strike me a little bit as being premature optimization for a case we don't have a real example of yet.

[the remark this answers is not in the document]
philippjfr: Pushed it now, revert it if you do object.

holoviews/element/graphs.py
```diff
@@ -0,0 +1,259 @@
...
+    dimension corresponding to the node index.
+    """
+
+    kdims = param.List(default=[Dimension('x'), Dimension('y'),
+                                Dimension('index')], bounds=(3, 3))
+
+    group = param.String(default='Nodes', constant=True)
+
+
+class EdgePaths(Path):
```
jlstevens: I'm wondering whether it is worth introducing ``EdgePaths``. Why not just use ``Path``?

holoviews/element/graphs.py
```diff
@@ -0,0 +1,259 @@
...
+    y = np.sin(circ)
+    return (x, y, nodes)
+
+
+class layout_nodes(Operation):
+    """
+    Accepts a Graph and lays out the corresponding nodes with the
+    supplied networkx layout function. If no layout function is
+    supplied uses a simple circular_layout function.
+    """
```
jlstevens: It's fine for this operation to live here for now. But if we end up with a bunch more operations, they should move to that module.

holoviews/operation/datashader.py
```diff
@@ -651,3 +658,32 @@ def _process(self, element, key=None):
...
+def split_dataframe(path_df):
+    """
+    Splits a dataframe of paths separated by NaNs into individual
+    dataframes.
+    """
+    splits = np.where(path_df.iloc[:, 0].isnull())[0]+1
+    return [df for df in np.split(path_df, splits) if len(df) > 1]
+
+
+class bundle_graph(Operation, hammer_bundle):
```
jlstevens: Another graph operation ... but this time depending on datashader. Not sure where these should go...

holoviews/plotting/bokeh/graphs.py
```diff
@@ -0,0 +1,193 @@
...
+      Index of the dimension from which the color will the drawn""")
+
+    selection_policy = param.ObjectSelector(default='nodes', objects=['edges', 'nodes', None], doc="""
+        Determines policy for inspection of graph components, i.e. whether to highlight
+        nodes or edges when selecting connected edges and nodes respectively.""")
+
+    inspection_policy = param.ObjectSelector(default='nodes', objects=['edges', 'nodes', None], doc="""
+        Determines policy for inspection of graph components, i.e. whether to highlight
+        nodes or edges when hovering over connected edges and nodes respectively.""")
+
```
jlstevens: Shorter names would be nice. Maybe ``select_by`` and ``inspect_by``?

holoviews/element/graphs.py
```diff
@@ -180,12 +178,31 @@ def range(self, dimension, data_range=True):
...
+            return dimensions+node_dims
         return dimensions
 
 
     @property
+    def nodes(self):
+        """
+        Computes the node positions the first time they are requested
+        if no explicit node information was supplied.
+        """
```
jlstevens: I still think this is a bit weird. I won't revert it though but I would like to keep the option to reconsider later...

holoviews/plotting/bokeh/graphs.py
jlstevens: Or ``select_mode`` and ``hover_mode``?

jlstevens:
The tests are now green and I'm very happy with the API overall. There are three things I want us to follow up on:

* Some of the new comments I have just made.
* The example with edge thicknesses representing weight.
* Dedicated ``Nodes`` and ``EdgePaths`` notebooks as these are elements too.

I don't think these issues should hold up this PR any longer as any PR adding a major new feature is going to need tweaks. I am happy to merge now so we can get testing and improving our new graph support!

PR #1832, opened 2017-08-30
Fixed scrubber bug and add support for discrete DynamicMap scrubber

philippjfr:
Addresses https://github.com/ioam/holoviews/issues/1567 fixing a bug in the way scrubber handled looping. Additionally also adds support for using the scrubber when a DynamicMap defines ``values`` for all its dimensions, which effectively makes it discrete and therefore the same as a Holomap.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -639,11 +639,7 @@ class LabelledData(param.Parameterized):
             for k, v in self.items():
                 new_val = v.map(map_fn, specs, clone)
                 if new_val is not None:
-                    # Ensure key validation doesn't cause errors
-                    try:
-                        deep_mapped[k] = new_val
-                    except KeyError:
-                        pass
+                    deep_mapped[k] = new_val
             if applies: deep_mapped = map_fn(deep_mapped)
             return deep_mapped
         else:
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -639,6 +639,7 @@ class LabelledData(param.Parameterized):
             for k, v in self.items():
                 new_val = v.map(map_fn, specs, clone)
                 if new_val is not None:
+                    # Ensure key validation doesn't cause errors
                     try:
                         deep_mapped[k] = new_val
                     except KeyError:
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -639,7 +639,10 @@ class LabelledData(param.Parameterized):
             for k, v in self.items():
                 new_val = v.map(map_fn, specs, clone)
                 if new_val is not None:
-                    deep_mapped[k] = new_val
+                    try:
+                        deep_mapped[k] = new_val
+                    except KeyError:
+                        pass
             if applies: deep_mapped = map_fn(deep_mapped)
             return deep_mapped
         else:
--- a/holoviews/core/traversal.py
+++ b/holoviews/core/traversal.py
@@ -8,7 +8,7 @@ from collections import defaultdict
 from operator import itemgetter
 
 from .dimension import Dimension
-from .util import merge_dimensions
+from .util import merge_dimensions, cartesian_product
 
 try:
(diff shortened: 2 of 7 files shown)


philippjfr:
@jlstevens Ready to review. The basics of it is that a DynamicMap which defines a discrete sampling which can be treated just the same as a HoloMap, which also means you scrub through the parameter space. The unique_dimkeys method computes the unique keys by taking the cartesian product of the DynamicMap dimension values. Additionally it fixes a small bug in ``.map`` and various bugs in the scrubber.

holoviews/core/dimension.py
```diff
@@ -639,7 +639,11 @@ def map(self, map_fn, specs=None, clone=True):
             for k, v in self.items():
                 new_val = v.map(map_fn, specs, clone)
                 if new_val is not None:
-                    deep_mapped[k] = new_val
+                    # Ensure key validation doesn't cause errors
```
jlstevens: I don't quite get why there would be key errors: ``deep_mapped`` is a clone of ``self`` and ``k`` comes from ``self.items()`` so why would the key ever be rejected?

holoviews/core/traversal.py
```diff
@@ -89,6 +89,11 @@ def unique_dimkeys(obj, default_dim='Frame'):
...
                 unique_keys.append(padded_key)
 
+    # Add cartesian product of DynamicMap values to keys
+    values = [d.values for d in all_dims]
+    if obj.traverse(lambda x: x, ['DynamicMap']) and values and all(values):
+        unique_keys += list(zip(*cartesian_product(values)))
+
     with item_check(False):
         sorted_keys = NdMapping({key: None for key in unique_keys},
                                 kdims=all_dims).data.keys()
```
jlstevens: Why do we do this? Is this more information than taking the set of keys then sorting them?

holoviews/plotting/renderer.py
```diff
@@ -301,15 +301,16 @@ def get_widget(self_or_cls, plot, widget_type, **kwargs):
         if not isinstance(plot, Plot):
             plot = self_or_cls.get_plot(plot)
         dynamic = plot.dynamic
+        # Whether dimensions define discrete space
+        discrete = all(d.values for d in plot.dimensions)
```
jlstevens: This seems to be the basic condition that determines when scrubber can be used (with ``DynamicMap``).

jlstevens:
Looks good - I just asked a few questions in the comments. Not sure why the ``s3-reference-data-cache`` isn't green yet but the other tests have passed and I am happy to merge.

holoviews/core/traversal.py
philippjfr: There used to be at least, not sure its the case anymore though. Might leave a not to simplify this.

holoviews/core/dimension.py
philippjfr: Just had to double check to figure out what's going on, apparently this can occur when working with multiple DynamicMaps in a Layout. I've not been able to track down the actual bug but it seems something else is bypassing the validation so you end up with a cache that contains values it shouldn't. I'll try to track down the original bug that leads to the invalid key being inserted in the first place.

philippjfr: Found it, DynamicMap is inserting stuff into the cache with checking first.

philippjfr:
Thanks for the comments, I've reverted the change to ``.map`` and ensured that DynamicMap correctly validates its keys and that it doesn't warn about ``KeyError`` because that's the accepted signal to declare a key value as invalid (and display an empty frame).

holoviews/core/traversal.py
philippjfr: Actually I'll open an issue because this is related to sorting issues.

philippjfr: Opened the issue here: https://github.com/ioam/holoviews/issues/1835

jlstevens:
Looks good. Merging.

PR #1836, opened 2017-08-31
Histogram improvements

philippjfr:
Cleaned up and made various outstanding fixes for PR https://github.com/ioam/holoviews/pull/1784.

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -12,7 +12,8 @@ from param import _is_number
 from ..core import (Operation, NdOverlay, Overlay, GridMatrix,
                     HoloMap, Dataset, Element, Collator, Dimension)
 from ..core.data import ArrayInterface, DictInterface
-from ..core.util import find_minmax, group_sanitizer, label_sanitizer, pd, basestring
+from ..core.util import (find_minmax, group_sanitizer, label_sanitizer, pd,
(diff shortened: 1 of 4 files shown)


philippjfr:
@jlstevens I've now got this passing. Do you still want a ``label_format`` parameter as you suggested in the original PR?

jlstevens:
I still think the ``label_format`` idea is a good one, so it would be nice to include if it doesn't add too much additional work.

philippjfr:
Not sure I like ``label_format`` for this, since it's the value dimension label but could be confused with the regular ``label``. Not sure I have much better suggestions though since I don't like ``vdim_label``.

jlstevens:
So your issue is not with the idea, just the parameter name? How about``binned_format`` or ``aggregate_label`` or something like that?

philippjfr:
``frequency_format`` is probably the most correct. Is that too long?

jlstevens:
It is a bit long but I think it is ok. I'll try to think of something shorter but if I can't come up with anything, it will do.

philippjfr:
Actually I probably prefer ``frequency_label``.

philippjfr:
Tests are now passing (apart from a single transient), so I'll merge. @timothydmorton Thanks again for contributing, sorry this took so long to merge but it'll be in the next release.

PR #1841, opened 2017-09-03
Completely remove NdElement

philippjfr:
We originally added a deprecation warning whenever an NdElement is constructed in v1.7. Removing it as part of v1.9 therefore seems safe. Will be very glad to finally be rid of it.


philippjfr:
@jlstevens Ready when tests pass (which they are). So happy about this :fire: :fire: :fire:

jlstevens:
Hoorah! 

This signals the end of the reign of ``NdElement`` and the dawn of the age of data interfaces! :-)

PR #1843, opened 2017-09-03
Added support for reverse sort of tabular data

philippjfr:
Add support for reverse sort to tabular interfaces as requested in https://github.com/ioam/holoviews/issues/1167. Adds a ``reverse`` argument to the ``Dataset.sort`` method, alternatively could rename it ``ascending``.


jlstevens:
Personally, I'm ok with ``reverse`` as I assume normal sort is ascending.

philippjfr:
In that case this is ready to merge, I've added a blurb about it in the Tabular Data user guide.

jlstevens:
Looks good. Merging.

PR #1845, opened 2017-09-03
Disable reorderable option on bokeh TablePlot

### LINKED ISSUE #1844: Table broken with bokeh 0.12.7
When displaying a Table with bokeh 0.12.7 I currently see the following error:

```
Javascript error adding output!
Error: SlickGrid's 'enableColumnReorder = true' option requires jquery-ui.sortable module to be loaded
See your browser Javascript console for more details.
```

philippjfr:
Fixes #1844, which is a compatibility issue with bokeh 0.12.7 caused by removing jQuery-UI from bokeh. Disabling reordering which isn't hugely useful anyway fixes it.


jlstevens:
Looks like a simple but necessary fix. Merging.

PR #1848, opened 2017-09-04
Bugs in bokeh Arrow implementation.

philippjfr:
When the number of Arrow Elements in an NdOverlay exceeds the ``legend_limit`` this condition would generate an error.


jlstevens:
Simple fix. Merging.

Issue #1834, opened 2017-08-31
No color with datashader backend


bicycle1885:
Hi,

I've tried the Large data example (https://github.com/ioam/holoviews/blob/aeccd50d41e64d412ed5ae80c80e10956648a51e/examples/user_guide/14-Large_Data.ipynb) on my local machine (macOS, Sierra) and found it does not show colors properly:
![screen shot 2017-08-31 at 12 45 05](https://user-images.githubusercontent.com/905683/29905549-4abda856-8e4a-11e7-94b3-3e4da41a15c9.png)

I expect each gaussian is shown in different colors as demonstrated in the docs (http://holoviews.org/user_guide/Large_Data.html):
![screen shot 2017-08-31 at 12 46 14](https://user-images.githubusercontent.com/905683/29905581-88178de8-8e4a-11e7-9ca5-3e640dc1e40f.png)

I installed holoviews, bokeh, and datashader using conda as follows:

    conda create -n holo -c ioam holoviews bokeh datashader
    source activate holo
    conda update -n holo -c bokeh datashader  # this is required to run the example notebook
    jupyter notebook

And these are the package versions installed:
```
$ conda list -n holo
# packages in environment at /Users/kenta/miniconda3/envs/holo:
#
appnope                   0.1.0                    py35_0  
bkcharts                  0.2                      py35_0  
bleach                    1.5.0                    py35_0  
bokeh                     0.12.7                   py35_0  
cloudpickle               0.4.0                    py35_0  
colorcet                  0.9.1                    py35_0    bokeh
cycler                    0.10.0                   py35_0  
dask                      0.15.2                   py35_0  
datashader                0.6.0                    py35_0    bokeh
datashape                 0.5.4                    py35_0  
decorator                 4.1.2                    py35_0  
entrypoints               0.2.3                    py35_0  
freetype                  2.5.5                         2  
holoviews                 1.8.3            py35hce7455e_0    ioam
html5lib                  0.9999999                py35_0  
icu                       54.1                          0  
ipykernel                 4.6.1                    py35_0  
ipython                   6.1.0                    py35_0  
ipython_genutils          0.2.0                    py35_0  
ipywidgets                6.0.0                    py35_0  
jbig                      2.1                           0  
jedi                      0.10.2                   py35_2  
jinja2                    2.9.6                    py35_0  
jpeg                      9b                            0  
jsonschema                2.6.0                    py35_0  
jupyter                   1.0.0                    py35_3  
[...]

PR #1850, opened 2017-09-04
Added separate color_key for datashader shade to use color cycle by default

jbednar:
PR https://github.com/ioam/holoviews/pull/1697 changed the default cmap for datashader's ``shade()`` operation to "fire".  @philippjfr wanted this replace datashader's default blue ``cmap``, but it also replaced datashader's default ``color_key`` of Sets1to3, causing all categorical datashded output to be mapped to the first few colors of the fire colormap, which all look nearly identical  (as reported in #1834).  

The original behavior for categories can be restored by explicitly passing ``Sets1to3`` to ``shade.cmap`` whenever a category is used, but that's awkward and mysterious; the defaults should not be covering up categories. 

This PR adds a separate ``color_key`` parameter, defaulting to None (and thus falling through to Sets1to3), leaving ``cmap`` at the default ``fire`` value for now. 

Personally, I think ``cmap`` should not be defaulting to ``fire`` either, because ``fire`` is intended only for black backgrounds, and the Bokeh defaults consistently use white backgrounds. So I wonder whether we should remove the "fire" default?  E.g. if we set ``shade.cmap="#30a2da"``, we get behavior that is similar to the non-datashaded colors used by hv/bokeh:

image

It's not very pretty, but seems better than fire:

image

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -445,7 +445,13 @@ class shade(Operation):
     """
 
     cmap = param.ClassSelector(default=fire, class_=(Iterable, Callable, dict), doc="""
-        Iterable or callable which returns colors as hex colors.
+        Iterable or callable which returns colors as hex colors, to
+        be used for the colormap of single-layer datashader output. 
+        Callable type must allow mapping colors between 0 and 1.""")
+
+    color_key = param.ClassSelector(class_=(Iterable, Callable, dict), doc="""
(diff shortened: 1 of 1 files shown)


jbednar:
Well, what's a good colormap for white backgrounds?  I don't really have a strong preference in that case, like I do for black backgrounds.

jbednar:
I'll go ahead and merge this to get the example working, but I still think "fire" is not a good default for datashaded results.

PR #1865, opened 2017-09-11
Fix for bokeh BoxWhiskerPlot category factor handling

philippjfr:
Fixes bug in bokeh BoxWhiskerPlot when one categorical dimension is used.


jlstevens:
Thanks. As this case wasn't caught by earlier tests it won't be caught now so there is no point waiting. @philippjfr That said, we should make sure to test this case in a later PR!

Merging.

PR #1872, opened 2017-09-12
Correctly handle invert_x/yaxis options on Image/RGB

philippjfr:
As reported in https://github.com/ioam/holoviews/issues/1853 (although that's not the right issue), Image and RGB currently do not handle the invert_xaxis and invert_yaxis options correctly.


philippjfr:
Ready to merge.

jlstevens:
I've recently bumped into this issue myself so it is good to see it fixed. Merging.

PR #1873, opened 2017-09-12
Fixed wrongly inherited code on Pointer callbacks

philippjfr:
The PointerX and PointerY callbacks are wrongly inheriting JS code that is meant only for the PointerXY callback.


jlstevens:
The code is a bit redundant but that is better than trying to parameterize for just two cases ('x' and 'y'). Merging.

PR #1874, opened 2017-09-12
Fixed bug inverting categorical axes

philippjfr:
The ``_get_factors`` returns the categorical factors for the axes and inverts them if ``invert_axes=True``. Since coords is a generator inverting it requires that it is a tuple or list not a generator, simple but important fix.


jlstevens:
Explanation makes sense. Merging.

PR #1875, opened 2017-09-12
Added support for initial_hooks to bokeh

philippjfr:
Matches the initial_hooks plot option that we've had in matplotlib for a long time.


jlstevens:
Maybe this is an opportunity to document an ``initial_hooks`` example in the user guide for both matplotlib and bokeh? We have the 'Plotting with matplotlib' and 'Plotting with bokeh' user guides iirc...

jlstevens:
@philippjfr Do you not like this idea anymore? This PR languished but the ideas seemed sound enough...

philippjfr:
I've never really used initial_hooks and they don't actually make much sense for bokeh since *I think* in matplotlib they are usually used to modify the Figure or axes in some way before artists start drawing on them.

PR #1876, opened 2017-09-12
Cleaned up Element range handling

philippjfr:
As described in https://github.com/ioam/holoviews/issues/1870, QuadMesh is currently not handling explicitly set dimension ranges. This PR adds a utility to combine the data range with the dimension ranges and uses it in various Element.range implementations.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -710,6 +710,18 @@ def max_range(ranges):
         return (np.NaN, np.NaN)
 
 
+def dimension_range(lower, upper, dimension):
+    """
+    Computes the range along a dimension by combining the data range
+    with the Dimension soft_range and range.
+    """
+    lower, upper = max_range([(lower, upper), dimension.soft_range])
+    dmin, dmax = dimension.range
+    lower = lower if dmin is None or not np.isfinite(dmin) else dmin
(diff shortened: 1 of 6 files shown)


holoviews/core/util.py
```diff
@@ -710,6 +710,18 @@ def max_range(ranges):
         return (np.NaN, np.NaN)
 
 
+def dimension_range(lower, upper, dimension):
+    """
+    Computes the range along a dimension by combining the data range
+    with the Dimension soft_range and range.
+    """
```
jlstevens: I would be surprised if we don't have very similar logic to this somewhere else in holoviews already...

jlstevens: I see you replaced two chunks of code with this utility below, is that everywhere this utility could be used?

jlstevens:
Looks good. 

It would be good if you could check that ``dimension_range`` can't be used in more places (I suspect so, but I might be wrong!). Otherwise happy to merge.

philippjfr:
I think this was long overdue tbh, there were various inconsistencies in the implementations.

jlstevens:
Tests are green, if you think you've used this new utility everywhere that makes sense, I'm happy to merge.

philippjfr:
Yes, that's every single range implementation now.

PR #1877, opened 2017-09-12
Ensure that title_format plot option is inherited by OverlayPlot

philippjfr:
As outlined in https://github.com/ioam/holoviews/issues/1859, the ``title_format`` plot option should be propagated upwards to the OverlayPlot so that it is not ignored when overlaying an Element.


jlstevens:
Makes sense. Merging.

PR #1878, opened 2017-09-12
Fixed bug regridding RGB/HSV types

philippjfr:
Fixes https://github.com/ioam/holoviews/issues/1857, ensuring that RGB and HSV types can be regridded correctly.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -400,8 +400,8 @@ class regrid(ResamplingOperation):
                 arrays.append(element.dimension_values(vd, flat=False))
             if len(arrays) > 1:
                 array = np.dstack(arrays).transpose([2, 0, 1])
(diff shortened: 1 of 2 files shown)


holoviews/operation/datashader.py
```diff
@@ -400,7 +400,7 @@ def _process(self, element, key=None):
                 arrays.append(element.dimension_values(vd, flat=False))
             if len(arrays) > 1:
                 array = np.dstack(arrays).transpose([2, 0, 1])
-                coords['layer'] = range(len(arrays))
+                coord_dict['layer'] = range(len(arrays))
                 dims = ['layer'] + dims
```
jlstevens: Could we use a more obscure name that conveys that users should never actually see 'layer'? This would also avoid clobbering a potentially user defined name...

philippjfr:

jlstevens:

PR #1881, opened 2017-09-13
Fixed incorrect scoping bug in widget template

philippjfr:
Fixes https://github.com/ioam/holoviews/issues/1871 addressing an issue with the scope of the widget value and label variables.


jlstevens:
Glad you managed to fix this quickly! I only wish we had a way to test this properly.

Ready to merge?

philippjfr:
Once we have new widgets we'll have to bite the bullet and implement some selenium tests.

PR #1884, opened 2017-09-13
v1.8.4 hotfix release

jlstevens:
This PR squashes the cherry picked commits on the release_1.8.4 branch (a temporary branch that will be deleted) in order to make a 1.8.4 hotfix release.


jlstevens:
As this PR represents a new workflow for making hotfix release, I think it is worth summarizing the process:

1. Each hotfix release branches should be branched off the previous one (e.g 1.8.4 is branched off 1.8.3)
2. The bug fixes are cherry picked onto a temporary hotfix branch (e.g ``release_1.8.4``)
3. Once we are happy with that, we can make a single diff between the head of that branch and the previous hotfix release tag to get a single diff file. In this case I used ``git diff v1.8.3``.
4. That diff can be applied on a second branch, also off the previous hotfix. In this case this is the ``hotfix_1.8.4`` branch used in this PR. The goal is to have a single commit with all the bug fixes with an annotated tag (here ``v1.8.4``).
5. The release can be made off that commit, then this PR can be merged into master with an appropriate merge commit (NOTE: DO *NOT* SQUASH MERGE!) and the temporary branch can be deleted.

jlstevens:
Tests are passing. Merging.

PR #1889, opened 2017-09-15
Fix for the removal of the notebook_type argument in bokeh

jlstevens:
Tiny PR for bokeh compatibility after the removal of the ``notebook_type`` argument since bokeh 0.12.9 (i.e current bokeh dev).


philippjfr:
Thanks, looks good.

PR #1894, opened 2017-09-18
Support for using bokeh protocol in the notebook including binary transfer

philippjfr:
The https://github.com/bokeh/bokeh/pull/6945 PR adds support for sending bokeh events in the notebook using the regular bokeh protocol. This means we can now take advantage of binary transfer for arrays providing a ~5x speedup, which is particularly noticeable for images. Additionally this will allow us to dynamically add new glyphs to a plot, which means a DynamicMap will be able to return varying numbers of layers with a bit of work.

For now I've left in support for the old way of generating JSON patches and moved some small deprecated bokeh functions into holoviews, as it's not quite clear to me yet whether we can use the protocol for statically embedded data.

### DIFF
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -1,6 +1,7 @@
+import json
 from itertools import groupby
-import numpy as np
 
+import numpy as np
 import param
 
 from bokeh.models import (ColumnDataSource, Column, Row, Div)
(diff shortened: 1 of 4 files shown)


jbednar:
> Additionally this will allow us to dynamically add new glyphs to a plot, which means a DynamicMap will be able to return varying numbers of layers with a bit of work.

That sounds really helpful, and is something that will greatly simplify a lot of user code.  What was the change that made that possible?

philippjfr:
>What was the change that made that possible?

It's not implemented yet in HoloViews, and will still require a bit of thought. The change in bokeh is that we can now easily send new models to the frontend in the notebook.

philippjfr:
I think we should get this PR merged as is and I'll separately investigate whether we could/should use the protocol (with or without binary encoding) for embedding data.

jlstevens:
Very happy with the speedup this will give us and the prospect of being able to push new models is very exciting. Merging.

PR #1896, opened 2017-09-19
Bokeh protocol improvements

philippjfr:
Three small fixes to improve event handling with bokeh protocol:

* Place hold on Document after adding root (avoid RootAdded events)
* Avoid sending updates without events entirely
* Only update colormapper if attributes have changed (avoid ModelChanged events and glyph redraws)


philippjfr:
Ready to merge. At some point we should go through the entire bokeh backend and ensure that we don't needlessly set bokeh model attributes, the colormapper was just a very obvious case to do this since every change to it triggers a full rerender of the glyph.

jlstevens:
Looks good! Merging.

PR #1899, opened 2017-09-20
Widget throttling fix

philippjfr:
The widget event throttling code was an odd mixture of an old approach and the newer approach we use elsewhere, this switches fully to the new approach which blocks until Python sends back a complete message. I've also tidied up the rest of the code a fair bit removing old, dead code and fixing indentation.


philippjfr:
Would be good to merge this sooner rather than later so we can test it properly.

jlstevens:
> Would be good to merge this sooner rather than later so we can test it properly.

Agreed. Merging.

PR #1900, opened 2017-09-20
Fixed regression in bokeh 0.12.9 compatibility

philippjfr:
In a recent commit I broke compatibility with bokeh 0.12.9, tiny fix to restore it.


jlstevens:
I restarted the pr build for what I assume is a transient. Happy to merge once it passes.

PR #1901, opened 2017-09-20
Datashade paths interface fix

philippjfr:
When introducing the path interface I did not update the datashade operation. This adds that support. I also included an inline import for ``hammer_bundle`` to ensure it raises an exception when dependencies are not met.


jlstevens:
Looks good. Merging.

PR #1902, opened 2017-09-21
Fixed Graph selection using edge information

philippjfr:
Fixes an oversight, where selecting a Graph by edge information would leave nodes unfiltered.


jlstevens:
Looks good. Merging.

PR #1905, opened 2017-09-21
Optimized Path update with static_source

philippjfr:
When a static element is updated in a DynamicMap ``get_data`` on the plot is still called which can be fairly slow for large paths. This prevents the path coordinates to be unpacked unnecessarily. For example in this [census gerrymandering example](https://anaconda.org/jbednar/census-hv-dask/notebook) it saves about 600ms per update.

### DIFF
--- a/holoviews/plotting/bokeh/path.py
+++ b/holoviews/plotting/bokeh/path.py
@@ -42,11 +42,14 @@ class PathPlot(ElementPlot):
 
 
     def get_data(self, element, ranges=None, empty=False):
-        xidx, yidx = (1, 0) if self.invert_axes else (0, 1)
-        paths = [p.array(p.kdims[:2]) for p in element.split()]
-        xs = [] if empty else [path[:, xidx] for path in paths]
-        ys = [] if empty else [path[:, yidx] for path in paths]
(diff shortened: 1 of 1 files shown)


jlstevens:
I thought telling whether a source is static has its own issues and can't always be relied on to be correct. I vaguely remember discussing some idea where we use the ``id`` of the element and the ``id`` of the data - was that implemented or am I confused about what ``static_source`` means here?

philippjfr:
>I vaguely remember discussing some idea where we use the id of the element and the id of the data - was that implemented or am I confused about what static_source means here?

That was a while back, we fixed that using the id of the element rather than the data, which is safe and hasn't caused any issues.

jlstevens:
OK great. In that case I am happy to merge once the tests pass.

philippjfr:
I feel like the rate of transients has gone up massively recently. I kind of think there are certain tests that fall into one of two conditions and in a recent test data update we captured the rarer condition.

jlstevens:
>  I kind of think there are certain tests that fall into one of two conditions and in a recent test data update we captured the rarer condition.

Certainly plausible! Anyway, the tests have passed now. Merging.

PR #1907, opened 2017-09-21
Add support for datashader Graph layouts

philippjfr:
The ``layout_nodes`` operation now supports datashader LayoutAlgorithm functions and allows returning a Graph containing the newly computed node positions.

### DIFF
--- a/holoviews/element/graphs.py
+++ b/holoviews/element/graphs.py
@@ -9,9 +9,9 @@ from .chart import Points
 from .path import Path
 
 try:
-    from datashader.layout import LayoutAlgorithm
+    from datashader.layout import LayoutAlgorithm as ds_layout
 except:
-    LayoutAlgorithm = None
+    ds_layout = None
 
 
 class graph_redim(redim):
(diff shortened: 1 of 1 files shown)


jbednar:
Looks good to me.

philippjfr:
@jlstevens Ready to merge.

holoviews/element/graphs.py
```diff
@@ -36,22 +42,37 @@ class layout_nodes(Operation):
     """
     Accepts a Graph and lays out the corresponding nodes with the
     supplied networkx layout function. If no layout function is
-    supplied uses a simple circular_layout function.
+    supplied uses a simple circular_layout function. Also supports
+    LayoutAlgorithm function provided in datashader layouts.
     """
 
+    as_graph = param.Boolean(default=False, doc="""
+        Whether to return Nodes or Graph.""")
```
jlstevens: I don't particularly like this name. Why not simply ``graph``?

holoviews/element/graphs.py
```diff
@@ -8,6 +8,12 @@
 from .chart import Points
 from .path import Path
 
+try:
+    from datashader.layout import LayoutAlgorithm as ds_layout
+except:
+    ds_layout = None
+
+
 class graph_redim(redim):
```
jlstevens: Minor point but I would call this ``redim_graph`` instead.

jlstevens:
Looks good other than two simple naming issues I mention in the comments above.

holoviews/element/graphs.py
jlstevens: I don't see this parameter used in this PR. Where are you expecting it to be used?

philippjfr: Sure, I'll rename. It's used in ``if self.p.as_graph:`` below.

holoviews/element/graphs.py
philippjfr: Not really part of this PR but I'm happy to fix it here if you want me to.

jlstevens: Please do!

holoviews/element/graphs.py
jlstevens: I know it is used within this ``redim`` utility. What I'm wondering is where is redim used with this parameter set to ``True`` (assuming it is used with the default value by default).

philippjfr: It's not it's used in the ``layout_nodes`` operation, this PR doesn't touch ``redim`` at all.

philippjfr: Anyway how about ``only_nodes`` or ``nodes_only`` instead of ``as_graph``?

jlstevens: Ah, I got confused and thought it was in ``redim`` for some reason so you can ignore that comment. I prefer ``only_nodes`` though as nodes are always returned how about ``with_edges``?

philippjfr: I've gone with ``only_nodes`` and defaulted to False.

holoviews/element/graphs.py
philippjfr: Done.

philippjfr:
Ready if you're happy with it.

jbednar:
Sounds good, thanks!

jlstevens:
Looks good. Happy to merge when the tests pass though I would be even happier if this PR added some new ones!

PR #1908, opened 2017-09-21
Add stack operation for compositing RGBs

philippjfr:
Wraps around the datashader stack operation adding support for compositing multiple RGBs one of the defined compositor operations. The operation takes an Nd(Overlay) of RGBs as input and spits out a single composited RGB.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -614,15 +614,21 @@ class stack(Operation):
 
         imgs = []
         for rgb in overlay:
+            if not isinstance(rgb, RGB):
+                raise TypeError('stack operation expect RGB type elements, '
+                                'not %s name.' % type(rgb).__name__)
+            rgb = rgb.rgb
             dims = [kd.name for kd in rgb.kdims][::-1]
             coords = {kd.name: rgb.dimension_values(kd, False)
                       for kd in rgb.kdims}
             imgs.append(tf.Image(self.uint8_to_uint32(rgb), coords=coords, dims=dims))
+
         try:
             imgs = xr.align(*imgs, join='exact')
         except ValueError:
             raise ValueError('RGB inputs to stack operation could not be aligned, '
                              'ensure they share the same grid sampling.')
+
         stacked = tf.stack(*imgs, how=self.p.compositor)
         arr = shade.uint32_to_uint8(stacked.data)[::-1]
         data = (coords[dims[1]], coords[dims[0]], arr[:, :, 0],
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -583,6 +583,51 @@ class datashade(aggregate, shade):
 
 
 
+class stack(Operation):
+    """
+    The stack operation allows compositing multiple RGB Elements using
+    the defined compositing operator.
+    """
+
+    compositor = param.ObjectSelector(objects=['add', 'over', 'saturate', 'source'],
+                                      default='over', doc="""
+        Defines how the compositing operation combines the images""")
+
+    def uint8_to_uint32(self, element):
+        img = np.dstack([element.dimension_values(d, flat=False)
+                         for d in element.vdims])
+        if img.shape[2] == 3: # alpha channel not included
+            alpha = np.ones(img.shape[:2])
+            if img.dtype.name == 'uint8':
+                alpha = (alpha*255).astype('uint8')
+            img = np.dstack([img, alpha])
+        if img.dtype.name != 'uint8':
+            img = (img*255).astype(np.uint8)
+        N, M, _ = img.shape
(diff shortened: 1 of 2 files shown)


jbednar:
Supporting stack is very useful, but it's irritating that it takes more code here than the entire stack() operation in datashader itself. Is there some of this that ought to move to datashader?  I had trouble following the details.

philippjfr:
>Supporting stack is very useful, but it's irritating that it takes more code here than the entire stack() operation in datashader itself.

That's not really true though is it, stack call various compositing operators.

>Is there some of this that ought to move to datashader?

No I think it's out of scope, most of the code is converting between data formats and uint8 vs. uint32 RGB representations.

jbednar:
Well, those same compositing operators are called here as well, so I think it's fair. :-)  I guess I don't see datashader trying to support other formats, so I suppose this is fine as is.

philippjfr:
I guess so :unamused: There is some work we can do for RGB to interpret stacked arrays better, which would simplify this somewhat. Might also be worth considering moving the uint conversion onto RGB itself as getting the uint32 representation is frequently useful.

jbednar:
That does sound useful, yes.

philippjfr:
Here's an example:

```python
points1 = hv.Points(np.random.randn(1000000,2)+2)
points2 = hv.Points(np.random.randn(1000000,2))
ranges = dict(x_range=(-5, 5), y_range=(-5, 5))
shaded1 = datashade(points1, cmap=['red'], **ranges)
shaded2 = datashade(points2, cmap=['blue'], **ranges)
hv.Layout([stack(shaded1*shaded2, compositor=comp).relabel(comp)
           for comp in ['add', 'over', 'saturate', 'source']]).cols(2)
```

screen shot 2017-09-21 at 8 16 30 pm

jbednar:
Nice!  This will come in very handy when I try to document what's supported for those operators in datashader. :-)

philippjfr:
Need to add a check to ensure all the arrays have the same gridding. To help out with that kind of thing it might be nice to have an ``equalize_grids`` operation that uses ``regrid`` to equalize the gridding of multiple Image/RGB types in an (Nd)Overlay.

philippjfr:
Another issue is that the code that resolves which streams should be linked does not seem to deal with operations that collapse an overlay, so ``stack(datashade(points) * datashade(points))`` does not respond to RangeXY events on the plot.

philippjfr:
Finally the output appears to be flipped, will fix that as well.

jlstevens:
There are an infinite number of ``stack`` operations but I think it is ok in this case as it is ``datashader.stack`` which means it really is only about ``RGB`` elements. Though doesn't ``aggregate`` output ``Image``s?

philippjfr:
I'd be okay with a different name or better validation for input types if you think that's necessary to avoid confusion.

jlstevens:
I think ``datashader.stack`` is fine as long as it works with the output of all the operations in the ``datashader`` submodule (which by nature should all be rasters of some sort).

philippjfr:
>I think datashader.stack is fine as long as it works with the output of all the operations in the datashader submodule (which by nature should all be rasters of some sort).

Well that's not the case, it is indeed only for RGBs because they are specifically RGB compositing operations.

jlstevens:
So applying the colormap to ``Images`` to make them into ``RGB``s doesn't make sense?

philippjfr:
>So applying the colormap to Images to make them into RGBs doesn't make sense?

Automatically without providing control over which colormap? Not really no. The user should use shade directly. Happy adding an exception stating that.

jlstevens:
Maybe ``stack_shaded``?

jbednar:
``stack_shaded`` seems like a stretch to me.  What's wrong with ``hv.operation.datashader.stack``?

philippjfr:
Agreed, also the whole point of stacking is usually that you want different shading on the individual inputs. I'd rather just check that the input is the right type for now and provide a useful exception.

jbednar:
Usually I would think that people will less often use ``stack(shade(m),shade(n))`` than ``shade(m)*shade(n)``. I.e. stacking is very important at the datashader level, but in HoloViews, seems only needed for unusual compositing operators since HV already does overlay itself.

philippjfr:
Sending image data is still expensive so I'd think stacking is still frequently preferable over sending multiple images at once.

jbednar:
Ok, sure; we can suggest it as an optimization.

jlstevens:
I prefer short names anyway so if we don't think we need to worry about any other ``stack`` operations, then great. Ready to merge when the tests go green?

philippjfr:
Yes, ready.

jbednar:
Sounds good!

jlstevens:
Looks good. Merging.

PR #1911, opened 2017-09-23
Fixed bug in array interface groupby

philippjfr:
Caught a bug in the ArrayInterface groupby implementation and added a test.

### DIFF
--- a/holoviews/core/data/array.py
+++ b/holoviews/core/data/array.py
@@ -138,7 +138,6 @@ class ArrayInterface(Interface):
         # Get dimension objects, labels, indexes and data
         dimensions = [dataset.get_dimension(d, strict=True) for d in dimensions]
         dim_idxs = [dataset.get_dimension_index(d) for d in dimensions]
-        ndims = len(dimensions)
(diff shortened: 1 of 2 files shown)


jlstevens:
This reminds me about the issue about giving the pandas interface priority over numpy, and also the issue about testing consistency between numpy and pandas. Did we come to a decision on that?

Otherwise looks good and I'll merge when the tests pass.

philippjfr:
>This reminds me about the issue about giving the pandas interface priority over numpy

Agreed, I think we should just go ahead with that in 1.9 tbh.

>also the issue about testing consistency between numpy and pandas

Not really related to this, all the dataset tests are run against all the interfaces, this particular thing just wasn't handled. The consistency issues applied to two functions in particular I think ``util.unique_array`` and ``util.groupby``.

jlstevens:
> The consistency issues applied to two functions in particular I think util.unique_array and util.groupby.

I seem to remember that there *is* an issue about that. I'll look for it and assign it to the 1.9 milestone (if it isn't already).

Anyway, the tests passed. Merging.

PR #1912, opened 2017-09-23
Fixed py3 bug in Stream.subscribers

philippjfr:
Fixes https://github.com/ioam/holoviews/issues/1887


jlstevens:
Looks good. I remember when this bug came up and thanks for getting round to fixing it!

PR #1914, opened 2017-09-23
Improved stream resolution to allow collapsed (Nd)Overlays

philippjfr:
The code that resolves which plot to attach a linked stream to previously gave up when some part of the graph reduces an (Nd)Overlay by one or more layers since it cannot easily determine which layers were dropped. However in the case of operations that collapse multiple input layers into a single layer, e.g. the new datashader stack operation which combines multiple RGBs into one, there is no ambiguity since information from all the input layers persists. This means that an operation like this:

```python
stream = [RangeXY()]
stack(datashade(points1, streams=[stream]) * datashade(points2, streams=[stream]))
```

now correctly resolves the linked stream and therefore responds to zoom events. In order to provide more flexibility I've also exposed the ``link_inputs`` argument on the ``DynamicMap.map`` method in case linking is not desired.

Here's an example:

collapse_streams

```
   Points              Points
     |                   |
 datashade           datashade
     |                   |
    RGB ----- * ------- RGB
              |
           Overlay          <-- Previously gave up resolving streams here
              |      ^
            stack    |
              |      |  Direction of resolution
             RGB     |
```

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -1062,11 +1062,13 @@ class DynamicMap(HoloMap):
         self[key] = val
 
 
-    def map(self, map_fn, specs=None, clone=True):
+    def map(self, map_fn, specs=None, clone=True, link_inputs=True):
         """
         Recursively replaces elements using a map function when the
         specification applies. Extends regular map with functionality
-        to dynamically apply functions.
+        to dynamically apply functions. By default all streams are
+        still linked to the mapped object, to disable linked streams
+        set linked_inputs=False.
         """
         deep_mapped = super(DynamicMap, self).map(map_fn, specs, clone)
         if isinstance(deep_mapped, type(self)):
@@ -1074,7 +1076,8 @@ class DynamicMap(HoloMap):
             def apply_map(obj, **dynkwargs):
                 return obj.map(map_fn, specs, clone)
             dmap = Dynamic(deep_mapped, operation=apply_map,
-                           streams=self.streams, shared_data=True)
+                           streams=self.streams, shared_data=True,
+                           link_inputs=link_inputs)
             return dmap
         return deep_mapped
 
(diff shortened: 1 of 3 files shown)


jbednar:
Great, thanks!

philippjfr:
Ready to merge.

holoviews/plotting/util.py
```diff
@@ -130,23 +130,26 @@ def compute_overlayable_zorders(obj, path=[]):
     # Process the inputs of the DynamicMap callback
     dmap_inputs = obj.callback.inputs if obj.callback.link_inputs else []
     for z, inp in enumerate(dmap_inputs):
+        no_increment = False
```
jlstevens: Should this be called ``no_z_increment``?

jlstevens: Or ``zorder_unchanged``?

philippjfr: Yes, sure.

holoviews/core/spaces.py
```diff
@@ -1062,19 +1062,22 @@ def _cache(self, key, val):
         self[key] = val
 
 
-    def map(self, map_fn, specs=None, clone=True):
+    def map(self, map_fn, specs=None, clone=True, link_inputs=True):
```
jlstevens: Seems useful though I just want to check this feature is unrelated to the rest of the PR...

jlstevens:
Happy to merge once my one query and naming suggestion in the comments above are addressed.

holoviews/core/spaces.py
philippjfr: It's related but not directly I guess. Basically if we are going to allow operations which collapse layers to still link the inputs to that operation we should also make sure that the user can disable that behavior.

jlstevens: That makes sense. Thanks.

jlstevens:
Looks good and the tests are passing.

PR #1915, opened 2017-09-23
Switched to Whisker and Band glyphs for ErrorBars and Spread

philippjfr:
Bokeh introduced glyphs for errorbars and confidence bands a while ago, which both look better and are more efficient than the ``multi_line`` and ``patches`` implementation we were using previously. Since the categorical overhaul this is also the only way these elements will work with categorical axes.

Fixes: https://github.com/ioam/holoviews/issues/1893

Old ErrorBars:

screen shot 2017-09-23 at 3 54 35 pm

New ErrorBars (note the addition of whiskers):

screen shot 2017-09-23 at 3 54 46 pm

Old Spread:

screen shot 2017-09-23 at 3 56 10 pm

New Spread (note the lack of border lines one left and right):

screen shot 2017-09-23 at 3 55 42 pm

### DIFF
--- a/holoviews/plotting/bokeh/__init__.py
+++ b/holoviews/plotting/bokeh/__init__.py
@@ -145,7 +145,7 @@ options.Spread = Options('style', color=Cycle(), alpha=0.6, line_color='black')
 options.Bars = Options('style', color=Cycle(), line_color='black', width=0.8)
 
 options.Spikes = Options('style', color='black', cmap='fire')
-options.Area = Options('style', color=Cycle(), line_color='black')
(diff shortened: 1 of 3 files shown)


jlstevens:
Makes sense to update the glyphs. Any other places where you think Bokeh might now have a more suitable glyph? Or maybe some glyph ideas could be posted as feature requests to bokeh?

philippjfr:
Thanks for mentioning that, ``Area`` is also better written using the Band glyph. Will rebuild test data and then this will be ready to merge.

jlstevens:
Test are now green. Merging.

PR #1922, opened 2017-09-24
Fixed initial frame bug for matplotlib animations

philippjfr:
The plot was being initialized twice causing the initial frame to persist throughout the animation.

Fixes: https://github.com/ioam/holoviews/issues/1621


jlstevens:
Thanks for tracking it down! 
Does that mean the gif output now looks correct?

Merging.

PR #1923, opened 2017-09-24
Removed prepending of param to syspath

philippjfr:
I think this must be leftover from some time ago, unless you can think of a reason this should still be here @jlstevens. May also fix https://github.com/ioam/holoviews/issues/1920.

### DIFF
--- a/holoviews/__init__.py
+++ b/holoviews/__init__.py
@@ -3,10 +3,6 @@ from __future__ import print_function, absolute_import
 import os, sys, pydoc
 
 import numpy as np # noqa (API import)
-
-_cwd = os.path.abspath(os.path.split(__file__)[0])
-sys.path.insert(0, os.path.join(_cwd, '..', 'param'))
-
 import param
 
 __version__ = param.Version(release=(1,8,4), fpath=__file__,


jbednar:
I would guess that was for making "import version ; version.Version()" work, but it's surely much cleaner just to use ``param.version.Version`` or ``param.Version``.

philippjfr:
Don't think it's that, it's been there since before DataViews became HoloViews and appears to have been added when param was a submodule of dataviews (see https://github.com/ioam/holoviews/commit/6953a803894c4313c12a96dd61ed5fef58a35aae). Can't believe it stuck around this long!

jlstevens:
> Can't believe it stuck around this long!

I agree! That said, I'm also a little surprised it *ever* caused a problem. There is no reason to keep it around anymore so I'll merge the PR.

philippjfr:
>I'm also a little surprised it ever caused a problem.

Think about the site-packages structure, once you install this, this would have added all the files in the param package to the PATH, overriding version and parameterized. Any library importing a version module would then pick up on param.version if they didn't properly use relative imports.

PR #1924, opened 2017-09-24
Disable shared axes when +axiswise

philippjfr:
As suggested in https://github.com/ioam/holoviews/issues/1906, setting axiswise=True should force shared axes to be disabled.


jlstevens:
Great!

I am very tempted to hit merge right away, but first we should consider any backwards compatibility implications. Should we have a compatibility switch in ``config``?

philippjfr:
>Should we have a compatibility switch in config?

No I don't think so, I'd consider this a bug {+axiswise} has clear semantics which weren't being respected before.

jlstevens:
Ok. Just make sure it is clearly noted in the releases notes for the next release.

Merging.

Issue #1838, opened 2017-09-01
BoxWhisker documentation could be improved


gcbgit:
The current documentation on BoxWhisker states: "BoxWhisker represent data as a distributions highlighting, the median, mean and various percentiles."  I don't see the mean represented on the plot, and it's not clear from the documentation which percentiles are shown.  I presume that the box is the first and third quartiles, but that should be stated in the documentation for clarity.  There isn't an established convention on where the whiskers should be, so that should also be described in the documentation.

Also, the link in the reference gallery to "Columnar Data Tutorial" is broken.  The reference gallery states that the BoxWhisker represents data through their quartiles, which is true for the box, but the examples on that page show outliers so the whiskers aren't quartile related (they'd have to be 0 and 4, which would capture the full extent of the data, leaving no outliers).

jlstevens:
The PR referenced above has been merged. Closing.

PR #1926, opened 2017-09-25
Updated BoxWhisker in Bokeh to have Tukey-style whisker bounds.

justinbois:
The Bokeh version of BoxWhisker currently uses the following definitions:

- Top of box: 75th percentile
- Bottom of box: 25th percentile
- Middle of box: 50th percentile (median)
- Bottom whisker: 25th percentile, minus 1.5 times the interquartile region (75th - 25th percentile)
- Top whisker: 75th percentile, plus 1.5 times the interquartile region
- Outliers: Anything that is 1.5 IQRs away from top or bottom of box

While the definitions for the box and the outliers is pretty standard (and definitely what I would choose), I have never seen that definition of the whisker. Rather, the "most standard" definition is what [Tukey proposed](https://en.wikipedia.org/wiki/Box_plot).

- Bottom whisker: The larger of:  (25th percentile minus 1.5 times the IQR, smallest data point)
- Top whisker: The smaller of:  (75th percentile plus 1.5 times the IQR, largest data point)

In fact, this is what we get using `hv.BoxWhisker` with Matplotlib and Plotly renderers. So, I think it is a bug in the Bokeh version. This PR fixes that bug.

This is also relevant for #1838.

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -1024,8 +1024,8 @@ class BoxWhiskerPlot(CompositeElementPlot, ColorbarPlot, LegendPlot):
             vals = g.dimension_values(g.vdims[0])
             qmin, q1, q2, q3, qmax = (np.percentile(vals, q=q) for q in range(0,125,25))
             iqr = q3 - q1
-            upper = q3 + 1.5*iqr
-            lower = q1 - 1.5*iqr
+            upper = min(q3 + 1.5*iqr, vals.max())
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks for the PR! I agree that is the more standard definition and I'm happy with this change. Going to rebuild the test data now and then merge it.

philippjfr:
Tests passing now, going to merge. Thanks again for the contribution!

justinbois:
My pleasure!  I *love* HoloViews!

PR #1927, opened 2017-09-25
Fix for Graph edge hover labels

philippjfr:
Fixes bug where it would not find hover info because edge start/end points in datasource is always start/end while Graph.kdims can be any names.


jlstevens:
Looks good. Merging.

PR #1934, opened 2017-09-27
Added Reset stream

philippjfr:
A Reset stream triggered when the reset tool has been hit in bokeh can be quite useful to reset selections and other custom interactive behavior.


jlstevens:
Looks good. 

Just wondering if ``ResetEvent`` is clearer? The fully explicit name is ``ResetToolEvent`` but that is too awkward. My issue is that the name doesn't convey that it is a linked stream it *sounds* like it could be a general stream (though I don't know what it would mean or do!).

philippjfr:
``ResetEvent`` doesn't really sound like a stream to me, personally I think ``Reset`` is fine, but ``PlotReset``might be clearer.

jlstevens:
I do prefer ``PlotReset`` over ``Reset``.

philippjfr:
Okay sounds good, matches ``PlotSize`` which already exists.

philippjfr:
Tests passing, merging.

PR #1940, opened 2017-09-29
Added support for updating and replacing axis properties

philippjfr:
As I was investigating getting the MercatorTicker working in GeoViews I realized that the code that handles updating of axis properties never correctly handled updating of ticker and formatter models. I have now added a utility which handles recursive updates to bokeh models, preventing unnecessary updates and even handling cases where the model is being replaced (which we now allow in live and dynamic modes as well as on the server). Once I've figured out whether we can use the new bokeh protocol to embed data we may eventually be able to switch out models at will and use this update function everywhere, e.g. to replace a linear axis with a log or categorical axis on the fly.

Turns out this was also the reason why I had to filter ticker and formatter model updates from the static json patches before, allowing me to simplify that now.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -26,7 +26,8 @@ from ..plot import GenericElementPlot, GenericOverlayPlot
 from ..util import dynamic_update
 from .plot import BokehPlot, TOOLS
 from .util import (mpl_to_bokeh, get_tab_title, bokeh_version,
-                   mplcmap_to_palette, py2js_tickformatter, rgba_tuple)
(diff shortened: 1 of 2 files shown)


jbednar:
Excellent!!!

jlstevens:
Looks great! Especially together with #1949.

Is this ready to merge and test or do you think we should cut a dev release before merging these two PRs?

philippjfr:
Go ahead and merge this. I'll do some more testing before #1949 is ready.

jlstevens:
Tests are passing and I'm looking forward to seeing this in action and simplifying the code with #1949. Merging.

PR #1946, opened 2017-10-01
Standardized constructor signature with dimensions as positional args

philippjfr:
As suggested in https://github.com/ioam/holoviews/issues/1938 this PR standardizes the constructors of Dimensioned objects such that ``kdims`` and ``vdims`` are the second and third argument respectively for Elements and ``kdims`` are the second argument for dimensioned containers (NdMapping, HoloMap, NdOverlay etc.). Currently ``Histogram`` and ``Annotations`` are the exceptions to this.


jlstevens:
Are there backwards compatibility implications? Otherwise I support the idea.

philippjfr:
Not as far as I know but I have yet to go through all the errors.

holoviews/core/dimension.py
```diff
@@ -783,35 +783,20 @@ class to be associated with dimensions. The contents associated
...
-
-    @property
-    def value_dimensions(self): return self.vdims
-
-    @property
-    def constant_dimensions(self): return self.cdims
-
-    @property
-    def deep_dimensions(self): return self.ddims
-
```
jlstevens: I suppose we haven't used them in a long time so they should be ok to remove now.

philippjfr:
Also addresses https://github.com/ioam/holoviews/issues/1619

philippjfr:
Okay, I've now gone through everything. There were two causes for errors:

1) In some very old tests we passed Image ``bounds`` as the second positional argument. Since ``bounds`` is now optional and is only used when passing a numpy array to Image I think moving away from that is sensible.
2) In some cases some internal code was passing NdMapping kdims as tuples, which used to be allowed but is now interpreted as a single dimension specification.

I think better exception messages will be sufficient to address both cases.

jbednar:
Sounds reasonable, but be sure to link to this PR in the changelog section on backwards compatibility.

holoviews/core/dimension.py
philippjfr: You would think that, turns out Columnar_Data still had a reference:

      3 error_hmap = hv.HoloMap({(i, j): hv.Image(j*np.random.randn(3, 3), bounds=extents)
      4                          for i, j in product(range(3), np.linspace(0, 1, 3))},
      5                         key_dimensions=['Observation', 'noise'])

jlstevens:
All sounds good to me!

I definitely wouldn't want bounds to be passed by position to ``Image`` anymore.

philippjfr:
Tests now passing, I suppose I'll add some tests to ensure Elements follow this signature.

philippjfr:
Ready to merge once test pass.

holoviews/element/chart.py
```diff
@@ -165,6 +165,10 @@ class Histogram(Element2D):
     vdims = param.List(default=[Dimension('Frequency')], bounds=(1,1))
 
     def __init__(self, values, edges=None, **params):
+        if edges is not None:
+            self.warning("Histogram edges should be supplied as a tuple "
+                         "along with the values, passing the edges will "
+                         "be deprecated in holoviews 2.0.")
```
jlstevens: Something to mention in the next changelog/release notes. It will be good to get histogram working consistently with everything else.

jlstevens:
Looks good to me. I made one comment about histogram but nothing that needs action. Merging.

PR #1949, opened 2017-10-02
Use the bokeh protocol for embedding static widgets

philippjfr:
Turns out completely switching to the bokeh protocol even for static plots was trivial. This means that all bokeh messages are now generated in exactly the same way whether on bokeh server, using dynamic or live widgets and statically embedded widgets. The PR is based off https://github.com/ioam/holoviews/pull/1940, which was required to get this working. You might particularly appreciate the large amount of hacky code we can now delete.

Edit: I'll likely reapply the changes in this PR once #1940 is merged.


jbednar:
Whoa, that's a lot of pink and not much new green! Sounds great!

philippjfr:
So I'm an idiot. We can't just merge this right now, because the version of bokeh this depends on hasn't even been released yet.

jbednar:
Sounds like we should pin the bokeh version when we do release, though, and just say that's how it has to be!

philippjfr:
If we do that we should also remove all the other code for compatibility with older versions. @jlstevens can chime in.

jlstevens:
It's a little tricky. 

Keeping the cruft to keep compatibility with older bokeh versions would mean that we can't clean things up just yet. As developers, we would love to see code removed and the codebase simplified but I'm not sure that is a justifiable reason to disrupt our users (forcing them to upgrade to the latest bokeh).

I think the best approach is to keep backwards compatibility for at least one more release cycle, marking the code path needed by the old bokeh versions in a way that makes it clear that it it is cruft that should be removed when the time comes. We can then announce the versions of bokeh we will be dropping support for to our uses in the release notes of version 1.9 and we can assign an issue to the 2.0 milestone to remind us to drop the cruft before then.

philippjfr:
I'm not entirely sure it's really that disruptive forcing people to upgrade and version 0.12.10 will have large benefits over older versions due to the binary protocol. We've been bad at pinning the our bokeh version requirement, which has consistently caused problems due to churn in bokeh. I'd be in favor doing this sooner rather than later and just pinning the version requirement for 1.9 to bokeh 0.12.10.

jlstevens:
In that case, maybe we should issue a warning recommending a bokeh upgrade? Otherwise, we need a helpful exception message for people using older versions to help them upgrade.

jbednar:
My gut feeling is that we will better serve our users if we simplify and focus on one target Bokeh version while Bokeh is churning so much, and now seems like a great opportunity to do that because there is now one particular version that gives massive gains in speed and simplicity.  We can work really hard to support all versions, but as we've seen from bokeh 0.12.5 onwards, we end up forcing people to upgrade hv with each new Bokeh release anyway, and if they are already doing that, they may as well upgrade Bokeh with each new hv release.  It's a matched set!

Instead of supporting multiple Bokeh versions, I'd rather spend our backwards-compatibility efforts on our own API, making sure our own users' hv code keeps working, because that's where changes will really hurt them. For hv users, upgrading bokeh is easy; updating their hv code is hard.

So, I vote for:
- clearing out the Bokeh-version-related cruft now, immediately, on hv master
- everyone on the hv team using 0.12.10devX or Bokeh master for our own work to make sure it's well tested before release
- not releasing hv 1.9 until Bokeh 0.12.10 is released
- pinning bokeh=0.12.10 for hv 1.9.0 (not >=, but specifically =)
- adding a warning if an older version of Bokeh is in use with hv 1.9
- probably *not* adding a warning if a *newer* version of Bokeh is in use? (pinning bokeh=0.12.10 in conda should be enough to ensure people don't accidentally get a newer version, but if they force conda to install 0.12.11, then there's probably no reason for us to be complaining further (it *could* work!)
- for all subsequent hv releases until Bokeh 1.0, do all the steps above, choosing one specific target Bokeh version only
- before bokeh 1.0 is released, have an extensive hv code review to try to detect any non-public API we're using from Bokeh (maybe not yet quite the time for that, given Bryan's efforts to clean up the public/private distinction in Bokeh itself?)
- after bokeh 1.0 is released, pin bokeh>=1.0,<2.0, and trust that (a) we are using only public API, and (b) bokeh will follow semantic versioning. 

Sound good?

bryevdv:
I don't have alot to add: 

* After 1.0, Bokeh's "public" API (for general use) and "internal" API (for use by other projects/integrations) will be locked down under test. They will be stable, modulo unintended bugs that creep past tests. 
* Until then, there will be some churn, especially around use of currently "private" APIs (i.e. leading underscore) After 1.0 all `_private` things in Bokeh will be truly private: no guarantees *at all* will be made, they may change at any time. They should not ever be used outside the module they are defined in. 
* So close to 1.0, I don't think anyone should be in the Bokeh version checking game for purporses of maintaining indefinite backward compat. At 1.0 a clean break should be made. Until then, whatever works well for you sounds good. Another approach might be to make your own internal APIs to wrap any current bokeh internal/private API usage. 
*  If you can give me a list of currently used private APIs we can prioritize stabilizing those first.

philippjfr:
@bryevdv Thanks, that's helpful. After this PR I actually think we've finally moved away from private APIs completely, to a large degree enabled by the your recent work on stabilizing the API of ``bokeh.io`` and ``bokeh.embed`` modules. That I think is another good reason for us to draw a clean break here and simply say we require bokeh=0.12.10. Another main reason for me is that I simply do not have the bandwidth to check compatibility with older releases, so while we say we support older versions of bokeh I cannot guarantee it because our tests run against one particular version and I do not have the time to do manual testing to see which version is actually supported.

jlstevens:
Based on this discussion, I agree that requiring 0.12.10 is a reasonable thing to do. We just need to issue helpful exception messages to help people upgrade if they are using earlier bokeh versions.

philippjfr:
Let's merge this now then, since 0.12.10 is released. I'll follow up with another PR cleaning up the remaining bits of compatibility code for older bokeh versions.

jbednar:
Sounds good!

jlstevens:
Happy to merge a PR that deletes a lot more code than it adds!

PR #1951, opened 2017-10-02
Various Graph improvements

philippjfr:
This PR includes a number of fixes and improvements for Graphs:

* The EdgePaths on a Graph may now be split (e.g. a separate array/dataframe for each edge) or nan separated (as is most efficient for datashading). Unlike before both are validated to have the correct number of edges.
* Applying selections and plotting with bokeh are now both allowed when nan separated data is used.
* Selection on a Graph selects all edges that connect to one of the selected nodes (and keeps both the selected nodes and the nodes that they connect to) by default. If you want to select only edges between the selected subset of nodes you can use ``selection_mode='nodes'``. This is much more useful behavior to select an important subgraph and provides more flexibility.
* Added a ``directly_connect_edges`` operation (based on datashader), which is much faster than the automatic behavior (and not split by default).
* Hide x, y location of nodes on hover

### DIFF
--- a/examples/user_guide/Network_Graphs.ipynb
+++ b/examples/user_guide/Network_Graphs.ipynb
@@ -358,7 +358,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "To select just nodes in 'circle15', get a list of the indices of the nodes in that circle and then select by the edge start (aka source) and end (aka target) points:"
+    "To select just nodes that are in 'circle15' set the ``selection_mode='nodes'`` overriding the default of 'edges':"
    ]
   },
   {
@@ -367,8 +367,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "nodes = list(bundled.nodes.select(circle='circle15')['index'])\n",
-    "bundled.select(start=nodes, end=nodes)"
+    "bundled.select(circle='circle15', selection_mode='nodes')"
    ]
   }
  ],
--- a/examples/user_guide/Network_Graphs.ipynb
+++ b/examples/user_guide/Network_Graphs.ipynb
@@ -340,7 +340,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Alternatively we can select the nodes and edges by an attribute that resides on either. In this case we will select the nodes and edges for a particular circle and then overlay just the selected part of the graph on the datashaded plot. In this way a smaller subgraph can be highlighted and the larger graph can be datashaded."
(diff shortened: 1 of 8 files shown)


examples/user_guide/Network_Graphs.ipynb
```diff
@@ -340,7 +340,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Alternatively we can select the nodes and edges by an attribute that resides on either. In this case we will select the nodes and edges for a particular circle and then overlay just the selected part of the graph on the datashaded plot. In this way a smaller subgraph can be highlighted and the larger graph can be datashaded."
+    "Alternatively we can select the nodes and edges by an attribute that resides on either. In this case we will select the nodes and edges for a particular circle and then overlay just the selected part of the graph on the datashaded plot. Note that selections on the ``Graph`` itself will select all nodes that connect to one of the selected nodes. In this way a smaller subgraph can be highlighted and the larger graph can be datashaded."
```
jbednar: Is that a policy that can be selected?  Seems likely that's what people want, but not always.

holoviews/element/graphs.py
```diff
@@ -124,17 +127,37 @@ def __init__(self, data, **params):
...
+                edges = self.edgepaths.split()[0]
+                vals = edges.dimension_values(0)
+                npaths = len(np.where(np.isnan(vals))[0])
+                if not np.isnan(vals[-1]):
+                    npaths += 1
+                mismatch = npaths != nedges
+            if mismatch:
+                raise ValueError('Ensure that the number of edges supplied to '
+                                 'the Graph matches the number of edgepaths: '
+                                 'expected %d, found %d' % (nedges, npaths))
```
jbednar: Probably clearer as ``Ensure that the number of edges supplied to the Graph (%d) matches the number of edgepaths (%d)' % (nedges, npaths)

examples/user_guide/Network_Graphs.ipynb
philippjfr: Not currently no, but I guess I could. I do provide an alternative way of doing it by selecting on the source and target instead. Any suggestion on what such an option could be called, don't like my suggestions much: ``only_selected`` or ``include_connected``.

jbednar: ``include_targets``?

philippjfr: Doesn't seem correct, the graph might be directed in which case that's just confusing because you may actually be selecting the source.

philippjfr: I'm going to go with ``selection_mode='nodes'/'edges'`` unless there is a better suggestion.

jbednar: Ok by me.

philippjfr:
@jlstevens Ready to merge once tests pass.

jlstevens:
Everything mentioned sounds good from the description. I will need to play around with graphs a bit more myself before I have ideas/suggestions for further improvements. Merging.

PR #1957, opened 2017-10-03
Avoid seaborn API import warning

philippjfr:
Since seaborn 0.8.0 the apionly import is the default and trying to ``import seaborn.apionly as sns`` raises a deprecation warning.


jlstevens:
Happy to merge when the tests go green.

PR #1958, opened 2017-10-03
Optimize RGB regridding

philippjfr:
This simplifies and optimizes the regridding of RGB elements by regridding each value dimension individually, which removes the overhead from dstacking the data.


jlstevens:
Looks good. Merging.

PR #1961, opened 2017-10-04
Improvements for ndloc and sampling

philippjfr:
A while ago we added ``ndloc`` as a means to provide array like integer indexing for gridded Datasets. This has worked out well overall and this PR simply builds on top of that. In particular this PR adds an out-of-bounds IndexError when sampling outside the Image bounds. Secondly it adds an xarray ndloc implementation, which should be more efficient than the generic implementation that we currently have, ensure that the datatype doesn't change when sampling and selecting and also ensure that any xarray variables that are not referenced by HoloViews aren't dropped after a slicing operation. Also added a number of additional unit tests ensuring that the implementations behave identically.

### DIFF
--- a/holoviews/core/data/xarray.py
+++ b/holoviews/core/data/xarray.py
@@ -194,6 +194,44 @@ class XArrayInterface(GridInterface):
 
 
     @classmethod
+    def ndloc(cls, dataset, indices):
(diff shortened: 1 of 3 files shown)


philippjfr:
Ready to merge.

PR #1964, opened 2017-10-05
Fix static source optimization on Graph Element

philippjfr:
The Graph.clone method was not retaining the ``_plot_id`` which is what the static source optimization uses resulting in data being sent unnecessarily.


jlstevens:
Simple fix that makes sense. I'll merge when you tell me it is ready.

philippjfr:
Ready.

PR #1971, opened 2017-10-06
Handle Bars with log axes

philippjfr:
Handling logarithmic bar plots is a special case because Bars usually start at zero, which bokeh does not allow. This PR uses the same code that we usually use to display logarithmic plots in bokeh, which will provide a warning that it automatically computed a logy lower bound and the user should override it. 

Fixes: https://github.com/ioam/holoviews/issues/1780


philippjfr:
Ready to merge.

PR #1976, opened 2017-10-07
Fix for Position and Tap stream on categorical axes

philippjfr:
Since bokeh 0.12.9 hover and tap events over a categorical axis no longer report the categorical coordinate but a float value instead. This restores the old behavior by looking up the coordinate in the ``FactorRange``. This also fixes at least two gallery examples with bokeh>=0.12.7.

Fixes https://github.com/ioam/holoviews/issues/1955


philippjfr:
Ready to merge.

jlstevens:
Looks good. I assume this approach worked with earlier bokeh versions as well. Merging.

PR #1978, opened 2017-10-08
Enable situate_axes in matplotlib for consistency

philippjfr:
Very early on when implementing holoviews we added a ``situate_axes`` option which disables axis normalization for Raster and Image types in matplotlib. This pops up as a user question fairly frequently since it is inconsistent behavior.

Fixes: https://github.com/ioam/holoviews/issues/692

### DIFF
--- a/holoviews/plotting/mpl/raster.py
+++ b/holoviews/plotting/mpl/raster.py
@@ -22,7 +22,7 @@ class RasterPlot(ColorbarPlot):
     colorbar = param.Boolean(default=False, doc="""
         Whether to add a colorbar to the plot.""")
 
-    situate_axes = param.Boolean(default=False, doc="""
+    situate_axes = param.Boolean(default=True, doc="""
         Whether to situate the image relative to other plots. """)
 
     style_opts = ['alpha', 'cmap', 'interpolation', 'visible',


jlstevens:
Seems fine though maybe we should have a switch in ``hv.config`` for backwards compatibility? The failing tests indicate that changing this will affect the behavior of people's existing code.

philippjfr:
I'm okay adding that but afaik ``hv.config`` is currently undocumented which means it might as well not exist.

jlstevens:
The most important place to mention ``hv.config`` is in the compatibility section of the release notes, like we did for [1.8.0](https://github.com/ioam/holoviews/releases/tag/v1.8.0). And it is documented [here](http://holoviews.org/user_guide/Installing_and_Configuring.html).

philippjfr:
Ah okay, great. So are we just going to add new keywords to the config argument like ``style_18``? Doesn't seem quite right, since it isn't necessarily a style.

jlstevens:
We can use any keywords we want in ``config`` and I imagine it would be something like ``config.situate_axes_18``.

philippjfr:
>We can use any keywords we want in config and I imagine it would be something like config.situate_axes_18.

That seems pretty pointless imo, how would anyone discover that keyword exists? Seems easier to just have backward compatibility notice that says you can revert to ``hv.plotting.mpl.RasterPlot.situate_axes=False``. I'm not against a ``style_18`` flag or similar but I don't think every single tiny change should have its own entry in ``hv.config``.

jlstevens:
> I'm not against a style_18 flag or similar but I don't think every single tiny change should have its own entry in hv.config.

I do agree but if there are enough changes then they should go in a ``hv.config`` flag.

philippjfr:
I don't think this needs its own flag and I'm not aware of any other changes this could be grouped in with, so I'd prefer just to get this merged.

jlstevens:
Ok. Though at the very least this will need to be mentioned wrt backwards compatibility in the new CHANGELOG.

PR #1979, opened 2017-10-08
Handle zero-sized array reductions

philippjfr:
When applying the histogram and datashader operations to an empty array it currently errors due to zero-sized array reductions and zero-range errors. This PR ensures that both operations default to a 0-1 range in that case and also handles colormapping issues in matplotlib for empty arrays. 

Fixes: https://github.com/ioam/holoviews/issues/1506 and https://github.com/ioam/holoviews/issues/1522


philippjfr:
Ready to merge.

jlstevens:
Looks good. Merging.

PR #1980, opened 2017-10-08
Make copies of supplied bokeh tool instances

philippjfr:
Since you can only ever supply a bokeh model to a single plot and HoloViews objects can be displayed multiple times user supplied tool instances should be copied before attaching them to a plot.

Fixes https://github.com/ioam/holoviews/issues/1851


philippjfr:
Tests passed, merging.

PR #1983, opened 2017-10-08
Add display_options parameter to bokeh widgets

### LINKED ISSUE #1660: Bokeh server warning message
the code server = renderer.app(dmap, show=True, new_window=True)
results in a warning message
WARNING:root:BokehServerWidgets24371: Setting non-parameter attribute display_options={'fps': 20} using a mechanism intended only for parameters

philippjfr:
This simply makes the API of bokeh widgets consistent, ensuring that they do not raise a warning when display_options are passed. The only actual display_option being passed is the ``fps``, which does not apply to these widgets since they do not have a "play" feature.

Fixes #1660


jlstevens:
Looks good and ready to merge.

PR #1984, opened 2017-10-09
Refactored bokeh plotting API

philippjfr:
This PR represents a major cleanup of the (internal) bokeh plotting API to make it more consistent with matplotlib and simplify some things. Basically it adds the dictionary of style options to the bokeh ``ElementPlot.get_data`` method signature and return types. This means that at least the input signature of these methods now matches matplotlib. It's also in many cases much cleaner because a) most ElementPlots need the style as input and b) many also modify the style depending on the element data or parameters, which currently requires subclassing a completely different method. This will make it easier to write a guide on extending HoloViews with new plots and also simplify the process itself.

Concretely the motivation here was to make it easier for me to write some of the statistical plot classes.

This does represent a major change and will force me to pin the latest version of holoviews in the new geoviews release, however in another recent change I already removed the ``empty`` keyword from the signature so that was already necessary.

### DIFF
--- a/holoviews/plotting/bokeh/raster.py
+++ b/holoviews/plotting/bokeh/raster.py
@@ -21,21 +21,13 @@ class RasterPlot(ColorbarPlot):
         if self.hmap.type == Raster:
             self.invert_yaxis = not self.invert_yaxis
 
-
-    def _glyph_properties(self, plot, element, source, ranges):
-        properties = super(RasterPlot, self)._glyph_properties(plot, element,
-                                                               source, ranges)
-        properties = {k: v for k, v in properties.items()}
-        val_dim = [d for d in element.vdims][0]
-        properties['color_mapper'] = self._get_colormapper(val_dim, element, ranges,
-                                                           properties)
-        return properties
-
-
-    def get_data(self, element, ranges=None):
+    def get_data(self, element, ranges, style):
         mapping = dict(image='image', x='x', y='y', dw='dw', dh='dh')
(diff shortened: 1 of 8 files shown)


philippjfr:
Ready to review/merge.

holoviews/plotting/bokeh/raster.py
```diff
@@ -21,21 +21,13 @@ def __init__(self, *args, **kwargs):
...
             self.invert_yaxis = not self.invert_yaxis
 
-
-    def _glyph_properties(self, plot, element, source, ranges):
-        properties = super(RasterPlot, self)._glyph_properties(plot, element,
-                                                               source, ranges)
-        properties = {k: v for k, v in properties.items()}
+    def get_data(self, element, ranges, style):
+        mapping = dict(image='image', x='x', y='y', dw='dw', dh='dh')
         val_dim = [d for d in element.vdims][0]
```
jlstevens: Not part of the PR but ``val_dim = [d for d in element.vdims][0]`` looks off to me. Why not ``val_dim = element.vdims[0]``?

jlstevens:
I have one question above unrelated to the changes in this PR. Otherwise, looks good.

Merging.

holoviews/plotting/bokeh/raster.py
philippjfr: True not sure what the point of that is. Assume it used to grab the dimension name and then changed.

PR #1985, opened 2017-10-09
Add bokeh statistics operations, elements and plots

philippjfr:
This PR will replace the seaborn interface with a set of statistical elements, operations and plots which will all work together. For now it's very much a WIP since there's still some groundwork to be laid, but it's a good start with the basic functionality in place for Distribution and Bivariate elements and corresponding operations.

Demo notebook can be seen [here](https://anaconda.org/philippjfr/statistical_elements/notebook)


holoviews/core/data/array.py
```diff
@@ -72,7 +72,7 @@ def init(cls, eltype, data, kdims, vdims):
 
     @classmethod
     def validate(cls, dataset):
-        ndims = len(dataset.dimensions())
+        ndims = dataset.ndims if dataset._virtual_vdims else len(dataset.dimensions())
```
jlstevens: I assume this is because the 'virtual' vdims are supposed to be hidden?

holoviews/core/data/__init__.py
```diff
@@ -164,6 +164,10 @@ class Dataset(Element):
     # to supplied data
     _auto_indexable_1d = True
 
+    # Determines whether value dimensions are in data or should be emulated
+    # Useful for elements which compute statistics from the data
+    _virtual_vdims = False
```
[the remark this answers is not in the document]
jlstevens: Or maybe these are 'derived' vdims?

holoviews/core/options.py
```diff
@@ -811,16 +811,24 @@ def collapse_element(cls, overlay, key=None, ranges=None, mode='data'):
...
+                result = result.relabel(group=applicable_op.group)
+                overlay = Overlay.from_values(values[:start]+[result]+values[stop:])
+                overlay.id = overlay.id
+            else:
+                values = overlay.items()
+                sliced = overlay.clone(values[start:stop])
+                result = applicable_op.apply(sliced, ranges, key=key)
+                result = result.relabel(group=applicable_op.group)
+                result = list(zip(sliced.keys(), [result]))
+                overlay = overlay.clone(values[:start]+result+values[stop:])
```
jlstevens: Not looked ahead yet - I'm hoping to see some new tests showing this new compositor functionality.

holoviews/core/options.py
```diff
@@ -1106,6 +1114,25 @@ def lookup(cls, backend, obj):
 
 
     @classmethod
+    def transfer_options(cls, obj, new_obj, drop=[]):
+        """
+        Transfers options for all backends from one object to another.
+        Drops any options defined in the supplied drop list.
+        """
```
jlstevens: Seems useful though it definitely needs some examples (even if only in unit tests).

holoviews/element/stats.py
```diff
@@ -0,0 +1,125 @@
+import param
+import numpy as np
+
+from ..core.dimension import Dimension, Dimensioned
+from .chart import Chart, Scatter
+
+
+class _StatisticsElement(Chart):
```
jlstevens: Not sure about the need for an underscore, I don't see why this shouldn't be called ``StatisticalElement`` even if it is a base class. I can imagine ``isinstance(foo, StatisticalElement)`` being useful.

holoviews/operation/stats.py
```diff
@@ -0,0 +1,135 @@
...
+from ..core.dimension import Dimension
+from ..core.operation import Operation
+from ..core.options import Compositor, Store, Options, StoreOptions
+from ..core.util import basestring, find_minmax, cartesian_product
+from ..element import Curve, Area, Image, Polygons
+
+from .element import contours
+
+
+class univariate_kde(Operation):
```
jlstevens: Will need a docstring. I know you were probably already planning to add one - just making sure!

holoviews/plotting/util.py
```diff
@@ -451,6 +453,31 @@ def get_min_distance(element):
     return 0
 
 
+class univariate_composite(Operation):
```
jlstevens: Minor point, I think you could call this ``univariate_compositor``.

holoviews/core/data/dictionary.py
```diff
@@ -101,7 +101,10 @@ def init(cls, eltype, data, kdims, vdims):
 
     @classmethod
     def validate(cls, dataset):
```
jlstevens: Why not make the ``derived_vdims`` flag (or similar) and explicit argument to ``validate``?

holoviews/element/stats.py
```diff
@@ -0,0 +1,125 @@
...
+    overrides standard Dataset methods emulating the existence
+    of the value dimensions.
+    """
+
+    _virtual_vdims = True
+
+    def __init__(self, data, kdims=None, vdims=None, **params):
+        super(_StatisticsElement, self).__init__(data, kdims, vdims, **params)
+        if not self.vdims:
+            self.vdims = [Dimension('Density')]
```
jlstevens: Is 'Density' going to be right for all statistical elements?

holoviews/plotting/util.py
```diff
@@ -451,6 +453,31 @@ def get_min_distance(element):
...
+
+    output_type = Area
+
+    def _process(self, element, key=None):
+        plot_opts = Store.lookup_options(Store.current_backend, element, 'plot').kwargs
+        bw = plot_opts.pop('bw', univariate_kde.bandwidth)
+        transformed = univariate_kde(element, bandwidth=bw)
+        Store.transfer_options(element, transformed, ['bw'])
+        return transformed
```
jlstevens: Transferring styles from input to output seems like it should be an option for the ``Compositor`` when you register the operation.

jlstevens:
Just to say that with a few tweaks discussed with @philippjfr (namely using the ``Compositor`` itself to transfer styles, passing the ``validate_vdims`` flag as an explicit boolean parameter), I am very happy with the approach taken here.

Using the compositor in the background and writing things in terms of operations is very clean and I find it to be a very powerful approach. It is along the lines of the 'plotting macro' idea we had before and this is an initial step in that direction (using the compositor to do that job for now).

Other than seeing those tweaks and the other minor comments above addressed, I would be happy to merge once some unit tests are added (doesn't have to be extensive!) and the corresponding element notebooks are added. I expect coverage would go up without new unit tests just because this would delete that old seaborn/pandas interface code!

holoviews/element/stats.py
philippjfr: Yeah, this logic isn't right at all, will get to it.

holoviews/core/data/dictionary.py
philippjfr: Now removed.

holoviews/plotting/util.py
philippjfr: Do you want that to hold up merging this PR?

holoviews/plotting/util.py
philippjfr: Agreed.

holoviews/core/data/__init__.py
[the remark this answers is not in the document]
philippjfr: Now removed.

philippjfr:
Docstrings, element references and unit tests left, should be able to finish that up in the morning. Compositor now handles everything, and I'm very happy with it overall. It did touch an awful lot of files though.

holoviews/operation/stats.py
philippjfr: Done.

philippjfr:
@jlstevens This is now ready for review. I've added element notebooks for both new Elements along with a new demo, added docstrings to the operations and added unit tests for transferring options, for the statistics elements themselves and for the compositing of the statistics elements.

Should get a very healthy boost in coverage overall.

philippjfr:
Almost a 0.9% coverage increase! Ready for final review and merge whenever.

holoviews/element/stats.py
```diff
@@ -0,0 +1,123 @@
+import param
+import numpy as np
+
+from ..core.dimension import Dimension, process_dimensions
+from ..core.element import Element
+from ..core.util import get_param_values
+from .chart import Chart, Scatter
+
+
+class _StatisticsElement(Chart):
```
jlstevens: I would still prefer to have this called ``StatisticalElement`` which matches the name used in the unit testing class.

holoviews/ipython/display_hooks.py
```diff
@@ -166,9 +167,10 @@ def element_display(element, max_frames):
         IPython.display.display(IPython.display.HTML(info))
         return
 
-
     backend = Store.current_backend
-    if type(element) not in Store.registry[backend]:
+    eltype = type(element)
+    if (eltype not in Store.registry[backend] and
+        all(eltype.__name__ != d.pattern for d in Compositor.definitions)):
```
jlstevens: How come we didn't need this before when the compositor was used?

holoviews/plotting/renderer.py
```diff
@@ -164,10 +164,20 @@ def get_plot(self_or_cls, obj, renderer=None):
...
+                obj = collate(obj)
+                initialize_dynamic(obj)
+            if any(len(c._pattern_spec) > 1 for c in Compositor.definitions):
+                obj = obj.map(lambda obj: Compositor.collapse_element(obj, mode='data',
+                                                                      backend=backend),
+                              [CompositeOverlay])
+            if any(len(c._pattern_spec) == 1 for c in Compositor.definitions):
+                obj = obj.map(lambda obj: Compositor.collapse_element(obj, mode='data',
+                                                                      backend=backend),
+                              [Element])
```
jlstevens: Maybe this stuff should be provided by Compositor itself? (i.e ``CompositeOverlay`` vs ``Element``)

holoviews/core/options.py
```diff
@@ -806,21 +807,45 @@ def strongest_match(cls, overlay, mode):
...
+            else:
+                result = list(zip(sliced.keys(), [result]))
+            overlay = overlay.clone(values[:start]+result+values[stop:])
+
+            # Guard against infinite recursion for no-ops
+            spec_fn = lambda x: not isinstance(x, CompositeOverlay)
+            new_ids = tuple(overlay.traverse(lambda x: id(x), [spec_fn]))
+            if new_ids == prev_ids:
+                return overlay
+            prev_ids = new_ids
```
jlstevens: I am happy that we have ``StatisticalCompositorTest`` but I don't yet have a mental model of how this code extends how Compositor works.

holoviews/core/dimension.py
```diff
@@ -41,6 +41,25 @@ def param_aliases(d):
...
+            continue
+        elif isinstance(dims, (tuple, basestring, Dimension)):
+            dims = [dims]
+        elif not isinstance(dims, list):
+            raise ValueError("%s must be a Dimension or list of dimensions, "
+                             "specified as tuples, string or Dimension instances, "
+                             "not %s." % (group, dims))
+        dimensions[group] = [d if isinstance(d, Dimension) else Dimension(d) for d in dims]
+    return dimensions
+
```
jlstevens: Just checking that this function is confined to ``core``....

holoviews/ipython/display_hooks.py
philippjfr: I can actually remove this again, it was needed when I didn't define plotting class stubs for these elements, which I then restored because I realized options wouldn't work correctly without them.

holoviews/element/stats.py
philippjfr: Sure, don't want that appearing in the top-level namespace though.

holoviews/core/dimension.py
philippjfr: Good thing to check but I'm fairly certain.

holoviews/core/options.py
philippjfr: I'll quickly describe it:

a) Compositors can now be applied to individual elements not just overlays.
b) Compositors are applied iteratively until there are no more matches in the compositor definitions (needed to reduce Overlays with multiple elements that should be transformed)
c) If transfer_options is true, the options are transferred from the input element to the output element. Additionally plot options that also apply to the operation are transferred there.

holoviews/core/options.py
```diff
@@ -778,17 +778,18 @@ class Compositor(param.Parameterized):
...
 
-    group = param.String(doc="""
+    group = param.String(allow_None=True, doc="""
        The group identifier for the output of this particular compositor""")
 
     kwargs = param.Dict(doc="""
        Optional set of parameters to pass to the operation.""")
 
+    transfer_options = param.Boolean(default=False, doc="""
+       Whether to transfer the options from the input to the output.""")
```
jlstevens: I think this is fine but maybe the part that transfers parameters (i.e plot options) to the operation should be a separate flag. Maybe ``transfer_parameters``?

holoviews/element/stats.py
jlstevens: Can't we just set ``__all__`` so if it is used it has to be explicitly imported?

holoviews/ipython/display_hooks.py
jlstevens: Ok, removing this would be good in that case.

holoviews/core/options.py
jlstevens: Ok. I think this is all stuff to demonstrate with examples when we finally expose ``Compositor`` as a useful thing to users.

jlstevens:
After a few more comments above are addressed, I'm happy to merge.

holoviews/plotting/renderer.py
jlstevens: I also think this should be done by compositor as you are accessing an underscore attribute, namely ``_pattern_spec``.

holoviews/element/stats.py
philippjfr: I'm now filtering abstract classes from ``hv.element``. Works fine.

philippjfr:
Okay all comments addressed, just waiting on tests now.

jlstevens:
Tests have passed. Merging!

PR #1988, opened 2017-10-09
Added support for scalars in dictionary interface

philippjfr:
As part of improving support for associating additional information with paths we decided that the dictionary interface should allow scalars as part of the format, avoiding having to duplicate the scalar value for each sample of a path/polygon. This PR makes this possible, while follow up PRs will start adding support for using this to represent scalar values associated with paths.

### DIFF
--- a/holoviews/core/data/dictionary.py
+++ b/holoviews/core/data/dictionary.py
@@ -76,13 +76,14 @@ class DictInterface(Interface):
         if not isinstance(data, cls.types):
             raise ValueError("DictInterface interface couldn't convert data.""")
         elif isinstance(data, dict):
-            unpacked = [(d, np.array(data[d])) for d in data]
-            if not cls.expanded([d[1] for d in unpacked]):
+            unpacked = [(d, vals if np.isscalar(vals) else np.asarray(vals))
+                        for d, vals in data.items()]
+            if not cls.expanded([d[1] for d in unpacked if not np.isscalar(d[1])]):
                 raise ValueError('DictInterface expects data to be of uniform shape.')
             if isinstance(data, odict_types):
                 data.update(unpacked)
             else:
-                data = OrderedDict([(d, np.array(data[d])) for d in dimensions])
+                data = OrderedDict(unpacked)
         return data, {'kdims':kdims, 'vdims':vdims}, {}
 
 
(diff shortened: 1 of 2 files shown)


jlstevens:
Glad to see this coming together!

I think we should work together on the documentation to make sure we convey how all these different threads come together (multi-interface, associated user API that still needs fleshing out, support for constant columns added in this PR). In theory, there could also be plotting code changes e.g supporting changing color along a path but that is less critical.

philippjfr:
There will be a path/polygon/geometry user guide to go alongside the gridded and columnar ones.  Once this PR is merged I can start working on that while I actually start implementing support for plotting scalar and continuously varying variables associated with paths. The user API, which I expect to be fairly minimal additions on top of what we currently have, can follow shortly after that. So the steps from here are:

1) Expand Path/Polygon key dimensions to allow additional scalar variables and add validation to ensure that these variables are indeed scalar.
2) Extend Path/Contour and Polygons plot implementations to support scalar and continuous variables and color the glyph/artists accordingly.
3) Start writing a path/geometry user guide to demonstrate how to construct and work with path data.
4) Decide on and implement any additional user level API necessary for easily working with paths.

philippjfr:
Anyway, this PR is ready to merge once tests pass.

jlstevens:
For your step 2) don't we also want the ``op``/``norm`` system in place first?

philippjfr:
>For your step 2) don't we also want the op/norm system in place first?

Not really, colormapping is never really going to fit into that since that's done client-side.

jlstevens:
I think there are possibilities other than just colormapping - for instance, mapping a vdim to the thickness along the path. Not about how technically feasible that might be but it does fit in conceptually with the improved vdim support for path types.

philippjfr:
>for instance, mapping a vdim to the thickness along the path.

Oh absolutely! There's a bunch of possibilities from line widths to alphas and other things I can't imagine right now, but that can happen independently from the path stuff I described above.

philippjfr:
Ready to merge.

philippjfr:
I think another optional improvement for the dictionary interface will be to allow combining multiple dimensions, letting you do something like this:

```python
{('x', 'y'): np.random.rand(10, 2), 'z': 1}
```

Might be a more natural way to represent paths.

jlstevens:
> ... letting you do something like this:

{('x', 'y'): np.random.rand(10, 2), 'z': 1}

I think we might have discussed something similar in the past (related, if not exactly the same). It seems like a reasonable thing to support at some point.

Anyway, tests have passed and I'll merge now.

philippjfr:
I think we discussed in context of the grid interfaces for the value dimensions, for example for RGBs.

PR #1989, opened 2017-10-09
Fix issue with bokeh legends

philippjfr:
Since bokeh 0.12.7 you can specify a column in the CDS to draw the legend from. This can result in issues like this when the Element ``label`` matches the column name:

screen shot 2017-10-09 at 6 43 41 pm

Thanks to Bryan I found out that "if the legend name is also a column name you need to explicitly pass value(name)", so that's what I've done here.


jlstevens:
Seems fine to merge once the tests pass.

philippjfr:
This PR is now a wholesale rewrite of a lot of the bokeh plotting code. The new capabilities can be seen here: https://anaconda.org/philippjfr/legend/notebook

Improvements include:

* legend_position options have been fixed (and are no longer in weird positions)
* OverlayPlot now has a multiple_legends options which splits the legends
* Find adjustment of legend position can be done using ``legend_offset`` option

philippjfr:
Going to go ahead and merge this, it's a huge improvement in terms of positioning and for actual correctness.

PR #1996, opened 2017-10-11
Ensure histogram operation does not raise deprecation warning

philippjfr:
In a recent PR I added a deprecation warning when constructing a Histogram with two separate arguments for edges and values but overlooked the fact that the histogram operation still does this.


philippjfr:
Ready to merge.

PR #2002, opened 2017-10-17
Improved support for adding hover info via Graph.from_networkx method

philippjfr:
Previously the ``Graph.from_networkx`` method allowed passing a dataset containing hover data which would be used to create the nodes. However the nodes in this dataset would have to match the edges in the supplied networkx graph exactly, which could at times be difficult. This filters the edges and the supplied nodes to use only the subset that are present in both so the user doesn't have to do the filtering.


philippjfr:
Ready to merge.

PR #2007, opened 2017-10-19
Disable deep hashing if memoization is off

philippjfr:
Currently ``Callable`` runs deep hashing on the stream values whether or not memoization is enabled. This PR simply disables it when memoization is off, which has no effect but can be more efficient.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -101,7 +101,7 @@ class HashableJSON(json.JSONEncoder):
         elif isinstance(obj, np.ndarray):
             return obj.tolist()
         if pd and isinstance(obj, (pd.Series, pd.DataFrame)):
-            return repr(sorted(list(obj.to_dict().items())))
+            return obj.to_csv().encode('utf-8')
         elif isinstance(obj, self.string_hashable):
             return str(obj)
         elif isinstance(obj, self.repr_hashable):
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -4,6 +4,8 @@ generate and respond to events, originating either in Python on the
 server-side or in Javascript in the Jupyter notebook (client-side).
 """
 
+import uuid
+
 import param
 import numpy as np
 from numbers import Number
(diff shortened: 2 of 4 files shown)


jlstevens:
Simple change that does make sense.

philippjfr:
>Simple change that does make sense.

Unfortunately it breaks some things. I'll have to think more about it.

philippjfr:
Okay figured it out, the ``memoize`` parameter was being abused by modifying it. I've now added an internal attribute that is used for this purpose instead and the ``Callable.memoize`` is now used solely to control whether memoization is enabled or not.

jlstevens:
Sounds like the right fix. 

That said it might be nice to give the concept a different name than ``self._memoize`` that helps express the other purposes you mention - otherwise it isn't clear why there is an underscore attribute of the same name that is distinct from the parameter.

jlstevens:
How about ``_stream_memoization`` as distinct from the state on the ``Callable``?

Other than that, I'm happy to merge once the tests pass (ideally with some new unit tests!)

holoviews/streams.py
```diff
@@ -404,11 +443,11 @@ class PointerXY(LinkedStream):
     the plot bounds, the position values are set to None.
     """
 
-    x = param.ClassSelector(class_=(Number, util.basestring), default=None,
+    x = param.ClassSelector(class_=(Number, util.basestring, tuple), default=None,
```
philippjfr: Unrelated change that snuck in. Allows the Pointer stream to work with nested categorical axes so worth keeping (and I can't be bothered to make a new PR).

holoviews/core/util.py
```diff
@@ -101,7 +101,7 @@ def default(self, obj):
         elif isinstance(obj, np.ndarray):
             return obj.tolist()
         if pd and isinstance(obj, (pd.Series, pd.DataFrame)):
-            return repr(sorted(list(obj.to_dict().items())))
+            return obj.to_csv().encode('utf-8')
```
philippjfr: A fair bit faster, although still not great, hence also adding a hashkey.

philippjfr:
Ready to merge.

jlstevens:
Looks good. Merging.

PR #2013, opened 2017-10-22
Ensure all clone methods correctly retain id

philippjfr:
Certain clone methods on Annotations, BaseShape and UniformNdMapping were not passing the ``id`` along resulting in styles getting dropped when different operations were applied to these objects. One issue that's caused by this is https://github.com/ioam/holoviews/issues/1904, but there may be others I missed.

### DIFF
--- a/holoviews/core/ndmapping.py
+++ b/holoviews/core/ndmapping.py
@@ -763,7 +763,7 @@ class UniformNdMapping(NdMapping):
             settings = {k: v for k, v in settings.items()
                       if k in new_params}
         settings = dict(settings, **overrides)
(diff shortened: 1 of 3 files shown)


jlstevens:
Well caught. We should keep close tabs on the clone methods that don't go up through super -  ``Annotation``, ``DynamicMap`` and ``BaseShape`` are the ones I remember. We should check to see if there are any more.

philippjfr:
DynamicMap doesn't matter since it does not have any options. The others are covered in this PR. Ready to merge unless you have other suggestions.

jlstevens:

PR #2019, opened 2017-10-23
Fixed empty data issues with multiinterface

philippjfr:
The MultiInterface for Paths was not handling empty data correctly in some cases. This fixes the issues and adds various data interface and plot tests.


philippjfr:
@jlstevens Would be great to merge this asap.

PR #2024, opened 2017-10-26
Fix minor typo

yuvallanger:


jbednar:
Thanks for noticing!  I saw one more typo in that paragraph, so I fixed that too.

PR #2027, opened 2017-10-27
Fixed Area.stack when passing in an NdOverlay

philippjfr:
The Area.stack classmethod provides an easy way to stack an Overlay of Areas, but currently it does not work with NdOverlays. This is an easy fix to address this issue.


philippjfr:
Very simple fix and tests are passing so I'll merge unilaterally.

PR #2028, opened 2017-10-27
Fixed bug in BoxWhisker coloring

philippjfr:
Fixes issues with categorical coloring of BoxWhisker plots.


philippjfr:
Another very simple fix and tests are passing so merging now.

PR #2029, opened 2017-10-27
Suppress warning when data empty in histogram operation

philippjfr:
When the data is empty on a histogram element it currently generates a true_div warning. This suppresses this warning.

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -548,16 +548,17 @@ class histogram(Operation):
         else:
             edges = np.linspace(hist_range[0], hist_range[1], self.p.num_bins + 1)
         normed = False if self.p.mean_weighted and self.p.weight_dimension else self.p.normed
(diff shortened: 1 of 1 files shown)


jbednar:
There's no easy way to check for empty data directly? Seems like that would be clearer and more reliable in the long term (in case the warning message ever changes).

philippjfr:
I assumed there was a reason I wasn't doing that already but it seems it works just fine if I do that. Let's see what the notebook tests say.

jbednar:
Did we decide len() was safe to call on a dask dataframe?

philippjfr:
The data here will always be a numpy array, you can't run np.histogram on other datatypes.

jbednar:
Seems like a better approach to me, then!

philippjfr:
Okay tests are passing and this is definitely a better approach, merging.

PR #2030, opened 2017-10-27
Various fixes and cleanup of bokeh BarsPlot

philippjfr:
The BarsPlot has gone through a lot of iteration due to fundamental changes to the way categoricals are handled in Bokeh. As a long term goal I think it should be rewritten again but for the time being this PR addresses several issues:

* Negative values are stacked correctly
* Stacked bars are colored correctly and stay in their original order
* Avoid warnings that the ColumnDataSource columns are not of matching lengths


philippjfr:
Tests passed, going to merge these fixes.

PR #2031, opened 2017-10-27
Allow setting explicit BoxWhisker range

philippjfr:
Currently there is no way to set the BoxWhisker range at all. While it makes sense that it auto-ranges by default it should respect explicitly set ranges. Partially addresses https://github.com/ioam/holoviews/issues/1937.


philippjfr:
Not much that can go wrong with this PR and tests passed, going to merge.

PR #2032, opened 2017-10-27
Allowed RGB elements to accept a PIL image

jbednar:
HoloViews RGB elements can plot PIL images, but only if you know to convert them to Numpy arrays first using ``numpy.array(pil_image)``.  This PR tries to detect PIL images and convert them automatically so that they can be passed directly to RGB:

image

We don't import PIL to do the check, for two reasons: most users of RGB are unlikely to need PIL, and PIL comes in two flavors, PIL and pillow, which can have strange interactions if both imported.  I *think* this approach is safe, in that this code should rarely be executed if not for actual PIL objects and vastly more rarely could the conversion to Numpy cause a problem in that case not resulting in an exception, but it's possible that I'm overlooking some scenario of importance.


jbednar:
I suppose if we want to be more paranoid we could check a combination of attributes:

```
if hasattr(data,'thumbnail') and hasattr(data,'tobitmap') and hasattr(data,'getpixel') :
```

jlstevens:
Why not attempt a conditional import, then do a proper isinstance check?

jbednar:
See above.

jlstevens:
What is wrong with ``hv.Image(np.array(image))``?

I don't think it is worth having a special case for PIL, even if we could do a robust isinstance check.

jlstevens:
I would be in favor of reverting the code change here and using this PR to add the documentation to show how to work with PIL images by casting to numpy array (you would need some documentation anyway, even if PIL images were supported directly).

jlstevens:
This would make sense as an extra little section in the RGB element notebooks.

jbednar:
I'll close this and open a new PR for the docs changes.

PR #2037, opened 2017-10-27
Removed width/height style opts from TabularPlot

philippjfr:
For some reason width and height style options were defined on bokeh TabularPlot leading to confusion and errors like in this issue: https://github.com/ioam/holoviews/issues/1583


philippjfr:
Tests passed, merging.

PR #2038, opened 2017-10-27
Updated Large Data user guide with datetime support

philippjfr:
As the title says, now that datetimes are supported by Images and the datashader operations the previous workarounds are no longer needed.


philippjfr:
Change to an untested notebook, going to merge without waiting on tests.

PR #2039, opened 2017-10-27
DynamicMap raises KeyError when out of bounds

philippjfr:
As discussed in https://github.com/ioam/holoviews/issues/482 DynamicMap should raise a KeyError not a StopIteration error here.


philippjfr:
Tests passed, merging.

PR #2040, opened 2017-10-28
Add axis handles to OverlayPlot

philippjfr:
The bokeh callbacks sometimes inspect the handles to determine whether to transform the data in some way, e.g. to convert datetimes. On an OverlayPlot this currently does not work because the axis is not defined in the handles.


philippjfr:
Tests are passing, merging.

PR #2042, opened 2017-10-28
Fix renderer issues with different matplotlib backends

philippjfr:
Certain matplotlib backends do not consistently define the renderer object but getting the renderer with ``fig.canvas.get_renderer()`` works consistently.

Fixes: https://github.com/ioam/holoviews/issues/1496


philippjfr:
Tests passed, merging.

PR #2043, opened 2017-10-28
Handle bokeh callback Comm issues

philippjfr:
Ensures that dynamic plots with bokeh callbacks can be instantiated outside a notebook context.


philippjfr:
Tests passed, merging.

PR #2045, opened 2017-10-28
Ensure that key dimensions are hashed correctly

philippjfr:
In an earlier PR I accidentally dropped the DynamicMap key dimensions values from the memoization hash, which meant it never updated. Somehow this wasn't caught at all so I've added some unit tests.


philippjfr:
Tests passing, merging.

PR #2046, opened 2017-10-28
Ensure overlaid BoxWhisker color cycles

philippjfr:
Currently when overlaying a BoxWhisker element the color cycle is lost. This simple looks up the color cycle in a way that works robustly.


philippjfr:
Tests passed, merging.

PR #2047, opened 2017-10-30
Fixed memoization bug when using generator in DynamicMap

philippjfr:
As title says, straightforward bugfix.


philippjfr:
Tests passed, merging.

PR #2048, opened 2017-10-30
Fixed bug in bokeh Graphs when plotting string index

philippjfr:
When the hover tool was enabled and the node indices were not integers Graph nodes would in some cases disappear.


philippjfr:
Tests passing, merging.

PR #2053, opened 2017-10-31
Changed kdims,vdims to positional arguments without lists where possible

jbednar:
Now that all Elements other than Histogram accept positional arguments for kdims and vdims, and accept both lists and single dimensions, changed nearly all examples to use the simplest format feasible.  

Added a bit of explanation of these positional arguments in the getting-started Introduction and in the Annotating user guide, but there's probably more explanation and maybe an example that could be added.

It would be great if someone else could review the diffs, looking for things like using positional arguments for HoloMap and DynamicMap (which I avoided to avoid confusion because they don't have quite the same argument format). None of this has been tested yet, but hopefully the PR tests will catch any errors?

Feel free to push fixes and merge this without me!


jbednar:
Hmm; it looks like the tests didn't even run.  Too many changes for Travis to handle?  Hopefully someone can debug that.

examples/reference/features/bokeh/table_hooks_example.ipynb
```diff
@@ -29,7 +29,7 @@
    "source": [
     "name = ['homepage', 'github', 'chat']\n",
     "link = ['http://holoviews.org', 'https://github.com/ioam/holoviews', 'https://gitter.im/ioam/holoviews']\n",
-    "table = hv.Table({'Name':name, 'Link':link},  vdims = ['Name', 'Link'],  kdims=[])"
+    "table = hv.Table({'Name':name, 'Link':link}, [], ['Name', 'Link'])"
```
jlstevens: I think we should keep the keywords when you need to set an empty list.

examples/user_guide/15-Streaming_Data.ipynb
```diff
@@ -422,7 +422,7 @@
     "multi_source = streamz.dataframe.Random(freq='5ms', interval='100ms')\n",
     "sdf = (multi_source-0.5).cumsum()\n",
     "hv.DynamicMap(hv.Table, streams=[Buffer(sdf.x, length=10)]) +\\\n",
-    "hv.DynamicMap(partial(hv.BoxWhisker, kdims=[], vdims=['x']), streams=[Buffer(sdf.x, length=100)])"
+    "hv.DynamicMap(partial(hv.BoxWhisker, [], 'x'), streams=[Buffer(sdf.x, length=100)])"
```
jlstevens: Another one with an empty list.

jlstevens:
@jbednar Thanks for doing this!

Looking over the diffs, it all *looks* fine apart from the two I mentioned in comments above. Now to look at what is happening on Travis...

jlstevens:
> ...  looking for things like using positional arguments for HoloMap and DynamicMap (which I avoided to avoid confusion because they don't have quite the same argument format).

I think that is fine as  ``HoloMap`` and ``DynamicMap`` only support kdims and I think it isn't a bad idea to be explicit for these container types (especially for ``DynamicMap`` where you often have other keyword arguments such as ``streams``).

jlstevens:
One last point: most of these changes aren't going to be caught by Travis anyway so really a doc build is what will tell us if everything is ok (or at least if there are no errors!)

philippjfr:
Okay I went through it all and it seems fine. As Jean-Luc said, the documentation build is likely to catch errors so I'll have to make sure to run it before release.

PR #2055, opened 2017-10-31
Rewrote SpikesPlot to use segment glyph

philippjfr:
Since bokeh no longer supports plotting ``multi_line`` glyphs on categorical axes this PR switches the Spikes element over to use the ``segment`` glyph which does work.


philippjfr:
Push build passed, merging.

PR #2059, opened 2017-11-01
Improved handling of DynamicMap transform methods

philippjfr:
This PR fixes an issue when applying .map to a DynamicMap where the mapped function transforms the type of element. Previously the mapping function was applied to the DynamicMap cache first and then the function was applied to the DynamicMap clone that had the converted data. This resulted in type mismatches between what was returned by the new DynamicMap callback and what was already in the cache. By simply copying over the cache we avoid this issue and get rid of one additional superfluous level of DynamicMap nesting, resulting in a cleaner graph.


philippjfr:
This may be better solved by integrating it properly with the ``Dynamic`` utility but I can currently not see how to do so properly. I'll probably file an issue about this.

philippjfr:
Tests passing now, and this is an important fix, merging.

PR #2061, opened 2017-11-01
Removed Collector and interface submodule

philippjfr:
We decided a while back that we'd move Collector elsewhere (probably topographica) and since the remaining files in ``interface`` are now gone, now would be a good opportunity to get rid of the entire ``interface`` submodule.


jlstevens:

philippjfr:
Okay, I've now moved it into featuremapper. Topographica was already importing it from there for some reason so it seemed easiest.

philippjfr:
Only a 0.4% gain, still 81% coverage!

philippjfr:
Going to merge, since you already agreed 😄

PR #2062, opened 2017-11-01
Fixed bounding box issues in regrid

philippjfr:
While adding datetime support to regrid I messed with the bounding boxes that were returned. This restores the old behavior while handling datetimes correctly.


jlstevens:
Feel free to go ahead and merge once you think this fix is ready.

philippjfr:
Merging this now, may add more tests and documentation for regrid in another PR before release.

PR #2063, opened 2017-11-01
Fixes for empty or missing frames

philippjfr:
Fixes a number of bugs handling empty frames:

* Fixed updating title when initial frame had no title (Bokeh)
* Ensured that glyphs with a renderer, e.g. Band and Whisker, toggle visibility (bokeh)
* When embedding plot update plot to last frame to ensure all change events are recorded correctly (bokeh)
* Do not adjust legend if the current frame is empty (matplotlib)


philippjfr:
All pretty safe changes, and tests are passing, merging.

PR #2064, opened 2017-11-01
Fixed updating of matplotlib ErrorBars

philippjfr:
It seems some recent version of matplotlib (probably 2.0) changed the API for updating ErrorBars. We should probably pin matplotlib version > 2.0 in the release.


philippjfr:
Tests passing.

PR #2065, opened 2017-11-02
Handle various error conditions in Interface constructor

philippjfr:
Handles some cases in the interfaces that would previously work even though they really shouldn't, includes:

```
Dataset({'x': np.zeros(5)}, kdims=['Test'], vdims=[])

Dataset({'x': 1}, kdims=['Test'], vdims=[])
```

Also handles elements which really only have one dimension, e.g. ``Distribution`` when passing in pandas dataframe with more than 2 dimensions.

```
Distribution(pd.DataFrame({'x': np.zeros(5), 'y': np.zeros(5), 'z': np.zeros(5)})
```

Once this is merged I'm going to make an issue about eventually dropping the dictionary mapping format that has clung on since ``NdElement``, examples include ``{0: 1}, {('A', 'B'): 1}, {('A', 'B'): (1, 2)}``, here both keys and values represent dimension values. It's terrible to validate and just not a great format. The only time it's semi-useful is for specifying HeatMaps.

### DIFF
--- a/holoviews/core/data/dictionary.py
+++ b/holoviews/core/data/dictionary.py
@@ -69,9 +69,16 @@ class DictInterface(Interface):
         elif not any(isinstance(data, tuple(t for t in interface.types if t is not None))
                      for interface in cls.interfaces.values()):
             data = {k: v for k, v in zip(dimensions, zip(*data))}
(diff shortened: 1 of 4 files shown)


philippjfr:
Going to go ahead and merge. I'll open an issue shortly.

PR #2066, opened 2017-11-02
Allow empty data on Bivariate element

philippjfr:
As the title says, this was previously not handled at all, resulting in errors inside scipy.


philippjfr:
PR build passed, merging.

PR #2067, opened 2017-11-02
Fix Bounds bokeh callbacks skipping

philippjfr:
The previous check apparently fails in bokeh 0.12.10, I'll follow up on that but this should work fine.


philippjfr:
Push build passed, merging.

PR #2070, opened 2017-11-02
Ensure Graph node colors persist when updated

philippjfr:
Small fix to ensure the Graph style does not override colormapped columns on Graph.


philippjfr:
Tests passed, merging.

PR #2072, opened 2017-11-02
Made data first argument for both Pipe and Buffer

philippjfr:
After just demoing the streaming API I realized that Pipe and Buffer constructors are inconsistent. This PR ensures that both accept the initial data as the first argument.


philippjfr:
Tests passed and the notebook doesn't need the updates, merging.

PR #2073, opened 2017-11-02
Fixed issues updating categorical axes and mappers

philippjfr:
Ensures that categorical axes are updated before the ColumnDataSource, avoiding errors in https://github.com/bokeh/bokeh/issues/7084. Also ensures that the colormapper correctly updates depending on the type of the data.


philippjfr:
PR build passed, and it's up to date so I'll merge.

PR #2074, opened 2017-11-02
Improved pandas dim handling further

philippjfr:
One more condition for pandas dataframe dimension extraction that I didn't handle correctly, including a test.


philippjfr:
I had to add a workaround that's only necessary because we haven't made this change https://github.com/ioam/holoviews/issues/1668 yet, should prioritize that in the next release.

philippjfr:
Tests passing, merging.

PR #2076, opened 2017-11-02
Fixed Surface plot without cmap

philippjfr:
Surface now errors when you try to add a colorbar with having a cmap.


philippjfr:
Push build is passing, merging.

PR #2078, opened 2017-11-02
Changed datashader default colormap to fit HV defaults

jbednar:
Currently, ``holoviews.operation.datashade.shade()`` and  ``holoviews.operation.datashade.datashade()`` use a default colormap of ``fire``, originally chosen to match the default Image colormap in HoloViews.  This made sense at first, because datashader produces images.  However, a datashader plot is rarely a dense image covering every pixel; instead it tends to have many transparent pixels, and is thus much more like a line or points plot in HoloViews.  When using ``datashade()`` on an ``hv.Points()``, etc. object using the HV/Jupyter/browser default white page background, a fire colormap is inappropriate, because the highest value maps to the page color.  This resulted in bits of datashader plots disappearing, which appeared to be a bug in datashader but was simply the result of using a colormap ending in white on a white page.  These bits sometimes disappear in the initial rendering, if there happen to be isolated pixels at the highest strength, or more often when zooming in far enough that a single color becomes visible, at which point the entire plot disappears.

This PR selects a colormap designed to be thematically compatible with the HV default blue color for lines and points:

image

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -497,10 +497,13 @@ class shade(Operation):
     Iterable or a Callable.
     """
 
-    cmap = param.ClassSelector(default=["deepskyblue","black"], class_=(Iterable, Callable, dict), doc="""
-        Iterable or callable which returns colors as hex colors, to
-        be used for the colormap of single-layer datashader output. 
-        Callable type must allow mapping colors between 0 and 1.""")
+    cmap = param.ClassSelector(class_=(Iterable, Callable, dict), doc="""
+        Iterable or callable which returns colors as hex colors
+        or web color names (as defined by datashader), to be used
+        for the colormap of single-layer datashader output. 
+        Callable type must allow mapping colors between 0 and 1.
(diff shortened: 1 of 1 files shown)


jlstevens:
@philippjfr @jbednar I'm ok with this as long as you are!

If we merge this PR now, I wouldn't want to change the colormap user here till 2.0. Using ``["deepskyblue","black"]`` fixes the issue Jim described but it is also a little simplistic -
 compared to the effort that went to fire for instance!

jbednar:
The fire colormap is excellent for black backgrounds, but I am not aware of a colormap for white backgrounds that is both perceptually intuitive (where increasing values are interpreted unambiguously by most people), and has high dynamic range.  Now reverted to the original default of None, which uses the default Datashader colormap of ["lightblue","darkblue"], which is always ok against a white background and is thematically compatible with HoloViews. Seems like the reasonable default choice, and restores behavior from HV 1.8.3, I believe.

philippjfr:
Okay, these transients are getting out of hand but the push build passed, so I'm merging.

PR #2080, opened 2017-11-02
Removed NdMapping cached variables

philippjfr:
All of these cached variables are completely pointless, we removed features that relied on them and now they are causing issues with DynamicMap changes I made in https://github.com/ioam/holoviews/pull/2059.


jlstevens:
Well, I certainly support removing code if it fixes bugs and everything else still works!

philippjfr:
Okay, tests passing. This should be the last actual code change in 1.9.

PR #2081, opened 2017-11-03
Added CSS class to logo block

philippjfr:
Adds a CSS class to the logo in the extension so it can be styled separately in the doc build.

### DIFF
--- a/holoviews/ipython/load_notebook.html
+++ b/holoviews/ipython/load_notebook.html
@@ -3,7 +3,7 @@
 {{ widgetcss }}
 
 {% if logo %}
-<div>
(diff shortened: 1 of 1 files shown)


philippjfr:
Okay I've confirmed this works and doesn't affect the tests so I'm merging.

jlstevens:
Could you paste the CSS rule here you used for 'logo-block' for future reference and in case anyone else needs it?

philippjfr:
```css
div.logo-block {
	display: inline-block;
}
```

jlstevens:
Thanks!

PR #2082, opened 2017-11-03
Ensure HeatMap correctly constructs from pandas

philippjfr:
Since ``HeatMap`` didn't define a dimension bounds the ``PandasInterface`` was incorrectly handling the kdims and vdims. So I've defined the ``bounds`` on the kdims and ensured the PandasInterface correctly infers dimensions for an Element that doesn't define the kdims bounds.


philippjfr:
Tests passe so I'll merge.

PR #2083, opened 2017-11-03
Fixed bug when raising Dimension error

philippjfr:
This error in Dimension previously raised errors about not all string arguments being converted because ``spec`` could be a tuple, to fix it I've simply cast the variable to a string.


philippjfr:
Tests passing, merging.

PR #2084, opened 2017-11-03
Fixed issues with Dataset conversions for stats elements

philippjfr:
The data conversion methods were still trying to import the elements from seaborn. Additionally the elements pose some challenges for the conversion interfaces in general so the interfaces really benefit from some additional strictness about the dimensions it tries to infer. That change only catches conditions which would otherwise have errored.

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -90,14 +90,22 @@ class DataConversion(object):
                                       "after version 1.7.")
                 groupby = kwargs.pop('mdims')
 
+        element_params = new_type.params()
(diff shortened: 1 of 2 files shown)


philippjfr:
Tests passing, merging. This is definitely the last PR to go into 1.9.0.

jlstevens:
Glad that this was caught! 

All those inline imports would have failed and it isn't complete support without the conversion interface. Partially my fault for not thinking about this when you removed the interface module - I'm guessing you have now grepped to check for any other imports that will no longer work...

PR #2085, opened 2017-11-03
Small updates for stats elements

philippjfr:
Some minor things that I forgot to update when adding the stats elements.


philippjfr:
Can't affect the tests, merging.

PR #2088, opened 2017-11-03
Fix plotting of empty image, closes #2087

### LINKED ISSUE #2087: Unable to plot empty Image or Raster
```python
hv.Image([])
```

This is possible with a lot of other `holoviews` plot objects.

basnijholt:


jbednar:
From Philipp: "if not data" won't work on an array.

philippjfr:
Is this definitely safe? I expect ``if not data``to error on an array. Looking on my phone though so may be missing something.

basnijholt:
You might be right. 

I changed it back to `if data is None`.

Now it will only work in the case of `hv.Image(None)`.

Do you have any suggestions how to do a more robust check?

basnijholt:
`if np.asarray(data).size == 0` would probably work, but I don't know whether you consider this too expensive?

philippjfr:
The specific case you're trying to handle here are empty arrays and empty lists right? I'd probably just go with something like this:

``python
if (isinstance(data, list) and not data) or (isinstance(data, np.ndarray) and np.product(data.shape) == 0):
    ...
```

**Comment by basnijholt** [comment]
@philippjfr thanks!
I went with:
```python 
        if (data is None
            or (isinstance(data, (list, tuple)) and not data)
            or (isinstance(data, np.ndarray) and data.size == 0)):
            data = np.zeros((2, 2))
```

not sure about the check for tuples, what do you think?

philippjfr:
I think that's very reasonable. Handling tuples (and xarray) like this may be too much and probably needs a different solution anyway. I'm +1 on merging this.

jlstevens:
Also seems like a good idea to me - this can be the first PR of 1.9.0 dev. :-)

philippjfr:
Thanks again for the contribution, too bad this didn't make it into 1.9.0 but I'm sure there will be a 1.9.1 in the not too distant future.

Issue #1869, opened 2017-09-12
Image should warn about unevenly spaced dimensions


rabernat:
Gridded scientific datasets commonly have dimensions that are unevenly spaced. (For example, an ocean temperature dataset has more vertical levels near the surface than in the deep.) We can easily visualize these datasets correctly with matlplotlib using a pcolormesh or contour plot. But I am unsure about the status of support for such datasets in holoviews. The documentation mentions support for [irregularly sampled continuous coordinates](http://holoviews.org/user_guide/Continuous_Coordinates.html), but I couldn't find a clear example.

Here is an example of what I mean using xarray. Consider the following dataset
```python
import xarray as xr
import numpy as np
# monotonically increasing but unevenly spaced coordinates
x = np.arange(20)**2
y = np.arange(10)**2
z = y[:,np.newaxis] * x
data = xr.DataArray(z,
            dims=['y', 'x'],
            coords={'y': y, 'x': x},
            name='randomdata')
data.plot()
```
xarray renders this correctly, with grid boxes of variable size
![image](https://user-images.githubusercontent.com/1197350/30306295-5d11391a-9745-11e7-8284-c3e51c2332e7.png)

However, in holoviews, the irregular spacing seems to be lost when I convert to image
```python
import holoviews as hv
%%opts Image (cmap='viridis')
hv.Image(data)
```
![image](https://user-images.githubusercontent.com/1197350/30306347-9b366ed6-9745-11e7-9034-cdedc112cb47.png)
all the grid boxes are the same size.

Am I missing something obvious?

jbednar:
Try [hv.QuadMesh](http://holoviews.org/reference/elements/bokeh/QuadMesh.html)?

rabernat:
Great @jbednar, `hv.QuadMesh((x, y, z[:-1,:-1]))` works! I would say that qualifies as "something obvious"

From a user perspective, coming from xarray to holoviews, I guess I would have liked to get some kind of error or warning when trying to create an image with unevenly spaced dimensions.

jbednar:
No problem. That does look like a case we should warn, which is even clearer if you increase the nonlinearity in your example (e.g. ``**7`` instead of ``**2``).

philippjfr:
Repurposed this issue to add such a warning, the warning/error should presumably also mention QuadMesh.

jbednar:
[...]

PR #2093, opened 2017-11-05
Enable compute_edges with unequally spaced data

mforbes:
This is a simple modification that allows one to use `QuadMesh` etc. with unevenly spaced points without having to explicitly provide the bin boundaries.  This simply uses the midpoints as bin boundaries and sets the endpoints as the midpoints of the outer cells.  This thus maintains the previous behaviour when equally-spaced points are provided, and now simply allows for unevenly spaced points.  (See also issue #1869.)

### DIFF
--- a/holoviews/element/util.py
+++ b/holoviews/element/util.py
@@ -29,20 +29,16 @@ except:
 
 def compute_edges(edges):
     """
-    Computes edges from a number of bin centers,
-    throwing an exception if the edges are not
-    evenly spaced.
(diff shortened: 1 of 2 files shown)


jlstevens:
Sounds great! I'll assign Philipp as a reviewer as he is the one who'll be able to reason about how your changes might affect other things (e.g casting between element types).

philippjfr:
I'm fine with making this change. It shouldn't affect anything else in terms of casting of elements and is usually a safe enough thing to do.

PR #2103, opened 2017-11-07
Improvements for kde operation error, dimension and range handling

philippjfr:
There were some bugs in the univariate_kde dimension handling but I've also included improvements to range and exception handling for both kde operations.

I'll add some more unit tests before this is ready to merge.


philippjfr:
Have added tests now, merging.

PR #2110, opened 2017-11-09
Bokeh plotting fixes

philippjfr:
A number of smaller fixes/enhancements for bokeh plots:

* CompositeElementPlot now allows drawing multiple glyphs from the same datasource
* CompositeElementPlot no longer errors when a plot doesn't return data for a particular glyph (fixes http://holoviews.org/gallery/demos/bokeh/dropdown_economic.html)
* BoxWhisker and BarPlot use regular Range1d models not DataRange1d, which has changed behavior in 0.12.11dev and no longer updates as one would expect (also DataRange1d is not actually required anymore)


philippjfr:
All safe fixes, merging.

PR #2112, opened 2017-11-10
Fixed throttling on bokeh server

philippjfr:
This PR fixes two bugs:

1. A left over and undefined ``bokeh_version`` variable causing errors
2. The previous approach to server side throttling never really worked and this is a huge improvement.


philippjfr:
All very straightforward fixes and addresses a major issue. Ideally we'd add bokeh server tests but not straight away. Going to merge.

PR #2113, opened 2017-11-10
Fix bokeh version import

suvarchal:
fixes bokeh_version import otherwise leads to

`  line 422, in set_server_callback
   if self.on_events and bokeh_version >= '0.12.5':
NameError: name 'bokeh_version' is not defined`

atleast while serving a bokeh app.


philippjfr:
Thanks for catching this, I've already made the fix in https://github.com/ioam/holoviews/pull/2112. This check is no longer needed, so I've just removed it. We'll likely merge the fix tomorrow and then release 1.9.1 with the fix on Sunday.

suvarchal:
thanks! my baby steps to atleast make some contribution.

philippjfr:
Your contribution was definitely appreciated, and really it's on us to provide some material for contributors to get started!

PR #2114, opened 2017-11-11
Add Violin Element

philippjfr:
Since we now have a univariate_kde operation implementing a Violin plot was pretty easy.

Here's what the element can do, it acts much like BoxWhisker: https://anaconda.org/philippjfr/violin/notebook

bokeh_plot

image

- [x] Added matplotlib plotting tests
- [x] Added bokeh plotting tests
- [x] Added Element reference notebooks
- [x] Added gallery examples


jlstevens:
Very happy to see this element added for the next release.

Other than the obvious merge conflict that needs resolving, I just want to add a reminder about that streams + compositor issue (as these stats elements use the compositor afaik). Just something extra to make sure is working as the compositor will need a bit more work to handle streams properly.

philippjfr:
This particular element does not use compositors. I think this is mostly ready to merge (after resolving merge conflicts). I have improvements to coloring the violins in mind but they will require some work in bokeh to allow ``multi_line`` and ``patches`` glyphs on categorical axes again.

philippjfr:
Ready to review and merge.

holoviews/element/chart.py
```diff
@@ -340,3 +319,20 @@ def stack(cls, areas):
...
         return stacked
+
+
+class BoxWhisker(Chart):
+    """
+    BoxWhisker represent data as a distributions highlighting the
+    median, mean and various percentiles. It may have a single value
+    dimension and any number of key dimensions declaring the grouping
+    of each violin.
+    """
```
jlstevens: I assume there a reason ``BoxWhisker`` can't also go into ``element/stats.py``?

holoviews/operation/stats.py
```diff
@@ -16,7 +16,7 @@ def _kde_support(bin_range, bw, gridsize, cut, clip):
     if clip[0] is not None and np.isfinite(clip[0]):
         kmin = max(kmin, clip[0])
     if clip[1] is not None and np.isfinite(clip[1]):
-        kmax = max(kmax, clip[1])
+        kmax = min(kmax, clip[1])
```
jlstevens: Given this function existed before this PR, what might have been affected by this bug? (if anything?)

holoviews/element/chart.py
philippjfr: Yes, I initially moved it there, but it breaks unpickling of our test data and I can't move it and import it into charts because that would be circular.

holoviews/operation/stats.py
philippjfr: ``Distribution`` elements would have been clipped incorrectly but I think clipping is disabled by default so not all that much would have changed.

holoviews/plotting/bokeh/stats.py
```diff
@@ -35,3 +47,377 @@ class BivariatePlot(PolygonPlot):
...
+    bandwidth = param.Number(default=None, doc="""
+        Allows supplying explicit bandwidth value rather than relying
+        on scott or silverman method.""")
+
+    clip = param.Boolean(default=True, doc="""
+        Whether to clip the violins at the max and min of the data.""")
+
+    cut = param.Number(default=5, doc="""
+        Draw the estimate to cut * bw from the extreme data points.""")
+
```
jlstevens: Looks like the matplotlib implementation doesn't have the ``clip`` and ``cut`` plot options?

holoviews/element/chart.py
jlstevens: Ok. Should this be marked as something to fix in version 2.0 when we can break backwards compatibility?

philippjfr: Yes, probably.

holoviews/plotting/bokeh/stats.py
philippjfr: Right, the matplotlib violin implementation doesn't expose it for some reason, it just always clips.

jlstevens: Ok. That is fine as long as these defaults match the behavior of the matplotlib version.

philippjfr: It does.

holoviews/element/chart.py
philippjfr: Should I just file an issue once this is merged?

philippjfr:
I think this is ready to merge.

jlstevens:
You replied to all my queries, thanks!

Looks good and tests are passing. Merging.

PR #2117, opened 2017-11-12
Image grid fixes

philippjfr:
Fixes for https://github.com/ioam/holoviews/issues/2094, the ImageInterface was not validating that the Image coordinates matched the array shape and the regrid operation was not ensuring the array had the correct orientation before regridding.


philippjfr:
Two important fixes and tests are passing. Merging.

PR #2120, opened 2017-11-12
Improved streaming updates on overlaid plots

philippjfr:
Previously the stream handling code in the bokeh backend was not correctly handling a condition where multiple subplots were attached to different streaming sources. In those cases it would fall back to sending a full update of the data for some subplots. It now appropriately skips updates to a streaming plot when it isn't being triggered.

Additionally it now forces the ranges to be updated in streaming mode, which previously caused issues when running a plot reset.


jlstevens:
Looks good. Merging.

PR #2122, opened 2017-11-13
Updates and fixes to user guides and demos

philippjfr:
Fixed some issues in the dashboarding user guide and used tornado PeriodicCallback for asynchronous streaming data example.

### DIFF
--- a/examples/user_guide/15-Streaming_Data.ipynb
+++ b/examples/user_guide/15-Streaming_Data.ipynb
@@ -558,7 +558,7 @@
    "source": [
     "## Asynchronous updates using the tornado ``IOLoop``\n",
     "\n",
-    "In most cases, instead of pushing updates manually from the same Python process, you'll want the object to update asynchronously as new data arrives. Since both Jupyter and Bokeh server run on [tornado](http://www.tornadoweb.org/en/stable/), we can use the tornado ``IOLoop`` in both cases to define a non-blocking co-routine that can push data to our stream whenever it is ready."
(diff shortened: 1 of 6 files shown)


jlstevens:
I know you might still be working on this but I noticed that this is now in the guide ``renderer._figure_data``. If people are going to use this, do we need a suitable public interface?

philippjfr:
>If people are going to use this, do we need a suitable public interface?

Yes probably.

examples/user_guide/15-Streaming_Data.ipynb
```diff
@@ -568,29 +568,51 @@
...
     "buffer = Buffer(np.zeros((0, 2)), length=50)\n",
     "\n",
     "@gen.coroutine\n",
     "def f():\n",
-    "    for x in range(500):\n",
-    "        gen.sleep(0.1)\n",
-    "        buffer.send(np.array([[x, np.random.rand()]]))\n",
-    "        \n",
-    "IOLoop.current().add_callback(f)\n",
+    "    global count\n",
```
jlstevens: Globals are the best thing to use here?

jlstevens:
Looks fine though I would rather avoid that ``global`` if possible and we need a better API than ``_figure_data``. If there is an issue to remind us to improve the latter, I am happy to see this PR merged though I would be even happier if we could get rid of what looks like internal API calls before the release.

examples/user_guide/15-Streaming_Data.ipynb
philippjfr: Coroutines don't allow generators so that's not an option, open to other suggestions though.

philippjfr:
>If there is an issue to remind us to improve the latter, I am happy to see this PR merged though I would be even happier if we could get rid of what looks like internal API calls before the release.

Not sure I want to add new API in this release and since this is just fixing broken docs I think we should merge and open an issue to discuss new API. I think there's more discussion to be had about the ``Renderer`` API anyway, particularly as it relates to bokeh server and perhaps providing some analogue to a ``show`` function.

jlstevens:
Sounds like a plan. In that case, please make sure that issue exists, mentioning ``_figure_data`` and opening up discussion regarding the renderer API changes you would like to see.

philippjfr:
Done in #2191.

jlstevens:
Thanks. Otherwise the PR looks good.

Merging.

PR #2133, opened 2017-11-17
Fixed yrotation and zrotation for mpl overlay plots.

ceball:
Maybe I shouldn't be trying to rotate axis labels, but I was, and I got quite confused...


philippjfr:
Thanks, clearly a bug.

PR #2138, opened 2017-11-18
Add Chord element

philippjfr:
As outlined in https://github.com/ioam/holoviews/issues/2137 this PR adds support for a ``Chord`` element. So far this is just a WIP, but basically this just extends the existing ``Graph`` element to lay things out in a circle and draw bezier splines between the nodes. The next step will be to allow scaling the line thickness by a value in a separate PR, which will allow scaling the chords representing the edges between nodes and then start drawing the bands around the circle.

Progress for what is supported can be tracked in: https://anaconda.org/philippjfr/chord/notebook

### DIFF
--- a/examples/reference/elements/bokeh/Chord.ipynb
+++ b/examples/reference/elements/bokeh/Chord.ipynb
@@ -48,7 +48,22 @@
    "outputs": [],
    "source": [
     "links = pd.DataFrame(data['links'])\n",
-    "print(links.head(3))\n",
+    "print(links.head(3))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In the simplest case we can construct the ``Chord`` by passing it just the edges"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
     "hv.Chord(links)"
    ]
   },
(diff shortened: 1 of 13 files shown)


jbednar:
Nice!  We sure need the automatic viewport padding...

philippjfr:
Since a Chord plot is always drawn on a unit circle I've now added some default padding which can be adjusted by using the ``.redim.range`` to adjust the range of the node x and y dimensions. This means it now provides reasonable defaults, while still letting you override them (the bkcharts version hardcoded the ranges), here's the default viewport of a full ChordPlot with labels:

bokeh_plot

philippjfr:
This PR is now feature complete, and behaves how I think it should. The upcoming 0.12.11 bokeh release will fix a bug updating the colormapping of the edges, which shouldn't be an issue as that will be released long before our next release. It's now just missing reference doc entries and plenty of unit tests.

philippjfr:
Ready to review and merge.

examples/reference/elements/bokeh/Chord.ipynb
```diff
@@ -0,0 +1,98 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "links = pd.DataFrame(data['links'])\n",
+    "print(links.head(3))\n",
```
jlstevens: I would prefer to see this cell split:

```python
links = pd.DataFrame(data['links'])
links.head(3)
```

Then:

```python
hv.Chord(links)
```

I also notice the kernel metadata is still present in the notebooks. Could you please clear that out?

philippjfr: Sure.

holoviews/element/util.py
```diff
@@ -235,6 +235,23 @@ def circular_layout(nodes):
     return (x, y, nodes)
 
 
+def quadratic_bezier(start, end, c0=(0, 0), c1=(0, 0), steps=50):
```
jlstevens: Ideally this should be handled by the plotting extension. The main advantage I see of us doing this is that it helps consistency between matplotlib/bokeh. That said, plotting libraries might have more performant/robust code (though this is using numpy so it shouldn't be too slow).

tests/plotting/matplotlib/testplot.py
```diff
@@ -9,6 +9,8 @@
 from holoviews.plotting import comms
 
 try:
+    from matplotlib import pyplot
+    pyplot.switch_backend('agg')
```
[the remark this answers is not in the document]
philippjfr: Just part of shuffling tests around.

holoviews/element/util.py
jlstevens: That would be nice. I might actually prefer holoviews handling this if the available bezier definitions in matplotlib/bokeh don't match up.

philippjfr: Afaik they do match.

philippjfr:
Made the requested change to the notebooks and resolved the merge conflicts. Ready to merge when tests pass.

jlstevens:
Looks good. Merging.

PR #2140, opened 2017-11-21
Fixes and improvements for CompositeElementPlot

philippjfr:
Fixes bug when using glyphs containing an underscore and allows defining an explicit draw order for glyphs.


philippjfr:
Going ahead and merging this since it's mostly a bug fix and other work depends on it.

PR #2143, opened 2017-11-22
Add TriMesh element

philippjfr:
Triangle meshes are a common form of data when working with complex polygons and are frequently employed in environmental modeling. The most common way to compute such a mesh is using Delaunay triangulation and the data is usually represented as two data structures:

1. The **simplices** representing each triangle, usually these are node indices.
2. The **nodes** or points representing each corner of a triangle.

This representation is in fact simply a specific type of graph and closely follow the data structures we already use for graphs, where the **simplices** represent the abstract connectivity of the mesh and the nodes represent the positions. It was therefore trivial to write a ``TriMesh`` element which simply reuses the same data structures, organization, and plotting code as existing graph elements.

Here is a simple example:

```python
import param
import numpy as np
import holoviews as hv
from scipy.spatial import Delaunay

from holoviews.element.graphs import Graph, EdgePaths

hv.extension('bokeh')

n_verts = 1000
pts = np.random.randint(1, n_verts, (n_verts, 2))
tris = Delaunay(pts)

hv.TriMesh((tris.simplices, tris.points))
```

bokeh_plot

This approach is reasonably fast for small meshes (~1 second/5000 triangles) and once you start plotting more than ~10k triangles you will want to use datashader anyway. It is also flexible enough to associate additional values both with the simplices and with the nodes.

Once https://github.com/bokeh/datashader/pull/525 is merged I will get on with allowing datashader operation such as aggregate to operate on the TriMesh element.

- [x] Reference gallery entries
- [x] Unit tests


philippjfr:
If you enable WebGL rendering in bokeh you can actually scale up to some pretty large meshes, here is an example of selecting on 40485 triangles:

screen shot 2017-11-23 at 12 38 42 am

philippjfr:
I've now allowed filling the triangles by a value, so here's the example above this time with filled triangles:

screen shot 2017-11-24 at 4 47 54 pm

I'd now consider this PR ready for review, I've added unit tests and reference notebooks. The datashading portion will follow in another PR once it's been merged into datashader.

philippjfr:
Yay, we even gained coverage.

basnijholt:
I am trying this out since I work a lot with triangulations, currently, I create plots like:
plot
(an Overlay of Polygon and Image)

Using this branch I am trying to get something similar, but it's not getting better:
```python
%%output size=200
%%opts TriMesh (node_size=0 edge_line_width=0.1 edge_nonselection_alpha=0.01)
plot * hv.TriMesh((tris.simplices, points))
```
mr

I am unable to set the `alpha` and `linewidth`, am I not understanding it, or is this a bug?

basnijholt:
Another question, the `hv.Image` I used in the `Overlay`s above are created from an interpolation that uses the `scipy.spatial.Delaunay` object, but looking at your example plot it seems to be possible to fill the triangles directly with the values at the vertices, but I can't see how exactly. Could you give a simple example?

philippjfr:
Looks like you're using the matplotlib backend but specifying bokeh options, try changing that to ``edge_linewidth`` and ``edge_alpha``.

>but looking at your example plot it seems to be possible to fill the triangles directly with the values at the vertices, but I can't see how exactly.

Have a look at [the example notebook](https://github.com/ioam/holoviews/blob/aac7fc8bac8611fe645b99e624432eb919a80e07/examples/reference/elements/matplotlib/TriMesh.ipynb), the last example demonstrates it. Going to be pushing a small fix for that shortly.

philippjfr:
@jlstevens @jbednar Requesting review.

holoviews/operation/datashader.py
```diff
@@ -478,6 +486,93 @@ def _process(self, element, key=None):
...
+
+        interpolate = bool(self.p.interpolation)
+        agg = cvs.trimesh(pts, simplices, agg=self.p.aggregator,
+                          interp=interpolate)
+        params = dict(get_param_values(element), kdims=[x, y],
+                      datatype=['xarray'], vdims=[vdim])
+        return Image(agg, **params)
+
+
+class rasterize(trimesh_rasterize):
```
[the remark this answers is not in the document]
philippjfr: True, it was easiest for inherit all the methods and parameters but a cleaner class hierarchy is worth the extra handling.

holoviews/operation/datashader.py
```diff
@@ -478,6 +486,93 @@ def _process(self, element, key=None):
...
+                               issubclass(x.type, Dataset)
+                               and not issubclass(x.type, Image))
+        element = element.map(dsrasterize, predicate)
+
+        # Rasterize other Dataset types
+        predicate = lambda x: (isinstance(x, Dataset) and
+                               (not isinstance(x, Image) or x in imgs))
+        element = element.map(dsrasterize, predicate)
+        return element
+
```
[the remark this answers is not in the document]
philippjfr: I was going to handle that in another PR once datashader as made the API more consistent.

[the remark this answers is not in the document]
jbednar: Ok. And that's on my plate, but after this conference prep...

basnijholt:
How can I use the interpolation here?


How do I go from:
```python
%%output size=200
%%opts TriMesh [filled=True edge_color_index='z' fig_size=200 ] (cmap='viridis' node_alpha=0 edge_line_alpha=0.4)
def plot_TriMesh(learner):
    ip = learner.ip()
    simplices = ip.tri.simplices
    nodes = ip.tri.points
    z = ip.values[simplices].mean(axis=1)
    return hv.TriMesh((np.column_stack([simplices, z]), nodes), vdims='z')
plot_TriMesh(learner)
```
trimesh

to

```
%%output size=200
%%opts Image [fig_size=200 ] (cmap='viridis')
%%opts Contours (color='k')
learner.plot(triangles_alpha=0.4)
```

adaptive


I would like to know because:
```python
%timeit plot_TriMesh(learner)
2.71 ms ± 309 µs per loop (mean ± std. dev. of 7 runs, 100 loops each)

%timeit learner.plot(triangles_alpha=True)
216 ms ± 13.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

basnijholt:
For the same reason `hv.Image([])` is useful, the same would go for `hv.TriMesh`.

Would it be easy to make this work with `hv.TriMesh(([], []))`or something?

My data is calculated and plotted live, and in be beginning there might not be data to plot yet.

philippjfr:
@basnijholt Both ``hv.TriMesh(([], []))`` and ``hv.TriMesh([])`` now work.

basnijholt:
How can I change the colors of the edges?

(I think) I've tried all style parameters with `*_color*` but I am not able to get it to work.

philippjfr:
@basnijholt Are your values on the vertices or simplexes? Currently it only allows coloring by values on the simplexes.

philippjfr:
@jlstevens Requesting review again.

basnijholt:
> Are your values on the vertices or simplexes

The values are at the vertices.

holoviews/operation/datashader.py
```diff
@@ -478,6 +487,103 @@ def _process(self, element, key=None):
...
+                      datatype=['xarray'], vdims=[vdim])
+        return Image(agg, **params)
+
+
+
+class rasterize(ResamplingOperation):
+    """
+    Rasterize is a high-level operation which will rasterize any
+    Element or combination of Elements supplied as an (Nd)Overlay by
+    aggregating with the supplied aggregation it with the declared
```
philippjfr: Need to fix this docstring.

holoviews/element/graphs.py
```diff
@@ -392,3 +392,119 @@ class EdgePaths(Path):
...
+        """
+        Allows selecting data by the slices, sets and scalar values
+        along a particular dimension. The indices should be supplied as
+        keywords mapping between the selected dimension and
+        value. Additionally selection_specs (taking the form of a list
+        of type.group.label strings, types or functions) may be
+        supplied, which will ensure the selection is only applied if the
+        specs match the selected object.
+        """
+        self.edgepaths
```
philippjfr: Will add a comment about this. By default ``edgepaths`` are not computed but .select expects it to be.

philippjfr:
>The values are at the vertices.

Okay, I think we could allow coloring by the mean value of the vertices. Interpolation would be nice to support as well but I currently don't see any way to do that in bokeh.

holoviews/element/graphs.py
```diff
@@ -392,3 +392,119 @@ class EdgePaths(Path):
...
+                                     "data with two or three columns representing the "
+                                     "x/y positions and optionally the node indices.")
+        if edgepaths is not None and not isinstance(edgepaths, self._edge_type):
+            edgepaths = self._edge_type(edgepaths)
+        super(TriMesh, self).__init__(edges, kdims=kdims, vdims=vdims, **params)
+        self._nodes = nodes
+        self._edgepaths = edgepaths
+
+    @classmethod
+    def from_points(cls, points):
```
jlstevens: Shouldn't this be ``from_nodes`` or ``from_vertices``?

``from_points`` doesn't sound like graph terminology...

holoviews/element/graphs.py
jlstevens: Yes, a comment next to this line explaining it would be a good idea.

holoviews/plotting/bokeh/__init__.py
```diff
@@ -197,6 +199,15 @@ def colormap_generator(palette):
...
                         edge_nonselection_line_color='black',
                         edge_hover_line_color='limegreen')
+options.TriMesh = Options('style', node_size=5, node_line_color='black',
+                          node_fill_color='white', edge_line_color='black',
+                          node_hover_fill_color='limegreen',
+                          edge_hover_line_color='limegreen',
+                          edge_nonselection_alpha=0.2,
+                          edge_nonselection_line_color='black',
+                          node_nonselection_alpha=0.2,
+                          edge_line_width=1)
```
jlstevens: That is a lot of options!

holoviews/plotting/bokeh/graphs.py
```diff
@@ -33,10 +33,20 @@ class GraphPlot(CompositeElementPlot, ColorbarPlot, LegendPlot):
...
+                  ['node_size', 'cmap', 'edge_cmap'])
+
+    # Filled is only supported for subclasses
+    filled = False
+
+    # Bezier paths
+    bezier = False
+
+    # Declares which columns in the data refer to node indices
+    _node_indices = [0, 1]
```
jlstevens: ``_node_columns`` maybe?

holoviews/plotting/bokeh/graphs.py
```diff
@@ -101,13 +111,31 @@ def _get_edge_colors(self, element, ranges, edge_data, edge_mapping, style):
...
+        path_data = {}
         xidx, yidx = (1, 0) if self.invert_axes else (0, 1)
+        if element._edgepaths is not None:
+            edges = element._split_edgepaths.split(datatype='array', dimensions=element.edgepaths.kdims)
+            if len(edges) == len(element):
+                path_data['xs'] = [path[:, xidx] for path in edges]
+                path_data['ys'] = [path[:, yidx] for path in edges]
+            else:
+                self.warning('Graph edge paths do not match the number of abstract edges '
+                             'and will be skipped')
```
jlstevens: Definitely a warning and not an error condition?

holoviews/element/graphs.py
philippjfr: I think the fact that it is a Graph type is an implementation detail but I agree ``from_vertices`` is better.

holoviews/plotting/bokeh/graphs.py
```diff
@@ -170,17 +193,29 @@ def get_data(self, element, ranges, style):
...
                 for d in element.dimensions():
                     path_data[dimension_sanitizer(d.name)] = element.dimension_values(d)
-        data = {'scatter_1': point_data, 'multi_line_1': path_data, 'layout': layout}
-        mapping = {'scatter_1': point_mapping, 'multi_line_1': edge_mapping}
+        data = {'scatter_1': point_data, self.edge_glyph: path_data, 'layout': layout}
+        mapping = {'scatter_1': point_mapping, self.edge_glyph: edge_mapping}
         return data, mapping, style
 
+    @property
+    def edge_glyph(self):
```
jlstevens: Little bit too long for what it does imho. Why not compress it a bit?

```python
if not any([self.bezier, self.filled):
   return  'multi_line_1'
return 'bezier_1' is self.bezier else 'patches_1'
```

holoviews/plotting/bokeh/graphs.py
philippjfr: Would be happy to make it an error, this has been part of the Graph implementation since first merging it and I've not run into it.

jlstevens:
Looks good!

My comments above are fairly minor and the only other thing is I would like to see the ``from_vertices`` classmethod mentioned in the two Trimesh element notebooks.

Happy to merge once these small issues are addressed.

holoviews/plotting/bokeh/__init__.py
philippjfr: Agreed, but they're mostly for interactive features when you're selecting or hovering over the graph.

philippjfr:
Addressed the comments, added vertex averaging and updated the reference notebooks. Ready for a final review.

jlstevens:
New changes look good and tests have passed. Merging!

PR #2144, opened 2017-11-22
Inherit parameters when casting Image to QuadMesh

philippjfr:
Due to an oversight the kdims, vdims, group and label aren't inherited when casting an Image to a QuadMesh.


philippjfr:
This is a simple bug fix and I've added a unit test, merging.

PR #2145, opened 2017-11-23
Improved Graph element

philippjfr:
Since I started writing the Chord (#2138) and TriMesh (#2143) elements, it's become clear that the Graph element is quite useful as a baseclass for other elements. However currently it isn't as solid, optimized and tested as I'd like it to be, so rather than mixing up a bunch of changes into the ``Chord`` and ``TriMesh`` PRs I've decided to first build on ``Graph``, ensuring that it is well tested and solid at it's core.

This will improve Graphs in a number of ways:

Graph Element Improvements
======================

1. Add an optimized pandas based implementation that handles drawing direct edge paths between nodes. The naive implementation relying just on HoloViews functionality is way to slow and this provides a ~100x speedup as long as pandas is installed.
2. Graphs allowed supplying node information without defining the explicit node positions. This code was brittle and relied on implicit assumptions on the ordering of nodes. I've hardened the exception handling here, being much stricter about how node information is merged and relying on pandas to handle actual merges between node indices.

Graph Plot improvements
===================

1. The Graph glyphs are now updated in a well defined order, which ensures that explicit paths are re-rendered correctly when updated.
2. The ``GraphPlot`` implementations now support an ``edge_color_index`` allowing them to be colored both by categorical attributes (e.g. by their source node) or by a value (e.g. a weight value).

- [x] Added tests
- [x] Added documentation

### DIFF
--- a/holoviews/element/graphs.py
+++ b/holoviews/element/graphs.py
@@ -9,7 +9,7 @@ from ..core.util import max_range
 from ..core.operation import Operation
 from .chart import Points
 from .path import Path
-from .util import split_path, pd
+from .util import split_path, pd, circular_layout, connect_edges, connect_edges_pd
 
 try:
     from datashader.layout import LayoutAlgorithm as ds_layout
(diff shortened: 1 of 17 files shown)


holoviews/element/graphs.py
```diff
@@ -34,13 +34,65 @@ def __call__(self, specs=None, **dimensions):
...
+    Lay out nodes on a circle and add node index.
+    """
     N = len(nodes)
     circ = np.pi/N*np.arange(N)*2
     x = np.cos(circ)
     y = np.sin(circ)
     return (x, y, nodes)
 
 
+def connect_edges_pd(graph):
```
jlstevens: Maybe move some of these functions to element.util?

philippjfr: Yes, sounds good.

holoviews/plotting/util.py
```diff
@@ -411,6 +412,59 @@ def map_colors(arr, crange, cmap, hex=True):
         return arr
 
 
+def mplcmap_to_palette(cmap, ncolors=None):
```
jlstevens: Seems like the right place for these utilities!

tests/testbokehgraphs.py
```diff
@@ -119,13 +121,52 @@ def test_graph_selection_policy_none(self):
...
+        factors = ['0', '1', '2', '3', '4', '5', '6', '7']
+        self.assertEqual(cmapper.factors, factors)
+        self.assertEqual(edge_source.data['start_str'], factors)
+        self.assertEqual(glyph.line_color, {'field': 'start_str', 'transform': cmapper})
+    
+    def test_graph_nodes_numerically_colormapped(self):
+        g = self.graph4.opts(plot=dict(edge_color_index='Weight'),
+                             style=dict(edge_cmap=['#FFFFFF', '#000000']))
+        plot = bokeh_renderer.get_plot(g)
+        print(plot.handles)
```
jlstevens: Stray print.

jlstevens:
Looks good! Other than some very minor comments I posted above, I am happy for you to go ahead and merge once you think it is ready.

philippjfr:
Yay, even got a gain in coverage. Thanks for the review, merging.

Issue #237, opened 2015-08-21
Transforming data and broadcasting


vascotenner:
Quite often, I want to substract a mean background from an Image or a Curve. This mean background is often calculated by reducing an image twice. Is it possible to substract a single number from an Image?

In numpy this would work like:
`image = img_raw - bg.mean()`

Or more complicated:

```
image.shape  = (10,20,300,300)
dark.shape  = (1,20,300,300)
image -= dark.mean(axis=-1)[...,np.newaxis]
```

I tried to accomplish this first in 1D (a Curve minus a number), but this is also not very easy.

Is this covered in the upcomming first part of the transformation tutorial?

philippjfr:
I'd recommend just doing the following:

```
img = hv.Image(np.random.rand(10,10))
img.clone(img.data - img.data.mean())
```

or on a HoloMap:

```
imgmap = hv.HoloMap({i: hv.Image(np.random.rand(10,10)) for i in range(3)})
imgmap.map(lambda x: x.clone(x.data - x.data.mean()), [hv.Image])
```

for Curves this is more difficult because of the way the data is stored (we're hoping to improve that in future), I can imagine something like this being supported:

jlstevens:
Perhaps there is a new way to do it using the Data API? Otherwise, if we feel this is resolved or don't have any new suggestions, we should close this issue.

philippjfr:
Hmm, I've considered whether Columns should support a ``setitem`` so you could do something like:

```python
curve = hv.Curve(np.random.rand(10))
curve['y'] = curve['y'] - curve['y'].mean()
```

We also still need to add the columns method which returns the Columns dictionary, that would at least allow:

```python
curve = hv.Curve(np.random.rand(10))
columns = curve.columns()
columns['y] = columns['y'] - columns['y].mean()
curve.clone(columns)
```

philippjfr:
Dataset.transform was merged in #4080 which allows specifying transforms. In future we may also add `Dataset.assign`.

PR #2152, opened 2017-11-25
Add support for dim transforms to set style options

philippjfr:
As discussed in https://github.com/ioam/holoviews/issues/768 this PR adds support for scaling various style options by values in the element. It adds an ``op`` object which allows defining deferred transforms on an element dimension implementing the standard mathematical operations and common NumPy operations.

You can see what is currently supported [here](https://anaconda.org/philippjfr/op_transforms/notebook).

As we had discussed in the issue we will also support tuple and string syntax so you only have to define an ``op`` to define complex transforms more easily.

Here are some example:

```python
scatter = hv.Scatter(np.arange(11))
o = (op('y')-op('y').min())/op('y').max()
print(o)
o.eval(scatter)
```

    truediv(sub('y', amin('y')), amax('y'))
    array([ 0. ,  0.1,  0.2,  0.3,  0.4,  0.5,  0.6,  0.7,  0.8,  0.9,  1. ])

```python
%%opts Points (alpha=op('z') size=np.log1p(op('z'))*20)
hv.Points(np.random.rand(100, 3), vdims=['z'])
```

bokeh_plot

```python
%%opts Curve [show_legend=False width=500] (line_width=10*op('y').mean())
hv.NdOverlay({i: hv.Curve(np.random.rand(10)) for i in range(5)})
```

bokeh_plot


jlstevens:
I'll be happy to see this supported though I do have one request...

I really want to see ``op`` used in the simplest cases, matching what we already support. One goal of ``op`` is to support more advanced control than has been possible so far (which your examples above demonstrate) but another goal should be to keep simple things looking simple. I want to make sure the simplest uses of ``op`` 1) work as a good default in many cases 2) remains succinct and readable.

philippjfr:
I'd also like this PR to address https://github.com/ioam/holoviews/issues/237.

jlstevens:
How do you think ``op`` can help with #237? It doesn't seem obvious to me yet...

philippjfr:
I've discussed this before but I was going to suggest an ``apply`` method (or similar) for elements, which would let you transform dimensions/columns using ops or simply replace it. I'll put together a fully fleshed out proposal at some point.

jlstevens:
I remember discussing ``apply`` with you but I didn't remember the idea that it could make use of ``op``. It seems like a reasonable idea but I don't see why that would need to be part of this PR...

philippjfr:
It wouldn't of course, could be a followup PR. I just wanted to establish a link between ``ops`` and that issue, so I don't forget about it.

philippjfr:
To track how complete this PR is I'm going to make a list of mappings that are supported:

Points/Scatter:
    - [x] size
    - [x] alpha
    - [x] color
    - [ ] marker
    
Curve:
    - [x] line width
    - [x] color
    - [x] dash
    - [x] alpha
    
Graph:
    - [ ] node color
    - [ ] node alpha
    - [ ] node size
    - [ ] edge color
    - [ ] edge alpha
    - [ ] edge line width

philippjfr:
This notebook also tracks what is and isn't supported yet:

https://anaconda.org/philippjfr/op_transforms/notebook

examples/user_guide/Styling_Plots.ipynb
```diff
@@ -179,8 +179,7 @@
     "\n",
     "## Colormapping\n",
     "\n",
-    "Color cycles and styles are useful for categorical plots and when overlaying multiple subsets, but when we want to map data values to a color it is better to use HoloViews' facilities for color mapping. Certain image-like types will apply colormapping automatically; e.g. for ``Image``, ``QuadMesh`` or ``HeatMap`` types the first value dimension is automatically mapped to the color. In other cases the values to colormap can be declared through the ``color_index``, which may reference any dimension by name or by numerical index.\n",
-    "\n",
+    "Color cycles and styles are useful for categorical plots and when overlaying multiple subsets, but when we want to map data values to a color it is better to use HoloViews' facilities for color mapping. Certain image-like types will apply colormapping automatically; e.g. for ``Image``, ``QuadMesh`` or ``HeatMap`` types the first value dimension is automatically mapped to the color. In other cases the values to colormap can be declared by mapping a ``color`` style option to an existing dimension.\n",
```
jbednar:
```suggestion
    "Color cycles and styles are useful for categorical plots and when overlaying multiple subsets, but when we want to map data values to a color it is better to use HoloViews' facilities for color mapping. Certain image-like types will apply colormapping automatically; e.g. for ``Image``, ``QuadMesh`` or ``HeatMap`` types the first value dimension is automatically mapped to the color. In other cases the values to colormap can be declared by providing a ``color`` style option that specifies which dimension to map into the color value.\n",
```

(I don't think it's meaningful to talk about mapping a color style option to a dimension; it's really the other way around, right?)

holoviews/util/transform.py
```diff
@@ -0,0 +1,293 @@
...
+            cat = categories.get(c, empty)
+        cats.append(cat)
+    return np.asarray(cats)
+
+
+
+class dim(object):
+    """
+    dim transform objects are a way to express deferred transformations
+    on HoloViews Datasets. It supports all mathematical operations,
```
jbednar:
```suggestion
    on HoloViews Datasets. Dims support all mathematical operations
```

holoviews/util/transform.py
```diff
@@ -0,0 +1,293 @@
...
+    NumPy ufuncs and provides a number of useful methods for normalizing,
```
jbednar:
```suggestion
    and NumPy ufuncs, and provide a number of useful methods for normalizing,
```

holoviews/util/transform.py
```diff
@@ -0,0 +1,293 @@
...
+            fn = self._op_registry.get(fn)
+            if fn is None:
+                raise ValueError('dim transform %s not found' % fn)
+        if fn is not None:
+            ops = ops + [{'other': other, 'fn': fn, 'kwargs': kwargs,
+                          'reverse': reverse}]
+        self.ops = ops
+
+    @classmethod
+    def resolve_spec(cls, op_spec):
```
jbednar: dim_from_spec?

holoviews/util/transform.py
```diff
@@ -0,0 +1,293 @@
...
+            obj, fn, other, kwargs = (o2 if o1 is None else o1 for o1, o2 in combined)
+            if isinstance(obj, tuple):
+                obj = cls.resolve_spec(obj)
+            return cls(obj, fn, other, **kwargs)
+        return op_spec
+
+    @classmethod
+    def register(cls, key, function):
+        """
+        Register a custom op transform function which can from then
```
jbednar:
```suggestion
        Register a custom dim transform function which can from then
```

holoviews/util/transform.py
```diff
@@ -0,0 +1,293 @@
...
+    from itertools import izip_longest as zip_longest
+
+import numpy as np
+
+from ..core.dimension import Dimension
+from ..core.util import basestring, unique_iterator
+from ..element import Graph
+
+
+def norm_fn(values, min=None, max=None):
```
jbednar: Can't this just be `norm`, and the same for `bin`?  I don't see what the `fn` buys here, and when I see that I expect to see a family of normalization functions, not just one simple normalizer.  I don't think there is any namespace clash with the `norm` and `bin` methods on dim, as these are just bare functions in the module namespace.

holoviews/util/transform.py
```diff
@@ -0,0 +1,293 @@
...
+        labels = np.asarray(labels)
+    dtype = 'float' if labels.dtype.kind == 'f' else 'O'
+    binned = np.full_like(values, (np.nan if dtype == 'f' else None), dtype=dtype)
+    for lower, upper, label in zip(bins[:-1], bins[1:], labels):
+        condition = (values > lower) & (values <= upper)
+        binned[np.where(condition)[0]] = label
+    return binned
+
+
+def cat_fn(values, categories, empty=None):
```
jbednar: Here I'd would have suggested just `cat` instead of `cat_fn`, and I do think that would be better, but I personally cannot think of `cat` on a computer as anything other than the UNIX command by that name, and so it expands to "concatenate" in my mind, not "categorize" or "categorical".  I'd vote for calling this function and the corresponding method `categorical`, because they turn what they are given into a categorical value, right?  I don't think it's such a common operation that it needs to shave off the last few characters, when that makes it confusing as in this case...

holoviews/util/transform.py
philippjfr: I guess, but these are not functions a user would ever access so I see very limited benefit. If you think I should rename them I'll change it to `_norm` and `_bin`.

holoviews/util/transform.py
philippjfr: I think I prefer `categorize`, but it is a bit of shame since ``dim('value').categorize(['o', 'x', '+'])`` is significantly longer than `dim('value').cat(['o', 'x', '+'])`. You're definitely right that `cat` usually refers to concatenate though, that's true even in pandas.

philippjfr:
So I think we are now in a place where this can be merged, but there are still a few outstanding items which will need to happen in later PRs:

* Matplotlib BarPlot needs rewriting from scratch, I'm not touching the current implementation with a 10 foot barge pole
* SankeyPlot does not yet support style mapping, this will require some work because the styles need to be carried through the operation
* I need to finish writing a style mapping tutorial and or integrate the material in some existing tutorial
* Eventually we want to support applying ops directly to elements to either transform existing dimension values or add new ones
* Write more unit tests for plot updates

Some other features this PR introduces:

* The normalization system now handles normalizing Graph edge start/end values along the node indices which is important for consistent coloration between nodes and edges
* The normalization system now handles computing discrete factors across plots

Another idea I've had, is that we could allow specifying a transform(s) directly on a `Dimension` object so when you query the dimension on an element the transform is applied. This could be **very** powerful as it could support a huge range of transformations:

* Unit transforms, e.g. when querying the values would be transformed to some fixed unit
* Coordinate system transforms
* Storing regularly spaced log transformed values from datashader on Image coordinates (solving https://github.com/ioam/holoviews/issues/2195)
* Jitter of axis locations
* Lots more I can't think of right now

jbednar:
Sounds great!

philippjfr:
Tests should pass shortly. Let's get this merged so we can get a few days of testing in before a release.

examples/gallery/demos/bokeh/autompg_violins.ipynb
```diff
@@ -50,7 +50,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "violin.options(height=500, width=900)"
+    "violin.options(height=500, width=900, violin_fill_color=('Year', str), cmap='Set1')"
```
jlstevens: I think this is a stray instance of the tuple format...

examples/gallery/demos/bokeh/boxplot_chart.ipynb
```diff
@@ -53,7 +53,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "boxwhisker.options(show_legend=False, width=400)"
+    "boxwhisker.options(show_legend=False, width=600, box_fill_color=('origin', str), cmap='Set1')"
```
jlstevens: And another tuple...

jlstevens:
I'm happy to merge but I think I spotted a few stray instances of the tuple format above.

jlstevens:
Tests are now green.

I'm very happy to merge this PR as I think it will greatly simplify things throughout while exposing a more powerful way of working. It will require a fair bit of testing and of course we may well have various tweaks to make. For the time being, I'll finally go ahead and merge...nearly a year since this PR was first opened!

Issue #2139, opened 2017-11-19
Add Radial Heatmap


mansenfranzen:
## Proposal

Radial heatmaps are great for visualizing periodicity. I would like to help adding radial heatmaps to holoviews.

I've implemented radial heatmaps in bokeh, as you can see here:
![radialheatmap](https://user-images.githubusercontent.com/18086180/32992237-917b0346-cd48-11e7-917c-f75570cef8b2.png)

The link to the notebook including the code can be found [here](https://github.com/mansenfranzen/holoviews/blob/feature_radial_heatmap/examples/gallery/demos/bokeh/radial_heatmap.ipynb ).

## First steps

I started to take a closer look on how to implement a radial heatmap in holoviews having a working bokeh example. Here are some considerations:

- Data wise, radial heatmaps are very much like a normal, rectular heatmap. Both have 2 key dimensions and one or more value dimension whereas the first value dimension is used for color. Hence, the [Heatmap](https://github.com/ioam/holoviews/blob/master/holoviews/element/raster.py#L772) Element class is the perfect data container.
- Plotting wise, the corresponding [HeatmapPlot](https://github.com/ioam/holoviews/blob/master/holoviews/plotting/bokeh/raster.py#L111) provides the basic functionality. The `get_data` method needs to be adjusted to compute radius, radients and so on (which is no big deal having a working example)

### Issues

However, I've ran into 2 concrete problems:

1. Since radial and rectular heatmaps are very similar, they could be represented within the same class (e.g. HeatmapPlot). To select either a radial or rectular heatmap, a parameter could be used like in grouped or stacked bar plots. However, the underlying glyphs for radial (annular_wedge) and rectular (rect) heatmaps differ. By design, a plot class may only have one *single* key for `_plot_methods`, right? Therefore, my current guess is to create a new plotting class (and hence a new element class).

2. The radial heatmap requires axis labels to be positioned manually since radial axis to not exist. Because of that, I need to add text glyphs to the annular_wedge glyphs. Furthermore, I would like to add some visual marker to increase readability of the chart, like those bold white lines in the following plot which results in 3 different glyphs (annular_wedge, text and multi_line). How can I integrate different glyphs in the same plot? Overlay?:

![radialheatmap2](https://user-images.githubusercontent.com/18086180/32992551-43f0e420-cd4c-11e7-876d-d2177b77ad8c.png)

Thanks for your advice.

philippjfr:
[...]

PR #2155, opened 2017-11-26
Feature radial heatmap

mansenfranzen:
This PR is not meant to be finished but is rather a work in progress for discussion purposes.

I implemented a first version of a radial heatmap which still has several limitations:

### Issues

- **Plotting order**: In order to enhance readability, one can define separation lines to distinguish groups of annular wedges more clearly via plot option `separate_nth_segment`. However, sometimes they show up, sometimes they don't. I guess this is due to the z-order in which glyphs are plotted. I added the `_draw_order` class attribute without any effect.

- **Match aspect**: Coordinates of annular wedges, lines and text do not result in corresponding alignments (same coordinates differ in visual field). This is a known issue in bokeh, see here [#7218](https://github.com/bokeh/bokeh/issues/7218). One workaround I've found is to completely remove any range and axis definition from a bokeh figure and set `match_aspect` to `True` in order to get the correct aspect ratio (however one has to overwrite many inherited methods...). Another way is to manually adjust the width and height to match the aspect ratio but this depends on toolbar location, colobar and other elements and hence is not really meaningful. Therefore, I guess we have to wait until a bokeh site of fix.

- **Colobar does not show up** if selected

- **Input key dimension differ to output dimension**: The key dimensions of the input data are categorical. However, the plot is based on numerical values a in interval scaled coordinate system. As far as I can understand the class design, holoviews expects input key dimension to map on the actual x and y ranges. I've thought about creating a custom `holoviews.Operation` to convert the categorical input into numerical radii and radiants just like `histogram`.

### Still missing:

- **Add labels for second key dimension**: At the moment, labels are only available for the first key dimension.

### Open question:

- **Separate** `RadialHeatMap` **class**: As discussed in #2139, I decided to create its own element class representation instead of using the `HeatMap` element class because the radial heatmaps also require custom layout options which are very different from the rectular heatmap requirements. WIthout the extra class, I wasn't able to set the visualization defaults in the *bokeh/__init__.py* properly.

Overall, the plain example works in *examples/gallery/demos/bokeh/radial_heatmap.ipynb*.

I'm very happy for every feedback.


philippjfr:
This is looking great already!

>Plotting order

Looks like this is a bug in the implementation, I'll have a PR fixing this tomorrow.

>Match aspect

I briefly investigated exposing ``match_aspect`` on all our plot classes but didn't really have much luck. I'll try to look into it again, so hopefully we can use it here.

>Colobar does not show up if selected

Looks like this is true for all CompositePlot classes. The problem lies with ColorbarPlot._init_glyph never getting called. Really ``ColorbarPlot`` should be a mixin class, looks like I'll have to do some refactoring here.

>Input key dimension differ to output dimension: The key dimensions of the input data are categorical. However, the plot is based on numerical values a in interval scaled coordinate system. As far as I can understand the class design, holoviews expects input key dimension to map on the actual x and y ranges. 

While this is true in general I don't think it's necessarily a huge problem for a radial heatmap because the heatmap is always a fixed size. Making space for the labels is still a problem though so I'll mull it over.

>Separate RadialHeatMap class: As discussed in #2139, I decided to create its own element class representation instead of using the HeatMap element class because the radial heatmaps also require custom layout options which are very different from the rectular heatmap requirements. WIthout the extra class, I wasn't able to set the visualization defaults in the bokeh/init.py properly.

I agree this is probably necessary, would be nice if we could find a shorter name though.

Some other to do items include (I'm happy to help or take on any or all of these):

- [ ] Add reference material in ``examples/reference/element/`` (the current demo is a good starting point)
- [ ] Add a few plot tests basically testing whether the datasources and glyphs are initialized correctly.
- [ ] Add an equivalent implementation for the matplotlib backend.

Thanks for the contribution, the code looks very clean and I'd be very happy to see this included with a bit more work.

holoviews/plotting/bokeh/raster.py
```diff
@@ -161,6 +164,239 @@ def get_data(self, element, ranges, style):
...
+
+    _draw_order = ["annular_wedge", "multi_line", "text"]
+
+    style_opts = (['separator_' + p for p in line_properties] + \
+                  ['annular_' + p for p in fill_properties + line_properties] + \
+                  ['label_' + p for p in text_properties] + ['width', 'cmap'])
+
+
+    @staticmethod
+    def _extract_implicit_order(array):
```
philippjfr: I think you can use ``holoviews.core.util.unique_array`` or ``holoviews.core.util.unique_iterator`` for this instead.

holoviews/plotting/bokeh/raster.py
```diff
@@ -161,6 +164,239 @@ def get_data(self, element, ranges, style):
...
+                return [dim.pprint_value(v) for v in vals]
+
+            return vals
+
+        xvals = _pprint(x, xvals)
+        yvals = _pprint(y, yvals)
+
+
+        # get orders
+        order_segment = self._extract_implicit_order(xvals)
```
philippjfr: I'd actually just use ``aggregate.dimension_values(x, expanded=False)`` and ``aggregate.dimension_values(y, expanded=False)``  here instead.

philippjfr:
I took the liberty of implementing the matplotlib version of this plot, which you can see [here](https://anaconda.org/philippjfr/radial_heatmap/notebook ). If it's okay with you I'll push that to the PR tomorrow.

Here's the output:

image

The only change I made to the parameters was using the ``xticks`` and ``yticks``  parameters to control the number of ticks.

jbednar:
Looks good! Is it intentional that the annular wedges are polygonal rather than actual annuli?  It's an odd effect even in the above plot, and I can't see how the above plot would work at all with fewer radial divisions. I.e., what happens when a radial section isn't convex (e.g. if one section covers more than 180 degrees), or even just if there are only two 180-degree sections?

Also, it might be nice to have an option for having the axial spacing be constant in terms of area, so that the space occupied by Tuesday 31 is the same as that occupied by Sunday 15 (which requires the ring spacing to be nonlinear).

philippjfr:
>Is it intentional that the annular wedges are polygonal rather than actual annuli?

No but I really don't know of a non-painful way to draw them correctly, all I can think of doing to handle the case of a low number of bins is to upsample the data by a scalar factor along that xaxis to ensure a minimum number of bins.

jbednar:
Right; I updated my comment to refer just to the mpl version after seeing the beautiful wedges on the Bokeh version.  Do the wedges illustrated [here](https://matplotlib.org/examples/api/patch_collection.html) not work?

philippjfr:
>Also, it might be nice to have an option for having the axial spacing be constant in terms of area, so that the space occupied by Tuesday 31 is the same as that occupied by Sunday 15 (which requires the ring spacing to be nonlinear).

Should be straightforward to do, but I'm unsure if that would be very interpretable in many cases.

philippjfr:
>Do the wedges illustrated here not work?

Possibly, I suppose the number of wedges will never be huge, but I can see this being slow. Also haven't yet found any example with polar axes but it's worth a shot.

philippjfr:
Here it is using matplotlib ``Wedges``:

image

holoviews/plotting/bokeh/raster.py
mansenfranzen: Thanks for the hint - no need to reimplement what is already there :-)

holoviews/plotting/bokeh/raster.py
mansenfranzen: Agreed - definitely better.

mansenfranzen:
@philippjfr 

> If it's okay with you I'll push that to the PR tomorrow.

Sure, go ahead - great to see you've implemented the MPL version so quickly :-).

>  Add a few plot tests basically testing whether the datasources and glyphs are initialized correctly.

I'm happy to do so: do you have an example test module of an already implemented bokeh plot class?

> Add reference material in examples/reference/element/ (the current demo is a good starting point)

Sure, I will add a reference documentation later this week.

---

@jbednar 

> Also, it might be nice to have an option for having the axial spacing be constant in terms of area, so that the space occupied by Tuesday 31 is the same as that occupied by Sunday 15 (which requires the ring spacing to be nonlinear).

Sounds like an interesting feature to implement. Would be great to see the effect of this in practice. Hope to add this soon, too.

And thx for the typo fixes ;-).

philippjfr:
>Sure, go ahead - great to see you've implemented the MPL version so quickly :-).

It actually turned out to be fairly easy, because I could mostly reuse your existing code and because matplotlib actually handles polar axes so I didn't have to implement custom labelling like you did.

philippjfr:
Just some quick thoughts on demos and reference material, I think if we could simplify the current example a little bit, it's fine as the reference example. For the gallery demo I would like to use global temperature anomaly data unless you can think of something better, something like this:

image

mansenfranzen:
> Just some quick thoughts on demos and reference material, I think if we could simplify the current example a little bit, it's fine as the reference example. For the gallery demo I would like to use global temperature anomaly data unless you can think of something better, something like this:

Sounds like a great visualization example. I think the viz could be even more striking if we restrict temperatures to be from one country/region (Germany or UK) in order reveal seasonal effects between winter and summer months, too.

jbednar:
> Here it is using matplotlib Wedges

Looks great!  I'd vote for that even if it's slower, because it's a lot clearer and should behave well for any wedge size.

It would be great to have a weekly example (seems like one could plot ridership numbers for the nyc_taxi data) where weekends stand out clearly, plus a monthly example like the temperature data, hopefully with clear cyclical variations in both cases to make it clear why to use this type of plot.

philippjfr:
The plotting order bug should be fixed by #2169, which I just merged.

mansenfranzen:
@philippjfr Thanks for fixing the plotting order bug. It works as expected now. I did a rebase on the current feature branch to rewrite the commit history and to include your fix.

I've simplified and refactored the existing code for better readability using more holoviews builtins as suggested by @philippjfr . In addition, I added support for annular tick labels. I dropped the `show_nth_segement` plot option in favour of `xticks`. Segment labels can now be set natively via `xticks` while annular labels can be defined via `yticks`.

### Reference example
I also added a reference example with the two examples of global surface temperatures and the NYC taxi dataset:

#### Absolute global temperatures since 1750
globaltemperatures

#### NYC Taxi pickup counts for 2016
@jbednar 
nyc_taxi_2016

For more detail on the plot examples, take a look at the reference examples. The NYC example is pretty good :-).

I haven't finished the reference example, yet. But please feel free to modify, comment or improve my English :-).

I added two datasets to the *examples/assets* directory which are used in the reference example. Both datasets are freely available. I modified them for easier usage. Creation and source can be found [here](https://github.com/mansenfranzen/notebooks/tree/master/SampleDatasets). I'm not completely sure on how to handle sample datasets and how this is supposed to be in holoviews. I've thought about adding the datasets to bokeh.sampledata as bokeh already has an existing sample data setup.

#### Still open

- Tests: Can you please provide an example of other plot classes having tests already? I need some guidance here.
- Annular separation lines: Basically the same as the segment separation lines.
- Match aspect workaround

philippjfr:
>@philippjfr Thanks for fixing the plotting order bug. It works as expected now. I did a rebase on the current feature branch to rewrite the commit history and to include your fix.

Great!

>I've simplified and refactored the existing code for better readability using more holoviews builtins as suggested by @philippjfr . In addition, I added support for annular tick labels. I dropped the show_nth_segement plot option in favour of xticks. Segment labels can now be set natively via xticks while annular labels can be defined via yticks.

Perfect!

>I also added a reference example with the two examples of global surface temperatures and the NYC taxi dataset:

Both look great, particularly the NYC taxi one. I think we can drop the temperature idea. I'm not sure about the dataset yet. I think we do need a mechanism to download larger datasets as we do want a wide range of examples.

>Tests: Can you please provide an example of other plot classes having tests already? I need some guidance here.

I'm planning on moving all of these into each respective plotting backend but for now https://github.com/ioam/holoviews/blob/master/tests/testbokehgraphs.py probably provides the best example.

mansenfranzen:
Here is the current update on the radial heatmaps:

#### Additions / Modifications:
1. The **reference example now focuses on the NYC example** providing some modification using several plot options to enhance the plot. It still contains the global temperature example at the end. Feel free to remove it if you want but it also provides some additional information about plot options.
2. The **`seperate_nth_segment` was renamed to `xmarks`** in analogy to `xticks`. Accordingly, `ymarks` was also added to add separation lines for annulars. The `y/xmarks` allow integers, lists, tuples and functions to support different requirements. 
3. **`x/yticks` now also support python functions** to define which ticks should be shown or not. A custom bokeh ticker does not make sense here because there is not *real* bokeh axis object because we have to fake the polar axis.
4. Several **tests were added** for the bokeh plot.
5. A new **`yticks_angle`** plot option was added to define the angle along which the yticks are drawn.

#### Open:
1. The MPL plot is missing the `ymarks` and `yticks_angle` plot options but this shouldn't be a great deal.
2. The MPL plot is also lacking tests. 
3. The MPL plot does not have a reference example yet.

#### Issues:
1. Text glyphs are used to create the xticks and yticks. It would be great to modify their visual text properties like size and color independently. However, it seems to as if this is not possible at the moment because `_style_groups` does not allow to specify the same glyph to multiple style groups. I can only change the visual property of text glyphs in general but not for a custom group of text glyphs. 
2. I just realized that tests failed for Python 2.7. I will have a look at that.

philippjfr:
>The reference example now focuses on the NYC example providing some modification using several plot options to enhance the plot. It still contains the global temperature example at the end. Feel free to remove it if you want but it also provides some additional information about plot options.

Great, I'll try to review this soon.

>The seperate_nth_segment was renamed to xmarks in analogy to xticks. Accordingly, ymarks was also added to add separation lines for annulars. The y/xmarks allow integers, lists, tuples and functions to support different requirements.

Sounds good.

>x/yticks now also support python functions to define which ticks should be shown or not. A custom bokeh ticker does not make sense here because there is not real bokeh axis object because we have to fake the polar axis.

Sounds good! I agree tickers don't make much sense here and I don't think they are used much. We do support supplying ticks in the form ``[position1, position2, ...]`` and ``[(position1, label), (position2, label), ...]`` but this does not work for categorical axes, so I wouldn't really expect to support it here.

>Several tests were added for the bokeh plot.

Sounds good!

>The MPL plot is missing the ymarks and yticks_angle plot options but this shouldn't be a great deal.
>The MPL plot is also lacking tests.
>The MPL plot does not have a reference example yet.

I'll handle these.

>Text glyphs are used to create the xticks and yticks. It would be great to modify their visual text properties like size and color independently. However, it seems to as if this is not possible at the moment because _style_groups does not allow to specify the same glyph to multiple style groups. I can only change the visual property of text glyphs in general but not for a custom group of text glyphs.

Yes, this is a definite problem I've also come across in other context. We'll open an issue to allow them to be styled separately once this is supported.

philippjfr:
@jbednar @jlstevens Any ideas for a shorter name than ``RadialHeatMap``?

jlstevens:
@mansenfranzen First off, these plots are beautiful and thanks for contributing!

> Any ideas for a shorter name than RadialHeatMap?

Good question, I'll have to think about it. Ideally, there is a unique name so it sounds like an element with its own identity but I can only think of compound names right now. ``AnnularBins``? A little bit of research might help come up with a better name...

jbednar:
What about ``HeatMap``?  Seems to me that whether it's radial or not is simply a plot option, and from a user perspective there's no semantic difference between a radial and a non-radial heat map.  I know that would take some fiddling with the code, but it seems like the implementation here has gradually become more like that of the regular HeatMap, so that they are no longer so very far apart in terms of their parameters and behavior.

philippjfr:
Thanks again for all your effort on this @mansenfranzen. I've been ill all week and have some other stuff to catch up on but hope to finish this off next week. I think the remaining bits are on me to handle.

jbednar:
> Rectangular heatmaps have categorical whereas radial heatmaps have numerical axes. 

Supporting numerical axes on rectangular heatmaps seems important to me anyway; people use them for both categorical and binned-continuous data.

Sounds to me like there need to be two separate objects at the implementation level, at least right now, but at the user level I think I still come down in favor of having only one Element type to worry about, even if some options only apply to some of the different variants of this plot.

jlstevens:
I think I am in favor of *not* unifying the elements and using plot options. Plot options should be choices about how the data is presented, not what it *means*. I am starting to think the fact the data is cyclic is about the data semantics so having a separate  ``RadialHeatMap`` element and corresponding plotting class might be a more sensible approach...

jbednar:
I considered that aspect as well, but I don't think it holds.  If we *always* wanted to see cyclic data as a radial heatmap, I think it would be true that cyclic <=> radial.  But I think people will very often try both ways, decide which one is clearer, and go with that one for the same data, which is basically the definition of something that should be a plot option.  Even for cyclic data, radial plots aren't always the right choice, because they are confusing to many people and because they don't ordinarily preserve the overall distribution of values perceptually (as recent values take up far more area than less recent ones).  So people have to decide based on stylistic, personal, or arbitrary criteria; it's not at all as simple as "is the data cyclic?".

philippjfr:
@mansenfranzen Sorry for letting this sit for so long, I'll try to get this finished this weekend.

mansenfranzen:
@philippjfr No worries. Please let me know if there is anything else you could need help with.

philippjfr:
Thanks, but it should be alright, I need to improve the matplotlib version a bit but I've now got HeatMap working with a ``radial`` plot option and have removed the ``RadialHeatMap`` element. Seems to work well, but I'll probably have to do some further tweaking.

philippjfr:
Now made matplotlib consistent with the bokeh implementation, adding ymarks and supporting the various tickers and markers that are supported in bokeh (ints, lists/tuples, functions, etc.). I've also copied the bokeh reference notebook and confirmed it visually matches the bokeh version. I'm happy to keep RadialHeatMap reference examples even though they are now combined on HeatMap, simply to make the functionality more visible.

Once I've added some matplotlib tests and we've added gallery examples I think this is ready to merge.

mansenfranzen:
@philippjfr Looks great - thanks. I like the `np.deg2rad` choice for the mpl plot. It is more readable than the current bokeh implementation which relies on radiants. I think we should also change this for the bokeh plot in order to make the bokeh and mpl plots consistent.

mansenfranzen:
@philippjfr Please let me know, if there's anything else to do in order to complete the current PR.

philippjfr:
No I think you're fine, I'll try to finish this up asap. Shouldn't take much now. It'll definitely go into 1.10.

philippjfr:
Ready for review.

examples/gallery/demos/bokeh/nyc_radial_heatmap.ipynb
```diff
@@ -0,0 +1,128 @@
...
+    "Let's see the result of these modifications:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "%%opts HeatMap [radial=True width=600 height=600 yticks=None xmarks=7 ymarks=3 start_angle=np.pi*19/14]\n",
```
jlstevens: Instead of having both ``radial=True`` (just a boolean) why not require ``start_angle`` which toggle radial if non-None? I would prefer to reduce the number of plot options and parameters involving ``angle`` only make sense to me for the radial case...

jlstevens: Looking at the plotting class definition there seem to be quite a few parameters specific to radial. I wonder if the ``is_heatmap`` property could detect the use of radial heatmap parameters. The problem then is you would probably need the defaults to be ``None``...

holoviews/plotting/bokeh/raster.py
```diff
@@ -156,10 +167,404 @@ def get_data(self, element, ranges, style):
...
+    start_angle = param.Number(default=np.pi/2, doc="""
+        Define starting angle of the first annulus segment. By default, begins 
+        at 12 o'clock.""")
+
+    padding_inner = param.Number(default=0.1, bounds=(0, 0.5), doc="""
+        Define the radius fraction of inner, empty space.""")
+
+    padding_outer = param.Number(default=0.05, bounds=(0, 1), doc="""
+        Define the radius fraction of outer space including the labels.""")
+
```
jlstevens: @philippjfr I think I would prefer different names for these as the use of ``padding`` can get confused with the more general idea of ``padding`` we want to introduce. Maybe ``ratio_inner`` or ``fraction_inner`` and the same for the ``outer`` parameter?

holoviews/plotting/bokeh/raster.py
```diff
@@ -156,10 +167,404 @@ def get_data(self, element, ranges, style):
...
+        separation lines at given annular values. If parameter is of type 
+        function, draw separation lines where function returns True for passed 
+        annular value.""")
+
+    max_radius = param.Number(default=0.5, doc="""
+        Define the maximum radius which is used for the x and y range extents.
+        """)
+
+    radial = param.Boolean(default=True, doc="""
+        Whether the HeatMap should be radial""")
```
jlstevens: Maybe ``max_radius`` could just be ``radius`` and a none ``None`` value would effectively switch ``radial`` to True? Probably nicer than using ``start_angle`` for this purpose...

holoviews/plotting/bokeh/raster.py
```diff
@@ -156,10 +167,404 @@ def get_data(self, element, ranges, style):
...
+        categorical input and numerical output axes.
+
+        Axis properties are handled separately for radial heatmaps because of
+        missing radial axes in bokeh.
+
+        """
+
+        return {}
+
+
```
jlstevens: There is a fair bit of wasted vertical space in the methods above (extra newline at the end of the docstring and one before the code starts).

examples/gallery/demos/bokeh/nyc_radial_heatmap.ipynb
philippjfr: Much prefer being explicit about radial, anything else is confusing especially enabling it via the angle. I do agree that having options on HeatMap that only apply to the radial plot isn't ideal though.

holoviews/plotting/bokeh/raster.py
philippjfr: Yeah I agree. Not sure which I'd prefer though.

holoviews/plotting/bokeh/raster.py
philippjfr: Also dont like this overloading and doesn't seem much more intuitive than using the angle. Both are options you wouldnt necessarily set and it's not immediately clear what you'd set it too.

holoviews/plotting/bokeh/raster.py
```diff
@@ -156,10 +167,404 @@ def get_data(self, element, ranges, style):
...
                                    for v in aggregate.dimension_values(vdim)]
+
+        # Filter radial heatmap options
+        style = {k: v for k, v in style.items() if not
+                 any(g in k for g in RadialHeatMapPlot._style_groups.values())}
         return (data, {'x': x, 'y': y, 'fill_color': {'field': 'zvalues', 'transform': cmapper},
                        'height': 1, 'width': 1}, style)
 
 
+class RadialHeatMapPlot(CompositeElementPlot, ColorbarPlot):
```
jlstevens: Between ``HeatMapPlot`` and ``RadialHeatMapPlot`` I wonder if it is worth introducing ``plotting/bokeh/heatmap.py`` (and the same for matplotlib).

philippjfr: No objection, they aren't really rasters anyway.

holoviews/plotting/bokeh/raster.py
jlstevens: ``ratio`` is shorter than ``fraction`` and I think more accurate.

examples/gallery/demos/bokeh/nyc_radial_heatmap.ipynb
jlstevens: I suppose I don't mind the ``radial`` parameter working and existing as a boolean - if you are happy with all the other defaults for radial heatmaps, you can use it. But then if you do want to set a parameter that does clearly apply only to radial heatmaps, it becomes unnecessary. This logic would probably all be in ``is_radial``.

Not sure it is worth the effort though...

holoviews/plotting/bokeh/raster.py
jlstevens: Agreed. I think it would be less confusing.

examples/gallery/demos/bokeh/nyc_radial_heatmap.ipynb
philippjfr: Seems reasonable, let me think about it. Maybe I'll try to reduce the number of parameters that are exclusive to radial heatmaps first, the xmarks and ymarks parameters for instance could be ported to the regular heatmap.

jlstevens: That was on my mind as another comment though I didn't bring it up. Glad you mentioned it as I agree!

mansenfranzen:
Just as a quick note: when I run the `\holoviews\examples\reference\elements\bokeh\RadialHeatMap.ipynb` example in the current feature branch, I get the following error:

```Traceback (most recent call last):

  File "C:\conda\envs\env_usage_prediction\lib\site-packages\holoviews\plotting\bokeh\element.py", line 925, in _init_glyphs
    renderer, glyph = self._init_glyph(plot, mapping.get(key, {}), properties, key)

  File "C:\conda\envs\env_usage_prediction\lib\site-packages\holoviews\plotting\bokeh\element.py", line 990, in _init_glyph
    renderer = getattr(plot, plot_method)(**dict(properties, **mapping))

  File "fakesource", line 9, in text

  File "C:\conda\envs\env_usage_prediction\lib\site-packages\bokeh\plotting\helpers.py", line 753, in func
    glyph = _make_glyph(glyphclass, kwargs, glyph_ca)

  File "C:\conda\envs\env_usage_prediction\lib\site-packages\bokeh\plotting\helpers.py", line 308, in _make_glyph
    return glyphclass(**kws)

  File "C:\conda\envs\env_usage_prediction\lib\site-packages\bokeh\model.py", line 219, in __init__
    super(Model, self).__init__(**kwargs)

  File "C:\conda\envs\env_usage_prediction\lib\site-packages\bokeh\core\has_props.py", line 236, in __init__
    setattr(self, name, value)

  File "C:\conda\envs\env_usage_prediction\lib\site-packages\bokeh\core\has_props.py", line 271, in __setattr__
    (name, self.__class__.__name__, text, nice_join(matches)))

AttributeError: unexpected attribute 'line_alpha' to Text, similar attributes are text_alpha
```

The `matplotlib` version works fine.

philippjfr:
Thanks. Was hoping to get to this this weekend but I've been distracted by JupyterLab support.

philippjfr:
Going to try to finish this off tonight. Sorry again @mansenfranzen for having left this for so long.

mansenfranzen:
@philippjfr No worries - I'm glad this feature is going to be available soon. Thanks for your help with it.

philippjfr:
>Just as a quick note: when I run the \holoviews\examples\reference\elements\bokeh\RadialHeatMap.ipynb example in the current feature branch, I get the following error:

Can't seem to reproduce this issue.

holoviews/plotting/bokeh/raster.py
philippjfr: Looking back at it now, I think I much prefer ``fraction``.

philippjfr: Decided to use ``radius_inner`` and ``radius_outer``.

philippjfr:
Ready to review again.

jlstevens:
Looks good to me and @mansenfranzen, thank you very much for your contribution!

PR #2157, opened 2017-11-27
Fixed shared_axis bug when plotting GridMatrix with Histogram

philippjfr:
This code was leftover from when Histogram's all had the same y-axis label and it was more convenient to ensure they were not linked when in GridMatrix. Fixes https://github.com/ioam/holoviews/issues/2131


philippjfr:
Very straightforward fix, merging.

jbednar:
Works well, thanks!

PR #2159, opened 2017-11-28
Handle ToolbarBox size for bokeh 0.12.11 compatibility

philippjfr:
Small change required for compatibility with bokeh 0.12.11.


philippjfr:
Straightforward fix that doesn't affect anything else.

Issue #2162, opened 2017-11-29
Changing Overlay Axis label based on Dimension's  unit attribute


johnzzzz:
When composing multiple Curves into a single chart using an Overlay, the y-axis is formatted as the FIRST Curve's  ydims Dimension's  "{label} ({unit})". 
This labeling convention might be fine to display a single Curve, but if there are multiple Curves per chart, the Y-axis label should not mention the First curve's label, but should instead be the Units that are common to all the Curves in the Overlay. 
1. How does one directly change the axis label of an Overlay?
2. Is there a way to get the Dimensions units to be used as an axis label?
3. Is there control over the axis label's format?

For example this code creates a chart comparing the acceleration  of two cars. The y-axis should be the unit "mph" and not "Porsche (mph)"

``` python
# matplotlib defaults
import matplotlib as mpl
mpl.rcParams['figure.facecolor']="white"
# Holoviews
import holoviews as hv
hv.extension('matplotlib')
# Compare Accelleration Data
import pandas as pd
raw_data = [{'time':0, 'yugo':0, 'porsche':0},
            {'time':1, 'yugo':10, 'porsche':10},
            {'time':2, 'yugo':20, 'porsche':30},
            {'time':3, 'yugo':30, 'porsche':90}]
data = pd.DataFrame(raw_data)
# Compare Chart
porsche = hv.Dimension('porsche', label='Porsche', range=(0,100), unit='mph')
yugo = hv.Dimension('yugo', label='Yugo', range=(0,55), unit='mph')
time = hv.Dimension('time', label='Time', range=(0,3), unit='sec')
porsche_curve=hv.Curve(data, kdims=[time], vdims=[porsche], label=porsche.label)
porsche_dot=hv.Scatter(data, kdims=[time], vdims=[porsche], label=porsche.label)
yugo_curve=hv.Curve(data, kdims=[time], vdims=[yugo], label=yugo.label)
yugo_dot=hv.Scatter(data, kdims=[time], vdims=[yugo], label=yugo.label)( style={'marker':'s'})
cmp_chart = (porsche_curve * porsche_dot * yugo_curve * yugo_dot).relabel('Compare Auto Accelleration')(plot={'legend_position':'top_left'})
cmp_chart
```
<img width="241" alt="holoviewaxislabelexample" src="https://user-images.githubusercontent.com/18220857/33356035-3a1ee788-d488-11e7-99e5-925a3772f86f.png">

jbednar:
[...]

PR #2164, opened 2017-12-01
Handling pandas Series and single column DataFrames

philippjfr:
This PR adds support for handling pandas and dask ``Series`` and single column ``DataFrame``s by resetting their index, turning it into a dimension. It should not have any backwards compatibility issues, except that constructing a ``Dataset`` from a single column ``DataFrame`` will now use the index and make it a dimension, while previously it would have just had the single column as a dimension. In all other cases it would have previously errored.

### DIFF
--- a/holoviews/core/data/dictionary.py
+++ b/holoviews/core/data/dictionary.py
@@ -110,7 +110,7 @@ class DictInterface(Interface):
 
     @classmethod
     def validate(cls, dataset, vdims=True):
-        dim_types = 'key' if vdims else 'all'
+        dim_types = 'all' if vdims else 'key'
         dimensions = dataset.dimensions(dim_types, label='name')
         not_found = [d for d in dimensions if d not in dataset.data]
         if not_found:
--- a/holoviews/core/data/pandas.py
+++ b/holoviews/core/data/pandas.py
@@ -56,6 +56,16 @@ class PandasInterface(Interface):
                              if d not in kdims]
             elif kdims == [] and vdims is None:
                 vdims = list(data.columns[:nvdim if nvdim else None])
+
+            # Handle reset of index if kdims reference index by name
+            for kd in kdims:
+                if isinstance(kd, Dimension): kd = kd.name
+                if kd in data.columns:
+                    continue
+                if any(kd == ('index' if name is None else name)
+                       for name in data.index.names):
+                    data = data.reset_index()
+                    break
             if any(isinstance(d, (np.int64, int)) for d in kdims+vdims):
                 raise DataError("pandas DataFrame column names used as dimensions "
                                 "must be strings not integers.", cls)
--- a/holoviews/core/data/pandas.py
+++ b/holoviews/core/data/pandas.py
@@ -35,7 +35,11 @@ class PandasInterface(Interface):
         element_params = eltype.params()
         kdim_param = element_params['kdims']
         vdim_param = element_params['vdims']
+        if util.is_series(data):
+            data = data.to_frame()
         if util.is_dataframe(data):
+            if eltype._auto_indexable_1d and len(data.columns) == 1:
+                data = data.reset_index()
             if isinstance(kdim_param.bounds[1], int):
                 ndim = min([kdim_param.bounds[1], len(kdim_param.default)])
             else:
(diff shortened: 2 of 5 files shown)


jbednar:
This is really helpful, thanks!  Given that the output of so many Pandas operations is a single-column dataframe or series with an index, we can now plot these without having to reset the index or specify the dimensions explicitly.

philippjfr:
So I've got some ideas on how to improve this a bit further. I'd suggest that if the user references an index as a key dimension by name we could also reset the index so it can be used by the interface. For dataframes with an index column with no name you should also simply be able to reference it by supplying 'index' as the key dimension.

So you might do something like this, a dataframe with three columns of which you want to plot one:

```python
df = pd.DataFrame({'A': range(10), 'B': np.random.rand(10), 'C': np.random.randn(10)})
hv.Curve(df, 'index', 'A')
```

This also got me thinking about "wide" datasets and what the best way would be to let the user get them into HoloViews. I feel fairly strongly that data should generally be tidy when working with HoloViews, since each column in a wide dataset does not correspond to a value dimension, instead the column names represent values along a secondary dimension. I'd therefore suggest we provide an easy way to "melt" a dataset. ``pd.melt`` provides very useful functionality to make datasets tidy, but it's always confused me. By adding a simple ``.melt`` method to datasets I think we could make things much more intuitive, e.g. take the example above:

```python
df = pd.DataFrame({'A': range(10), 'B': np.random.rand(10), 'C': np.random.randn(10)})
hv.Curve(df, 'index').melt()
```

image

It's gone from a wide to a tiny format and given the value dimension the name ``'Value'`` and the columns have become individual Curve indexed by the 'Group' dimension. Now let's look at a more realistic example, here's a mock stock timeseries datasets with three columns:

```
df = pd.DataFrame(np.random.randn(30, 3), columns=['MSFT', 'IBM', 'AAPL'],
                  index=pd.date_range('2017-01-01', '2017-01-30', name='Date')).cumsum()

hv.Curve(df, 'Date').melt('Close ($)', 'Stock')
```

bokeh_plot

This time we've given the index, value and group dimensions actual names.

jbednar:
>  I'd suggest that if the user references an index as a key dimension by name we could also reset the index so it can be used by the interface. For dataframes with an index column with no name you should also simply be able to reference it by supplying 'index' as the key dimension.

Sounds good!  

> I'd therefore suggest we provide an easy way to "melt" a dataset.

Your examples look good, and I think they will help address #2162.  But it's hard for me to see what ``hv.Curve(df, 'Date')`` would do *before* the melting; is that a meaningful object?  I was expecting to see you propose for .melt to be a method on hv.Curve, not on hv.Dataset.

philippjfr:
>But it's hard for me to see what hv.Curve(df, 'Date') would do before the melting; is that a meaningful object?

It would display just the first column, so I do think it's meaningful. It's even useful for gridded data where you might have multiple value arrays in your ``xr.Dataset``.

```python
hv.Image(xrds).melt()
```

>I was expecting to see you propose for .melt to be a method on hv.Curve, not on hv.Dataset.

Assume you meant a method on Dataset not Curve? Anyway, (almost) all elements except annotations are ``Dataset`` subclasses, so yes it is a method on Dataset and Curve.

jbednar:
Hmm.  It seems more natural to me to do that sort of data manipulation *before* the object is visualizable, but I'm not sure.  Can you make an example where you melt an hv.Dataset, then display it as a table, and only then make a Curve out of it, so that it's clear what the melting did?

philippjfr:
These two would be equivalent:

```python
hv.Dataset(df, 'index').melt().to(hv.Curve, 'index').overlay()
```

and:

```python
hv.Curve(df, 'index').melt().overlay()
```

jbednar:
Makes sense, thanks!  Might be good to introduce it that way in the docs so that , then point out that the second way is more convenient and does the same thing.  Shouldn't melt() just assume that it should use the index column as the kdim, as the default?

jlstevens:
I definitely support this feature though I would need to think carefully about how intuitive this behavior would be. The goal is to have things 'just work' without messing around people who might be making certain assumptions about the index on their dataframes.

Whatever we do, we need a new user guide talking about tidy vs wide formats and the melt method when it is introduced.

philippjfr:
>The goal is to have things 'just work' without messing around people who might be making certain assumptions about the index on their dataframes.

More complex and automatic handling of pandas indexes is going to have to wait a while longer as it has potential backward compatibility implications. For now this will allow using an index if supplied and automatically using it only in cases where it would have errored before. I'd be happy to rework the existing user guides with material on working with different datasets.

philippjfr:
Let's deal with my suggestion for a ``melt`` method in another PR. For now I think we should merge this immediately.

jlstevens:
Looks good and I agree it should be merged ASAP, just to get some real world testing.

It would be good to update the docs about this in the PR implementing the ``melt`` method.

Merging.

PR #2166, opened 2017-12-01
WIP: js fixups

ceball:


jbednar:
Looks good to me! Much uglier, but if it's not officially allowed and only supported by some browsers, seems like the ugly way is the only way.

ceball:
It's one way. The other way is to write e.g. es2015 and 'compile to js' with e.g. babel.

ceball:
Also, I expected this PR to be a misc collection of fixes like the one commit so far. However, I only found one fix was necessary for the code I'm trying to run, so I guess this is it for the PR - but there may be other cases still left to find.

philippjfr:
Should I go ahead and merge then?

ceball:
That would be great, although I have personally only tested the result of the change on IE11.

philippjfr:
Looks safe to me.

PR #2169, opened 2017-12-02
Allow defining CompositeElementPlot glyph order

philippjfr:
Previously the glyph ordering was not applied correctly, which is causing issues in https://github.com/ioam/holoviews/pull/2155. I've now fixed that and moved the ordering out to a separate function with a doctest.


philippjfr:
Tests passed and this is a bug fix holding up other development work.

jlstevens:
We never really decided on a policy on doctests. I am certainly in favor of them when they make sense and don't dominate the docstring. Certainly not complaining about the usage here which does seems appropriate...

PR #2171, opened 2017-12-03
Provide control over bokeh aspect

philippjfr:
This PR adds support for exposing the ``match_aspect`` option in bokeh as requested in https://github.com/ioam/holoviews/issues/1953. While it isn't ideal it provides at least some control over the screen vs. data space aspect. The main problem being that it doesn't work when explicitly set ranges are used, so for now it simply skips setting the holoviews computed ranges and relies on bokeh's data range calculations. This is mostly useful for plots like this https://github.com/ioam/holoviews/pull/2155, which require a fixed screen vs. data range aspect ratio.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -49,6 +49,13 @@ legend_dimensions = ['label_standoff', 'label_width', 'label_height', 'glyph_wid
 
 class ElementPlot(BokehPlot, GenericElementPlot):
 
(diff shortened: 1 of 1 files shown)


jlstevens:
I am optimistically assigning this to the next milestone though I believe it might require a fair bit of discussion.

philippjfr:
Closing since stale and not working, new solution will build on the layout work about to be merged into bokeh.

PR #2174, opened 2017-12-04
Allow dynamically adding new subplots

philippjfr:
This PR allows OverlayPlots to dynamically add new subplots when a DynamicMap returns more elements than were present on initialization. So far this is very much WIP and requires thorough testing. It does seem to work based on initial testing and should be a huge generalization of DynamicMaps.

Addresses: https://github.com/ioam/holoviews/issues/2097, https://github.com/ioam/holoviews/issues/1388, https://github.com/ioam/holoviews/issues/493

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -17,7 +17,7 @@ except ImportError:
     LogColorMapper, ColorBar = None, None
 from bokeh.plotting.helpers import _known_tools as known_tools
 
-from ...core import DynamicMap, CompositeOverlay, Element, Dimension
+from ...core import DynamicMap, CompositeOverlay, Element, Dimension, Overlay
 from ...core.options import abbreviated_exception, SkipRendering
 from ...core import util
 from ...streams import Buffer
@@ -628,6 +628,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         allowed_properties = glyph.properties()
         properties = mpl_to_bokeh(properties)
         merged = dict(properties, **mapping)
+        legend = merged.pop('legend', None)
         for glyph_type in ('', 'selection_', 'nonselection_', 'hover_', 'muted_'):
             if renderer:
                 glyph = getattr(renderer, glyph_type+'glyph', None)
@@ -635,6 +636,12 @@ class ElementPlot(BokehPlot, GenericElementPlot):
                 continue
             filtered = self._filter_properties(merged, glyph_type, allowed_properties)
             glyph.update(**filtered)
+
+        if legend is not None:
+            for leg in self.state.legend:
+                for item in leg.items:
+                    if renderer in item.renderers:
+                        item.label = legend
 
 
     def _postprocess_hover(self, renderer, source):
@@ -1449,9 +1456,11 @@ class OverlayPlot(GenericOverlayPlot, LegendPlot):
                     el = element
                 # If not batched get the Element matching the subplot
                 elif element is not None:
-                    idx = dynamic_update(self, subplot, k, element, items)
+                    idx, spec, exact = dynamic_update(self, subplot, k, element, items)
                     if idx is not None:
                         _, el = items.pop(idx)
+                        if not exact:
+                            self._update_subplot(subplot, spec)
 
                 # Skip updates to subplots when its streams is not one of
                 # the streams that initiated the update
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -17,7 +17,7 @@ except ImportError:
     LogColorMapper, ColorBar = None, None
 from bokeh.plotting.helpers import _known_tools as known_tools
 
-from ...core import DynamicMap, CompositeOverlay, Element, Dimension
+from ...core import Store, DynamicMap, CompositeOverlay, Element, Dimension, Overlay
 from ...core.options import abbreviated_exception, SkipRendering
 from ...core import util
 from ...streams import Buffer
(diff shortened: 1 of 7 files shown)


philippjfr:
Here is a simple test case:

```python
def cb(X):
    return hv.Overlay([hv.Curve(np.arange(10)+i) for i in range(X)])
hv.DynamicMap(cb, kdims=['X']).redim.range(X=(1, 10))
```

jbednar:
Fabulous, thanks!

jlstevens:
Very happy to see we will be able to support this finally.

I agree this needs thorough review though maybe it might be better to merge to master (after 1.9.2) ASAP to make sure it is thoroughly tested before the next release? I suspect there are only so many issues we will be able to uncover if we leave this feature in this PR too long...

philippjfr:
Agreed, I'll write some unit tests as soon as I have a minute and then we can hopefully merge this. I don't think it should change anything for things that were already supported but I expect there may be some things that could be improved when adding new subplots.

jlstevens:
It might also be worth adding a few examples to the docs to demonstrate that the old constraint is now removed. I'm not quite sure where though...

basnijholt:
This is (almost) exactly what I need 👍 

In [adaptive](https://github.com/python-adaptive/adaptive) we have a live plotting functionality and depending on the data that is returned we need to decide how to plot it. Sometimes we for example need a `Path` and othertimes a `Scatter`, however before any data is known a user might have already called the `live_plot`, so we might need to change the plot type.

The logic is in the following code:
```python
import holoviews as hv

def plot():
    while True:
        if path:
            yield hv.Path([])
        else:
            yield hv.Scatter([])

path = True
dm = hv.DynamicMap(lambda: plot(path), streams=[hv.streams.Stream.define('Next')()])
dm
```

```python
dm.event()  # works
path = False
dm.event()  # should plot a Scatter, but errors
```

With this PR a `DynamicMap` only allows `Overlay`s to be changed, so my example would work if we do:
```python
        if path:
            yield hv.Overlay([hv.Path([])])
        else:
            yield hv.Overlay([hv.Scatter([])])
```
which seems a bit unnessecary.

basnijholt:
Just to strengthen my point, I would like to describe how we solve the above issue at the moment.

We do it by creating an overlay:
```python
    def plot(self):
        if not self.data:
            return hv.Scatter([]) * hv.Path([])
        elif self.vdim > 1:
            return hv.Scatter(self.data) * hv.Path([])
        else:
            xs = list(self.data.keys())
            ys = list(self.data.values())
            return hv.Path((xs, ys)) * hv.Scatter([])
```
This leads to all kinds of annoying problems, like having to access attributes as `learner.plot().Scatter.I.opts(style=...)` instead of just `learner.plot().opts(style=...)`.

Or for example this issue https://github.com/ioam/holoviews/issues/2347. 

Allowing plot types to be changed in a `DynamicMap` would make life easier :)

philippjfr:
Due to a regression in bokeh this is once again blocked: https://github.com/bokeh/bokeh/issues/7590

philippjfr:
I'm sort of tempted to go ahead with this despite the bokeh regression. If a user doesn't make use of the functionality there is no change and if they do they lose the toolbar but otherwise should see no issue. Since this is a feature this would have to wait on an 1.11 release, while bokeh could fix the underlying issue quite soon.

basnijholt:
Then I would say to just include it! 😁

jlstevens:
I suppose if the release notes/changelog explicitly mention the problem with the toolbar disappearing, I would also be inclined to include it sooner rather than later.

basnijholt:
I see you amended there commit recently, did you by any chance address my comment https://github.com/ioam/holoviews/pull/2174#comment-366222952 (sorry, I'm AFK so can't test it.)

philippjfr:
@basnijholt For some reason I can't tell exactly which comment you're referring to but what you describe in the comments above is exactly what this PR should address.

philippjfr:
Okay the main reason I'm +1 on merging this is that it doesn't really affect cases that were previously already supported at all, so will not cause any regressions for existing use cases. The release notes should clearly point out the limitations with bokeh but otherwise I think this is in good shape.

In addition to the new functionality I have also improved existing functionality. OverlayPlots already supported remapping existing subplots if they were not being used, while this worked it also had some bugs, e.g. legend and the cyclic_index were not updated appropriately resulting in the wrong color being used and the legend not updating. I've now addressed this and added unit tests for this case as well.

philippjfr:
Ready to review.

philippjfr:
I see, I believe this will separately be fixed by https://github.com/ioam/holoviews/pull/2286.

basnijholt:
I just tried rebasing on the branch of https://github.com/ioam/holoviews/pull/2286 but I get the same error.

philippjfr:
Damn, thanks for checking I'll have a look.

philippjfr:
Sorry I should have read your question more clearly, wrapping the elements in an Overlay to be able to change the types will indeed continue to be required. I'd be happy to potentially address this limitation by **always** wrapping individual elements in an Overlay, but that's definitely out of scope for this PR, as it's a far more wide-reaching change with the potential for a lot of unintended side-effects.

jlstevens:
Looks ok. Ready to merge when the tests pass?

philippjfr:
Yes, let's go ahead. It's a relatively safe change on its own, we just need to remember to mention the caveats in the release announcement and I'll open an issue to continue discussion on the issue @basnijholt mentioned.

philippjfr:
Have to rebase and start tests over unfortunately, the colormap test changes causing issues.

philippjfr:
I've opened the issue now, and tests appear to be passing.

jlstevens:
Merging.

PR #2181, opened 2017-12-08
Fixed bug when using mpl cmap instance

philippjfr:
In a previous PR I introduced a small regression in colormap handling which resulted in matplotlib colormap instances not being handled anymore. The PR fixes the issue and adds a test to avoid future regression.


philippjfr:
Straightforward bug fix plus unit test, going to merge.

PR #2185, opened 2017-12-08
Issue template

ceball:
Please edit! This is just to get it started...

### DIFF
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,0 +1,60 @@
+Please edit either the 'Bug report' or 'Feature request' template
+below, attempting to fill in as much info as possible (and deleting
+the other). Thanks!
+
+# Feature request
+
+If you are proposing a feature, it's great if you can include as much
+of the following as possible, but we are always happy to discuss even
+brief feature requests to flesh them out. (We may already want and
+have thought about how to implement the feature you are proposing!)
+
+  * Motivation - what are you trying to enable? can you already do it,
+    but in an awkward way? etc
+
+  * Proposed outline of an implementation
+
+  * ...
+
+
+# Bug report
+
+## Description of the problem
+
+Things to consider including:
+
+  * What went wrong?
+
+  * What did you expect to happen?
+
+  * A screen shot or short screen cast.
+
+## Errors and warnings
+
+If relevant, please include:
+
+  * Python tracebacks and/or warnings.
+
+  * Any messages/tracebacks in your browser's developer tools console.
+
+## How to reproduce
+
+Ideally, please include a complete, minimal, self-contained python
+script or notebook that reproduces the issue.
+
+## Your environment
+
+(very draft)
+
+  * Output of `pyver python holoviews bokeh matplotlib notebook
+    ipython datashader pandas xarray conda TODO` (`pip install pyver`
+    if you do not already have it) ? or `pip freeze` or `conda env
+    export + conda info?` or ?
+
+  * Or maybe separately, your package manager and version?
+
+  * Your operating system and version.
+
+  * If relevant, your browser and version.
+
+
--- a/.github/issue_template.md
+++ b/.github/issue_template.md
@@ -1,6 +1,15 @@
 Please edit either the 'Bug report' or 'Feature request' template
 below, attempting to fill in as much info as possible (and deleting
-the other). Thanks!
+the other).
+
+If pasting code and or/text output, please enclose it in backticks, as
+shown below:
+
+```
+your code goes here
+```
+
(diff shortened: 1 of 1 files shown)


ceball:
In particular, I'm not sure of the best way (if there is one way?) to have someone describe their environment.

jbednar:
>* Output of `pyver python holoviews bokeh matplotlib notebook
 +    ipython datashader pandas xarray conda TODO` (`pip install pyver`
 +    if you do not already have it) ? or `pip freeze` or `conda env
 +    export + conda info?` or ?

jupyter-notebook instead if ipython?

ceball:
I probably meant `ipykernel`, not `ipython` (I already listed `notebook`). I'm not sure if there's one package you can use to get enough info about someone's jupyter system (any ideas, anyone?).

But really that was just a list that came to my fingers as I typed, as a demo of a specific list of packages to ask about. I'm not sure if it's best to ask for the versions of specific packages that matter for holoviews, or just to ask for the whole environment - and I'm not sure of the best way to ask for either of those things!

jlstevens:
I think we should aim to keep this template short to encourage contribution. Maybe we can link to a wiki page for suggestions but I would keep the main message focused as possible. The main ones I think of is to 1) suggest gitter for general questions 2) request short reproducible examples for bug reports.

jbednar:
Sure, but asking for a list of at least some versions is crucial for things to be reproducible, and it also helps people realize that they are using old versions, thereby fixing their own problems.

jlstevens:
Sure, three things then! I would just avoid front-loading too much text and point to the wiki for details.

ceball:
Just some thoughts:

* I totally agree about keeping it as short as possible, and that the proposal is currently too long. However, I also think that pointing to a wiki is probably pointless (most people will not read it).

* Obviously it's up to you how you want to provide support (thanks for providing support!), so if you prefer to interactively interrogate on gitter rather than request that users do certain things before reporting, that's your choice. I personally would be able to answer some questions if only more information were provided up front, but obviously what you prefer is more important as you are the main providers of support! I also think github is a better (more visible, more easily searchable) record.

* A short reproducible example is still something you (or other/future viewers of the issue) would have to run, while a screenshot/cast can often save you from that.

jlstevens:
>  ... I also think that pointing to a wiki is probably pointless (most people will not read it).

I think that is also true of an overlong template! (most people not reading it)

jlstevens:
I think we can shorten it much further:

```diff
+  * What went wrong, or what were you unable to do?
 +
 +  * What did you expect or hope to happen?
```

To me, these questions are so general they don't mean much but maybe that is because I am used to filing issues! Even so, I have rarely encountered an issue where the person hasn't expressed these things.

 ```
 +  * A screen shot, short screen cast, or sketch.
 ```

Helpful, I agree. The next two sections are also helpful but I don't think they need to be sections.

I would be happy with just two paragraphs:

> Please consider including a screenshot or short screencast when describing your issue and include any relevant Python tracebacks and/or warnings. If you don't see a clear Python error, please report any warnings/error messages you see in your web browser's developer tools console. [LINK here?]

> In order for us to address the issue, we need to be able to reproduce the problem. Ideally, please include a minimal, self-contained example that illustrates the issue you are encountering. Finally, please include the holoviews, matplotlib and/or bokeh version you are using as well as your operating system version and browser.

If the holoviews/bokeh/mpl versions are reported as requested, I'm sure we can address most issues without filling up the issue with  reams of `conda env export`, `conda info` and  `pip freeze` output.

philippjfr:
Any chance we can finish this up soon? I agree we should make it shorter, @jlstevens do you want to take this over?

jlstevens:
I was wondering if we should ask people only to report bugs against tagged versions, especially now we plan to tag and build a lot more frequently. That said, PRs should only be merged with passing tests so master should be kept in a good state which means we do want bug reports against master that can be at any point in the history.

Anyway template would be a good place to ask for the version and point people to the install instructions so that pip users use pip + git clone and not archive. Note that this would be recommended even if we were using that *other* versioning approach (that I shall not name!).

jbednar:
Seems like we want bug reports against latest release or against master, to me.  Both those are useful; any other version from the past seems less useful.

jlstevens:
Sometimes it is nice to get feedback from users willing to check out a PR though...but that should normally be comments on the PR as opposed to a new issue. So I agree with that...

philippjfr:
Going to close, issue templates can now be created via the GitHub UI. @jsignell will likely add the same templates used by bokeh in our triaging session over the next couple of weeks.

PR #2186, opened 2017-12-11
Improved error message for DataArray without name

philippjfr:
Improves error message when a DataArray without a name is supplied and the default value dimension name clashes with a coordinate. Addresses https://github.com/ioam/holoviews/issues/2147.

### DIFF
--- a/holoviews/core/data/xarray.py
+++ b/holoviews/core/data/xarray.py
@@ -50,9 +50,18 @@ class XArrayInterface(GridInterface):
                 vdim = vdims[0]
             elif len(vdim_param.default) == 1:
                 vdim = vdim_param.default[0]
+                if vdim.name in data.dims:
+                    raise DataError("xarray DataArray does not define a name, "
+                                    "and the default of '%s' clashes with a "
+                                    "coordinate dimension. Give the DataArray "
+                                    "a name or supply an explicit value dimension."
+                                    % vdim.name, cls)
             else:
-                raise DataError("If xarray DataArray does not define a name "
-                                "an explicit vdim must be supplied.", cls)
+                raise DataError("xarray DataArray does not define a name "
+                                "and %s does not define a default value "
+                                "dimension. Give the DataArray a name or "
+                                "supply an explicit vdim." % eltype.__name__,
+                                cls)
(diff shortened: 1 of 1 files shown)


holoviews/core/data/xarray.py
```diff
@@ -50,9 +50,18 @@ def init(cls, eltype, data, kdims, vdims):
                 vdim = vdims[0]
             elif len(vdim_param.default) == 1:
                 vdim = vdim_param.default[0]
+                if vdim.name in data.dims:
+                    raise DataError("xarray DataArray does not define a name, "
+                                    "and the default of '%s' clashes with a "
+                                    "coordinate dimension. Give the DataArray "
+                                    "a name or supply an explicit value dimension."
+                                    % vdim.name, cls)
```
jlstevens: Are you sure this works? I only see one ``%s`` and two things after ``%``. I also prefer an explicit tuple after ``%``.

holoviews/core/data/xarray.py
```diff
@@ -50,9 +50,18 @@ def init(cls, eltype, data, kdims, vdims):
...
             else:
-                raise DataError("If xarray DataArray does not define a name "
-                                "an explicit vdim must be supplied.", cls)
+                raise DataError("xarray DataArray does not define a name "
+                                "and %s does not define a default value "
+                                "dimension. Give the DataArray a name or "
+                                "supply an explicit vdim." % eltype.__name__,
+                                cls)
```
jlstevens: Same comment as above.

holoviews/core/data/xarray.py
philippjfr: Yes, I'm sure. ``cls`` is the second argument to the error. Personally I don't like using superfluous parentheses when only one formatting argument is used.

jbednar: Ah. I favor having the parentheses in a case like this, as I don't think everyone has memorized the precedence of %.

jlstevens: Ah, didn't notice this was in the context of ``DataError``. Thanks for clarifying.

jlstevens:
Looks good. Merging.

PR #2187, opened 2017-12-11
Box whisker improvements

philippjfr:
Improved BoxWhisker element docstrings as requested in https://github.com/ioam/holoviews/issues/1838 and fixed hover issues as requested in https://github.com/ioam/holoviews/issues/2142.

### DIFF
--- a/holoviews/element/chart.py
+++ b/holoviews/element/chart.py
@@ -137,8 +137,14 @@ class Bars(Chart):
 
 class BoxWhisker(Chart):
     """
-    BoxWhisker represent data as a distributions highlighting
-    the median, mean and various percentiles.
+    BoxWhisker allows representing the distribution of data grouped
+    into one or more groups by summarizing the data using quartiles.
+    The boxes of a BoxWhisker element represent the first, second and
+    third quartiles. The whiskers follow the Tukey boxplot definition
+    representing the lowest datum still within 1.5 IQR of the lower
+    quartile, and the highest datum still within 1.5 IQR of the upper
+    quartile. Any points falling outside this range are shown as
+    distinct outlier points.
     """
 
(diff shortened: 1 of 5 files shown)


holoviews/element/chart.py
```diff
@@ -137,8 +137,14 @@ class Bars(Chart):
...
-    BoxWhisker represent data as a distributions highlighting
-    the median, mean and various percentiles.
+    BoxWhisker allows representing the distribution of data grouped
+    into one or more groups by summarizing the data using quartiles.
+    The boxes of a BoxWhisker element represent the first, second and
+    third quartiles. The whiskers follow the Tukey boxplot definition
+    representing the lowest datum still within 1.5 IQR of the lower
+    quartile, and the highest datum still within 1.5 IQR of the upper
+    quartile. Any points falling outside this range are shown as
+    distinct outlier points.
```
jlstevens: maybe 'standard Tukey boxplot definition' if it is standard? Otherwise sounds like it is just *a* definition for boxplots...

jlstevens:
One minor docstring comment. Otherwise looks good.

holoviews/element/chart.py
jlstevens: Might be worth expanding the name to 'John W. Tukey'...

philippjfr: Tukey box plots is the common way to refer to this style of plot, don't think the full name adds anything.

philippjfr: To quote Wikipedia:

> But the ends of the whiskers can represent several possible alternative values, among them:

>the lowest datum still within 1.5 IQR of the lower quartile, and the highest datum still within 1.5 IQR of the upper quartile (often called the Tukey boxplot)

jlstevens:
Looks good. Merging.

PR #2189, opened 2017-12-11
Added support for date types on Pointer derived streams

philippjfr:
This was already handled at the bokeh callback level but parameter validation meant that datetime types were not allowed on Pointer derived streams. This ensures that Pointer derived streams all consistently support the various types that might be returned by a bokeh callback.

Addresses https://github.com/ioam/holoviews/issues/2124.

### DIFF
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -14,6 +14,9 @@ from .core import util
 
 from contextlib import contextmanager
 
+# Types supported by Pointer derived streams
+pointer_types = (Number, util.basestring, tuple)+util.datetime_types
+
 
 @contextmanager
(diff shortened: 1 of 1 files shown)


holoviews/streams.py
```diff
@@ -14,6 +14,9 @@
 
 from contextlib import contextmanager
 
+# Types supported by Pointer derived streams
+pointer_types = (Number, util.basestring, tuple)+util.datetime_types
```
jlstevens: Super minor comment: spaces around operators like ``+`` are preferred unless you are clearly trying to save horizontal space.

jlstevens:
Looks good. Merging.

PR #2190, opened 2017-12-11
Allow empty Distribution and Bivariate elements

philippjfr:
Empty Distribution and Bivariate elements weren't handled correctly.

### DIFF
--- a/holoviews/operation/stats.py
+++ b/holoviews/operation/stats.py
@@ -192,7 +192,7 @@ class bivariate_kde(Operation):
         elif ymin == ymax:
             ymin, ymax = ymin-0.5, ymax+0.5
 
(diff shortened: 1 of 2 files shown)


jlstevens:
Looks good. Ideally we would automate testing of empty data across all elements as has been suggested before. Maybe there should be an issue about that though it will be easiest once *all* non-annotation elements use interfaces (including histogram types and raster).

philippjfr:
>Maybe there should be an issue about that though it will be easiest once all non-annotation elements use interfaces (including histogram types and raster).

Agreed.

philippjfr:
Here's the issue: https://github.com/ioam/holoviews/issues/2192

Issue #2129, opened 2017-11-16
Selection1D streams dont work on ndoverlay source?


suvarchal:
While trying to find a simple example to raise an issue with DynamicMap and streams (another issue), i went back to simplest example closest to my problem and turned out that the [example on curve selection doesnt work anymore?](http://build.holoviews.org/reference/streams/bokeh/curve_selection.html) on tap `selection` always returns  `None`.

jlstevens:
Sorry about the delay getting back to you and thanks for reporting this! It does seem like there is a bug here for us to fix...

philippjfr:
This should work now but I'm going to repurpose this issue to fix it more generally.

PR #2194, opened 2017-12-11
Ensure Compositor does not make superfluous copies

philippjfr:
A partial workaround/fix for https://github.com/ioam/holoviews/issues/2129. The correct fix will be to ensure that a Compositor transfers the ID a stream is registered to when it is applied but since we do not ship any actual Overlay compositors by default this fix should be sufficient for 1.9.2.


jlstevens:
Looks like the correct fix to me! I'll wait for the tests to pass anyhow...

jlstevens:
To clarify that I am not contradicting you, this is the right thing to do (don't touch the overlay unless the patterns apply) but it isn't the *complete* fix which is what #2129 is about.

jlstevens:
Tests have passed. Merging.

PR #2197, opened 2017-12-14
Matplotlib plotting fixes

philippjfr:
In recent versions of pandas the matplotlib unit conversions for pandas datetime types is no longer automatic, this ensures that if pandas is available they are registered with matplotlib. The PR also fixes a bug handling legends in Python 3.


philippjfr:
Will merge these simple fixes.

PR #2198, opened 2017-12-14
Small notebook fixes

philippjfr:
Minor fixes/improvements to existing notebooks.


jlstevens:
One thing I was meaning to do was to add some info about ``hv.help`` to the element notebooks. I am happy to do that in this PR once you are done with your fixes.

philippjfr:
I'm done for now, so go ahead.

jlstevens:
I'll do that now but because I will need to touch *all* the element notebooks I want to decide on the message now because I won't want to change it later.

1. Should the message be at the start or the end of the element notebooks? I am leaning towards the end.
2. Should the message be generic so it can copy/pasted or should it mention the element name? Latter is nicer but more effort (so I guess I'll do that?).

Assuming it is something I append to the element notebooks maybe something like this, e.g for ``Curve``:

To get additional help as well as a list of available style and plot options, use ``hv.help(hv.Curve)``.

Is that any good and do you think that would assist people?

jbednar:
I think at the end is easier to ignore once people get it, and it isn't the first thing they will want to do even if it's new.  I'd vote for:

> For full documentation and the available style and plot options, use ``hv.help(hv.Curve)``.

It would be nice if the output of ``hv.help(hv.Curve)`` could be linked to the text ``hv.help(hv.Curve)``, but that maybe too much work for today?

jlstevens:
> It would be nice if the output of hv.help(hv.Curve) could be linked to the text hv.help(hv.Curve), but that maybe too much work for today?

Not today, but I'll use the format you suggested, thanks!

jlstevens:
It was pretty tedious going but all the element notebooks are now updated with a pointer to ```hv.help`` at the end.

Happy to see this PR merged if @philippjfr agrees it is ready.

philippjfr:
Thanks, looks good but tedious indeed. Will merge now.

PR #2199, opened 2017-12-14
Fixed bug validating stats elements in PandasInterface

philippjfr:
The PandasInterface was trying to validate the value dimensions which do not exist on stats elements like Distribution and Bivariate causing errors.

### DIFF
--- a/holoviews/core/data/pandas.py
+++ b/holoviews/core/data/pandas.py
@@ -108,7 +108,7 @@ class PandasInterface(Interface):
 
     @classmethod
     def validate(cls, dataset, vdims=True):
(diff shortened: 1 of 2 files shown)


jlstevens:
If I remember right, stats elements 'fake' their vdims to get things to work. Isn't there something that could be faked on stat elements to get it to work as before?

Otherwise, looks like a relatively simple fix and the four new unit tests seem to be passing.

philippjfr:
You originally suggested this workaround for stats elements, I just messed up the logic implementing it.

philippjfr:
Going to merge.

PR #2200, opened 2017-12-14
Fixed bug in bokeh BarPlot colormapping integer values

philippjfr:
When using a CategoricalColorMapper integer values need to be transformed. The logic to apply this transformation was incorrect leading to BokehJS errors in certain cases.


philippjfr:
Will go ahead and merge since the change is tested.

PR #2206, opened 2017-12-15
Implement precomputing of datashader input data

philippjfr:
An initial go at https://github.com/ioam/holoviews/issues/2196. It allows datashader operation to precompute results, which can be used by the operation. For now only implemented for TriMesh rasterization but should be straightforward enough to do for other aggregation (although unnecessary for rasters).

I'm still considering how to improve this further. I think memoizing the precomputed results based on the element ID would be sufficient but will continue thinking about it.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -12,6 +12,7 @@ import xarray as xr
 import datashader as ds
 import datashader.transfer_functions as tf
 import dask.dataframe as dd
+from param.parameterized import bothmethod
 
 ds_version = LooseVersion(ds.__version__)
 
@@ -89,29 +90,16 @@ class ResamplingOperation(Operation):
         Disable when you do not want the resulting plot to be interactive,
         e.g. when trying to display an interactive plot a second time.""")
 
-    precompute = param.Boolean(default=True, doc="""
+    precompute = param.Boolean(default=False, doc="""
        Whether to apply precomputing operations, if available.
        Precomputing is in some cases useful to avoid repeating
        expensive operations on the same inputs.""")
 
-    _precomputed = None
-
-
-    def __call__(self, element, **params):
-        if params.get('precompute', self.precompute):
-            self._precomputed = self._precompute(element)
-        return super(ResamplingOperation, self).__call__(element, **params)
-
-
-    def _precompute(self, obj):
-        """
-        Precompute the input to an operation returning data that will
-        be held on the _precomputeed attribute of the operation instance.
-        Allows pre-computing data for operations that will be executed
-        multiple times, allowing caching of certain results.
-        """
-        return None
-
+    @bothmethod
+    def instance(self_or_cls,**params):
+        inst = super(ResamplingOperation, self_or_cls).instance(**params)
+        inst._precomputed = {}
+        return inst
 
     def _get_sampling(self, element, x, y):
         target = self.p.target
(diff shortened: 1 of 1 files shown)


jlstevens:
I know it isn't ready for review yet but I will say that at a glance, this looks like a very reasonable approach.

philippjfr:
I'm now using the ``element._plot_id`` to memoize the precomputed data on as that's specifically used in the plotting code to denote static data.

holoviews/operation/datashader.py
```diff
@@ -502,27 +519,41 @@ class trimesh_rasterize(aggregate):
                                          objects=['bilinear', None], doc="""
         The interpolation method to apply during rasterization.""")
 
+    def _precompute(self, element):
+        from datashader.utils import mesh
```
jlstevens: Why the inline import?

holoviews/operation/datashader.py
```diff
@@ -567,21 +598,27 @@ def _process(self, element, key=None):
...
                                and not issubclass(x.type, Image))
         element = element.map(dsrasterize, predicate)
+        self._precomputed = trirasterize._precomputed
 
         # Rasterize other Dataset types
         predicate = lambda x: (isinstance(x, Dataset) and
                                (not isinstance(x, Image) or x in imgs))
         element = element.map(dsrasterize, predicate)
+        self._precomputed = trirasterize._precomputed
+
```
jlstevens: If any more of these appear I would recommend a context manager.

holoviews/operation/datashader.py
```diff
@@ -89,6 +90,17 @@ class ResamplingOperation(Operation):
         Disable when you do not want the resulting plot to be interactive,
         e.g. when trying to display an interactive plot a second time.""")
 
+    precompute = param.Boolean(default=False, doc="""
+       Whether to apply precomputing operations, if available.
```
jlstevens: Why 'if available'?

jlstevens: When do you *not* want to precompute? I assume the default is False because precomputation has disadvantages?

philippjfr: Sure, caching always has disadvantages, namely increased memory usage.

jlstevens: Ok, then the docstring should mention that.

philippjfr: >Why 'if available'?

regridding doesn't have or need precomputing.

holoviews/operation/datashader.py
philippjfr: Requires a very specific version of datashader. Could move it to the top in a try/except, up to you.

holoviews/operation/datashader.py
jlstevens: I'm also confused, this is not the sort of caching I was thinking of. I thought you cache from the previous call and if it is the same element, the data is available. If the data changes, then the cache is invalid and cleared.

Maybe you mean the operation *holds* onto memory between calls but from what I can tell, the transient memory usage when the operation is being applied should be the same.

philippjfr: >I'm also confused, this is not the sort of caching I was thinking of. I thought you cache from the previous call and if it is the same element, the data is available. If the data changes, then the cache is invalid and cleared.

That's correct.

>Maybe you mean the operation holds onto memory between calls but from what I can tell, the transient memory usage when the operation is being applied should be the same.

Correct.

jlstevens: I think the user doesn't need to know about regridding needing/not needing precomputation - whatever setting there is can be safely ignores so I wouldn't mention it, just to simplify the docstring and minimize confusion.

holoviews/operation/datashader.py
jlstevens: Ok. That is fine I think.

holoviews/operation/datashader.py
philippjfr: Sounds good.

jlstevens: Maybe something like:

> Whether to apply precomputing operations. Precomputing can speed up resampling operations by avoiding recomputation when the supplied element does not change at the cost of holding onto the memory used for this internal state between calls.

jlstevens:
Other than clarifying the docstring, this PR looks good to me. Happy to merge once it is updated.

jlstevens:
Python 2 builds have passed and we are having some issues with conda so I won't wait for the Python 3 ones. Merging.

PR #2210, opened 2017-12-18
Addition of Bokeh theme support to BokehRenderer

jlstevens:
On current bokeh master, a PR has been merged which (re)enables bokeh themes in the notebook: bokeh/bokeh#7325. Bokeh themes in general is something HoloViews should support for extra control over options that might not be exposed via plot/style options. In addition, we do support style files for matplotlib so it makes sense to support Bokeh themes as the analogous concept.

This PR shows one way this could work (you will need the latest bokeh master for this to work) using a theme file:

image

Or using a JSON spec (not a great theme but it gets the point across!):

image

I'm not committed to this API (though it seems ok to me) and I wonder if I should do a bokeh version check? This won't cause any errors/exceptions if you use an earlier bokeh version but the theming won't work either.

### DIFF
--- a/holoviews/plotting/bokeh/renderer.py
+++ b/holoviews/plotting/bokeh/renderer.py
@@ -23,6 +23,7 @@ from .util import attach_periodic, compute_plot_size
 from bokeh.io.notebook import load_notebook
 from bokeh.protocol import Protocol
 from bokeh.embed.notebook import encode_utf8, notebook_content
+from bokeh.themes.theme import Theme
 
 NOTEBOOK_DIV = """
 {plot_div}
@@ -33,6 +34,9 @@ NOTEBOOK_DIV = """
 
 
 class BokehRenderer(Renderer):
+
+    theme = param.ClassSelector(default=None, class_=Theme, allow_None=True, doc="""
+       The applicable Bokeh Theme object (if any).""")
 
     backend = param.String(default='bokeh', doc="The backend name.")
 
@@ -212,6 +216,11 @@ class BokehRenderer(Renderer):
         for m in model.references():
             m._document = None
         doc.add_root(model)
+
+        if self.theme:
+            from bokeh.io import curdoc
+            curdoc().theme = self.theme
+
         comm_id = plot.comm.id if plot.comm else None
         # Bokeh raises warnings about duplicate tools and empty subplots
         # but at the holoviews level these are not issues
--- a/holoviews/plotting/bokeh/renderer.py
+++ b/holoviews/plotting/bokeh/renderer.py
@@ -83,7 +83,7 @@ class BokehRenderer(Renderer):
         backend. The output is not a file format but a suitable,
         in-memory byte stream together with any suitable metadata.
         """
-        plot, fmt =  self._validate(obj, fmt)
+        plot, fmt =  self._validate(obj, fmt, doc=doc)
         info = {'file-ext': fmt, 'mime_type': MIME_TYPES[fmt]}
 
         if self.mode == 'server':
@@ -110,6 +110,7 @@ class BokehRenderer(Renderer):
             return '\n'.join(self_or_cls.html_assets()).encode('utf8')
         return
 
+
     @bothmethod
     def get_plot(self_or_cls, obj, doc=None, renderer=None):
         """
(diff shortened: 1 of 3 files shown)


holoviews/plotting/bokeh/renderer.py
```diff
@@ -212,6 +216,11 @@ def _figure_data(self, plot, fmt='html', doc=None, **kwargs):
         for m in model.references():
             m._document = None
         doc.add_root(model)
+
+        if self.theme:
+            from bokeh.io import curdoc
+            curdoc().theme = self.theme
```
philippjfr: Seems like this should be ``doc.theme`` not ``curdoc().theme``, but maybe I'm confused?

jlstevens: Well spotted! I had this comment in my code:

```
# Using curdoc because of conditional in embed.py _ModelInEmptyDocument
```

I'll try using ``doc`` again but I suspect it might not work for this reason - I forgot the details exactly. If ``doc`` does work, I also prefer using that.

jlstevens: I can also restore that comment if that helps.

jlstevens: Right, I just checked and using ``doc`` does not work.  I'm not sure which conditional I am referring to, maybe this one https://github.com/bokeh/bokeh/blob/master/bokeh/embed/notebook.py#L121?

Only in that branch does it look like the theme is transferred from one document to another in the way you expect from the bokeh API (setting the theme on the document). Perhaps there is a way to pass ``apply_theme`` into the ``_ModelInEmptyDocument`` context manager somehow but I would not know what that is.

philippjfr: Okay, I'll have a poke around as well. Want to at least make sure it works when a user supplies a custom Document, that's not equivalent to curdoc().

jlstevens: Good idea!

holoviews/plotting/bokeh/renderer.py
```diff
@@ -152,6 +157,8 @@ def app(self_or_cls, plot, show=False, new_window=False, websocket_origin=None):
             return doc
 
         def modify_doc(doc):
+            if self_or_cls.theme:
+                doc.theme = self_or_cls.theme
```
philippjfr: Don't think this is actually needed thinking about it again.

holoviews/plotting/bokeh/renderer.py
```diff
@@ -208,10 +215,17 @@ def server_doc(self_or_cls, obj, doc=None):
 
     def _figure_data(self, plot, fmt='html', doc=None, **kwargs):
         model = plot.state
-        doc = Document() if doc is None else doc
+        if plot.document:
+            doc = plot.document
+        else:
+            doc = Document() if doc is None else doc
```
philippjfr: This is internal API now and I don't think ``doc`` should be passed into this method at all now. We were already discussing providing a public API that does something similar. I'll have to look at some paramnb/parambokeh apps that use this codepath to figure out the best approach for migrating to the new API without breaking old apps.

jlstevens:
Your changes look good to me. Ready to merge?

holoviews/plotting/bokeh/renderer.py
jlstevens: That makes sense though I think that should be in a PR about refactoring, not one adding a new feature...

philippjfr:
No, I want to address the issues mentioned above and would like to see some documentation for applying themes.

jlstevens:
I am happy to write some docs in the 'Plotting with Bokeh' user guide. If I do that and it turns out the change to ``modify_doc`` can be removed, would you be happy to merge then, leaving the rest of the refactor for another PR?

jlstevens:
@philippjfr I have now added a theming section to the 'Plotting with Bokeh' user guide.

philippjfr:
Thanks, I'll not get to this today unfortunately. Should be able to resolve the issues this weekend though.

philippjfr:
Happy for this to be merged once tests are passing.

philippjfr:
I have no idea what's going on with the tests here, the PR has been rebased but push builds are failing.

jlstevens:
Odd. No idea what is up with the tests so I restarted the push builds to see if it fails in the same way again.

philippjfr:
Seems to have fixed itself.

jlstevens:
Great! I'll be happy to see this merged. :-)

PR #2218, opened 2017-12-19
Fixes for bokeh server widgets

philippjfr:
Fixes and unit tests for bokeh server widgets


philippjfr:
Added a test as well and the fix is straightforward. Merging.

PR #2219, opened 2017-12-19
Trisurface renamed to TriSurface

jlstevens:
The branch is misnamed after deciding that ``TriMesh`` was not the inconsistent name but ``Trisurface`` was.


jlstevens:
@philippjfr Not sure there is much more to this PR other than having you check that the plotly TriSurface/Trisurface works for you (I can't test it because of some odd webgl error!)

philippjfr:
There are demo notebooks, which haven't been updated:

http://holoviews.org/gallery/demos/matplotlib/trisurf3d_demo.html
http://holoviews.org/gallery/demos/plotly/trisurf3d_demo.html

jlstevens:
Thanks..I'll fix those shortly.

jlstevens:
That should be the rest of them. Happy to merge when the tests pass?

jlstevens:
There are two failures that look like they will require the test data to be updated.

philippjfr:
>There are two failures that look like they will require the test data to be updated.

Seems likely if you updated the ``group``.

jlstevens:
I have now updated the test data.

philippjfr:
PR build passing, merging.

PR #2222, opened 2017-12-20
Added double_pendulum example

philippjfr:
Adds double_pendulum examples as suggested in: https://github.com/ioam/holoviews/issues/1544

screen shot 2017-12-20 at 8 51 19 pm

pendulum

### DIFF
--- a/examples/gallery/demos/bokeh/double_pendulum.ipynb
+++ b/examples/gallery/demos/bokeh/double_pendulum.ipynb
@@ -1,0 +1,177 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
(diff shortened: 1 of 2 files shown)


jlstevens:
Looks good though for real usage I would recommend a dynamicmap for simulations instead of a holomap. For the website it has to be a holomap anyway but I think this might be worth mentioning in this example.

jbednar:
Could it be a DynamicMap cast to a HoloMap to make that clearer?

philippjfr:
Sure, I could sample a DynamicMap.

hoxbro:
I will close this PR and the corresponding issue; if it hasn't happened by now, it is very unlikely to happen.

Issue #2221, opened 2017-12-20
Add hv.HTML Element?


jbednar:
Linked streams in HoloViews make it feasible to make richly linked interactive apps, where hovering or clicking in one plot can control which information is shown in other plots.  E.g. the [Tap](http://holoviews.org/reference/streams/bokeh/Tap.html) example shows how a click on a heatmap can open an associated curve, and the [LandSat example](https://bokeh.github.io/datashader-docs/topics/landsat.html) shows how to do the same by just hovering.  This functionality is really powerful, but it's limited by the scope of what can be contained in a HoloViews element, and so what can be linked right now are just plots and tables.  

It could be very useful to be able to bring up arbitrary HTML information, if we add an HV element that can accept arbitrary HTML-renderable content as its .data.  Such content might include a textual description of a selected data point, a news article about that item, some web page links associated with it, and so on -- anything that can be put into a ``<div>`` and displayed along with the main plot.  

Bokeh should be able to handle arbitrary HTML like this, but the matplotlib renderer would not have an HTML parser, in which case it could possibly show the .data characters as a plot with a text annotation, just to give it some value under mpl as well.  It's not ideal, but exploiting the power of HTML for formatting and hyperlinking in the Bokeh case seems to override the goal of having the backends have similar functionality.

jlstevens:
I don't think I would ever approve of ``hv.HTML``but I would consider ``hv.Div`` given a good set of compelling examples. I would also make ``hv.Div`` a Bokeh only element analogous to how matplotlib/plotly have the 3D elements but bokeh does not.

jbednar:
What's the difference between hv.Div and hv.HTML, apart from div being more obscure (and slightly inaccurate, in that the data supplied need not be a div (as it will be placed *into* a div)?

jlstevens:
There is a bokeh ``Div`` model and not a bokeh ``HTML`` model and ``Div`` is a type of HTML node in the DOM whereas HTML is the markup language itself. We don't want to support HTML headers, body or script tags either.

philippjfr:
Here's my prototype: https://anaconda.org/philippjfr/div/notebook

jlstevens:
The prototype looks good: a relatively thin wrapper around Bokeh's ``Div`` model. 

[...]

PR #2223, opened 2017-12-21
Add Div Element

philippjfr:
As suggested in #2221 this PR adds a Div element wrapping around the bokeh Div model, which opens up all kinds of possibilities for nice demos.

- [x] Add plotting tests
- [x] Add element reference entry

### DIFF
--- a/examples/reference/elements/bokeh/Div.ipynb
+++ b/examples/reference/elements/bokeh/Div.ipynb
@@ -1,0 +1,108 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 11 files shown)


philippjfr:
Let's merge this now, we can add nice demos later.

jlstevens:
Looks good to me. Merging.

jlstevens:
Actually, are you planning to mention this in the user guide? Probably in the bokeh specific one?

jlstevens:
Anyway, that can be a separate task and assigned an issue if necessary. Merging.

PR #2233, opened 2017-12-30
Fix link to homepage in introduction

jbampton:


jlstevens:
Both www.holoviews.org and http://holoviews.org work for me when I click on those links. Was the former (www version) not working from the notebook/website?

Anyway, as the updated version is better regardless, I'll go ahead and merge. Thanks for the fix!

jbampton:
Yes it was not working on the notebook website.  Just returned a 404 not found.

PR #2238, opened 2018-01-03
Changed Image regular sampling validation

philippjfr:
As is evident in https://github.com/ioam/holoviews/issues/2228, the new Image sampling validation which relies on machine floating point precision does not work correctly. This PR switches the validation strategy to one that ensures that the maximum difference in the sampling is one billionth (1e9) of the smallest sampling step. This works consistently based on my now more thorough testing and is still a very strict constraint to ensure regular sampling.

### DIFF
--- a/holoviews/element/util.py
+++ b/holoviews/element/util.py
@@ -282,3 +282,21 @@ def connect_edges(graph):
         end = end_ds.array(end_ds.kdims[:2])
         paths.append(np.array([start[0], end[0]]))
     return paths
+
+
+def validate_regular_sampling(img, dimension, rtol=10e-9):
+    """
+    Validates regular sampling of Image elements ensuring that
+    coordinates the difference in sampling steps is at most rtol times
+    the smallest sampling step. By default ensures the largest
+    sampling difference is less than one billionth of the smallest
+    sampling step.
+    """
+    dim = img.get_dimension(dimension)
+    diffs = np.diff(img.dimension_values(dim, expanded=False))
+    vals = np.unique(diffs)
+    if len(vals) > 1 and np.abs(vals.min()-vals.max()) > diffs.min()*rtol:
+        raise ValueError("%s dimension %s is not evenly sampled, "
+                         "please use the QuadMesh element for "
+                         "unevenly or irregularly sampled data." %
+                         (type(img).__name__, dim))
--- a/holoviews/element/raster.py
+++ b/holoviews/element/raster.py
@@ -262,14 +262,17 @@ class Image(Dataset, Raster, SheetCoordinateSystem):
                 raise ValueError("Input array has shape %r but %d value dimensions defined"
                                  % (self.shape, len(self.vdims)))
 
-        xvals = np.unique(np.diff(self.dimension_values(0, expanded=False)))
-        if len(xvals) > 1 and np.abs(xvals.min()-xvals.max()) > sys.float_info.epsilon*100:
+        xdiffs = np.diff(self.dimension_values(0, expanded=False))
+        xvals = np.unique(xdiffs)
+        if len(xvals) > 1 and np.abs(xvals.min()-xvals.max()) > xdiffs.min()*10e-9:
             raise ValueError("%s dimension %s is not evenly sampled, "
                              "please use the QuadMesh element for "
                              "unevenly or irregularly sampled data." %
                              (type(self).__name__, self.get_dimension(0)))
-        yvals = np.unique(np.diff(self.dimension_values(1, expanded=False)))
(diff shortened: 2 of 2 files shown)


jlstevens:
Looks good although given this is a heuristic check that has already had to change once, wouldn't it be worth pulling this out into a utility?

philippjfr:
>Looks good although given this is a heuristic check that has already had to change once, wouldn't it be worth pulling this out into a utility?

Sure, happy to do that.

holoviews/element/util.py
```diff
@@ -282,3 +282,21 @@ def connect_edges(graph):
         end = end_ds.array(end_ds.kdims[:2])
         paths.append(np.array([start[0], end[0]]))
     return paths
+
+
+def validate_sampling(img, dimension, rtol=10e-9):
```
jlstevens: Not worth holding up the PR unless you agree, but I think ``check_regularly_sampled`` might be a better name...

jlstevens: Or ``check_regular_sampling``

philippjfr: I'll go with a mixture, ``validate_regular_sampling``

ea42gh:
can't wait to try it: I currently have plots
that no longer work...

jlstevens:
The failing test looks like a transient failure. Merging.

stonebig:
Sorry if I raise a wrong alarme, if I didn't mistaken the latest commit, it seems an issue remains, on this test.

````
# Datashader (holoviews+Bokeh)
import numpy as np
import pandas as pd
import holoviews as hv
import datashader as ds
from holoviews.operation.datashader import aggregate, shade, datashade, dynspread
from bokeh.models import DatetimeTickFormatter
hv.extension('bokeh')

def time_series(T = 1, N = 100, mu = 0.1, sigma = 0.1, S0 = 20):  
    """Parameterized noisy time series"""
    dt = float(T)/N
    t = np.linspace(0, T, N)
    W = np.random.standard_normal(size = N) 
    W = np.cumsum(W)*np.sqrt(dt) # standard brownian motion
    X = (mu-0.5*sigma**2)*t + sigma*W 
    S = S0*np.exp(X) # geometric brownian motion
    return S

def apply_formatter(plot, element):
    plot.handles['xaxis'].formatter = DatetimeTickFormatter()
    
drange = pd.date_range(start="2014-01-01", end="2016-01-01", freq='1D') # or '1min'
dates = drange.values.astype('int64')/10**6 # Convert dates to ints
curve = hv.Curve((dates, time_series(N=len(dates), sigma = 1)))
````

````
%%opts RGB [finalize_hooks=[apply_formatter] width=800]
%%opts Overlay [finalize_hooks=[apply_formatter] width=800] 
%%opts Scatter [tools=['hover', 'box_select']] (line_color="black" fill_color="red" size=10)

from holoviews.operation.timeseries import rolling, rolling_outlier_std
smoothed = rolling(curve, rolling_window=50)
outliers = rolling_outlier_std(curve, rolling_window=50, sigma=2)
datashade(curve, cmap=["blue"]) * dynspread(datashade(smoothed, cmap=["red"]),max_px=1) * outliers
````

````
C:\WinPython\basedir36\buildQt5\winpython-64bit-3.6.x.0\python-3.6.4.amd64\lib\site-packages\holoviews\element\raster.py in __init__(self, data, kdims, vdims, bounds, extents, xdensity, ydensity, **params)
    268                              "please use the QuadMesh element for "
    269                              "unevenly or irregularly sampled data." %
--> 270                              (type(self).__name__, self.get_dimension(0)))
    271         yvals = np.unique(np.diff(self.dimension_values(1, expanded=False)))
    272         if len(yvals) > 1 and np.abs(yvals.min()-yvals.max()) > sys.float_info.epsilon*100:

ValueError: RGB dimension x is not evenly sampled, please use the QuadMesh element for unevenly or irregularly sampled data.
````

File: holoviews/element/util.py [hunk 2]
```diff
@@ -282,3 +282,21 @@ def connect_edges(graph):
         end = end_ds.array(end_ds.kdims[:2])
         paths.append(np.array([start[0], end[0]]))
     return paths
+
+
+def validate_regular_sampling(img, dimension, rtol=10e-9):
+    """
+    Validates regular sampling of Image elements ensuring that
+    coordinates the difference in sampling steps is at most rtol times
+    the smallest sampling step. By default ensures the largest
```
jbednar: I can't parse this docstring; "ensuring that coordinates the difference"?

Also, I think rtol isn't useful as a function argument if this utility is primarily called from hv's own code that doesn't supply that argument.  It could be a module-level variable or this could be a ParameterizedFunction, either of which would allow the tolerance to be overridden in cases where it isn't appropriate.

PR #2245, opened 2018-01-05
QuadMesh and irregular xarray data fixes

philippjfr:
Three separate fixes for irregular gridded data and QuadMeshes. 

1) Ensures that xarray data is correctly canonicalized as previously constant dimensions were not appropriately squeezed out by the XArrayInterface.

2) Improved QuadMesh.trimesh method ensuring that key dimensions are inherited

3) Simplifies the matplotlib QuadMeshPlot

4) Add inference for xarray multi-dimensional (irregular) coordinate variables

These are all issues I ran into while implementing https://github.com/ioam/geoviews/pull/116.

- [x] Add unit test for canonicalizing xarray Dataset with constant dimension

### DIFF
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -198,7 +198,7 @@ class GridInterface(DictInterface):
 
 
     @classmethod
-    def canonicalize(cls, dataset, data, coord_dims=None):
+    def canonicalize(cls, dataset, data, coord_dims=None, irregular_dims=[]):
         """
         Canonicalize takes an array of values as input and
         reorients and transposes it to match the canonical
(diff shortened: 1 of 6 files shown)


philippjfr:
@jlstevens Ready to merge after review, tests should pass shortly.

holoviews/core/data/grid.py
```diff
@@ -198,7 +198,7 @@ def coords(cls, dataset, dim, ordered=False, expanded=False, edges=False):
 
 
     @classmethod
-    def canonicalize(cls, dataset, data, coord_dims=None):
+    def canonicalize(cls, dataset, data, coord_dims=None, irregular_dims=[]):
```
jlstevens: After discussing what ``irregular_dims`` really is, we agreed that we need a better name that makes it clearer that this is more of an xarray concept of dimensions than a holoviews one.

jlstevens: Perhaps to make it clear,  this could be ``irregular_xrdim`` where ``xr`` helps show this is the xarray concept of dimensions. When dealing with xarray 'coords', that name should then be used to make the distinction clear.

holoviews/element/raster.py
```diff
@@ -677,13 +678,17 @@ def trimesh(self):
...
         # Construct TriMesh
-        params = {k: v for k, v in util.get_param_values(self).items()
-                  if k != 'kdims'}
-        return TriMesh(((ts,), vertices), **params)
+        params = util.get_param_values(self)
+        params['kdims'] = params['kdims'] + TriMesh.node_type.kdims[2:]
+        node_params = {k: v for k, v in params.items() if k != 'vdims'}
+        nodes = TriMesh.node_type(vertices+(np.arange(len(vertices[0])),),
+                                  **node_params)
+        params = {k: v for k, v in params.items() if k != 'kdims'}
```
jlstevens: Very minor point: I would inline this dictionary comprehension to avoid overriding the ``params`` variable again with something derived from itself.

jlstevens: You could also do that to avoid defining ``node_params``...

holoviews/core/data/grid.py
jlstevens: This is complicated by the fact this method doesn't only apply to xarray...so we might need some other suggestion.

jlstevens:
Other than a naming situation that I would like to improve and one very minor suggestion, the PR looks good to merge.

philippjfr:
Ready to merge again.

jlstevens:
Looks good!

PR #2248, opened 2018-01-06
Ensure no information is lost when casting Image and Graph types

philippjfr:
Generally casting between different or the same type is straightforward, however for Image and Graph types copying the ``.data`` and parameters is not sufficient. This PR adds special handling ensuring that Image bounds and densities and Graph/TriMesh nodes and edgepaths are inherited during casting.

### DIFF
--- a/holoviews/element/graphs.py
+++ b/holoviews/element/graphs.py
@@ -127,8 +127,11 @@ class Graph(Dataset, Element2D):
         if isinstance(data, tuple):
             data = data + (None,)* (3-len(data))
             edges, nodes, edgepaths = data
+        elif isinstance(data, type(self)):
+            edges, nodes, edgepaths = data, data.nodes, data._edgepaths
         else:
             edges, nodes, edgepaths = data, None, None
+
         if nodes is not None:
             node_info = None
             if isinstance(nodes, self.node_type):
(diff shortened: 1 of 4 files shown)


philippjfr:
@jlstevens Ready to review/merge once tests pass.

holoviews/element/graphs.py
```diff
@@ -422,6 +425,8 @@ def __init__(self, data, kdims=None, vdims=None, **params):
         if isinstance(data, tuple):
             data = data + (None,)*(3-len(data))
             edges, nodes, edgepaths = data
+        elif isinstance(data, type(self)):
+            edges, nodes, edgepaths = data, data.nodes, data._edgepaths
```
jlstevens: This duplicates the code you would hope to be handled in ``Graph.__init__`` in the call to ``super``. This suggests some refactoring might be possible but as that might be tricky, it is just a wishlist task and not something to hold up the PR.

tests/testelementconstructors.py
```diff
@@ -179,3 +180,22 @@ def test_quadmesh_string_signature(self):
...
+    Tests whether casting an element will faithfully copy data and
+    parameters. Important to check for elements where data is not all
+    held on .data attribute, e.g. Image bounds or Graph nodes and
+    edgepaths.
+    """
+
+    def test_image_casting(self):
+        img = Image([], bounds=2)
+        self.assertEqual(img, Image(img))
+
```
jlstevens: Could we have a test of ``RGB`` as a subclass of ``Image``?

jlstevens:
Looks good. 

I made two comments but only one needs action before merging - the request for one additional unit test.

jlstevens:
Thanks for the addition test (which is now green). Merging.

PR #2249, opened 2018-01-07
Add scroll_to_selection bokeh Table style option

philippjfr:
Adds ``scroll_to_selection`` style option for bokeh TablePlot which is often useful when streaming data.


philippjfr:
Simply adds a style option, merging.

PR #2252, opened 2018-01-09
Ensure linked streams are resolved on HoloMap and Table objects

philippjfr:
There was a bug in the code that caused non-static and non-dynamic plots not to correctly resolve streams registered on an Element or HoloMap, this PR ensures that this is now resolved correctly. Additionally it adds support for resolving linked streams on bokeh's TablePlot which was previously not possible (because it was not implemented).

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -179,39 +179,6 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         self._shared = {'x': False, 'y': False}
 
 
-    def _construct_callbacks(self):
-        """
-        Initializes any callbacks for streams which have defined
-        the plotted object as a source.
-        """
-        if isinstance(self, OverlayPlot):
-            zorders = []
-        elif self.batched:
-            zorders = list(range(self.zorder, self.zorder+len(self.hmap.last)))
-        else:
-            zorders = [self.zorder]
-
-        if isinstance(self, OverlayPlot) and not self.batched:
-            sources = []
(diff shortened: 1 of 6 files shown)


philippjfr:
Ready to merge once tests pass. Added some unit tests for the HoloMap stream resolution.

jlstevens:
Thanks! I've really surprised the HoloMap * DynamicMap bug wasn't mentioned earlier!

Supporting streams on ``TablePlot`` is good too. I'll merge once the tests pass.

ayoungs:
Does this fix an issue where plotting DynamicMap in Jupiter notebooks work, but as soon as I run with bokeh serve --show       it will not detect the stream?  I've been working this for last few days trying to get that to work.  I'm hoping this fixes it!

ayoungs:
[InteractiveBoxTest.zip](https://github.com/ioam/holoviews/files/1618597/InteractiveBoxTest.zip)
Here's a test I've been trying to get to work.  I downloaded and installed the latest from master, this doesn't work in the App.
The jupyter notebook works as expected.
Then adding the two bottom lines:
doc = renderer.server_doc(layout)
doc.title = 'HoloViews App'
to a .py file with the same code to make an App, the App won't provide the BoundsX stream callback.
This sounds very similar to the issue you reference here in #2252

philippjfr:
>Does this fix an issue where plotting DynamicMap in Jupiter notebooks work, but as soon as I run with bokeh serve --show it will not detect the stream? I've been working this for last few days trying to get that to work. I'm hoping this fixes it!

I believe a separate PR fixed that, we can try to have a dev release so you can easily try it out.

PR #2257, opened 2018-01-12
Using Pylint to lint the code base against Python best practices or P…

jbampton:
…EP8.

https://www.pylint.org/

For consistency, always use """triple double quotes""" around docstrings.
Surround top-level function and class definitions with two blank lines.
Method definitions inside a class are surrounded by a single blank line.
Always surround these binary operators with a single space on either side: assignment (=), augmented assignment (+=, -= etc.), comparisons (==, <, >, !=, <>, <=, >=, in, not in, is, is not), Booleans (and, or, not).
Don't use spaces around the = sign when used to indicate a keyword argument or a default parameter value.

https://www.python.org/dev/peps/pep-0008/#blank-lines
https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements
https://www.python.org/dev/peps/pep-0257/

### DIFF
--- a/holoviews/__init__.py
+++ b/holoviews/__init__.py
@@ -58,6 +58,7 @@ for rcfile in [os.environ.get("HOLOVIEWSRC", ''),
                 print("Warning: Could not load %r [%r]" % (filename, str(e)))
         break
 
+
 def help(obj, visualization=True, ansi=True, backend=None,
          recursive=False, pattern=None):
     """
@@ -74,9 +75,9 @@ def help(obj, visualization=True, ansi=True, backend=None,
     info = Store.info(obj, ansi=ansi, backend=backend, visualization=visualization,
                       recursive=recursive, pattern=pattern, elements=elements_list)
 
-    msg = ( "\nTo view the visualization options applicable to this "
-            "object or class, use:\n\n"
-            "   holoviews.help(obj, visualization=True)\n\n")
+    msg = ("\nTo view the visualization options applicable to this "
+           "object or class, use:\n\n"
+           "   holoviews.help(obj, visualization=True)\n\n")
     if info:
         print((msg if visualization is False else '') + info)
     else:
--- a/holoviews/__init__.py
+++ b/holoviews/__init__.py
@@ -58,7 +58,6 @@ for rcfile in [os.environ.get("HOLOVIEWSRC", ''),
                 print("Warning: Could not load %r [%r]" % (filename, str(e)))
         break
 
-
 def help(obj, visualization=True, ansi=True, backend=None,
          recursive=False, pattern=None):
     """
--- a/holoviews/element/annotation.py
+++ b/holoviews/element/annotation.py
@@ -49,6 +49,7 @@ class Annotation(Element2D):
         ystart, ystop = ykey.start, ykey.stop
         return self.clone(self.data, extents=(xstart, ystart, xstop, ystop))
 
+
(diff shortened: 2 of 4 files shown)


philippjfr:
Thanks these changes mostly look good. We've already decided to enforce most of PEP8 using pyflakes on Travis but never got around to adding it.

>Surround top-level function and class definitions with two blank lines.
>Method definitions inside a class are surrounded by a single blank line.

We've inherited a different convention from a previous codebase and generally use two blank lines for methods and three for class definitions. That said this is not consistently enforced in any real way and there's probably a mixture in the codebase. I'm generally leaning toward keeping the non-PEP8 compliant convention here.

jbampton:
Hi @philippjfr ok I understand your views on blank lines.

Would you like me to remove the work done on the blank lines then by pushing another commit to this branch ?

philippjfr:
>Would you like me to remove the work done on the blank lines then by pushing another commit to this branch ?

Sounds perfect but you can also just push another commit on top of the last one, we generally squash merge so it makes little difference. Thanks again!

philippjfr:
I'll let @jlstevens review, but I'm happy to merge.

jlstevens:
Thanks for reverting those blank line changes!

At a glance, most of the changes look good and I'll do a more detailed review shortly.

jlstevens:
Ok, I've looked through the changes and the only thing I am not convinced about are the triple quoted docstrings. I am not convinced that:

```python
def foo():
   """A short docstring"""
   return 'foo'
```

Is really more appropriate or more readable than:

```python
def foo():
   "A short docstring"
   return 'foo'
```

I do realize that you will still want triple quotes for longer docstrings but this is one case where I don't mind inconsistency in quote style: I value short functions/methods with short readable docstrings (i.e without the extra visual noise of the triple quotes) more than I value always making use of the same quote style.

This is just an expression of my personal preference and I realize it is a matter of opinion. What do you think @philippjfr? Jim can act as the tie-breaker if we disagree...

philippjfr:
I think consistency is good, so I'm +1 on those changes.

jlstevens:
Actually, I am inclined to agree with two of the quote changes, namely for the class docstrings of ``VLine`` and ``HLine`` where triple quotes do seem appropriate. That just leaves this one:

```diff
     @property
      def subscribers(self):
 -        " Property returning the subscriber list"
 +        """Property returning the subscriber list"""
```

I don't mind merging this PR now but be warned, I'm not in the habit of using triple quoted docstrings for this sort of short function/method and I am unlikely to be able to get out of that habit!

philippjfr:
>I don't mind merging this PR now but be warned, I'm not in the habit of using triple quoted docstrings for this sort of short function/method and I am unlikely to be able to get out of that habit!

We can debate whether we want to do some basic automated linting at some later point. For now I'd be happy adding the basic pyflakes checks we already agreed on. I think far more important than docstring style consistency will be updating/improving the actual content, I suspect a fairly large percentage of docstrings are outdated and/or refer to old class names like ``View``, ``ViewMap``, ``Layer`` and so on.

+1 on merging now.

jlstevens:
> I suspect a fairly large percentage of docstrings are outdated and/or refer to old class names

Yes, I agree that updating the docstrings is more important and I also agree we should add the basic pyflake checks for things that are almost certainly coding errors.

Anyway, PR seems good so I will merge it now.

PR #2259, opened 2018-01-14
Fixed categorical coloring of Contours in matplotlib

philippjfr:
As the title says, Contours/Polygons in matplotlib did not handle categorical values for the ``color_index``.

### DIFF
--- a/holoviews/core/data/dictionary.py
+++ b/holoviews/core/data/dictionary.py
@@ -213,7 +213,7 @@ class DictInterface(Interface):
         if np.isscalar(values):
             if not expanded:
                 return np.array([values])
-            values = np.full(len(dataset), values)
+            values = np.full(len(dataset), values, dtype=np.array(values).dtype)
         else:
             if not expanded:
                 return util.unique_array(values)
--- a/holoviews/plotting/mpl/path.py
+++ b/holoviews/plotting/mpl/path.py
@@ -85,9 +85,12 @@ class ContourPlot(PathPlot):
             return (paths,), style, {}
 
         if element.level is not None:
-            style['array'] = np.full(len(paths), element.level)
+            array = np.full(len(paths), element.level)
         else:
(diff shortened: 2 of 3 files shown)


jlstevens:
I can't tell if this PR is ready yet, but at a glance it looks good. Happy to merge once you tell me it is ready and the tests are green.

philippjfr:
Can't be ready since the tests aren't green right :-) Locally it seems to pass but Travis isn't happy yet. Will try to figure this out today.

jlstevens:
> Can't be ready since the tests aren't green right...

Could have just been unlucky with transient failures you know...

philippjfr:
>Could have just been unlucky with transient failures you know...

The unit test I added is failing.

holoviews/core/data/dictionary.py
```diff
@@ -213,7 +213,7 @@ def values(cls, dataset, dim, expanded=True, flat=True):
         if np.isscalar(values):
             if not expanded:
                 return np.array([values])
-            values = np.full(len(dataset), values)
+            values = np.full(len(dataset), values, dtype=np.array(values).dtype)
```
philippjfr: This is the most bizarre bug I've come across in a long time. Before I added this ``dtype`` approach it would error (but only on Travis) saying that:

>ValueError: could not convert string to float: 'B'

But internally ``np.full`` does **exactly** the same thing I've done here: 

```python
def full(shape, fill_value, dtype=None, order='C'):
    if dtype is None:
        dtype = array(fill_value).dtype
    a = empty(shape, dtype, order)
    multiarray.copyto(a, fill_value, casting='unsafe')
    return a
```

Absolutely baffling, but it seems to work now.

philippjfr:
Ready now.

jlstevens:
Weird 'fix'!

Anyway, looks good. Merging.

PR #2265, opened 2018-01-20
Mesh improvements

philippjfr:
Improvements for handling QuadMesh and TriMesh. Both @jbednar and I have been writing up documentation on dealing with meshes of different kinds and in doing so both of us hit some things that make it awkward to work with TriMeshes in particular at the moment. In particular in order to pass in a set of vertices with associated values you previously would have to construct a set of Nodes or Points yourself and pass those into the ``TriMesh`` constructor. With this change you can now pass in a pandas DataFrame of vertices and it will correctly infer the value dimension.

Secondly this PR refactors the regrid operation in such a way that a bunch of the code can be reused by regridding operations in other packages (specifically xESMF based regridding in GeoViews).

Finally this PR changes the way bokeh QuadMesh plots are created avoiding binary serialization. Sending a bunch of small array chunks seems to confuse bokeh and cause errors and actually adds more overhead than simply sending the data without binary transfer.

### DIFF
--- a/holoviews/plotting/bokeh/raster.py
+++ b/holoviews/plotting/bokeh/raster.py
@@ -166,7 +166,7 @@ class QuadMeshPlot(ColorbarPlot):
     show_legend = param.Boolean(default=False, doc="""
         Whether to show legend for the plot.""")
 
-    _plot_methods = dict(single='rect')
+    _plot_methods = dict(single='quad')
     style_opts = ['cmap', 'color'] + line_properties + fill_properties
 
     def get_data(self, element, ranges, style):
@@ -197,8 +197,8 @@ class QuadMeshPlot(ColorbarPlot):
             XS, YS = [], []
             for x, y, zval in zip(X, Y, zvals):
                 if np.isfinite(zval):
-                    XS.append(x)
-                    YS.append(y)
+                    XS.append(list(x[:-1]))
+                    YS.append(list(y[:-1]))
             data = {'xs': XS, 'ys': YS, z.name: zvals[np.isfinite(zvals)]}
         else:
             xc, yc = (element.interface.coords(element, x, edges=True),
--- a/tests/testgraphelement.py
+++ b/tests/testgraphelement.py
@@ -1,8 +1,11 @@
 """
 Unit tests of Graph Element.
 """
+from unittest import SkipTest
+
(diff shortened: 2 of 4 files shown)


philippjfr:
Ready for review.

jbednar:
Looks good to me, thanks!

holoviews/plotting/bokeh/raster.py
```diff
@@ -197,8 +197,8 @@ def get_data(self, element, ranges, style):
             XS, YS = [], []
             for x, y, zval in zip(X, Y, zvals):
                 if np.isfinite(zval):
-                    XS.append(x)
-                    YS.append(y)
+                    XS.append(list(x[:-1]))
+                    YS.append(list(y[:-1]))
```
jlstevens: I assume there is a good reason though it seems odd to me you are slicing off the last element...

tests/testgraphelement.py
```diff
@@ -130,13 +133,15 @@ def test_graph_redim_nodes(self):
 class TriMeshTests(ComparisonTestCase):
 
     def setUp(self):
-        self.nodes = [(0, 0, 0), (0.5, 1, 1), (1., 0, 2), (1.5, 1, 3)]
+        self.nodes = [(0, 0, 0), (0.5, 1, 1), (1., 0, 2), (1.5, 1, 4)]
```
jlstevens: The last value wasn't tested before?

jlstevens:
Made two minor comments but otherwise it looks good to me!

holoviews/plotting/bokeh/raster.py
philippjfr: When drawing a polygon it automatically connects the first and last point, so the last point for each quad geometry is superfluous.

jlstevens: Ok, that does make sense. Thanks!

tests/testgraphelement.py
philippjfr: It was, but it wasn't a very good test. I've adjusted the test so it properly tests how the Graph index dimension is handled.

jlstevens:
I'll merge once the tests are green.

philippjfr:
Seems to be transients, I've restarted them.

PR #2266, opened 2018-01-21
Updated Deploying_Bokeh_Apps user guide

philippjfr:
Fixes two bugs in the bokeh apps documentation, ``_figure_data`` is now private (for consistency) and bokeh's ``Server`` no longer requires/accepts an explicit ioloop. The API around ``_figure_data`` still needs to be improved but ``BokehRenderer.figure_data`` no longer exists so at least things will work now.


jlstevens:
Looks good to me!

philippjfr:
Ready to merge.

PR #2268, opened 2018-01-23
Added draw streams for new bokeh drawing tools

philippjfr:
Adds drawing streams which are linked to the new bokeh edit tools (introduced in https://github.com/bokeh/bokeh/pull/7415). The tools+stream allow drawing boxes, points and polygons/paths on a plot as well as editing the vertices of existing paths and polygons. Whenever the tool edits the bokeh data sources the results are synced back to Python letting you access the data that has been edited/drawn.

- [x] Add tests
- [x] Add reference examples for each stream
- [x] Add gallery examples


philippjfr:
This is now almost ready to merge, I'll just add some unit tests. One thing that's worth mentioning/discussing is the BoxEdit stream. Internally the BoxEditTool in bokeh uses ``Rect`` glyphs to represent the boxes but at the HoloViews level we do not have any element which is rendered using ``Rect`` glyphs. This means that I cannot simply attach the tool to an existing renderer but instead have to attach it to a Polygons element (which uses a Patches glyph), then create a Rect based renderer and appropriately convert the data from polygons -> boxes and then from boxes -> polygons. This is obviously less than ideal but currently there is no way around it. While for the time being I think it's a reasonable approach and will make most of the basics work, in the long term it should appropriately use the correct glyph.

In the longer term I would like to see HoloViews elements that wrap around simple geometry types like boxes, ellipses etc. and indeed we have something like that already in form of the ``BaseShape`` annotation elements. The main problem with these is a) they are really just wrappers around Path data and b) they are not vectorized. In the long term I would really like it if all annotation types were vectorized but the migration path to get there is not a straightforward one, because the constructors are set up to draw a single annotation. 

Therefore I would really like to make decisions on this issue, do we add vectorized equivalents of the existing annotations as we are already planning for ``Text`` and ``MultiText`` or do we wait until 2.0 (which might be a fair while off) to change the Annotation constructors to allow vectorized data? In either case some low level vectorized geometry elements would be easy to implement, allow users to build complex custom plots from these lower level primitives without having to construct Overlays with hundreds of layers for each element. Here are the list of geometries I'd consider exposing:

* Rect
* Oval
* Wedge
* Arc
* Annulus
* VBar/HBar

In bokeh these would be directly mappable to glyph types while in matplotlib they would become PatchCollections and LineCollections of the appropriate type.

jlstevens:
I will need to think about it more but my initial reaction is to say we don't need to wait for 2.0 if the new vectorized elements have new names that don't clash (your list of glyphs don't overlap with any element names). Then we could discourage/add deprecation warnings for the old annotations without having to wait for 2.0 for the new functionality.

Again, I haven't thought too deeply about this question just yet...

philippjfr:
>I will need to think about it more but my initial reaction is to say we don't need to wait for 2.0 if the new vectorized elements have new names that don't clash (your list of glyphs don't overlap with any element names)

That's true although they duplicate some of the same functionality. Personally I have no issue with this but I know @jbednar was not too happy about the duplication of Text and MultiText elements. Personally I think it's fine and we just need to come up with good, non-clashing names for these elements.

jlstevens:
I would have an issue with the functionality duplication if we don't issue deprecation warnings for the elements that we will eventually want to get rid of (while offering a way to let users turn those warnings off).

jbednar:
I think that we shouldn't have duplication between Text and Multitext, etc. in the long run, and I think our docs should only ever present the one best way we currently recommend.  But we can of course keep the old objects around for compatibility, at least until 2.0.  

The only annoying bit that I see is that I don't want to have to use the name MultiText forever into the future, as it is ugly and not intuitive.  The rest seem fine...

philippjfr:
>The only annoying bit that I see is that I don't want to have to use the name MultiText forever into the future, as it is ugly and not intuitive.

We could just pick a better name, e.g. ``Labels``?

jlstevens:
Labels, labels, labels everywhere! ;-p

It is so overloaded anyway, maybe it doesn't matter?

philippjfr:
Are you talking ``Dimension.label`` and ``Dimensioned.label``? Don't think that's likely to get confused with ``Labels`` as an element.

jlstevens:
And ``group``/``label`` too...

philippjfr:
That's what I was referring to with ``LabelledData.label``.

jbednar:
I guess Labels is intuitive, though indeed it's vastly overloaded already!

philippjfr:
I suppose the other thing to discuss is passing additional parameters to streams. It's something we discussed before but keeps coming up. In this case there are certain options to set, e.g. what value to insert on value dimension columns when you add a new point, poly or box. Those are not parameters which should be included in the callback but it seems odd not to parameterize them. Last time we discussed using precedences to include and exclude parameters from the Stream contents but @jlstevens had objections iirc.

philippjfr:
Apart from that this is ready for review, I'll write some tests in the meantime.

jbednar:
I still don't have objections to using precedence in that way. :-)

jlstevens:
> I still don't have objections to using precedence in that way

It isn't about precedence but about what parameters mean on Stream objects. Anyway, I think we should come back to this discussion some other time.

philippjfr:
>Anyway, I think we should come back to this discussion some other time.

So you're saying I should make them simple attributes for now?

jbednar:
If you don't want to use precedence to indicate which parameters are part of the public Stream interface, then presumably there would need to be some other way to indicate that, which would need to be added to Param.  I'm generally reluctant to add new concepts to Param without a really clear motivation, so I'd favor slightly abusing precedence for this purpose rather than add some new fuzzy concept.  Seems better than using regular attributes for what is very clearly a Parameter, in any case.

jlstevens:
Looks good. I know you want this merged ASAP so I'll merge when the tests are green.

My main comment is that we still need a few more pieces for the docs:

* Another more ``DynamicMap`` focused example of how these streams are used as currently most examples use ``source``.
* I think there should be a section of the user guide discussing these types of streams explicitly, not just examples.

These things can be put into a separate PR where I might make some minor comments about the code that I haven't here.

jlstevens:
The pr builds are being slow but the push builds have passed. Merging.

jlstevens:
The key point is that I never want to reason about whether a parameter on a stream is a 'stream parameter' or a 'parameter on a stream', precedence or not precedence. It is just cleaner and simpler semantics to say 'parameters of stream classes are passed to the callbacks' rather than mixing things up.

jlstevens:
I would suggest a ``settings`` sub-object which groups all the parameters that control the stream but are not part of the stream semantics. This lets these things be parameterized without mixing them with the stream parameters.

PR #2270, opened 2018-01-24
Set bgcolor default to None

jlstevens:
This is a tiny change which I think is an improvement:

1. This allows bokeh themes to change the background color instead of always setting it to white.
2. The default background color without theming is white anyway, so it should not affect anything that isn't themed.

I think the same logic applies to matplotlib but a quick rgrep suggests it is more complicated:

image

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -49,7 +49,7 @@ legend_dimensions = ['label_standoff', 'label_width', 'label_height', 'glyph_wid
 
 class ElementPlot(BokehPlot, GenericElementPlot):
 
-    bgcolor = param.Parameter(default='white', doc="""
+    bgcolor = param.Parameter(default=None, allow_None=True, doc="""
         Background color of the plot.""")
 
     border = param.Number(default=10, doc="""


holoviews/plotting/bokeh/element.py
```diff
@@ -49,7 +49,7 @@
 
 class ElementPlot(BokehPlot, GenericElementPlot):
 
-    bgcolor = param.Parameter(default='white', doc="""
+    bgcolor = param.Parameter(default=None, allow_None=True, doc="""
```
philippjfr: ``default=None`` implies ``allow_None`` so ``allow_None`` is superfluous here.

    As a special case, if allow_None=True (which is true by default if
    the parameter has a default of None when declared) then a value
    of None is also allowed.

philippjfr: Don't mind though tbh.

jlstevens: I don't really mind either way but I am in the habit of explicitly setting ``allow_None``. I'll get rid of it if you ask me again.

jlstevens:
I think I'll go ahead and rebase off master and update the test data because I think this PR is ready to merge as 1. the handling of bgcolor in matplotlib is more complicated 2. the issue solved here is to allow bokeh themes to work properly which is bokeh specific 3. merging and testing is the best way to see things are working (which they should be as the default is white anyway).

jlstevens:
@philippjfr PR tests are now passing. I think this PR is ready to merge.

PR #2271, opened 2018-01-25
Fix bug in bokeh server setup when using RangeXY stream

philippjfr:
Starting a bokeh server app would fail if a RangeXY stream was attached to a subplot triggering changes to the ranges before the plot was even fully set up. In a scenario like this it would trigger change events during the plot instantiation and before the Document was set, causing errors.

This PR ensures callback events aren't processed until the plot is fully set up and adds some initial bokeh server unit tests.

### DIFF
--- a/tests/plotting/bokeh/testserver.py
+++ b/tests/plotting/bokeh/testserver.py
@@ -1,0 +1,157 @@
+from unittest import SkipTest
+
+from holoviews.core.spaces import DynamicMap
+from holoviews.core.options import Store
+from holoviews.element import Curve, Polygons, Path, Image, HLine
+from holoviews.element.comparison import ComparisonTestCase
+from holoviews.streams import RangeXY, PlotReset
+
+try:
+    from bokeh.application.handlers import FunctionHandler
+    from bokeh.application import Application
+    from bokeh.client import pull_session
+    from bokeh.document import Document
+    from bokeh.io import curdoc
+    from bokeh.server.server import Server
+
+    from holoviews.plotting.bokeh.callbacks import (
+        Callback, RangeXYCallback, ResetCallback
+    )
+    from holoviews.plotting.bokeh.renderer import BokehRenderer
+    bokeh_renderer = BokehRenderer.instance(mode='server')
+except:
+    bokeh_renderer = None
+
+
+class TestBokehServerSetup(ComparisonTestCase):
+
+    def setUp(self):
+        self.previous_backend = Store.current_backend
+        if not bokeh_renderer:
+            raise SkipTest("Bokeh required to test plot instantiation")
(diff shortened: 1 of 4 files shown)


philippjfr:
Ready to merge.

tests/testplotinstantiation.py
```diff
@@ -356,7 +356,7 @@ def setUp(self):
     def tearDown(self):
         Store.current_backend = self.previous_backend
         Callback._comm_type = comms.JupyterCommJS
-        mpl_renderer.comms['default'] = self.default_comm
+        bokeh_renderer.comms['default'] = self.default_comm
         Callback._callbacks = {}
```
jlstevens: Bit surprised by this. I would have thought either you want to set it for both mpl and bokeh...or alternatively it was only ever meant for bokeh and was always wrong?

tests/plotting/bokeh/testserver.py
```diff
@@ -0,0 +1,157 @@
...
+        server.start()
+        url = "http://localhost:" + str(server.port) + "/"
+        pull_session(session_id='Test', url=url, io_loop=server.io_loop)
+        self.assertTrue(launched)
+
+    def test_launch_server_with_stream(self):
+        obj = Curve([])
+        stream = RangeXY(source=obj)
+
+        global launched
```
jlstevens: If you need a global, maybe it should be something like a dict so you can say that it was launched for a particular unit test? Or do you not see any need for this sort of test anywhere else?

jlstevens:
Looks good and I am happy to merge once my two questions are addressed.

tests/testplotinstantiation.py
philippjfr: It was wrong before. I just came across it and snuck the fix in here.

tests/plotting/bokeh/testserver.py
philippjfr: Well I specifically set it to False each time before the test, so I don't think that's needed.

tests/testplotinstantiation.py
jlstevens: I suppose the incorrect teardown didn't mess up the tests (which I suppose it could have) otherwise it would have been immediately spotted...

tests/plotting/bokeh/testserver.py
jlstevens: As an alternative to using a global, something like this should work:

```
launched = []
def modify_doc(..):
   ...
   launched.append(True)
...
self.assertEquals(len(launched)==1)
```

philippjfr: Happy to change it to that.

philippjfr:
Thanks for fixing it up.

PR #2273, opened 2018-01-25
Fixed bogus overlay warnings in bokeh OverlayPlot

philippjfr:
In #2253 I ensured that OverlayPlot does not rerender its subplots unless necessary. Since there is a check to ensure that all elements in the overlay were plotted correctly this causes bogus warnings to appear, this change ensures that it correctly detects that the subplots were indeed processed.


jlstevens:
Looks good to me.

philippjfr:
Ready to merge.

PR #2278, opened 2018-01-28
Use nanosecond resolution for datashader sampling

philippjfr:
Datashader operations sometimes have issues with regular sampling. This is because ``Image`` types (or rather SheetCoordinateSystems) only handle microsecond resolution, however the issue can be avoided by having the datashader operation to use nanosecond resolution internally and converting back to microseconds before constructing the Image.

Also ended up having to define a lower sampling tolerance again because SheetCoordinateSystem only handles microsecond resolution.

- [x] Fixes https://github.com/ioam/holoviews/issues/2276 and ioam/holoviews#2280

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -135,7 +135,7 @@ class ResamplingOperation(Operation):
 
         xtype = 'numeric'
         if isinstance(xstart, datetime_types) or isinstance(xend, datetime_types):
-            xstart, xend = dt_to_int(xstart), dt_to_int(xend)
+            xstart, xend = dt_to_int(xstart, 'ns'), dt_to_int(xend, 'ns')
             xtype = 'datetime'
         elif not np.isfinite(xstart) and not np.isfinite(xend):
             if element.get_dimension_type(x) in datetime_types:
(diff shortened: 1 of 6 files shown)


philippjfr:
Ready for review.

holoviews/operation/datashader.py
```diff
@@ -135,7 +135,7 @@ def _get_sampling(self, element, x, y):
 
         xtype = 'numeric'
         if isinstance(xstart, datetime_types) or isinstance(xend, datetime_types):
-            xstart, xend = dt_to_int(xstart), dt_to_int(xend)
+            xstart, xend = dt_to_int(xstart, 'ns'), dt_to_int(xend, 'ns')
```
jlstevens: Wondering if life might be easier/ less error-prone if instead of ``dt_to_int(x, 'ns')`` and ``dt_to_int(x)`` you had two separate utilities: ``dt_to_ns`` and ``dt_to_us``...

philippjfr: Hmm, not sure how that's either easier or less error-prone. The utility is general purpose and in almost all cases the default of 'us' is appropriate.

jbednar: A single utility seems appropriate, as physicists might someday want time resolutions other than ns and us...

jlstevens: It just seems to be a pain to specify 'ns' everywhere. Anyway, it was just a suggestion and I don't really know how to make things easier...

jlstevens:
Looks good overall though it does seem to be a lot of fiddly little changes where you might miss one somewhere. At least the additional unit tests should help make sure it keeps working correctly.

Happy to merge unless @jbednar spots something to address.

jbednar:
Looks good to me!

PR #2281, opened 2018-01-29
Handled transposed xarray shapes correctly

philippjfr:
The XArray interface shape method was not taking into account data that was transposed relative to the key dimensions of the element resulting in wrong densities and causing further issues with selecting on the resulting object.

- [x] Fixes https://github.com/ioam/holoviews/issues/2258

### DIFF
--- a/holoviews/core/data/xarray.py
+++ b/holoviews/core/data/xarray.py
@@ -38,6 +38,22 @@ class XArrayInterface(GridInterface):
 
 
     @classmethod
+    def shape(cls, dataset, gridded=False):
+        array = dataset.data[dataset.vdims[0].name]
+        if not any(cls.irregular(dataset, kd) for kd in dataset.kdims):
+            names = [kd.name for kd in dataset.kdims
+                     if kd.name in array.dims][::-1]
+            if not all(d in names for d in array.dims):
+                array = np.squeeze(array)
+            array = array.transpose(*names)
+        shape = array.shape
+        if gridded:
+            return shape
+        else:
+            return (np.product(shape), len(dataset.dimensions()))
+
+
+    @classmethod
     def init(cls, eltype, data, kdims, vdims):
         element_params = eltype.params()
         kdim_param = element_params['kdims']
--- a/holoviews/core/data/xarray.py
+++ b/holoviews/core/data/xarray.py
@@ -111,9 +111,10 @@ class XArrayInterface(GridInterface):
             if vdims is None:
                 vdims = list(data.data_vars.keys())
             if kdims is None:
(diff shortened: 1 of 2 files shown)


philippjfr:
Ready to review and merge.

holoviews/core/data/xarray.py
```diff
@@ -37,6 +37,22 @@ def dtype(cls, dataset, dim):
         return dataset.data[name].dtype
 
 
+    @classmethod
+    def shape(cls, dataset, gridded=False):
```
jlstevens: This is fine if you are sure that the ``GridInterface`` implementation needs to be overridden. If not, it might be more correct to fix things in a more general way there.

tests/testimageinterfaces.py
```diff
@@ -413,6 +413,22 @@ class ImageXArrayInterfaceTest(ImageGridInterfaceTest):
...
+
+    def test_select_on_transposed_dataarray(self):
+        import xarray as xr
+        x = np.linspace(-3, 7, 533)
+        y = np.linspace(-5, 8, 894)
+        z = np.exp(-1*(x**2 + y[:, np.newaxis]**2))
+        array = xr.DataArray(z, coords=[y, x], dims=['x', 'y'])
+        img = Image(array)[1:3]
+        self.assertEqual(img.data, Image(array.sel(x=slice(1, 3))).data)
+
```
jlstevens: Might be nice to have one more unit test that relies on the updated ``shape`` property implementation.

jlstevens:
Other than the two comments above, looks good to me.

holoviews/core/data/xarray.py
philippjfr: I was remembering things incorrectly when I mentioned this, the ``GridInterface`` enforces that the array dimensions must match the shape as defined as ``tuple(len(element.dimension_values(kd)) for kd in element.kdims)[::-1]`` so it does not allow transposing the dimensions without explicitly transposing  the data.

philippjfr:
Added those tests, ready to merge.

jlstevens:
Does the test data need updating or are they failing for some other reason? I'll merge when the tests are green...

philippjfr:
Should be ready to merge now, once tests pass.

jlstevens:
All the tests are failing so there is probably some real issue, not just transients...

philippjfr:
Should be handled now.

jlstevens:
I'll merge when the tests are green.

jlstevens:
After a number of restarts due to transients, the tests are now green. Merging.

PR #2283, opened 2018-01-30
Setting the theme to None now disables theming

jlstevens:
Simple PR that completes the theming support by allowing you to disable theming by setting ``theme`` on the renderer to ``None``.

I suppose I'll add mention of this in the user guide.

### DIFF
--- a/holoviews/plotting/bokeh/renderer.py
+++ b/holoviews/plotting/bokeh/renderer.py
@@ -119,10 +119,7 @@ class BokehRenderer(Renderer):
         combining the bokeh model with another plot.
         """
         doc = curdoc() if doc is None else doc
-        if self_or_cls.theme:
-            doc.theme = self_or_cls.theme
-        elif self_or_cls.theme is None:
-            doc.theme = None
+        doc.theme = self_or_cls.theme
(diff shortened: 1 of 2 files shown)


holoviews/plotting/bokeh/renderer.py
```diff
@@ -121,6 +121,8 @@ def get_plot(self_or_cls, obj, doc=None, renderer=None):
         doc = curdoc() if doc is None else doc
         if self_or_cls.theme:
             doc.theme = self_or_cls.theme
+        elif self_or_cls.theme is None:
+            doc.theme = None
```
philippjfr: Would it be easier just to do:

```
doc.theme = self_or_cls.theme
```

BokehRenderer.theme can only be an Theme instance or None.

holoviews/plotting/bokeh/renderer.py
```diff
@@ -228,6 +230,9 @@ def _figure_data(self, plot, fmt='html', doc=None, **kwargs):
             m._document = None
         if self.theme:
             doc.theme = self.theme
+        elif self.theme is None:
+            doc.theme = None
```
philippjfr: Same as above applies here.

philippjfr:
Looks good and seems to be passing.

PR #2286, opened 2018-01-31
Implemented dynamic split_overlays

philippjfr:
This PR ensures that ``DynamicMap.split_overlays`` does something that at least makes sense and matches what the plotting code is already doing internally. None of the logic here is new so I'm pretty confident about the change. 

- [x] Fixes https://github.com/ioam/holoviews/issues/2782
- [x] Fixes https://github.com/ioam/holoviews/issues/2284
- [x] Implements  https://github.com/ioam/holoviews/issues/2254

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -1229,8 +1229,8 @@ class DynamicMap(HoloMap):
         """
         if not len(self):
             raise ValueError('Cannot split DynamicMap before it has been initialized')
-        elif not issubclass(self.type, Overlay):
-            return None, self
+        elif not issubclass(self.type, CompositeOverlay):
+            return [None], self
 
         from ..util import Dynamic
         keys = list(self.last.keys())
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -1236,12 +1236,20 @@ class DynamicMap(HoloMap):
         keys = list(self.last.data.keys())
         dmaps = []
         for key in keys:
-            def cb(obj, overlay_key=key, **kwargs):
+            el = self.last.data[key]
+            def split_overlay_callback(obj, overlay_key=key, overlay_el=el, **kwargs):
+                spec = util.get_overlay_spec(obj, overlay_key, overlay_el)
+                items = list(obj.data.items())
                 specs = [(i, util.get_overlay_spec(obj, k, v))
-                         for i, (k, v) in enumerate(obj.data.items())]
-                match = util.closest_match(overlay_key, specs)
-                return obj.data[match]
-            dmap = Dynamic(self, streams=self.streams, operation=cb)
+                         for i, (k, v) in enumerate(items)]
+                match = util.closest_match(spec, specs)
+                if match is None:
+                    raise KeyError('{spec} spec not found in {otype}. The split_overlays method '
+                                   'only works consistently for a DynamicMap where the '
+                                   'layers of the {otype} do not change.'.format(
+                                       spec=spec, otype=type(obj).__name__))
+                return items[match][1]
+            dmap = Dynamic(self, streams=self.streams, operation=split_overlay_callback)
             dmap.data = OrderedDict([(list(self.data.keys())[-1], self.last.data[key])])
             dmaps.append(dmap)
         return keys, dmaps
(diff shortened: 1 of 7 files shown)


jlstevens:
All tests passed! Very happy to see this as I wasn't expecting to see this attempted before the next release.

My main suggestion is unit tests (I realize those are probably planned with this PR marked as WIP) and to merge ASAP to maximise general testing.

philippjfr:
Main reason I tackled this now is that it's really a prerequisite for https://github.com/ioam/holoviews/pull/2174. The logic here has been used for a long time but it's never been directly tested and I think it needs at least a few changes to be completely safe. Will need lots of tests before I'm happy with it.

philippjfr:
Given that DynamicMap.split_overlay is completely broken atm, and causing errors in certain cases I think we should merge this now. It will also fix: https://github.com/ioam/holoviews/issues/2782

jlstevens:
> ...  I think we should merge this now. It will also fix: #2782

Sounds good. At a glance it looks ok but I'll have another look once the tests are passing.

philippjfr:
Seems to require some more changes. Will add those on Sunday.

philippjfr:
Fixed the last few issues, should be ready to review and merge once tests pass.

holoviews/core/util.py
```diff
@@ -1776,3 +1776,35 @@ def numpy_scalar_to_python(scalar):
     elif np.issubclass_(scalar_type, np.int_):
         return int(scalar)
     return scalar
+
+
+def closest_match(match, specs, depth=0):
```
jlstevens: I'm going to assume ``closest_match`` has been moved but not changed.

holoviews/plotting/bokeh/element.py
```diff
@@ -1451,7 +1451,7 @@ def update_frame(self, key, ranges=None, element=None):
         elif element is not None:
             self.current_frame = element
             self.current_key = key
-        items = element.items() if element else []
```
jlstevens: Was this broken for Python 3? How come this wasn't raised by the tests?

holoviews/core/util.py
jlstevens: There doesn't seem to be any unit tests for ``closest_match`` right now. Would be good to have and it looks like the sort of utility that could be easily tested.

holoviews/core/spaces.py
```diff
@@ -1221,6 +1220,31 @@ def dynamic_relabel(obj, **dynkwargs):
...
 
+    def split_overlays(self):
+        """
+        Splits a DynamicMap into its components. Only well defined for
+        DynamicMap with consistent number and order of layers.
+        """
+        if not len(self):
+            raise ValueError('Cannot split DynamicMap before it has been initialized')
+        elif not issubclass(self.type, CompositeOverlay):
+            return [None], self
```
jlstevens: ``[None]`` is a little weird as a list of keys and I don't think you could get this from splitting a dynamic map.

holoviews/core/spaces.py
```diff
@@ -1221,6 +1220,31 @@ def dynamic_relabel(obj, **dynkwargs):
...
+        from ..util import Dynamic
+        keys = list(self.last.data.keys())
+        dmaps = []
+        for key in keys:
+            def cb(obj, overlay_key=key, **kwargs):
+                specs = [(i, util.get_overlay_spec(obj, k, v))
+                         for i, (k, v) in enumerate(obj.data.items())]
+                match = util.closest_match(overlay_key, specs)
+                return obj.data[match]
+            dmap = Dynamic(self, streams=self.streams, operation=cb)
```
jlstevens: As ``operation`` is a parameter, it will be possible to get a handle on this callback after the dynamic maps are returned. For this reason, it is worth giving this callback a better name e.g ``dynamic_split_overlays`` or similar.

holoviews/plotting/bokeh/element.py
philippjfr: No, not broken. I just changed it so that the tuples are guaranteed to be tuples ensuring that I don't have to call ``wrap_tuple`` all over the place.

holoviews/core/spaces.py
philippjfr: True, it just returns ``None`` in that case.

jlstevens:
Other than a few comments this looks good.

philippjfr:
Ready to merge.

jlstevens:

PR #2288, opened 2018-01-31
Small bokeh plot fixes

philippjfr:
Fixes two bugs in bokeh plots:

1. Introduced recently when improving TablePlot erroneously assuming the data is non-empty.
2. Batched plots are ending up with the wrong colors due to static source optimization. Disabled for the time being, correctness more important than speed.


philippjfr:

jlstevens:
Looks good to me. Merging.

PR #2290, opened 2018-01-31
Allow sharing of PolyEditTool across multiple streams

philippjfr:
The PolyEditTool can edit multiple renderers at once, so this exposes an option to share the tool across multiple PolyEdit streams. Also since we're not using parameters for Stream settings I've added docstrings to describe the options.

- [x] Add tests
- [x] Update PolyEdit reference notebook

### DIFF
--- a/examples/reference/streams/bokeh/PolyEdit.ipynb
+++ b/examples/reference/streams/bokeh/PolyEdit.ipynb
@@ -50,7 +50,7 @@
     "\n",
     "    After selecting one or more vertices press BACKSPACE while the mouse cursor is within the plot area.\n",
     "    \n",
-    "As a simple example we will draw a number of boxes by displaying them using a ``Polygons`` element and then link that element to a ``PolyEdit`` stream. You may also supply a ``vertex_style`` dictionary defining the visual attributes of the vertices once you double tapped a polygon:"
+    "As a simple example we will draw a number of boxes and ellipses by displaying them using a ``Polygons`` element and then link that element to two ``PolyEdit`` streams. Enabling the ``shared`` option allows editing multiple ``Polygons``/``Paths`` with the same tool.  You may also supply a ``vertex_style`` dictionary defining the visual attributes of the vertices once you double tapped a polygon:"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


jlstevens:
> Also since we're not using parameters for Stream settings I've added docstrings to describe the options.

I think I've recently made a suggestion for a parameterized ``settings`` subobject on ``Streams`` for such parameters. Forgot which PR/issue I mentioned that on though....

philippjfr:
>I think I've recently made a suggestion for a parameterized settings subobject on Streams for such parameters

Can't say I like the suggestion.

jbednar:
I don't either; the subobject would need to be ignored, all accesses would awkwardly be to the subobject, and none of the documentation for those parameters would be on the main object.  Seems to me that stream parameters should be the special cases, not configuration parameters.  

Could we maybe add an optional list to Stream that if None, assumes all Parameters are stream parameters, and if a list, is a list of the names of the stream parameters (all others being normal parameters)?

jlstevens:
Maybe a naming convention on a regular parameter would be acceptable to me. e.g all such non-stream parameters need to be called``setting_X`` for instance. Not sure yet, I haven't thought about it but I insist on *something* to separate these types of parameter other than just precedence.

I think I would be happy if these parameters would be called ``setting_X`` and we use tests to enforce negative ``precedence`` on all parameters with this prefix and non-negative precedence for all other parameter names.

jbednar:
Ugh!

jlstevens:
> Ugh!

It is horrible but mixing stream parameters and other parameters is even worse.

philippjfr:
I don't like any of these suggestions really. I know you think Streams shouldn't have options at all but I don't think that's feasible for a wide array of streams whether that's the ``Buffer`` stream, ``PlotSize`` streams, or the new plot tool data streams. I think we should move the discussion elsewhere, for now I think using attributes is okay and matches what existing Stream types do.

So I think this is ready to merge now.

jlstevens:
I think streams *should* have parameterized options as long as they are distinct from stream parameters at a glance with no risk of confusion between these two parameter types.

> ... for now I think using attributes is okay and matches what existing Stream types do.

Agreed. Merging.

jbednar:
Is there an issue open about user-settable options for streams already?  I can see https://github.com/ioam/holoviews/issues/1719 is slightly relevant, but don't see an issue directly about that that we could link here.

PR #2292, opened 2018-02-01
Make Raster a Dataset type

philippjfr:
This is my initial attempt at making Raster a Dataset type. So far all I've done is make it a Dataset subclass and delete all the custom method implementations and surprisingly not that much broke. Will have to see how much more is required to get this working.


jlstevens:
> ... and surprisingly not that much broke

I hope the good luck continues as I think ``Raster`` is now the last holdout wrt dataset types so the sooner it can be updated, the better.

PR #2293, opened 2018-02-01
Added option for dimension range padding

philippjfr:
Adds support for dimension range padding as suggested in https://github.com/ioam/holoviews/issues/1090. Was reasonably straightforward, for backward compatibility no padding is added by default. Logic for computing padding on Overlays not yet quite correct.

All changes should for now be backward compatible and tests should pass.


jbednar:
Thanks! Looks like annoying code to have to write. :-)  Can you show examples of how to turn it on, and e.g. how it will apply only to non-raster types?

jlstevens:
Very happy to see this PR and I think having ``padding=0`` makes sense for now - the tests should still pass for that value. I would like to echo what Jim said of showing some examples, ideally in the docs.

What I worry about is that it could potentially affect everything so I just hope we can get in enough testing before the next release. The important thing is that ``padding=0`` works as before but having the ability to set padding too is going to be really valuable!

If all goes well, we might want to set a non-zero padding in this release (at least for some non-raster plot types) and a config option for backwards compatibility.

philippjfr:
For now I'm focused on getting tests passing, then I'll look at the API again (I suspect xpadding and ypadding are needed/useful). Don't think I'm in favor of enabling it by default for this release, it's too large a change in behavior to turn on by default. I think it should wait for 2.0 or at a minimum we should test it out for one release cycle before committing to default padding.

jlstevens:
> ...  we should test it out for one release cycle before committing to default padding.

I think this option is the most reasonable then.

philippjfr:
I think we can probably get this into the 1.11 release. One thing I'm wondering is whether we should add ``xpadding`` and ``ypadding`` options or whether ``padding`` could optionally accept a tuple. Any thoughts @jbednar and @jlstevens?

philippjfr:
This is now working as expected (the few test failures are expected and inconsequential) so before I start adding tests and documentation the API needs to be sorted out. The ``padding`` option clashes with one existing usage of ``padding`` on the matplotlib BarPlot, but since that needs a full rewrite anyway to be made consistent with the bokeh implementation I don't think that's a major issue. We do have to decide whether to add ``xpadding`` and ``ypadding``, allow a tuple on ``padding`` though or something else though.

holoviews/plotting/plot.py
```diff
@@ -597,6 +610,9 @@ class GenericElementPlot(DimensionedPlot):
     logy = param.Boolean(default=False, doc="""
         Whether the y-axis of the plot will be a log axis.""")
 
+    padding = param.Number(default=0, doc="""
+        Amount of padding to apply to data ranges.""")
```
jbednar: Some more detail would be good here, at least explaining that it's fractional and whether it is applied only to auto-computed ranges, or also to user-provided ranges.  If the former, maybe "Amount of padding to apply to auto-computed data ranges, as a fraction of the autocomputed range.  E.g. a value of 1.2 will result in a range 20% larger than the auto-computed range."?

philippjfr:
>From what I can tell, the value is a fraction applied to the current range?

Yes, if you see the linked issue that's what we had discussed. Matplotlib, bokeh, ggplot all use a default value of 5% padding, which I would express as ``0.05`` rather than ``1.05``.

>If that's true, then it won't depend on the axis scales, so I favor the tuple option for setting the padding per axis when needed.

Makes sense.

>Also, I'm not sure the current defaults will make sense for extreme aspect ratios. If e.g. x_range==10*y_range, it seems like a setting of 1.2 will result in a much larger absolute padding length on x than on y, which I think will look very bad.

I don't understand this, if it's a fraction then a value of ``1.2`` or rather ``0.2`` will look visually the same whatever the aspect ratio of the scales is:

screen shot 2018-07-05 at 11 39 44 am

ea42gh:
This has been on my wishlist for quite a while!

re tuple option, how would you handle twin axes?

philippjfr:
>re tuple option, how would you handle twin axes?

We currently have no real support for twin axes, but it is worth keeping in mind I suppose. Neither the xpadding/ypadding or the padding approach has any serious advantage for that case. In the vast majority of cases one value for both axes would be sufficient as well. If you have any suggestions that would make separate values for twin axes easier I'd love to hear them but atm I see no strong reason to change the design on account of them.

jbednar:
> 0.2 will look visually the same whatever the aspect ratio of the scales is:

Sorry to be unclear -- I wasn't talking about the aspect ratio of the *scales*, but of the plot itself.  I.e. I should have said `width` and `height`, not `x_range` and `y_range`, and that we need an example where e.g. `height=width/10`.  Seems like we'd want the same padding on both axes in such a case, yet if I understand it correctly the current implementation will give 10x the padding in width than in height instead.

philippjfr:
I see, yes that is true but I don't really consider it a major issue. If you do that you'll be able to set different padding values using the tuple format. Also I would consider extreme aspect ratios (>5) pretty rare cases. Here is an example with an aspect ratio of 8:

screen shot 2018-07-05 at 2 21 18 pm

As far as I'm aware other libraries do not provide any special handling for the padding if the aspect isn't square so I think my vote is not to overcomplicate things and make the padding dependent on the aspect.

jbednar:
Well, the extreme case just makes it clear what's going on, which is that the padding won't really be appropriate for anything but square aspects. I don't think it's anything to do with overcomplicating things, I think it's that currently the padding is being done with respect to entirely the wrong quantities, which becomes evident only when looking at an appropriate test case like you just showed.  Sure, for square aspects it's ok, but don't you think that when people want padding, they want approximately the same padding around the whole plot, i.e. the same number of inches on the page or pixels on the screen?  Your two examples above just make that clear -- a padding that's about the fraction of the plot range really isn't what people really want when they ask for padding.  So why should we follow the incorrect example of some other library, if it only accidentally ever works?

philippjfr:
Basically you want the padding to occur in screen space, but recall once again that we do not have access to the actual plot dimensions ahead of time (in bokeh at least), which means that there is currently no real way to achieve that with accuracy.

jlstevens:
I agree with Jim's desire (do it in screen space) but I also agree with Philipp that there are technical hurdles why we can't currently do it that way (unfortunately!). We need something to address this problem which is why I support this PR even though I also would love to have screen space padding.

ea42gh:
Here is a use case I have:
streaming data, non-negative x and y values.
I typically keep the axes centered on the origin, and pad the upper range of the x and y axes.

Looking forward to trying this out!

jbednar:
Ok, now the issue is clearer, thanks.  Do we have *no* access to the screen space, or do we have a nominal width and height that may or may not be respected given space for the axes, labels, etc.  If we do have some nominal values, I would vote for using those values as being better than using a fraction of the data range; it seems like those values are *nearly* correct in general, while the fraction of data range is only correct in the special case of square aspects.

philippjfr:
Okay, I've now adjusted the padding to use approximate screen space by computing the aspect ratio of the plot and adjusting the padding values accordingly. I'll be adding unit tests and documentation now.

jbednar:
Woohoo!  Sometimes I just have to complain loudly enough, and magic happens. :-)  Thanks, @philippjfr. From the code it now looks like it will behave properly in a much wider range of circumstances, and I look forward to trying it out.

philippjfr:
Will just have to make it very clear that's what it's doing, neither bokeh nor matplotlib uses screenspace to compute the padding.

holoviews/plotting/plot.py
```diff
@@ -597,6 +613,9 @@ class GenericElementPlot(DimensionedPlot):
     logy = param.Boolean(default=False, doc="""
         Whether the y-axis of the plot will be a log axis.""")
 
+    padding = param.ClassSelector(default=0, class_=(int, float, tuple), doc="""
+        Amount of padding to apply to data ranges.""")
```
jbednar: This still needs a clearer docstring; it's not clear what units it's in, whether it only applies to auto-ranged data or also to explicit ranges, etc.  Maybe:

```
Amount of extra screen space to add as a buffer around the specified or auto-ranged extents.
Specified as a fractional addition to width and heights, e.g., a padding of 0.2 will make an x,y plot 20% larger than the widest axis (x or y).  If specified as a tuple, the int or float values in the tuple will be used for padding in each axis, in order (x,y or x,y,z).
```

(not sure precisely how to describe it, but something like that?).

philippjfr: Not even sure about the name ``padding`` yet, matplotlib uses the term ``margin`` while bokeh uses ``range_padding``.

> Amount of extra screen space to add as a buffer around the specified or auto-ranged extents.

This is not quite correct, no buffer is added to user specified values. Otherwise sounds fine.

jbednar: I don't think I like `margin`, as I expect that to be outside the plot, not in it.  Padding I don't have much of an expectation, and am more likely to look up what that means.  `range_padding` presumably isn't appropriate for calculations in screen space, so I prefer `padding`.  

My point was mainly that it's not clear whether it applies to specified values or not, which you've clarified.  (I could imagine an argument for applying it to user extents too, i.e. that you could automatically add a bit of space to user-computed values, but on balance I think it's probably least surprising not to.)  So it sounds like:

```
Amount of extra screen space to add as a buffer around auto-ranged extents.
Specified as a fractional addition to width and heights, e.g., a padding of 0.2 will make an x,y plot 20% larger than the widest axis (x or y).  If specified as a tuple, the int or float values in the tuple will be used for padding in each axis, in order (x,y or x,y,z).
```

If that's correct, then I think that's enough information.

philippjfr: >Specified as a fractional addition to width and heights

Don't think this is quite the correct way of phrasing this, it sounds like it's actually adding to the width and height. Maybe something like

>Amount of extra screen space to add as a buffer around auto-ranged extents. Specified as a fraction of the width and height of an axis, ensuring equal padding for non-square plots. For example a padding of 0.2 will ensure that 20% of each axis range is devoted to padding. If specified as a tuple, the int or float values in the tuple will be used for padding in each axis, in order (x,y or x,y,z).

philippjfr:
While I'm already dealing with this logic I'd also like to discuss the addition of ``xlim``, ``ylim`` (and ``zlim``) plot options. While the padding will address a lot of user issues around setting ranges, users asking how to set ranges is probably the most frequent question we get. Setting the dimension range using redim has improved things a bit in that regard, but it is still a significant conceptual leap and lots of users get confused about which dimension is which. It's also not a very programmatically friendly because it requires knowledge about the name of dimension, which in some cases cannot be programmatically inspected, e.g. on DynamicMaps.

I've hesitated making this addition for a long time, hoping to deprecate ``Element.extents`` fully first. However doing that will be a significant chunk of work and I seriously think this is turning some users off HoloViews.

ea42gh:
To me, `Amount of extra screen space` sounds weird,
Extend axis range by a given percentage of the axis length?

holoviews/plotting/plot.py
jbednar: I don't quite get it.  How will a value of 0.2 ensure that 20% of each axis range is devoted to padding, if we want the same padding all around, even for non-square axes?  Doesn't it need to be 20% of the axis range of the axis whose width or height is larger?

philippjfr: >Doesn't it need to be 20% of the axis range of the axis whose width or height is larger?

You're right that's still correct, my objection was specifically to:

>Specified as a fractional addition to width and heights

But we do need to keep the bit about the larger axis.

jbednar: Ah, ok.  It's tough for me to write good docs when I'm not sure of the precise details of the algorithm, but that's why we need good docs!  I can't see how to say this concisely, so how about:

```
Fraction by which to increase auto-ranged extents to make datapoints more visible
around borders.

To compute padding, the axis whose screen size is largest is chosen, and the range
of that axis is increased by the specified fraction.  To ensure equal padding for all
axes, other axes are then padded so that their screen size will increase by roughly
the same amount.  If specified as a tuple, the int or float values in the tuple will be
used for padding in each axis, in order (x,y or x,y,z).

For example, for `padding=0.2` on a 800x800-pixel plot, an x axis with the range 
[0,10] will be padded by 20% to be [-1,11], while a y axis with a range [0,1000] will
be padded to be [-100,1100], which should make the padding be approximately the
same number of pixels.  But if the same plot is changed to have a height of only
200, the y range will then be [-400,1400] so that the y-axis padding will match that
of the x axis.
```

philippjfr: >To ensure equal padding for all axes, other axes are then padded so that their screen size will increase by roughly the same amount.

This again makes it sound like the screen space is increased, when it simply devotes a certain percentage of the existing screen space to the padding. Your description walking through an example of what it's actually doing is correct though. How about changing that to

>Other axes are then padded ensuring that the amount of screen space devoted to padding is equal for all axes.

philippjfr: >Fraction by which to increase auto-ranged extents to make datapoints more visible
around borders.

>To compute padding, the axis whose screen size is largest is chosen, and the range
of that axis is increased by the specified fraction.  Other axes are then padded ensuring that the amount of screen space devoted to padding is equal for all axes.  If specified as a tuple, the int or float values in the tuple will be used for padding in each axis, in order (x,y or x,y,z).

>For example, for `padding=0.2` on a 800x800-pixel plot, an x axis with the range 
[0,10] will be padded by 20% to be [-1,11], while a y axis with a range [0,1000] will
be padded to be [-100,1100], which should make the padding be approximately the
same number of pixels.  But if the same plot is changed to have a height of only
200, the y range will then be [-400,1400] so that the y-axis padding will match that
of the x axis.

jbednar: Good catch; yes, that's what I meant to convey, now much more clearly. I'm happy with both the description and the algorithm now, and look forward to seeing it in action!

philippjfr:
This needs further unit tests to ensure all plot types are behaving correctly but the core logic is now complete and the tests are passing. Suggestions on where best to document this would be helpful. Also worth noting that out of the current 1500 lines about 1000 are unit tests and I expect that to go to at least 2000 lines.

philippjfr:
I'm also now wondering whether we should allow different padding values for left and right on the x-axis and the bottom and top on the y-axis. As @ea42gh suggested sometimes it makes to pad only the top and right of a plot. The fact that tuples are used to denote x- and y-axis complicates things and it would probably be confusing to use a list, e.g. ``[lower_pad, upper_pad]`` and ``([left_pad, right_pad], [bottom_pad, top_pad])``, but I can't really think of alternatives.

jbednar:
The behavior for one-sided padding sounds good apart from seemingly using twice the padding along the top for a given padding fraction.  Unless I counted the tick marks incorrectly in the above plots (which is very possible!), it seems like a fraction of 0.2 will result in all 20% of the padding being on the top in the one-sided case, while it gives 10% padding on the top and 10% on the bottom in the two-sided case.  

To make this all semantically clean it seems like the padding value should be redefined consistently as the one-sided padding value in all cases, such that a padding of 0.1 gives 10% padding on the top and 10% padding on the bottom in the two-sided case, and 10% padding on the top and 0% padding on the bottom in the one-sided case.  I think that's the behavior that people would want and expect, so that they don't have to re-define the numerical value depending on whether it is one or two sided.  

In any case, the docstring needs updating to explain the one-sided case.

philippjfr:
>The behavior for one-sided padding sounds good apart from seemingly using twice the padding along the top for a given padding fraction. Unless I counted the tick marks incorrectly in the above plots (which is very possible!), it seems like a fraction of 0.2 will result in all 20% of the padding being on the top in the one-sided case, while it gives 10% padding on the top and 10% on the bottom in the two-sided case.

You must have counted incorrectly, it's adding 10% on top.

>To make this all semantically clean it seems like the padding value should be redefined consistently as the one-sided padding value in all cases

I agree with that and will make the necessary changes.

jbednar:
As for the manual settings, seems like we could force it to accept only a scaar or tuples in all cases, if we are willing for some cases to be more verbose. Here I would assume that manual one-sided padding is less common than manual per-axis padding, so that it's ok for that case to be ugly to specify:

1. pad : two-sided padding on all axes
2. (x_pad, y_pad, z_pad) : two-sided padding separately on each axis
3. ((x_low_pad, x_high_pad), (y_low_pad, y_high_pad), (z_low_pad, z_high_pad)) : arbitrary-sized padding separately on each side of each axis
4. ((low_pad, high_pad), (low_pad, high_pad), (low_pad, high_pad)) : equal one-sided padding on all axes

Case 1 is short and presumably the most common, case 2 is reasonably concise, and case 3 is fully general.  Case 4 is awkward and duplicative, but this seems like a very rare case anyway?  Because the distinction between tuples and lists is very subtle, and because JSON doesn't support tuples directly, I'd prefer to avoid depending on a semantic distinction between tuple and list notation except where it's truly unavoidable.

philippjfr:
@jlstevens @jbednar Ready for a first round of reviews, it now behaves as desired with all the options outlined in @jbednar's comment above being supported. Before this can be merged I'll have to make a PR against geoviews for compatibility.

philippjfr:
Oh and I need to add some documentation about this, any idea where best to put it?

jbednar:
Whew!  That looks like a lot of work.  But I think it's totally worth it; it will be great to have this!

My guess (but it's only a guess) is that a very short mention of the docs (including basic x/y/zlim examples) should be in getting-started/Customization, which would point to a longer section in user-guide/Customization.

holoviews/core/data/__init__.py
```diff
@@ -274,22 +274,25 @@ def sort(self, by=[], reverse=False):
         return self.clone(sorted_columns)
 
 
-    def range(self, dim, data_range=True):
+    def range(self, dim, data_range=True, dimension_range=True):
```
jlstevens: Would be good to update the docstring to explain this new flag.

holoviews/core/util.py
```diff
@@ -809,36 +809,80 @@ def find_range(values, soft_range=[]):
             return (None, None)
 
 
-def max_range(ranges):
+def is_finite(value):
+    """
+    Safe check whether a value is finite, only None and NaN values are
+    considered non-finite and allows checking all types not restricted
+    to numeric types.
```
jlstevens: And infinity isn't also non-finite? ;-p

holoviews/core/util.py
```diff
@@ -809,36 +809,80 @@ def find_range(values, soft_range=[]):
...
+            else:
+                return (np.nanmin(arr[:, 0]), np.nanmax(arr[:, 1]))
     except:
         return (np.NaN, np.NaN)
 
 
-def dimension_range(lower, upper, dimension):
+def range_pad(lower, upper, padding=None, log=False):
+    """
+    Pads the range by a fraction of the interval
```
jlstevens: I didn't spot any unit tests explicitly for these utilities. I know they should be tested indirectly by the more comprehensive unit tests in this PR but a couple of utility unit tests might be useful as examples of the expected input/output...

holoviews/plotting/bokeh/annotation.py
```diff
@@ -61,7 +61,7 @@ def get_batched_data(self, element, ranges=None):
                 data[k].extend(eld)
         return data, elmapping, style
 
-    def get_extents(self, element, ranges=None):
+    def get_extents(self, element, ranges=None, range_type='combined'):
```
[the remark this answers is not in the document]
philippjfr: >I'm not sure where this should go, but one of the get_extents methods should mention that range_type can be 'data' or 'combined' (are there others?). I found out those are the two expected values by searching the code...

Presumably the baseclass should document this in detail, the full list include:

* 'data': Just the data ranges
* 'combined': All the range types combined correctly
* 'extents': Element.extents
* 'soft': Dimension.soft_range values
* 'hard': Dimension.range values

holoviews/plotting/util.py
```diff
@@ -309,6 +309,45 @@ def compute_sizes(sizes, size_fn, scaling_factor, scaling_method, base_size):
...
+    """
+    Process a padding value supplied as a tuple or number and returns
+    padding values for x-, y- and z-axis.
+    """
+    if isinstance(padding, tuple):
+        if len(padding) == 2:
+            xpad, ypad = padding
+            zpad = 0
+        elif len(padding) == 3:
+            xpad, ypad, zpad = (padding, padding, padding)
```
jlstevens: This is confusing to me...if the input is a 3-tuple, the return is that 3-tuple repeated 3 times in another 3-tuple?

holoviews/core/data/__init__.py
philippjfr: Agreed, should be done here and in the Dimensioned baseclass at least.

holoviews/core/util.py
philippjfr: Good point.

holoviews/plotting/plot.py
```diff
@@ -582,6 +595,9 @@ class GenericElementPlot(DimensionedPlot):
     bgcolor = param.ClassSelector(class_=(str, tuple), default=None, doc="""
         If set bgcolor overrides the background color of the axis.""")
 
+    default_span = param.ClassSelector(default=2.0, class_=(int, float, tuple), doc="""
+        Default span if the spans zero range.""")
```
jlstevens: I'm not quite sure that a `span` is supposed to be...

holoviews/core/util.py
philippjfr: Agreed, should be straightforward.

holoviews/plotting/plot.py
```diff
@@ -597,6 +613,32 @@ class GenericElementPlot(DimensionedPlot):
     logy = param.Boolean(default=False, doc="""
         Whether the y-axis of the plot will be a log axis.""")
 
+    padding = param.ClassSelector(default=0, class_=(int, float, tuple), doc="""
```
jlstevens: Personally, I would just specify float and not make special allowances for int...

holoviews/plotting/plot.py
```diff
@@ -597,6 +613,32 @@ class GenericElementPlot(DimensionedPlot):
...
+        pixels. But if the same plot is changed to have a height of only
+        200, the y-range will then be [-400,1400] so that the y-axis
+        padding will still match that of the x-axis.
+
+        It is also possible to declare non-equal padding value for the
+        lower and upper bound of an axis by supplying nested tuples,
+        e.g. padding=(0.1, (0, 0.1)) will pad the x-axis lower and
+        upper bound as well as the y-axis upper bound by a fraction of
+        0.1 while the y-axis lower bound is not padded at all.""")
+
```
jlstevens: Not sure about this format but it is good to allow extra flexibility when required.

holoviews/plotting/util.py
philippjfr: Hmm, yeah offhand this does look incorrect. Reminds me that I haven't actually written a single test for 3d plots yet, which probably would have uncovered an issue here.

holoviews/plotting/bokeh/annotation.py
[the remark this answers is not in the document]
jlstevens: And there is 'soft' and 'hard'?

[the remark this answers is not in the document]
jlstevens: Again..github doesn't show your comments when you post them!

holoviews/plotting/plot.py
```diff
@@ -1096,9 +1225,63 @@ def get_extents(self, overlay, ranges):
...
+            hx0, hy0, hx1, hy1 = hard_extent
+            z0, z1 = np.NaN, np.NaN
+
+        xspan, yspan, zspan = (v/2. for v in get_axis_padding(self.default_span))
+        if util.is_number(x0) and x0 == x1: x0, x1 = x0-xspan, x1+xspan
+        if util.is_number(x0) and y0 == y1: y0, y1 = y0-yspan, y1+yspan
+        if util.is_number(z0) and z0 == z1: z0, z1 = z0-zspan, z1+zspan
+        xpad, ypad, zpad = self.get_padding((x0, y0, z0, x1, y1, z1))
+
+        print('<<<', hard_extents)
```
jlstevens: Stray print.

holoviews/plotting/plot.py
```diff
@@ -1071,14 +1199,15 @@ def _update_subplot(self, subplot, spec):
             subplot.overlay_dims = util.OrderedDict(new_dims)
 
 
-    def get_extents(self, overlay, ranges):
-        extents = []
+    def get_extents(self, overlay, ranges, range_type='combined'):
```
jlstevens: This method is a bit of a monster...

holoviews/plotting/plot.py
philippjfr: Doesn't help that the current docstring isn't english...

Basically it defines what happens if the data has no range, e.g. when you have a single point like ``hv.Points([(0, 0)])`` it means the axes will span from -1 to 1 (i.e. a span of 2).

holoviews/plotting/plot.py
philippjfr: Thanks.

holoviews/plotting/plot.py
philippjfr: So you would want a user specifying ``0`` to raise an error?

holoviews/plotting/plot.py
philippjfr: You should see it before I refactored it :-) I'll have another go at factoring stuff out though.

jlstevens:
Looks good to me though this is a pretty huge PR. For that reason, I would actually like to merge it soon to give us a chance to uncover bugs.

I've made a few comments and once you've added a few examples of the padding options to the docs, I'll check out this branch and test it for myself.

holoviews/plotting/plot.py
philippjfr: Since you haven't replied I'm going to leave this, I don't think ``0`` should be an error.

jbednar: I don't think 0 should be an error either.

holoviews/plotting/plot.py
jlstevens: I see. I would like to have a better name for this concept but I don't have one to suggest at the moment. Conceptually it is something like `fallback_span` but that name is horrible...

holoviews/plotting/plot.py
```diff
@@ -584,6 +597,12 @@ class GenericElementPlot(DimensionedPlot):
     bgcolor = param.ClassSelector(class_=(str, tuple), default=None, doc="""
         If set bgcolor overrides the background color of the axis.""")
 
+    default_span = param.ClassSelector(default=2.0, class_=(int, float, tuple), doc="""
+        Defines the span of an axis if the axis range is zero, i.e. if
+        the lower and upper end of an axis are equal. For example if
+        there is a single datapoint at 0 a default_span of 2.0 will
+        result in axis ranges spanning from -1 to 1.""")
```
jbednar: The equivalent of `default_span` in datashader is used both when min==max and also when there is no data (e.g. if someone filtered the data dynamically and no datapoints were in the filter range).  Is that true here?  If so, presumably it should be mentioned in the docs.

holoviews/plotting/plot.py
ea42gh: a fallback_range maybe?  Is this triggered only if the axis extent is [a,a], i.e., if the span is 0?
I could think of a tolerance, e.g. when plotting residuals (with bounds [-1e-12, 1e-8] for example).

Looking at this, I used the terms range, extent, bounds....

philippjfr: I used ``default_span`` because that's what bokeh uses, ``fallback_span`` is semantically okay but I also don't like the name. ``fallback_range`` is incorrect because it's not a range or extent since it specifies the span or interval not explicit upper and lower bounds.

holoviews/plotting/plot.py
philippjfr: Sure, I can add that once we've decided on a name for the parameter.

philippjfr: Both the name and the default value of 2 were copied from bokeh. @jlstevens indicated he didn't like the name in a comment above but we both dislike his only suggestion even more:

>I see. I would like to have a better name for this concept but I don't have one to suggest at the moment. Conceptually it is something like fallback_span but that name is horrible... 

My vote is therefore to stick with ``default_span``.

jbednar: Same here.  Ok, I'll change datashader to 2.0.

philippjfr:
Once tests pass this will be ready to merge. I've addressed the majority of the comments. It would be good to have an immediate dev release after this, so I can merge the corresponding changes into geoviews.

holoviews/plotting/mpl/raster.py
```diff
@@ -150,6 +150,9 @@ class RasterGridPlot(GridPlot, OverlayPlot):
     equivalent using subplots.
     """
 
+    padding = param.Number(default=0.1, doc="""
+        The amount of padding as a fraction of the total Grid size""")
```
jlstevens: This `padding` parameter is a bit different from the common one...

jlstevens:
Looks good and I'm happy to merge (though I've made one new comment for you to think about first)...

holoviews/plotting/mpl/raster.py
philippjfr: Yes, thanks for catching that. It might cause confusion but I don't think it's a huge issue, it's on ``GridPlot`` and therefore won't actually clash, and also only exists in the matplotlib backend.

philippjfr:
I think we can merge, while it might cause confusion the ``padding`` parameter on GridSpace does not actually clash with that on ElementPlot and also only applies to the matplotlib backend. I think I'll open a quick issue to rename the GridSpace parameter.

philippjfr:
Actually I don't think we need to or should rename, ``padding`` on a GridPlot is a very similar concept.

jlstevens:
Ok, sounds good! Merging.

PR #2295, opened 2018-02-02
Fixed bug computing categorical datashader aggregates

philippjfr:
Recently I added some validation to xarray Datasets that disallows coords (kdims) and data variables (vdims) with the same name, which was apparently being used by the datashader ``aggregate`` operation for categorical aggregates. This handles them correctly.

- [x] Add tests for categorical aggregates

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -752,7 +752,20 @@ class shade(Operation):
         return "#{0:02x}{1:02x}{2:02x}".format(*(int(v*255) for v in rgb))
 
 
+    @classmethod
+    def to_xarray(cls, element):
+        if issubclass(element.interface, XArrayInterface):
+            return element
+        data = tuple(element.dimension_values(kd, expanded=False)
+                     for kd in element.kdims)
+        data += tuple(element.dimension_values(vd, flat=False)
+                      for vd in element.vdims)
+        dtypes = [dt for dt in element.datatype if dt != 'xarray']
+        return element.clone(data, datatype=['xarray']+dtypes)
+
+
     def _process(self, element, key=None):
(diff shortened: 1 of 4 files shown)


philippjfr:
Now added tests, ready to merge once passing.

philippjfr:
I've also pushed a fix for https://github.com/ioam/holoviews/issues/2299 to this PR and added a unit test.

holoviews/operation/datashader.py
```diff
@@ -374,9 +373,10 @@ def _process(self, element, key=None):
                 raise ValueError("Aggregation column %s not found on %s element. "
                                  "Ensure the aggregator references an existing "
                                  "dimension." % (column,element))
+            name = column
             if isinstance(agg_fn, ds.count_cat):
-                name = '%s Count' % agg_fn.column
-            vdims = [dims[0](column)]
+                name = '%s Count' % column
```
jlstevens: ``name = '%s Count' % column if isinstance(agg_fn, ds.count_cat) else column``?

Personally, I don't like it when I see simple variable reassignments e.g ``x=y`` used in this way. Up to you, I won't hold the merge up because of this.

jlstevens:
Happy to merge once you reply to my comment. I won't hold up the merge if you don't want to make the suggested change.

holoviews/operation/datashader.py
philippjfr: There is more to the if/else block than just this single line.

philippjfr: Suppose I see what you're getting at.

philippjfr:
Made the change, ready to merge once tests pass again.

jlstevens:
Tests are passing. Merging.

PR #2300, opened 2018-02-03
Fixed bug computing cartesian product of keys

philippjfr:
The ``cartesian_product`` implementation it was using is not safe for all types since it uses NumPy so I replaced it with itertools.product. Fixes a bug when using booleans as keys.


philippjfr:
Ready to merge

jlstevens:
Looks good though I see you haven't removed ``cartesian_product``? Is it safe wherever it is still being used? 

Otherwise, happy to merge.

philippjfr:
>Looks good though I see you haven't removed cartesian_product? Is it safe wherever it is still being used?

Yes absolutely, it's meant for NumPy arrays, just not appropriate to use in this case.

jlstevens:
Ok, thanks for the clarification. Merging.

PR #2305, opened 2018-02-05
Added support for dask arrays in GridInterface

philippjfr:
As requested in https://github.com/ioam/holoviews/issues/2176 this PR adds support for using dask arrays with the gridded interface. This turned out to be completely trivial (as you can see most of the new code is tests)  and let users use dask arrays without relying on the heavier xarray dependency. @jakirkham Would you mind testing this and seeing whether it addresses your requirements?

### DIFF
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -5,7 +5,16 @@ try:
 except ImportError:
     pass
 
+
 import numpy as np
+array_types = (np.ndarray,)
+
+try:
+    import dask.array as da
+    array_types += (da.Array,)
+except ImportError:
+    da = None
+
 
 from .dictionary import DictInterface
 from .interface import Interface, DataError
@@ -78,7 +87,7 @@ class GridInterface(DictInterface):
             name = dim.name if isinstance(dim, Dimension) else dim
             if name not in data:
                 raise ValueError("Values for dimension %s not found" % dim)
-            if not isinstance(data[name], np.ndarray):
+            if not isinstance(data[name], array_types):
                 data[name] = np.array(data[name])
 
         kdim_names = [d.name if isinstance(d, Dimension) else d for d in kdims]
@@ -226,18 +235,18 @@ class GridInterface(DictInterface):
                 invert = True
             else:
                 slices.append(slice(None))
-        data = data[slices] if invert else data
+        data = data[tuple(slices)] if invert else data
 
         # Transpose data
         dims = [name for name in data_coords
-                if isinstance(cls.coords(dataset, name), np.ndarray)]
+                if isinstance(cls.coords(dataset, name), array_types)]
         dropped = [dims.index(d) for d in dims
                    if d not in dataset.kdims+virtual_coords]
         if dropped:
             data = data.squeeze(axis=tuple(dropped))
 
         if not any(cls.irregular(dataset, d) for d in dataset.kdims):
-            inds = [dims.index(kd.name)for kd in dataset.kdims]
+            inds = [dims.index(kd.name) for kd in dataset.kdims]
             inds = [i - sum([1 for d in dropped if i>=d]) for i in inds]
             if inds:
                 data = data.transpose(inds[::-1])
@@ -301,6 +310,8 @@ class GridInterface(DictInterface):
         if dim in dataset.vdims or dataset.data[dim.name].ndim > 1:
             data = dataset.data[dim.name]
             data = cls.canonicalize(dataset, data)
+            if da and isinstance(data, da.Array):
+                data = data.compute()
             return data.T.flatten() if flat else data
         elif expanded:
             data = cls.coords(dataset, dim.name, expanded=True)
@@ -364,7 +375,7 @@ class GridInterface(DictInterface):
     def key_select_mask(cls, dataset, values, ind):
         if isinstance(ind, tuple):
             ind = slice(*ind)
-        if isinstance(ind, np.ndarray):
+        if isinstance(ind, array_types):
             mask = ind
         elif isinstance(ind, slice):
             mask = True
@@ -491,7 +502,10 @@ class GridInterface(DictInterface):
                 data[d].append(arr)
             for vdim, array in zip(dataset.vdims, arrays):
                 flat_index = np.ravel_multi_index(tuple(int_inds)[::-1], array.shape)
-                data[vdim.name].append(array.flat[flat_index])
+                if da and isinstance(array, da.Array):
+                    data[vdim.name].append(array.flatten()[tuple(flat_index)])
+                else:
+                    data[vdim.name].append(array.flat[flat_index])
         concatenated = {d: np.concatenate(arrays).flatten() for d, arrays in data.items()}
         return concatenated
 
--- a/tests/core/data/testdataset.py
+++ b/tests/core/data/testdataset.py
@@ -7,6 +7,11 @@ from nose.plugins.attrib import attr
 from itertools import product
 
 import numpy as np
+try:
+    import dask.array as da
+except ImportError:
+    da = None
+
 from holoviews import Dataset, HoloMap, Dimension, Image
 from holoviews.element import Distribution, Points, Scatter
 from holoviews.element.comparison import ComparisonTestCase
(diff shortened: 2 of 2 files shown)


jakirkham:
Thanks @philippjfr. Sorry for getting back to you so late. How would I use this to say instantiate a `hv.Dataset` using a Dask Array?

philippjfr:
It should work something like this:

```python
darr = da.from_array(np.random.rand(10, 20))

# Tuple constructor
hv.Dataset((range(20), range(10), darr), ['x', 'y'], 'z')

# Dict constructor
hv.Dataset({'x': range(20), 'y': range(10), 'z': darr}, ['x', 'y'], 'z')
```

Basically you declare coordinate arrays describing the values along the axes of your dask array and then pass them in as a tuple or dictionary. The order of key dimensions should be the reverse order of the axes of the array.

jakirkham:
Thanks for the tips. Managed to get that to work.

Minor note: `from_array` needs a `chunks` argument

jakirkham:
Anything else needed before this can go in?

philippjfr:
The Dataset tests for the ``aggregate`` method were highlighting some issues using numpy functions to aggregate the dask arrays. Haven't actually looked into what the problem is though.

philippjfr:
Looks like the errors were in the test code itself:

```
  File "/home/travis/build/ioam/holoviews/tests/testdataset.py", line 1522, in test_dataset_2D_aggregate_partial_hm
    Dataset({'x':self.xs, 'z': np.mean(array, axis=0).compute()},
  File "/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages/numpy/core/fromnumeric.py", line 2937, in mean
    return mean(axis=axis, dtype=dtype, out=out, **kwargs)
TypeError: mean() got an unexpected keyword argument 'out'
```

philippjfr:
This is ready to merge once tests pass.

holoviews/core/data/grid.py
```diff
@@ -301,6 +310,8 @@ def values(cls, dataset, dim, expanded=True, flat=True):
         if dim in dataset.vdims or dataset.data[dim.name].ndim > 1:
             data = dataset.data[dim.name]
             data = cls.canonicalize(dataset, data)
+            if da and isinstance(data, da.Array):
+                data = data.compute()
```
jakirkham: When does this get called? On what data is this being called on?

jakirkham:
Thanks @philippjfr. Had one question above. Otherwise LGTM.

holoviews/core/data/grid.py
philippjfr: This should only be called during plotting, and only on the subset of data that you are actually displaying. If you use a datashader operation this won't end up being called at all, and all the processing should happen out-of-core.

jlstevens:
Looks good to me as well and I am happy to see a good number of unit tests. I'm also happy to see that ``compute`` only needs to be called in one place. Merging.

philippjfr:
>happy to see a good number of unit tests

Note that what you are seeing is only a small fraction of unit tests, it's running the entire dataset test suite, which now includes about 130 tests.

holoviews/core/data/grid.py
jakirkham: Ok, thanks for clarifying. Sounds reasonable.

Does this get cached?

jakirkham:
Thanks all. :)

jakirkham:
@mrocklin, this may be of interest. ;)

holoviews/core/data/grid.py
philippjfr: No, but if you've got ideas about caching I'd love to hear them. I'm currently running into issues with some remote datasets, which end up being downloaded multiple times.

mrocklin: For the local scheduler there is http://dask.pydata.org/en/latest/caching.html

No opportunistic caching exists today for the distributed scheduler.  It's wouldn't be particularly hard to do though if someone wants to dive in.  Relevant issue here: https://github.com/dask/distributed/issues/681

jakirkham: Naively was thinking we might be able to call `persist` here. It would still be blocking in some cases (single machine scheduler), but no worse than calling `compute`. Though given we would want the array as well (in cases where we would get Futures), we would need somewhere to hang on to the Futures. Perhaps in Holoviews (i.e. an object of some kind), a Distributed Client (if using Distributed), somewhere else?

philippjfr: Is the use case here a multi-dimensional array which should be loaded into memory incrementally as you are exploring it? For example let's say you had a stack of images with x, y and z axes and you are browsing through the z-stack, you want each 2D array to persist in memory once it has been displayed? I think that would be a very useful option but I'd have to think about the best approach to support that. 

Currently our interfaces are stateless but I could imagine passing through a ``persist`` option to the interfaces in the constructor, which tells them to call ``.persist()`` before calling ``.compute()``. To make this concrete:

```python
darr = da.from_array(np.random.rand(30, 20, 20), (1, 1, 3))
ds = hv.Dataset((range(20), range(20), range(30), darr),
                ['x', 'y', 'z'], 'Value', persist=True)

z_stack = ds.to(hv.Image, ['x', 'y'])
z_stack[10]['Value']
```

In this example we create the 3D array, use ``.to`` to group it by the z-axis and then fetch the 2D array corresponding to z-value 10 and then use getitem to load the 'Value' array into memory. If we add the persist flag the array would then stay in memory and subsequent access to ``z_stack[10]['Value']`` would be much faster. Is this what you had in mind? If so we'll need to discuss whether interfaces could/should have state for backend specific options like this.

philippjfr:
Going to have to follow up on this, it seems the groupby implementation is loading all the data into memory.

holoviews/core/data/grid.py
mrocklin: I think that medium-term this is the sort of thing that should be handled on the Dask side.  We're well-positioned to solve this with minimal code-complexity elsewhere.  Again, you can probably do this on the local (non-distributed) scheduler today.

mrocklin: It would be interesting to see how calling these lines ahead of time affected performance:

```python
from dask.cache import Cache
cache = Cache(2e9).register()
```

jakirkham: To answer your question, @philippjfr, yes, basically as you described. We are looking at time instead of z (and the axis order differs), but for all intensive purposes this difference is merely semantic as far as Holoviews is concerned.

FWIW, as @mrocklin points out, we do have other options on the Dask side that could be helpful here or could be extended a bit (e.g. Cache support on Distributed). Issue ( https://github.com/dask/distributed/issues/681 ) outlines this problem a bit more. Alternatively there are some existing Dask options, which could allow this to be solved in user code.

There's a separate discussion that is a bit more interesting, which essentially involves having something like Holoviews drive computational priority in Dask. IOW one requests a large array be persisted (in memory, `persist`, or [on disk]( http://dask.pydata.org/en/latest/array-api.html#dask.array.store ), with `return_stored=True`), but would like to prioritize computations based on what a user would like to look at first. Outlined a bit more in issue ( https://github.com/dask/distributed/issues/1753 ).

holoviews/core/data/grid.py
```diff
@@ -491,7 +502,10 @@ def sample(cls, dataset, samples=[]):
                 data[d].append(arr)
             for vdim, array in zip(dataset.vdims, arrays):
                 flat_index = np.ravel_multi_index(tuple(int_inds)[::-1], array.shape)
-                data[vdim.name].append(array.flat[flat_index])
+                if da and isinstance(array, da.Array):
+                    data[vdim.name].append(array.flatten()[tuple(flat_index)])
```
jakirkham: Think the indexing is the issue here. If you switch to `vindex`, IIUC what `flat_index` means, this should be lazy.

jakirkham: ref: http://dask.pydata.org/en/latest/array-api.html#dask.array.Array.vindex

philippjfr: Thanks, the same issue also applied in ``select`` which was causing most of the problems. Followed up in https://github.com/ioam/holoviews/pull/2424.

PR #2307, opened 2018-02-05
Allowed custom sampling tolerance

jlstevens:
The tolerance for all types is now consistent with datetimes, you can now specify a global tolerance value using ``hv.config`` and tolerance values per ``Image`` by specifying ``rtol`` in the constructor.

TODO:

- [x] Add unit tests.
- [x] Add docs.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -73,10 +73,11 @@ class Config(param.ParameterizedFunction):
        recommended that users switch this on to update any uses of
        __call__ as it will be deprecated in future.""")
 
-    rtol = param.Number(default=10e-6, doc="""
-      The tolerance used to enforce regular sampling for gridded data
-      where regular sampling is expected. Expressed as the maximal
-      allowable sampling difference between sample locations.""")
+    image_rtol = param.Number(default=10e-6, doc="""
+      The tolerance used to enforce regular sampling for regular,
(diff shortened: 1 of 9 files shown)


holoviews/core/util.py
```diff
@@ -73,6 +73,11 @@ class Config(param.ParameterizedFunction):
        recommended that users switch this on to update any uses of
        __call__ as it will be deprecated in future.""")
 
+    rtol = param.Number(default=10e-6, doc="""
```
philippjfr: The config option should probably have a more specific name. Also it's not for all gridded data but specifically for Images.

jlstevens: Sure. ``image_rtol`` then? Or do you have an alternative suggestion?

jbednar:
Looks good, but ``image_rtol`` seems very obscure.  Maybe ``sampling_tolerance`` or something more readable like that?

philippjfr:
>Maybe sampling_tolerance or something more readable like that?

Sampling tolerance of what though, this is an option on the global config, so naming it something non-specific makes it seem like this will be used globally for all kinds of things.

jlstevens:
Ready for final review/merge when the tests go green.

philippjfr:
Looks good, tests are passing and coveralls seems stuck.

PR #2312, opened 2018-02-06
Fix/ignore matplotlibrc

laleph:
Overwrite the matplotlibrc setting `image.origin : lower` for RasterPlot and its children
Remove orphaned ImagePlot


philippjfr:
Thanks! Also good catch on the orphaned plot, must have been orphaned for a while.

PR #2314, opened 2018-02-06
Optimize VectorField min distance calculation

philippjfr:
Makes multiple improvements:

- Minimum distance only computed when required
- Matplotlib now consistent with bokeh (distances rescaled per frame)
- Attempts to use scipy's pdist functionality first, then falls back to own implementation (15x speedup)
- Slightly optimized existing minimum distance calculation (3x speedup)
- Filter annoying/useless warnings

- [x] Addresses https://github.com/ioam/holoviews/issues/2313


jlstevens:
Looks good. My only comment is it might be nice to have a few unit tests for ``get_min_distance`` just to double check the minimum value is consistent with and without scipy installed. Of course, it *should* be but a test will make sure that stays true.

jlstevens:
Tests are green. Merging.

PR #2329, opened 2018-02-10
Bumped Travis testing dependencies to latest versions

philippjfr:
Bumps the matplotlib version on Travis to 2.1.2 (i.e. latest). This is long overdue and **hopefully** will finally make the transient test failures disappear. Will require rebuilding test data of course.

### DIFF
--- a/tests/testrenderclass.py
+++ b/tests/testrenderclass.py
@@ -92,11 +92,6 @@ class MPLRendererTest(ComparisonTestCase):
         w, h = self.renderer.get_size(plot)
         self.assertEqual((w, h), (288, 288))
 
-    def test_get_size_tables_in_layout(self):
-        table = Table(range(10), kdims=['x'])
-        plot = self.renderer.get_plot(table+table)
-        w, h = self.renderer.get_size(plot)
-        self.assertEqual((w, h), (576, 230))
 
 @attr(optional=1)
 class BokehRendererTest(ComparisonTestCase):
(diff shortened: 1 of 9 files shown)


jlstevens:
Hopefully this works well. Just to be safe, we should make an explicit backup of the testing data before merging this PR. It will take a little while before we have a feel about the transients being better or worse so it is worth making sure we have the old testing data in case we have to revert in the worst case scenario.

philippjfr:
Looking quite hopeful that this resolves our transient issues. I restarted builds once so that's now 8 runs without a transient failure. Now that I've updated matplotlib it's time to start doing the same with pandas, dask, datashader and bokeh.

philippjfr:
I'd also like to switch from listing the travis dependencies in the .travis.yaml to using the environment.yml, with strict version requirements which will be guaranteed to work for testing and doc building.

philippjfr:
Uncovered a few bugs as part of this update. I expect one unit test to fail on this build, which I'll resolve in the morning. Once this is merged (along with the table PR) I think we're good to go for 1.9.4.

philippjfr:
Ready to merge now. I'll be restoring the disabled test in this PR https://github.com/ioam/holoviews/pull/2338. After a lot of test runs I can confidently say that transients are reduced but unfortunately not eliminated.

tests/testrenderclass.py
```diff
@@ -73,12 +73,12 @@ def test_get_size_single_plot(self):
     def test_get_size_row_plot(self):
         plot = self.renderer.get_plot(self.image1+self.image2)
         w, h = self.renderer.get_size(plot)
-        self.assertEqual((w, h), (576, 258))
+        self.assertEqual((w, h), (576, 255))
```
jlstevens: At least the changes here are small...

jlstevens:
Looks good. I'm a little surprised at the changes in the size tests but at least the changes are small. As you said, this is long overdue so I am happy to merge now.

PR #2330, opened 2018-02-10
Add Element.__contains__ method

philippjfr:
As discussed in https://github.com/ioam/holoviews/issues/540 this PR adds a ``__contains__`` method to Element types. This is consistent with the way xarray and pandas handle the ``__contains__`` method, and makes the common task of checking whether an element contains a particular dimension easier.

- [x] Implements https://github.com/ioam/holoviews/issues/540
- [x] Add tests

### DIFF
--- a/doc/Tutorials/Columnar_Data.ipynb
+++ b/doc/Tutorials/Columnar_Data.ipynb
@@ -315,7 +315,7 @@
    "source": [
     "xs = np.arange(10)\n",
     "curve = hv.Curve(zip(xs, np.exp(xs)))\n",
-    "curve * hv.Scatter(zip(xs, curve)) + curve.table()"
+    "curve * hv.Scatter(curve) + curve.table()"
    ]
   },
   {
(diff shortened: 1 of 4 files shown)


philippjfr:
The issue also mentions a ``__iter__`` method but I can't quite decide on the semantics I want. The only one that makes some sense is an iterator that emits a tuple for each observation in the flattened representation of the element. I've wanted that fairly frequently but I'm wondering whether that's too weird when dealing with gridded data. If we decide not to implement ``__iter__``, we should still implement it because currently ``list(hv.Dataset(...))`` simply stalls forever because defining ``__getitem__`` implicitly makes it an iterator, just not a valid one.

philippjfr:
I'll implement the ``__iter__`` to raise errors and add some tests.

jlstevens:
Looks good to me and I'm happy to merge when the tests are green - or is there a good reason not to expect the tests to be passing?

philippjfr:
No, I was expecting them to but something weird happened. Will try to figure that out now.

philippjfr:
Finally ready.

PR #2331, opened 2018-02-10
Ensured Tree types do not create lowercase node on attribute access

philippjfr:
As discussed in https://github.com/ioam/holoviews/issues/1182, Tree based objects (e.g. Overlay and Layout) should not insert new lower-case nodes when the attribute is lower case.

### DIFF
--- a/holoviews/core/tree.py
+++ b/holoviews/core/tree.py
@@ -221,20 +221,23 @@ class AttrTree(object):
 
         if not any(identifier.startswith(prefix)
                    for prefix in type(self)._disabled_prefixes):
-            identifier = type(self)._sanitizer(identifier, escape=False)
+            sanitized = type(self)._sanitizer(identifier, escape=False)
+        else:
+            sanitized = identifier
 
-        if identifier in self.children:
-            return self.__dict__[identifier]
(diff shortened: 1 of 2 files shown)


jlstevens:
I'm always nervous when I see changes to Tree behaviour. I agree with the changes as described but would you mind adding a few unit tests to check node insertion with lowercase names? There should already be tests inserting nodes starting with upper-case letters but if there aren't, a couple of those would be good to add as well.

philippjfr:
>I agree with the changes as described but would you mind adding a few unit tests to check node insertion with lowercase names?

No sounds good, mostly wanted you to review first in case you had any immediate objections in mind.

jlstevens:
Ok great. I'll merge once new unit tests are added and the tests are green.

philippjfr:
Turns out that AttrTree didn't implement delitem even though some methods referenced it so I ended up implementing it. I've also added a number of tests for AttrTree.

philippjfr:
Ready to merge.

philippjfr:
Let's merge this now.

jlstevens:
Ok. Looks good and worth getting some real testing out of it.

PR #2334, opened 2018-02-11
Pass operation to Operation hooks

philippjfr:
Recently we added hooks to operations which could do some pre- and post-processing on the data. The current implementation only passes the processed object itself when it would be more general if it also passed in the operation itself, which can be useful e.g. to apply pre-processing the operation parameters. The specific use-case I have in mind is ensuring that GeoViews returns RangeXY stream bounds in the correct coordinate system.


philippjfr:
Right, this "feature" hasn't been in a release yet so technically it's not an API change.

PR #2335, opened 2018-02-12
Fixed bug handling dask dataframe indexes

philippjfr:
Fixed bug handling dask dataframe types, since unlike pandas DataFrames their index does not have a ``names`` attribute in recent versions of dask.

I may also bump the dask version in this PR because this bug slipped by because we are currently only testing against dask=0.13.0.


jlstevens:
Looks good to me. Merging.

PR #2336, opened 2018-02-12
Fixed bug in bokeh TablePlot not escaping dimensions correctly

philippjfr:
The dimension names were not escaped on bokeh's TablePlot, resulting in no data being displayed when plotting Table containing columns with spaces or special characters in them.


philippjfr:
Ready to merge.

jlstevens:
Looks good. Merging.

PR #2338, opened 2018-02-13
Deprioritized array interface

philippjfr:
Finally implements https://github.com/ioam/holoviews/issues/1236, this is long overdue since the ArrayInterface mostly just causes troubles due to mixed types. With this change you should only ever end up with an ArrayInterface when you pass in an ``NxM`` array explicitly.

### DIFF
--- a/doc/Tutorials/Columnar_Data.ipynb
+++ b/doc/Tutorials/Columnar_Data.ipynb
@@ -46,7 +46,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "xs = range(10)\n",
+    "xs = np.arange(10)\n",
     "ys = np.exp(xs)\n",
     "\n",
     "table = hv.Table((xs, ys), kdims=['x'], vdims=['y'])\n",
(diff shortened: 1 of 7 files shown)


philippjfr:
Should be merged after https://github.com/ioam/holoviews/pull/2329. Should also reenable the Graph test that was disabled in that PR.

jlstevens:
Agreed this is overdue. #2329 is now merged so those tests should be re-enabled and the tests seem to have failed right now. Maybe fixing that (or a rebase) will get the tests green so I can merge?

I'm a little worried that this change might cause problems as it affects so many things. But I also believe we really should do it now...

philippjfr:
>I'm a little worried that this change might cause problems as it affects so many things. But I also believe we really should do it now...

It's revealing at least a few bugs caused by the ArrayInterface, which should be fixed by using the other interfaces. Tests also need updating because type comparisons are failing.

philippjfr:
Ready to merge.

jlstevens:
Looks good. I think this is important to do and hope it won't cause us any backwards compatibility headaches! Merging.

PR #2340, opened 2018-02-14
Pyflakes

philippjfr:
Adds this line to the travis.yml:

    flake8 --ignore=E,W,F999,F405 holoviews tests

Also fixes all current flake issues, fixing two undefined variable bugs in the process. In another PR we should clean up the install and consider moving some of the code out into scripts so we can clean up a bit and start using separate build stages for linting, tests and eventually package building.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
   - conda update -q conda
   # Useful for debugging any issues with conda
   - conda info -a
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION scipy=1.0.0 numpy freetype nose pandas=0.22.0 jupyter ipython=4.2.0 param matplotlib=2.1.2 xarray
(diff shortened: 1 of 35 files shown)


jlstevens:
There are a number of flakes I am in favour of detecting and raising as errors:

* Unreferenced variables
* Unused variables
* Unused imports

These typically do indicate errors or sensible ways the code should be simplified and cleaned up. I am strongly opposed to having other types of flakes, such as those to do with formatting or style issues being raised as errors on Travis.

I just want to be clear about this and I will not consider having other types of flakes being raised as errors unless someone can make a *very* compelling case.  As this PR does only raise errors for the key flake types listed above, I am happy to merge.

jbednar:
It would be nice if a github PR could have a friendly little button that one could click on to get warnings, i.e. things that shouldn't break the build but are possibly of interest or worth dealing with.  Right now it's a strict binary decision; if it doesn't break the build, we will basically never see anything about it. :-(

jlstevens:
@jbednar That would be a lovely Github feature to have indeed!

PR #2345, opened 2018-02-15
Handle range on heterogeneously typed Dataset column

philippjfr:
Calling range on a heterogeneously typed column currently causes all kinds of issues (including errors) due to python 3 sorting semantics. While range should generally not be called on object or string dtype columns it currently is so this needs to work robustly for the time being. Therefore this PR uses the python2sort utility to do the sorting robustly in python3.

### DIFF
--- a/holoviews/core/data/pandas.py
+++ b/holoviews/core/data/pandas.py
@@ -141,6 +141,7 @@ class PandasInterface(Interface):
                 column = column.sort(inplace=False)
             else:
                 column = column.sort_values()
+            column = column[~column.isin([None])]
             return column.iloc[0], column.iloc[-1]
         else:
             return (column.min(), column.max())
--- a/holoviews/core/data/pandas.py
+++ b/holoviews/core/data/pandas.py
@@ -142,6 +142,8 @@ class PandasInterface(Interface):
             else:
                 column = column.sort_values()
             column = column[~column.isin([None])]
+            if not len(column):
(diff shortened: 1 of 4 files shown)


jlstevens:
This is one semantic change in Python 3 that I find to be downright annoying. Anyway, the suggested fix makes sense.

holoviews/core/data/pandas.py
```diff
@@ -291,6 +294,16 @@ def dframe(cls, columns, dimensions):
...
 
+    @classmethod
+    def array(cls, dataset, dimensions):
+        if not dimensions:
+            dimensions = dataset.dimensions(label='name')
+        else:
+            dimensions = [dataset.get_dimensions(d).name for d in dimensions]
+        inds = [dataset.data.columns.index(dim.name) for dim in dimensions]
+        return dataset.data.values[:, inds]
+
```
jlstevens: I don't quite see how this relates to the python 2 vs 3 sorting issue. Is this method implementation an unrelated fix?

philippjfr: Yes, accidentally pushed this. Will revert.

philippjfr:
Going to need rebuilt test data, but let's get https://github.com/ioam/holoviews/pull/1978 merged first since that also requires new test data.

jlstevens:
Happy to merge when the tests go green.

PR #2348, opened 2018-02-16
Fallback for unknown dask categories

philippjfr:
When the categories on a categorical column aren't known it throws an NotImplementedError, in this case this falls back to getting the categories the usual way.


philippjfr:
Ready to merge.

PR #2354, opened 2018-02-19
Option error refactor

philippjfr:
Picks up where https://github.com/ioam/holoviews/pull/2306 left off, completely refactoring option error handling. All methods of setting options will now either warn or error when given an invalid option. Could consider readding a skip option for the OptionsMagic, because that won't let you be explicit about the backend you're trying to set options on yet.

- [ ] Update existing unit tests
- [ ] Add new unit tests

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -1132,7 +1132,7 @@ class Dimensioned(LabelledData):
         If no opts are supplied all options on the object will be reset.
         Disabling clone will modify the object inplace.
         """
-        backend = backend or Store.current_backend
+        selected_backend = backend or Store.current_backend
         if isinstance(options, basestring):
             from ..util.parser import OptsSpec
             try:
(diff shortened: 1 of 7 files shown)


philippjfr:
This is far from ready but I wanted to discuss an idea. I'd like to make option validation as strict as possible in future so you always get immediate feedback if you supplied an invalid option. By default this PR will ensure that ``.opts``, ``.options`` and ``hv.opts`` will be strict about validation but because the magic does not yet have a way of being explicit about the backend I would not yet want that to be strict. Therefore I'd like to add two config options something like ``hv.config.strict_options`` and ``hv.config.strict_magics``.

For testing I'd want both to be enabled, which will catch any examples that are using incorrect options.

philippjfr:
I also want to enable warnings when using ``__call__`` to set options in this PR. Finally as a reminder to myself ``.options()`` without any arguments doesn't seem to be handled currently.

PR #2357, opened 2018-02-20
Decode bytes in bokeh plots

philippjfr:
Attempt to decode bytes types passed to bokeh plots. As part of the Intake project I'm building a HoloViews based plotting API, many of the data sources in this project includes ``bytes`` types even in Python3 which currently cannot be serialized by bokeh causing errors. In this API its not really practical to convert the bytes types to strings, so it would be helpful if we could just assume utf-8 encoding (which will be most common). This will of course not always work but based on my testing it works in most cases, which is better than the current case where it works in no cases.


philippjfr:
Ready to review/merge.

jlstevens:
Looks good though I think this issue needs more thorough investigation before we can be confident that we have the most general/robust fix. If you file an issue as a reminder to do that, I'm happy to merge this as a fix (even if it only a partial fix, it will handle more inputs without erroring).

jlstevens:
Thanks for filing the issue. Merging.

PR #2360, opened 2018-02-21
Fixed clip parameter on Violin

philippjfr:
The clip option on ViolinPlot made no sense as it would actually override any custom ranges.

### DIFF
--- a/holoviews/plotting/bokeh/stats.py
+++ b/holoviews/plotting/bokeh/stats.py
@@ -277,8 +277,8 @@ class ViolinPlot(BoxWhiskerPlot):
         Allows supplying explicit bandwidth value rather than relying
         on scott or silverman method.""")
 
-    clip = param.Boolean(default=True, doc="""
-        Whether to clip the violins at the max and min of the data.""")
+    clip = param.NumericTuple(default=None, length=2, doc="""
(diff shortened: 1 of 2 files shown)


jlstevens:
Looks good to me. Is ``clip`` a parameter unique to ``ViolinPlot``? Is so, I am happy to merge now, otherwise I hope this parameter is consistent with any other parameters called ``clip``. And if it is the former, I suppose ``clip`` does sounds like a very general parameter name that we might want for something more general than just violin plots?

philippjfr:
It is in fact unique as of right now but it would also be useful to have on Distribution plots. Would do that in a new PR though.

jlstevens:
Ok. Sounds like ``clip`` will apply to multiple plot types in future. That said, the notion of clipping to the min/max seems like something more general than even just the statistical plots...

philippjfr:
>That said, the notion of clipping to the min/max seems like something more general than even just the statistical plots...

Hmm, in those cases you can select on the data, but for statistical plots you might want to include outliers in the density calculation but not sample the density estimate beyond a certain clipping range. So I'm not in favor of adding a clip parameter to other plot types.

jlstevens:
Ok. Given that you seem happy with this parameter name, I'll go ahead and merge.

PR #2363, opened 2018-02-21
fixed color option in code snippet

ea42gh:
the bokeh reference page for BoxWhisker has an element option that needed updating


philippjfr:
Looks good thanks!

PR #2368, opened 2018-02-22
Support for JupyterLab

philippjfr:
This PR adds initial support for working in JupyterLab. So far it supports plotting static plots and embedded widgets by piggybacking on top of the bokeh JupyterLab extension. To make that work I have to do a number of pretty terrible things such as dynamically loading jQuery and require, so in the long term this isn't a good solution. However in the long term we plan on rewriting the widgets anyway, so this is a stopgap solution to get our foot in the door.

The main changes that were required to get this to work were:

* Put the HoloViews widgets and JS utilities into a global HoloViews namespace (rather than making them all global)
* Split the core JS code from the HTML and JS templates
* Add ``Renderer.components`` method which returns data and metadata dictionaries both indexed by ``MIME_TYPE``
* Have the display hooks render the HTML with publish_display_data and then publish the JS code in a separate step

Current issues:

* The display hooks were designed to return the HTML rather than publishing it using ``publish_display_data``, however an HTML blob cannot work in this new approach since the HTML needs to be published before the JS code.
* No handling of Comms yet, this will follow in a later PR which isolates all comm code on ``HoloViews.CommManager`` so we can ship a CommManager that handles Jupyter Notebook & JupyterLab but also allow overriding the CommManager for other environments.

screen shot 2018-02-22 at 1 46 49 pm

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -11,7 +11,7 @@ from ..core.tree import AttrTree
 from ..core.options import Store
 from ..element.comparison import ComparisonTestCase
 from ..util import extension
-from ..plotting.renderer import Renderer
+from ..plotting.renderer import Renderer, MIME_TYPES
 from .magics import load_magics
 from .display_hooks import display  # noqa (API import)
 from .display_hooks import set_display_hooks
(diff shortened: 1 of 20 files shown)


jbednar:
Can you clarify what the user-level impact of the "Current issues" is?  E.g. with no comms, presumably no DynamicMap and no Streams?

jlstevens:
> Can you clarify what the user-level impact of the "Current issues" is? E.g. with no comms, presumably no DynamicMap and no Streams?

Right.

philippjfr:
Going to close as it is now superseded by https://github.com/ioam/holoviews/pull/2377.

PR #2372, opened 2018-02-23
Add CommManager classes 

philippjfr:
Builds on https://github.com/ioam/holoviews/pull/2368 abstracting all Comm handling onto so called CommHandler classes. By default HoloViews will ship with a JupyterCommHandler, which will handle both the classic notebook and JupyterLab.

Here is the current status of JupyterLab support:

Static plots
-------------

* Static plots work (all backends)
* Embedded widgets work (all backends)

Backend -> Frontend
-------------------------

* Can send data from backend to frontend and update matplotlib and plotly plots
* Cannot yet update bokeh plots

Frontend -> Backend
-------------------------

* Bokeh callbacks work
* Dynamic and Live widgets work

- [x] Handle bokeh messages (either by exposing the API in bokeh or by duplicating some of that code in HoloViews)
- [ ] Add tests
- [x] Review API and docstrings with @jlstevens


philippjfr:
I've now added a bokeh message handler, which will now allow custom Comm types to trigger updates in a bokeh plot. To make this work I had to make a one-line change in bokeh's main.ts to make he ``Receiver`` class publicly available:

```
export {Receiver}              from "./protocol/receiver"
```

This is responsible for decoding binary data and combining it with the JSON message, and is not something we would want to duplicate in HoloViews.

philippjfr:
@jlstevens I think it would be easiest to simply merge this PR and close https://github.com/ioam/holoviews/pull/2368, unless you really think reviewing them completely separately would help. In terms of the actual code this is now ready, but it would be great if you could write the docstrings for the ``CommManager`` and review the API as we discussed.

tests/testutils.py
```diff
@@ -40,6 +44,8 @@ def tearDown(self):
         if bokeh:
             Store.renderers['bokeh'] = bokeh.BokehRenderer.instance()
         OutputSettings.options =  OrderedDict(OutputSettings.defaults.items())
+        for renderer in Store.renderers.values():
+            renderer.comm_manager = CommManager
```
jlstevens: Why does ``comm_manager`` need to be set explicitly here? Isn't it the default?

tests/ipython/testmagics.py
```diff
@@ -9,12 +9,15 @@
     raise SkipTest("Required dependencies not satisfied for testing magics")
 
 from holoviews.operation import Compositor
+from holoviews.plotting.comms import CommManager
```
jlstevens: Lots of setting of ``comm_manager`` across the tests (made the same comment again below). Why is this needed?

holoviews/ipython/__init__.py
```diff
@@ -224,19 +224,25 @@ def load_hvjs(cls, logo=False, bokeh_logo=False, mpl_logo=False, plotly_logo=Fal
...
-                                      'mpl_logo':    mpl_logo,
-                                      'plotly_logo': plotly_logo,
-                                      'message':     message})))
+        html = template.render({'widgetcss':   widgetcss,
+                                'logo':        logo,
+                                'bokeh_logo':  bokeh_logo,
+                                'mpl_logo':    mpl_logo,
+                                'plotly_logo': plotly_logo,
+                                'message':     message})
+        publish_display_data(data={'text/html': html})
```
jlstevens: I assume ``display(HTML(...))`` still worked and was equivalent to this, but this is more consistent now so I am happy with this change.

holoviews/ipython/__init__.py
```diff
@@ -224,19 +224,25 @@ def load_hvjs(cls, logo=False, bokeh_logo=False, mpl_logo=False, plotly_logo=Fal
...
+        if JS:
+            publish_display_data(data={
+                MIME_TYPES['js']   : widgetjs,
+                MIME_TYPES['jlab-hv-load'] : widgetjs
+            })
```
jlstevens: I'll add a comment to explain that jupyterlab will only process the non-standard mime type.

holoviews/ipython/__init__.py
philippjfr: Right, ``display`` calls ``publish_display_data`` internally.

tests/testutils.py
philippjfr: Various tests run the notebook extension which registers the Jupyter comms.

holoviews/ipython/display_hooks.py
```diff
@@ -59,7 +60,13 @@ def render(obj, **kwargs):
     # Drop back to png if pdf selected, notebook PDF rendering is buggy
     if renderer.fig == 'pdf':
         renderer = renderer.instance(fig='png')
-    return renderer.html(obj, **kwargs)
+
+    data, metadata = renderer.components(obj, **kwargs)
```
jlstevens: ``renderer.components`` makes sense to me though I can imagine it could be useful for giving the components in situations outside of jupyterlab/notebook. Is the output dependent on the renderer ``mode``? Is the plan to make ``components`` useful in different contexts or for different environments?

holoviews/ipython/display_hooks.py
```diff
@@ -175,7 +182,13 @@ def element_display(element, max_frames):
...
     if renderer.fig == 'pdf':
         renderer = renderer.instance(fig='png')
-    return renderer.html(element, fmt=renderer.fig)
+
+    data, metadata = renderer.components(element)
+    html = data.pop('text/html')
+    publish_display_data({'text/html': html})
+    if data or metadata:
+        publish_display_data(data, metadata=metadata)
+    return ''
```
jlstevens: If this code were to repeat once more, I would suggest a helper function.

holoviews/plotting/bokeh/bokehwidgets.js
```diff
@@ -13,26 +13,48 @@ BokehScrubberWidget.prototype = Object.create(ScrubberWidget.prototype);
 
 // Define methods to override on widgets
 var BokehMethods = {
```
jlstevens: In ES6 we would make this a base class I think...

holoviews/ipython/display_hooks.py
philippjfr: It's not specific to the notebook, e.g. in theory you could use the output in a flask app, and with a ``SocketIOCommManager`` it would all work.

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -210,32 +206,23 @@ class CustomJSCallback(MessageCallback):
...
           return
         }}
 
         // Initialize event queue and timeouts for Comm
-        var comm_state = HoloViewsWidget.comm_state["{comm_id}"];
+        var comm_state = HoloViews.comm_state["{comm_id}"];
         if (comm_state === undefined) {{
             comm_state = {{event_buffer: [], blocked: false, time: Date.now()}}
-            HoloViewsWidget.comm_state["{comm_id}"] = comm_state
+            HoloViews.comm_state["{comm_id}"] = comm_state
```
jlstevens: Would rename to ``comm_status`` but I don't mind if that doesn't make it into this PR (though that would be good).

holoviews/plotting/bokeh/plot.py
```diff
@@ -74,6 +74,11 @@ def document(self):
         return self._document
 
 
+    @property
+    def id(self):
+        return self.state._id
```
jlstevens: Sure using ``_id`` is fine here? I assume this is ``_id`` in bokeh not holoviews...

holoviews/plotting/bokeh/renderer.py
```diff
@@ -31,6 +30,28 @@
 </script>
 """
 
+bokeh_msg_handler = """
+var plot = Bokeh.index["{plot_id}"];
```
jlstevens: This is the block of JS that becomes the body of the message handler callback? I might add a comment if so...

holoviews/plotting/bokeh/renderer.py
```diff
@@ -31,6 +30,28 @@
 </script>
 """
 
+bokeh_msg_handler = """
```
jlstevens: This is the block of JS that becomes the body of the message handler callback? I might add a comment if so...

holoviews/plotting/bokeh/renderer.py
```diff
@@ -71,11 +92,11 @@ class BokehRenderer(Renderer):
     backend_dependencies = {'js': CDN.js_files if CDN.js_files else tuple(INLINE.js_raw),
                             'css': CDN.css_files if CDN.css_files else tuple(INLINE.css_raw)}
 
-    comms = {'default': (JupyterComm, None),
-             'server': (Comm, None)}
-
     _loaded = False
 
+    # Define the handler for updating matplotlib plots
```
jlstevens: Outdated/incorrect comment? Or does this really involve matplotlib somehow?

holoviews/plotting/bokeh/renderer.py
```diff
@@ -236,14 +257,16 @@ def _figure_data(self, plot, fmt='html', doc=None, **kwargs):
         try:
             js, div, _ = notebook_content(model, comm_id)
```
jlstevens: Before the variable ``div`` just got clobbered and returned? If so it seems like a bug that could have broken things!

holoviews/plotting/bokeh/renderer.py
```diff
@@ -306,6 +329,8 @@ def load_nb(cls, inline=True):
         """
         Loads the bokeh notebook resources.
         """
-        from bokeh.io.notebook import curstate
+        LOAD_MIME_TYPE = bokeh.io.notebook.LOAD_MIME_TYPE
+        bokeh.io.notebook.LOAD_MIME_TYPE = MIME_TYPES['jlab-hv-load']
```
jlstevens: Is this something bokeh expects you to do?

philippjfr: No, it's me temporarily switching which JupyterLab extension to publish to. It means you won't have to install both the bokeh and the holoviews extension.

holoviews/plotting/bokeh/widgets.py
```diff
@@ -273,13 +275,10 @@ def _plot_figure(self, idx, fig_format='json'):
         first call and
         """
         self.plot.update(idx)
-        if self.embed or fig_format == 'html':
-            if fig_format == 'html':
```
jlstevens: I assume ``fig_format == 'html'`` is still handled but somewhere else I haven't looked at yet...

holoviews/plotting/bokeh/renderer.py
philippjfr: No, logic was fine before, just overwrote the variable.

holoviews/plotting/bokeh/renderer.py
philippjfr: No, definitely wrong, must be leftover from the very beginning of the bokeh interface!

holoviews/plotting/comms.py
```diff
@@ -222,3 +212,93 @@ def send(self, data=None, buffers=[]):
...
+        comm = cls.server_comm(plot, id, on_msg)
+        cls._comms[comm.id] = comm
+        return comm
+
+    @classmethod
+    def get_client_comm(cls, plot, on_msg=None, id=None):
+        comm = cls.client_comm(plot, id, on_msg)
+        cls._comms[comm.id] = comm
+        return comm
+
```
jlstevens: These two classmethods look nice and symmetrical but I assume that the comm needs to be either registered in Python or JS respectively with the registering happening in the right order...

holoviews/plotting/bokeh/widgets.py
philippjfr: Previously it was used both for the initial frame and for the bokeh plot update messages, now the initial frame is handled separately.

holoviews/plotting/bokeh/renderer.py
philippjfr: Definitely a good idea.

holoviews/plotting/comms.py
```diff
@@ -222,3 +212,93 @@ def send(self, data=None, buffers=[]):
...
+
+
+class JupyterCommManager(CommManager):
+    """
+    The CommManager is an abstract baseclass for establishing
+    Websocket Comms on the client and the server.
+    """
```
jlstevens: I'll be expanding on this docstring.

holoviews/plotting/bokeh/plot.py
philippjfr: Yeah, this is bokeh, not sure why bokeh doesn't make ``_id`` public, it's pretty central to a lot of things. I'll double check there isn't a public API.

holoviews/plotting/mpl/comms.py
```diff
@@ -13,8 +13,8 @@
 
 mpl_msg_handler = """
 /* Backend specific body of the msg_handler, updates displayed frame */
-target = $('#fig_{comm_id}');
-img = $('<div />').html(msg);
+var target = $('#fig_{plot_id}');
+var img = $('<div />').html(msg);
```
jlstevens: Ouch! This could have gone badly wrong I think!

holoviews/plotting/comms.py
philippjfr: Yeah, the actual Comm classes handle that part internally.

holoviews/plotting/mpl/renderer.py
```diff
@@ -177,11 +176,11 @@ def diff(self, plot):
                 figure_format = self.params('fig').objects[0]
             else:
                 figure_format = self.fig
-            data = self.html(plot, figure_format, comm=False)
+            data = self.html(plot, figure_format)
         return data
 
 
-    def _figure_data(self, plot, fmt='png', bbox_inches='tight', **kwargs):
+    def _figure_data(self, plot, fmt='png', bbox_inches='tight', as_script=False, **kwargs):
```
jlstevens: Need to document ``as_script`` appropriately. I assume it is about whether the return should be split into a separate JS script? How would you describe it?

holoviews/plotting/mpl/comms.py
philippjfr: Yes, making sure all the ids match up was the hardest part of this PR. If you want to do some testing I've got a notebook with all the various combinations of widgets/comms/backends.

holoviews/plotting/plot.py
```diff
@@ -527,7 +532,7 @@ def init_comm(self):
         """
         comm = None
         if self.dynamic or self.renderer.widget_mode == 'live':
-            comm = self.renderer.comms[self.renderer.mode][0](self)
```
jlstevens: I assume it was parameterized by ``mode`` was because of nbagg support? Does that still work?

holoviews/plotting/mpl/renderer.py
philippjfr: Correct. I'm not hugely happy shoving this onto ``_figure_data`` but it seemed like the only reasonable approach without duplicating a lot of code.

holoviews/plotting/plot.py
philippjfr: Ah, forgot about that, I was going to suggest deprecating it, I doubt it works right now.

holoviews/plotting/renderer.py
```diff
@@ -171,7 +177,10 @@ def get_plot(self_or_cls, obj, renderer=None):
                 initialize_dynamic(obj)
             obj = Compositor.map(obj, mode='data', backend=self_or_cls.backend)
 
-        if not renderer: renderer = self_or_cls.instance()
+        if not renderer:
+            renderer = self_or_cls
+            if inspect.isclass(self_or_cls):
```
jlstevens: Is ``inspect`` really needed? Would something like ``if isinstance(self_or_cls, type) and issubclass(self_or_cls, Renderer)`` not work?

holoviews/plotting/renderer.py
```diff
@@ -276,16 +285,37 @@ def html(self, obj, fmt=None, css=None, comm=True, **kwargs):
...
+            plot_id = plot.plot_id
         else:
-            return html
+            html, js = self._figure_data(plot, as_script=True, **kwargs)
+            plot_id = plot.id
+            if comm and plot.comm is not None and self.comm_msg_handler:
+                msg_handler = self.comm_msg_handler.format(plot_id=plot_id)
+                html = plot.comm.html_template.format(init_frame=html,
+                                                      plot_id=plot_id)
+                js += plot.comm.js_template.format(msg_handler=msg_handler,
```
jlstevens: Simply concatenating the JS is a little risky if you don't have newlines at the transition.

philippjfr: Good point, worth fixing.

holoviews/plotting/renderer.py
philippjfr: Tbh, even ``not isinstance(self_or_cls, Renderer)`` should be enough.

holoviews/plotting/renderer.py
```diff
@@ -386,7 +419,7 @@ class needed to render it with the current renderer.
 
 
     @classmethod
-    def html_assets(cls, core=True, extras=True, backends=None):
+    def html_assets(cls, core=True, extras=True, backends=None, script=False):
```
jlstevens: Is this the only place where we will have to worry about embedding JS into script tags?

holoviews/plotting/widgets/__init__.py
```diff
@@ -206,7 +227,6 @@ def update(self, key):
         else:
             self.plot.update(key)
             self.plot.push()
-        return 'Complete'
```
jlstevens: Wasn't this an ad hoc 'message' that meant something?

holoviews/plotting/renderer.py
philippjfr: I think so.

holoviews/plotting/widgets/__init__.py
philippjfr: It used to be the ACK for kernel.execute, that's now done via a comm.

holoviews/plotting/widgets/__init__.py
```diff
@@ -338,13 +358,12 @@ def get_widgets(self):
...
                 dim_vals = escape_list(escape_vals(dim_vals))
                 next_vals = escape_dict({k: escape_vals(v) for k, v in next_vals.items()})
-
             visibility = '' if visible else 'display: none'
             dim_str = dim.pprint_label
             escaped_dim = dimension_sanitizer(dim_str)
             widget_data = dict(dim=escaped_dim, dim_label=dim_str,
                                dim_idx=idx, vals=dim_vals, type=widget_type,
-                               visibility=visibility, step=step, next_dim=next_dim,
+                               visibility=visibility, step=step, next_dim=next_dim or None,
```
jlstevens: ``or None`` is an unrelated bug fix?

philippjfr: Not unrelated no, to do with templating the variable now that it is a function argument rather than templated directly into the JS code.

holoviews/plotting/widgets/widgets.js
```diff
@@ -13,14 +10,13 @@ HoloViewsWidget.prototype.init_slider = function(init_val){
 }
 
 HoloViewsWidget.prototype.populate_cache = function(idx){
-  this.cache[idx].html(this.frames[idx]);
+  this.cache[idx].innerHTML = this.frames[idx];
```
jlstevens: ``innerHTML`` can be quite different from ``html`` - I think something to do with escaping. So this is a bugfix?

holoviews/plotting/widgets/widgets.js
```diff
@@ -36,83 +32,77 @@ HoloViewsWidget.prototype.dynamic_update = function(current){
...
     } else {
-      frame = i;
+      var frame = i;
     }
     if(!(frame in this.cache) || force) {
       if ((frame in this.cache) && force) { this.cache[frame].remove() }
-      this.cache[frame] = $('<div />').appendTo("#"+"_anim_img"+this.id).hide();
+      var div = document.createElement("div");
+      var parent = document.getElementById("_anim_img"+this.id);
+      div.style.display = "none";
```
jlstevens: Looks like some static styling that should be done by CSS?

holoviews/plotting/widgets/widgets.js
philippjfr: No, this is converting jQuery to pure JS.

holoviews/plotting/widgets/widgets.js
philippjfr: No, not static, this is just emulating what ``.hidden()`` in jQuery does using vanilla JS.

holoviews/plotting/widgets/widgets.js
```diff
@@ -139,8 +129,8 @@ SelectionWidget.prototype.get_key = function(current_vals) {
   var key = "(";
   for (var i=0; i<this.slider_ids.length; i++)
   {
-    val = this.current_vals[i];
-    if (!(typeof val === 'string')) {
+    let val = this.current_vals[i];
```
jlstevens: Should probably still be ``var``. I am strongly in favor of switching to ``let`` and ES6 but not in this PR!

philippjfr: Right, that was accidental.

holoviews/plotting/widgets/__init__.py
philippjfr: On a more general note, there is probably a relatively clean path to progressively moving to the widget system we envision. I'd probably start by moving away from templating and encapsulate all this stuff in a JSON blob.

holoviews/plotting/widgets/widgets.js
```diff
@@ -346,3 +337,198 @@ function update_widget(widget, values) {
...
+    noConflict = !!oldjQuery;
+  } else {
+    // register the current jQuery
+    define('jquery', [], function() { return jQuery; });
+  }
+  if (!jQueryUI) {
+	paths.jQueryUI = jQueryUI_CDN.slice(null, -3);
+  } else {
+    define('jQueryUI', [], function() { return jQuery.ui; });
+  }
```
jlstevens: This is a hacky JQueryUI/regexp/require nightmare! I won't suggest touching it if it works but we should try to get rid if we ever see the opportunity!

tests/testutils.py
jlstevens: Ok. And those won't be available in this context. Seems fine.

holoviews/ipython/display_hooks.py
jlstevens: Ok. I was just making sure the intention behind the method was to be as general as possible.

holoviews/plotting/bokeh/renderer.py
jlstevens: Ok. Anyway, it looks cleaner now.

holoviews/plotting/bokeh/renderer.py
jlstevens: Ok. That sounds good as long as we feel this approach will be robust enough.

holoviews/plotting/mpl/comms.py
jlstevens: Sounds horrible. I'll pass, thanks! ;-p

holoviews/plotting/plot.py
jlstevens: I'm fully in favor of deprecating nbagg myself. @jbednar ?

holoviews/plotting/widgets/widgets.js
philippjfr: Yes, absolutely the first thing I want to get rid of. As I pointed out above, some of the widget code is salvageable. One of the first steps to doing that will be to get rid of the jQuery(UI) dependency. One of the reasons why I actually like a JupyterLab extension is because I don't have to fight require anymore.

holoviews/plotting/renderer.py
jlstevens: I prefer not to use ``inspect`` unless it is really necessary...

holoviews/plotting/widgets/__init__.py
jlstevens: Ok. Glad to see it go.

holoviews/ipython/display_hooks.py
philippjfr: Sort of unfortunate you decided to reuse the IPython comms in labmode as it would have made for a good test. May have a go at a Flask+SocketIO example at some point.

holoviews/plotting/widgets/widgets.js
jlstevens: Ok. But I am not 100% sure that these two things are exactly equivalent. They may be but subtle annoyances like escaping can creep into things that look simple like this. I'm happy if it works and it probably does if the HTML isn't expected to have special characters that need escaping.

holoviews/plotting/widgets/widgets.js
jlstevens: Got it.

jlstevens:
Done a pass over the PR. Overall, I am very happy with it: things are now more extensible and also more general without adding lots more code to support jupyterlab. There are still a lot of moving parts and it isn't trivial to understand but there is less repetition now and once you get your head around it, the design is much saner. I think things could be further improved with a widget refactor and switching to ES6 but this is already a very nice improvement for one PR!

After you've made the fairly trivial changes I have suggested, I'll think about where to add some description to the docstrings (probably in the comms/comm manager) and then I am happy to merge.

One thing that might hold things up is if you want to get the bokeh tests working again before merging?

jlstevens:
The description of the ``as_script`` argument isn't really right for matplotlib as it just returns ``html`` and an empty string. This seems a little odd/inconsistent as I assume JupyterLab would enforce the HTML/JS split for matplotlib too? Or is the entire JS component handled by the extension  for matplotlib?

jlstevens:
I've now updated the comments and docstrings. Please check that what I say is accurate and I am happy to elaborate on anything you think I should.

After that, once the minor tweaks mentioned in the comments are addressed (and ideally once the tests are working again), I am happy to merge.

philippjfr:
>The description of the as_script argument isn't really right for matplotlib as it just returns html and an empty string. This seems a little odd/inconsistent as I assume JupyterLab would enforce the HTML/JS split for matplotlib too? Or is the entire JS component handled by the extension for matplotlib?

``_figure_data`` is only meant to return the data required to display the plot, not any widget or comm code around it, which is handled by other methods. Perhaps returning None would be a bit more explicit?

holoviews/plotting/bokeh/callbacks.py
philippjfr: Done.

holoviews/plotting/bokeh/plot.py
philippjfr: ``obj.ref['id']`` works

jlstevens:
> Perhaps returning None would be a bit more explicit?

It is fine, I was just wondering what was going on in the matplotlib case.

holoviews/plotting/plot.py
philippjfr: I've now removed nbagg.

philippjfr:
The tests should be passing shortly at which point we're only waiting on https://github.com/bokeh/bokeh/pull/7567 to be merged at which point I'll release ``jupyterlab_holoviews`` 0.1.0 and will let @jlstevens test the JupyterLab support.

jlstevens:
@philippjfr I have now tested it with Jupyterlab and it seems to work. I tested a simple streams example and a linked stream (``PositionXY``).

Other than needing to build bokeh against the relevant PR, the steps were:

```bash
conda install -c conda-forge jupyterlab
jupyter labextension install @pyviz/jupyterlab_holoviews
jupyter-lab
```
This is pretty horrible as there is now a second step users need to run and it is *dog slow*. It took 10 minutes or so to install the extension! I don't think this is related to this PR or the holoviews extension in particular so I am happy to merge.

stonebig:
installing jupyterlab extension is dead slow on my PC too, I fear the slowness is a proof of success.

jlstevens:
Tests have passed now. Merging.

PR #2377, opened 2018-02-27
Fixed bug finding Layout label

philippjfr:
Fixes a bug revealed by https://github.com/ioam/holoviews/pull/2331 currently messing with some tests. Before that PR it would try to access Layout.label and simply return a new Layout node, but now this causes issues.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -1195,7 +1195,7 @@ def get_ndmapping_label(ndmapping, attr):
             el = next(els)
         except StopIteration:
             return None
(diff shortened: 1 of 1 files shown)


jlstevens:
How come ``getattr`` doesn't also create a node?

philippjfr:
It would have before, since the fix it doesn't, causing this issue. Turns out getattr wasn't validated at all before, only ``__getitem__`` was.

philippjfr:
Mind merging this? Then I can fix the test data in the comms PR.

PR #2382, opened 2018-03-01
Fixed Renderer widget export

philippjfr:
After adding JupyterLab support the widgets were returning separate HTML and JS components which were not appropriately assembled into a single document on export.

This PR restores the old behavior on ``NdWidgets`` and adds a keyword to return the components separately matching the keyword on ``Renderer._figure_data``.

- [x] Fixes https://github.com/ioam/holoviews/issues/2381


philippjfr:
@jlstevens This is ready to merge, the push build won't pass because it's still pulling the old data but the PR build passed.

jlstevens:
Looks good to me. Merging.

PR #2383, opened 2018-03-01
Workaround for datetime streaming bug in bokeh

philippjfr:
Very ugly workaround for https://github.com/bokeh/bokeh/issues/7587 in bokeh datetime streaming.

### DIFF
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -15,10 +15,12 @@ from ..plot import (DimensionedPlot, GenericCompositePlot, GenericLayoutPlot,
                     GenericElementPlot, GenericOverlayPlot)
 from ..util import attach_streams
 from .util import (layout_padding, pad_plots, filter_toolboxes, make_axis,
-                   update_shared_sources, empty_plot, decode_bytes)
+                   update_shared_sources, empty_plot, decode_bytes,
+                   bokeh_version)
 
 from bokeh.layouts import gridplot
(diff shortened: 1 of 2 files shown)


stonebig:
so my code with bokeh-0.12.5.dev1 will break  also ? would it be possible to include it in the test ?

philippjfr:
>so my code with bokeh-0.12.5.dev1 will break also ? would it be possible to include it in the test ?

Probably a good idea.

philippjfr:
Ready to merge.

philippjfr:
Looks like this change will require pinning to bokeh>=0.12.14, which also affects 1.9.5 since I ported this change there.

jlstevens:
> Looks like this change will require pinning to bokeh>=0.12.14...

Are you going to pin bokeh in this PR? I would also like to see an issue to remove this hack once it is fixed in bokeh to make sure we remove this code when we can.

Otherwise, happy to merge.

philippjfr:
>I would also like to see an issue to remove this hack once it is fixed in bokeh to make sure we remove this code when we can.

Unless we want to remove support for bokeh 0.12.14 we can't really delete this.

jlstevens:
> Unless we want to remove support for bokeh 0.12.14 we can't really delete this. 

We will be happy to remove support for bokeh 0.12.14 *eventually* so I would still file an issue. Maybe assign it to the 2.0 milestone.

jlstevens:
Looks good. Merging.

Issue #2380, opened 2018-02-28
Image Element does not store rtol parameter


scaine1:
When the rtol parameter is supplied to the hv.Image element, rtol does not seem to be stored or associated with the element instance in a permanent way.
This means that when the Image element is cloned, the rtrol parameter is lost and further use of the image leads to failure.

Below is some code that I have mocked up to highlight the example. 

From what I can tell any command or operation that uses the clone method will cause the image element to fail as the rtol will not be copied

```python
import numpy as np
import xarray as xr
import holoviews as hv

def create_data():
    """
    create fake example data
    """
    times = [np.datetime64('2018-02-22 00:00:00'),
            np.datetime64('2018-02-23 00:00:00')]
    x = np.arange(0, 10, 1.)
    y = np.arange(0, 10, 1.)

    # now add some randomness at 1e-4 to test rtol
    x += np.random.randint(0, 9, 10) * 1e-4
    dims = ('time', 'x', 'y')
    shape = (len(times), len(x), len(y))
    coords = {'x':x, 'y':y, 'time':times}
    field = np.zeros(shape)
    field[0, 3:5, 3:5] = 10
    field[1, 3:5, 3:5] = 20
    field_xr = xr.DataArray(field, dims=dims, coords=coords)
    xr_ds = xr.Dataset({'time': times, 'x': x,
        'y': y,  'field': (dims, field)})
    return xr_ds

hv.extension('bokeh')
renderer = hv.renderer('bokeh')
hv.output('size=200')
hv.Dimension.type_formatters[np.datetime64] = '%Y-%m-%d %H'

xr_ds = create_data()
vdims = ['field']
kdims= ['time', 'x', 'y']
ds = hv.Dataset(xr_ds, kdims=kdims, vdims=vdims)

# actually applying these globally or else we will hit our bug!
hv.opts("Image [colorbar=True] (cmap='viridis')")

# here we can convert to a ds and plot, IF we use rtol
img = ds.to(hv.Image, ['x', 'y'], rtol=1e-2)
renderer.save(img, 'test_rtol')

# now use any function that clones the image, I will use redim because
# thats how I found the problem, but actually img.opts(opts) also causes
# a failure, anything that clones the image will
img = img.redim.range(field=(0, 40))
renderer.save(img, 'test_rtol_redim')
```

PR #2386, opened 2018-03-01
Fix for issue #2380

scaine1:
changed rtol to a param of Image, so that it when the Image is cloned rtol is included in the clone. Also added a unit test to catch the problem described in issue #2380 report

### DIFF
--- a/holoviews/element/raster.py
+++ b/holoviews/element/raster.py
@@ -231,6 +231,11 @@ class Image(Dataset, Raster, SheetCoordinateSystem):
                        bounds=(1, 1), doc="""
         The dimension description of the data held in the matrix.""")
 
+    rtol = param.Number(config.image_rtol, doc="""The tolerance used to enforce 
+            regular sampling for regular, gridded data where regular sampling is 
+            expected. Expressed as the maximal allowable sampling difference 
+            between sample locations.""")
+
     def __init__(self, data, kdims=None, vdims=None, bounds=None, extents=None,
                  xdensity=None, ydensity=None, rtol=None, **params):
         if isinstance(data, Image):
@@ -274,9 +279,10 @@ class Image(Dataset, Raster, SheetCoordinateSystem):
 
         # Ensure coordinates are regularly sampled
 
-        rtol = config.image_rtol if rtol is None else rtol
-        validate_regular_sampling(self, 0, rtol)
-        validate_regular_sampling(self, 1, rtol)
+        self.set_param(rtol=config.image_rtol) if rtol is None else self.set_param(rtol=rtol)
+
+        validate_regular_sampling(self, 0, self.rtol)
+        validate_regular_sampling(self, 1, self.rtol)
 
 
     def __setstate__(self, state):
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,7 +30,7 @@ install:
   - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION flake8 scipy=1.0.0 numpy freetype nose pandas=0.22.0 jupyter ipython=4.2.0 param matplotlib=2.1.2 xarray
   - source activate test-environment
   - conda install -c conda-forge  iris plotly flexx --quiet
-  - conda install -c bokeh datashader dask=0.16.1 bokeh=0.12.14 selenium
+  - conda install -c bokeh datashader dask bokeh=0.12.14 selenium
   - if [[ "$TRAVIS_PYTHON_VERSION" == "3.4" ]]; then
       conda install python=3.4.3;
     fi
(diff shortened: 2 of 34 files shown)


jlstevens:
Thanks for this PR.

I am not entirely convinced that things like``rtol`` should be element parameters: they aren't really 'about' the data - how often do people know about suitable tolerance values for regular sampling? It is an ad hoc validation criterion that most people don't care about. The only time they care about it is when our automatic attempt at validation goes wrong.

So generally I am not in favor of adding such things as element parameters but in *this* particular case I think it is ok. Issue #2380 does need fixing and this is the simplest approach and the default value of rtol does comes from ``hv.config``. 

For these reasons, I think I am ok with this fix though I will be very unhappy if I see our existing elements gaining new parameters after this one!

jlstevens:
Actually, I suppose you can argue that ``rtol`` is part of the semantics of ``Image`` which is designed for regularly sampled data. This means having ``rtol`` parameter always defined for elements that expect regular sampling is reasonable.

So I am happy with this approach here though in general I am very wary about adding parameters to elements!

philippjfr:
I'm also *okay* with this although not entirely happy about it. I've got a minor comment, and this PR will have to be rebased but after those issues are addressed I'll be happy to merge.

holoviews/element/raster.py
```diff
@@ -274,9 +279,10 @@ def __init__(self, data, kdims=None, vdims=None, bounds=None, extents=None,
 
         # Ensure coordinates are regularly sampled
 
-        rtol = config.image_rtol if rtol is None else rtol
-        validate_regular_sampling(self, 0, rtol)
-        validate_regular_sampling(self, 1, rtol)
+        self.set_param(rtol=config.image_rtol) if rtol is None else self.set_param(rtol=rtol)
```
philippjfr: Rather than using ``set_param`` here, the ``rtol`` value should be passed to the Dataset constructor above, something like:

```
rtol = (config.image_rtol if self.rtol is None else self.rtol) if rtol is None else rtol
Dataset.__init__(self, data, kdims=kdims, vdims=vdims, extents=extents, **params)
```

and then change the default ``rtol`` to None.

jlstevens: Agreed.

jbednar:
| I am not entirely convinced that things like rtol should be element parameters: they aren't really 'about' the data - how often do people know about suitable tolerance values for regular sampling?

I would argue that rtol is *entirely* about the data.  It's metadata describing a particular dataset, at least if it's been changed for that data.  It's a declaration of how regularly sampled you expect this data to be, allowing us to warn or error if that declaration turns out to be false.  Seems entirely within the semantics of an Element.

jlstevens:
>  It's a declaration of how regularly sampled you expect this data to be...

It is a declaration that no one ever knows an appropriate value for or ever really things about. That is the problem with it!

If everyone always knew their appropriate rtol, it would be an easy decision. As it is, I do agree it is in the semantics of ``Image`` (but an annoying aspect of it!) and at least if our default is normally good enough, most people won't ever need to think about it, most of the time.

jbednar:
> It is a declaration that no one ever knows an appropriate value for or ever really thinks about. That is the problem with it!

That's a problem with reality, not with this parameter. :-)  Pick a fight with the universe, and you're going to lose!

holoviews/element/raster.py
scaine1: I don't quite understand and I am probably missing something
if `set_param` is not used then `self.rtol = None` always, so in the conditional statement the `config.image_rtol if self.rtol is None`  is always True and we never hit the ` else self.rtol `

scaine1: Further, unless `rtol` is a `param` the clone method in holoviews/core/dimension.py wont copy it and I get the following issue, which is what caused me to make rtol a param in.
```
  File "my_path/holoviews/core/dimension.py", line 572, in clone
    return clone_type(data, *args, **{k:v for k,v in settings.items()
```

philippjfr: >if set_param is not used then self.rtol = None always, so in the conditional statement the config.image_rtol if self.rtol is None is always True and we never hit the else self.rtol

You can override parameters at the class level, e.g. if you do ``hv.Image.rtol = 0.000001`` then ``self.rtol`` will not be None.

>Further, unless rtol is a param the clone method in holoviews/core/dimension.py wont copy it and I get the following issue, which is what caused me to make rtol a param in the first place.

This should not be the case but Image is a bit special in terms of multiple inheritance so I'd have to look more closely.

philippjfr:
Looking at this now.

philippjfr:
Going to close and try to reopen.

PR #2389, opened 2018-03-03
Correctly handle comms in bokeh<0.12.15

philippjfr:
In the JupyterLab PR we switched to handling the comm messages ourselves which does not work in bokeh versions <=0.12.14 however I did not guard against this case appropriately. This now ensures that Bokeh exposes the ``protocol`` module and only creates a comm for bokeh versions < 0.12.15.


philippjfr:
Ready to merge.

jlstevens:
Looks good to me.

PR #2391, opened 2018-03-03
Replaced badge with monthly download badge

philippjfr:
The current conda download badge only takes into account downloads from the ioam channel even though that is not even the most popular channel for installing holoviews (conda-forge is). This replaces the badge with a badge that shows rolling monthly download stats aggregating over conda and PyPI. Reporting 6-7k downloads per month looks a lot better than 20k all time downloads which is just totally wrong.


philippjfr:
Ready to merge.

jlstevens:
Looks good. Merging.

PR #2396, opened 2018-03-04
Add support for Dimension.default

philippjfr:
Implements ``Dimension.default`` to allow initializing a plot to something other than the first key value.

- [x] Addresses https://github.com/ioam/holoviews/issues/704 and https://github.com/ioam/holoviews/issues/1258
- [x] Add new tests
- [x] Add example to docs

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -746,7 +746,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
             element = [el for el in self.hmap.data.values() if el][-1]
         else:
             element = self.hmap.last
-        key = self.keys[-1]
+        key = util.wrap_tuple(self.hmap.last_key)
         ranges = self.compute_ranges(self.hmap, key, ranges)
         self.current_ranges = ranges
         self.current_frame = element
@@ -1390,7 +1390,7 @@ class OverlayPlot(GenericOverlayPlot, LegendPlot):
 
 
     def initialize_plot(self, ranges=None, plot=None, plots=None):
-        key = self.keys[-1]
+        key = util.wrap_tuple(self.hmap.last_key)
         nonempty = [el for el in self.hmap.data.values() if el]
         if not nonempty:
             raise SkipRendering('All Overlays empty, cannot initialize plot.')
--- a/holoviews/plotting/bokeh/widgets.py
+++ b/holoviews/plotting/bokeh/widgets.py
@@ -70,8 +70,6 @@ class BokehServerWidgets(param.Parameterized):
         self.mock_obj = NdMapping([(k, None) for k in self.keys],
                                   kdims=list(self.dimensions))
         self.widgets, self.lookups = self.get_widgets()
-        self.reverse_lookups = {d: {v: k for k, v in item.items()}
-                                for d, item in self.lookups.items()}
         self.subplots = {}
         if self.plot.renderer.mode == 'default':
             self.attach_callbacks()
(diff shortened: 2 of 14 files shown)


philippjfr:
Ready to review.

holoviews/plotting/mpl/plot.py
```diff
@@ -240,7 +240,7 @@ def anim(self, start=0, stop=None, fps=30):
 
 
     def update(self, key):
-        if len(self) == 1 and key == 0 and not self.drawn:
+        if len(self) == 1 and ((key == 0) or (key == self.keys[0])) and not self.drawn:
```
jlstevens: Is the idea that ``self.keys[0]`` here normally matches ``init_key``?

philippjfr: Correct usually plots are initialized to the first frame.

holoviews/plotting/bokeh/element.py
```diff
@@ -746,7 +746,7 @@ def initialize_plot(self, ranges=None, plot=None, plots=None, source=None):
             element = [el for el in self.hmap.data.values() if el][-1]
         else:
             element = self.hmap.last
-        key = self.keys[-1]
+        key = util.wrap_tuple(self.hmap.last_key)
```
jlstevens: Trying to understand the different between these two...

holoviews/plotting/bokeh/widgets.py
```diff
@@ -95,52 +93,80 @@ def create_widget(self, dim, holomap=None, editable=False):
         label, mapping = None, None
         if holomap is None:
             if dim.values:
+                if dim.default is None:
+                    default = dim.values[0]
+                elif dim.default not in dim.values:
```
jlstevens: Isn't this already validated at the dimension level?

holoviews/plotting/bokeh/widgets.py
```diff
@@ -95,52 +93,80 @@ def create_widget(self, dim, holomap=None, editable=False):
...
+                elif isinstance(dim_range, int) or int_type:
+                    step = 1
                 else:
                     step = 10**((round(math.log10(dim_range))-3))
+
+                if dim.default is None:
+                    default = start
+                elif (dim.default < start or dim.default > end):
+                    raise ValueError("%s dimension default %r is not in the provided range: %s"
+                                     % (dim, dim.default, (start, end)))
```
jlstevens: Same question about it not already being validated? Isn't ``start`` and ``end`` defined by the dimension range?

holoviews/plotting/renderer.py
```diff
@@ -185,7 +185,10 @@ def get_plot(self_or_cls, obj, renderer=None):
             plot_opts = self_or_cls.plot_options(obj, self_or_cls.size)
             plot = self_or_cls.plotting_class(obj)(obj, renderer=renderer,
                                                    **plot_opts)
-            plot.update(0)
```
jlstevens: I suppose handling ``default`` means this type of update by integer index is no longer recommended?

holoviews/plotting/bokeh/widgets.py
philippjfr: True it is now

holoviews/plotting/widgets/__init__.py
```diff
@@ -290,95 +293,137 @@ class SelectionWidget(NdWidget):
...
+            widget_data = dict(widget_data, dim=escaped_dim, dim_label=dim_str,
+                               dim_idx=idx, visibility=visibility)
             widgets.append(widget_data)
             dimensions.append(escaped_dim)
         init_dim_vals = escape_list(escape_vals(init_dim_vals, not self.plot.dynamic))
         return widgets, dimensions, init_dim_vals
 
 
+    @classmethod
+    def _get_static_widget(cls, idx, dim, mock_obj, hierarchy, init_dim_vals):
```
jlstevens: Good to see it split up a bit!

holoviews/plotting/widgets/widgets.js
```diff
@@ -390,16 +390,22 @@ function init_slider(id, plot_id, dim, values, next_vals, labels, dynamic, step,
...
     } else {
       var min = 0;
       if (dynamic) {
         var max = Object.keys(vals).length - 1;
+        var init_label = labels[value];
+        var default_value = values[value];
       } else {
         var max = vals.length - 1;
+        var init_label = labels[value];
+		var default_value = value;
```
jlstevens: Very minor point: wonky indentation here.

holoviews/plotting/bokeh/element.py
philippjfr: The former is a global key across a whole layout, the latter the last actually existing key on the subplot. The former would have been causing errors in some cases since you can't initialize on a non-existent key.

holoviews/plotting/bokeh/widgets.py
philippjfr: Yes, at least the dynamic cases are validated now.

holoviews/plotting/renderer.py
philippjfr: A user can do what they like I guess, but generally we've internally been moving away from them.

holoviews/plotting/widgets/widgets.js
philippjfr: Thanks.

holoviews/plotting/mpl/mplwidgets.js
```diff
@@ -19,7 +19,11 @@ var MPLMethods = {
     } else {
       this.update_cache();
     }
-    this.update(0);
+    if (this.dynamic | !this.cached) {
+      this.update(0)
```
jlstevens: So does ``update(0)`` refer to the initial key that includes defaults?

philippjfr: In the dynamic case that should be the case yes, in the embedded case below it's not, hence adding this branch.

jlstevens:
Ok. Very happy to finally see this feature though unfortunately it does looks like it was a pain to do! :-(

Other than this oft-requested functionality, I think this PR is good for a few other reasons:

* Some of the logic has been split up a bit better.
* Getting rid of all the hard coded ``0`` indices paves the way to a more powerful and general widget system. That said, the widget refactor looks like it will be a nightmare!

I'm hoping this is the last widget feature we add before the eventual refactor - the more features we add to this code, the harder it will be to improve.

Anyway, other than some questions and comments above, I am happy to merge and start testing it out! One task I can assign myself (doesn't need to be in this PR) is to set defaults in some of the dynamicmap user guides as there are places where the default would show what is going on better before touching the widgets.

holoviews/plotting/bokeh/element.py
jlstevens: Thanks for clarifying...

philippjfr:
>That said, the widget refactor looks like it will be a nightmare!

I'm getting more and more confident that we can move to a better widget system gradually. The next steps from here should be:

* Instead of templating the JS pack the widget data into a JSON blob with a well defined API from which widgets can be instantiated
* Deprecate old functionality to deal with hierarchical indexes to make the widgets truly independent
* Separate the widget manager from the specific widget implementation
* Rewrite a cleaner implementation of the widget manager with well thought out API for the widgets to communicate with (probably in ES6)  
* Factor out parts of bokeh callback machinery that deal with syncing bokeh models (to be reused by a bokeh widget implementation)
* Build new widget implementation based on streams and callbacks

philippjfr:
Okay made the suggested changes, ready to merge when tests pass.

jlstevens:
> * Instead of templating the widget JS code, pack the widget data into a JSON blob with a well defined API from which widgets can be instantiated, this blob can either be used by a JS widget or by a Python widget (e.g. bokeh)

Good idea. The goal would be to have widgets that can update their state dynamically from the JSON.

> * Deprecate old functionality to deal with hierarchical indexes to make the widgets truly independent

Sounds good though won't there be a corresponding loss of functionality?

> * Separate the widget manager from the specific widget implementation
> * Rewrite a cleaner implementation of the widget manager with well thought out API for the widgets to communicate with (probably in ES6)

Agreed though using ES6 means we will need to use babel to generate ES5 at some point...

> * Factor out parts of bokeh callback machinery that deal with syncing bokeh models (to be reused by a bokeh widget implementation)

I saw such a callback recently for widgets that apply the bokeh JSON patches. It looked almost identical to the normal callback so it shouldn't be hard to update.

> * Build new widget implementation based on streams and callbacks

That is the hardest step! :-)

Sounds like a plan and I agree that getting rid of templating is probably the first and most important step that can be hopefully be done in the background without changing how things fundamentally work (for the time being).

philippjfr:
>That is the hardest step! :-)

I think it's the simplest actually, I hacked a version of this together in an hour. Once the other parts are done it won't need all the hacks.

jlstevens:
By a 'new widget implementation' I assumed that means swappable widget libraries (not just JQueryUI) new architecture and API to work with the widget manager etc. The streams and callbacks part isn't the hard bit, I agree.

jlstevens:
Anyway, the tests are green now and I'm very happy to see this functionality finally make it into master. Merging.

PR #2397, opened 2018-03-04
Move shared plot parameters to base classes

philippjfr:
As the title says, where ever possibly this moves plot parameters to the baseclasses.

- [x] Addresses https://github.com/ioam/holoviews/issues/2369


philippjfr:

jlstevens:
Great! This helps make sure the API is nice and consistent. We should always aim to move parameters to the base classes when they are added instead of afterwards.

Merging.

PR #2404, opened 2018-03-06
Miscellaneous plot fixes

philippjfr:
Smaller bug fixes along with tests to avoid regressions.

- [x] Fixes https://github.com/ioam/holoviews/issues/1757, https://github.com/ioam/holoviews/issues/1921 and https://github.com/ioam/holoviews/issues/2347

### DIFF
--- a/holoviews/plotting/bokeh/annotation.py
+++ b/holoviews/plotting/bokeh/annotation.py
@@ -38,6 +38,8 @@ class TextPlot(ElementPlot):
         style['text_align'] = element.halign
         baseline = 'middle' if element.valign == 'center' else element.valign
         style['text_baseline'] = baseline
+        if 'text_font_size' not in style:
+            style['text_font_size'] = '%dPt' % element.fontsize
         if 'color' in style:
             style['text_color'] = style.pop('color')
         return (data, mapping, style)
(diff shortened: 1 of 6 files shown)


holoviews/plotting/bokeh/annotation.py
```diff
@@ -35,9 +35,13 @@ def get_data(self, element, ranges, style):
...
         data['text'] = [element.text]
-        style['text_align'] = element.halign
+        if 'text_align' not in style:
+            style['text_align'] = element.halign
         baseline = 'middle' if element.valign == 'center' else element.valign
-        style['text_baseline'] = baseline
+        if 'text_baseline' not in style:
+            style['text_baseline'] = baseline
+        if 'text_font_size' not in style:
+            style['text_font_size'] = '%dPt' % element.fontsize
```
jlstevens: One day we can try to allow the user to specify font sizes in something other than points. For now this is fine though...

jlstevens:
Looks good. Happy to merge when the tests are green.

holoviews/plotting/bokeh/annotation.py
philippjfr: Right, this is the semantics that matplotlib uses so I stuck with Pts. If you set the style option you can use a different unit.

jlstevens: To be honest using points as a default is probably fine and not worth changing (or even exposing). If anything the problem is documenting that the default is in points.

jlstevens:
The pr build is now passing. Merging.

PR #2405, opened 2018-03-06
Fixed bug setting bokeh toolbar location

philippjfr:
The LayoutPlot and GridPlot were not using the correct keyword to set the toolbar location on a gridplot.

- [x] Fixes https://github.com/ioam/holoviews/issues/2370
- [x] Adds tests

### DIFF
--- a/tests/plotting/bokeh/testgridplot.py
+++ b/tests/plotting/bokeh/testgridplot.py
@@ -7,7 +7,8 @@ from holoviews.operation import gridmatrix
 from .testplot import TestBokehPlot, bokeh_renderer
 
 try:
-    from bokeh.models import Div
+    from bokeh.layouts import Column
+    from bokeh.models import Div, ToolbarBox
 except:
     pass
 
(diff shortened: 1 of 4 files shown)


tests/plotting/bokeh/testgridplot.py
```diff
@@ -7,7 +7,8 @@
 from .testplot import TestBokehPlot, bokeh_renderer
 
 try:
-    from bokeh.models import Div
+    from bokeh.layouts import Column, Row
+    from bokeh.models import Div, ToolbarBox
```
jlstevens: ``Row` is imported here but I don't see it used? Won't that be a flake?

tests/plotting/bokeh/testlayoutplot.py
```diff
@@ -202,3 +202,15 @@ def test_layout_empty_subplots(self):
...
+    def test_layout_set_toolbar_location(self):
+        layout = (Curve([]) + Points([])).options(toolbar='left')
+        plot = bokeh_renderer.get_plot(layout)
+        self.assertIsInstance(plot.state, Column)
+        self.assertIsInstance(plot.state.children[0], ToolbarBox)
+
+    def test_layout_disable_toolbar(self):
+        layout = (Curve([]) + Points([])).options(toolbar=None)
+        plot = bokeh_renderer.get_plot(layout)
+        self.assertIsInstance(plot.state, Row)
```
jlstevens: ``Row`` is used here but this is a different file...

tests/plotting/bokeh/testgridplot.py
philippjfr: Yes, good catch.

jlstevens:
Looks good other than what looks like an unused import. Happy to merge when the tests are green.

philippjfr:
Requires changes in the test data so should be merged after https://github.com/ioam/holoviews/pull/2404.

PR #2418, opened 2018-03-07
Bokeh renderer fixes

philippjfr:
Fixes two bugs introduced in the JupyterLab refactor:

1) Fixes png export by correctly handling it in ``_figure_data`` (where it should be handled)
2) Fixes wrong logic in bokeh message handler, which incorrectly skipped messages when Bokeh.protocol was defined.


jlstevens:
Thanks for the message handler fix and the PNG changes seem fine to me. Merging.

PR #2424, opened 2018-03-09
Ensure select on dask arrays in GridInterface is lazy

philippjfr:
Followup from https://github.com/ioam/holoviews/pull/2305 which ensures that select and sample operations are applied lazily and do not load the dask array into memory.

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -621,7 +621,7 @@ class Dataset(Element):
         a new_type, which will inherit all shared parameters.
         """
         if 'datatype' not in overrides:
-            datatypes = [self.interface.datatype]+self.datatype
+            datatypes = [self.interface.datatype] + self.datatype
             overrides['datatype'] = list(unique_iterator(datatypes))
         return super(Dataset, self).clone(data, shared_data, new_type, *args, **overrides)
 
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -9,7 +9,7 @@ import numpy as np
 import param
 
 from ..dimension import redim
-from ..util import dimension_range
+from ..util import dimension_range, unique_iterator
 from .interface import Interface, iloc, ndloc
 from .array import ArrayInterface
 from .dictionary import DictInterface
@@ -609,6 +609,21 @@ class Dataset(Element):
         convert to other Element types.
         """
         return self._conversion_interface(self)
+
+
+    def clone(self, data=None, shared_data=True, new_type=None, *args, **overrides):
+        """
+        Returns a clone of the object with matching parameter values
+        containing the specified args and kwargs.
+
+        If shared_data is set to True and no data explicitly supplied,
+        the clone will share data with the original. May also supply
+        a new_type, which will inherit all shared parameters.
+        """
+        if 'datatype' not in overrides:
+            datatypes = [self.interface.datatype]+self.datatype
+            overrides['datatype'] = list(unique_iterator(datatypes))
+        return super(Dataset, self).clone(data, shared_data, new_type, *args, **overrides)
 
 
     @property
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -243,7 +243,7 @@ class GridInterface(DictInterface):
         dropped = [dims.index(d) for d in dims
                    if d not in dataset.kdims+virtual_coords]
         if dropped:
-            data = data.squeeze(axis=tuple(dropped))
+            data = np.squeeze(data, axis=tuple(dropped))
 
         if not any(cls.irregular(dataset, d) for d in dataset.kdims):
             inds = [dims.index(kd.name) for kd in dataset.kdims]
@@ -354,7 +354,8 @@ class GridInterface(DictInterface):
                 group_data = group_data if np.isscalar(group_data) else group_data.columns()
             else:
                 group_data = cls.select(dataset, **select)
-            if np.isscalar(group_data):
+
+            if np.isscalar(group_data) or (isinstance(group_data, array_types) and da.group_data.shape == ()):
                 group_data = {dataset.vdims[0].name: np.atleast_1d(group_data)}
                 for dim, v in zip(dim_names, unique_key):
                     group_data[dim] = np.atleast_1d(v)
@@ -460,15 +461,25 @@ class GridInterface(DictInterface):
         int_inds = [np.argwhere(v) for v in value_select][::-1]
         index = np.ix_(*[np.atleast_1d(np.squeeze(ind)) if ind.ndim > 1 else np.atleast_1d(ind)
                          for ind in int_inds])
+
         for kdim in dataset.kdims:
             if cls.irregular(dataset, dim):
-                data[kdim.name] = np.asarray(data[kdim.name])[index]
+                if da and isinstance(dataset.data[kdim.name], da.Array):
+                    data[kdim.name] = dataset.data[kdim.name].vindex[index]
+                else:
+                    data[kdim.name] = np.asarray(data[kdim.name])[index]
+
         for vdim in dataset.vdims:
-            data[vdim.name] = np.asarray(dataset.data[vdim.name])[index]
+            if da and isinstance(dataset.data[vdim.name], da.Array):
+                data[vdim.name] = dataset.data[vdim.name].vindex[index]
+            else:
+                data[vdim.name] = np.asarray(dataset.data[vdim.name])[index]
 
         if indexed:
             if len(dataset.vdims) == 1:
                 arr = np.squeeze(data[dataset.vdims[0].name])
+                if da and isinstance(arr, da.Array):
+                    arr = arr.compute()
                 return arr if np.isscalar(arr) else arr[()]
             else:
                 return np.array([np.squeeze(data[vd.name])
@@ -503,7 +514,7 @@ class GridInterface(DictInterface):
             for vdim, array in zip(dataset.vdims, arrays):
                 flat_index = np.ravel_multi_index(tuple(int_inds)[::-1], array.shape)
                 if da and isinstance(array, da.Array):
-                    data[vdim.name].append(array.flatten()[tuple(flat_index)])
+                    data[vdim.name].append(array.flatten().vindex[tuple(flat_index)])
                 else:
                     data[vdim.name].append(array.flat[flat_index])
         concatenated = {d: np.concatenate(arrays).flatten() for d, arrays in data.items()}
@@ -547,7 +558,7 @@ class GridInterface(DictInterface):
                 if len(axes) > 1:
                     vdata = vdata.transpose(axes[::-1])
                 if dropped_axes:
-                    vdata = vdata.squeeze(axis=dropped_axes)
+                    vdata = np.squeeze(vdata, axis=dropped_axes)
                 data[vdim.name] = vdata
             return data
         elif dropped_kdims:
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -616,7 +616,7 @@ class GridInterface(DictInterface):
             column = cls.values(dataset, dimension, flat=False)
         if column.dtype.kind == 'M':
             dmin, dmax = column.min(), column.max()
-            if isinstance(column, da.Array):
+            if da and isinstance(column, da.Array):
                 return dmin.compute(), dmax.compute()
             return dmin, dmax
         elif len(column) == 0:
@@ -624,7 +624,7 @@ class GridInterface(DictInterface):
         else:
             try:
                 dmin, dmax = (np.nanmin(column), np.nanmax(column))
-                if isinstance(column, da.Array):
+                if da and isinstance(column, da.Array):
                     return dmin.compute(), dmax.compute()
                 return dmin, dmax
             except TypeError:
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -1,3 +1,5 @@
+from __future__ import absolute_import
+
 from collections import OrderedDict, defaultdict, Iterable
 
 try:
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -619,7 +619,7 @@ class GridInterface(DictInterface):
         if column.dtype.kind == 'M':
             dmin, dmax = column.min(), column.max()
             if da and isinstance(column, da.Array):
-                return dmin.compute(), dmax.compute()
+                return da.compute(dmin, dmax)
             return dmin, dmax
         elif len(column) == 0:
             return np.NaN, np.NaN
@@ -627,7 +627,7 @@ class GridInterface(DictInterface):
             try:
                 dmin, dmax = (np.nanmin(column), np.nanmax(column))
                 if da and isinstance(column, da.Array):
-                    return dmin.compute(), dmax.compute()
+                    return da.compute(dmin, dmax)
                 return dmin, dmax
             except TypeError:
                 column.sort()
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -297,7 +297,7 @@ class GridInterface(DictInterface):
         for d in dataset.dimensions():
             if d in dataset.kdims and not cls.irregular(dataset, d):
                 continue
-            arr = dataset.dimension_values(d, flat=False)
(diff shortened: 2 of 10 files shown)


holoviews/core/data/grid.py
```diff
@@ -243,7 +243,7 @@ def canonicalize(cls, dataset, data, data_coords=None, virtual_coords=[]):
         dropped = [dims.index(d) for d in dims
                    if d not in dataset.kdims+virtual_coords]
         if dropped:
-            data = data.squeeze(axis=tuple(dropped))
+            data = np.squeeze(data, axis=tuple(dropped))
```
jakirkham: So calling `np.squeeze` will probably make this non-lazy. Not sure if that matters in this case. Just a friendly note.

holoviews/core/data/grid.py
```diff
@@ -354,7 +354,8 @@ def groupby(cls, dataset, dim_names, container_type, group_type, **kwargs):
                 group_data = group_data if np.isscalar(group_data) else group_data.columns()
             else:
                 group_data = cls.select(dataset, **select)
-            if np.isscalar(group_data):
+
+            if np.isscalar(group_data) or (len(dimensions) == dataset.ndims and len(dataset.vdims) == 1):
```
jakirkham: FYI `np.isscalar` is always `False` on Dask Arrays. Not sure if these are Dask Arrays though.

holoviews/core/data/grid.py
```diff
@@ -503,7 +514,7 @@ def sample(cls, dataset, samples=[]):
             for vdim, array in zip(dataset.vdims, arrays):
                 flat_index = np.ravel_multi_index(tuple(int_inds)[::-1], array.shape)
                 if da and isinstance(array, da.Array):
-                    data[vdim.name].append(array.flatten()[tuple(flat_index)])
+                    data[vdim.name].append(array.vindex[tuple(flat_index)])
```
jakirkham: FWIW `flatten` is lazy and returns a Dask Array. So you could do both if needed.

philippjfr: >FWIW flatten is lazy and returns a Dask Array. So you could do both if needed.

Thanks, I think that might be needed here. I'm wondering why the tests didn't error though, going to have to make sure the ``sample`` tests are actually working.

holoviews/core/data/grid.py
jakirkham: NVM. Turns out I'm wrong. `np.squeeze` works just fine on Dask Arrays. Sorry for the noise.

holoviews/core/data/grid.py
philippjfr: Right, that's the main thing I struggled with, I think the ``or`` branch here should handle that.

holoviews/core/data/grid.py
philippjfr: It seems to work okay in my simple test case at least, at least it's still a dask array after ``np.squeeze``. Could it still be non-lazy? Here's an example:

```python
darr = da.from_array(np.random.rand(30, 20, 20), (3, 2, 2))
ds = hv.Dataset((range(20), range(20), range(30), darr),
                ['x', 'y', 'z'], 'Value')
z_stack = ds.to(hv.Image, ['x', 'y'])
z_stack.last.data
```

```
{'Value': dask.array<squeeze, shape=(20, 20), dtype=float64, chunksize=(20, 20)>,
 'x': array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15, 16,
        17, 18, 19]),
 'y': array([ 0,  1,  2,  3,  4,  5,  6,  7,  8,  9, 10, 11, 12, 13, 14, 15, 16,
        17, 18, 19]),
 'z': array([29])}
```

philippjfr: Great, thanks! I noticed that since dask 1.7.0 you can use ``np.squeeze`` with an ``axis`` argument but ``da.Array.squeeze`` does not accept ``axis``, is this an oversight?

holoviews/core/data/grid.py
jakirkham: So testing `group_data.shape == ()` should be valid for both.

Admittedly NumPy has some weird scalar like entities (e.g. 0-D Arrays), which would pass this test and fail `np.isscalar`. Though would suspect that distinction is irrelevant here. Would be interesting to know if that assumption is wrong.

tests/core/data/testdataset.py
```diff
@@ -1551,8 +1557,8 @@ def test_dataset_2D_reduce_hm(self):
         array = da.from_array(np.random.rand(11, 11), 3)
         dataset = Dataset({'x':self.xs, 'y':self.y_ints, 'z': array},
                           kdims=['x', 'y'], vdims=['z'])
-        self.assertEqual(np.array(dataset.reduce(['x', 'y'], np.mean)),
-                         np.mean(array))
+        self.assertEqual((dataset.reduce(['x', 'y'], np.mean)),
+                         np.mean(array).compute())
```
jakirkham: FWIW we have utility functions in Dask like [`assert_eq`]( https://github.com/dask/dask/blob/83c3129ce24f71e13e97219fb5c4da1b7d774668/dask/array/utils.py#L73 ). They aren't public technically (would like to change that personally). However they do a lot of things right, which might not be obvious initially.

holoviews/core/data/grid.py
jakirkham: Thanks for pointing that out. Agree it is probably an oversight. Raised as issue ( https://github.com/dask/dask/issues/3259 ). Should be trivial to fix.

holoviews/core/data/grid.py
philippjfr: Thanks, I changed it. I also don't think that 0-D arrays would be relevant here, groupby should only produce scalars or ndarrays.

philippjfr:
Okay, I've now gone through the remaining methods and all the non-lazy operations are now intentional, including:

* scalar indexing: For consistency when you index a specific value using ``.select`` or ``__getitem__`` it will call compute
* ``dimension_values``: This is the main way that plotting code accesses the data and therefore needs to be non-lazy, the underlying interface ``values`` method now has a ``compute`` keyword however, so that internal operations can access the reshaped and correctly oriented dask arrays
* ``.range``: Computes the range out-of-core but does call compute

Lazy operations:

* ``.iloc``: Indexing into the flattened and normalized representation of the data
* ``.ndloc``: Integer indexing into the normalized dask arrays
* ``.groupby``: Splits dataset into chunks along one or more axes
* ``.sample``: Samples the array using vectorized, fancy indexing
* ``.select``: Labeled indexing by value
* ``.aggregate/.reduce``: Allows computing statistics along one or more axes

Another major benefit of this PR is that it finally handles datatypes a bit better, in the past any operation that triggered a ``clone`` (which is pretty much every operation) had the potential side-effect of changing the type of the data. This behavior was very problematic in this case because it had the additional side-effect of converting dask arrays to numpy arrays in many cases. It will now always start by trying to maintain the current interface and only fall back to another interface if that fails.

philippjfr:
Ready to review/merge now.

jakirkham:
With the scalar indexing, are we selecting and then computing or vice versa?

tests/core/data/testdataset.py
philippjfr: Thanks that's useful to know. I don't think they are needed in this case but I can think of other places where that would come in handy.

holoviews/core/data/__init__.py
```diff
@@ -611,6 +611,21 @@ def to(self):
...
+        """
+        Returns a clone of the object with matching parameter values
+        containing the specified args and kwargs.
+
+        If shared_data is set to True and no data explicitly supplied,
+        the clone will share data with the original. May also supply
+        a new_type, which will inherit all shared parameters.
+        """
+        if 'datatype' not in overrides:
+            datatypes = [self.interface.datatype]+self.datatype
```
jakirkham: nit: Would add spaces around the `+`.

holoviews/core/data/grid.py
```diff
@@ -600,14 +615,20 @@ def range(cls, dataset, dimension):
             expanded = cls.irregular(dataset, dimension)
             column = cls.coords(dataset, dimension, expanded=expanded, edges=True)
         else:
-            column = dataset.dimension_values(dimension)
+            column = cls.values(dataset, dimension, flat=False)
         if column.dtype.kind == 'M':
-            return column.min(), column.max()
+            dmin, dmax = column.min(), column.max()
+            if da and isinstance(column, da.Array):
+                return dmin.compute(), dmax.compute()
```
jakirkham: Would be better to do `da.compute(dmin, dmax)`. That reuses any common intermediate results for both computations. Also only optimizes the underlying Dask graph once.

holoviews/core/data/grid.py
```diff
@@ -600,14 +615,20 @@ def range(cls, dataset, dimension):
...
+            return dmin, dmax
         elif len(column) == 0:
             return np.NaN, np.NaN
         else:
             try:
-                return (np.nanmin(column), np.nanmax(column))
+                dmin, dmax = (np.nanmin(column), np.nanmax(column))
```
jakirkham: This should be lazy as long as the user has NumPy 1.13.0+ and Dask 0.15.0+. If we use the associated Dask Array functions, we can relax the version constraints needed to make this lazy.

holoviews/core/data/grid.py
```diff
@@ -600,14 +615,20 @@ def range(cls, dataset, dimension):
...
+                if da and isinstance(column, da.Array):
+                    return dmin.compute(), dmax.compute()
```
jakirkham: Would use `da.compute(dmin, dmax)` here as well. Reasoning is the same as above.

holoviews/element/chart.py
```diff
@@ -317,7 +317,7 @@ def stack(cls, areas):
             x, y = (area.dimension_values(i) for i in range(2))
             stacked[k] = area.clone((x, y+baseline, baseline), vdims=vdims,
                                     new_type=Area)
-            baseline += y
+            baseline = baseline+y
```
jakirkham: nit: Would add spaces around the `+`.

jakirkham: That said, would think `+=` works ok here. Was there an issue with it?

holoviews/core/data/grid.py
philippjfr: Thanks, I keep forgetting about this!

holoviews/core/data/grid.py
philippjfr: Okay, I'll use those, although I think we rely on dask>=1.7.0 anyway due to the np.squeeze ``axis`` argument, which is used in a few places.

holoviews/element/chart.py
philippjfr: The array ended up being shared between multiple stacked Area elements, which means they all incorrectly shared the same baseline. This was previously covered up by the fact that certain interfaces would copy the array anyway.

jakirkham:
Looks pretty good.

Made a few style notes that most linters would complain about. Then again this may be consistent with the code style (as I saw in one diff). So if the general code style matches, would ignore these.

Separately saw a `a += b` converted to `a = a + b`. Was trying to understand the motivation for this. Also issue ( https://github.com/dask/dask/issues/2588 ) may be relevant. Please see [this comment]( https://github.com/dask/dask/issues/2588#issuecomment-321079105 ) in particular.

When it comes to calling `compute`, it is best to pass as much stuff being computed into one call as possible. This is important as Dask does a graph optimization pass before computing, which can then reuse optimizations on the shared graph. Also this is important as the scheduler will take into account shared intermediate computations.

philippjfr:
Thanks for the review, it's really appreciated. We don't enforce rigorous linting atm, but it's still the preferred style so I'll change it.

>Separately saw a a += b converted to a = a + b. Was trying to understand the motivation for this. Also issue ( dask/dask#2588 ) may be relevant. Please see this comment in particular.

I replied to that, it's not related to dask, it's just a bug that was uncovered when I changed how the interfaces are prioritized.

philippjfr:
Would be good to get this into the alpha release today, so @jlstevens let's merge once the tests pass. Since we don't require dask directly I'm not sure how to express the requirement for dask>=1.7.0 yet, but we don't need to solve that right away.

jakirkham:
Guessing you mean `dask>=0.7.0`? ;) Haven't hit 1.0.0 yet, but it is being discussed.

holoviews/element/chart.py
jakirkham: Interesting. Guessing there is a test for this?

philippjfr:
Ah 😄 Guess I was confused by this http://dask.pydata.org/en/latest/array-api.html#dask.array.squeeze, which mentions 1.7.0, any idea what that is actually referring to?

jakirkham:
> Ah 😄 Guess I was confused by this http://dask.pydata.org/en/latest/array-api.html#dask.array.squeeze, which mentions 1.7.0, any idea what that is actually referring to?

Oops, that's on us then. :) It's copying the docstring from NumPy. Definitely a downside to doing this.

That said, this code is using `np.squeeze` instead of `da.squeeze` (which worked lazily). This probably depends on NumPy, but am not sure how that behavior has existed. @shoyer might have some idea (as he may have made this change if it wasn't already there).

cc @mrocklin @jcrist

holoviews/core/data/grid.py
jakirkham: Just to clarify, the reason relying on NumPy ufuncs requires both NumPy and Dask version constraints is because we need the `__array_ufunc__` attribute that NumPy 1.13 added and the Dask 0.15.0 version that defines `__array_ufunc__ ` for Dask Arrays. If we use the Dask ufuncs, this doesn't occur.

philippjfr:
@jlstevens I think we can merge this now, I'll follow up on the dask version issue in a subsequent PR.

shoyer:
> This probably depends on NumPy, but am not sure how that behavior has existed. @shoyer might have some idea (as he may have made this change if it wasn't already there).

Sorry, which behavior specifically are you referring to here? (It's not quite clear to me from the immediate context)

philippjfr:
>Sorry, which behavior specifically are you referring to here? (It's not quite clear to me from the immediate context)

@shoyer Thanks for responding so quickly, the question was about using ``np.squeeze`` on dask Arrays and I guess there are two questions:

 1) Since when is the axis argument supported?
 2) Since when can you use numpy functions on dask arrays lazily?

shoyer:
> Since when is the axis argument supported?

@jakirkham is only adding that now -- see https://github.com/dask/dask/pull/3261

> Since when can you use numpy functions on dask arrays lazily?

For quite some time for most functions that support it (usually by calling methods of the same name on the underlying array object). In the case of squeeze, Git shows it hasn't changed in 7 years:
https://github.com/numpy/numpy/blob/bd882cb6486f3d9dd3141dd58f0fb2e3a6c7cb3c/numpy/core/fromnumeric.py#L1275-L1284

philippjfr:
Thanks again, my confusion stemmed from the dask docs page: http://dask.pydata.org/en/latest/array-api.html#dask.array.squeeze, which made it appear like the support for the ``axis`` argument was only added to dask recently, but it's clear to me now that it's referring to NumPy.

philippjfr:
Okay, by the sounds of it there isn't too much to worry about in terms of dask version requirements. Thanks again @jakirkham and @shoyer. @jlstevens Go ahead and merge once the py3 build passes.

jlstevens:
Tests have passed. Merging.

jakirkham:
Seems that NumPy code uses a [C-bound Python method]( https://github.com/numpy/numpy/blob/e06d3614182e7b97d5e0d90291642027d147744b/numpy/core/src/multiarray/methods.c#L2684-L2686 ), which [dispatches]( https://github.com/numpy/numpy/blame/e06d3614182e7b97d5e0d90291642027d147744b/numpy/core/src/multiarray/methods.c#L215-L238 ) to a [C function]( https://github.com/numpy/numpy/blame/c90d7c94fd2077d0beca48fa89a423da2b0bb663/numpy/core/src/multiarray/shape.c#L572-L626 ). All of which seems basically static for the past 7yrs (excepting occasional message tweaks for errors and the like). Interesting that the C function works lazily on Dask Arrays in the first place. 🤔

jakirkham:
Raised issue ( https://github.com/dask/dask/issues/3265 ) to improve docstrings, @philippjfr.

jakirkham:
Thanks all. :)

PR #2425, opened 2018-03-09
Made pandas autoindexing behavior backwards compatible

philippjfr:
Recent changes to the way pandas indexes are handled caused a regression in the behavior of the interface when passing in a single column dataframe. This restores the old behavior while still allowing the use of the index on a single column dataframe.

- [x] Fixes https://github.com/ioam/holoviews/issues/2400

### DIFF
--- a/holoviews/core/data/pandas.py
+++ b/holoviews/core/data/pandas.py
@@ -38,8 +38,13 @@ class PandasInterface(Interface):
         if util.is_series(data):
             data = data.to_frame()
         if util.is_dataframe(data):
-            if eltype._auto_indexable_1d and len(data.columns) == 1:
-                data = data.reset_index()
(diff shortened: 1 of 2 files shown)


philippjfr:

jlstevens:
Looks good to me and I like that this change manages to avoid pandas version checks.

philippjfr:
> I like that this change manages to avoid pandas version checks.

It's not pandas' behavior that changed, it's ours since we now support using the pandas index in certain cases.

jlstevens:
> It's not pandas' behavior that changed, it's ours since we now support using the pandas index in certain cases.

Ah. I misread your description and thanks for clarifying. Anyway, I'll merge when the tests are green.

PR #2426, opened 2018-03-09
Switch versioning system to use autover

jlstevens:
Currently the PR is relying on ``embed_version`` during the build step and the JSON fallback in ``__init__.py`` which should be sufficient for packaging an alpha release.

This should also be sufficient once a param release is available includes ``Version``. Until then, the version won't be up-to-date when working from a git repository which means this PR shouldn't be merged until either 1) a new param is released and pinned 2) we decide to temporarily bundle autover  instead.


jlstevens:
I believe @philippjfr would like to see this merged sooner, rather than later...even if we will delete ``version.py`` when ``Version`` is available via param.

Things to do in future PRs:

- [ ] Remove ``version.py`` and use param instead, updating ``setup.py`` accordingly.
- [ ] Pin to the upcoming param release.
- [ ] Set up building on travis when tags get pushed.

I think this is ready to review/test/merge once travis goes green.

philippjfr:
Looks good, very happy to finally have sane version numbers.

PR #2428, opened 2018-03-09
Fixed empty Element visibility bug in bokeh plots

philippjfr:
When using the drawing tools you often initialize the plot with an empty element, but currently empty elements make the glyph renderer invisible. This ensures that only missing elements set visible=False on the renderer.


jlstevens:
Fix looks correct. Merging.

Issue #2319, opened 2018-02-08
Improved xarray support


drs251:
I noticed some points concerning xarray support in holoviews which could be improved:

- [x] Both xarray and holoviews support measurements units, but it appears that they are lost during in- and exporting. I think it would be pretty straightforward to add this feature in the holoviews importing and exporting code, I could look into it.

- [x] When importing a netcdf file via xarray and passing it to ``hv.Dataset``, the key dimensions are sometimes ordered differently, depending whether the file is imported as a ``xr.DataArray``or ``xr.Dataset``. This is not a gigantic problem, especially since one can manually specify the key dimensions' order, although it is still pretty annoying when dealing with images...
EDIT: (actually, this seems to have been fixed already sometime between 1.9.3 and the current master)

- [x] I've never had any luck converting the data storage backend from 'xarray' to anything else, writing ``img.clone(datatype=['image'])``, as shown in the gridded datasets guide: http://holoviews.org/user_guide/Gridded_Datasets.html Perhaps it would be nice to show this more explicitly in the documentation, or fix it if it is a bug.

EDIT:
- [x] I've noticed that dimensions without coordinates are not accepted by `XArrayInterface` - although omitting coordinates seems to be the recommended way in xarray to deal with dimensions without meaningful coordinates (such as pixels in an image). In this case, it should be possible to just automatically assign simple range coordinates to these dimensions when making a holoviews Dataset.

- [ ] The treatment of non-dimension coordinates could be more concise. Take for example the "rasm" example that comes with xarray, which looks like this
```
Coordinates:
  * time     (time) datetime64[ns] 1980-09-16T12:00:00 1980-10-17 ...
    xc       (y, x) float64 189.2 189.4 189.6 189.7 189.9 190.1 190.2 190.4 ...
    yc       (y, x) float64 16.53 16.78 17.02 17.27 17.51 17.76 18.0 18.25 ...
Dimensions without coordinates: x, y
Data variables:
    Tair     (time, y, x) float64 nan nan nan nan nan nan nan nan nan nan ...
```
Right now, coordinates are converted into kdims, and dimensions without coordinates are ignored. IMO, it would be better to let the user select if they want (time, y, c) or (time, yc, xc) coordinates when making the holoviews Dataset, where the default should be the proper dimensions.

[...]

PR #2431, opened 2018-03-11
XArrayInterface improvements: dimension units and labels

drs251:
This addresses the first two points in #2319. As @philippjfr suggested, it's relatively straightforward to read units and dimension labels from xarray and attach them to the appropriate key and value dimensions (as far as I can tell, the attributes `units` and `long_name` are a standard for netcdf files). I could not reproduce the the problem with the swapped kdims for xarray.DataArray vs. xarrayDataset, but I added a test to guard against it.

I would like to try to add some stuff to this PR with your input (see #2319):
- ~~Support for dimensions without coordinates~~
- ~~Perhaps a better way to deal with non-coordinate dimensions~~
- [x] When everything else is done: some extra documentation.


philippjfr:
Looks good so far and the other suggestions sound good too although I suspect it might get tricky.

drs251:
I'm tempted to agree; the first two points should probably be implemented together, and the second one could be a real head scratcher. I'll try to put some thought into it, but if it's too ambitious, I'll just add the documentation I had in mind and leave the issue open for another time. Anyway, I would like to get this in the next release, if possible. Any ideas when that will be?

drs251:
OK, then I'll probably concentrate on the docs for now.

drs251:
@philippjfr OK, I added a simple example of how visualize an xarray Dataset. I also found that something similar already exists for GeoViews, so I added some links. I don't think I'll have time right now to work on the other ideas I have for XArrayInterface, so I would consider this PR to be finished, but I'll leave issue #2319 open and hopefully I'll manage to come back to it sometime.

philippjfr:
At a quick glance this looks great! Would you mind clearing the notebook and committing that. We generally don't store notebooks with output and it will make it easier to review.

examples/user_guide/08-Gridded_Datasets.ipynb
```diff
@@ -292,6 +292,69 @@
...
+    "%%opts Image [colorbar=True]\n",
+    "%%output size=200\n",
+    "hv_ds.to(hv.Image, kdims=[\"lon\", \"lat\"])[:, 220:320, :]"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note, that this approach immediately converts all available data to images, which will take up a lot of RAM for large datasets. For these situations, consider using a [DynamicMap](./06-Live_Data.ipynb) in conjunction with [xarray's dask support](http://xarray.pydata.org/en/stable/dask.html) instead.\n",
```
philippjfr: Maybe mention the ``dynamic`` argument to ``.to`` here, it will generate a DynamicMap for you.

philippjfr:
One minor comment, otherwise looks good to merge.

philippjfr:
You mean changing a label after it has already been set? It was a deliberate decision not to allow that although @jlstevens was more insistent on that than I was. You can override the existing label like this though:

```python
xr_ds = xr.tutorial.load_dataset("air_temperature")
hv.Dataset(xr_ds, vdims=('air', 'air_temp')).redim(air=('air', 'air_temperature')).vdims[0].label
```

drs251:
Sorry, but I still don't get it: I'm absolutely unable to replace the rather long-winded "4xDaily Air temperature at sigma level 995" axis label in the plot within Holoviews. On the other hand, it's not essential for this PR though.

philippjfr:
Right, that's a separate issue in any case. I'll play around once merged and if I can reproduce I'll raise an issue.

PR #2434, opened 2018-03-12
Optimize range computations

philippjfr:
In benchmarking HoloViews for the pangeo project which will predominantly use dask arrays wrapped using xarray I identified a number of bottlenecks in the range calculations, which are being done inefficiently or unnecessarily.

### DIFF
--- a/holoviews/core/data/xarray.py
+++ b/holoviews/core/data/xarray.py
@@ -166,8 +166,8 @@ class XArrayInterface(GridInterface):
         else:
             data = dataset.data[dim]
             dmin, dmax = data.min().data, data.max().data
-            if dask and isinstance(dmin, dask.array.Array):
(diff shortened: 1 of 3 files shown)


philippjfr:
Ready to merge, tested on a 100GB image stack with a 5x improvement on initialization and 3x improvement for plot updates.

jlstevens:
Looks good and I am happy to merge. My only comment is that it would be good to eventually have performance tests for things like this. Also, how well do the unit tests cover the different code paths that can be used for calculating the range?

philippjfr:
>My only comment is that it would be good to eventually have performance tests for things like this

Completely agree.

>Also, how well do the unit tests cover the different code paths that can be used for calculating the range?

The same suite of tests is run for numpy and dask backed xarray Datasets.

PR #2436, opened 2018-03-12
Fixing version fetched via git archive

jlstevens:
I think I found the issue though I need to check it is working now before this PR is merged.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,2 @@
  __init__.py export-subst
+ setup.py export-subst


jlstevens:
I've tested ``pip+git`` against this branch/PR and it is now working as intended. When I run:

```bash
pip install https://github.com/ioam/holoviews/archive/gitattributes_archive_fix.zip
```

The install finishes with:

```
Successfully installed holoviews-0.0.0+g8374cef6f.gitarchive
```

Which reports the correct SHA. We may have a few last things to tweak regarding versioning but at least this much is working as intended. Merging.

PR #2438, opened 2018-03-12
Linked to and updated install guide

jlstevens:
Currently a draft and I'm not 100% sure the rst markup is correct. That said, it is worth updating this before 1.10.


jlstevens:
Unless anyone has any other suggestions for better instructions, I think this can be merged now.

philippjfr:
Looks good.

PR #2091, opened 2017-11-04
Notebook flake fixes

ceball:
I'm testing a notebook linting tool. These are the issues it found in the `doc/` directory, running it with python3.

It's not clear to me that everything it's found is correct, e.g.

* Is there a notebook about pandas that doesn't use pandas? Or is it because you need to import pandas to get features of the pandas backend?
* Do notebooks work in python 2 and python 3 with both the print statement and function? One notebook had both in.
* Are some things in notebooks deliberately imported and unused?

This is what I ran:
```
pip install pytest-nbsmoke
pytest --tb=short --nbsmoke-lint doc/
```

`--tb=short` is necessary because currently the combined pytest/nbconvert output is huge, unfortunately.

If I run nbsmoke on `examples/`, I get a lot more to fix. But first I wanted to see how you feel about these changes...

PR #2439, opened 2018-03-12
Notebook flake fixes.

ceball:
Following on from #2091, I did some more work on a notebook flakes tool, and ran it on examples/. This PR is the result.

Notes:
* examples/user_guide/Deploying_Bokeh_Apps.ipynb: `loop` comes up as being an undefined name; is it?
* examples/user_guide/Plots_and_Renderers.ipynb: `holoviews.plotting.mpl` is imported near the start, but is unused. I made it be used. You might not like the change. An alternative supported by the flakes tool would be to add `#noqa: reason` to the `holoviews.plotting.mpl` import, and reverting the change I've made here.

### DIFF
--- a/examples/user_guide/08-Gridded_Datasets.ipynb
+++ b/examples/user_guide/08-Gridded_Datasets.ipynb
@@ -13,7 +13,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import xarray as xr\n",
     "import numpy as np\n",
     "import holoviews as hv\n",
     "hv.extension('matplotlib')\n",
--- a/examples/user_guide/Deploying_Bokeh_Apps.ipynb
+++ b/examples/user_guide/Deploying_Bokeh_Apps.ipynb
@@ -387,6 +387,7 @@
    "source": [
     "def show_callback():\n",
     "    server.show('/')\n",
+    "TODO: where does loop come from?\n",
     "loop.add_callback(show_callback)\n",
     "server.start()\n",
     "# Outside the notebook ioloop needs to be started\n",
@@ -551,8 +552,6 @@
     "import numpy as np\n",
     "import holoviews as hv\n",
     "\n",
-    "from bokeh.application.handlers import FunctionHandler\n",
-    "from bokeh.application import Application\n",
     "from bokeh.io import show\n",
     "from bokeh.layouts import layout\n",
     "from bokeh.models import Slider, Button\n",
--- a/examples/user_guide/Deploying_Bokeh_Apps.ipynb
+++ b/examples/user_guide/Deploying_Bokeh_Apps.ipynb
@@ -385,12 +385,12 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "def show_callback():\n",
-    "    server.show('/')\n",
-    "TODO: where does loop come from?\n",
-    "loop.add_callback(show_callback)\n",
     "server.start()\n",
+    "server.show('/')\n",
+    "\n",
     "# Outside the notebook ioloop needs to be started\n",
+    "# from tornado.ioloop import IOLoop\n",
+    "# loop = IOLoop.current()\n",
     "# loop.start() "
    ]
   },
--- a/examples/user_guide/Plots_and_Renderers.ipynb
+++ b/examples/user_guide/Plots_and_Renderers.ipynb
@@ -83,7 +83,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "renderer = holoviews.plotting.mpl.MPLRenderer.instance(dpi=120)"
+    "renderer = hv.plotting.mpl.MPLRenderer.instance(dpi=120)"
    ]
   },
   {
--- a/examples/user_guide/Plots_and_Renderers.ipynb
+++ b/examples/user_guide/Plots_and_Renderers.ipynb
@@ -83,8 +83,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from holoviews.plotting.mpl import MPLRenderer\n",
-    "renderer = MPLRenderer.instance(dpi=120)"
+    "renderer = holoviews.plotting.mpl.MPLRenderer.instance(dpi=120)"
    ]
   },
   {
--- a/examples/getting_started/4-Gridded_Datasets.ipynb
+++ b/examples/getting_started/4-Gridded_Datasets.ipynb
@@ -14,7 +14,6 @@
    "outputs": [],
    "source": [
     "import numpy as np\n",
-    "import pandas as pd\n",
     "import holoviews as hv\n",
     "hv.extension('bokeh', 'matplotlib')"
    ]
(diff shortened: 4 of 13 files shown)


examples/user_guide/Deploying_Bokeh_Apps.ipynb
```diff
@@ -387,6 +387,7 @@
    "source": [
     "def show_callback():\n",
     "    server.show('/')\n",
+    "TODO: where does loop come from?\n",
```
philippjfr: Thanks for catching this, the IOLoop import must have gotten lost somewhere.

philippjfr: It used to be:

```python
from tornado.ioloop import IOLoop
loop = IOLoop.current()
```

I wonder if this is needed though or if we can just call ``server.show()`` right after ``server.start()``.

examples/user_guide/Plots_and_Renderers.ipynb
```diff
@@ -83,8 +83,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "from holoviews.plotting.mpl import MPLRenderer\n",
-    "renderer = MPLRenderer.instance(dpi=120)"
+    "renderer = holoviews.plotting.mpl.MPLRenderer.instance(dpi=120)"
```
jbednar: An alternative would be to do `from holoviews.plotting import mpl as hpm` at the top, then `hpm.MPLRenderer.instance` here...

philippjfr:
I've shortened it to ``hv.plotting.mpl.MPLRenderer``, I'm guessing that won't make the flakes tool happy but I'm personally not entirely convinced that level of strictness is appropriate for a notebook. Missing or completely unused (i.e. pointless) imports should be fixed, but I don't think we should start making the notebook purposefully verbose or add ``#noqa`` flags to satisfy it.

What's the flake tool called and can I start using it myself? I'd like to make a release checklist which would include things like running nbsmoke and this flake tool.

jlstevens:
Looks good to me. I agree we do *not* want to have ``#noqa`` in notebooks or as you say, make notebooks more verbose than necessary just to satisfy flake checks.

Merging.

ceball:
> I agree we do not want to have #noqa in notebooks or as you say, make notebooks more verbose than necessary just to satisfy flake checks.

The intent of supporting `# noqa: explanation` in nbsmoke is to encourage genuinely unavoidable or genuinely desirable 'mysterious imports' (i.e. ones where there's no subsequent usage of the name you've imported) to be clarified. E.g. if you're importing something for its side effects, it's very helpful to inform the reader of that. I guess I hope there wouldn't be such imports in the first place, but no doubt they make sense in places.

Anyone got a better suggestion?

And how do you feel about the above, @jbednar?

examples/user_guide/08-Gridded_Datasets.ipynb
```diff
@@ -13,7 +13,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "import xarray as xr\n",
```
ceball: Unfortunately I made a manual error here when dealing with the flakes caught by nbsmoke - `xr` is used. I'll make a new PR (probably alongside doing python 2 lint checks).

philippjfr:
> if you're importing something for its side effects, it's very helpful to inform the reader of that.

That's a good point and given that there seems to be exactly one such occurrence in the code I won't object marking it explicitly.

examples/user_guide/08-Gridded_Datasets.ipynb
philippjfr: Thanks!

ceball:
> What's the flake tool called and can I start using it myself? I'd like to make a release checklist which would include things like running nbsmoke and this flake tool.

It's nbsmoke: `pytest --nbsmoke-lint` (rather than `pytest --nbsmoke-run`). However, note that you'd need a recent version of nbsmoke (to get recent improvements to flake checking when there's magics involved). And also note that I am currently running nbsmoke lint checks on holoviews, pyviz, datashader to test nbsmoke itself :) Once that's done, I hope all our projects that have notebooks could use nbsmoke flakes checking on travis. (I *think* that's preferable to running it as part of a release checklist, but you probably know better than I do, as I've never released holoviews...)

philippjfr:
>Once that's done, I hope all our projects that have notebooks could use nbsmoke flakes checking on travis.

No objection from me, as long as the checks aren't any more strict than ``flake8 --ignore=E,W,F999,F405``.

jbednar:
>  how do you feel about the above, @jbednar?

I'd have to see examples of where noqa in notebooks was really necessary. If we can avoid it altogether, I think we should, because notebooks are generally for showing people how things work, and any obscure syntax like that will confuse people.  Of course, adding bogus actual statements like 0*val also confuses people, so it's a judgement call whether noqa is better if it comes down to that.

ceball:
> No objection from me, as long as the checks aren't any more strict than flake8 --ignore=E,W,F999,F405.

It uses pyflakes, so the idea is it should "never complain about style, and it will try very, very hard to never emit false positives."

ceball:
> I'd have to see examples of where noqa in notebooks was really necessary. If we can avoid it altogether, I think we should, because notebooks are generally for showing people how things work, and any obscure syntax like that will confuse people. Of course, adding bogus actual statements like 0*val also confuses people, so it's a judgement call whether noqa is better if it comes down to that.

By 'avoid it altogether', do you mean, 'avoid importing something that's not explicitly used, so noqa is never necessary in the first place', or do you mean, 'avoid adding noqa where something is imported but not explicitly used, and instead shut the linter up some other way'?

jlstevens:
Personally I would prefer it if we just make sure our notebook are written so noqa is never necessary. If we find noqa is often necessary, that suggests that there might be a legitimate case that it shouldn't be considered a flake.

jlstevens:
Notebooks are scary and confusing enough without magic comments. These comments will scare people who aren't developers.

jbednar:
>  'avoid importing something that's not explicitly used, so noqa is never necessary in the first place'

That's what I meant, yet.

ceball:
Ok, sounds like we're all in agreement, then. Nobody wants to see noqa in notebooks.

I think I've only encountered one time where I've used noqa, and it was to mark an apparently unused import as being necessary for a particular side effect, where I did not understand why it had to be that way, but it was outside my control.

ceball:
Philipp was right that the pyflakes would complain about the change he made to examples/user_guide/Plots_and_Renderers.ipynb:

```
$ cat test.py
import holoviews as hv
import holoviews.plotting.mpl
renderer = hv.plotting.mpl.MPLRenderer.instance(dpi=120)
$ pyflakes test.py
test.py:2: 'holoviews.plotting.mpl' imported but unused
```

Is that the best way to do it?

I admit I'm now no longer asking in order to make this notebook clearer than it already is, but because I may need to do something in nbsmoke if we have code in notebooks that pyflakes complains about.

Relevant pyflakes issues: https://github.com/PyCQA/pyflakes/issues/159, https://github.com/PyCQA/pyflakes/issues/248

philippjfr:
@ceball Tbh, I think we should just replace that with:

``renderer = hv.renderer('matplotlib').instance(dpi=120)``

and remove the import entirely. Probably needs some other edits to explain that properly and remove the other mention of ``hv.renderer``.

PR #2442, opened 2018-03-13
Fixed transposed expanded grid coords

philippjfr:
In investigating a bug in the ``Surface`` element I discovered that expanded coordinates are seemingly transposed from the expected orientation. Additionally the ImageInterface.coords method did not allow returning coordinate edges (rather than centers) like the other grid based interfaces.

If this PR passes there's clearly a gap in the unit tests, if it fails I'll have to investigate why this seemingly didn't cause issues or remove workarounds that were put in place elsewhere to make this work.

### DIFF
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -317,7 +317,7 @@ class GridInterface(DictInterface):
             return data.T.flatten() if flat else data
         elif expanded:
             data = cls.coords(dataset, dim.name, expanded=True)
-            return data.flatten() if flat else data
+            return data.T.flatten() if flat else data
         else:
             return cls.coords(dataset, dim.name, ordered=True)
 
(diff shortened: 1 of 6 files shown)


philippjfr:
There were indeed quite a few tests checking for the wrong thing. I've now also updated the tests and added two new tests (which are executed against all grid interfaces) to act as a sanity check.

philippjfr:
This is unlikely to have affected much user code since this is a fairly specialized API used mostly by the plotting code. It should however be included in the release notes.

philippjfr:

philippjfr:
Although I suspect some plotting code may have to be updated.

jlstevens:
Tests are green so I'll merge. Happy you can reason about when to transpose things because this sort of logic tends to melt my brain!

PR #2443, opened 2018-03-13
Ensure HoloMap.opts and .options apply to elements

philippjfr:
Makes HoloMap consistent with DynamicMap such that any options applied to it, get passed down to the objects contained by the HoloMap.

- [x] Fixes https://github.com/ioam/holoviews/issues/2408
- [x] Adds tests

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -72,6 +72,55 @@ class HoloMap(UniformNdMapping, Overlayable):
         return self.groupby(dimensions, container_type=NdLayout, **kwargs)
 
 
+    def opts(self, options=None, backend=None, clone=True, **kwargs):
+        """
+        Applies options on an object or nested group of objects in a
+        by options group returning a new object with the options
+        applied. If the options are to be set directly on the object a
+        simple format may be used, e.g.:
+
+            obj.opts(style={'cmap': 'viridis'}, plot={'show_title': False})
+
+        If the object is nested the options must be qualified using
+        a type[.group][.label] specification, e.g.:
+
+            obj.opts({'Image': {'plot':  {'show_title': False},
+                                'style': {'cmap': 'viridis}}})
+
+        If no opts are supplied all options on the object will be reset.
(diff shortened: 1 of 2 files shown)


jlstevens:
Happy to see the behavior made more consistent. I have two questions:

* Do we need to think about the implications wrt backwards compatibility?
* Is there any common code between ``HoloMap`` that can be used by ``DynamicMap``? (pretty sure the answer is 'no' but I want to check anyway).

Otherwise, tests are green and I am happy to merge.

philippjfr:
>Do we need to think about the implications wrt backwards compatibility?

I too wondered about this, I suspect not because if you were setting options using ``HoloMap.opts`` then you either explicitly defined the type, e.g. ``hmap.opts({'Image': ...})``, which still works, or your options would have been ignored entirely. In the latter case the original intention probably matches the new behavior.

> Is there any common code between HoloMap that can be used by DynamicMap? (pretty sure the answer is 'no' but I want to check anyway).

Right the answer is no, there is very little code anyway since it's mostly docstring, but I don't want users to lose the docstring. I really wish Python provided a clean way to inherit the docstring but override the method.

jlstevens:
I think I agree: no change to the old behavior which worked (and still works) and a new useful behavior which didn't used to to work so people wouldn't have used.

> I really wish Python provided a clean way to inherit the docstring but override the method.

I wish this were possible as well. I bet you could write a decorator to handle this but even a one line decorator would be repeated hundreds of times. By maybe it would be worth it for methods where the copied docstring is more than one line?

philippjfr:
>I bet you could write a decorator to handle this but even a one line decorator would be repeated hundreds of times. 

This is the the recipe that usually gets passed around for this: http://code.activestate.com/recipes/576862/ I think something like it might be worth considering but probably not until we have overhauled the docstrings more generally.

jlstevens:
> ... it might be worth considering but probably not until we have overhauled the docstrings more generally.

Agreed. Anyway, merging now.

PR #2444, opened 2018-03-13
Improved validation for nested DynamicMaps and other invalid nestings

philippjfr:
This PR extends the validation of nesting structures we perform during plotting. Currently when plotting a DynamicMap which returns another DynamicMap you receive an obscure error, this PR ensures you get a sensible error, which describes the two most common scenarios under which this might occur: 1) the callback actually returns a DynamicMap 2) an operation with dynamic=True was applied.

This PR also extends the validation of invalid nesting structures to Layouts/GridSpaces which contain objects which are invalidly nested.

- [x] Addresses https://github.com/ioam/holoviews/issues/2392

### DIFF
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -7,13 +7,14 @@ import param
 from bokeh.models import (ColumnDataSource, Column, Row, Div)
 from bokeh.models.widgets import Panel, Tabs
 
-from ...core import (OrderedDict, Store, AdjointLayout, NdLayout,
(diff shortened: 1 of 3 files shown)


philippjfr:
I think that might be trying to be too smart, using Dynamic with an operation/function is a possible solution in this situation but it might not necessarily apply so could just be more confusing.

philippjfr:
Ready to merge if you agree.

jlstevens:
This is a definite improvement over the old behavior. If we decide that a URL to a user guide would be more helpful (e.g if we find that users find this message confusing) then we can do that in a later PR.

Merging.

PR #2445, opened 2018-03-13
Fixed bokeh bug with offset being undefined

philippjfr:
Fixes a small bug in bokeh which occurs in certain layouts containing ``Empty`` objects.

- [x] Fixes https://github.com/ioam/holoviews/issues/2378


jlstevens:
Looks good. Merging.

PR #2454, opened 2018-03-16
Fixed MPLRenderer anim output regression

philippjfr:
During the update to the renderer code to make it work in JupyterLab a regression was introduced to rendering animation output using matplotlib (e.g. gif, mp4 and webm output). This PR fixes the regression.

Additionally this switches to the pillow animation writer for matplotlib>=2.2.0, which provides an easy way to generate gifs on windows.

- [x] Added unit tests
- [x] Addresses https://github.com/ioam/holoviews/issues/385, https://github.com/ioam/holoviews/issues/2095
- [x] Improved documentation of animation support

### DIFF
--- a/examples/user_guide/Plotting_with_Matplotlib.ipynb
+++ b/examples/user_guide/Plotting_with_Matplotlib.ipynb
@@ -133,16 +133,29 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Animated GIF support"
+    "## Animation support"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The ``'matplotlib'`` backend supports animated GIF output.  This is useful for output to web pages that users can view without needing to interact with.  It can also be useful for creating descriptive pages for HoloViews constructs that require a live Python/Jupyter server rather than just a web page - see for example [DynamicMap](../reference/containers/matplotlib/DynamicMap.ipynb).\n",
+    "The ``'matplotlib'`` backend supports animated outputs either as video (using mp4 or webm formats) or as animated GIFS. This is useful for output to web pages that users can view without needing to interact with.  It can also be useful for creating descriptive pages for HoloViews constructs that require a live Python/Jupyter server rather than just a web page - see for example [DynamicMap](../reference/containers/matplotlib/DynamicMap.ipynb).\n",
     "\n",
-    "Note that GIF support requires [ImageMagick](http://www.imagemagick.org) which is installed by default on many Linux installations and may be installed on OSX using [brew](http://brew.sh) . For more information on how to install ImageMagick (including Windows instructions) see the [installation page](http://www.imagemagick.org/script/binary-releases.php)."
+    "### GIF\n",
+    "\n",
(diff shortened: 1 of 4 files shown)


philippjfr:
Ready to merge.

examples/user_guide/Plotting_with_Matplotlib.ipynb
```diff
@@ -133,16 +133,29 @@
...
     "\n",
-    "Note that GIF support requires [ImageMagick](http://www.imagemagick.org) which is installed by default on many Linux installations and may be installed on OSX using [brew](http://brew.sh) . For more information on how to install ImageMagick (including Windows instructions) see the [installation page](http://www.imagemagick.org/script/binary-releases.php)."
+    "### GIF\n",
+    "\n",
+    "Note that GIF support requires [ImageMagick](http://www.imagemagick.org) which is installed by default on many Linux installations and may be installed on OSX using [brew](http://brew.sh) . For more information on how to install ImageMagick (including Windows instructions) see the [installation page](http://www.imagemagick.org/script/binary-releases.php).\n",
```
jlstevens: I am tempted to deprecate the code path that uses imagemagick entirely. Installing imagemagick on systems where it isn't already available is a pain.

philippjfr: matplotlib 2.2.0 with pillow support is only a few weeks old, so strong -1 from me on deprecating that code path now.

jbednar: I'm a bit confused; this PR says that it's using pillow rather than imagemagick, but the change above makes it sound like imagemagick is still needed?  I too vote for never using imagemagick if we don't need to, as it's a non-python dependency that's surprising to people and hard to manage.

philippjfr: I'm happy to make the documentation clearer on ImageMagick being required for earlier versions of matplotlib and pillow being fine for recent versions, but due to the matplotlib release being so recent I still don't think it should be deprecated.

jlstevens: I think it would be ok to tell people to upgrade their matplotlib and install pillow if they want gif support. On the plus side, the actual GIF support is handled by matplotlib so it doesn't really change how much code we have to maintain.

I'm happy to go with Philipp's suggestion as long an issue is filed to remove imagemagick support (and mention from the docs) in 1.11.

philippjfr: >On the plus side, the actual GIF support is handled by matplotlib so it doesn't really change how much code we have to maintain.

We do declare which backend it uses for rendering GIFs, maybe we can find a way to fall back to other backends cleanly though.

>I'm happy to go with Philipp's suggestion as long an issue is filed to remove imagemagick support (and references from the docs) in 1.11.

That seems sensible, but deprecating something in favor of a replacement that's only been around 2-3 weeks just seems too quick.

jbednar: Sounds good.

jlstevens:
I'll merge now with the understanding that Philipp will file an issue to remove the references to imagemagick when the time is right (I agree it is probably too early to do that now).

PR #2455, opened 2018-03-17
Avoid computing frame ranges for dynamic plots

philippjfr:
In the past Plot.keys would only initialize with the first key when plotting DynamicMaps, however at some point it was generalized such that it would contain all possible dynamic key combinations. This causes issues for the range computation because if framewise it will compute evaluate the entire dynamic parameter space on initialization, which is completely infeasible in many cases. This restores the old behavior of only computing the first frame range on initialization.

- [x] Fixes https://github.com/ioam/holoviews/issues/2373

### DIFF
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -422,8 +422,9 @@ class GridPlot(CompositePlot, GenericCompositePlot):
     def _create_subplots(self, layout, ranges):
         subplots = OrderedDict()
         frame_ranges = self.compute_ranges(layout, None, ranges)
+        keys = self.keys[:1] if self.dynamic else self.keys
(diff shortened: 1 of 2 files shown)


philippjfr:
Ready to merge.

jlstevens:
> In the past Plot.keys would only initialize with the first key when plotting DynamicMaps..

How does this interact with the ``default`` parameter of ``Dimension`` which should now be the initial key when set?

philippjfr:
>How does this interact with the default parameter of Dimension which should now be the initial key when set?

Shouldn't make any difference, it always initializes DynamicMaps on the first key, and then updates them to the default key if it differs. This is for consistency since the default key does not actually have to be defined.

jlstevens:
Ok. That makes sense.

Merging.

PR #2456, opened 2018-03-19
Define custom css for JupyterLab slider

philippjfr:
Has no effect on the CSS design in the classic notebook but overrides improves the styling for JupyterLab which does not include bootstrap.css.

screen shot 2018-03-19 at 12 54 32 am

There's still some things that could be improved but I haven't been able to figure out a way without affecting the styling in the classic notebook.

### DIFF
--- a/holoviews/plotting/widgets/jsslider.css
+++ b/holoviews/plotting/widgets/jsslider.css
@@ -22,6 +22,9 @@ form.holoform {
     padding-right: 0.8em;
     padding-top: 0.4em;
     padding-bottom: 0.4em;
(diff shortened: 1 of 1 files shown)


philippjfr:
Ready to merge.

holoviews/plotting/widgets/jsslider.css
```diff
@@ -1,71 +1,89 @@
 div.hololayout {
-    display: flex;
```
jlstevens: Could you add some comments to mark which bits of CSS map to the sliders in jupyterlab and which map to the classic notebook?

PR #2463, opened 2018-03-20
Apply utc offset to bokeh callback datetimes

philippjfr:
This fixes an issue dynamically regridding data. The problem was that when bokeh sent datetimes back to Python those were in the local timezone and therefore did not necessarily match the data. By applying an UTC offset to the callback values it seems the bug is fixed. However I don't pretend to know enough about timezones in general to know whether this will cause other issues.

- [x] Fixes https://github.com/ioam/holoviews/issues/2459

### DIFF
--- a/holoviews/plotting/bokeh/util.py
+++ b/holoviews/plotting/bokeh/util.py
@@ -53,7 +53,8 @@ def convert_timestamp(timestamp):
     """
     Converts bokehJS timestamp to datetime64.
     """
-    return np.datetime64(dt.datetime.fromtimestamp(timestamp/1000.))
+    datetime = dt.datetime.fromtimestamp(timestamp, dt.timezone.utc)
+    return np.datetime64(datetime.replace(tzinfo=None))
 
 
 def rgba_tuple(rgba):


philippjfr:
I think we should tentatively merge since it does fix a concrete bug and I can reproduce it, which I've been able to reproduce in different scenarios. It simply undoes the UTC offset that bokeh applies on serialization.

Handling timezones properly can be done and would probably be handled at the HoloViews level. Once this is merged I'll add an issue about it.

jlstevens:
>  It simply undoes the UTC offset that bokeh applies on serialization...

As long as you sure that serialization is fixed to UTC, I am happy to merge.

philippjfr:
Okay I just realized it has nothing to do with bokeh, ``datetime.fromtimestamp`` parses a timestamp in local time, I've now explicitly parsed the timestamp in UTC resulting in no offset being applied during the conversion. This behavior means we are timezone naive just like bokeh, we just display and return what we are given.

jlstevens:
I agree this is a more obvious and correct fix.

philippjfr:
It's effectively the same fix, except the first one was basically me stumbling on the fix by guessing, while this one is actually explicit about it.

philippjfr:
Okay, should be ready to merge now.

jlstevens:
Looks good. Merging.

jbednar:
Perfect, thanks!

PR #2465, opened 2018-03-20
Improved rasterize API

philippjfr:
This PR ensures that the ``rasterize`` and ``datashade`` operations correctly apply datashading operations to all elements that support it, specifically regridding of Image types can now also be invoked in this way. Secondly this PR improves handling for the aggregator functions, which may now be defined as strings or ``ds.reductions.Reduction`` types.
 
Adding documentation is a separate task which @jbednar volunteered for, specifically these two issues should be addressed by documentation updates:

https://github.com/ioam/holoviews/issues/1909
https://github.com/ioam/holoviews/issues/2158

The rasterize operation works by dispatching each Element type to a specific underlying operation with a specific default aggregator:

- ``Curve``/``Scatter``/``Points``/``Path``: ``aggregate`` (count)
- ``Image``/``RGB``/``HSV``: ``regrid`` (mean)
- ``TriMesh``: without vdims ``aggregate`` (count), with vdims ``trimesh_rasterize`` (mean)
- ``QuadMesh``: ``quadmesh_rasterize`` (mean)

- [x] Addresses https://github.com/ioam/holoviews/issues/2461
- [x] Add tests

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -198,7 +198,7 @@ class AggregationOperation(ResamplingOperation):
                     'min':   rd.min,
                     'max':   rd.max}
 
-    def _get_aggregator(self, element):
+    def _get_aggregator(self, element, add_field=True):
         agg = self.p.aggregator
         if isinstance(agg, basestring):
             if agg not in self._agg_methods:
@@ -223,7 +223,7 @@ class AggregationOperation(ResamplingOperation):
                                  '%s operation to. Declare the dimension '
                                  'to aggregate as part of the datashader '
                                  'aggregator.' % type(self).__name__)
-            agg = type(agg)(field)
+            agg = type(agg)(field if add_field else None)
         return agg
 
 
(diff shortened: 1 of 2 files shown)


jbednar:
Looks great!  I'm slightly confused about why regrid still has so much code in it.  I grudgingly can see why it still exists, in that regrid involves interpolation and the others don't, but I don't see why it has a lot of very specific code that does more than that.  Maybe some explanation of that would help.

philippjfr:
Sure, I'll add that.  It's almost exclusively datetime handling and converting non-xarray data to xarray.

philippjfr:
Ready to review and merge.

philippjfr:
Topics to cover in documentation:

* Explain ``rasterize`` and ``datashade`` operation being extra clear about 'eq_hist', I think it's currently not clear enough that ``shade`` does histogram normalization, leading people to draw false conclusions.
* Explain what rasterization does for Image types, including ``expand`` and ``upsample`` options which stop the regridded Image from expanding beyond bounds of original data, and disable upsampling by default, respectively
* Explain what ``precompute`` does and when to use it, i.e. it collapses the data into a monolithic datastructure which is faster to aggregate, if plotting a single element containing pandas/dask dataframe it's probably not worth it. When rasterizing images precompute does nothing.

jlstevens:
Looks good to me though you might want to consider moving ``ResamplingOperation`` out of this file as it is more general than just datashader.

Merging.

PR #2471, opened 2018-03-21
Updated docstrings and cleaned up inheritance for datashader operations

jbednar:
Minor updates following on from #2465:

1. Removed unnecessarily duplicated docstrings in subclasses (param docstrings are inherited even if the parameter is redefined with a different default in a subclass)
2. Created a new abstract base class to hold the "linked_inputs" parameter to avoid duplicated docstrings and allow users to set the value for all such plots as a single class parameter. 
3. Minor typo fixes and docs clarifications.

I would assume that only item 2 would be potentially controversial, but that parameter was duplicated in  three different places, with various errors from copying the docstring and not updating it, so I think it's important to have a single definition for what that parameter does.  Plus, it seems like a parameter that people might want to disable uniformly for all such plots, and right now to do that one would have to search the code to find all places where it's used and set each such class parameter.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -31,7 +31,21 @@ from ..element import (Image, Path, Curve, RGB, Graph, TriMesh, QuadMesh)
 from ..streams import RangeXY, PlotSize
 
 
-class ResamplingOperation(Operation):
+class LinkableOperation(Operation):
+    """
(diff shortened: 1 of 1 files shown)


philippjfr:
Looks good to me, will merge once tests pass.

jlstevens:
Looks good to me too.

Issue #2250, opened 2018-01-09
Warnings in the Live Data user guide


jlstevens:
I am seeing the following warnings:

![image](https://user-images.githubusercontent.com/890576/34723878-e93c1f84-f543-11e7-82b7-d78ad70b45e9.png)

It seems to be a zero division error (the warnings disappear if ``0`` isn't the first argument) although the warnings don't say that. For some reason, the error seems to originate from ipykernel instead of from the callback...

jlstevens:
I didn't look closely enough..it does say it is zero division - I had expected a dedicated exception. Apparently this is how numpy always raises such errors. In that case, the question is why the warning doesn't say it is issues from the callback. Anyhow, the notebook does need updating.

philippjfr:
Maybe just change the alpha value not to be zero.

PR #2472, opened 2018-03-22
Avoid zero division warnings in live data user guide

jlstevens:
Addresses #2250. I also thought it was worth making the backend explicit when loading the extension. Was there a good reason not to?


philippjfr:
>Was there a good reason not to?

No don't think so.

jlstevens:
I skimmed the text and I don't think I saw the actual values referenced in the markdown. The rest of the notebook doesn't have this issue so I think it is ready to review/merge.

philippjfr:
Looks good, merging.

PR #2473, opened 2018-03-22
Implemented and standardized GridSpace/NdLayout overlay operations

philippjfr:
This PR makes ``NdLayout.__mul__`` consistent with ``GridSpace.__mul__``. Both are homogeneous layout containers and should therefore overlay in the same way. Adding ``__mul__ `` support for Layout as discussed in the corresponding issue is imo a different topic that would require more discussion. In that case I think the suggested semantics are more confusing than helpful while the semantics for NdLayout/GridSpace are well defined. In the ``GridSpace*GridSpace`` or ``NdLayout*NdLayout`` items can be matched up based on their keys while zipping Layouts will surprise at least some people.

- [x] Addresses https://github.com/ioam/holoviews/issues/2075
- [x] Adds unit tests

### DIFF
--- a/holoviews/core/layout.py
+++ b/holoviews/core/layout.py
@@ -354,37 +354,6 @@ class NdLayout(UniformNdMapping):
         return clone
 
 
-    def __mul__(self, other, reverse=False):
-        if isinstance(other, NdLayout):
-            from .overlay import Overlay
-            if self.kdims != other.kdims:
-                raise KeyError("Can only overlay two NdLayouts with "
-                               "non-matching key dimensions.")
-            items = []
-            self_keys = list(self.data.keys())
-            other_keys = list(other.data.keys())
-            for key in unique_iterator(self_keys+other_keys):
-                self_el = self.data.get(key)
-                other_el = other.data.get(key)
-                if self_el is None:
-                    item = [other_el]
-                elif other_el is None:
-                    item = [self_el]
-                elif reverse:
-                    item = [other_el, self_el]
-                else:
-                    item = [self_el, other_el]
-                items.append((key, Overlay(item)))
-            return self.clone(items)
-
-        overlayed_items = [(k, other * el if reverse else el * other)
-                           for k, el in self.items()]
-        return self.clone(overlayed_items)
-
-
-    def __rmul__(self, other):
-        return self.__mul__(other, reverse=True)
-
 
 
 # To be removed after 1.3.0
--- a/holoviews/core/layout.py
+++ b/holoviews/core/layout.py
@@ -16,7 +16,7 @@ import param
 from .dimension import Dimension, Dimensioned, ViewableElement
 from .ndmapping import OrderedDict, NdMapping, UniformNdMapping
 from .tree import AttrTree
-from .util import unique_array, get_path, make_path_unique, int_to_roman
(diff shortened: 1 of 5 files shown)


philippjfr:
Ready to review and merge.

jlstevens:
What happens in the error conditions? E.g trying to overlay ``GridSpace`` or ``NdLayout`` objects with mismatched shapes or lengths?

philippjfr:
It's only allowed if the keys match exactly otherwise throws an exception. In future we could further improve it by actually matching the keys up. Matching by key is much better defined than matching them up by length, which is what we'd have to do for Layout and also the reason why I think we never should do that.

jlstevens:
> In future we could further improve it by actually matching the keys up. 

Could you file an issue about that? Once you do that, I'll merge.

philippjfr:
Decided to just implement partial matches and added some tests.

holoviews/core/layout.py
```diff
@@ -16,7 +16,10 @@
 from .dimension import Dimension, Dimensioned, ViewableElement
 from .ndmapping import OrderedDict, NdMapping, UniformNdMapping
 from .tree import AttrTree
-from .util import (unique_array, get_path, make_path_unique, int_to_roman)
+from .util import (
+    unique_array, get_path, make_path_unique, int_to_roman,
+    unique_iterator
```
jlstevens: At this point isn't it better to use ``util`` as a qualified namespace?

holoviews/core/spaces.py
```diff
@@ -1458,24 +1458,38 @@ def __init__(self, initial_items=None, kdims=None, **params):
...
+                if self_el is None:
+                    item = [other_el]
+                elif other_el is None:
+                    item = [self_el]
+                elif reverse:
+                    item = [other_el, self_el]
+                else:
+                    item = [self_el, other_el]
+                items.append((key, Overlay(item)))
+            print(items)
```
jlstevens: Stray print.

philippjfr: Thanks.

holoviews/core/layout.py
philippjfr: Sure.

jlstevens: Definitely makes sense when you have 4+ imports from ``util``.

jlstevens:
Other than my two comments, my main question is whether this could be done at the ``UniformNdMapping`` level? The block of code doing the key matching is pretty much repeated and both these classes inherit from ``UniformNdMapping``.

I suppose there might be import issue and maybe this isn't appropriate for other ``UniformNdMapping`` classes?

philippjfr:
>I suppose there might be import issue and maybe this isn't appropriate behavior for other UniformNdMapping classes?

Certainly isn't appropriate for HoloMaps or NdOverlays, but I suppose those will override the methods. I'll have a look at how much pain that would be.

jlstevens:
> I'll have a look at how much pain that would be.

Thanks for investigating. I'll leave it up to you to decide whether or not it makes sense.

philippjfr:
Had to change the inheritance order for ``NdOverlay`` slightly but I don't foresee any issues. Ready to merge.

jlstevens:
Looks good - less duplicated code than before and lots of new tests. Merging.

Issue #2099, opened 2017-11-07
Feature request: expose setting levels for hv.Bivariate


ea42gh:
I would like to be able to specify the levels of hv.Bivaraiate:
 I have a use case where I want to contrast the results of 2 algorithms
with 2 Bivariate plots next to each other

PR #2474, opened 2018-03-22
Allow specifying Bivariate levels

philippjfr:
This PR allows specifying levels for the Bivariate element as requested in #2099. Additionally it also ensures that the contours operation actually generates the requested number of levels, which was not the case before. Probably changes some test data but it's a clear bug fix.

- [x] Addresses https://github.com/ioam/holoviews/issues/2099
- [x] Added unit tests

### DIFF
--- a/tests/operation/testoperation.py
+++ b/tests/operation/testoperation.py
@@ -57,8 +57,9 @@ class OperationTests(ComparisonTestCase):
     def test_image_contours(self):
         img = Image(np.array([[0, 1, 0], [3, 4, 5.], [6, 7, 8]]))
         op_contours = contours(img, levels=[0.5])
-        contour = Contours([[(-0.5,  0.416667, 0.5), (-0.25, 0.5, 0.5)],
-                             [(0.25, 0.5, 0.5), (0.5, 0.45, 0.5)]],
+        contour = Contours([[(-0.5,  0.416667, 0.5), (-0.25, 0.5, 0.5),
+                             (np.NaN, np.NaN, 0.5), (0.25, 0.5, 0.5),
+                             (0.5, 0.45, 0.5)]],
                             vdims=img.vdims)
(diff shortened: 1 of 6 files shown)


jlstevens:
Looks good to me. Happy to merge when the tests are green.

philippjfr:
Ready to merge, push build failed due to updated test data.

tests/operation/testoperation.py
```diff
@@ -57,17 +57,18 @@ def test_image_gradient(self):
     def test_image_contours(self):
         img = Image(np.array([[0, 1, 0], [3, 4, 5.], [6, 7, 8]]))
         op_contours = contours(img, levels=[0.5])
-        contour = Contours([[(-0.5,  0.416667, 0.5), (-0.25, 0.5, 0.5)],
-                             [(0.25, 0.5, 0.5), (0.5, 0.45, 0.5)]],
+        contour = Contours([[(-0.5,  0.416667, 0.5), (-0.25, 0.5, 0.5),
+                             (np.NaN, np.NaN, 0.5), (0.25, 0.5, 0.5),
```
jlstevens: I don't quite understand where the NaNs come from...

jlstevens:
Change looks good to me - ``levels`` should be exposed. 

As long as you have checked that the notebooks with the updated test data now have the correct number of contours, I'm happy to merge once the tests are green.

tests/operation/testoperation.py
philippjfr: ``contours`` now returns a Contours element with one array per level, if there's multiple contours at a level they will be NaN separated.

philippjfr:
>As long as you have checked that the notebooks with the updated test data now have the correct number of contours, I'm happy to merge once the tests are green.

I have, also added unit tests to check this.

jlstevens:
Looks good and the pr tests are green. Merging.

PR #2479, opened 2018-03-23
Fixed expansion of Cycles of unequal lengths

philippjfr:
This PR implements the suggestion in https://github.com/ioam/holoviews/issues/2333, ensuring that cycles of unequal color are expanded by cycling over each item individually. It will try to expand the styles up to the defined max_cycles or until the combined cycle starts repeating. To prevent combinatorial explosion the total number of cycles is clipped at 1000.

- [x] Addresses https://github.com/ioam/holoviews/issues/2333 and fixes https://github.com/ioam/holoviews/issues/1852
- [x] Added unit tests

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -392,7 +392,7 @@ class Options(param.Parameterized):
        skipping over invalid keywords or not. May only be specified at
        the class level.""")
 
-    def __init__(self, key=None, allowed_keywords=[], merge_keywords=True, **kwargs):
+    def __init__(self, key=None, allowed_keywords=[], merge_keywords=True, max_cycles=None, **kwargs):
 
         invalid_kws = []
         for kwarg in sorted(kwargs.keys()):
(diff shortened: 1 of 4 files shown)


philippjfr:
@jlstevens Managed to implement this using a generator instead, it now lazily evaluates the cycle up to the maximum requested value and caches the intermediate results.

holoviews/core/options.py
```diff
@@ -474,21 +485,26 @@ def max_cycles(self, num):
         """
         kwargs = {kw: (arg[num] if isinstance(arg, Cycle) else arg)
                   for kw, arg in self.kwargs.items()}
-        return self(**kwargs)
```
jlstevens: I think this docstring should say this is about discretizing palettes. I also think the method name is very confusing but that would be an API change so it would not be something to fix in this PR...

jlstevens:
Unfortunately the current notebook tests don't seem to test this change but at least some of the unit tests do. @philippjfr If you are happy with my changes, please go ahead and merge.

philippjfr:
Looks good to me. Will merge once tests pass.

Issue #449, opened 2016-02-05
Notebook archive should support interactive holomaps


jlstevens:
Right now, HoloMaps are exported as a single frame. There are several issues with exporting holomaps in archives:
- HoloMaps can take a lot more disk space than static images.
- They are harder to test if we want to use archive testing in the future.
- Saving out separate HTML+JS for our javascript widgets is hard to include in `index.html` due to the same origin policy.

Nonetheless, we now have static HTML export and it would be nice to be able to use it. Ideally we could have standalone HTML files that display nicely by themselves but that can also be linked into an overall notebook document. Unfortunately, this looks like it might be quite tricky to get right.

Lastly, this might link to the proposal of storing the main holomap data in JSON files....

philippjfr:
I don't really understand the proposal here, we can and do support HoloMap exports now, but maybe not in archives? Please close if you think this has been addressed.

jlstevens:
I know we can export standalone html files for holomaps now but I would need to check how this applies in the context of archives.

PR #2480, opened 2018-03-23
Adding HTML holomap export to notebook archive

jlstevens:
This WIP PR is attempting to address #449.

Unfortunately, it isn't working yet as bokeh JS isn't included in the output. I see various issues:

* I think ``__call__`` should simply be the data saved to file by ``save``.
* The ``__call__`` method should not be used internally e.g in the ``html`` method.

By making ``__call__`` closely related to ``save``, users using ``__call__`` will get useful, self-contained output which is currently not the case.

If we fix those issues, I'll have to think about whether the changes I've currently made to the renderers needs to stick or go away.


philippjfr:
Closing, will try to fix this in #3836

PR #2482, opened 2018-03-24
Add HexTiles element

philippjfr:
This PR adds a HexTiles element which is very useful as a density map. The functionality is all there now, but this depends on the unmerged bokeh PR https://github.com/bokeh/bokeh/pull/7682. So far bokeh does not yet support hexagons with unequal scales along the two axes which is required for consistency with matplotlib and for it to be generally useful.

By default hex bins simply count the values in each bin but it may also be used for weighted binning.

screen shot 2018-03-24 at 2 28 15 pm

screen shot 2018-03-24 at 3 05 28 pm

On an implementation level we are letting matplotlib handle the binning internally while in the bokeh case we generate the hex binning using a Compositor and an operation. Unlike in other cases this operation is not user facing since the hex coordinates are not useful to a user. To make the Compositor work just for the bokeh backend it now includes an option to apply the compositor only for certain backends.

Notebook with examples: https://anaconda.org/philippjfr/hextiles/notebook

- [x] Adds HexTiles element
- [x] Addresses https://github.com/ioam/holoviews/issues/1141
- [x] Add unit tests
- [x] Add reference notebook entry


philippjfr:
Ready to review, and can be merged once the corresponding bokeh PR is merge: https://github.com/bokeh/bokeh/pull/7682.

examples/gallery/demos/matplotlib/hextile_movie_ratings.ipynb
```diff
@@ -0,0 +1,78 @@
...
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.6.4"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 2
```
jlstevens: Please clear out this metadata.

philippjfr: Will do.

holoviews/plotting/mpl/stats.py
```diff
@@ -204,3 +205,49 @@ def teardown_handles(self):
         for group in self.handles['box'].values():
             for v in group:
                 v.remove()
+
+
+class HexTilesPlot(ColorbarPlot):
```
jlstevens: Why is this in ``stats.py`` when the bokeh equivalent is in ``hex_tiles.py``?

philippjfr: Can create a new file here too, didn't seem worth it though given that it's fairly short and doesn't have to implement the aggregation/tiling code. Happy to move it though.

jlstevens: I would prefer consistency in structure when the elements supported match across backends.

philippjfr: Sounds good, I'll move it.

philippjfr:
All done, should be ready to merge when tests pass.

jlstevens:
Merging.

PR #2489, opened 2018-03-27
Added check for duplicate dataframe columns

philippjfr:
- [x] Addresses https://github.com/ioam/holoviews/issues/2476
- [x] Adds unit test

### DIFF
--- a/holoviews/core/data/pandas.py
+++ b/holoviews/core/data/pandas.py
@@ -80,6 +80,14 @@ class PandasInterface(Interface):
                 if eltype._auto_indexable_1d and ncols == 1 and kdim not in data.columns:
                     data = data.copy()
                     data.insert(0, kdim, np.arange(len(data)))
+
+            for d in kdims+vdims:
+                if isinstance(d, Dimension): d = d.name
+                if len([c for c in data.columns if c == d]) > 1:
+                    raise DataError('Dimensions may not reference duplicated DataFrame '
+                                    'columns (found duplicate %r columns). If you want to plot '
+                                    'a column against itself simply declare two dimensions '
+                                    'with the same name. '% d, cls)
         else:
             # Check if data is of non-numeric type
             # Then use defined data type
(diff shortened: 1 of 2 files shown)


philippjfr:
Ready to merge.

holoviews/core/data/pandas.py
```diff
@@ -80,6 +81,16 @@ def init(cls, eltype, data, kdims, vdims):
...
                     data = data.copy()
                     data.insert(0, kdim, np.arange(len(data)))
+
+            col_counter = Counter(data.columns)
+            for d in kdims+vdims:
+                if isinstance(d, Dimension): d = d.name
+                count = col_counter[d]
+                if d in col_counter and count > 1:
+                    raise DataError('Dimensions may not reference duplicated DataFrame '
+                                    'columns, found %d %r columns. If you want to plot '
```
jbednar: Maybe `columns (found %d %r columns).`

holoviews/core/data/pandas.py
```diff
@@ -80,6 +81,16 @@ def init(cls, eltype, data, kdims, vdims):
...
+
+            col_counter = Counter(data.columns)
+            for d in kdims+vdims:
+                if isinstance(d, Dimension): d = d.name
+                count = col_counter[d]
+                if d in col_counter and count > 1:
+                    raise DataError('Dimensions may not reference duplicated DataFrame '
+                                    'columns (found %d %r columns). If you want to plot '
+                                    'a column against itself simply declare two dimensions '
+                                    'with the same name. '% (count, d), cls)
```
jlstevens: I don't really see a need to use ``Counter`` for this. I think I would have done something like:

```python
columns = []
for d in kdims+vdims:
   d = d.name if isinstance(d, Dimension) else d
    if d in columns:
        raise DataError('Dimensions may not reference duplicated DataFrame '
                                     'column  %r. If you want to plot '
                                    'a column against itself simply declare two dimensions '
                                    'with the same name. '% d, cls)
   columns.append(d)
```

It is just as long as avoids an import.

philippjfr: That logic is wrong, we're checking for referenced columns that are duplicated in the dataframe, not duplicate dimensions. Happy to rewrite it without ``Counter`` though.

philippjfr:
Some issues with test data, seems some bad data was captured or merged at some point causing a bokeh elements test to fail, going to rebuild and merge this to keep other stuff moving. I've addressed all the comments and it's not a controversial PR.

PR #2492, opened 2018-03-29
Colormap categories (WIP)

jbednar:
This WIP PR is mainly for discussion, building on the unmerged PR for colormap_consistency.  Only the last two commits are new, so please only discuss those two.

The code isn't quite done, but I wanted to get feedback before going further. What it mainly adds is the ability to pull out any particular category, provider, or source of colormaps (e.g. `lookup_cmaps(category='Diverging',provider='bokeh', source='d3'))`, which I think will be very useful when building GUIs.  Usually, only a subset of the types are relevant, and it's very painful to list them out by name, yet it's also very painful for users to page through long lists of mostly inappropriate colormaps.  I think that this approach will allow a given app or dashboard to select just the subset of colormaps that would be relevant for it.

I've also put in the bare minimum version of listing the colormaps by category in the Styling user guide:

image

But note that to do this, I had to list each of the various colormaps provided, declaring their category, etc., because that information is not otherwise available programmatically except for colorcet.  This ended up being less verbose than I expected, but it does encode knowledge about the external backends that we should be aware that we are encoding.

There are still several tasks remaining:

- [x] Fix handling for reversed ('_r') colormaps
- [x] Define a useful default for colormaps not found in the lookup table
- [x] Find a better way to break up the big list of colormaps with big category labels (presumably not possible in the same cell, given the lack of hierarchical layouts?)
- [ ] Add examples of pulling out categories of colormaps, e.g. for a gui (maybe to pyviz?)
- [x] Sort colormaps in the displayed list
- [x] Clean up categorization of Bokeh maps to make sure they end up in the same category as similar Matplotlib maps
- [x] Consider merging list_cmaps with lookup_cmaps, e.g. making the current list_cmaps private and making lookup_cmaps be a superset of its functionality
- [x] There's no obvious way to specify a colormap by provider, and so when providers have matching-named colormaps the example currently shows only one of them, twice; must be fixed.
- [ ] Allow categorizing by background type as well, to make it simpler for dashboards to select the right set of useful colormaps?
- [ ] Allow excluding certain colors (white, black), for the same reason?


jbednar:
Other questions:

1. Does anyone see a better way to do any of this?
2. Is there a way to fix transpose=True so that I can specify the true columns (horizontally) regardless of how many rows there are?  I want a large number of separate sections, all with 5 or so columns left to right and plots laid out top to bottom in each column.  However, because each one has a different number of items, I have to do `transpose=True).cols(x)` with a different number `x` for each one, depending on the number of items, to get 5 columns consistently.  Isn't there some way I wouldn't have to do this?

philippjfr:
1. I certainly can't think of one
2. Probably not, not sure if there's already an issue about this, but there probably should be.

jbednar:
Ok, thanks.  I'll open a new PR shortly based on master and addressing many of these issues.

PR #2498, opened 2018-03-30
Colormap categories

jbednar:
This PR replaces #2492 (rebasing it on master).  What it mainly adds is the ability to pull out any particular category, provider, or source of colormaps (e.g. `list_cmaps(category='Diverging',provider='bokeh', source='d3'))`, which I think will be very useful when building GUIs.  Usually, only a subset of the types are relevant for any given plot, and it's very painful to list them out by name in an app or dashboard, yet it's also very painful for users to page through long lists of mostly inappropriate colormaps.  I think that this approach will allow a given app or dashboard to select just the subset of colormaps that would be relevant for it much more easily.

Using this facility, I've also created a new [Colormaps user guide](https://anaconda.org/jbednar/colormaps) that breaks down all of the available colormaps by type and explains how and when to use each one.

For any of this to work, I had to list each of the various colormaps provided across backends, declaring their category, etc., because that information is not otherwise available programmatically except for colorcet.  This ended up being less verbose than I expected, but it does encode knowledge about the external backends that we should be aware that we are encoding. Colormaps that are deleted from the packages in the future will simply fail to show up in these lists, and those that are added will appear in the 'Miscellaneous' category until someone adds some other category and source information for them.

Tasks required:

- [x] Fix handling for reversed ('_r') colormaps
- [x] Define a useful default for colormaps not found in the lookup table
- [x] Find a better way to break up the big list of colormaps with big category labels (presumably not possible in the same cell, given the lack of hierarchical layouts?)
- [x] Sort colormaps in the displayed list
- [x] Clean up categorization of Bokeh maps to make sure they end up in the same category as similar Matplotlib maps
- [x] Consider merging list_cmaps with lookup_cmaps, e.g. making the current list_cmaps private and making lookup_cmaps be a superset of its functionality
- [x] There's no obvious way to specify a colormap by provider, and so when providers have matching-named colormaps the example currently shows only one of them, twice; must be fixed.
- [x] Add examples of pulling out categories of colormaps, e.g. for a gui (maybe to pyviz?)
- [x] Allow categorizing by background type as well, to make it simpler for dashboards to select the right set of useful colormaps?
- [x] Consider allowing excluding certain colors (white, black), for the same reason? Decided against, as people can already do that programmatically by searching in the colormap itself.

### DIFF
--- a/holoviews/plotting/util.py
+++ b/holoviews/plotting/util.py
@@ -515,8 +515,9 @@ def bokeh_palette_to_palette(cmap, ncolors=None):
         reverse = True
 
     # Some colormaps are inverted compared to matplotlib
-    inverted = not (categorical or cmap.capitalize() in palettes.mpl)
-    reverse = (reverse and not inverted) or inverted
+    inverted = (not categorical and not cmap.capitalize() in palettes.mpl)
+    if inverted:
+        reverse=not reverse
     ncolors = ncolors or 256
 
     # Alias mpl tab cmaps with bokeh Category cmaps
@@ -676,13 +677,13 @@ def list_cmaps(provider=None, records=False, name=None, category=None, source=No
                        elif r.bg=='dark': 
                            r=r._replace(bg='light')
 
-                   if ((    name==None or name in r.name) and
-                       (provider==None or r.provider==provider) and
-                       (category==None or r.category==category) and
-                       (  source==None or r.source==source) and
-                       (      bg==None or r.bg==bg)):
+                   if ((    name is None or     name in r.name) and
+                       (provider is None or provider in r.provider) and
+                       (category is None or category in r.category) and
+                       (  source is None or   source in r.source) and
+                       (      bg is None or       bg in r.bg)):
                        matches.add(r)
-            if not matched and (category==None or category=='Miscellaneous'):
+            if not matched and (category is None or category=='Miscellaneous'):
                 # Return colormaps that exist but are not found in cmap_info
                 # under the 'Miscellaneous' category, with no source or bg
                 r = CMapInfo(aname,provider=avail.provider,category='Miscellaneous',source=None,bg=None)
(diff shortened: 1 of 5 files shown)


jbednar:
Ready for review and merge.

philippjfr:
Still a bit annoying that we have to list all the colormaps like that, but I don't see a way around it. Apart from that this is missing an entry in the user guide index. Otherwise I'd be happy to see this merged.

philippjfr:
There also seems to be a small flake:

>holoviews/plotting/util.py:695:56: F812 list comprehension redefines 'r' from line 688

jbednar:
Ok, I added it to the index, fixed the flake, and systematically evaluated all the reversed Bokeh colormaps, which showed that there was still a logic error in the reversing code (now fixed).  I also tidied up the categorizations a bit, to make sure all combinations of queries produced reasonable results.  Should be ready to merge if the tests pass.  If the tests don't, particularly the new Bokeh ones, it's worth checking that the tests are correct, as they were definitely letting through some unreversed `_r` Bokeh colormaps.

holoviews/plotting/util.py
```diff
@@ -515,8 +515,9 @@ def bokeh_palette_to_palette(cmap, ncolors=None):
         reverse = True
 
     # Some colormaps are inverted compared to matplotlib
-    inverted = not (categorical or cmap.capitalize() in palettes.mpl)
-    reverse = (reverse and not inverted) or inverted
+    inverted = (not categorical and not cmap.capitalize() in palettes.mpl)
+    if inverted:
+        reverse=not reverse
```
jbednar: I didn't mean to switch 518 around; that change has no effect.  The one on 519 is the important one.

philippjfr: Are you definitely sure the new logic is correct? I tried exactly what you did and I was fairly sure that was inverting them in some cases. I'll have another look.

philippjfr: I suppose you didn't touch the tests so those must be okay.

philippjfr: Yeah, so I forgot to add a sequential colormap reverse test and that's what's failing for me now.

philippjfr: Nevermind I hadn't pulled. I've added unit tests to cover each type of colormap now, just for consistency.

philippjfr:
Passing now, merging.

PR #2500, opened 2018-03-30
Improved notebook MIME type handling and added cleanup on cell deletion

philippjfr:
Currently the ``Renderer`` publishes JS code on two different MIME_TYPEs, one used by the classic notebook (``application/js``) and the other used by the JupyterLab extension (``application/vnd.holoviews_exec.v0+json``). The main issue with this is that all the data is sent both times a) doubling network traffic and b) doubling the size of the notebook.

This PR ensures that the Renderer will now only publish the actual data to the ``application/js`` MIME type. Additionally it adds a MIME handler for the classic notebook which processes the empty ``application/vnd.holoviews_exec.v0+json`` output and replaces it with ``application/js`` output, (which is what the jupyterlab extension was already doing). 

Finally this lays the groundwork to notify Python when a cell is deleted, which will allow us to clean up unused plots and stream callbacks.

- [x] Tested in classic notebook
- [x] Tested in JupyterLab
- [x] Tested in nbconvert export

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -252,7 +252,8 @@ class notebook_extension(extension):
     @classmethod
     def _process_comm_msg(cls, msg):
         """
-        Processes global comm messages to process deletions.
+        Processes comm messages to handle global actions such as
+        cleaning up plots.
         """
         if msg['event_type'] == 'delete':
             Renderer._delete_plot(msg['id'])
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -161,6 +161,9 @@ class notebook_extension(extension):
         Renderer.load_nb()
         for r in [r for r in resources if r != 'holoviews']:
             Store.renderers[r].load_nb(inline=p.inline)
+        if hasattr(ip, 'kernel'):
+            Renderer.comm_manager.get_client_comm(self._process_comm_msg,
+                                                  "hv-extension-comm")
 
         # Create a message for the logo (if shown)
         self.load_hvjs(logo=p.logo,
@@ -237,9 +240,6 @@ class notebook_extension(extension):
                                 'plotly_logo': plotly_logo,
                                 'message':     message})
         publish_display_data(data={'text/html': html})
-
-        Renderer.comm_manager.get_client_comm(cls._process_comm_msg,
-                                              "hv-extension-comm")
 
         # Vanilla JS mime type is only consumed by classic notebook
         # Custom mime type is only consumed by JupyterLab
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -238,6 +238,9 @@ class notebook_extension(extension):
                                 'message':     message})
         publish_display_data(data={'text/html': html})
 
+        Renderer.comm_manager.get_client_comm(cls._process_comm_msg,
+                                              "hv-extension-comm")
+
         # Vanilla JS mime type is only consumed by classic notebook
         # Custom mime type is only consumed by JupyterLab
         if JS:
@@ -245,6 +248,14 @@ class notebook_extension(extension):
                 MIME_TYPES['js']           : widgetjs,
                 MIME_TYPES['jlab-hv-load'] : widgetjs
             })
+
+    @classmethod
+    def _process_comm_msg(cls, msg):
+        """
+        Processes global comm messages to process deletions.
+        """
+        if msg['event_type'] == 'delete':
+            Renderer._delete_plot(msg['id'])
 
 
     @param.parameterized.bothmethod
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -15,6 +15,7 @@ from ...streams import Stream
 from ..plot import (DimensionedPlot, GenericCompositePlot, GenericLayoutPlot,
                     GenericElementPlot, GenericOverlayPlot)
 from ..util import attach_streams, displayable, collate
+from .callbacks import Callback
 from .util import (layout_padding, pad_plots, filter_toolboxes, make_axis,
                    update_shared_sources, empty_plot, decode_bytes,
                    bokeh_version)
@@ -245,6 +246,9 @@ class BokehPlot(DimensionedPlot):
         plots = self.traverse(lambda x: x, [GenericElementPlot])
         for plot in plots:
             for callback in plot.callbacks:
+                callbacks = {k: cb for k, cb in callback._callbacks.items()
+                            if cb is not callback}
+                Callback._callbacks = callbacks
                 for stream in callback.streams:
                     subscribers = [
                         (p, subscriber) for p, subscriber in stream._subscribers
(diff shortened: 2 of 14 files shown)


holoviews/plotting/widgets/widgets.js
```diff
@@ -539,3 +539,64 @@ for (var k in _namespace) {
...
+  }
+  events.on('output_added.OutputArea', handleAddOutput);
+
+  OutputArea.prototype.register_mime_type(EXEC_MIME_TYPE, append_mime, {
+    safe: true,
+    index: 0
+  });
+}
+
+if (window.Jupyter !== undefined) {
```
philippjfr: @jlstevens Any way of stopping this from running in labmode? IIRC you do define ``window.Jupyter``.

philippjfr:
@jlstevens I've decided to add the plot cleanup in this PR, it's not much code and I wanted to get your thoughts on it.

philippjfr:
I now have this working in JupyterLab as well, going to push a jupyterlab_holoviews 0.3.3 release as soon as this is merged.

philippjfr:
Ready for review.

holoviews/plotting/widgets/widgets.js
philippjfr: Added the try/catch you suggested.

holoviews/ipython/__init__.py
```diff
@@ -246,6 +249,14 @@ def load_hvjs(cls, logo=False, bokeh_logo=False, mpl_logo=False, plotly_logo=Fal
                 MIME_TYPES['jlab-hv-load'] : widgetjs
             })
 
+    @classmethod
+    def _process_comm_msg(cls, msg):
```
jlstevens: Maybe this method could have a more specific name like ``_delete_plot_by_id``?

holoviews/plotting/comms.py
```diff
@@ -57,12 +57,11 @@ class Comm(object):
 
     js_template = ''
 
-    def __init__(self, plot, id=None, on_msg=None):
+    def __init__(self, id=None, on_msg=None):
```
jlstevens: Nice to see that all these ``plot`` arguments and the ``_plot`` attribute were unnecessary!

holoviews/plotting/plot.py
```diff
@@ -63,6 +63,12 @@ def state(self):
         raise NotImplementedError
 
 
+    def cleanup(self):
+        """
+        Allows defining cleanup actions to perform on plot deletion.
+        """
```
jlstevens: Why not raise ``NotImplementedError`` here?

holoviews/plotting/widgets/widgets.js
```diff
@@ -539,3 +541,88 @@ for (var k in _namespace) {
...
+  if (comm !== null) {
+    comm.send({event_type: 'delete', 'id': id});
+  }
+  if ((window.Bokeh !== undefined) & (id in window.Bokeh.index)) {
+    window.Bokeh.index[id].model.document.clear();
+    delete Bokeh.index[id];
+  }
+}
+
+function register_renderer(events, OutputArea) {
```
jlstevens: Bit of an inconsistent mix of snake case and camel case going on here...

holoviews/plotting/bokeh/plot.py
```diff
@@ -241,6 +242,21 @@ def current_handles(self):
         return []
 
 
+    def cleanup(self):
```
jlstevens: Would be good to have a few more comment and/or a docstring describing what state is being cleaned up here...

Looks like you simply filter out the subscribers so the unreferenced elements can be garbage collected?

jlstevens:
@philippjfr Other than a few minor comments it looks good!

holoviews/ipython/__init__.py
philippjfr: I purposefully kept this general, I could imagine sending other types of messages, e.g. to shut down bokeh server sessions launched inside the notebook and if we ever find it's needed we could actually distinguish between JupyterLab and classic notebooks using a message.

holoviews/plotting/comms.py
philippjfr: Agreed, this has annoyed me in the past but I didn't think to look it was actually used.

holoviews/plotting/plot.py
philippjfr: Sure.

holoviews/plotting/widgets/widgets.js
philippjfr: Intentional, ``OutputArea`` is the name of a class.

holoviews/plotting/bokeh/plot.py
philippjfr: Agreed, I'll add a comment now.

holoviews/plotting/plot.py
philippjfr: Actually not totally convinced, a ``Plot`` doesn't necessarily have any cleanup to do, so they shouldn't have to define it.

holoviews/ipython/__init__.py
jlstevens: Sure. Personally I would use a specific name for the method while it *is* specific and rename it once it does handle the more general case. I don't mind leaving it as it is though...

holoviews/plotting/plot.py
jlstevens: Ok. In that case at least put in ``pass`` to make it explicit that something isn't missing.

holoviews/plotting/widgets/widgets.js
jlstevens: I'm talking about the functions in this file.

holoviews/plotting/plot.py
philippjfr: Nevermind the baseclass should actually clean up streams itself so this is a non-issue.

holoviews/plotting/bokeh/plot.py
philippjfr: >Looks like you simply filter out the subscribers so the unreferenced elements can be garbage collected?

Btw, elements or containers will almost never be garbage collected without user intervention since they will usually be in the global IPython namespace. All this cleanup has two main reasons:

* Streams should not push updates to plots which are no longer displayed.
* Plot/Widget/Callback instances can be cleaned up once the plot is no longer displayed.

holoviews/plotting/widgets/widgets.js
philippjfr: Ah, thanks for clarifying, that was confusing.

philippjfr: Fixed.

holoviews/plotting/bokeh/plot.py
jlstevens: You're right...I wasn't thinking carefully enough as subscribers don't point to elements but to plot methods (iirc). So your explanation makes sense and just need to be expressed in the comment...

jlstevens:
Looks good and you seem to have addressed all my suggestions/comments. Happy to merge when the tests are green and you tell me it is ready.

philippjfr:
Just pushed some unit tests, it's now ready to merge once tests pass. Fixing an issue in the JLab extension now, but that's pretty much independent from this PR.

jlstevens:
Tests are green. Merging.

Issue #2467, opened 2018-03-21
Datetime Support for Vectorfield


ahuang11:
![image](https://user-images.githubusercontent.com/15331990/37692710-6e93ae7e-2c88-11e8-9d88-003edab7c606.png)

Also I think it would be helpful to state that angles must be in radians before plotting here: http://holoviews.org/reference/elements/matplotlib/VectorField.html

jlstevens:
Seems like a simple enough change to the docs that I might as well get it out the way now.

philippjfr:
Seems like this is two distinct issues. I've bumped the issue the title references but adding a quick note about radians could happen in 1.10.

jlstevens:
Agreed. I didn't even notice that the title was a separate issue! 

I am going to fix the docs issue today. No idea about how much effort datetime support would be...

philippjfr:
Probably not too difficult, just need to use the same tricks I used in the datashader code, but that seems like a fix that can go in a 1.10.x release.

jlstevens:
The PR referenced above addresses the docs portion of this issue.

PR #2506, opened 2018-04-03
Clarified that VectorField angles are in radians

jlstevens:
Small tweak to the docs to address #2467, clarifying that the angles in ``VectorField`` are expressed in radians.


philippjfr:
Looks good, happy to merge.

philippjfr:
Will go ahead.

Issue #2262, opened 2018-01-18
Documentation unclear for installing bokeh 'pip install holoviews[recommended]'


lioumens:
The [installation documentation](http://holoviews.org/install.html) are somewhat unclear about what libraries the recommended install includes vs. extras. The documentation states, 

>This recommended installation includes the default Matplotlib plotting library backend, the more interactive Bokeh plotting library backend, and the Jupyter/IPython Notebook .
>A similar set of packages can be installed using pip , if that command is available on your system:
>pip install 'holoviews[recommended]'

This seems to imply Bokeh is installed with holoviews[recommended], whereas it is included with "extras". Since Bokeh is used in the [introduction](http://holoviews.org/getting_started/Introduction.html), can we include this in the recommended pip install?

jlstevens:
Thanks for raising this. I agree we should revisit this and update our docs and probably the extra requires - installing bokeh with 'recommended' now seems advisable given the direction HoloViews has moved in since those options were defined.

jlstevens:
@jbednar  @philippjfr Should I just move bokeh from the 'extras' list to the 'recommended' list?

jbednar:
Yes.

PR #2508, opened 2018-04-03
Moved bokeh from 'extras' to 'recommended' require list

jlstevens:
PR to address #2262.

@jbednar @philippjfr Do you think this is sufficient to address the above issue or is a change to the docs needed too? I think this will now behave as described....


jbednar:
Looks good to me...

philippjfr:
Seems reasonable to me as well, ready to merge (I'll just ignore the transient)?

jlstevens:
> Seems reasonable to me as well, ready to merge (I'll just ignore the transient)?

Sure!

PR #2511, opened 2018-04-03
Ensured Callback plot_id matches top-level

philippjfr:
The recent changes to Comms mean that it is now necessary to use the plot id to establish a Comm in JupyterLab. In bokeh this plot id is synonymous with the ID of the root-level bokeh model. In a composite plot the individual ElementPlots do not have access to the root level model because it isn't created until a later point. This PR inserts a placeholder into the callbacks CustomJS code, which the composite plot (i.e. LayoutPlot or GridPlot) replaces with the actual id once it is known.

- [x] Fixes https://github.com/ioam/holoviews/issues/2510

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -184,7 +184,7 @@ class CustomJSCallback(MessageCallback):
             comm_status.event_buffer = [];
         }}
 
-        function on_msg(msg){{
+        function on_msg(msg) {{
           // Receives acknowledgement from Python, processing event
           // and unblocking Comm if event queue empty
           msg = JSON.parse(msg.content.data);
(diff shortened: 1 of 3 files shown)


jlstevens:
Seems a bit fiddly but I don't have a better suggestion.

philippjfr:
>Seems a bit fiddly but I don't have a better suggestion.

Agreed, I'm not hugely happy with it but I'm still brainstorming. Perhaps we can generate a unique id at the top-level and simply assign it to the bokeh model, but that almost seems more fiddly/dangerous.

philippjfr:
Don't think I want to mess with bokeh model ids tbh, so this is the best approach I can think of. Ready to review.

jlstevens:
Looks fine. Merging.

PR #2514, opened 2018-04-03
Add a gitfilter to clean the output and metadata from an ipynb

basnijholt:
Regarding https://github.com/ioam/holoviews/issues/1617 and https://github.com/ioam/holoviews/pull/2507.

What about this?

You activate the filter with:
```bash
git config filter.nbstrip_jq.clean './nbstrip_jq.sh'
git config filter.nbstrip_jq.smudge cat
git config filter.nbstrip_jq.required true
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,3 @@
  __init__.py export-subst
  setup.py export-subst
+*.ipynb filter=nbstrip_jq
--- a/nbstrip_jq.sh
+++ b/nbstrip_jq.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# `nbstrip_jq.sh`:
+# This is a git filters that strips out Jupyter notebook outputs and meta data.
+# Execute the following lines in order to activate this filter:
+# conda install --yes jq  # or `brew install jq` or `apt-get install jq`
+# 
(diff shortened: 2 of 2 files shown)


jlstevens:
Thanks @basnijholt for the suggestion!

It is good to have a concrete proposal like this but I think there is more discussion to be had:

1. We would prefer to have a solution that doesn't require ``jq``
2. Can the filter only apply to specific directories?
3. Very occasionally we might want to commit notebooks with *some* output. How would you handle that if such a situation arises?
4. We probably need an approach that helps contributors who aren't regular enough developers that they would want to set up the git filters.
5. My ideal git filter would be one which applies when you *stage* a change, before you make the commit. This would give me the option to see the diff e.g if I work on a new machine where I haven't got the git filters configured.

At any rate, I think once we've thought it all through, git filters should be documented for the developers who want to use them.

basnijholt:
This is another solution [using `nbconvert`](https://github.com/basnijholt/ipynb_git_filters/blob/master/ipynb_filter.py) although it doesn't do all of the things @jlstevens mentioned above.

philippjfr:
Will close and maybe use your nbconvert solution.

PR #2516, opened 2018-04-04
Update conda.recipe to noarch

jlstevens:
- [x] Updates the recipe to build noarch
- [x] Makes things more consistent with the recipe on conda-forge.
- [x] Removes conda-forge from extra channels (only used at build time)

### DIFF
--- a/conda.recipe/build.sh
+++ b/conda.recipe/build.sh
@@ -1,2 +1,0 @@
-#!/bin/bash
-$PYTHON setup.py --quiet install --single-version-externally-managed --record=record.txt
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -10,7 +10,13 @@ source:
 extra:
   channels:
     - conda-forge
-    
+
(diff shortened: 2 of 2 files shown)


philippjfr:
Can't tell if the build.sh is now empty or whether you actually deleted it (which is what I'd expect). Otherwise this looks good to me.

philippjfr:
Looks like it's deleted, GitHub should really highlight that somehow.

jlstevens:
Odd. Github shows the deletion for me:

image

Anyway, the idea is that the last line there replaces the need for a script. As it is now noarch, there shouldn't be any worries about running across platforms (which I assume conda handles for you anyway).

jbednar:
How does that screenshot show the difference between deleting the contents and deleting the file itself?  If it does, it's subtle...

jlstevens:
I guess it doesn't make it clear that the whole file is removed. Anyway, ready to merge this PR?

philippjfr:
Merging.

PR #2519, opened 2018-04-04
Fixed bugs using groupby on transposed gridded data

philippjfr:
When using ``.to`` or ``.groupby`` to group a dataset by one or more dimensions it is possible to reorder the dimensions by passing explicit kdims in. This works fine for xarray datasets but when transposing gridded data in the dictionary format the underlying data actually needs to be transposed appropriately, otherwise it raises an error about mismatching shapes.

- [x] Add unit tests

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -527,6 +527,7 @@ class Dataset(Element):
             kdims = [self.get_dimension(d) for d in group_dims]
             group_kwargs = dict(util.get_param_values(self), kdims=kdims)
             group_kwargs.update(kwargs)
+            kdims = group_kwargs['kdims']
             drop_dim = len(kdims) != len(group_kwargs['kdims'])
             def load_subset(*args):
                 constraint = dict(zip(dim_names, args))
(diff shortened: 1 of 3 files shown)


philippjfr:
Ready to review/merge.

holoviews/core/data/__init__.py
```diff
@@ -527,14 +527,15 @@ def groupby(self, dimensions=[], container_type=HoloMap, group_type=None,
...
+            kdims = group_kwargs['kdims']
             drop_dim = len(kdims) != len(group_kwargs['kdims'])
             def load_subset(*args):
                 constraint = dict(zip(dim_names, args))
                 group = self.select(**constraint)
                 if np.isscalar(group):
                     return group_type(([group],), group=self.group,
                                       label=self.label, vdims=self.vdims)
-                data = group.reindex(group_dims)
+                data = group.reindex(kdims)
```
jlstevens: ``kdims`` is shadowed a number of times. If ``len(kdims)`` is the same for line 527 and 530, I would inline the access on ``group_kwargs`` (which doesn't seem to be mutated anywhere). i.e:

``data = group.reindex(group_kwargs['kdims'])``

If the lengths are different, I am happy to discard this suggestion.

jlstevens:
Made one comment. Otherwise happy to merge.

holoviews/core/data/__init__.py
philippjfr: Yah, I can probably clean this up a bit.

philippjfr: Cleaned it up differently.

jlstevens:
Just to say that it looks like the tests are about to go green, at which point I'll merge.

philippjfr:
Thanks!

Issue #2151, opened 2017-11-24
question: saving entire layout to a single png


icarus75:
When opting to save a layout to png via the toolbar, each hv element of the layout will be converted to a png individually. Is there an option to export the entire layout to a single png?

philippjfr:
Yes, you can render directly to png, using bokeh's export capabilities. This is a bit slow because it relies on Selenium, but it does what you want, e.g. in a notebook cell you can do this:

```python
%%output fig='png' filename='test'
layout = hv.Curve(range(10)) + hv.Area(range(10))
layout
```

or in a script:

```python
layout = hv.Curve(range(10)) + hv.Area(range(10))
hv.renderer('bokeh').save(layout, 'test', fmt='png')
```

jlstevens:
Can this issue be closed now? Last I remember you get separate pieces from a bokeh layout, but only when exporting SVG...

philippjfr:
Yes, exporting to png is possible using my answer above. Let's keep this open to document this though.

sfrodrigues:
How can you save a plot, via script, to obtain the save output as when you use the save button, .i.e without the bokeh toolbar on the side?

philippjfr:
The png export capability will always capture the toolbar if it is there, if you don't want it in the export just disable the toolbar entirely by setting the ``toolbar`` plot option to ``None``.

jlstevens:
I've always wondered why we shouldn't do this automatically. Is there any good reason you would want to see the bokeh toolbar in the export?

jlstevens:
Actually, I'll link to #2370 which should be fixed before we would consider that suggestion...

jbednar:
There may be cases where you would want the toolbar in the exported image, e.g. if you are capturing screenshots automatically for a tutorial, but I agree that it would be vastly preferable *not* to have the toolbar in an exported plot, by default.

sfrodrigues:
I agree that the default should be to not have the toolbar..

philippjfr:
I think all we need here is a short example or a link to a short example showing how to export using the magic and the renderer.

jlstevens:
The PR above specifically mentions Bokeh's PNG export and points to the Bokeh docs to find out about the extra dependencies.  The "Plots and Renderers" user guide now points to the exporting user guide where there is already an example of exporting to disk using the magic.

I think this issue can be closed once that PR is merged.

berkdemir:
[...]

PR #2520, opened 2018-04-04
Mentioned bokeh PNG export and saving to disk using the magic

jlstevens:
PR to address #2151 by mentioning bokeh's PNG export capability and pointing to a use of the ``%%output`` magic to save to file.


jlstevens:
This should be discussed more when we have a vanilla python user guide but I think this addition addresses  #2151 well enough for the time being. Ready to review/merge.

philippjfr:
Looks good.

PR #2522, opened 2018-04-04
Disabled auto-indexable on Dataset baseclass

philippjfr:
A while back it seem I accidentally enabled auto-indexing for the Dataset baseclass. The feature was meant as a convenience for 1D Chart types but absolutely should not apply to the baseclass which should accurately represent the underlying data and not fiddle with it. This needs to be called out explicitly in the release notes but should be done immediately since it is a bug and can make it awkward to work with ``Graph`` types which allow passing in a list or array of node indices to define an explicit order. This currently will not work correctly because it will add an index automatically.

- [x] Updated existing tests

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -173,7 +173,7 @@ class Dataset(Element):
 
     # In the 1D case the interfaces should not automatically add x-values
     # to supplied data
-    _auto_indexable_1d = True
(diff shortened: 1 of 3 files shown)


philippjfr:
Ready to merge.

jlstevens:
Looks good. I assume you checked ``_auto_indexable_1d`` attribute has the right boolean state everywhere it needs it.

Merging.

PR #2524, opened 2018-04-04
Now reading in README.md for PyPI description

jlstevens:
PYPI now supports markdown so we can get rid of the ugly, hardcoded blurb.


philippjfr:
Looks good.

PR #2525, opened 2018-04-04
Fixes for TriMesh rasterization

philippjfr:
This PR fixes issues determining the aggregator for ``TriMesh`` elements when no value dimensions are defined or the value dimensions are defined on the ``TriMesh.nodes``.

- [x] Fixes https://github.com/ioam/holoviews/issues/2523
- [x] Adds unit tests

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -227,6 +227,8 @@ class AggregationOperation(ResamplingOperation):
             inner_element = elements[0]
             if inner_element.vdims:
                 field = inner_element.vdims[0].name
+            elif isinstance(inner_element, TriMesh) and inner_element.nodes.kdims:
+                field = inner_element.nodes.vdims[0].name
             elif isinstance(element, NdOverlay):
                 field = element.kdims[0].name
             else:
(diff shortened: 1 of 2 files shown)


jbednar:
That helps, thanks.  Previously, only plot A would work:

```
import numpy as np, holoviews as hv, datashader as ds
hv.extension('matplotlib')
```

```
%%output backend='matplotlib'
%%opts TriMesh [filled=False]

np.random.seed(12)
N=100
pts = [(0.1*i, np.sin(0.1*i)) for i in range(N)]

opts = dict(filled=True, edge_color_index='z')
tri = hv.TriMesh.from_vertices(hv.Points(np.random.randn(N,3), vdims='z')).options(**opts)
(tri + tri.edgepaths + datashade(tri, aggregator=ds.mean('z')) + datashade(tri.edgepaths)).cols(2)

curves = {2:hv.Curve(pts),1:hv.Curve([(x,0.5*y) for x,y in pts])}

datashade(tri, aggregator=ds.mean('z')) + datashade(tri.options(filled=True)) + datashade(tri.options(filled=False))
```
image

However, I expected plot C to differ from plot B?

philippjfr:
We've never respected plot options in operations, and I wouldn't really want to start now. That said, we do need to come up with some way to do it.

philippjfr:
Should I maybe just switch to wiremeshes for ds.count and ds.any aggregators? This behavior doesn't seem particularly useful:

screen shot 2018-04-04 at 10 53 55 pm

jbednar:
The behavior for `.any()` is useful for making a mask for doing arithmetic, and `.count()` could be useful if we allow overlapping triangles (can't remember if we specifically allow or disallow that). In general for consistency I don't think I'd want to mess with either of those behaviors.  Seems like we need a special switch to change to wireframe, or else just show people how to specify that there is no vdim?

philippjfr:
>Seems like we need a special switch to change to wireframe, or else just show people how to specify that there is no vdim?

Or just explain that they should use ``tris.edgepaths``, because that can take a little while to compute and then doesn't have to be separately cached.

philippjfr:
IMO we should go ahead and merge this, ``datashade(tri.edgepaths)`` is the correct thing to recommend if the TriMesh defines value dimensions.

PR #2530, opened 2018-04-05
Support index_position for hv.Table

Dr-Irv:
Support bokeh 0.12.15 feature of `index_position` for `hv.Table`


jlstevens:
Looks good to me as well and the tests are green. Merging.

PR #2528, opened 2018-04-05
Removed embedded autover version.py

jlstevens:
I'll now test this PR against the param 1.6 alpha versions.

PR #2532, opened 2018-04-05
Maded embed_version function more robust

jlstevens:
Addresses a point made by @ceball in #2528 about how the ``embed_version`` function can be made more robust.


jlstevens:
Seems to be working using:

```
pip install git+https://github.com/ioam/holoviews.git@improved_embed
```

ceball:
The change looks good to me. I can't comment on whether or not it actually works, but at some point will update autover itself (https://github.com/ioam/autover/issues/35) and see if its tests pass.

Issue #1819, opened 2017-08-22
using holoviews and bokeh to generate html file without using iPython Notebook


LuckyGem:
I have a question about using holoviews and bokeh together to generate html file without using iPython Notebook. I have used bokeh before to generate interactive html file but I am not sure how to do it with holoviews. I want to have a time series graph and a dropdown menu(for example sales of fruit across year, the drop down menu will contain the different kind of frute), the graph will change according to fruite type.
For example, if I want to have a time series graph, with bokeh alone we can 
plot=TimeSeries(data,index='Date',title="fruit X",ylabel='sales')
output_file("graph.html")
show(plot)
I wonder if I can do something similar without using iPython Notebook? Really appreciate you help, thank you.

philippjfr:
Sorry for the delay, we really need to document this better. There are two ways of saving out data:

```
curve = hv.Curve(range(10)) # Could be any HoloViews object

renderer = hv.renderer('bokeh')

# Using renderer save
renderer.save(curve, 'graph.html')

# Convert to bokeh figure then save using bokeh
plot = renderer.get_plot(curve).state

from bokeh.io import output_file, save, show
save(plot, 'graph.html')
# OR
output_file("graph.html")
show(plot)
```

Sieboldianus:
Wow, this should really be added to the Holoviews Userguide [here](http://holoviews.org/user_guide/Deploying_Bokeh_Apps.html)!

hygull:
***Only specify html file name without an extension***
otherwise file will be saved as `graph.html.html`

elmbeech:
@philippjfr - this is semi related.

The explanation was really helpful for me!
I tried to use the same technique to store a hv.Scatter object using **plotly** as backed.
(Because there is a sudden feature on the ploty scatter plot I prefer over the bokeh one.)

This is my attempt and observation: 
```{python}
# load libraries
import pandas as pd
import plotly
import holoviews as hv
hv.extension('plotly')

# generate holoviews scatter object
df = pd.DataFrame([[1,2,3],[4,5,6]]).T
df.columns = ['x', 'y']
obj = hv.Scatter(df, 'x', 'y')
```

First I tried:
```{python}
hv.renderer('plotly').save(obj, 'output')
```
this generates a *output_1.html* file, 
but open it in a webbrowser only shows a **Drawing...** webpage (png attached).

[...]

PR #2534, opened 2018-04-06
Bokehapps

jbednar:
Addresses suggestion made in #1819. 

A remaining issue is that there are various tornado messages printed in the notebook throughout, along with a Bokeh deprecation warning in cell 19 after pressing play and then pause:

screen shot 2018-04-06 at 8 59 04 am

If anyone knows a fix for those issues, feel free to push them to this PR.  Otherwise ready for review and merge.

### DIFF
--- a/examples/user_guide/Deploying_Bokeh_Apps.ipynb
+++ b/examples/user_guide/Deploying_Bokeh_Apps.ipynb
@@ -34,7 +34,7 @@
     "\n",
     "This command will generate a file `out.html` that you can put on any web server, email directly to colleagues, etc.; it is fully self-contained and does not require any Python server to be installed or running.  \n",
     "\n",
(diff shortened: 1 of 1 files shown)


jbednar:
Actually, is there some reason we need casual users to know about the concept of a "renderer"?  Shouldn't we provide a very simple top-level function for them to export an object using the current backend?  E.g. something like this in the top-level hv namespace:

```python
def export(obj, filenamebase, backend=None):
    if backend is None:
        renderer = # look up current renderer
    else:
        renderer = hv.renderer(backend)
    renderer.save(obj,filenamebase)
```

So that `hv.export(obj,'out3')` would create `out3.html` if Bokeh is the current default plotting extension or `out3.png` if Matplotlib is the default (or if it is selected explicitly using  `hv.export(obj,'out4',backend='matplotlib')`)?  I'm not sure how to look up the current backend or the default renderer, but it seems like this would be much more obvious and convenient for users and would help dispel the notion that HoloViews is only really meant for use with Jupyter.

PR #2535, opened 2018-04-07
Replace display hooks with mimebundle

philippjfr:
This PR is an initial attempt at replacing the IPython display hooks with a ``_repr_mimebundle_`` method. There is still a bit of cleanup to do but this seems to work.
 
- [x] Fixes https://github.com/ioam/holoviews/issues/2462

### DIFF
--- a/holoviews/ipython/display_hooks.py
+++ b/holoviews/ipython/display_hooks.py
@@ -60,8 +60,7 @@ def render(obj, **kwargs):
     if renderer.fig == 'pdf':
         renderer = renderer.instance(fig='png')
 
-    data, metadata = renderer.components(obj, **kwargs)
-    return data, metadata
+    return renderer.components(obj, **kwargs)
 
 
 def single_frame_plot(obj):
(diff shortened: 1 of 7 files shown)


philippjfr:
This now works correctly and seems pretty clean to me. Once this is merged I will also have to push out jupyterlab_holoviews 0.4.0. I'm not sure if I have to update the notebook testing machinery yet, but it's definitely possible.

holoviews/ipython/display_hooks.py
```diff
@@ -127,40 +121,64 @@ def option_state(element):
         if raised_exception:
             dynamic_optstate(element, state=optstate)
 
+
+def mimebundle_to_html(bundle):
+    """
+    Converts a MIME bundle into HTML.
+    """
```
jlstevens: Minor and won't hold up the PR: would be good to say this is for adding to the notebook archive.

jlstevens:
Looks good and we already discussed improvements together which have now been implemented. I like the idea of ``Store.render`` which I feel was a missing piece of the puzzle relating renderers to the store. 

Unfortunately, the quality of our notebook testing output has decreased in order to get the notebook tests running at all: you no longer get feedback about which cells are running. This isn't too important as the key thing is to get the tests working and we plan to switch to ``nbval`` based notebook testing as soon as possible.

jlstevens:
The pr tests have passed. Merging.

PR #2536, opened 2018-04-07
Use HoloViews.plot_index to track bokeh models

philippjfr:
In a recent PR I added a ``HoloViews.plot_index`` which did little more than keep track of the ids of plots that were displayed. Since then I've found another issue related to the ``Bokeh.index``, which is bokeh's way of keeping track of the root models. The problem with ``Bokeh.index`` is that it is reset every time the extension is loaded (nothing we can do about that), which ends up breaking HoloMaps because they require access to the root model. This PR simply uses the ``HoloViews.plot_index`` to mirror the ``Bokeh.index``, which means that HoloMaps will continue  working even when the extension is reloaded.


jlstevens:
Looks good. Merging.

PR #2539, opened 2018-04-08
Fixed handling of categorical colormapping

philippjfr:
In the recent PR to make colormapping consistent across bokeh and matplotlib I adopted the categorical colormapping behavior that was used by matplotlib, which linearly interpolated the colormap. However for categorical values you instead want to cycle through the colors. This PR ensures that when colormapping a categorical dimension with a categorical colormap it cycles through the colors instead of interpolating between the colors.


jlstevens:
This does seem like a better policy to me. Merging.

PR #2542, opened 2018-04-08
XArray: enable overriding dimension names

drs251:
Referencing issue #2319, I made some small changes so that the automatic extraction of units and labels from Dataset and DataArrays is only done if no key or value dimensions are supplied, respectively. This makes it easy to override labels and units e.g. for a DataArray loaded from a netcdf file.


philippjfr:
Thanks for this, would be great to get this into 1.10. Any chance you could rebase the PR?

drs251:
OK, I think I managed to do a rebase without ruining my repo.

PR #2543, opened 2018-04-08
Add environment variable to combine mime bundle into HTML

philippjfr:


holoviews/ipython/display_hooks.py
```diff
@@ -154,18 +139,21 @@ def wrapped(element):
         try:
             max_frames = OutputSettings.options['max_frames']
             mimebundle = fn(element, max_frames=max_frames)
+            if mimebundle is None:
+                return {}, {}
+
+            html = mimebundle_to_html(mimebundle)
+            if os.environ.get('HV_DOC_HTML', False):
+                mimebundle = {'text/html': html}, {}
```
jbednar: For our future selves, it would be helpful to add a detailed comment explaining what this achieves, who is expected to set that variable, and why it's done this way.

jlstevens: Agreed.

In summary, the changes we've been forced to make for Jupyterlab don't work for doc building without this hack. In other words, you need to set this environment variable in doc building only to get working output.

jbednar: Sounds fine, as long as we leave an explanation of that.

jlstevens: The hope is that this problem only applies to our current doc building infrastructure and not when using ``nbsite`` which we hope to use in future. To my knowledge, we haven't checked yet.

PR #2544, opened 2018-04-08
Tweaks to docs for 1.10

jlstevens:
I am currently going through the notebooks to make sure they work correctly with 1.10. This PR makes a number of small fixes and tweaks.

One issue I found in the gridded data user guide:

image


jlstevens:
I'm going to update this notebook to use the ``.opts`` method. That said, two examples aren't working:

This is isn't being linked:

image

The tap example isn't working either:

image

jlstevens:
@jbednar I can't see a change moving this slider in the Large Data user guide:

image

I also don't quite get why matplotlib is being used here?

jlstevens:
In the Streaming Data user guide, the call to ``clear`` seems to have no effect:

image

jlstevens:
In the dashboards user guide I am seeing our favorite warning:

image

Everything seems to be working regardless except for the last example. Moving the slider generates these warnings:

image

jlstevens:
You can't see it now, but in the "Installing and Configuring" user guide, running:

image

Sometimes shows a parameter warning about setting the state of ``AdjointLayout.show_frame``.

jlstevens:
The tabs don't seem to be working in the 'Plotting with Bokeh' user guide. Preliminary investigation suggests it *might* be a bokeh issue:

image

jlstevens:
In Plotting with Bokeh, only the diagonal seems to update:

image

jlstevens:
In the Plotting with Matplotlib user guide, I get an error rendering to mp4:

image

jlstevens:
Now I'm going through the examples.  I'm getting some really weird behavior in ``box_draw_roi_editor.ipynb``:

Initially it displays but box select doesn't do anything:


image

If you run it again:

image

jlstevens:
In hex movie ratings:

image

jlstevens:
In iris density grid (bokeh):

image

Note this is after rebasing with #2547.

jlstevens:
The matplotlib measles example demo is broken:

image

jlstevens:
In ``route_chord.ipynb`` there is a problem getting the airports dataset:

image

End of the traceback:

image

jlstevens:
There is also a problem getting the matplotlib sample data for ``topographic_hillshading.ipynb`` (both bokeh and matplotlib versions):

image

jlstevens:
Matplotlib heatmap isn't working in the ``us_unemployment.ipynb`` example:

image

jbednar:
> I'm going to update this notebook to use the .opts method

Why .opts and not .options?

> @jbednar I can't see a change moving this slider in the Large Data user guide:

Yes, I already reported that to @philippjfr, who said that he'd update something about bokeh indexes, I think?

> I also don't quite get why matplotlib is being used here?

I used matplotlib just to match the previous much larger set of examples (where I didn't want to make the load time for the page huge) and so I wouldn't have to explain that the data would be dynamically updated if you have a live Python server.  But that's only a minor concern; feel free to change it to Bokeh if you prefer.

> Important: It looks like upgrading to matplotlib 2.2.2 fixed a large number of unit tests running locally.

Should we pin matplotlib to >= 2.2.2?

jlstevens:
> Why .opts and not .options?

That would require more work and thought. The goal was just to quickly get rid of the deprecation warnings about using ``__call__`` and get through testing all the example notebooks manually.

The good news is that ``.opts(`` is easy to grep so we can convert it to ``.options`` in a separate PR if we have time.

> Should we pin matplotlib to >= 2.2.2?

I would rather we didn't have to but fixing things to be backwards compatible might be more effort than it is worth.

philippjfr:
>Yes, I already reported that to @philippjfr, who said that he'd update something about bokeh indexes, I think?

@jbednar As far as I can tell it's not a HoloViews bug but directly related to your example, not 100% sure though.

philippjfr:
>Should we pin matplotlib to >= 2.2.2?

I think matplotlib>=2.1 should be sufficient.

jlstevens:
@philippjfr I see a number of errors in the ``sri_model.ipynb`` topic example.

philippjfr:
>@philippjfr I see a number of errors in the sri_model.ipynb topic example. e.g hitting play does nothing:

Worth checking the scrubber widget works elsewhere, I'll do that tomorrow morning.

PR #2545, opened 2018-04-08
Fixed bug in GridInterface.groupby

philippjfr:
Fixes bug in ``GridInterface.groupby`` when only a single value dimension is used which can be reproduced by an example in the Gridded Datasets user guide.


jlstevens:
Good to have a fix. Will merge when the tests pass and I'll test the user guide again.

PR #2547, opened 2018-04-08
Fixed contours operation empty level bug

philippjfr:
Fixes bug in contours operation when a level is empty.


jlstevens:
I believe this change is safe. Merging.

PR #2549, opened 2018-04-09
Fixed matplotlib HeatMap extents

philippjfr:
HeatMap extents weren't being set correctly when the values were numeric.

- [x] Adds unit test


philippjfr:
So I think matplotlib HeatMaps have been broken for quite a while, this needs a test data update.

PR #2552, opened 2018-04-09
Fixes for cleanup of LayoutPlots

philippjfr:
The plot cleanup code was not correctly cleaning up LayoutPlots which caused events to be sent to non-existent plots on the JS side.


jlstevens:
Tests are green and I believe this is PR is now ready. Merging.

PR #2553, opened 2018-04-09
Use different bokeh Documents in notebook context

philippjfr:
In the notebook all bokeh plots were attached to curdoc(), resulting in references to deleted plots to be kept and causing issues sending updates to multiple plots at once.


jlstevens:
@philippjfr I suppose the test data needs updating?

PR #2554, opened 2018-04-10
Ensure all tabular interfaces treat empty list equally

philippjfr:
This PR ensures that all tabular interfaces treat an empty list equally generating data columns of float64 types. Previously the pandas interface would interpret an empty list as 'object' type data which is inconsistent with other interfaces and caused issues in some of our examples (specifically the Tap example in the Custom Interactions user guide).


jlstevens:
I agree the object type is something to be avoided. Tests are green so I'll go ahead and merge.

PR #2557, opened 2018-04-10
Switched to parambokeh in Dashboards user guide

philippjfr:
Latest versions of ipywidgets are basically broken as far as I can tell and since we aim to switch to paramBokeh more generally I've replaced the usage of paramnb with parambokeh in the Dashboards notebook.


jlstevens:
Looks good and the push build passed. Merging.

PR #2561, opened 2018-04-11
Updated bokeh version used on travis

jlstevens:
Updates the bokeh version in travis.yml.


jlstevens:
This PR could also pin IPython (to avoid the bad exception problem) and matplotlib. We just need to be confident that the pinning is appropriate.

philippjfr:
I think we should go ahead with this and upgrade IPython and matplotlib as well.

jlstevens:
For reference, the matplotlib version that appears to be specified in the ``.travis.yml`` is matplotlib=2.1.2.

jlstevens:
Tests are updated and now green. Merging.

PR #2574, opened 2018-04-17
Add support for HTML export after moving to MIME bundle

philippjfr:
When we switched to rendering our plots using the ``_repr_mimebundle`` we lost all support for exporting bokeh plots and holoviews widgets to standalone HTML files. This is because nbconvert will only render a single MIME type, whichever is highest in its priority ordering. Since ``application/javascript`` has the highest priority only that gets rendered and doesn't end up finding the HTML tags it's meant to render into.

The only approach I could possible see how to make it work is to dynamically create the HTML DOM nodes that the JS renders into. So this PR adds a bit of JS code to each plot that includes ``application/javascript``, which looks for an existing DOM element with the matching ``element_id`` and if that doesn't exist it dynamically creates it on its own parent node. This means that in the notebook nothing changes but in a static HTML export the DOM nodes are appropriately created.

I have to say I really dislike this approach but it is absolutely the only way I see to make this possible, since I can only publish a single MIME type in the nbconvert output and ``application/javascript`` will always take precedence.

### DIFF
--- a/holoviews/plotting/renderer.py
+++ b/holoviews/plotting/renderer.py
@@ -70,8 +70,9 @@ static_template = """
 """
 
 embed_js = """
-// Sorry for this hack! See #2574 if you want to be sad
+// Ugly hack - see #2574 for more information
 if (!(document.getElementById('{plot_id}')) && !(document.getElementById('_anim_img{widget_id}'))) {{
+  console.log("Creating DOM nodes dynamically for assumed nbconvert export. To generate clean HTML output set HV_DOC_HTML as an environment variable.")
   var htmlObject = document.createElement('div');
   htmlObject.innerHTML = `{html}`;
   var scriptTags = document.getElementsByTagName('script');
--- a/holoviews/plotting/renderer.py
+++ b/holoviews/plotting/renderer.py
@@ -70,7 +70,8 @@ static_template = """
 """
 
 embed_js = """
-if (!(document.getElementById('{plot_id}')) && !(document.getElementById('_anim_img{element_id}'))) {{
+// Sorry for this hack! See #2574 if you want to be sad
+if (!(document.getElementById('{plot_id}')) && !(document.getElementById('_anim_img{widget_id}'))) {{
   var htmlObject = document.createElement('div');
   htmlObject.innerHTML = `{html}`;
   var scriptTags = document.getElementsByTagName('script');
@@ -318,11 +319,12 @@ class Renderer(Exporter):
         else:
             plot, fmt = self._validate(obj, fmt)
 
+        widget_id = None
         data, metadata = {}, {}
         if isinstance(plot, NdWidget):
             js, html = plot(as_script=True)
             plot_id = plot.plot_id
-            element_id = plot.id
+            widget_id = plot.id
         else:
             html, js = self._figure_data(plot, fmt, as_script=True, **kwargs)
             plot_id = plot.id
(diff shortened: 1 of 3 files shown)


jlstevens:
Isn't this arguably a bug in nbconvert? Shouldn't the output of ``_repr_mimebundle`` have the highest precedence?

philippjfr:
>Isn't this arguably a bug in nbconvert?

No.

>Shouldn't the output of _repr_mimebundle have the highest precedence?

No, the output of ``_repr_mimebundle`` contains a bunch of things and only the thing with the highest precedence that is supported by the output type (in this case HTML) will make it into the converted output. According to the nbconvert docs each MIME type should be an alternative but standalone representation of the plot. So what this PR does is make the highest priority item (namely ``application/javascript``) be able to render itself as a standalone thing.

jlstevens:
> According to the nbconvert docs each MIME type should be an alternative but standalone representation of the plot.

That doesn't work when you have to separate HTML from JS which you now have to do for Jupyterlab. I agree this is by 'design' but that design now looks broken to me.

holoviews/plotting/renderer.py
```diff
@@ -322,7 +334,10 @@ def components(self, obj, fmt=None, comm=True, **kwargs):
                                                        comm_id=plot.comm.id,
                                                        plot_id=plot_id)
                 js = '\n'.join([js, comm_js])
-            html = "<div style='display: table; margin: 0 auto;'>%s</div>" % html
+            element_id = plot_id
+            html = "<div id='%s' style='display: table; margin: 0 auto;'>%s</div>" % (element_id, html)
+        if not os.environ.get('HV_DOC_HTML', False) and js is not None:
+            js = embed_js.format(element_id=element_id, plot_id=plot_id, html=html) + js
```
jlstevens: Why define ``element_id`` if you can just replace this with ``plot.id`` (or ``plot_id``)?

jlstevens: Nevermind, I see it now. Though isn't it ``plot.id`` in both branches?

holoviews/plotting/renderer.py
```diff
@@ -69,6 +69,17 @@
 </html>
 """
 
+embed_js = """
```
jlstevens: I suggest leaving a comment pointing to this PR where we can explain why such a nasty thing exists.

philippjfr: Agreed, don't want anyone blaming us for this.

holoviews/plotting/renderer.py
philippjfr: Bit confusing but in the widget case they aren't identical, in that case ``plot`` variable is actually the widget and the two are different. I might just call it ``widget_id`` instead of element_id`` and then default it to None in the non-widget case.

jlstevens: Ok. If it can be made clearer, then great!

jlstevens:
Looks as good as it is going to get and the tests have finally gone green.

For anyone directed to this PR from the JS comment, this is a very ugly hack that we did not want to include in our code. Unfortunately, the alternatives were even less palatable 1) prematurely drop support for classic notebook in favor of JupyterLab 2) send all the data to the notebook twice. 

Given that we found those alternatives unacceptable, we opted for this approach which should only ever execute in the context of HTML exported using nbconvert.

PR #2584, opened 2018-04-19
Fixed unreferenced variable in matplotlib colormapping

philippjfr:
Fixes bug when clims is passed to a colormapped matplotlib plot and the ``categorical`` variable is then not defined.

- [x] Added unit test


jlstevens:
Looks good. Merging.

PR #2585, opened 2018-04-19
Changed Image.rtol to a warning and increased default tolerance

philippjfr:
An exception turned out to quite disruptive to a large number of users and the error was less than informative. We've now further increased the tolerance and switched from an exception to more informative warning.

- [x] Updates unit tests


philippjfr:
Needs updated test data due to the Options notebook unfortunately.

philippjfr:
Looks like there's some unit test failures from some really old unit tests which still compare ``.data``, will update those first thing tomorrow.

PR #2588, opened 2018-04-20
Allowing use of upcoming 0.12.16 bokeh release

jlstevens:
Simple update to how bokeh is pinned in the meta.yaml.


philippjfr:
Looks good thanks.

PR #2589, opened 2018-04-20
Support Plot in ipython display function

philippjfr:
In the past we allowed Plot objects to display themselves, this was a bad idea. For easier debugging I therefore suggest we instead support it in the ipython ``display`` function.


jlstevens:
This looks good. Merging.

PR #2597, opened 2018-04-21
Add support for Tool.renderers == 'auto'

philippjfr:
In the upcoming bokeh 0.12.16 some tools may now have ``Tool.renderers == 'auto'``, something that is not handled by the current code. This PR ensures that if the renderers are set to 'auto' HoloViews will explicitly override that.


jlstevens:
Looks fine. Merging.

PR #2598, opened 2018-04-21
Generalized display function again

philippjfr:
In the past the ``hv.ipython.display`` function simply extended the IPython version to add support for displaying HoloViews objects. When refactoring the MIME types that generality was lost and non-holoviews objects were simply rendered as their ``repr``, which disrupted various usages of the function inside the extension, e.g. to adjust the ``width`` or display various warnings.

- [x] Fixes https://github.com/ioam/holoviews/issues/2548


jlstevens:
Makes sense. Happy to merge when the tests pass.

Issue #2411, opened 2018-03-06
Plot and style option name clashes


philippjfr:
The new .options method relies on unique names for plot and style options. At least in one case that is currently not true, namely the width option on Bars is both a plot options (controlling the width of the plot) and a style option (setting the width of the bars). We need to validate that these do not clash, and in the case of width rename the width style option to ``bar_width``.

jlstevens:
I hope this is the only example of a clash we find.

philippjfr:
There's a few more, but it's not too bad:

```python
from itertools import combinations

def option_intersections(backend):
    options = hv.Store.options(backend)
    for k, opts in options.items():
        if len(k) > 1: continue
        eltype = k[0]
        valid_options = {k: set(o.allowed_keywords)
                         for k, o in opts.groups.items()}
        for g1, g2 in combinations(['plot', 'style', 'norm'], 2):
            intersection = valid_options[g1] & valid_options[g2]
            if intersection:
                print('%s element %s and %s %s backend options intersect on: %s'
                      % (eltype, g1, g2, backend, intersection))
       
for backend in hv.Store.renderers:
    option_intersections(backend)
```

Bars element plot and style **bokeh** backend options intersect on: {'width'}
BoxWhisker element plot and style **bokeh** backend options intersect on: {'width'}

Arrow element plot and style **matplotlib** backend options intersect on: {'fontsize'}
Text element plot and style matplotlib backend options intersect on: {'fontsize'}

Curve element plot and style **plotly** backend options intersect on: {'width'}
ErrorBars element plot and style **plotly** backend options intersect on: {'width'}

jlstevens:
Maybe for ``Arrow`` it could be called ``textsize``. Not sure about all the others though.

philippjfr:
Bokeh Bars: width -> bar_width
Bokeh BoxWhisker: width -> box_width
Matplotlib Arrow: fontsize -> textsize
Matplotlib Text: fontsize -> textsize
Plotly Curve/ErrorBars: delete width as a style option

jlstevens:
Those suggestions seem reasonable enough. The problem then is retaining the old names (and therefore the clashes) for backwards compatibility...

philippjfr:
[...]

PR #2604, opened 2018-04-22
Deal with clashing style and plot options

philippjfr:
A number of options exist both as plot and style, which results in issues when using ``.options``. The clashing options are listed in https://github.com/ioam/holoviews/issues/2411 

- [x] Fixes https://github.com/ioam/holoviews/issues/2599
- [x] Fixes https://github.com/ioam/holoviews/issues/2411
- [x] Adds unit tests to avoid future option clashes

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -660,7 +660,7 @@ class BarPlot(ColorbarPlot, LegendPlot):
        Index of the dimension in the supplied Bars
        Element, which will stacked.""")
 
-    style_opts = line_properties + fill_properties + ['width', 'cmap']
+    style_opts = line_properties + fill_properties + ['width', 'bar_width', 'cmap']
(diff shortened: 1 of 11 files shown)


philippjfr:
As I described in #2411 I've done one of three things with the clashes:

1. Bokeh HeatMap ``width`` and matplotlib Arrow/Text ``fontsize`` style options removed completely because they were not used.
2. Bokeh BoxWhisker and Bars ``width`` style option given an alias to allow eventual deprecation of clashing option.

I've also added unit tests to validate clashes on all backends, except those aliased in 2.

philippjfr:
Ready to review/merge.

jlstevens:
Looks good. Testing to ensure no new clashes are introduced is a good thing and the remaining changes seem sensible. We will want to deprecate the aliases and I think a new issue will be created to remind us to do that. Merging.

PR #2607, opened 2018-04-23
Handled signal interrupt in bokeh apps

philippjfr:
Added callbacks to handle interrupts for bokeh apps in ``BokehRenderer.app`` method. This allows a server to be cleanly stopped on KeyboardInterrupt when working in a Python REPL, making interactive use from a Python interpreter possible.


philippjfr:
Ready to merge.

jlstevens:
Looks good. Merging.

PR #2612, opened 2018-04-24
Add update_display_data event handler

philippjfr:
This PR makes it possible to update existing plotting output using Jupyter ``update_display_data`` messages. While we no longer aim to support or work much with paramnb, this will provide a much cleaner mechanism than it currently uses and can be used more generally to replace displayed plots in the notebook, e.g.:

```python
curve = hv.Curve(np.random.rand(100))
handle = hv.ipython.display(curve, display_id='A')
```

Can be updated with:

```python
scatter = hv.Scatter(curve)
handle.update(scatter)
```

Note: depends on https://github.com/jupyter/notebook/pull/3560, which ~~should be merged soon~~ is now merged and will make it into the imminent 5.5 release, but this approach is already supported by our JupyterLab extension.

### DIFF
--- a/holoviews/plotting/widgets/widgets.js
+++ b/holoviews/plotting/widgets/widgets.js
@@ -596,9 +596,20 @@ function handle_clear_output(event, handle) {
   }
 }
 
+/**
+ * Handle kernel restart event
+ */
 function handle_kernel_cleanup(event, handle) {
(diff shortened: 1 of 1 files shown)


jlstevens:
Seems fine to me. Merging.

PR #2617, opened 2018-04-25
Added validation for Image bounds

philippjfr:
Previously the Image element would allow passing in bounds which did not match the coordinates that were declared. This adds validation ensuring that if both bounds and coordinates are passed they are appropriately validated.

- [x] Add unit test
- [x] Fixes https://github.com/ioam/holoviews/issues/2605

### DIFF
--- a/holoviews/element/raster.py
+++ b/holoviews/element/raster.py
@@ -217,7 +217,7 @@ class Image(Dataset, Raster, SheetCoordinateSystem):
     bounds = param.ClassSelector(class_=BoundingRegion, default=BoundingBox(), doc="""
        The bounding region in sheet coordinates containing the data.""")
 
-    datatype = param.List(default=['image', 'grid', 'xarray', 'cube', 'dataframe', 'dictionary'])
+    datatype = param.List(default=['grid', 'xarray', 'image', 'cube', 'dataframe', 'dictionary'])
 
     group = param.String(default='Image', constant=True)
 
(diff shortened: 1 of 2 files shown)


jlstevens:
Makes sense and I don't *think* there is any scope for problems regarding floating point precision matching issues...

philippjfr:
>Makes sense and I don't think there is any scope for problems regarding floating point precision matching issues...

Seems like that is not the case, I think this is too hairy to attempt right now. There are certain cases where the bounds indeed differ slightly, which would be fine but there are some cases using the ImageInterface where it is not and ``np.isclose`` does not work on datetimes.

The temporary solution I'm going to suggest is threefold:

1. I'm going to validate all interfaces except the ``ImageInterface``
2. I'm not going to validate datetime types for the time being because I don't currently have an implementation akin to np.isclose
3. I'm going to de-prioritize the ImageInterface (which imo is the correct thing anyway), if you pass in a tuple of coordinates along with the array it should not throw away the coordinates and use bounds instead

This addresses the vast majority of cases but I'll file an issue to follow up later.

jlstevens:
Steps 1 and 2 are sensible. Although I agree with step 3 as well, that is the riskiest step as it might have some impact on backwards compatibility.

philippjfr:
I've now also addressed point 1, leaving just 2 and 3. I'd be okay postponing the change to the precedence order and merge it once 1.10.2 is out to give it some of testing. I do consider it a bug though and would prefer to switch it before 1.11.

philippjfr:
This PR suffers from this issue: https://github.com/ioam/holoviews/issues/2619 forcing me to update the test data.

philippjfr:
Ready to review/merge, I'll open a new issue about dropping the precedence of the image interface once this is merged.

jlstevens:
Looks good and I am glad we aren't changing the precedence just yet.

> I'll open a new issue about dropping the precedence of the image interface once this is merged.

Please do. Merging.

Issue #2576, opened 2018-04-19
color cycle not working with string format of hv.opts


scaine1:
I dont seem to be able to get color cycles working when supplying the string format to hv.opts
Using the dictionary format works fine.
When using the string format the following warning is given

Warning: Ignoring keyword pair that fails to evaluate: 'color=cycle'

Example code below.

```python
import numpy as np
import holoviews as hv

hv.extension('bokeh')
renderer = hv.renderer('bokeh')

def curve(val):
    xs = range(-10,11)
    ys = [val-x**2 for x in xs]
    return hv.Curve((xs, ys))

curves = hv.Overlay([curve(x) for x in range(0, 100, 10)])

cycle = hv.Cycle('BuGn')
hv.opts("Curve (color=cycle)")
renderer.save(curves, 'string_opts')

hv.opts({'Curve': {'style': dict(color=cycle())}})
renderer.save(curves, 'dict_opts')

```

philippjfr:
Not that in the notebook this does work:

```python
cycle = hv.Cycle()
%opts Curve (color=cycle)
```

So there must be some difference in the scope in which the string format is evaluated in the two cases.

philippjfr:
This is definitely a more general issue, if you look at the gallery example of filled_contours in the GeoViews gallery you'll see the thumbnail is not respecting the specified ``crs`` https://ioam-docs.github.io/geoviews-master/gallery/index.html, presumably because it's not available in whatever scope it's evaluated.

PR #2627, opened 2018-04-26
Fixed namespace issue in opts utility

jlstevens:
Simple fix to address #2576.

In the ipython module where the magic is defined, ``self.shell.user_ns`` is used as the namespace passed to the parser which is safe as the entire subpackage assumes you are in an IPython environment. In this PR, the IPython namespace is used if available, otherwise it falls back to ``globals`` which should be fine outside of an IPython context.

### DIFF
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -85,7 +85,9 @@ class opts(param.ParameterizedFunction):
 
         if isinstance(options, basestring):
             from .parser import OptsSpec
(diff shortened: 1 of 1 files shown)


philippjfr:
Looks good, should I merge?

jlstevens:
I think so. If there are any more issues I expect they would be subtle things to do with namespaces - this new behavior should be correct in the vast majority of cases.

PR #2629, opened 2018-04-27
Updates opts to options in documentation

philippjfr:
Updates existing uses of the ``.opts`` method to the ``.options`` method, significantly simplifying many examples.


jlstevens:
@philippjfr Grepping ``/examples`` for ``.opts(``, the only notebook left with an inappropriate ``.opts`` is ``sri_model.ipynb`` which fails at the end with an exception. As far as I can tell, all the other notebooks are now updated.

Not sure why the tests are failing. I see some display test failures but these example notebooks aren't tested and I haven't touched them here. Transient errors maybe?

jlstevens:
@philippjfr Ready to merge? I think that is all the ``.opts`` updated where required.

philippjfr:
We should figure out what's going on with https://github.com/ioam/holoviews/issues/2633 first.

PR #2630, opened 2018-04-27
Fix for empty options method call

philippjfr:
When calling the ``.options`` method without any arguments it would previously be an error. This PR ensures that condition is handled appropriately.


jlstevens:
Looks good to me. Merging.

jlstevens:
Unless you want to add a unit test first?

philippjfr:
Thanks, I'll go ahead and merge.

PR #2636, opened 2018-04-30
Fixed bug inverting matplotlib heatmap

philippjfr:
Matplotlib handles inverting the data automatically if an axis is inverted which means that the code to invert the HeatMap data was actually undoing the inversion.

- [x] Fixed https://github.com/ioam/holoviews/issues/2633

### DIFF
--- a/holoviews/element/comparison.py
+++ b/holoviews/element/comparison.py
@@ -115,7 +115,8 @@ class Comparison(ComparisonInterface):
         cls.equality_type_funcs[OrderedDict] =  cls.compare_dictionaries
 
         # Numpy array comparison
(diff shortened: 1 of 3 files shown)


jbednar:
Good catch.  Do you know if mpl has *always* done that, or if it's a change in some more recent version?

philippjfr:
>Do you know if mpl has always done that

I think it's always done that, no other plot classes implement this inversion, so I think it was accidentally adding while I added support for it in the bokeh backend.

jlstevens:
Glad this is fixed although I worry about how many people might have been affected by this without realizing it.

jlstevens:
Looks good. I'll merge when the tests pass.

PR #2639, opened 2018-04-30
Pinned to param 1.6.1 or greater

jlstevens:
This will make sure the latest version of ``Version`` is used which is friendlier when using ``python setup.py develop`` (among other relatively minor, but useful changes).


jlstevens:
The push build has passed and the pr build should too. Merging.

PR #2643, opened 2018-05-02
Fixed bug disabling axis labels with labelled option

philippjfr:
The ``labelled`` plot option allows disabling axis labels, however in bokeh it is being ignored by the ``_update_plot`` method. This fixes the issue.

- [x] Add unit tests


philippjfr:
@jlstevens Ready to merge. This one is pretty unfortunate and may prompt a 1.10.3 release (along with some other fixes I haven't created PRs for yet).

jlstevens:
Is this a regression or just something that wasn't properly supported when it should have been? I'm fine with a 1.10.3 release though and the changes here look good so I'll merge.

philippjfr:
>Is this a regression or just something that wasn't properly supported when it should have been?

I think it's a regression, but I think it's one that snuck in at some point in the 1.9.x series (i.e. I don't think it's not a new bug in 1.10).

PR #2646, opened 2018-05-04
Optimize display of large DynamicMap parameter space

philippjfr:
As described in https://github.com/ioam/holoviews/issues/2645, a DynamicMap with a large parameter space can be very slow to initialize, this is because we have started to process its keys more consistently but did not take into account the fact that the DynamicMap parameter spaces can be much larger. This now avoids a number of computations that are not necessary and optimized some others:

```python
def test(t, z, cmap, vmin, vmax):
    return hv.Image(np.random.rand(100, 100))
    
dmap = hv.DynamicMap(test, kdims=['t','z', 'cmap', 'vmin','vmax']).redim.values(t=range(10), z=range(10) ,cmap=['viridis','Greys'], vmin=range(100), vmax=range(100,200))
```

This DynamicMap defines a parameter space with 2 million keys, previously this would take 52 seconds to display, with this PR this drops to 0.5 seconds.

bokeh_plot

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -1329,7 +1329,8 @@ def drop_streams(streams, kdims, keys):
     inds, dims = zip(*[(ind, kdim) for ind, kdim in enumerate(kdims)
                        if kdim not in stream_params])
     get = itemgetter(*inds)
-    return dims, [get(k) for k in keys]
+    keys = (get(k) for k in keys)
+    return dims, ([wrap_tuple(k) for k in keys] if len(inds) == 1 else list(keys))
 
 
 def itervalues(obj):
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -11,6 +11,7 @@ from collections import defaultdict
 from functools import partial
 from contextlib import contextmanager
 from distutils.version import LooseVersion
+from operator import itemgetter
 
 from threading import Thread, Event
 import numpy as np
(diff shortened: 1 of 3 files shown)


philippjfr:
@jlstevens Another good reason to get a 1.10.3 release out.

holoviews/core/util.py
```diff
@@ -1327,7 +1328,9 @@ def drop_streams(streams, kdims, keys):
     stream_params = stream_parameters(streams)
     inds, dims = zip(*[(ind, kdim) for ind, kdim in enumerate(kdims)
                        if kdim not in stream_params])
-    return dims, [tuple(wrap_tuple(key)[ind] for ind in inds) for key in keys]
+    get = itemgetter(*inds)
```
jlstevens: Don't often see a use for ``itemgetter`` and I'm trying to think whether you really need it here. I don't immediately have a better suggestion so it is probably ok...

jlstevens:
I've made one comment and I'm not sure whether you are done with it, but at a glance this PR seems fine.

holoviews/core/util.py
philippjfr: itemgetter is about 100x faster than the previous tuple comprehension.

jlstevens: That seems to be a good reason. Could you put in a comment to that effect?

philippjfr: Sure, that said itemgetter is used frequently throughout utils, so I'm not sure this one deserves a special note.

philippjfr:
Ready to merge.

jlstevens:
Looks good. Merging.

PR #2653, opened 2018-05-04
Fixed unicode bug in XArrayInterface

philippjfr:
Fixes bug in XArrayInterface handling unicode dimensions in python2.


jlstevens:
Looks good. Merging.

PR #2654, opened 2018-05-04
Sort dynamic dimension values in widget constructor

philippjfr:
Fixes two separate bugs caused by Dimension.values potentially being unsorted while the slider widget always expects sorted values.

- [x] Fixes https://github.com/ioam/holoviews/issues/2641
- [x] Added unit tests

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -807,8 +807,13 @@ class DynamicMap(HoloMap):
         for kdim in self.kdims:
             if str(kdim) in stream_params:
                 key.append(None)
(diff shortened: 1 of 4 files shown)


philippjfr:
Ready to review/merge when tests pass.

jlstevens:
Looks good. One comment is that ``values`` in ``Dimension`` acts more like a set than a list but it is probably too late to change that. Merging.

philippjfr:
Only when the values are numeric, the explicit ordering is respected for dropdowns.

PR #2657, opened 2018-05-06
Made Div signature consistent

philippjfr:
As discussed in https://github.com/ioam/holoviews/issues/2656 the ``Div`` element signature differs from other elements in that the ``data`` argument is named ``html``. This is inconsistent and as highlighted by that issue it's surprising. Since 1.10 hasn't been out long @jbednar and I agreed that it would be better to make it consistent now.

### DIFF
--- a/holoviews/element/annotation.py
+++ b/holoviews/element/annotation.py
@@ -257,16 +257,17 @@ class Text(Annotation):
 
 class Div(Element):
     """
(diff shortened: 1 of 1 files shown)


philippjfr:
Ready to review/merge if you agree @jlstevens

jlstevens:
Yes, I am happy to see this made consistent. Merging.

PR #2663, opened 2018-05-07
Ensure PolyDraw data has consistent keys on initialization

philippjfr:
Previously the PolyDraw stream renamed the keys in its .data dictionary, but only after initialization. This is inconsistent in two ways, first of all other draw/edit streams give the data consistent keys, such as x0/x1/y0/y1 in the case of ``BoxEdit`` or 'xs' and 'ys' in the case of ``PolyEdit``, while this renamed the keys to match the dimensions. We already offer a ``.element`` property on the stream, which transforms the columns so it's preferable to give the stream.data consistent naming.

- [x] Fixes https://github.com/ioam/holoviews/issues/2650


jlstevens:
Looks good. Merging.

PR #2664, opened 2018-05-07
GridSpace and GridPlot fixes

philippjfr:
Fixes three bugs related to grids:

*  The outer axes label ordering of the grid (which was in some cases wrong)
*  Indexing on a GridSpace with non-numeric keys
* Disable zooming on the GridPlot outer axes


jlstevens:
Looks good though maybe you would like to add a unit test for indexing with non-numeric keys?

philippjfr:
Probably for the others too.

jlstevens:
Tests are green. Merging.

PR #2665, opened 2018-05-08
Fixed PlotReset stream parameter

philippjfr:
The reset parameter on the PlotReset stream was shadowing the reset method on Streams, causing errors when the parameter was reset to None on transient events.

### DIFF
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -750,7 +750,7 @@ class PlotReset(LinkedStream):
     A stream signalling when a plot reset event has been triggered.
     """
 
-    reset = param.Boolean(default=False, constant=True, doc="""
+    resetting = param.Boolean(default=False, constant=True, doc="""
         Whether a reset event is being signalled.""")
 
     def __init__(self, *args, **params):


jlstevens:
Some other ideas for the new name:  ``plotreset`` or ``reset_event``?

philippjfr:
Out of the three I much prefer ``resetting``, it implies an action/event and is fairly short.

jlstevens:
I'm just brainstorming to see if we can come up with something better. As it is a boolean, I'm thinking``isreset`` or ``hasreset``.

My main issue with ``resetting`` is that it sounds like it is in the process of resetting which I don't think is strictly true: the plot has *finished* resetting and this boolean is informing the user code of this fact.

If we can't find something better, I agree we can stick with ``resetting`` though.

philippjfr:
>My main issue with resetting is that it sounds like it is in the process of resetting which I don't think is strictly true: the plot has finished resetting and this boolean is informing the user code of this fact.

I'd argue whatever action you want to trigger on reset becomes part of the reset so it is in fact ``resetting``.

jlstevens:
Would be good to add an example to the streams reference gallery. That can be done is a separate PR though, merging.

PR #2666, opened 2018-05-08
Allow changes to columns in bokeh CDS

philippjfr:
This PR eliminates a long-standing limitation in updating plots with elements where the dimensions change. Specifically the code that updates a CDS now checks whether the data needs to be replaced completely, which occurs in one of two conditions 1) there is a new column which does not match the length of old columns or 2) there is an old column which is not being updated which does not match the updated columns in length.

Here are some examples of what this allows (note the changes in the axis label and table column header):

curve_update
table_update

Technically we could just always replace the data completely but a) that can be inefficient and b) it breaks ``shared_datasource``. I've double checked shared sources are not affected by this change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,7 +29,7 @@ install:
   - conda info -a
   - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION flake8 scipy=1.0.0 numpy freetype nose pandas=0.22.0 jupyter ipython=5.4.1 param matplotlib=2.1.2 xarray
   - source activate test-environment
-  - conda install -c conda-forge  iris plotly flexx ffmpeg --quiet
+  - conda install -c conda-forge  iris plotly flexx ffmpeg netcdftime --quiet
   - conda install -c bokeh datashader dask bokeh=0.12.15 selenium
   - if [[ "$TRAVIS_PYTHON_VERSION" == "3.4" ]]; then
       conda install python=3.4.3;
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -29,7 +29,7 @@ install:
   - conda info -a
   - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION flake8 scipy=1.0.0 numpy freetype nose pandas=0.22.0 jupyter ipython=5.4.1 param matplotlib=2.1.2 xarray
   - source activate test-environment
-  - conda install -c conda-forge  iris plotly flexx ffmpeg --quiet
+  - conda install -c conda-forge  iris plotly flexx ffmpeg netcdftime --quiet
```
philippjfr: Unrelated to this PR but apparently required to get tests passing again.

holoviews/plotting/bokeh/plot.py
```diff
@@ -225,6 +225,18 @@ def _update_datasource(self, source, data):
...
+            return
+
+        # Determine if the CDS.data requires a full update or simply needs
+        # to be updated, this is done by determining whether newly added
+        # or not updated columns have the same length
+        new_length = [len(v) for v in data.values() if isinstance(v, (list, np.ndarray))]
+        length = [len(v) for v in source.data.values() if isinstance(v, (list, np.ndarray))]
+        not_updated = [k for k in source.data if k not in data]
+        new = [k for k in data if k not in source.data]
+        if ((not_updated and new_length and any(len(source.data[n]) != new_length[0] for n in not_updated))
```
jlstevens: Not sure you need the initial ``not_updated`` as ``any([])`` evaluates to false.

jlstevens: I think that is also true for the check against ``new``.

PR #2668, opened 2018-05-08
Fixes bug sampling 2D Datasets

philippjfr:
As the title says, fixes ``sample`` on non-Image 2D data types.

- [x] Fixes https://github.com/ioam/holoviews/issues/2667
- [x] Adds unit test

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -416,7 +416,8 @@ class Dataset(Element):
             if np.isscalar(selection):
                 selection = [samples[0]+(selection,)]
             else:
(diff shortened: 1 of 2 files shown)


jlstevens:
> ... on non-Image 2D data types.

That would be ``QuadMesh`` and what else? I assume that doesn't include  ``Raster`` as that isn't a ``Dataset`` yet?

Anyway, looks good and I'll merge once the tests pass.

philippjfr:
>That would be QuadMesh and what else? I assume that doesn't include Raster as that isn't a Dataset yet?

QuadMesh and Dataset itself mostly.

PR #2669, opened 2018-05-08
Added check to ensure empty subplot is not added

philippjfr:
In 1.10 we started supporting dynamically adding subplots, however the code that adds them did not check whether a subplot was actually created correctly, potentially adding Nones to the subplots dictionary, causing downstream issues.


jlstevens:
Looks good. I'll merge once the tests are green.

PR #2674, opened 2018-05-11
Fixed undefined variable bug in XArrayInterface

philippjfr:
Fixed undefined variable bug when slicing with tuples and added unit tests for grid slicing.

- [x] Add unit tests


philippjfr:
Ready to merge, test data was updated due to previous issues (so we can ignore the failing push build).

jlstevens:
I restarted the push build anyway and it is now green. Merging.

PR #2675, opened 2018-05-11
Added restructured text releases page to link to blog

jlstevens:
Currently it contains the CHANGELOG and looks like this:

image

It is mostly the CHANGELOG converted to rst from pandoc. There was some annoyance when backticks appear in markdown links (I made comments in the file about both these issues).

My two remaining concerns:

* Releases hasn't appeared in the side bar for me, despite adding an entry to the TOC (how is this different from the entry for the 'about' page?). **Note**: the other entries are missing as I deleted a bunch of rst files before building, so that much is expected.
* The link to the blog is rather underwhelming. There needs to be a way to really encourage people to click through to the blog. Another image maybe?

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -76,6 +76,7 @@ html_context = {
         ('User Guide', 'user_guide/index'),
         ('Gallery', 'gallery/index'),
         ('Reference Gallery', 'reference/index'),
+        ('Releases', 'releases'),
         ('API', 'Reference_Manual/index'),
         ('FAQ', 'FAQ')
     ),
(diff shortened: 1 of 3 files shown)


jlstevens:
Maybe to address the fact that the blog link doesn't stick out, we could just have an image link showing a screenshot of the top of the blog page. e.g:



As long as we stick to this format in the blog (date shown, HoloViews X.X.X as the tile and an initial collage), I think this could work to drive traffic to the blog page.

philippjfr:
>Releases hasn't appeared in the side bar for me, despite adding an entry to the TOC

Worked fine for me, see http://dev.holoviews.org/FAQ.html (note that I accidentally had Releases in there twice and both showed up)

>The link to the blog is rather underwhelming.

I've replaced the little snippet of text which imo was somewhat redundant with section headers, I think that makes the link a bit more prominent but I'd be okay adding something else.

jlstevens:
Your edit makes sense and I'm happy to see it merged now. We can always improve it later if we agree on some change we all think is worthwhile.

PR #2678, opened 2018-05-12
Enable Travis checking for Python linting rules we are currently not …

jbampton:
…violating.


jbampton:
Well the flake 8 tests are passing in both Python 2 and 3, but in Python 3 there is an error:

The command "cd doc/nbpublisher; chmod +x test_notebooks.py; QT_QPA_PLATFORM='offscreen' BOKEH_DEV=True ./test_notebooks.py" exited with 1.

I thought some basic checks would stop regressions from happening.  For example I enabled W291 Trailing whitespace check as I previously fixed the trailing whitespace in Holoviews.

philippjfr:
Sorry for never replying here, the linting rules will have to be discussed a bit first and since this PR has gone stale I'm going to close.

PR #2680, opened 2018-05-12
Allow setting alpha on bokeh Raster/Image/RGB

philippjfr:
Recently bokeh added support for setting a global alpha for image and rgb plots (I think this was in 0.12.15). This PR simply defines the style option so the option can be set.


philippjfr:
Ready to merge.

jlstevens:
Great! I'm a little surprised that such an obvious feature hasn't existed till now but it is good to support it now it does.

philippjfr:
>I'm a little surprised that such an obvious feature hasn't existed till now but it is good to support it now it does.

matplotlib doesn't support this at all so maybe not so surprising.

PR #2681, opened 2018-05-12
Correctly check if hover data should be added

philippjfr:
This PR addresses a class of bugs in the bokeh backend where the data required to display hover information for the element was added if any subplot of the overlay enabled hover information even if the hover for a particular element was not actually displayed. Rather than checking if the bokeh plot object has a HoverTool, it now checks whether the hover tool has been added to the holoviews plot handles, which is the correct way of checking whether hover was enabled.


jlstevens:
Very happy with this PR: it fixes the bug while making the code simpler and easier to read. Merging.

PR #2683, opened 2018-05-14
OverlayPlot passes projection to subplots

philippjfr:
Very simple change which ensures that an OverlayPlot forwards the projection to all subplots. This was not needed for matplotlib since the OverlayPlot (or Layout/GridPlot) creates the axis with the appropriate projection but is needed for bokeh where each plot needs to know its projection independently.

- [x] Adds unit tests


jlstevens:
I know this PR is probably still WIP but the change does make sense and currently looks good to me. I assume you are planning to add unit tests at which point I would be happy to merge.

philippjfr:
Codewise the PR is complete, but sure I'd be happy to add a few unit tests.

philippjfr:
Ready to merge.

PR #2685, opened 2018-05-14
Fixed bug initializing empty data in combined dictionary format

philippjfr:
The DictInterface allows combining two columns into one array, however when this array is empty it currently fails and raises an error. This PR ensures that this works:

```python
Dataset({('x', 'y'): []}, kdims=['x', 'y'])
```

- [x] Adds unit test


jlstevens:
Tests are green. Merging.

PR #2686, opened 2018-05-14
Added CHANGELOG for 1.10.4

philippjfr:


jlstevens:
Why would coverage decrease adding a changelog entry? It isn't code!

philippjfr:
>Why would coverage decrease adding a changelog entry? It isn't code!

One of the tests failed with a transient, which triggered the coverage drop.

jlstevens:
> One of the tests failed with a transient, which triggered the coverage drop.

I suppose I would prefer if coverage was only updated when the tests pass.

jlstevens:
Anyway, all the tests are green now. Merging.

PR #2700, opened 2018-05-19
Fixed ResetCallback

philippjfr:
The ResetCallback for the PlotReset stream was still incorrectly returning {'reset': True}, rather than the updated ``{'resetting': True}``.

- [x] Adds unit test


philippjfr:
Ready to merge.

PR #2701, opened 2018-05-19
Fixes for datashader reduction error

philippjfr:
- [x] Fixes bug in datashade/rasterize operation error handling
- [x] Defines missing 'sum': rd.sum reduction


jbednar:
Whoa; how did that ever work before?  Needs a new test?

philippjfr:
What do you mean? This was not a critical error, the error message was slightly wrong, and you could not declare the sum reduction as a string, that's about it.

philippjfr:
Ready to merge.

jbednar:
Ah, I didn't see that it was only in an error message; that's why it didn't matter much before.

PR #2706, opened 2018-05-20
Fix bug dropping streams in DynamicMap.groupby

### LINKED ISSUE #2705: DynamicMap drops streams when faceting methods are called
When constructing a DynamicMap with kdims and streams and calling a faceting method like ``.grid``, ``.layout`` or ``.overlay`` on it, it will simply drop the streams if no other key dimensions are left. 

```
def plot_function(mydim, data):
    return hv.Curve(data[data.channel==mydim], 'x', 'y')

my_buffer = hv.streams.Buffer(data=pd.DataFrame({'x': [], 'y': [], 'channel': np.array([], dtype='S1')}))
hv.DynamicMap(plot_function, streams=[my_buffer], kdims=[hv.Dimension('mydim', values=['A', 'B', 'C'])]).layout(['mydim'])
```

If we print the repr we can see what's going on, it has simply unpacked the DynamicMap:

```
:NdLayout   [mydim]
   :Curve   [x]   (y)
```

There needs to be a conditional to ensure that if there are any streams defined, it should not unpack the DynamicMap.

philippjfr:
- [x] Fixes #2705 
- [x] Adds unit test

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -1378,7 +1378,7 @@ class DynamicMap(HoloMap):
                         inner_dims = zip(inner_kdims, util.wrap_tuple(key))
                         inner_vals = [(d.name, k) for d, k in inner_dims]
                         return self.select(**dict(outer_vals+inner_vals)).last
-                    if inner_kdims:
+                    if inner_kdims or self.streams:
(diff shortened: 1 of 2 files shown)


philippjfr:
Ready to merge.

jlstevens:
Looks good, merging.

PR #2715, opened 2018-05-22
Add a isfinite utility to use throughout code

philippjfr:
Throughout our code we check whether a value is finite using the np.isfinite utility, in almost all of those cases this usage is accompanied by additional checks whether the value is a None, datetime or string type. This also caused various bugs when we did not check for these extra types (e.g. https://github.com/ioam/holoviews/issues/2708). I've therefore added a ``isfinite`` utility that accounts for these corner cases.

- [x] Fixes https://github.com/ioam/holoviews/issues/2708
- [x] Fixes https://github.com/ioam/holoviews/issues/2712
- [x] Adds unit tests

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -46,7 +46,7 @@ try:
         from pandas.core.dtypes.dtypes import DatetimeTZDtypeType
     else:
         from pandas.types.dtypes import DatetimeTZDtypeType
-    datetime_types = datetime_types + (pd.Timestamp, DatetimeTZDtypeType)
+    datetime_types = datetime_types + (pd.Timestamp, DatetimeTZDtypeType, pd.Period)
     timedelta_types = timedelta_types + (pd.Timedelta,)
 except ImportError:
     pd = None
@@ -703,7 +703,7 @@ def isfinite(val):
         if val.dtype.kind in 'USMO':
             return np.ones_like(val, dtype=bool)
         return np.isfinite(val)
-    elif isinstance(val, datetime_types+(basestring,)):
+    elif isinstance(val, datetime_types+timedelta_types+(basestring,)):
         return True
     return np.isfinite(val)
 
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -692,6 +692,22 @@ def isnumeric(val):
         return False
 
 
+def isfinite(val):
+    """
+    Helper function to determine if scalar or array value is finite extending
+    np.isfinite with support for None, string, datetime types.
+    """
+    if val is None:
+        return False
+    elif isinstance(val, np.ndarray):
+        if val.dtype.kind in 'USMO':
+            return np.ones_like(val, dtype=bool)
+        return np.isfinite(val)
+    elif isinstance(val, datetime_types+(basestring,)):
+        return True
+    return np.isfinite(val)
+
+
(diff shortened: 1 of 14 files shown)


jlstevens:
Makes sense to me and the utility looks good. Happy to merge when the tests pass.

holoviews/core/util.py
```diff
@@ -703,7 +705,7 @@ def isfinite(val):
         if val.dtype.kind in 'USMO':
             return np.ones_like(val, dtype=bool)
         return np.isfinite(val)
-    elif isinstance(val, datetime_types+(basestring,)):
+    elif isinstance(val, datetime_types+timedelta_types+(basestring,)):
```
jlstevens: Is this really an ``isfinite`` fix?

philippjfr: Yes, it's in the ``isfinite`` function isn't it?

jlstevens: Sorry, the Github diff was misleading! Isn't there another recent issue about ``pd.Period``?

philippjfr: Yes, but that's also added to datetime_types to exclude it from isfinite checks.

philippjfr:
Ready now.

jlstevens:
Looks good. Merging.

AlbertDeFusco:
Sorry to revive this, but perhaps `pd.isna()` would help since it directly supports not-a-time `NaT`.

https://pandas.pydata.org/pandas-docs/stable/generated/pandas.isna.html

philippjfr:
>Sorry to revive this, but perhaps pd.isna() would help since it directly supports not-a-time NaT.

Thanks for bringing that up, we do not have a core dependency on pandas currently so we can't just use ``pd.isna`` but we certainly should be supporting ``NaT``.

AlbertDeFusco:
NumPy 1.14 has `np.isnat`, but that only really works for `np.datetime64` and `np.timedelta64` and not the Pandas types, but those implement `.to_datetime64()`.

philippjfr:
Thanks, I'm digging through the logic now, here's roughly what I'm coming up with:

```python
def isnat(val):
    """
    Checks if the scalar value is a NaT. Should only be called on datetimelike objects.
    """
    if isinstance(val, (np.datetime64, np.timedelta64)):
        if numpy_version >= '1.13':
            return np.isnat(val)
        else:
                warnings.simplefilter("ignore")
                return nat == np.datetime64('NaT')
    elif pd and isinstance(val, pandas_datetime_types+pandas_timedelta_types):
        return pd.isna(val)
    else:
        return False


def isfinite(val):
    """
    Helper function to determine if scalar or array value is finite extending
    np.isfinite with support for None, string, datetime types.
    """
    val = asarray(val, strict=False)

    if val is None:
        return False
    elif isinstance(val, np.ndarray):
        if val.dtype.kind == 'M':
            return pd.isna(val) if pd else np.array([isnat(v) for v in val], dtype=bool)
        elif val.dtype.kind == 'O':
            return np.array([isfinite(v) for v in val], dtype=bool)
        elif val.dtype.kind in 'US':
            return np.ones_like(val, dtype=bool)
        return np.isfinite(val)
    elif isinstance(val, datetime_types+timedelta_types):
        return not isnat(val)
    elif isinstance(val, basestring):
        return True
    return np.isfinite(val)
```

PR #2716, opened 2018-05-22
Handle NaT types in isfinite utility

philippjfr:
As mentioned in https://github.com/ioam/holoviews/pull/2715 we currently do not handle NaT types at all. Unfortunately this is a bit of a pain since it has only been consistently supported in NumPy >= 1.13. Additionally this PR improves the ``isfinite`` utility to work correctly on various arraylike types including pandas index and series types.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -37,17 +37,29 @@ except AttributeError:
     _getargspec = inspect.getargspec
     get_keywords = operator.attrgetter('keywords')
 
+numpy_version = LooseVersion(np.__version__)
+
 datetime_types = (np.datetime64, dt.datetime, dt.date)
 timedelta_types = (np.timedelta64, dt.timedelta,)
+arraylike_types = (np.ndarray,)
 
 try:
     import pandas as pd
-    if LooseVersion(pd.__version__) > '0.20.0':
+    pandas_version = LooseVersion(pd.__version__)
+    if pandas_version > '0.20.0':
         from pandas.core.dtypes.dtypes import DatetimeTZDtypeType
+        from pandas.core.dtypes.generic import ABCSeries, ABCIndexClass
     else:
         from pandas.types.dtypes import DatetimeTZDtypeType
-    datetime_types = datetime_types + (pd.Timestamp, DatetimeTZDtypeType, pd.Period)
-    timedelta_types = timedelta_types + (pd.Timedelta,)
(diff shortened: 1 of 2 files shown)


holoviews/core/util.py
```diff
@@ -692,18 +704,69 @@ def isnumeric(val):
...
+def isnat(val):
+    """
+    Checks if the value is a NaT. Should only be called on datetimelike objects.
+    """
+    if isinstance(val, (np.datetime64, np.timedelta64)):
+        if numpy_version >= '1.13':
+            return np.isnat(val)
+        else:
+            return val.view('i8') == nat_as_integer
+    elif pd and val is pd.NaT:
```
jbednar: Has pd.NaT always been available, or only since a certain Pandas release?  If the latter, is that Pandas release a minimum requirement for HoloViews?

holoviews/core/util.py
```diff
@@ -692,18 +704,69 @@ def isnumeric(val):
...
         return False
     elif isinstance(val, np.ndarray):
-        if val.dtype.kind in 'USMO':
+        if val.dtype.kind == 'M':
+            if pd:
+                return ~pd.isna(val)
+            elif numpy_version >= '1.13':
+                return ~np.isnat(val)
+            else:
+                return val.view('i8') != nat_as_integer
```
jbednar: And this logic can't be folded into isnat above?

holoviews/core/util.py
philippjfr: I see StackOverFlow issues about pd.NaT going back to 2014, so that doesn't seem a major worry.

holoviews/core/util.py
philippjfr: Done.

holoviews/core/util.py
jbednar: Ok.

holoviews/core/util.py
jbednar: Looks good; thanks.

holoviews/core/util.py
philippjfr: That said we should make a list of optional dependencies and determine minimum versions for them.

jlstevens:
Looks good although having to do those version checks for both numpy and pandas is rather annoying. Ready to merge?

philippjfr:
Yes, would be good to merge now since this is also needed to make https://github.com/ioam/holoviews/pull/2719 work properly.

jlstevens:
Merging.

PR #2719, opened 2018-05-23
Support datetimes in histogram operation

philippjfr:
This PR makes it possible to generate a histogram from datetime data as requested in https://github.com/ioam/holoviews/issues/2713 and https://github.com/pyviz/holoplot/issues/12. This is consistent with ``plt.hist`` which also handle datetimes and can be useful to determine the sampling frequency of a dataset.

- [x] Implements https://github.com/ioam/holoviews/issues/2713
- [x] Adds unit tests

### DIFF
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -204,7 +204,7 @@ class GridInterface(DictInterface):
         if edges and not isedges:
             data = cls._infer_interval_breaks(data)
         elif not edges and isedges:
-            data = np.convolve(data, [0.5, 0.5], 'valid')
+            data = data[:-1] + np.diff(data)/2.
         return data
 
 
--- a/holoviews/plotting/mpl/chart.py
+++ b/holoviews/plotting/mpl/chart.py
@@ -261,8 +261,8 @@ class HistogramPlot(ChartPlot):
 
         # Get plot ranges and values
         dims = hist.dimensions()[:2]
-        edges, hvals, widths, lims, datetime = self._process_hist(hist)
-        if datetime and not dims[0].value_format:
+        edges, hvals, widths, lims, isdatetime = self._process_hist(hist)
+        if isdatetime and not dims[0].value_format:
             dt_format = Dimension.type_formatters[np.datetime64]
             dims[0] = dims[0](value_format=DateFormatter(dt_format))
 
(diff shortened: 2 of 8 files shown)


philippjfr:

holoviews/core/data/grid.py
```diff
@@ -204,7 +204,7 @@ def coords(cls, dataset, dim, ordered=False, expanded=False, edges=False):
         if edges and not isedges:
             data = cls._infer_interval_breaks(data)
         elif not edges and isedges:
-            data = np.convolve(data, [0.5, 0.5], 'valid')
+            data = data[:-1] + np.diff(data)/2.
```
jlstevens: Much simpler than a weird and confusing ``convolve`` call!

holoviews/plotting/mpl/chart.py
```diff
@@ -288,9 +295,15 @@ def _process_hist(self, hist):
...
+        ylim = hist.range(1)
+        datetime = False
+        if edges.dtype.kind == 'M' or isinstance(edges[0], datetime_types):
+            edges = date2num([v.tolist() if isinstance(v, np.datetime64) else v for v in edges])
+            xlim = date2num([v.tolist() if isinstance(v, np.datetime64) else v for v in xlim])
+            datetime = True
         widths = np.diff(edges)
-        lims = hist.range(0) + hist.range(1)
-        return edges[:-1], hist_vals, widths, lims
+        return edges[:-1], hist_vals, widths, xlim+ylim, datetime
```
jlstevens: Instead of ``True/False`` maybe the more general thing would be to return the type itself? This could allow for other conditional dimension type formatters in future.

If nothing else, grepping for ``datetime`` is a bit of a nightmare so it should be called ``isdatetime`` or similar...

philippjfr:
Done.

holoviews/plotting/mpl/chart.py
```diff
@@ -288,9 +295,15 @@ def _process_hist(self, hist):
...
+        ylim = hist.range(1)
+        isdatetime = False
+        if edges.dtype.kind == 'M' or isinstance(edges[0], datetime_types):
+            edges = date2num([v.tolist() if isinstance(v, np.datetime64) else v for v in edges])
+            xlim = date2num([v.tolist() if isinstance(v, np.datetime64) else v for v in xlim])
+            isdatetime = True
         widths = np.diff(edges)
-        lims = hist.range(0) + hist.range(1)
-        return edges[:-1], hist_vals, widths, lims
+        return edges[:-1], hist_vals, widths, xlim+ylim, isdatetime
```
jlstevens: I can still conceive of wanting support for other numeric types one day (e.g rationals?) but this is fine for now. We can use a general type specifier when we find ourselves needing to do something similar for another type. This this fine for now.

jlstevens:
Looks good. Merging.

PR #2720, opened 2018-05-23
Importing comms machinery from pyviz_comms

jlstevens:
This PR is to update holoviews to use the new [pyviz_comms](https://github.com/pyviz/pyviz_comms) package. Note that the ``builder`` and ``nbpublisher`` submodules in ``doc`` will need updating.


jlstevens:
@philippjfr I added ``pyviz_comms`` to the ``install_requires`` even though it technically isn't part of 'core' holoviews. I think this makes sense as this package has no dependencies and we want everyone to get it.

philippjfr:
Looks good, will merge after we release 1.10.5.

jlstevens:
Now that ``pyviz_comms`` is on conda-forge and will be on defaults in the next hour or so, I would consider including this PR in 1.10.5.

holoviews/plotting/widgets/widgets.js
```diff
@@ -522,6 +522,7 @@ function init_dropdown(id, plot_id, dim, vals, value, next_vals, labels, next_di
 
 if (window.HoloViews === undefined) {
   window.HoloViews = {}
+  window.PyViz = {}
```
philippjfr: Shouldn't ``window.HoloViews`` and ``window.PyViz`` be referencing the same object, rather than two distinct objects?

philippjfr: The jlab extension for example still references ``windows.HoloViews`` so if they are not the same object things will not work.

jlstevens: Good point.

jlstevens:
@philippjfr I've tested it using the ``pyviz_comms`` package from ``defaults`` both with classic notebook and Jupyterlab for both holoviews and parambokeh. It all seems to work fine.

I think it is ready to merge now unless you want to check out this branch to double check it works properly first.

philippjfr:
Will test it before release.

PR #2722, opened 2018-05-24
Fixed bug defining callbacks on synced sources

philippjfr:
After datasources are synced, i.e. shared across a layout any callbacks defined on the datasource is currently not given the correct plot id. When the top level plot syncs the datasources, the callbacks are reinitialized at which point we can directly pass the plot_id to the callback.

- [x] Adds unit test

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -284,13 +284,14 @@ class CustomJSCallback(MessageCallback):
         return code
 
 
-    def get_customjs(self, references):
+    def get_customjs(self, references, plot_id=None):
         """
         Creates a CustomJS callback that will send the requested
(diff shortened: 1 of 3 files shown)


philippjfr:

jlstevens:
Looks good. Is there an issue/description of the problem(s) caused by the old behavior?

philippjfr:
There is not, but I can describe it here, all drawing tools were broken when you tried to apply them to a shared datasource, e.g. all annotators in EarthSim were broken.

jlstevens:
> There is not, but I can describe it here, all drawing tools were broken when you tried to apply them to a shared datasource, e.g. all annotators in EarthSim were broken.

Thanks. Anyway, the changes look fine (just passing ``plot_id`` through mostly) and the tests are green. Merging.

PR #2724, opened 2018-05-24
Ensure vdims are initialized on PointDraw stream

philippjfr:
When plotting a Points object with a PointDraw stream various value dimensions may be defined which are not necessarily defined on the ColumnDataSource (because vdims are usually only sent when the hover tool is requested). This can cause issues downstream because the PointDraw tool won't know to add values along that dimension. This PR ensures that the PointDrawCallback initializes those columns if the plot hasn't already done so.

- [x] Fixes https://github.com/pyviz/EarthSim/issues/87
- [x] Add unit test


philippjfr:

jlstevens:
Looks good though you might want to add a comment above that block as a reminder of why you need this. Otherwise happy to merge.

PR #2726, opened 2018-05-24
Avoid making clone in DynamicMap.options method if requested

philippjfr:
This is a tricky issue, we recently added the ability to use ``.options`` without making a clone, which has worked out great and is used quite frequently (at least by me) since it is very convenient when working with an element that is a stream source. However the ``DynamicMap.options`` method still made clones even when ``clone=False`` was set, this is because our general approach to applying any operation to a DynamicMap is to wrap the DynamicMap in another DynamicMap either by writing a wrapping callback or by using the ``Dynamic`` utility. In this particular case we have to make an exception and wrap the user supplied callback function directly without wrapping the DynamicMap. It's not ideal and doesn't fit the model we use elsewhere but it's the only way to avoid inconsistencies like the one raised in https://github.com/ioam/holoviews/issues/2698.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -961,9 +961,16 @@ class DynamicMap(HoloMap):
         Disabling clone will modify the object inplace.
         """
         from ..util import Dynamic
-        dmap = Dynamic(self, operation=lambda obj, **dynkwargs: obj.options(options, backend,
-                                                                            clone, **kwargs),
+        obj = self if clone else self.clone()
+        dmap = Dynamic(obj, operation=lambda obj, **dynkwargs: obj.options(options, backend,
+                                                                           clone, **kwargs),
                        streams=self.streams, link_inputs=True)
+        if not clone:
+            with util.disable_constant(self):
+                self.callback = dmap.callback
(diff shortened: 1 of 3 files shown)


jlstevens:
I didn't object to the ``clone`` switch in ``.options`` when I saw it, but only because I hadn't thought about this consequence. In retrospect, I now think it was a bad idea.

philippjfr:
I think I can rewrite this in a way that makes me a little happier about this and doesn't break the model of having DynamicMaps build a graph of the operations that were applied to it. I'll simply transfer the original callback onto another DynamicMap, which will become what ``self`` wraps around and be added to its ``inputs``.

jlstevens:
I'll have to see it and think about it some more. At least ``clone=True`` by default so I hope most calls to ``options`` were cloning, avoiding this issue.

philippjfr:
>I'll have to see it and think about it some more. At least clone=True by default so I hope most calls to options were cloning

That's right. I've now rewritten the fix for this, which I'm much happier with.

philippjfr:
I think this is the best approach for this I'm likely going to come up with, so please review.

PR #2728, opened 2018-05-25
Added support for cftime types

philippjfr:
This PR adds support for ``cftime`` types which are a custom time type used by xarray for dates that cannot be stored in a datetime64 type. A small example:

```python
ref_date = '0181-01-01'
ds = xr.DataArray([1, 2, 3], dims=['time'],
                  coords={'time': ('time', [1, 2, 3],
                                   {'units': 'days since %s' % ref_date})}
                  ).to_dataset(name='foo')
with xr.set_options(enable_cftimeindex=True):
    ds = xr.decode_cf(ds)
print(ds)
hv_ds = hv.Dataset(ds)
hv_ds.to(hv.Curve)
```

```
<xarray.Dataset>
Dimensions:  (time: 3)
Coordinates:
  * time     (time) object 0181-01-02 00:00:00 0181-01-03 00:00:00 ...
Data variables:
    foo      (time) int64 ...
```

bokeh_plot

This approach seems to work well but there are a questions I have, which @rabernat and @spencerclark hopefully can answer:

1. Which types does it make sense to support, and do they all support conversion to a timetuple? As far as I could tell this is the list of types ``cftime`` provides:

cftime._cftime.Datetime360Day
cftime._cftime.DatetimeGregorian
cftime._cftime.DatetimeJulian
cftime._cftime.DatetimeNoLeap
cftime._cftime.DatetimeProlepticGregorian

2. Secondly in what context can I expect these types to be used? Will they mostly be used by xarray or should I also handle them elsewhere?

- [x] Addresses https://github.com/ioam/holoviews/issues/2709
- [ ] Add unit tests
- [ ] Handle matplotlib
- [ ] Handle datashader (if possible)

### DIFF
--- a/holoviews/core/data/xarray.py
+++ b/holoviews/core/data/xarray.py
@@ -18,6 +18,12 @@ from ..element import Element
 from .grid import GridInterface
 from .interface import Interface, DataError
 
+try:
+    import cftime
+    util.datetime_types += (cftime._cftime.DatetimeGregorian,)
+except:
+    pass
+
 
 class XArrayInterface(GridInterface):
 
--- a/holoviews/plotting/bokeh/util.py
+++ b/holoviews/plotting/bokeh/util.py
@@ -1,4 +1,4 @@
-import re, time, sys
+import re, time, sys, calendar
 from distutils.version import LooseVersion
 from collections import defaultdict
 import datetime as dt
@@ -617,8 +617,8 @@ def date_to_integer(date):
         date = dt64_to_dt(date)
     elif pd and isinstance(date, pd.Timestamp):
         date = date.to_pydatetime()
-    if isinstance(date, (dt.datetime, dt.date)):
-        dt_int = time.mktime(date.timetuple())*1000
+    if hasattr(date, 'timetuple'):
+        dt_int = calendar.timegm(date.timetuple())*1000
     else:
         raise ValueError('Datetime type not recognized')
     return dt_int
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -26,7 +26,7 @@ from ..util import dynamic_update, process_cmap
 from .plot import BokehPlot, TOOLS
 from .util import (mpl_to_bokeh, get_tab_title,  py2js_tickformatter,
                    rgba_tuple, recursive_model_update, glyph_order,
-                   decode_bytes)
+                   decode_bytes, date_to_integer)
 
 property_prefixes = ['selection', 'nonselection', 'muted', 'hover']
 
@@ -528,6 +528,8 @@ class ElementPlot(BokehPlot, GenericElementPlot):
             if util.isfinite(high):
                 updates['end'] = (axis_range.end, high)
             for k, (old, new) in updates.items():
+                if isinstance(new, util.datetime_types):
+                    new = date_to_integer(new)
                 axis_range.update(**{k:new})
                 if streaming:
                     axis_range.trigger(k, old, new)
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -10,7 +10,8 @@ from bokeh.models.widgets import Panel, Tabs
 from ...core import (OrderedDict, Store, AdjointLayout, NdLayout, Layout,
                      Empty, GridSpace, HoloMap, Element, DynamicMap)
 from ...core.options import SkipRendering
-from ...core.util import basestring, wrap_tuple, unique_iterator, get_method_owner
+from ...core.util import (basestring, wrap_tuple, unique_iterator,
+                          get_method_owner, datetime_types)
 from ...streams import Stream
 from ..plot import (DimensionedPlot, GenericCompositePlot, GenericLayoutPlot,
                     GenericElementPlot, GenericOverlayPlot)
@@ -18,7 +19,7 @@ from ..util import attach_streams, displayable, collate
 from .callbacks import Callback
 from .util import (layout_padding, pad_plots, filter_toolboxes, make_axis,
                    update_shared_sources, empty_plot, decode_bytes,
-                   bokeh_version)
+                   bokeh_version, date_to_integer)
 
 from bokeh.layouts import gridplot
 from bokeh.plotting.helpers import _known_tools as known_tools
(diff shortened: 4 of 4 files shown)


rabernat:
Thanks so much for taking this on! It will make climate modelers very happy.

I think I need to defer to @spencerclark or @shoyer to answer your questions, as they understand the details better.

philippjfr:
I've tried playing with nc-time-axis a bit but I can't actually get it to plot cftime objects with non-standard reference units, e.g. I haven't found any way to plot dates like this:

```
dates = xr.cftime_range(start='0001', periods=24, freq='MS', calendar='noleap')
```

My vote is to merge this without matplotlib support for now.

holoviews/plotting/bokeh/util.py
```diff
@@ -542,10 +543,10 @@ def date_to_integer(date):
...
     if isinstance(date, np.datetime64):
         date = dt64_to_dt(date)
-    if pd and isinstance(date, pd.Timestamp):
-        dt_int = date.timestamp()*1000
-    elif isinstance(date, (dt.datetime, dt.date)):
-        dt_int = time.mktime(date.timetuple())*1000
+    elif pd and isinstance(date, pd.Timestamp):
+        date = date.to_pydatetime()
+    if hasattr(date, 'timetuple'):
+        dt_int = calendar.timegm(date.timetuple())*1000
```
spencerkclark: Seeing this makes me a little uneasy with respect to use with cftime types.  I think the built-in Python `calendar` module always assumes the Gregorian calendar is used; in cftime, different date types operate on different calendars, so the "seconds since 1970-01-01T00:00:00" value could be different for the same time tuple, depending on the calendar.

But perhaps there's good reason for doing things this way (I'm not familiar with how dates are handled in bokeh/holoviews)?

philippjfr: Thanks, definitely need another look at this. Is there a simple way to convert the values to a Gregorian calendar?

spencerkclark: Ah, I see; this is appropriate if that is what you are looking to do.  

However, I'm not sure if this is the ideal approach in general; for instance this will cause there to be extra space (and potentially a tick label) in the place of February 29th, 2000 for a no leap calendar (even though in that calendar that date does not exist).  See [an example in a gist with this branch](https://nbviewer.jupyter.org/gist/spencerkclark/477ccac28db88c54004fd9a4c5f2d4fe).  There are other possible edge cases too (in a 360-day calendar some dates do not exist in a Gregorian calendar, like February 30th).

I think nc-time-axis handles these edge cases properly, because it does not rely on converting dates to the Gregorian calendar.  [See how it handles the example above in this gist](https://nbviewer.jupyter.org/gist/spencerkclark/79e790a6c59c6655c687fade3f00a370) (I've also included an example of how to plot with dates generated from `xarray.cftime_range`).

I understand if it would be a lot more work to solve this in general though.

philippjfr: Thanks again, I'll look at nc_time_axis in a bit more detail, running your notebook is currently giving me an error though, which I had already filled here: https://github.com/SciTools/nc-time-axis/issues/40.

```
ValueError: invalid year provided in cftime.DatetimeNoLeap(0, 11, 27, 1, 12, 0, 4, 3, 331)
```

What version of nc_time_axis and cftime have? I **think** I have the latest version of both:

```
print(cftime.__version__)
print(nc_time_axis.__version__)
1.0.0
1.1.0
```

spencerkclark: Sure thing, thanks for the link to that issue.  In my notebook I was using cftime 1.0.2.1 and nc_time_axis 1.1.0.  I can reproduce your issue if I downgrade cftime to version 1.0.0.

philippjfr: Thanks, I'll try upgrading.

philippjfr: Okay, I think we'll likely have to push ahead with the subtly wrong behavior in bokeh, as a proper solution would require writing custom bokeh models to correctly format dates with different calendars. Would you expect a warning when plotting dates in a non-Gregorian calendar until we've got a proper fix?

holoviews/plotting/mpl/util.py
```diff
@@ -334,3 +337,36 @@ def polygons_to_path_patches(element):
...
+    @classmethod
+    def convert(cls, value, unit, axis):
+        if not nc_axis_available:
+            raise ValueError('In order to display cftime types with '
+                             'matplotlib install the nc_time_axis '
+                             'library using pip or from conda-forge '
+                             'using:\n\tconda install -c conda-forge '
+                             'nc_time_axis')
+        if isinstance(value, cftime_types):
+            value = CalendarDateTime(cftime.datetime(*value.timetuple()[0:6]), value.calendar)
```
spencerkclark: Is there a reason for converting to the generic `cftime.datetime` here and below?  I think you can just pass `value` and `value.calendar` to the `CalendarDateTime` constructor.

holoviews/plotting/bokeh/util.py
```diff
@@ -536,16 +534,53 @@ def append_refresh(dmap):
...
+    return utime.date2num(date)*86400000
+
+
+def date_to_integer(date):
+    """Converts support date types to milliseconds since epoch
+
+    Attempts highest precision conversion of different datetime
+    formats to milliseconds since the epoch (1970-01-01 00:00:00).
+    If datetime is a cftime with a non-standard calendar the
+    caveats describd in cftime_to_timestamp apply.
```
spencerkclark: Typo in "described"

holoviews/plotting/bokeh/util.py
```diff
@@ -536,16 +534,53 @@ def append_refresh(dmap):
...
 
-def date_to_integer(date):
+def cftime_to_timestamp(date):
+    """Converts cftime to timestamp since epoch in milliseconds
+
+    Non-standard calendars (e.g. Julian or no leap calendars)
+    are converted to standard Gregorian calendar. This can cause
+    extra space to be added for dates that don't exist in the original
+    calendar. In order to handle these dates correctly a custom bokeh
+    model with support for other calendars would have to be defind.
```
spencerkclark: Typo in "defined."

holoviews/core/data/xarray.py
```diff
@@ -12,6 +12,19 @@
...
 from .interface import Interface, DataError, dask_array_module
 
+try:
+    import cftime
+    cftime_types = (
+        cftime._cftime.DatetimeGregorian,
+        cftime._cftime.Datetime360Day,
+        cftime._cftime.DatetimeJulian,
+        cftime._cftime.DatetimeNoLeap,
+        cftime._cftime.DatetimeProlepticGregorian
```
spencerkclark: I think you are missing `cftime.DatetimeAllLeap` here.  That said, all of these are subclasses of `cftime.datetime`, so I think you could get away with just using `cftime.datetime` here, rather than enumerating all of the different subclasses (since `cftime_types` is only used for instance checks).

holoviews/plotting/mpl/util.py
philippjfr: Thanks, don't think there was any good reason.

holoviews/core/data/xarray.py
philippjfr: That's much easier thanks.

philippjfr:
>If it's not too much trouble I do think a warning would be helpful in the bokeh case. For example we now warn in xarray whenever this kind of conversion is done.

Thanks, I've adapted the error message from xarray for our purposes, e.g.:

>Converting cftime.datetime from a non-standard calendar (noleap) to a standard calendar for plotting. This may lead to subtle errors in formatting dates, for accurate tick formatting switch to the matplotlib backend.

philippjfr:
@jlstevens This should be ready to review and merge now.

If you're wondering about all the other changes, the dt_to_int utility was all kinds of horrendously wrong with its unit conversions, which was made up for by counter-acting it with various scaling factors in the histogram and datashading operations. The operation now actually converts between time resolutions correctly.

spencerkclark:
Thanks again @philippjfr!

philippjfr:
Thank you for being so helpful @spencerkclark!

PR #2732, opened 2018-05-25
Add baseclass for Overlay and Layout

philippjfr:
As was rightly pointed out in https://github.com/ioam/holoviews/issues/2673 ``Overlay`` currently inherits from ``Layout`` which is not a sensible class hierarchy and also leads ``Overlay`` to inherit a number of methods and behaviors which do not make sense, e.g. ``.cols``, ``.shape`` and ``.grid_items``. This needs pretty thorough testing so it'll have to go in the 1.11 release.


jlstevens:
I agree with the principle of having a base class though I hope we can express the idea better than ``DimensionedTree`` and I also agree this will need a lot of testing.

The core concept is that these are the results of the general compositions operators (``+`` and ``*``) that apply to heterogenous collections of elements.

philippjfr:
>I agree with the principle of having a base class though I hope we can express the idea better than DimensionedTree

I think ``DimensionedTree`` is the core idea here.

>The core concept is that these are the results of the general compositions operators (+ and *) that apply to heterogenous collections of elements.

While that's true for Overlay + Layout none of that functionality would actually live on DimensionedTree so that's not what defines a DimensionedTree. What defines a DimensionedTree is that it is a tree type which provides the common Dimensioned API.

jlstevens:
There seems to be a deeper issue. How are ``Layout`` and ``Overlay`` dimensioned? We never set kdims or vdims on these containers so ``LabelledTree`` seems more correct to me. I realize we probably use some of the methods of ``Dimensioned`` so at the minimum we should give the ``kdims``, ``vdims`` and ``cdims`` negative precedence and lock their maximum length to zero.

philippjfr:
>How are Layout and Overlay dimensioned?

They are dimensioned because they allow deep indexing, i.e. they define ``.ddims``, ``.redim`` and ``.select``. I do agree that them being Labelled is just as important as that though, so maybe the most sensible name is ``ViewableTree``, since that's what they really are a tree of Viewables.

jlstevens:
I think ``ViewableTree`` is the best suggestion so far.

jbednar:
`Viewable` seems good, but aren't both Layouts and Overlays restricted to a single non-hierarchical level?  "Tree" is confusing in that context, even if it's the underlying data structure.

philippjfr:
>but aren't both Layouts and Overlays restricted to a single non-hierarchical level?

No attribute access on both is two-level by default and nestable to whatever level you want if you construct it manually, e.g.:

```
>>> hv.Overlay([hv.Element([])])
:Overlay
   .Element.I :Element

>>> hv.Layout([('A.Deeply.Nested.Layout', hv.Image([]))])
:Layout
   .A.Deeply.Nested.Layout :Image   [x,y]   (z)
```

Plus this is still a baseclass and the previous baseclass was AttrTree, so I don't think referencing the datastructure is really an issue.

PR #2734, opened 2018-05-26
Interface test reorganization

philippjfr:
Reorganizes the interface tests into several files, previously a single file contained >1000 tests, which was less than ideal and quite annoying if you were simply testing changes to one interface. Secondly the class hierarchies of the tests were in some cases not correct and in other cases it wasn't testing the correct interface. The PR expands the test suite by about 200 tests by running more of the tests against a wider set of interfaces. Additionally this fixes two bugs uncovered by running the tests against other interfaces.


philippjfr:
Ready to merge.

jlstevens:
Splitting the tests up into more manageable chunks is sensible and helps make it easier to pull interfaces out of holoviews when it makes sense. Merging.

PR #2739, opened 2018-05-27
Allow the conversion interface to discover pandas indexes

philippjfr:
Allows referencing pandas indexes in .to interface even when Dataset does not declare them as dimensions.

- [x] Fixes https://github.com/ioam/holoviews/issues/2717

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -25,8 +25,9 @@ try:
     datatypes = ['dataframe', 'dictionary', 'grid', 'ndelement', 'array']
     DFColumns = PandasInterface
 except ImportError:
-    pass
+    pd = None
 except Exception as e:
+    pd = None
     param.main.warning('Pandas interface failed to import with '
                        'following error: %s' % e)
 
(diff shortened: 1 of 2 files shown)


philippjfr:
Ready to merge.

holoviews/core/data/__init__.py
```diff
@@ -132,7 +133,15 @@ def __call__(self, new_type, kdims=None, vdims=None, groupby=None,
...
-            selected = self._element.reindex(groupby+kdims, vdims)
+            if pd and issubclass(self._element.interface, PandasInterface):
+                ds_dims = self._element.dimensions()
+                ds_kdims = [self._element.get_dimension(d) if d in ds_dims else d
+                            for d in groupby+kdims]
+                ds_vdims = [self._element.get_dimension(d) if d in ds_dims else d
+                            for d in vdims]
+                selected = self._element.clone(kdims=ds_kdims, vdims=ds_vdims)
+            else:
+                selected = self._element.reindex(groupby+kdims, vdims)
```
jlstevens: Is the same logic used elsewhere in a context other than the conversion interface? Is any code shared and is it worth abstracting out?

jlstevens:
Other than one question in the comment above, this looks fine and I'm happy to merge once you respond.

holoviews/core/data/__init__.py
philippjfr: No, I cannot think of anywhere this could be reused atm.

PR #2740, opened 2018-05-27
Fixed bokeh LayoutPlot bugs

philippjfr:
Turns out that the bokeh LayoutPlot implementation was wrong in all kinds of subtle ways so I've completely rewritten the implementation from scratch. The main issue was that AdjointLayout plots were not correctly positioned on the grid. Secondly when enabling ``tabs`` AdjointLayout plots were split across multiple tabs.

- [x] Fixes https://github.com/ioam/holoviews/issues/1489
- [x] Fixes https://github.com/ioam/holoviews/issues/2568
- [x] Fixes https://github.com/ioam/holoviews/issues/2652
- [x] Adds unit test

### DIFF
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -876,7 +876,7 @@ class LayoutPlot(CompositePlot, GenericLayoutPlot):
         kwargs = dict(sizing_mode=self.sizing_mode)
         if self.tabs:
             plots = filter_toolboxes([p for t, p in tab_plots])
-            panels = [Panel(child=child, title=title) for title, child in tab_plots]
+            panels = [Panel(child=child, title=t) for t, child in tab_plots]
             layout_plot = Tabs(tabs=panels)
         else:
             plots = layout_padding(plots, self.renderer)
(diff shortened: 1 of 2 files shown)


philippjfr:
Ready to review and merge.

jlstevens:
> Turns out that the bokeh LayoutPlot implementation was wrong in all kinds of subtle ways...

Why do I feel deja vu? ;-p

To be fair it has always been very tricky code and I am always happy to see it refactored and improved. I've had to restart a test build for what I assume was a transient but otherwise I am happy to merge.

The only thing I would ask you to consider is whether there is another unit test or two you could add. The more unit tests the better given how tricky it is to get right!

philippjfr:
>Why do I feel deja vu? ;-p

I think this code has been around since I first wrote the bokeh backend, so it has actually held up surprisingly well, unlike the matplotlib version.

>The only thing I would ask you to consider is whether there is another unit test or two you could add. The more unit tests the better given how tricky it is to get right!

Added a unit test for the tabs issue, layouts themselves have enough unit tests covering them.

jlstevens:
> Added a unit test for the tabs issue, layouts themselves have enough unit tests covering them.

Thanks. I'll merge when the tests go green.

PR #2741, opened 2018-05-27
Fixed bug in js conversion of value formatters

philippjfr:
Ensures that only the formatter function is modified.

- [x] Fixes https://github.com/ioam/holoviews/issues/2707


jlstevens:
This is more specific than before but also more brittle e.g if the ``formatter`` argument were ever to change, this would break. I would love to avoid regular expressions here but I don't have a suggestion for how that could be done so I'll go ahead and merge.

PR #2745, opened 2018-05-28
Move DynamicMap unbounded error onto Renderer

philippjfr:
The validation for unbounded DynamicMaps previously lived inside the display hooks which meant that when working on bokeh server it did not raise a reasonable error message. This moves the validation onto the Renderer which ensures that it is properly validated in and outside the notebook.

- [x] Fixes https://github.com/ioam/holoviews/issues/2309

### DIFF
--- a/holoviews/plotting/renderer.py
+++ b/holoviews/plotting/renderer.py
@@ -12,7 +12,7 @@ import param
 from ..core.io import Exporter
 from ..core.options import Store, StoreOptions, SkipRendering, Compositor
 from ..core.util import find_file, unicode, unbound_dimensions, basestring
-from .. import Layout, HoloMap, AdjointLayout
+from .. import Layout, HoloMap, AdjointLayout, DynamicMap
 from .widgets import NdWidget, ScrubberWidget, SelectionWidget
 
 from . import Plot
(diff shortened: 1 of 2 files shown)


philippjfr:
Ready to review and merge.

holoviews/plotting/renderer.py
```diff
@@ -484,7 +492,7 @@ def open_and_read(path, f):
         if extras:
             dependencies.update(cls.extra_dependencies)
         for backend in backends:
-            dependencies['backend'] = Store.renderers[backend].backend_dependencies
+            dependencies[backend] = Store.renderers[backend].backend_dependencies
```
philippjfr: Separate fix that snuck in because I've had it on my working tree for a week. Happy to make a new PR, doesn't seem worthwhile though.

jlstevens: What was going on here before?! What was the broken behavior and what is the new behavior?

holoviews/plotting/renderer.py
```diff
@@ -183,6 +183,14 @@ def get_plot(self_or_cls, obj, renderer=None):
         """
         Given a HoloViews Viewable return a corresponding plot instance.
         """
+        if isinstance(obj, DynamicMap) and obj.unbounded:
+            dims = ', '.join('%r' % dim for dim in obj.unbounded)
+            msg = ('DynamicMap cannot be displayed without explicit indexing '
+                   'as {dims} dimension(s) are unbounded. '
+                   '\nSet dimensions bounds with the DynamicMap redim.range '
+                   'or redim.values methods.')
+            raise SkipRendering(msg.format(dims=dims))
```
jlstevens: I guess bokeh server will have the ``SkipRendering`` exception raised in this case which it has no special handling for. It doesn't have a way to actually 'skip' rendering afaik but this is just a minor issue/comment and it is fine to leave it as is.

jlstevens:
Moving this logic out the display hooks is a good idea. Other than the question about your separate fix, this looks good. Merging.

holoviews/plotting/renderer.py
philippjfr: Nothing was actually broken in any usual scenario, since you'd usually only request to embed the JS/CSS dependencies for one backend in an HTML file. In theory though you could do something crazy like embed a plotly plot in a bokeh Div and try to export that to HTML.

PR #2746, opened 2018-05-28
Grid interface reindex

philippjfr:
This ensures that GridInterface.reindex has the same behavior as NdMapping.reindex, which is to drop scalar dimensions automatically. Also fixes a bug in ``XArrayInterface.ndloc`` uncovered in the process of writing tests for the new reindex functionality.

- [x] Fixes https://github.com/ioam/holoviews/issues/2362
- [x] Adds unit test


philippjfr:
Ready to merge.

jlstevens:
Any worries about people relying on the old behavior? Otherwise looks good. Merging.

PR #2748, opened 2018-05-29
Fix dynamic ScrubberWidget and add throttling

philippjfr:
Fixes two issues with the ScrubberWidget:

1. A few releases ago we added the ability to use the ScrubberWidget on a DynamicMap as long all  its dimensions define ``values``, however for this to work the plots had to expand the values into their cross-product. We subsequently reverted the expansion change because it was very slow and expensive for large cross products but did not disable the scrubber. This PR ensures that this works even without expanding the whole cross-product.

2. This PR adds throttling to the ScrubberWidget by using the same approach used by the selection widgets.

- [x] Fixes https://github.com/ioam/holoviews/issues/2571

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -1550,6 +1550,26 @@ def cartesian_product(arrays, flat=True, copy=False):
     return tuple(arr.copy() if copy else arr for arr in arrays)
 
 
+def cross_index(values, index):
+    """
+    Allows efficiently indexing into a cartesian product without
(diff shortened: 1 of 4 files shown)


philippjfr:
Ready for review and merge.

jlstevens:
Looks good and I'm glad the most tricky it is contained within a well-defined utility. I'll merge once the pr build goes green too.

jlstevens:
The failing pr test was a transient (now restarted). Merging.

PR #2750, opened 2018-05-29
fix typo

jordansamuels:


jordansamuels:
No more right now, it was a spur of the moment thing 😋

PR #2755, opened 2018-06-01
Nested legend

philippjfr:
This PR ensures that nested Nd(Overlays) generate a legend. Previously nesting Overlays inside an NdOverlay would mean two things:

1) The styling group_counter would be reset inside the nested OverlayPlot meaning that color cycles would not work.
2) The topmost legend plot would not find matplotlib artists nested further down and the bokeh version would not pass its dimensions down to the individual elementplots meaning that no legend would be generated.

```python
%%opts Spread (alpha=0.1) Path [show_legend=True]

d = {}
for i in range(1, 5):
    xs = np.linspace(0, 1)
    ys = xs**2 / i + 5 * i
    d[i] = hv.Path((xs, ys), kdims=['X', 'Y']) * hv.Spread((xs, ys, 2))
hv.NdOverlay(d)
```

screen shot 2018-06-01 at 1 27 20 pm

image

- [x] Fixes https://github.com/ioam/holoviews/issues/2754

### DIFF
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -765,7 +765,7 @@ class OverlayPlot(LegendPlot, GenericOverlayPlot):
             else:
                 if isinstance(subplot, OverlayPlot):
                     legend_data += subplot.handles.get('legend_data', {}).items()
-                if element.label and handle:
+                elif element.label and handle:
                     legend_data.append((handle, element.label))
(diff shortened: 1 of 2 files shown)


jlstevens:
Looks good to me as well. Happy to merge once the tests are green.

philippjfr:
Ready to merge now.

PR #2762, opened 2018-06-02
Implement grid concatenation and standardize datatype casting

philippjfr:
This PR has two main aims:

1) Standardize and simplify how casting between different datatypes works
2) Implement concatenation for all datatypes

Before this PR was applied past both casting and concatenation were limited to columnar data formats, which meant that certain operations could not be applied to gridded data, e.g. a HoloMap collapse. Having a dedicated concat implementation for both columnar and gridded data also allows much more efficient concatenation than what is currently in use by methods like ``.table`` and ``.dframe`` and will generalize them so that we can eventually replace the column specific ``.table`` implementation with a general one that returns a dataset of arbitrary type.

Implementing concatenation along HoloMap dimensions also means that Dataset.groupby operations are now reversible and fixes HoloMap.collapse.

- [x] Fixes https://github.com/ioam/holoviews/issues/1417
- [x] Adds unit tests

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -68,13 +68,15 @@ from .. import util
 
 def concat(datasets, datatype=None):
     """
-    Concatenates multiple datasets wrapped in an NdMapping type
-    along all of its dimensions. Before concatenation all datasets
-    are cast to the same datatype. For columnar data concatenation
-    adds the columns for the dimensions being concatenated along
-    and then concatenates all the old and new columns. For gridded
-    data a new axis is created for each dimension being concatenated
-    along and then hierarchically concatenates along each dimension.
+    Concatenates multiple datasets wrapped in an NdMapping type along
+    all of its dimensions. Before concatenation all datasets are cast
+    to the same datatype, which may be explicitly defined or
+    implicitly derived from the first datatype that is
+    encountered. For columnar data concatenation adds the columns for
+    the dimensions being concatenated along and then concatenates all
+    the old and new columns. For gridded data a new axis is created
+    for each dimension being concatenated along and then
+    hierarchically concatenates along each dimension.
 
     Signature
     ---------
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -84,26 +84,7 @@ def concat(datasets, datatype=None):
 
     Returns: Dataset
     """
-    if isinstance(datasets, NdMapping):
-        dimensions = datasets.kdims
-        datasets = datasets.data
-    if isinstance(datasets, (dict, OrderedDict)):
-        datasets = datasets.items()
-    keys, datasets = zip(*datasets)
-    template = datasets[0]
-    datatype = datatype or template.interface.datatype
-
-    # Handle non-general datatypes by casting to general type
-    if datatype == 'array':
-        datatype = default_datatype
-    elif datatype == 'image':
-        datatype = 'grid'
-
-    datasets = template.interface.cast(datasets, datatype)
-    template = datasets[0]
-    data = list(zip(keys, datasets))
-    concat_data = template.interface.concat(data, dimensions, vdims=template.vdims)
-    return template.clone(concat_data, kdims=dimensions+template.kdims, new_type=Dataset)
+    return Interface.concatenate(datasets, datatype)
 
 
 class DataConversion(object):
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -61,9 +61,42 @@ if 'array' not in datatypes:
 
 from ..dimension import Dimension, process_dimensions
 from ..element import Element
-from ..ndmapping import OrderedDict
+from ..ndmapping import OrderedDict, NdMapping
 from ..spaces import HoloMap, DynamicMap
 from .. import util
+
+
+def concat(datasets, datatype=None):
+    """
+    Concatenates multiple datasets wrapped in an NdMapping type
+    along all of its dimensions. Before concatenation all datasets
+    are cast to the same datatype. For columnar data concatenation
+    adds the columns for the dimensions being concatenated along
+    and then concatenates all the old and new columns. For gridded
+    data a new axis is created for each dimension being concatenated
+    along and then hierarchically concatenates along each dimension.
+
+    Signature
+    ---------
+
+    datasets: NdMapping of Datasets defining dimensions to concatenate on
+    datatype: Datatype to cast data to before concatenation
+
+    Returns: Dataset
+    """
+    if isinstance(datasets, NdMapping):
+        dimensions = datasets.kdims
+        datasets = datasets.data
+    if isinstance(datasets, (dict, OrderedDict)):
+        datasets = datasets.items()
+    keys, datasets = zip(*datasets)
+    template = datasets[0]
+    datatype = datatype or template.interface.datatype
+    datasets = template.interface.cast(datasets, datatype)
+    template = datasets[0]
+    data = list(zip(keys, datasets))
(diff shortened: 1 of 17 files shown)


holoviews/core/util.py
```diff
@@ -1532,7 +1531,7 @@ def groupby_python(self_or_cls, ndmapping, dimensions, container_type,
         selects = get_unique_keys(ndmapping, dimensions)
         selects = group_select(list(selects))
         groups = [(k, group_type((v.reindex(idims) if hasattr(v, 'kdims')
-                                  else [((), (v,))]), **kwargs))
+                                  else [((), v)]), **kwargs))
```
philippjfr: These are NdMapping implementation details left over from when we had an NdElement, the precursor to datasets, which now lead to strange behavior.

holoviews/core/data/__init__.py
```diff
@@ -17,11 +17,13 @@
 from .multipath import MultiInterface         # noqa (API import)
 from .image import ImageInterface             # noqa (API import)
 
+default_datatype = 'dictionary'
 datatypes = ['dictionary', 'grid']
 
 try:
     import pandas as pd # noqa (Availability import)
     from .pandas import PandasInterface
+    default_datatype = 'dataframe'
```
philippjfr: When converting from gridded to columnar data throughout the code it usually has to cast the data to a specific datatype. Various places in the code hardcoded ``['pandas', 'dictionary']`` in these places, defining a default_datatype avoids having to hardcode this all over the place.

holoviews/core/data/grid.py
```diff
@@ -115,6 +117,31 @@ def init(cls, eltype, data, kdims, vdims):
...
+    @classmethod
+    def concat_dim(cls, datasets, dim, vdims):
+        values, grids = zip(*datasets.items())
+        new_data = {k: v for k, v in grids[0].items() if k not in vdims}
+        new_data[dim.name] = np.array(values)
+        for vdim in vdims:
+            arrays = [grid[vdim.name] for grid in grids]
+            stack = np.stack if any(is_dask(arr) for arr in arrays) else da.stack
+            new_data[vdim.name] = stack(arrays, -1)
+        return new_data
```
philippjfr: Since arrays cannot be concatenated along multiple axes at once the implementation of ``concat`` on gridded interfaces has two components. A general ``concat`` method coordinates hierarchical concatenation along each dimension and uses the interface specific ``concat_dim`` method implementations to concatenate along one particular axis or dimension.

philippjfr:
Made some additional comments to clarify certain implementation details.

holoviews/core/data/interface.py
```diff
@@ -105,30 +105,21 @@ class Interface(param.Parameterized):
...
-
-        return [co.clone(co.columns(), datatype=[datatype], new_type=cast_type)
-                for co in dataset]
-
+        datatype = datatype or cls.datatype
+        interfaces = list(util.unique_iterator((d.interface for d in datasets)))
+        cast = []
+        for ds in datasets:
+            if cast_type is not None or ds.interface.datatype != datatype:
+                ds = ds.clone(ds, datatype=[datatype], new_type=cast_type)
```
philippjfr: Casting works quite simply, if the ``Interface.initialize`` is passed another dataset and it finds a mismatch between the supplied datatype and the requested datatype it will deconstruct the original dataset into the columnar or gridded tuple format, which is supported by all interfaces. In this way a dataset can easily be cast to any other datatype, except for columnar -> gridded conversions.

holoviews/core/data/__init__.py
jbednar: Shouldn't this be "default_columnar_datatype', then?  Or are there no cases where columnar data needs to be cast into some gridded data type?

philippjfr: Columnar data cannot be cast to gridded data without some kind of aggregation occurring. So that's correct. Would still be okay with changing it to ``default_columnar_datatype``.

philippjfr:
Ready for review.

holoviews/core/data/__init__.py
```diff
@@ -64,6 +66,27 @@
 from .. import util
 
 
+def concat(datasets, datatype=None):
+    """
+    Concatenates multiple datasets wrapped in an NdMapping type
+    along all of its dimensions. Before concatenation all datasets
+    are cast to the same datatype. For columnar data concatenation
```
jlstevens: 'same datatype' determined how?

philippjfr: Either explicitly defined or the type of the first dataset that was passed in.

holoviews/core/data/interface.py
```diff
@@ -307,25 +297,35 @@ def range(cls, dataset, dimension):
...
+        from . import Dataset, default_datatype
+        new_type = new_type or Dataset
+        if isinstance(datasets, NdMapping):
+            dimensions = datasets.kdims
+            datasets = datasets.data
+        if isinstance(datasets, (dict, OrderedDict)):
+            datasets = datasets.items()
+            keys, datasets = zip(*datasets)
+        elif isinstance(datasets, list) and not any(isinstance(v, tuple) for v in datasets):
+            keys = [()]*len(datasets)
```
jlstevens: What are all these empty tuple keys for? Just to get things in the right format?

philippjfr: Right, concatenate is usually meant for concatenating along some dimension but you can also concatenate a simple list of datasets without concatenating along some dimensions. For that case we generate empty tuple keys. Happy to add a comment. Separately I also need to assert that this only happens for tabular data, since gridded data must be concatenated along some dimension.

holoviews/core/data/iris.py
```diff
@@ -4,6 +4,9 @@
 from itertools import product
 
 import iris
+from iris.coords import DimCoord
+from iris.cube import CubeList
+from iris.experimental.equalise_cubes import equalise_attributes
```
jlstevens: Will be good to have the iris interface moved to geoviews. Could this be done for 1.10.6?

holoviews/core/spaces.py
```diff
@@ -324,16 +325,18 @@ def collapse(self, dimensions=None, function=None, spreadfn=None, **kwargs):
...
+                    agg = group_data.aggregate(group.last.kdims, function, spreadfn, **kwargs)
+                    group_data = group.type(agg)
             else:
+                group_data = [el.data for el in group]
+                args = (group_data, function, group.last.kdims)
                 data = group.type.collapse_data(*args, **kwargs)
-                col_data = group.last.clone(data)
-            collapsed[key] = col_data
-        return collapsed if self.ndims > 1 else collapsed.last
+                group_data = group.last.clone(data)
```
jlstevens: ``group_data`` can be a whole load of different things at different times. Not critical but I would prefer to have something that isn't clobbered so much.

jlstevens:
Other than a few minor comments this looks good and I'm happy to merge.

holoviews/core/data/__init__.py
jlstevens: Would be good to state that bit about it being chosen from the first one if not explicitly set.

holoviews/core/data/iris.py
philippjfr: Tests need to be moved into the holoviews package first.

jlstevens: You mean 'geoviews' package?

philippjfr: No I mean, the ``/tests`` need to move to ``/holoviews/tests``.

philippjfr: The interface tests are defined as mix-in classes, so if I want to run them in geoviews I have to be able to import them from holoviews. We also promised this to the bokeh folks so they can run our bokeh unit tests easily.

jlstevens:
Tests are green. Merging.

PR #2766, opened 2018-06-04
Theme fix

jlstevens:
Unfortunately theming broke in the notebook again. This is because theming lacks unit tests, partially because it is difficult to test - checking Bokeh's ``.state`` in the usual way does not seem to work.

This PR restores theming in the notebook and I seem to have found a way to test it (at least for the notebook context) at the display hooks level. For some reason, everything has to go through the full notebook machinery for theming to be applied.

Although it *looks* like it is working again this PR is still WIP:

- [x] I'm not sure if this is the right fix - it seems to work in notebook but maybe it is incorrect or breaks something else?
- [x] Need to test theming keeps working in the bokeh server context.
- [x] Need to improve this new unit test (e.g skip if bokeh is unavailable, simplify test theme, tear it down nicely etc)


philippjfr:
The fix seems fine, I'd prefer it if bokeh correctly respected the theming applied to the Document actually being plotted when working in the notebook but this fix shouldn't cause any other issues so I'm happy with it.

As for testing here's what seems to work. For the notebook mode do this:

```python
renderer.theme = theme
curve = hv.Curve([])
plot = renderer.get_plot(curve)
renderer(plot)
self.assertEqual(plot.state.outline_line_color, '#444444')
```

And for server mode I'd probably actually start up a server:

```python
    def test_server_theming(self):
        obj = Curve([])
        bokeh_renderer.theme = theme
        plots = []
        def modify_doc(doc):
            plot = bokeh_renderer.get_plot(obj, doc=doc)
            plots.append(plot)
            server.stop()
        handler = FunctionHandler(modify_doc)
        app = Application(handler)
        server = Server({'/': app}, port=0)
        server.start()
        url = "http://localhost:" + str(server.port) + "/"
        pull_session(session_id='Test', url=url, io_loop=server.io_loop)
        self.assertTrue(plots[0].state.outline_line_color, '#444444')
```

tests/ipython/testdisplayhooks.py
```diff
@@ -73,3 +73,47 @@ def test_store_render_combined(self):
                       'application/vnd.holoviews_exec.v0+json',
                       'image/svg+xml', 'image/png'}
         self.assertEqual(set(data), mime_types)
+
+
+class TestBokehTheming(TestDisplayHooks):
```
philippjfr: Please move the tests inside ``tests/plotting/bokeh``.

philippjfr:
Lastly, I'd probably use a custom ``bokeh_renderer`` instance or restore the previous theme after the tests so the theming state doesn't leak.

tests/ipython/testdisplayhooks.py
jlstevens: Fixed in 61cb658

jlstevens:
It turns out the server test is not working after all: it isn't testing the server machinery properly and is redundant with the test in renderer. It doesn't look like the theme has been applied after ``pull_session``. Any suggestions? And if not should I just remove the test?

philippjfr:
Ready to merge?

jlstevens:
> Ready to merge?

I think so unless you've figured out a way to do the tests in the bokeh server context...

philippjfr:
I have not.

PR #2770, opened 2018-06-04
Update the bokeh reset ranges

philippjfr:
In https://github.com/bokeh/bokeh/pull/7948 I added ``reset_start`` and ``reset_end`` parameters for bokeh ranges to allow updating the range the reset tool resets to. This is a longstanding annoyance for HoloViews plots and this PR simply updates the reset_range whenever the actual ranges are updated.

- [x] Fixes https://github.com/ioam/holoviews/issues/2371

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -26,7 +26,7 @@ from ..util import dynamic_update, process_cmap
 from .plot import BokehPlot, TOOLS
 from .util import (mpl_to_bokeh, get_tab_title,  py2js_tickformatter,
                    rgba_tuple, recursive_model_update, glyph_order,
-                   decode_bytes)
+                   decode_bytes, bokeh_version)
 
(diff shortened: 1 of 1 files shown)


jlstevens:
So in short, it no longer jumps the viewport about when using the reset tool?

philippjfr:
Yes, it resets to the most recently applied range rather than the initially applied range.

jlstevens:
I suppose my only question is...what if you explicitly want the original behavior? How do you disable it? And if you can disable it, we should document it...

philippjfr:
>I suppose my only question is...what if you explicitly want the original behavior?

I don't think the original behavior was in any way desirable and consider it a bug. If you can think of an example where the original behavior is desirable I'd consider adding an option, but I really can't think of one.

jlstevens:
> I don't think the original behavior was in any way desirable and consider it a bug.

I may have initialized the plot in some sensible reference frame that you may want to reset to. Anyway, I agree this is not the common case so I'll go ahead and merge.

PR #2789, opened 2018-06-09
Handle kdims referencing multi-index in dask

philippjfr:
Since dask dataframes do not support multi-indexes the current code does not detect when a multi-index is referenced. The only way of checking whether the multi-index exists is to speculatively reset the index (which is almost free when using dask) and check whether the columns exist on the reindexed dataframe. This code now resets the index and uses the reindexed dataframe if the columns are present there (otherwise the validation code will complain about missing columns as before).

- [x] Adds unit tests

### DIFF
--- a/holoviews/core/data/dask.py
+++ b/holoviews/core/data/dask.py
@@ -11,6 +11,7 @@ import dask.dataframe as dd
 from dask.dataframe import DataFrame
 
 from .. import util
+from ..dimension import Dimension
 from ..element import Element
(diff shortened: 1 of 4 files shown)


philippjfr:

jlstevens:
Looks good though I think it is worth adding a comment to the code to explain the approach (i.e '# speculatively reset index to detect....' or something similar).

jlstevens:
Looks good and the addition of the comment definitely shouldn't affect the tests. Merging.

PR #2790, opened 2018-06-10
Cleaned up Dimension casting and name lookup

philippjfr:
The code is currently littered with ``isinstance(dim, Dimension)`` checks, which is not optimal because it's duplicative but more importantly because it does not check for the full range of valid dimension specs which include tuples and dictionaries. This PR adds two utilities one to cast Dimension-like objects to Dimensions if it isn't already one and the other that looks up the name on Dimensions or Dimension-like objects. It then uses these utilities throughout the code.

- [x] Adds unit test

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -41,6 +41,38 @@ def param_aliases(d):
     return d
 
 
+def as_dimension(dimension):
+    """
+    Converts tuple, dict and basestring types to Dimension and leaves
+    Dimension types untouched.
+    """
+    if isinstance(dimension, Dimension):
+        return dimension
+    elif isinstance(dimension, (tuple, dict, basestring)):
+        return Dimension(dimension)
+    else:
+        raise ValueError('%s type could not be interpreted as Dimension. '
+                         'Dimensions must be declared as a string, tuple, '
+                         'dictionary or Dimension type.')
+
+def dimension_name(dimension):
+    """
+    Looks up the dimension name on a Dimension or Dimension-like object.
+    """
+    if isinstance(dimension, Dimension):
+        return dimension.name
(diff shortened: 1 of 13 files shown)


philippjfr:
Ready to review/merge.

jlstevens:
So this is a refactor as opposed to a feature or bug fix?

holoviews/core/dimension.py
```diff
@@ -41,6 +41,41 @@ def param_aliases(d):
     return d
 
 
+def as_dimension(dimension):
```
jlstevens: Could this not simply be called ``dimension``?

jlstevens: Nevermind...I see ``dimension`` is frequently used as name even though ``dim`` is a lot shorter.

philippjfr: I think I prefer ``as_dimension`` or ``asdimension``, akin to numpy ``asarray``, ``dimension`` is far too commonly used as a variable name so it would cause issues and be confusing.

jlstevens: How about ``asdim``?

philippjfr:
>So this is a refactor as opposed to a feature or bug fix?

It also fixes bugs when constructing Datasets with tuple dimensions.

holoviews/core/dimension.py
philippjfr: Can't say I prefer it but if you insist I'll do it.

jlstevens: It is a lot shorter so I do prefer it..unless you have any specific objections to it?

philippjfr: Okay, pushed.

jlstevens:
Looks good. I'll merge when the tests are green.

PR #2793, opened 2018-06-11
Dropping gridded data non-scalar coordinates casts datatype 

philippjfr:
Currently when dropping non-scalar gridded data dimensions it still attempts to maintain the datatype, which can cause issues due to duplicate coordinates. This PR cleans up the detection of scalar coordinates (i.e. key dimensions) and ensures that if a non-scalar dimension is dropped the data is cast from a gridded to a tabular datatype.

- [x] Adds unit test

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -327,14 +327,20 @@ class Dataset(Element):
         Create a new object with a re-ordered set of dimensions.  Allows
         converting key dimensions to value dimensions and vice versa.
         """
+        gridded = self.interface.gridded
+        scalars = []
+        if gridded:
+            coords = [(d, self.interface.coords(self, d)) for d in self.kdims]
+            scalars = [d for d, vs in coords if len(vs) == 1]
+
         if kdims is None:
             # If no key dimensions are defined and interface is gridded
             # drop all scalar key dimensions
-            gridded = self.interface.gridded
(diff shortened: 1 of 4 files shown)


philippjfr:

holoviews/core/data/__init__.py
```diff
@@ -22,7 +22,7 @@
 try:
     import pandas as pd # noqa (Availability import)
     from .pandas import PandasInterface
-    datatypes = ['dataframe', 'dictionary', 'grid', 'ndelement', 'array']
+    datatypes = ['dataframe', 'dictionary', 'grid', 'array']
```
jlstevens: I certainly hope nothing is still using 'ndelement' at this point!

jlstevens:
Looks good and I'll merge when the tests are green.

philippjfr:
Stupidly need to rebuild the test data because it contains references to ndelement.

jlstevens:
Had to restart the Python 2 tests twice. Hope that was because of two transients as opposed to one captured in the tests. Anyway, it is green now so I'll merge.

philippjfr:
May have been a captured transient since master is failing too but I need to rebuild test data for single Image pixel change anyway (once again that exception in the Exporting tutorial that's been captured in the test data, I'll be updating nbpublisher to fix it).

PR #2794, opened 2018-06-11
Handle single pixel Image types

philippjfr:
In https://github.com/ioam/holoviews/issues/2785 an issue describes how when creating a single-pixel Image datashading breaks. This is because Image cannot compute a density from a single coordinate. This can be worked around if the resampling operation directly supplies the Image ``bounds``. Additionally this PR adds an explicit check for invalid densities, which arise when a single-pixel image is declared without supplying bounds.

### DIFF
--- a/holoviews/element/raster.py
+++ b/holoviews/element/raster.py
@@ -279,6 +279,11 @@ class Image(Dataset, Raster, SheetCoordinateSystem):
         l, b, r, t = bounds.lbrt()
         xdensity = xdensity if xdensity else compute_density(l, r, dim1, self._time_unit)
         ydensity = ydensity if ydensity else compute_density(b, t, dim2, self._time_unit)
+        if not np.isfinite(xdensity) or not np.isfinite(ydensity):
+            raise ValueError('Density along Image axes could not be determined. '
+                             'If the data contains only one coordinate along the '
+                             'x- or y-axis ensure you declare the bounds and/or '
+                             'density.')
         SheetCoordinateSystem.__init__(self, bounds, xdensity, ydensity)
         self._validate(data_bounds, supplied_bounds)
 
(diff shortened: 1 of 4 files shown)


jlstevens:
Looks good though isn't working with a single pixel normally an error (or at least, highly unusual)? In any case, this PR is fine and I'll merge when the tests pass.

philippjfr:
>Looks good though isn't working with a single pixel normally an error (or at least, highly unusual)?

No, there's nothing wrong with an Image which has a single pixel along one or both axes as long as the density is known or can be computed.

holoviews/element/raster.py
```diff
@@ -279,6 +279,11 @@ def __init__(self, data, kdims=None, vdims=None, bounds=None, extents=None,
         l, b, r, t = bounds.lbrt()
         xdensity = xdensity if xdensity else compute_density(l, r, dim1, self._time_unit)
         ydensity = ydensity if ydensity else compute_density(b, t, dim2, self._time_unit)
+        if not np.isfinite(xdensity) or not np.isfinite(ydensity):
+            raise ValueError('Density along Image axes could not be determined. '
+                             'If the data contains only one coordinate along the '
+                             'x- or y-axis ensure you declare the bounds and/or '
+                             'density.')
```
jbednar: And compute_density will really return a floating-point inf value, not raise an exception?

jbednar:
Is there a similar issue that needs fixing with datashader itself still, or is this sufficient?

philippjfr:
>Is there a similar issue that needs fixing with datashader itself still, or is this sufficient?

This is sufficient.

jbednar:
Thanks!

philippjfr:
Datetime issues now resolved, we're now also validating datetime bounds. Ready to merge once passing.

philippjfr:
Will need rebuilt tests due to the Exporting tutorial notebook exception.

philippjfr:
Updated nbpublisher to avoid the exception in the exporting notebook, so we don't have to update the tests every time the line numbers in element/raster.py change.

jlstevens:
Looks good. Merging.

PR #2803, opened 2018-06-13
Ensure that Dataset.groupby does not sort by default

philippjfr:
Seemingly all the tabular data interfaces sorted the groups when applying a groupby, which is not always desirable, and is inconsistent with the gridded interfaces, which already maintain the supplied ordering. It also causes issues like https://github.com/ioam/holoviews/issues/2792, where the user expected the data to stay in the order it was defined in. While I think the previous behavior was incorrect I'm also somewhat wary of changing it right away since it is a definite change in behavior. Any thoughts @jbednar and @jlstevens?

- [x] Fixes https://github.com/ioam/holoviews/issues/2792

### DIFF
--- a/holoviews/core/data/array.py
+++ b/holoviews/core/data/array.py
@@ -8,7 +8,7 @@ import numpy as np
 from .interface import Interface, DataError
 from ..dimension import dimension_name
 from ..element import Element
-from ..ndmapping import NdMapping, item_check
+from ..ndmapping import NdMapping, item_check, sorted_context
 from .. import util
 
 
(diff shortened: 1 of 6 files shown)


jbednar:
I think it should change but at a fairly major release, so presumably 1.11?

philippjfr:
That seems fine, I can fix #2793 by disabling sorting inside the plot, so we can fix the immediate issue and wait to merge this until the next major release.

jlstevens:
> I think it should change but at a fairly major release, so presumably 1.11?

Agreed.

philippjfr:
Now that we're working towards 1.11 I think we should just merge this now.

philippjfr:
Merge?

PR #2805, opened 2018-06-14
Do not sort data in Violin and BoxWhisker plots

### LINKED ISSUE #2792: Conversion of xarray to BoxWhisker breaks coordinate ordering
With an xarray Dataset `xarr` with a categorical coordinate `label` and a data variable `value`, I have reordered the coordinate labels to have a particular order. If I apply
```
xarr.to(hv.Scatter, [`label`], [`value`])
```
the categorical labels are in the correct order along the x-axis. If I do  
```
xarr.to(hv.BoxWhisker, [`label`], [`value`])
```
instead, then the labels appear in a different (seemingly alphabetical) order. There is a `FutureWarning` which hints at issues with preserving key order: 

```
/home/name/anaconda3/envs/tftt/lib/python3.5/site-packages/holoviews/core/data/pandas.py:130: FutureWarning: from_items is deprecated. Please use DataFrame.from_dict(dict(items), ...) instead. DataFrame.from_dict(OrderedDict(items)) may be used to preserve the key order.
  zip(columns, data)])
```

Is there a way to reorder the categorical labels of an axis post hoc?

philippjfr:
Stops the Violin and BoxWhisker plots from sorting the data on groupbys which provides a temporary fix for #2792 until we disable sorting on groupbys in general in https://github.com/ioam/holoviews/pull/2803.

- [x] Fixes #2792

### DIFF
--- a/holoviews/plotting/bokeh/stats.py
+++ b/holoviews/plotting/bokeh/stats.py
@@ -7,6 +7,7 @@ import numpy as np
 from bokeh.models import FactorRange, Circle, VBar, HBar
 
 from ...core.dimension import Dimension
+from ...core.ndmapping import sorted_context
 from ...core.util import (basestring, dimension_sanitizer, wrap_tuple,
                           unique_iterator)
 from ...operation.stats import univariate_kde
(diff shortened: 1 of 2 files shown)


jlstevens:
Looks good as a temporary solution till the general fix is ready.

PR #2808, opened 2018-06-18
Fix for non-categorical discrete color levels in bokeh

philippjfr:
Fixes bug that was introduced in the recent PR to support color intervals. Also improves the unit test to ensure it would have caught this regression.


philippjfr:
Ready to review/merge.

jlstevens:
Looks good. Merging.

PR #2812, opened 2018-06-19
Histogram enhancements

philippjfr:
Various enhancements for the histogram operation:

* Remove ``individually`` parameter which didn't do anything
* Added ``cumulative`` parameter to compute cumulative histogram (https://github.com/ioam/holoviews/issues/2811)
* Added option to pass in explicit list/array of bins
- [x] Adds unit tests

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -495,9 +495,6 @@ class histogram(Operation):
     groupby = param.ClassSelector(default=None, class_=(basestring, Dimension), doc="""
       Defines a dimension to group the Histogram returning an NdOverlay of Histograms.""")
 
-    individually = param.Boolean(default=True, doc="""
-      Specifies whether the histogram will be rescaled for each Element in a UniformNdMapping.""")
-
     log = param.Boolean(default=False, doc="""
       Whether to use base 10 logarithmic samples for the bin edges.""")
 
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -486,6 +486,12 @@ class histogram(Operation):
     bin_range = param.NumericTuple(default=None, length=2,  doc="""
       Specifies the range within which to compute the bins.""")
 
+    bins = param.ClassSelector(default=None, class_=(np.ndarray, list), doc="""
+      An explicit set of bin edges.""")
+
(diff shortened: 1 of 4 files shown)


holoviews/operation/element.py
```diff
@@ -486,6 +486,12 @@ class histogram(Operation):
     bin_range = param.NumericTuple(default=None, length=2,  doc="""
       Specifies the range within which to compute the bins.""")
 
+    bins = param.ClassSelector(default=None, class_=(np.ndarray, list), doc="""
```
jlstevens: What about tuples or pandas series? Do we want to support lots of different types or force a single type?

jlstevens:
Made one comment. Otherwise looks good.

holoviews/operation/element.py
philippjfr: I'd be okay adding tuples I suppose, although that doesn't seem particularly common.

jlstevens: We should use a known set of types for 'list-like' types. Could you open an issue about this? Doesn't need to be fixed in this PR.

philippjfr: After considering this I'm sure this makes sense tbh. It's simply too dependent on the particular use case. In many cases ndarray and lists might be used interchangeably but often supporting arrays doesn't really make sense. Then there's lists and tuples, which again can be used interchangeably in some cases but are treated completely differently in others.

I've now added tuples here, but I don't think that generalizes in any real sense.

jbednar: You're *not* sure it makes sense?

philippjfr: Right, fixed.

jlstevens: My (not particularly strongly held) opinion is that it should only support lists: ``ndarray`` depends on numpy (not a built-in) and as Philipp points out we have cases where we treat tuples and lists differently. For parameters we would often enforce lists with ``param.List`` instead of a class selector which is what would have probably made more sense here.

jlstevens: Anyway, this is a minor point which we can decide on later and fix in a subsequent PR if we want to. I'll go ahead and merge.

jlstevens:
Looks good. Merging.

PR #2824, opened 2018-06-23
Update plotly/raster.py to fix Raster for plotly

thoth291:
This should fix an issue with Raster for plotly backend which was reported here:
https://github.com/ioam/holoviews/issues/2821

### DIFF
--- a/holoviews/plotting/plotly/raster.py
+++ b/holoviews/plotting/plotly/raster.py
@@ -2,7 +2,7 @@ import numpy as np
 import plotly.graph_objs as go
 
 from ...core.options import SkipRendering
-from ...element import Image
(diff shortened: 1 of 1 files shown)


philippjfr:
LGTM, will merge when tests pass.

thoth291:
if ci would be so kind and report which of the notebooks are failing to be tested - I would probably take a look at it...
Situation is getting even more strange given that this test failure is Python3 sensitive. Python2 test is passed with no issues.

philippjfr:
Don't worry about it, don't think it has anything to do with your change.

philippjfr:
I'll merge and fix it up later if the issue persists.

PR #2829, opened 2018-06-26
Fixed bug datashading NdOverlay with datetime axis

philippjfr:
The datashading code was not converting coordinates back to datetimes when applied to an NdOverlay. This PR simply ensures the same conversion is applied as in all other cases.

- [x] Fixes https://github.com/ioam/holoviews/issues/2819
- [x] Adds unit test


jlstevens:
Looks good. Merging.

PR #2830, opened 2018-06-26
Fix for matplotlib BoxWhisker and Violin overlays

philippjfr:
The legend code in the matplotlib looks at the artist in the ``handles`` dictionary and tries to hash it in order to generate a unique legend entry for each artist. In the case of BoxWhisker and Violin elements the artist is a dictionary which cannot be hashed. This PR adds each artist to the handles separately ensuring and aliases the 'boxes' and 'bodies' artists to the 'artist' handle, which allows the legend code to correctly process the legend for these elements.

- [x] Fixes https://github.com/ioam/holoviews/issues/2826
- [x] Adds unit tests


jlstevens:
Shame you have to list the artists explicitly but this is still better than it not working. Merging.

PR #2836, opened 2018-06-27
Removed bokeh 0.12.14 streaming fix

philippjfr:
Removes a fix introduced to patch over a bug in bokeh 0.12.14 since we now require bokeh >= 0.12.15.


philippjfr:
I'll just go ahead.

PR #2837, opened 2018-06-27
Allow setting HLine/VLine level in bokeh

philippjfr:
Makes sure that the overlay zorder is respected by default but old behavior can be restored.


jlstevens:
Nice simple fix. Merging.

PR #2838, opened 2018-06-27
Improved handling of scalar and non-scalar handling in Paths

philippjfr:
Ensures that the ``DictInterface.isscalar`` function also handles object types correctly and improves handling for non-scalar values in ContourPlot.


philippjfr:
Yes, let's merge.

PR #2842, opened 2018-06-28
Improve handling of zero range datashader aggregations

philippjfr:
Ensures datashader aggregation on zero-range data is handled correctly. Previously there were some workarounds that did one of two things:

1. If the range was zero-width then it was padded by an arbitrary factor of 0.5
2. If the width or height were 0 then it would simply increase it to 1

Both were only workarounds and led to unexpected behavior so the operations now return Images with zero sized arrays. This is also related to https://github.com/bokeh/datashader/pull/612 but doesn't require it since the operations are doing all the guarding against the bad cases that datashader doesn't handle correctly.

- [x] Fixes https://github.com/ioam/holoviews/issues/2817
- [x] Adds unit tests

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -126,16 +126,16 @@ class ResamplingOperation(LinkableOperation):
                 else:
                     x0, x1 = self.p.x_range
                     ex0, ex1 = element.range(x)
-                    x_range = np.max([x0, ex0]), np.min([x1, ex1])
-                if x_range[0] == x_range[1]:
-                    x_range = (x_range[0]-0.5, x_range[0]+0.5)
+                    x_range = (np.min([np.max([x0, ex0]), ex1]),
+                               np.max([np.min([x1, ex1]), ex0]))
 
                 if self.p.expand or not self.p.y_range:
                     y_range = self.p.y_range or element.range(y)
                 else:
                     y0, y1 = self.p.y_range
                     ey0, ey1 = element.range(y)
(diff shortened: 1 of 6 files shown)


jbednar:
Looks good, thanks.

jbednar:
New changes look good too.

philippjfr:
Ready to review.

jbednar:
Looks good to me, but would be helpful if @jlstevens looks it over too.

philippjfr:
One arbitrary thing I've done here is set the density for zero sized dimensions to 1, technically the density should be ``np.inf`` or something but that causes all kinds of trouble.

holoviews/operation/datashader.py
```diff
@@ -468,6 +458,27 @@ def _process(self, element, key=None):
             vdims = Dimension('Count')
         params['vdims'] = vdims
 
+        if x is None or y is None or width == 0 or height == 0:
+            xarray = xr.DataArray(np.full((height, width), np.NaN, dtype=np.float32),
```
jlstevens: Why does it have to be ``np.float32`` in these two cases?

philippjfr: No good reason, pointless optimization. Can take it out.

jlstevens: I wouldn't specify a precision unless there is a good reason.

jlstevens:
Looks good to me and the tests are passing. Merging.

PR #2843, opened 2018-06-28
Fixed bug rendering gif using mpl

philippjfr:
Fixes an annoying bug that's been causing tests to fail about 80% of the time recently. Not sure why or how it worked sometimes but the previous approach was definitely wrong.


jlstevens:
Looking at it, I am surprised it ever worked!

philippjfr:
It seems to still sporadically fail but seemingly less often (granted that's based on 3 observations).

jlstevens:
Best way to find out how often it is failing now is to merge. So I'll go ahead...

PR #2849, opened 2018-06-29
Do not sort by default when concatenating pandas data

philippjfr:
Avoids very large futurewarning when concatenating pandas datasets.


philippjfr:
>maybe when running nosetests or ideally when building docs?

Unlikely to catch anything since the tests pin libraries to versions that are known to be supported.

jlstevens:
> Unlikely to catch anything since the tests pin libraries to versions that are known to be supported.

True but we might want to consider adding some tests that check against the latest version of everything...

PR #2855, opened 2018-07-04
DictInterface only validates referenced columns

philippjfr:
Disables validation for columns which are not referenced as a dimension in the DictInterface. It can be useful to store other data in the dictionary and validation should not prevent non-standard data from being stored when HoloViews will just ignore that data anyway.


jlstevens:
I suppose that makes sense though you'll hit an error eventually if you decide to visualize an incorrectly formatted vdim (for instance). Anyway, it does seem worth being permissive about data not actually being used.

Ready to merge?

philippjfr:
Yes, ready.

PR #2860, opened 2018-07-06
Constant value datashader aggregation

philippjfr:
Further fixes for datashader aggregation on an axis with a constant value (i.e. zero-range). Also fixes a datetime issue and some issues when cloning Image types.

- [x] Fixes https://github.com/ioam/holoviews/issues/2858
- [x] Adds unit test


jlstevens:
Looks good although it might be nice to avoid subclassing the ``clone`` method if possible. Happy to merge once the test data is updated and the tests are passing.

philippjfr:
Ready to merge.

PR #2861, opened 2018-07-06
Fixed bug in frame key lookup

philippjfr:
Fixes issue when plotting multiple HoloMaps/DynamicMaps with non- or partially overlapping dimensions.


jlstevens:
Looks good. Merging.

PR #2869, opened 2018-07-11
Handle zero ranges correctly in regrid operation

philippjfr:
While I fixed the zero range handling in the aggregate operation I seemingly did not do it properly in the regrid operation. Additionally this PR allows Images to contain arrays with a (0, 0) shape, which previously required a hack which would instead store a (1, 1) image.

- [x] Fixes https://github.com/ioam/holoviews/issues/2817
- [x] Adds unit test

### DIFF
--- a/holoviews/core/data/image.py
+++ b/holoviews/core/data/image.py
@@ -156,14 +156,14 @@ class ImageInterface(GridInterface):
             l, b, r, t = dataset.bounds.lbrt()
             dim2, dim1 = dataset.data.shape[:2]
             xdate, ydate = isinstance(l, util.datetime_types), isinstance(b, util.datetime_types)
-            if l == r:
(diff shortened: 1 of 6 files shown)


stevievb:
@philippjfr that seemed better in that all of the images don't disappear when you zoom in but now there is a new problem. Once the image is entirely out of the viewport it disappears and never comes back.

philippjfr:
>Once the image is entirely out of the viewport it disappears and never comes back.

Odd, I haven't been able to reproduce that. I'll have another look.

philippjfr:
Nevermind ran into the issue in a completely separate context and fixed it. Thanks for your help @stevievb.

philippjfr:
This is ready to merge once tests pass.

jlstevens:
Looks good to me. Merging.

PR #2883, opened 2018-07-20
Flip y axis for QuadMesh as trimesh

jbednar:
Possible fix for https://github.com/pyviz/hvplot/issues/59


jbednar:
Some tests fail with this change, but I can't tell whether that means there are tests with bogus (flipped) data, or whether there are cases that *shouldn't* be flipped but now are being flipped.

philippjfr:
I think this probably has to inspect whether the ys are ascending or descending.

philippjfr:
Superseded by https://github.com/ioam/holoviews/pull/3017

Issue #2895, opened 2018-07-26
Accidentally using numpy.int32 for indexing on 64-bit windows


ceball:
I think that anywhere python's int or np.int_ is used as a dtype, that means on windows it'll end up as np.int32. numpy's default int type is C long, which on windows is 4 bytes (e.g. https://docs.scipy.org/doc/numpy-1.13.0/user/basics.types.html). At best this unnecessarily restricts sizes of stuff on windows (e.g. getting 'array size too large' or whatever), but could also lead to garbage results.

One example of the problem in hv, from https://github.com/ioam/holoviews/blob/master/holoviews/core/data/image.py#L82, on 64-bit windows: 

<img width="312" alt="screen shot 2018-07-26 at 10 17 30 am" src="https://user-images.githubusercontent.com/1929/43253557-2b395f86-90bd-11e8-88b6-478a933038e6.png">

* `a` shows that python int going in for the shape results in np.int32 coming back
* `b` shows a silent problem that will lead to garbage later
* `c` shows explicitly requesting output dtype of numpy's indexing integer type (C ssize_t), i.e. what I think is the fix in this case
* `d` just shows that if one of the dimensions is large enough, there won't be a problem

We should consider if there are other places where this kind of thing could be happening, and fix them all.

jbednar:
Thanks. 

For this particular problem, I think this code reproduces the issue:

```
import numpy as np, holoviews as hv
shape = (10000, 300000)
img = hv.Image(np.ndarray(shape))
print(shape, img.data.shape, type(img.data.shape[0]), img.shape, type(img.shape[0]))
```

If this is called on a Mac OS X platform and run using `python shape.py`, the result is:

```
(10000, 300000) (10000, 300000) <class 'int'> (3000000000, 3) <class 'numpy.int64'>
```

I don't have a Windows 10 platform accessible to test, but I'm told it produces numpy.int32 and a negative shape.  I had thought that making this change might help force the output type to numpy.int64, but apparently not:

```
shape = (np.int64(10000), np.int64(300000))
```

But adding `, dtype=np.intp` to https://github.com/ioam/holoviews/blob/master/holoviews/core/data/image.py#L82  does fix the problem?

jbednar:
As an aside, does anyone know why the shape is returned as a scalar 300000000 in the above case? It's clear what's happening [in the code](https://github.com/ioam/holoviews/blob/master/holoviews/core/data/image.py#L72):

```
[...]

PR #2903, opened 2018-07-31
Fix shape and length reporting of gridded interfaces for Python 3.6 on Win 64

ceball:
Part of #2895 (this PR just covers grid-based interfaces).

I used the below test files to check these changes. After switching to pyctdev and getting windows tests, these tests could be included in hv's test suite (but presumably re-written into class-based unit tests, which would reduce some boilerplate overlap). (Note for the future: the tests are intentionally set up to work on a machine with less than 4GB ram, for travis/appveyor.)

test_grid.py:
```python
import numpy as np, holoviews as hv
shape = (100, 1+np.iinfo(np.int32).max//100)
img = hv.Image( (range(shape[1]), range(shape[0]), np.empty(shape,dtype=np.uint8)), datatype=['grid'])
img = img.clone(datatype=['grid'])

print(shape, img.shape, type(img.shape[0]))
s = shape[0]*shape[1]

grid_shape = img.interface.shape(img)
# fails because Grid.shape() is wrong
assert grid_shape[0] == s, "%s != %s"%(grid_shape[0],s)
```

test_image.py:
```python
import numpy as np, holoviews as hv
shape = (100, 1+np.iinfo(np.int32).max//100)
img = hv.Image(np.ndarray(shape,dtype=np.uint8))


print(shape, img.data.shape, type(img.data.shape[0]), img.shape, type(img.shape[0]))
s = shape[0]*shape[1]

length = img.interface.length(img)
# fails because Image.length() is wrong
assert length==s, "%s != %s"%(length,s)
assert img.shape[0]==s, "%s != %s"%(img.shape[0],s)
```

test_iris.py:
```python
import numpy as np, holoviews as hv
shape = (100, 1+np.iinfo(np.int32).max//100)
img = hv.Image( (range(shape[1]), range(shape[0]), np.empty(shape,dtype=np.uint8)), datatype=['grid'])
img = img.clone(datatype=['cube'])

print(shape, img.data.shape, type(img.data.shape[0]), img.shape, type(img.shape[0]))
s = shape[0]*shape[1]

# fails because CubeInterface.length() is wrong
assert img.shape[0]==s, "%s != %s"%(img.shape[0],s)
```

test_xarray.py:
```python
import numpy as np, holoviews as hv
shape = (100, 1+np.iinfo(np.int32).max//100)
img = hv.Image( (range(shape[1]), range(shape[0]), np.empty(shape,dtype=np.uint8)), datatype=['grid'])
img = img.clone(datatype=['xarray'])

print(shape, img.shape, type(img.shape[0]))
s = shape[0]*shape[1]

# fails because XArrayInterface.shape() is wrong
assert img.shape[0]==s, "%s != %s"%(img.shape[0],s)
# fails because XArrayInterface.length() is wrong
assert img.interface.length(img)==img.shape[0] # TODO verify that definition of length is right
```

### DIFF
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -180,7 +180,7 @@ class GridInterface(DictInterface):
         if gridded:
             return shape
         else:
-            return (np.product(shape), len(dataset.dimensions()))
+            return (np.product(shape,dtype=np.intp), len(dataset.dimensions()))
 
 
     @classmethod
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -180,7 +180,7 @@ class GridInterface(DictInterface):
         if gridded:
             return shape
         else:
-            return (np.product(shape,dtype=np.intp), len(dataset.dimensions()))
+            return (np.product(shape, dtype=np.intp), len(dataset.dimensions()))
 
 
     @classmethod
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
   - conda update -q conda
   # Useful for debugging any issues with conda
   - conda info -a
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION flake8 scipy=1.0.0 numpy freetype nose pandas=0.22.0 jupyter ipython=5.4.1 "param<1.7" matplotlib=2.1.2 xarray networkx
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION flake8 scipy=1.0.0 numpy freetype nose pandas=0.22.0 jupyter ipython=5.4.1 param matplotlib=2.1.2 xarray networkx
   - source activate test-environment
   - conda install -c conda-forge filelock iris plotly=2.7 flexx ffmpeg netcdf4=1.3.1 --quiet
   - conda install -c bokeh datashader dask bokeh=0.12.15 selenium
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ install:
   - conda update -q conda
   # Useful for debugging any issues with conda
   - conda info -a
-  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION flake8 scipy=1.0.0 numpy freetype nose pandas=0.22.0 jupyter ipython=5.4.1 param matplotlib=2.1.2 xarray networkx
+  - conda create -q -n test-environment python=$TRAVIS_PYTHON_VERSION flake8 scipy=1.0.0 numpy freetype nose pandas=0.22.0 jupyter ipython=5.4.1 "param<1.7" matplotlib=2.1.2 xarray networkx
   - source activate test-environment
   - conda install -c conda-forge filelock iris plotly=2.7 flexx ffmpeg netcdf4=1.3.1 --quiet
   - conda install -c bokeh datashader dask bokeh=0.12.15 selenium
--- a/holoviews/core/data/image.py
+++ b/holoviews/core/data/image.py
@@ -79,7 +79,7 @@ class ImageInterface(GridInterface):
 
     @classmethod
     def length(cls, dataset):
-        return np.product(dataset.data.shape[:2],dtype=np.intp)
+        return np.product(dataset.data.shape[:2], dtype=np.intp)
 
 
     @classmethod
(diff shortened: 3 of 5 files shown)


holoviews/core/data/grid.py
```diff
@@ -180,7 +180,7 @@ def shape(cls, dataset, gridded=False):
         if gridded:
             return shape
         else:
-            return (np.product(shape), len(dataset.dimensions()))
+            return (np.product(shape,dtype=np.intp), len(dataset.dimensions()))
```
philippjfr: Prefer to leave a space between arguments. Same applies below.

philippjfr:
Looks good to me apart from a minor code formatting issue.

ceball:
Are the test failures my fault? I'll investigate if so (I couldn't run the test suite locally on the machine I was using for this PR).

holoviews/core/data/grid.py
ceball: Sorry, will fix those. 

hv has flake8 whitespace checking turned off on travis. I also couldn't see any instructions for contributors about whitespace (whether it's important, and if so, what the rules are). And when I skimmed the hv code, I saw inconsistent whitespace in general, including spacing between arguments. So I think that makes it a bit tricky for contributors to get whitespace right, unless I've failed to see something?

If you can point me to the rules hv's using, I'll make a PR in docs and/or testing, plus clean up the existing whitespace.

philippjfr:
The errors are pretty bizarre, it looks like the Dimension ``label`` parameters are set to None now when they should just mirror the ``name`` by default. Can't see what that has to do with the PR though. Is the PR otherwise up-to-date with master?

ceball:
> Is the PR otherwise up-to-date with master?

Probably not, as I did most of this last week. I'll investigate! 

Also so far I only ran my own tests on these changes, not the existing HV tests (and I only ran on windows).

philippjfr:
Don't think anything would have changed since last week. Really not sure what could have changed though.

holoviews/core/data/grid.py
philippjfr: We don't have consistent rules mainly (although I think we really should), but that's a whole argument to have. That said I'm very strongly in favor of separating arguments with whitespace and would be in favor of enabling this particular rule as a flake8 check immediately.

ceball:
I had a horrible thought on my way home...the errors somehow look like they'd be something more to do with recent param changes. However, I've just checked, and those changes to param were released before the end of June. There must have been successful tests of hv since then, right...?

ceball:
Hmm, my understanding had always been that holoviews is using param master on travis, but it's not. holoviews on travis gets its dependencies from either conda defaults, conda conda-forge, conda bokeh, or easy_install (via `python setup.py develop`) - varies by dependency. param's coming from conda defaults. And looks like param 1.7 made it to defaults 7 days ago. There hasn't been a successful travis build of holoviews since then, as far as I can tell.

ceball:
I've just pushed a commit to test that theory (I'm away for a while - will come back later).

ceball:
I was imagining something to do with the namespace stuff. I should be able
to look later (I'm away now).

jlstevens:
That is worrying indeed. That said, wouldn't param 1.7 have been pulled in for the scipy pyviz tutorial? It seems to have worked fine then...

ceball:
Preliminary thoughts...

The problem appears to be with the unpickled reference data, i.e. with backwards compatibility of pickles, rather than a problem with "current" hv or param. 

On unpickling, the recently added param namespace object is not being set up correctly from the state in the pickle (of the old param). Specifically, the param object's `self` isn't set (so is `None`). That of course leads to various problems (e.g. get_param_values(), used in hv's comparisons, reports the class's values rather than the instance's values).

To test that idea out, I thought it'd be a one line fix in Parameterized.__setstate__ to set self on the param object, but one or more hv things (e.g. LabelledData) appear to have their own independent state handling for pickle, and don't use Parameterized's, so I need to discuss if that's deliberate or an oversight. (Which I can do once I verify this really is the problem by creating a test case.)

I also need to see what param's current policy is regarding backwards compatibility of pickles (and consider if that's how it should be). I used to make pickles backwards compatible when param was part of topographica, but not sure if I ever moved any of that stuff to param itself, or what's happened since then...

ceball:
About the tests failing: moved to #2914

PR #2912, opened 2018-08-01
Fix clickable legends for some elements

ahuang11:
Test notebook
https://anaconda.org/ahuang11/test_muted_alpha/notebook


philippjfr:
Thanks, rebuilding the test data now. Will merge once it passes.

PR #2914, opened 2018-08-01
WIP: Demo unpickling of pre-namespace parameterized objects.

ceball:
WIP to see what changes are required to get hv data tests to pass. (The alternative would be updating the hv reference data.)

Explanation (copied from #2903)...

On unpickling, the recently added param namespace object is not being set up correctly from the state in the pickle (of the old param). Specifically, the param object's self isn't set (so is None). That of course leads to various problems (e.g. get_param_values(), used in hv's comparisons, reports the class's values rather than the instance's values).

To test that idea out, I thought it'd be a one line fix in Parameterized.setstate to set self on the param object, but one or more hv things (e.g. LabelledData) appear to have their own independent state handling for pickle, and don't use Parameterized's, so I need to discuss if that's deliberate or an oversight.

I also need to see what param's current policy is regarding backwards compatibility of pickles (and consider if that's how it should be). I used to make pickle loading backwards compatible with old pickles when param was part of topographica, but not sure if I ever moved any of that stuff to param itself, or what's happened since then...

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -754,7 +754,6 @@ class LabelledData(param.Parameterized):
         When unpickled, restore the saved style and plotting options
         to ViewableElement.options.
         """
-        self.param.self = self # TODO: super's setstate not called?
         d = param_aliases(d)
         try:
             load_options = Store.load_counter_offset is not None
@@ -781,7 +780,7 @@ class LabelledData(param.Parameterized):
         except:
             self.warning("Could not unpickle custom style information.")
         self.__dict__.update(d)
-
+        self.param.self = self # TODO: super's setstate not called?
 
 
 class Dimensioned(LabelledData):
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -754,6 +754,7 @@ class LabelledData(param.Parameterized):
         When unpickled, restore the saved style and plotting options
         to ViewableElement.options.
         """
+        self.param.self = self # TODO: super's setstate not called?
         d = param_aliases(d)
         try:
             load_options = Store.load_counter_offset is not None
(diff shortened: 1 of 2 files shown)


holoviews/core/dimension.py
```diff
@@ -780,7 +780,9 @@ def __setstate__(self, d):
         except:
             self.warning("Could not unpickle custom style information.")
         self.__dict__.update(d)
-
+        # TODO: super's setstate not called?
```
philippjfr: Seems like an oversight but I don't think the super would do anything else in this case unless I'm mistaken.

jlstevens:
@ceball Thanks for investigating this and identifying the error!

I may not be following everything exactly, but if I understand correctly, this PR is just a test and https://github.com/ioam/param/pull/248 is the real fix at the param level (hopefully!)?

Obviously we need pickles to work but ideally, I would have a chance to work on a more robust serialization format for holoviews objects so we could avoid using them whenever possible.

holoviews/core/dimension.py
philippjfr: Okay, thanks for explaining that. I have no objection to merging this without a super call due to the ``setattr`` issue. I can see how that would break on Overlay and Layout objects.

ceball:
I'm completely confused now about the hv test data; it doesn't seem to be the same every time. (Has it been updated? I can't tell from git.) Anyway, whatever happened has solved this problem for hv.

philippjfr:
>I'm completely confused now about the hv test data; it doesn't seem to be the same every time. (Has it been updated? I can't tell from git.)

I've sent you a message on slack notifying you of the test data update, which I had to do for a separate PR.

jlstevens:
With this PR and the param PR both closed, what is the correct fix? I do think we should keep pickle backwards compatibility - regenerating the test data simply masks the problem...

PR #2923, opened 2018-08-04
Update Customizing Plots Docs with more description about .options

ahuang11:
Somewhat addresses https://github.com/ioam/holoviews/issues/2916

This task turned out to be more difficult than I initially imagined (kudos to you guys for writing all the docs!)

There seems to be redundancy across places in the docs, and I'm not sure if it's appropriate to remove them or leave them there.  Maybe it's best to wait for https://github.com/ioam/holoviews/issues/2485 so it's not necessary to show other methods of customization in detail besides `.options` besides a brief mention? I think, that way, the documentation can be more straightforward and clear for the average user.

(I'm not sure why the muted alpha stuff is still showing  as commits if it's already merged into master).

Perhaps, it's more helpful to update the FAQ or create a cheat sheet first while the core developers can discuss the best way forward for the customization docs?


philippjfr:
Thanks so much for going through these. Would you mind clearing the notebooks? Makes it easier/possible to review the changes and avoids bloating the repository.

examples/user_guide/03-Customizing_Plots.ipynb
```diff
@@ -479,9 +556,51 @@
...
+    }
+   },
+   "types_to_exclude": [
+    "module",
+    "function",
+    "builtin_function_or_method",
+    "instance",
+    "_Feature"
+   ],
+   "window_display": false
```
philippjfr: Assume these are from various notebook extensions you have? Would you mind clearing these. Check out https://github.com/ioam/holoviews/issues/1617 for a small script to do this.

examples/user_guide/03-Customizing_Plots.ipynb
```diff
@@ -63,13 +63,87 @@
...
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "There is, however, one advantage in using the `%%opts` magic when working in the notebook environment, and that is %%opts magic allows tab-completion of the options which is currently unavailable with the `.options` method.\n",
+    "f\n",
+    "Additionally, across the docs, you may encounter a legacy method `.opts`, which is now superseded by the `.options` method. \n",
+    "\n",
+    "`.opts` is essentially the Python script version of `%%opts`: in both, you'll have to manually categorize whether an option is a style, plot, or norm option.\n",
+    "\n",
+    "Ultimately, `.options` method is preferred because using `.opts` requires extreme verbosity in nested dictionaries, which can get  overwhelming fast."
```
philippjfr: Small typo in form of a double space here.

philippjfr:
These changes look great to me, just had too small comments.

ahuang11:
I'm having trouble with using jq to clear the notebook output (presumably because I'm using a Windows Subystem Linux). Maybe you can help me do it?

I tried installing from `sudo apt-get install jq` and also `conda install jq`; both end up `jq: 1 compile error`

philippjfr:
For now you could just push the changes in this commit to your branch: https://github.com/ioam/holoviews/commit/68b23ed6c18a616f84ce2b32d6d6145d9cc183de

Not sure what's wrong with ``jq`` but it would be good to work out a cross-platform solution.

philippjfr:
I'm happy to see this merged but will give @jbednar and @jlstevens a chance to review.

examples/user_guide/03-Customizing_Plots.ipynb
```diff
@@ -63,13 +63,87 @@
...
+    "curve = hv.Curve( spike_train, 'milliseconds', vdims='Hertz')\n",
+    "spikes = hv.Spikes(spike_train, 'milliseconds', vdims=[])\n",
+    "(curve + spikes).cols(1)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Here it is, again, with the `.options` method, which is the simplest way of customizing because it will automatically deduce whether an option is a style, plot, or norm option. In addition, `.options` is portable across notebook environments and `.py` scripts, and also will be saved on export, unlike `%%opts`."
```
jbednar: What does it mean to be "saved on export"?  To a .py script?  If so, maybe "is portable across notebook environments and `.py` scripts, including exporting the notebook to a `.py` file, unlike `%%opts`.".

examples/user_guide/03-Customizing_Plots.ipynb
```diff
@@ -63,13 +63,87 @@
...
+    "spikes = hv.Spikes(spike_train, 'milliseconds', [])\n",
+    "\n",
+    "(curve.options(**curve_opts) + spikes.options(**spike_opts)).cols(1)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "When using ``.options`` to apply options directly to an individual object we do not have to explicitly declare which object the options apply to, however often it is useful to set options on a composite object. In these cases the options can be declared as a dictionary of the type name and the options. The code below is therefore equivalent to the syntax we used above:"
```
jbednar: @philippjfr, does .options only accept a type name, or does it also accept a fully qualified type.group.label string?

philippjfr: ``type.group.label`` is valid.

examples/user_guide/03-Customizing_Plots.ipynb
ahuang11: image

I remember when I first started using HoloViews, this was one of the main issues I encountered, and I had little clue on how to use port the magic to .py script.

jbednar: Ah, here you need to move `hv.renderer` to a subsequent cell.  The `%%opts` cell magic is only applied to the *output* of the cell, when it's displayed, not to everything in it.

examples/user_guide/03-Customizing_Plots.ipynb
jbednar: Ok, that's what I thought.  The text needs to be updated here to reflect that.

examples/user_guide/03-Customizing_Plots.ipynb
jbednar: Sorry I wasn't clear -- renderer.save() *should* keep *all* options that have been modified on the object; if it didn't, that would be a very strange bug.  But the point here is that you're doing renderer.save() on an object that has *not yet been modified by cell magic*, which is why you are failing to see any effect.  At that point, the cell magic application is still in the future! You have to do the renderer.save() *after* the cell has been displayed, because cell magics are applied only on the output of the cell, which hasn't yet occurred by the time you are saving it.  So all of the wordings you are suggesting are incorrect; %%opts *does* change the object in a way that will export; you just have to make sure that (a) the object is in fact affected by the cell magic (by returning it as the output of the cell), and (b) that you don't save it until *after* it has been returned as the output of the cell.  Does that make sense now?

ahuang11: Okay, thanks for clarifying! I see your point now and learned something new! I wonder if I'm the only one who encountered this before? If not maybe should have it as another FAQ?

```
Q: Why isn't my %%opts magic applied when exported?
A: You're probably doing renderer.save() on an object that has not yet been modified by cell magic, which is why you are failing to see any effect. At that point, the cell magic application is still in the future! You have to do the renderer.save() after the cell has been displayed, because cell magics are applied only on the output of the cell, which hasn't yet occurred by the time you are saving it. So all of the wordings you are suggesting are incorrect; %%opts does change the object in a way that will export; you just have to make sure that (a) the object is in fact affected by the cell magic (by returning it as the output of the cell), and (b) that you don't save it until after it has been returned as the output of the cell. Does that make sense now?
```
image

jbednar: Maybe make it more general:

Q: Why isn't my %%opts cell magic being applied to my HoloViews object?
A: %%opts is convenient because it tab-completes, but it can be confusing because of the "magic" way that it works.  Specifically, if you use it at the top of a Jupyter notebook cell, the indicated options will be applied to the *return value* of that cell, if it's a HoloViews object.  So, if you want a given object to get customized, you need to make sure it is returned from the cell, or the options won't *ever* be applied, and you should only access it *after* it has been returned, or the options won't *yet* have been applied.  For instance, if you use `renderer.save()` to export an object and only *then* return that object as the output of a cell, the exported object won't have the options applied, because they don't get applied until the object is returned (during IPython's "display hooks" processing).  So to make sure that options get applied, (a) return the object from a cell, and then (b) access it (e.g. for exporting) *after* the object has been returned.

jbednar: Also, another FAQ:

Q: Why are my `.options()` settings not having any effect?
A: By default, `.options()` returns a *copy* of your object, rather than modifying your original object. In HoloViews, making a copy of the object is cheap, because only the metadata is copied, not the data, and returning a copy makes it simple to work with a variety of differently customized versions of any given object.  You can pass `clone=False` to `.options()` if you wish to modify the object in place.

jbednar:
Thanks, @ahuang11!!

PR #2932, opened 2018-08-08
Add exception to the error message printed from DynamicMap

jbednar:
RIght now, if an error is raised in a DynamicMap, we get a warning, but it doesn't indicate anything about what the problem was:

```
import holoviews as hv, datashader as ds, pandas as pd, dask.dataframe as dd
from holoviews.operation.datashader import datashade, rasterize, dynspread
hv.extension('bokeh')
df = dd.read_parquet('../data/nyc_taxi_wide.parq').persist()
points = hv.Points(df, ['dropoff_x', 'dropoff_y'], 'dropoff_hour')
rasterize(points, aggregator=ds.last('dropoff_hour'))
```

```
WARNING:root:dynamic_operation: Exception raised in callable 'dynamic_operation' of type 'function'.
Invoked as dynamic_operation(height=400, scale=1.0, width=400, x_range=None, y_range=None)
```

It's quite unclear what a user is meant to do in this case.  With this change, the type and text of the exception will be shown, letting the user know what to do:

```
WARNING:root:dynamic_operation: Exception "NotImplementedError('last is currently implemented only for rasters',)" raised in callable 'dynamic_operation' of type 'function'.
Invoked as dynamic_operation(height=400, scale=1.0, width=400, x_range=None, y_range=None)
```

I don't think we want to show a full backtrace in this case, because there may be many such errors raised dynamically at different points of a DynamicMap's lifetime, but at least knowing the error is vastly better than just guessing...

### DIFF
--- a/examples/user_guide/03-Customizing_Plots.ipynb
+++ b/examples/user_guide/03-Customizing_Plots.ipynb
@@ -287,7 +287,7 @@
     "    plot:  {interpolation: 'steps-mid'}\n",
     "```\n",
     "\n",
-    "The reason HoloViews does not encourage the use of yaml is that yaml to literals and cannot express richer objects that can be used in options such as ``Cycle`` or ``Palette``."
+    "The reason HoloViews does not encourage the use of yaml is that yaml is limited to literals and cannot express richer objects that can be used in options such as ``Cycle`` or ``Palette``."
    ]
   },
   {
@@ -339,7 +339,7 @@
     "\n",
     "#### ``%%opts``\n",
     "\n",
-    "As shown in the examples above, customizes a particular HoloViws output displayed in a code cell. This application is not global and persists for that object which means settings stay in place if the object is re-displayed. Only accepts the string specification format. All cell magics need to appear *above* any code in the cells they are used in and cannot be used if there is no code in the cell. Only accepts the string specification syntax.\n",
+    "As shown in the examples above, customizes a particular HoloViews output displayed in a code cell. This application is not global and persists for that object which means settings stay in place if the object is re-displayed. Only accepts the string specification format. All cell magics need to appear *above* any code in the cells they are used in and cannot be used if there is no code in the cell. Only accepts the string specification syntax. Only applied to the output of the cell when it is displayed by IPython, and thus will not apply to objects exported or displayed using a separate mechanism before the cell completes, and does not affect objects created but not displayed in the output of the cell.\n",
     "\n",
     "#### ``%opts``\n",
     "\n",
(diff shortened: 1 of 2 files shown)


examples/user_guide/03-Customizing_Plots.ipynb
```diff
@@ -339,7 +339,7 @@
...
     "\n",
-    "As shown in the examples above, customizes a particular HoloViws output displayed in a code cell. This application is not global and persists for that object which means settings stay in place if the object is re-displayed. Only accepts the string specification format. All cell magics need to appear *above* any code in the cells they are used in and cannot be used if there is no code in the cell. Only accepts the string specification syntax.\n",
+    "As shown in the examples above, customizes a particular HoloViews output displayed in a code cell. This application is not global and persists for that object which means settings stay in place if the object is re-displayed. Only accepts the string specification format. All cell magics need to appear *above* any code in the cells they are used in and cannot be used if there is no code in the cell. Only accepts the string specification syntax. Only applied to the output of the cell when it is displayed by IPython, and thus will not apply to objects exported or displayed using a separate mechanism before the cell completes, and does not affect objects created but not displayed in the output of the cell.\n",
```
jbednar: Oops; I thought this was just a typo fix, but it's actually a clarification that I meant to have as a separate PR.  @jlstevens should verify the text here.

philippjfr:
Seems a bit redundant could we change the formatting to:

```
WARNING:root:dynamic_operation: NotImplementedError('last is currently implemented only for rasters') raised in callable 'dynamic_operation' of type 'function'.
```

jbednar:
What about just:
```
WARNING:root:dynamic_operation: NotImplementedError('last is currently implemented only for rasters').
```
(I'm not sure what knowing the type `function` does for us, and `dynamic_operation` seems to be printed already anyway.)

philippjfr:
Fine with me.

jbednar:
I changed it to:

> WARNING:root:dynamic_operation: Callable raised "NotImplementedError('first is currently implemented only for rasters',)".
> Invoked as dynamic_operation(height=400, scale=1.0, width=400, x_range=None, y_range=None)

I personally find it helpful if error messages can be found in the codebase, and here "Callable raised" is searchable to find where the error is raised (in the "Callable" class), while "first is currently" is searchable to find where it originated.  I would have liked to mention DynamicMap in the message, but it looks like Callable is called from many locations, so it isn't just for DynamicMap.

examples/user_guide/03-Customizing_Plots.ipynb
jlstevens: Looks correct to me.

jlstevens:
This PR seems fine once the merge conflict is resolved.

PR #2937, opened 2018-08-10
Add new drawing tool functionality

philippjfr:
This PR adds multiple drawing tool features I recently added to bokeh:

* Adds ``num_objects`` option to limit the number of glyphs that can be drawn
* Adds ``show_vertices`` option on ``PolyDraw`` enabling display of and snapping to vertices
* Adds new ``FreehandDraw`` stream

* Tests will have to wait until we have support for Selenium

- [x] Add documentation

### DIFF
--- a/examples/reference/streams/bokeh/FreehandDraw.ipynb
+++ b/examples/reference/streams/bokeh/FreehandDraw.ipynb
@@ -1,0 +1,103 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import holoviews as hv\n",
+    "from holoviews import streams\n",
+    "hv.extension('bokeh')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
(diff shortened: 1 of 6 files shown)


philippjfr:
There aren't really any tests I can add here without selenium. I'd very much like to set that up soon but I think we should switch to pytest first, since that will allow us to reuse a lot of bokeh's tooling.

So this is ready to review now.

examples/reference/streams/bokeh/FreehandDraw.ipynb
```diff
@@ -0,0 +1,103 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Alternatively we can use the ``element`` property to get an Element containing the returned data:"
+   ]
+  },
+  {
```
jlstevens: Please clear out the notebook metadata...

jlstevens:
Now the metadata is cleared from the new notebook, I'm happy to merge. No need to wait on tests as the last commit passed and the removal of the notebook metadata is the only change.

PR #2941, opened 2018-08-15
Faq updates

jbednar:
Minor updates.

### DIFF
--- a/doc/FAQ.rst
+++ b/doc/FAQ.rst
@@ -264,10 +264,15 @@ for that HoloViews object:
   
 .. code:: python
 
+  import holoviews as hv
+  from bokeh.models.tickers import FixedTicker
+  hv.extension('bokeh')
+  
   def update_axis(plot, element):
       b = plot.state
       b.axis[0].ticker = FixedTicker(ticks=list(range(0, 10)))
   
+  h = hv.Curve([1,2,7], 'x_col', 'y_col')
   h = h.options(finalize_hooks=[update_axis])
   h
 
--- a/doc/FAQ.rst
+++ b/doc/FAQ.rst
@@ -74,9 +74,6 @@ Example code below:
   %%opts Curve [width=1000]
   # preceding cell
   curve = hv.Curve([1, 2, 3])
-
-.. code:: python
-
   # next cell
   hv.renderer('bokeh').save(curve, 'example_curve')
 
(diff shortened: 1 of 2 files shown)


jbednar:
@ahuang11, does this existing FAQ answer address any of what you were proposing in **Q: How can I access all the options that aren't exposed in HoloViews, but are available in the backend?**?

```
  from holoviews import Store
  Store.add_style_opts(Image, ['filternorm'], backend='matplotlib')
```

ahuang11:
Ah, I wasn't aware of that, but that's pretty neat!

For the question `Q: How can I access all the options that aren't exposed in HoloViews, but are available in the backend?`, I used `finalize_hooks` / `final_hooks`  in the answer because it can be applied to more cases. For instance one example from gitter:

"I need plain integer number on x/y axis labels rather than scientific notation. How can I format the axes ?"
```
def _add_int_axis(plot, element):
    p = plot.state
    axis = p.xaxis[0]
    axis.ticker = FixedTicker(ticks=list(range(0, 10)))

hv_obj.options(finalize_hooks=[_add_int_axis])
```
However, I believe your answer is more suited for the question, and maybe my answer of using finalize_hooks can be put somewhere else?

ahuang11:
Also, I noticed some of the code is not showing up in the preview, but I can't find what's different about this code block from other ones...

image

image

jbednar:
@ahuang11 , I think I addressed your suggestions; can you please see if that's now clearer?

doc/FAQ.rst
```diff
@@ -232,20 +231,74 @@ to matplotlib's ``imshow`` command when displaying an ``Image``
...
+
+Here, you've wrapped your Bokeh-API calls into a function, then
+supplied that to HoloViews so that it can be run automatically
+whenever object ``h`` is viewed.
+
+  
+**Q: What I want to change is about how HoloViews works, not about the
+underlying backend.  Is that possible?**
+
+Sure, if you need more customization and configurability than is
```
ahuang11: I think missing `**A:**`

ahuang11:
Yes, it looks great! Thanks!

One additional suggestion I have; we probably should try our best to sort the FAQ by how frequent they're actually asked so that newer users don't have to read through the entire doc to find what they're looking for.

doc/FAQ.rst
jbednar: Fixed, thanks!

jbednar:
Right; the order is quite random right now. Ok, let's merge this one and then if you'd like to take a stab at identifying which are the most common questions and put those first, that can be a separate PR where it's just moving things around rather than adding new content.

ahuang11:
Sounds good!

PR #2942, opened 2018-08-15
Ensure initial PolyDraw and FreehandDraw data is projected

philippjfr:
When the PolyDraw and FreehandDraw stream is initialized the initial data stored in the ColumnDataSource is used to set the .data on the stream. If the data has been projected (as is the case in GeoViews) this initial data will be wrong, which means it needs to be passed through ``_process_msg`` first to ensure it is projected first.


philippjfr:
The code is under test, and the only way I can test these changes is in GeoViews so this is ready to merge now.

jlstevens:
The new calls to ``super`` makes sense and the change to ``initialize`` seems a reasonable way to allow geoviews to do an initial projection. Merging.

PR #2944, opened 2018-08-15
Multi interface improvements

philippjfr:


jlstevens:
Pulling out ``isscalar`` certainly makes sense and it is good to see ``add_dimension`` implemented. I don't quite get what is happening in ``callbacks.py`` so maybe it is worth adding a comment there.

As for supporting ``None`` as a scalar to the dict interface, that seems ok as long as it is documented somewhere with an explanation of what it means.

jlstevens:
After a bunch of transients, the tests have finally gone green. Merging.

PR #2950, opened 2018-08-20
Handle empty and flat distribution Violin plots

philippjfr:
- [x] Fixes https://github.com/ioam/holoviews/issues/2897
- [x] Add unit tests

### DIFF
--- a/holoviews/operation/stats.py
+++ b/holoviews/operation/stats.py
@@ -66,6 +66,7 @@ class univariate_kde(Operation):
 
         try:
             from scipy import stats
+            from scipy.linalg import LinAlgError
         except ImportError:
             raise ImportError('%s operation requires SciPy to be installed.' % type(self).__name__)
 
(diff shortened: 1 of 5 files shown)


philippjfr:
Ready to merge.

holoviews/operation/stats.py
```diff
@@ -100,8 +101,13 @@ def _process(self, element, key=None):
         elif bin_range[0] == bin_range[1]:
             bin_range = (bin_range[0]-0.5, bin_range[1]+0.5)
 
+        element_type = Area if self.p.filled else Curve
         data = data[np.isfinite(data)] if len(data) else []
         if len(data) > 1:
+            try:
+                kde = stats.gaussian_kde(data)
+            except LinAlgError:
```
jlstevens: Are flat and empty distributions the only things that would cause a ``LinAlgError``?

philippjfr: Only flat distributions as far as I know (as they result in a singular matrix), empty distributions are handled by the conditional on the line above.

jlstevens: Actually, now I'm thinking that the empty case handled in this PR is just for violin plots. But maybe there are still other things that cause this exception?

philippjfr: As I said I'm not aware of any other error conditions that would result in a linear algebra error, invalid datatypes would cause other errors and a singular matrix is the only error condition I'm aware of that could occur here.

jlstevens: Github didn't show me your second comment before I posted my second one. Seems fine then.

PR #2961, opened 2018-08-27
Add functionality to stream from parameterized objects and methods

philippjfr:
Requires https://github.com/ioam/param/pull/253 to work

- [x] Adds ability to directly stream from methods decorated with ``param.depends``
- [x] Adds ability from parameterized classes
- [x] DynamicMap allows declaring whether to dependencies on parameterized method
- [ ] Add unit test
- [ ] Update docs

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -17,7 +17,7 @@ from .layout import Layout, AdjointLayout, NdLayout, Empty
 from .ndmapping import UniformNdMapping, NdMapping, item_check
 from .overlay import Overlay, CompositeOverlay, NdOverlay, Overlayable
 from .options import Store, StoreOptions
-from ..streams import Stream, ParameterizedStream
+from ..streams import Stream, ParamStream
 
 
 
@@ -740,8 +740,11 @@ class DynamicMap(HoloMap):
 
     def __init__(self, callback, initial_items=None, streams=None, **params):
         streams = (streams or [])
-        if util.is_param_method(callback):
+
+        # If callback is a parameterized method and watch is disabled add as stream
+        if util.is_param_method(callback) and params.get('watch', True):
             streams.append(callback)
+
         if isinstance(callback, types.GeneratorType):
             callback = Generator(callback)
         elif not isinstance(callback, Callable):
@@ -752,11 +755,21 @@ class DynamicMap(HoloMap):
                          'and no longer needs to be specified.')
             del params['sampled']
 
+        # Validate streams by ensuring parameterized objects and methods are wrapped in ParamStream
         valid, invalid = [], []
+        parameterizeds = [s.parameterized for s in streams if isinstance(s, ParamStream)]
         for s in streams:
             if not isinstance(s, Stream):
-                if isinstance(s, param.Parameterized) or util.is_param_method(s):
-                    s = ParameterizedStream(s)
+                if isinstance(s, param.Parameterized):
+                    if s not in parameterizeds:
+                        s = ParamStream(s)
+                    else:
+                        continue
+                elif util.is_param_method(s):
+                    if not hasattr(s, "_dinfo") or util.get_method_owner(s) in parameterizeds:
+                        continue
+                    else:
+                        s = ParamStream(s)
                 else:
                     invalid.append(s)
                     continue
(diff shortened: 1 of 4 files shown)


philippjfr:
Tempted to leave testing and docs until there is a param release to update to but I'll defer to your judgement @jlstevens. If you agree this is ready to review/merge.

jlstevens:
I'm not sure whether I want to merge it yet but one thing that would make me happier is if there were a param version check so that the new execution path would only apply for current param dev versions or higher.

holoviews/core/spaces.py
```diff
@@ -749,14 +755,34 @@ def __init__(self, callback, initial_items=None, **params):
                          'and no longer needs to be specified.')
             del params['sampled']
 
-        super(DynamicMap, self).__init__(initial_items, callback=callback, **params)
-        invalid = [s for s in self.streams if not isinstance(s, Stream)]
+        # Validate streams by ensuring parameterized objects and methods are wrapped in ParamStream
```
jlstevens: How about a utility here? Something like:

```
valid, invalid = util.parameterized_streams(streams)
```

I'm not quite sure what you need `invalid` for though...

holoviews/core/util.py
```diff
@@ -1263,6 +1263,13 @@ def get_param_values(data):
     return params
 
 
+def is_param_method(obj):
```
jlstevens: Seems like something that could live in param...

holoviews/streams.py
```diff
@@ -547,6 +547,56 @@ def hashkey(self):
...
+            parameters = [p.name for p in parameterized.param.params_depended_on(method.__name__)]
+        elif parameters is None:
+            parameters = [p for p in parameterized.params()]
+
+        super(ParamStream, self).__init__(
+            parameterized=parameterized,
+            parameters=parameters, **params
+        )
+        for p in self.parameters:
+            self.parameterized.param.watch(self._listener, p)
```
jlstevens: I don't see any use of `unwatch` yet. Surely there are situations where clean up is needed?

jlstevens: Instead of defining ``_listener``, wouldn't a lambda be sufficient?

holoviews/streams.py
```diff
@@ -547,6 +547,56 @@ def hashkey(self):
...
+    def _listener(self, change):
+        self.trigger([self])
+
+    def update(self, **kwargs):
+        for k, v in kwargs.items():
+            setattr(self.parameterized, k, v)
+
+    @property
+    def contents(self):
+        if self.is_method: return {}
```
jlstevens: If it isn't a method, I assume the ``contents`` is based on what `ParamValues` used to do? I'm not sure yet, but I expected this stream to be just for triggering, i.e always return ``{}``. That might suggest to me that we can leave ``ParamValues`` as it was an make ``ParamStream`` its own thing...

jlstevens: In other words *maybe* parameterized methods should have their own stream class independent of arbitrary parameterized things...

holoviews/core/spaces.py
philippjfr: If you look below you'll see that invalid streams will raise errors. Happy to factor that out if you want.

holoviews/core/util.py
philippjfr: Agreed, I'll add that in a PR, I'm also duplicating it in ``panel``, so that would definitely make sense.

holoviews/core/spaces.py
jlstevens: Oops missed that. I think the utility could raise the exception for invalid inputs.

holoviews/streams.py
```diff
@@ -547,6 +574,81 @@ def hashkey(self):
...
+
+    @property
+    def contents(self):
+        return {}
+
+
+
+# Backward compatibility
+def ParamValues(*args, **kwargs):
+    param.main.warning('ParamValues stream is deprecated, use ParamStream instead.')
```
jlstevens: Now called ``Params``?

jlstevens:
Thanks for splitting it into ``Params`` and ``ParamMethod``: I think this is semantically cleaner.

I am now happy to merge even though there are two items left on the TODO list (unit tests and docs) as we can now test this out with a dev release. Of course, we will then have to address those two items (and any new issues we encounter) later.

PR #2962, opened 2018-08-28
Fix issue with value dimensions on PolyDraw stream

philippjfr:
This PR addresses issues when attaching a Polygons/Path element which contains value dimensions to a PolyDrawTool. The issue arises because value dimensions are only added to the CDS when they are actually needed but the drawing streams expects the data to be in the data it receives. This PR simply ensures that the stream itself adds any missing dimensions to the CDS.

- [x] Fixes bug in PolyDraw stream on element with value dimensions
- [x] Adds unit tests


philippjfr:
Ready to review and merge.

jlstevens:
Looks good. Merging.

PR #2973, opened 2018-09-01
Fix theme's title

ahuang11:
This should not be merged until bokeh's master is released because `from bokeh.themes import built_in_themes` doesn't exist yet in the latest stable version.

https://github.com/ioam/holoviews/issues/2871
image
image
image

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -10,7 +10,6 @@ from bokeh.models import (HoverTool, Renderer, Range1d, DataRange1d, Title,
 from bokeh.models.tickers import Ticker, BasicTicker, FixedTicker, LogTicker
 from bokeh.models.widgets import Panel, Tabs
 from bokeh.models.mappers import LinearColorMapper
-from bokeh.themes import built_in_themes
 try:
     from bokeh.models import ColorBar
     from bokeh.models.mappers import LogColorMapper, CategoricalColorMapper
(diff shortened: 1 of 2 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -355,12 +356,24 @@ def _title_properties(self, key, plot, element):
...
-        opts = dict(text=title, text_color='black')
+        theme = self.renderer.theme
+        if theme is not None:
+            if theme in built_in_themes:  # built-in-theme name
+                opts = built_in_themes[theme]._json['attrs']['Title']
+            else:  # theme object
+                opts = theme._json['attrs']['Title']
+            opts['text'] = title
+        else:
+            opts = dict(text=title, text_color='black')
```
ahuang11: I feel like there's a way to simplify this logic, but my brain isn't working right now...

philippjfr: Since the default text_color is black anyway I think we can simply remove this bit entirely.

ahuang11:
I feel like I tried to do that but the color wasn't black. I'll have to
test again.

philippjfr:
You are right, the default color is ``'#444444'``. Not sure what to think, it won't be that noticeable a difference, but it's also not ideal.

ahuang11:
It kind of was to me. Bold gray letters kind of is distracting especially
when everything else is black.

philippjfr:
Another option would be to define a default theme, which defines a black title.

PR #2975, opened 2018-09-01
Fix misleading example boxplot and change default

ahuang11:
Fixes misleading plot (title says colored by cylinder, but the coloring looks random) https://github.com/ioam/holoviews/issues/2974

Also, changes the default of box and violin's fill color from Cycle() to 'lightgray' because the colors don't really add any value unless there are multiple categories. See https://serialmentor.com/dataviz/color-pitfalls.html Figure 15.3

This has a misleading title
image

This is correct but looks a bit ugly to me
image

This looks the cleanest to me
image


jlstevens:
Makes sense to me and thanks for these improvements!

PR #2990, opened 2018-09-07
Split all value dimensions when splitting by path color

philippjfr:
When displaying a path with a varying value dimension used to color each line segment the path is split into little chunks. However previously other value dimensions would not be split in this way which would cause several issues:

1) When enabling hover on the path the CDS would end up with mismatched shapes
2) When enabling a drawing tool on the path the CDS would also end up with mismatching shapes

The solution is to split all value dimensions by default.

### DIFF
--- a/holoviews/plotting/util.py
+++ b/holoviews/plotting/util.py
@@ -805,7 +805,27 @@ def process_cmap(cmap, ncolors=None, provider=None, categorical=False):
 
 def color_intervals(colors, levels, clip=None, N=255):
     """
-    Maps a set of intervals to colors given a fixed color range.
+    Maps the supplied colors into bins defined by the supplied levels.
+    If a clip tuple is defined the bins are clipped to the defined
+    range otherwise the range is computed from the levels and returned.
+
+    Arguments
+    ---------
+    colors: list
+      List of colors (usually hex string or named colors)
+    levels: list or array_like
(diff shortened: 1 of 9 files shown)


philippjfr:
@jlstevens Ready to review and once merged a dev release would be helpful.

holoviews/plotting/util.py
```diff
@@ -822,8 +822,14 @@ def color_intervals(colors, levels, clip=None, N=255):
...
-        cmap = cmap[lidx:N-uidx]
-    return cmap
+        uidx = N-uidx
+        if lidx == uidx:
+            uidx = lidx+1
+        cmap = cmap[lidx:uidx]
+        if clmin == clmax:
+            idx = np.argmin(np.abs(np.array(levels)-clmin))
+            clip = levels[idx: idx+2] if len(levels) > idx+2 else levels[idx-1: idx+1]
+    return cmap, clip
```
jlstevens: Now the return value has changed (or at least now includes `clip`) it might be worth updating the docstring...

jlstevens:
This change makes a lot of sense. Other than one minor comment to address, I am happy to merge.

holoviews/plotting/util.py
philippjfr: Okay, done.

jlstevens:
Looks good to me.

PR #2996, opened 2018-09-14
Fixed plot extent calculation when apply_ranges=False

philippjfr:
Fixes small bug in extent calculations for plots with ``apply_ranges=False`` which skip extent calculation.

- [x] Add unit test


philippjfr:
@jlstevens This small fix is needed after the padding PR, currently things break if you overlay plots which all have ``apply_ranges=False`` set.

jlstevens:
Looks like a relatively small fix that now has associated tests. Merging.

PR #3023, opened 2018-09-29
Do not expand full DynamicMap cartesian product

philippjfr:
I previously removed code that computed the cartesian product of DynamicMap dimension values which is completely infeasible for large parameter spaces. It seems this was still being done for layout plots, I'll have to have another look but afaik I already ensured that the fact that these keys aren't expanded doesn't cause any issues.

### DIFF
--- a/holoviews/core/traversal.py
+++ b/holoviews/core/traversal.py
@@ -8,7 +8,7 @@ from collections import defaultdict
 from operator import itemgetter
 
 from .dimension import Dimension
-from .util import merge_dimensions, cartesian_product
+from .util import merge_dimensions
 
 try:
     import itertools.izip as zip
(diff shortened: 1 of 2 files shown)


philippjfr:
Had to remove existing unit tests but since this behavior was already the default behavior for non-layout plots I don't think there's much to worry about.

jlstevens:
Given layout plots were the remaining exception, this looks fine. I'll merge once the tests are green.

jlstevens:
Oddly, the pr build for python 2.7 seems to have this error:

image

philippjfr:
I think the environment simply got messed up here. We still want to update to scipy 1.1 but that seems unrelated to this PR. So this should be ready to merge now.

jlstevens:
All the Travis tests are green now. Merging.

PR #3025, opened 2018-09-30
Add normalization faq

ahuang11:
https://github.com/ioam/holoviews/issues/3016

### DIFF
--- a/doc/FAQ.rst
+++ b/doc/FAQ.rst
@@ -48,21 +48,42 @@ This same method is applicable to adjust the range of a color bar. Here
 z_col is the color bar value dimension and is bounded from 0 to 5.
 
 
-**Q: How do I keep all the axes limits consistent/normalized across
-all frames in a HoloMap/all elements in a layout?
+**Q: How do I control the auto-ranging/normalization of axis limits 
+across frames in a HoloMap or objects in a Layout?**
 
-**A:** You may use the use the normalization options as specified below:
+**A:** Where feasible, HoloViews defaults to normalizing axis ranges
+across all objects that are presented together, so that they can be 
+compared directly. If you don't want objects that share a dimension to 
+be normalized together in your layout, you can change the `axiswise`
+normalization option to True, making each object be normalized 
+independently:
 
 ..code:: python
-    # for consistent limits across frames
-    your_holomap.options(framewise=False)
+    your_layout.options(axiswise=True)
 
-    # for consistent limits across elements
-    your_layout.options(axiswise=False)
+Similarly, if you have a HoloMap composed of multiple frames in an
+animation or controlled with widgets, you can make each frame be normalized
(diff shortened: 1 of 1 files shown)


doc/FAQ.rst
```diff
@@ -48,6 +48,22 @@ This same method is applicable to adjust the range of a color bar. Here
...
+all frames in a HoloMap/all elements in a layout?
+
+**A:** You may use the use the normalization options as specified below:
+
+..code:: python
+    # for consistent limits across frames
+    your_holomap.options(framewise=False)
+
+    # for consistent limits across elements
+    your_layout.options(axiswise=False)
```
philippjfr: These are the defaults so it might be nice to expand on this and explain the axiswise=True and framewise=True cases. Particularly for DynamicMaps setting framewise=True means that the ranges are updated on every frame, which is a very frequently asked question.

jbednar: I'll reword it and push to the branch.

doc/FAQ.rst
```diff
@@ -48,6 +48,45 @@ This same method is applicable to adjust the range of a color bar. Here
...
+of objects that might be returned by future calls to a DynamicMap's 
+callback function, and so there is no way for it to implement 
+``axiswise=False`` normalization (even though such normalization 
+is otherwise normally the default in HoloViews). Thus as a special 
+case, a DynamicMap (whether created specifically or as the return
+value of various operations that accept a ``dynamic=True`` argument)
+always acts as if ``axiswise=True`` has been set on it.  If you want
+to normalize across all frames of a DynamicMap, you will need to manually
+determine the appropriate axis range yourself and set that, e.g. with 
+``.redim.range()`` as described above.
```
philippjfr: This is all very confused, I think where ever you say ``axiswise`` you actually mean ``framewise``. Additionally this should explain that ``framewise=False`` by default which means that if your axis ranges change between frames in a DynamicMap the axes won't update to match unless you set ``framewise=True``.

jbednar: Oops; I had to retype it and started saying axiswise instead of framewise!  Can you explain precisely what the behavior is in the last case? I thought the behavior was to act as if framewise  is True, but it sounds like instead it just takes the values from the first frame???

philippjfr: >I thought the behavior was to act as if framewise is True, but it sounds like instead it just takes the values from the first frame???

We cannot change the default of framewise depending on whether or not the object is in a HoloMap or DynamicMap, so this is indeed the case. I agree it's potentially very confusing and it is one of the most frequently asked questions, but without this behavior there would be no way of toggling whether the  ranges should be updated on each frame or not. This probably needs some significant thought but for now we need to document it better.

jbednar: Done; please re-review @philippjfr .

PR #3041, opened 2018-10-03
Fix layout titles

ahuang11:
https://github.com/ioam/holoviews/issues/3039

```
r.theme = 'caliber'  # with theme
hv.Curve([0, 1, 2])  # single no label
(hv.Curve([0, 1, 2]) * hv.Curve([4, 5, 6]))  # overlay no title
(hv.Curve([0, 1, 2]) * hv.Curve([4, 5, 6])).relabel('This title')  # labeled overlay
(hv.Curve([0, 1, 2]) + hv.Curve([4, 5, 6]))  # layout no title
(hv.Curve([0, 1, 2]) + hv.Curve([4, 5, 6])).relabel('This title')  # labeled layout
```
image
image

```
import holoviews as hv
hv.extension('bokeh')
r = hv.renderer('bokeh')
# ensure other plots still work without a theme
hv.Curve([0, 1, 2])  # single no label
(hv.Curve([0, 1, 2]) * hv.Curve([4, 5, 6]))  # overlay no title
(hv.Curve([0, 1, 2]) * hv.Curve([4, 5, 6])).relabel('This title')  # labeled overlay
(hv.Curve([0, 1, 2]) + hv.Curve([4, 5, 6]))  # layout no title
(hv.Curve([0, 1, 2]) + hv.Curve([4, 5, 6])).relabel('This title')  # labeled layout
```
image
image

Works for other layouts too:
Without theme:
image

With:
image

### DIFF
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -23,10 +23,10 @@ from .util import (layout_padding, pad_plots, filter_toolboxes, make_axis,
 
 from bokeh.layouts import gridplot
 from bokeh.plotting.helpers import _known_tools as known_tools
-try:
-    from bokeh.themes import built_in_themes
-except ImportError:
+if bokeh_version < '0.13.0':
     built_in_themes = {}
+else:
+    from bokeh.themes import built_in_themes
 
 TOOLS = {name: tool if isinstance(tool, basestring) else type(tool())
          for name, tool in known_tools.items()}
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -21,6 +21,7 @@ from .callbacks import Callback, LinkCallback
 from .util import (layout_padding, pad_plots, filter_toolboxes, make_axis,
                    update_shared_sources, empty_plot, decode_bytes)
 
+from bokeh.themes import built_in_themes
 from bokeh.layouts import gridplot
 from bokeh.plotting.helpers import _known_tools as known_tools
 
@@ -349,7 +350,7 @@ class CompositePlot(BokehPlot):
     to such a plot.
     """
 
-    fontsize = param.Parameter(default={'title': '16pt'}, allow_None=True,  doc="""
+    fontsize = param.Parameter(default={'title': '15pt'}, allow_None=True,  doc="""
        Specifies various fontsizes of the displayed text.
 
        Finer control is available by supplying a dictionary where any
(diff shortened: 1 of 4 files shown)


philippjfr:
Thanks for this, it looks good but will require a few changes. You're importing `built_in_themes` from bokeh which has only been added recently. To make this mergeable just put the import in a try/except block or import `bokeh_version` from `holoviews.plotting.bokeh.util` and check that the version meets the minimum requirement ('0.13.0' is just a guess):

```python
if bokeh_version < '0.13.0':
    built_in_themes = {}
else:
    from bokeh.themes import built_in_themes
```

holoviews/plotting/bokeh/plot.py
```diff
@@ -349,23 +350,50 @@ class CompositePlot(BokehPlot):
...
+        '"color:{color};font-family:{font};'
+        'font-style:{fontstyle};font-weight:{fontstyle};'  # italic/bold
+        'font-size:{fontsize}">'
+        '{title}</span>'
+    )
 
     def _get_title(self, key):
         title_div = None
         title = self._format_title(key) if self.show_title else ''
         if title:
```
philippjfr: To avoid the nesting I'd change this to:

```python
if not title:
    return title_div
```

and then un-indent the rest of the code in this section.

holoviews/plotting/bokeh/plot.py
```diff
@@ -23,6 +23,10 @@
 
 from bokeh.layouts import gridplot
 from bokeh.plotting.helpers import _known_tools as known_tools
+try:
+    from bokeh.themes import built_in_themes
```
ahuang11: philippjfr suggested:
```
if bokeh_version < '0.13.0':
    built_in_themes = {}
else:
    from bokeh.themes import built_in_themes
```

But I wasn't sure if could compare whether less than/greater than with a string so try and except seemed more straightforward to me

jbednar: bokeh_version (from plotting/bokeh/util.py) is of type [LooseVersion](http://epydoc.sourceforge.net/stdlib/distutils.version.LooseVersion-class.html), which does support comparisons with strings and should do the right thing.

jbednar: (note that I have no opinion between that and try/except; just pointing out that string comparison is a valid suggestion here.)

philippjfr: No objection to the try/except, but @jbednar is correct about string comparisons being supported.

ahuang11: Too late, learned something new heh

philippjfr:
There are a number of tests that need updates:

https://github.com/ioam/holoviews/blob/master/tests/plotting/bokeh/testlayoutplot.py#L31
https://github.com/ioam/holoviews/blob/master/tests/plotting/bokeh/testlayoutplot.py#L40
https://github.com/ioam/holoviews/blob/master/tests/plotting/bokeh/testlayoutplot.py#L57
https://github.com/ioam/holoviews/blob/master/tests/plotting/bokeh/testgridplot.py#L19
https://github.com/ioam/holoviews/blob/master/tests/plotting/bokeh/testgridplot.py#L29

philippjfr:
Thanks, looks good. The test failure is probably down to some change in the environment that seems to have broken matplotlib gif rendering.

PR #3042, opened 2018-10-03
Add tick formatter plot options

philippjfr:
Adds plot options to control the axis formatting directly rather than using the `Dimension.value_format` which only accepts formatter functions. These accept:

* Format strings like '%d', '%.3f' etc.
* Python functions which are compiled to JS with flexx (or converted to a FuncFormatter in mpl)
* TickFormatter instances (or matplotlib.ticker.Formatter instances for mpl)

- [x] Add unit tests
- [x] Add documentation (to styling plots notebook)

### DIFF
--- a/examples/user_guide/Plotting_with_Bokeh.ipynb
+++ b/examples/user_guide/Plotting_with_Bokeh.ipynb
@@ -144,32 +144,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "### Controlling axes"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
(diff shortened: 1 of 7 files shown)


philippjfr:
@jlstevens Ready to review and merge once tests are passing.

jlstevens:
Looks good! 

I feel that maybe we should have used bokeh's `PrintfTickFormatter` earlier but better late than never! I'll merge once the push build is green.

philippjfr:
>I feel that maybe we should have used bokeh's PrintfTickFormatter earlier but better late than never! 

Definitely, I've never been able to give a good answer to questions about tick formatting.

jlstevens:
Hmm, one thing...shouldn't matplotlib match the plot options here?

philippjfr:
>Hmm, one thing...shouldn't matplotlib match the plot options here?

Aren't they?

PR #3043, opened 2018-10-03
Minor fixes to examples and docs

jbednar:
- Fix "size" option that may have been accepted by some Matplotlib versions, but is supposed to be "s" as far as I know
- Clarify that we decided that "finalize_hooks" was the correct spelling, and that "final_hooks" is the deprecated spelling, even though we all prefer "final_hooks", because "finalize_hooks" has been accepted by both bokeh and mpl backends for many releases while "final_hooks" is only accepted by the mpl backend.


philippjfr:
Looks good.

philippjfr:
I've also pushed an update to the nyctaxi_hover app example to this PR because a) it was using geoviews and b) it no longer worked with latest geoviews.

PR #3045, opened 2018-10-04
Use robust isfinite utility throughout

philippjfr:
We use np.isfinite in many places across the code, however the numpy version of this function is not robust to many different types which is why we created the ``hv.core.util.isfinite`` utility. Unfortunately this is not yet used everywhere throughout the code, leading to bugs like https://github.com/ioam/holoviews/issues/3044 and https://github.com/ioam/holoviews/issues/3040

- [x] Add unit tests

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -810,20 +810,6 @@ def find_range(values, soft_range=[]):
             return (None, None)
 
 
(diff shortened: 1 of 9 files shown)


philippjfr:
Ready to review and merge.

jlstevens:

PR #3047, opened 2018-10-05
Address categorical labels not responding to fontsizes

ahuang11:
https://github.com/ioam/holoviews/issues/3007

Okay I think it's ready?
image

image

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -19,7 +19,7 @@ try:
     from bokeh.models.mappers import LogColorMapper, CategoricalColorMapper
 except ImportError:
     LogColorMapper, ColorBar = None, None
-from holoviews.plotting.bokeh.util import theme_attr_json
+
 from bokeh.plotting.helpers import _known_tools as known_tools
 
 from ...core import DynamicMap, CompositeOverlay, Element, Dimension
@@ -31,7 +31,7 @@ from ..util import dynamic_update, process_cmap, color_intervals
 from .plot import BokehPlot, TOOLS
 from .util import (mpl_to_bokeh, get_tab_title,  py2js_tickformatter,
                    rgba_tuple, recursive_model_update, glyph_order,
-                   decode_bytes, bokeh_version)
+                   decode_bytes, bokeh_version, theme_attr_json)
 
 property_prefixes = ['selection', 'nonselection', 'muted', 'hover']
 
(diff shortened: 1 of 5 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -434,6 +445,20 @@ def _axis_properties(self, axis, key, plot, dimension=None,
                 if jsfunc:
                     formatter = FuncTickFormatter(code=jsfunc)
                     axis_props['formatter'] = formatter
+
+        if axis == 'x':
```
philippjfr: If you use the `invert_axes` then the y-axis can be the categorical axis. Additionally the arguments already provide you with the desired axis, so I think this is sufficient:

```python
if isinstance(axis, CategoricalAxis):
    for key in list(axis_props):
        if key.startswith('major_label'):
            new_key = key.replace('major_label', 'group')
            axis_props[new_key] = axis_props[key]
```

You may also have to pop the old key from ``axis_props``, but since you weren't doing that maybe not.

holoviews/plotting/bokeh/element.py
```diff
@@ -389,7 +395,12 @@ def _axis_properties(self, axis, key, plot, dimension=None,
         Returns a dictionary of axis properties depending
         on the specified axis.
         """
-        axis_props = {}
+        try:
+            axis_props = (built_in_themes[self.renderer.theme]
```
philippjfr: If we'll be doing this a lot maybe we should have a utility to get the default properties for a particular model type.

holoviews/plotting/bokeh/element.py
ahuang11: Okay, but I believe I will still get the error:

```
/mnt/c/Users/user/GOOGLE~1/Bash/bokeh/bokeh/core/has_props.py in apply_theme(self, property_values)
    589         old_values = dict()
    590         for k in added.union(removed):
--> 591             old_values[k] = getattr(self, k)
    592 
    593         if len(property_values) > 0:

AttributeError: 'LinearAxis' object has no attribute 'group_text_font_style'
```

ahuang11: Also axis in this case is a string 'x' or 'y' I think...

philippjfr:
>If so, how should I approach it in the inputtable fontsize dictionary; major and minor ticks?

That makes sense to me.

philippjfr:
Looks good to me, ready to merge?

philippjfr:
Sorry for changing my mind here but could you change this to a try/except after all?

```python
if bokeh_version <= '0.13.0':
    built_in_themes = {}
else:
    from bokeh.themes import built_in_themes
```

ahuang11:
Looks good to me.

philippjfr:
Okay, let's merge once tests are passing.

jlstevens:
One build had a transient and needed to be restarted but all the other builds are now green. I'll go ahead and merge.

PR #3049, opened 2018-10-05
Improve Params stream handling

philippjfr:
This PR makes twofold improvements for Params streams:

* First of all it improves hashing on Params streams, disabling memoization temporarily when a ``param.Action`` is triggered.

* Just like a DynamicMap will listen to the dependencies of a parameterized method ``hv.util.Dynamic`` will do the same if it is passed a parameterized method as an operation.

- [x] Adds unit tests

### DIFF
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -609,7 +609,6 @@ class Params(Stream):
         return mapping
 
     def _listener(self, *events):
-        params = self.parameterized.params()
         self._memoize = not any(e.type == 'triggered' for e in events)
         self.trigger([self])
         self._memoize = True
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -597,7 +597,7 @@ class Params(Stream):
         super(Params, self).__init__(parameterized=parameterized, parameters=parameters, **params)
         self._memoize = True
         if watch:
-            self.parameterized.param.watch(self._listener, self.parameters)
+            self.parameterized.param.watch(self._watcher, self.parameters)
 
     def _validate_rename(self, mapping):
         for k, v in mapping.items():
@@ -608,7 +608,7 @@ class Params(Stream):
                                'stream parameter of the same name' % v)
         return mapping
 
-    def _listener(self, *events):
+    def _watcher(self, *events):
         self._memoize = not any(e.type == 'triggered' for e in events)
         self.trigger([self])
         self._memoize = True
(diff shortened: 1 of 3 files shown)


philippjfr:
@jlstevens Please review and then merge.

jlstevens:
After some discussion, I think we agreed that this PR should be built on top of the changes in https://github.com/ioam/param/pull/291.

philippjfr:
I had already updated it when you wrote that 😄

Just pushed a flake fix though, should be ready to merge.

holoviews/streams.py
```diff
@@ -608,8 +608,21 @@ def _validate_rename(self, mapping):
                                'stream parameter of the same name' % v)
         return mapping
 
-    def _listener(self, change):
```
jlstevens: I think it might be clearer if this method were called `watcher`.

holoviews/streams.py
```diff
@@ -595,9 +595,9 @@ def __init__(self, parameterized, parameters=None, watch=True, **params):
         if parameters is None:
             parameters = [p for p in parameterized.params() if p != 'name']
         super(Params, self).__init__(parameterized=parameterized, parameters=parameters, **params)
+        self._memoize = True
         if watch:
```
jlstevens: Little bit confused as the the `watch` argument: it doesn't seem to be state (either an attribute or a parameter) but is set through the constructor. Are there cases where you expect users to set `watch=False` even if watch could be supported by the param version?

If not, I would remove this argument so that the param version check is sufficient to know whether watch is being used or not (something you need to determine in `Dynamic` later...)

philippjfr: >Are there cases where you expect users to set watch=False even if watch could be supported by the param version?

It's used by the backward compatible `ParamValues` stream.

jlstevens: Alternatively, `watch` *could* be state which you would need to check in `Dynamic`...

jlstevens: > It's used by the backward compatible `ParamValues` stream.

(GitHub is laggy) In that case, shouldn't `watch` be an attribute to check in `Dynamic`?

philippjfr: Don't follow how does this relate to `Dynamic`? That's unrelated.

jlstevens: Nevermind, I get it now.

jlstevens:
I think this PR has now been greatly improved, thanks! I'll merge once the tests pass.

PR #3050, opened 2018-10-05
Change trimesh_rasterize value dimension handling

philippjfr:
Makes a number of changes and fixes to trimesh rasterization:

* A value dimension defined on the nodes takes precedence over one defined on the tris
* If an explicit aggregator is defined that column is actually respected


jlstevens:
Changes seem reasonable though are there any backwards compatibility implications of changing how the value dimensions are handled?

philippjfr:
>Changes seem reasonable though are there any backwards compatibility implications of changing how the value dimensions are handled?

Yes, but @jbednar and I agreed that this behavior is definitely preferable. We should simply make sure to document this in the release notes.

jlstevens:
Ok, as long as we don't forget to mention this in the release notes I'm fine with this change. Merging.

PR #3051, opened 2018-10-05
Drop gridplot width kwarg

philippjfr:
For some reason we were passing a total width to the bokeh gridplot function, which it simply ignored internally. In bokeh 1.0dev this hole in kwarg validation was fixed and passing in the width now errors, so this PR is required for forward compatibility.


jlstevens:
Looks good. Merging.

PR #3052, opened 2018-10-06
Moved tests inside holoviews package

philippjfr:
Moves tests inside the HoloViews package, which will allow upstream libraries (like bokeh) to run our test suite to ensure compatibility and downstream libraries (like geoviews) to reuse interface tests, which will allow us to finally move the iris interface.

- [x] Closes https://github.com/ioam/holoviews/issues/2580


philippjfr:
If this passes I'll probably just go ahead and merge so I can make progress on moving the iris interface.

PR #3055, opened 2018-10-06
Improve import times

philippjfr:
This PR attempts to improve import times by deferring imports of:

* xarray
* dask
* datashader and therefore numba (these were accidentally imported automatically in ``element/graphs.py``)

Here are the current import timings:

* Before this PR (with iris): 2.5 seconds
* Before this PR (without iris): 1.5 seconds
* After this PR: 0.55 seconds

In addition to some utilities to load the backends this PR works by declaring two new methods to `Interfaces`. The first method ``Interface.loaded`` is called to check whether an interfaces dependencies have been loaded, if not the interface is skipped, this defaults to true for backward compatibility. The second method ``Interface.applies`` is used to check whether the Interface applies to the supplied data object. By default this method simply checks whether the data is one of the ``Interface.types``, maintaining backward compatibility. However interfaces with heavy dependencies can subclass this method to defer the import.  

- [x] Closes https://github.com/ioam/holoviews/issues/1892


philippjfr:
Ready to review and merge.

philippjfr:
Just going to note that this PR shaved about 10 minutes off running nbsmoke on all our current notebook examples, user guides etc. (244 notebooks in total), unfortunately it's still slightly to long a job for travis.

philippjfr:
Also please take a second to appreciate the coverage of 88.888% 😃

jlstevens:
> ... unfortunately it's still slightly to long a job for travis.

So roughly how long does it take?

holoviews/core/data/__init__.py
```diff
@@ -8,8 +8,12 @@
...
 import param
 
-from ..dimension import redim
-from ..util import unique_iterator
+from ..dimension import redim, Dimension, process_dimensions
+from ..element import Element
+from ..ndmapping import OrderedDict
+from ..spaces import HoloMap, DynamicMap
+from ..util import (basestring, dimension_range as d_range, get_param_values,
+                    isfinite, process_ellipses, unique_iterator, wrap_tuple)
```
philippjfr: I had some weird issues when importing ``from .. import util`` getting the wrong utilities, hence I did this.

philippjfr:
>So roughly how long does it take?

Travis jobs time out at 50 minutes, by that time it's about 90% through the notebooks.

holoviews/core/util.py
```diff
@@ -1274,6 +1269,10 @@ def is_dataframe(data):
     """
     Checks whether the supplied data is of DataFrame type.
     """
+    if 'dask' in sys.modules:
+        import dask.dataframe as dd
+    else:
+        dd = None
```
jlstevens: Could also do:

```
dd = None
if 'dask' in sys.modules:
   import dask.dataframe as dd
```

Saves a line and maybe even more explicit that ``dd` always exists...

holoviews/core/data/util.py
```diff
@@ -0,0 +1,26 @@
...
+import numpy as np
+
+def get_array_types():
+    array_types = (np.ndarray,)
+    if 'dask' in sys.modules:
+        import dask.array as da
+        array_types += (da.Array,)
+    return array_types
+
+def get_dask_array():
```
jlstevens: I think `get_dask_array_module` or `dask_array_module` or even `da_module` or `get_da_module` would be clearer. You aren't getting the data type, you are getting a module...

philippjfr: Sounds good.

holoviews/core/data/__init__.py
```diff
@@ -210,7 +208,7 @@ class Dataset(Element):
 
     def __init__(self, data, kdims=None, vdims=None, **kwargs):
         if isinstance(data, Element):
-            pvals = util.get_param_values(data)
+            pvals = get_param_values(data)
```
jlstevens: Why remove the namespace qualification? Does it affect the import times? (It shouldn't!)

philippjfr: See my comment above, I can have another look at this but:

>I had some weird issues when importing from .. import util getting the wrong utilities, hence I did this.

holoviews/core/data/__init__.py
```diff
@@ -281,15 +279,15 @@ def range(self, dim, data_range=True, dimension_range=True):
...
+        elif all(isfinite(v) for v in dim.range) and dimension_range:
             return dim.range
         elif dim in self.dimensions() and data_range and len(self):
             lower, upper = self.interface.range(self, dim)
         else:
             lower, upper = (np.NaN, np.NaN)
         if not dimension_range:
             return lower, upper
-        return util.dimension_range(lower, upper, dim.range, dim.soft_range)
+        return d_range(lower, upper, dim.range, dim.soft_range)
```
jlstevens: I think I prefer the old version...I don't see why the name `d_range` should be introduced for a single use (at least in this PR diff...)

holoviews/core/data/xarray.py
```diff
@@ -4,28 +4,34 @@
...
 
 class XArrayInterface(GridInterface):
 
-    types = (xr.Dataset, xr.DataArray)
+    types = ()
 
     datatype = 'xarray'
 
+    @classmethod
+    def loaded(cls):
```
jlstevens: This method makes sense but based on the way it is used, shouldn't it be called `available` instead of `loaded`? I thought the idea was that you can check `sys.modules` for availability and it is only loaded once actually imported...

holoviews/core/data/__init__.py
jlstevens: I see. I would prefer to figure out the issue rather than switching to the unqualified version...

holoviews/core/data/__init__.py
philippjfr: Again, see the comment above.

holoviews/core/data/xarray.py
philippjfr: Slightly confused here, what is the difference between the semantics of ``available`` and ``loaded`` in your mind? Loaded currently checks if the dependency has been loaded, it does not check whether the library is importable (which is my interpretation of "available") because there is no way to check whether a library is importable without importing it.

jlstevens: I understand it now, `loaded` is correct.

holoviews/core/data/__init__.py
jlstevens: Well, really it is two things:

1. For some reason you have problems with the qualified import (the root problem which would ideally be fixed)
2. You renamed to `d_range` to avoid a local variable name clash.

This PR would be greatly simplified if 1 can be fixed!

holoviews/core/util.py
jlstevens: This isn't a strong opinion but I do think my suggested version is slightly better. Don't you?

holoviews/core/data/__init__.py
philippjfr: Having nightmares with this, but I should be able to fix it.

holoviews/core/util.py
philippjfr: Basically have no preference here.

jlstevens:
I've made a few suggestions and everything is clearer now that I understand you expect that the user will populate `sys.modules` (unless they use an interface with a literal constructor).

The main issue is the removal of the qualified namespace. Restoring that would greatly simplify this PR and keep things cleaner overall imho.

Otherwise, I'm very happy with the import time speed up and am happy to merge.

jlstevens:
> Travis jobs time out at 50 minutes, by that time it's about 90% through the notebooks.

I think things would be better if we could at least nbsmoke the user/getting started guides. If we skip the elements (or maybe only do those on doc builds?) can we get it to around 20 minutes (i.e around how long it takes to run test builds now)

philippjfr:
>I think things would be better if we could at least nbsmoke the user/getting started guides. If we skip the elements (or maybe only do those on doc builds?) can we get it to around 20 minutes (i.e around how long it takes to run test builds now)

I think we can probably get them all working in a travis cron job.

jlstevens:
All looks good now, thanks! I'll merge when the tests pass.

jbednar:
> Travis jobs time out at 50 minutes, by that time it's about 90% through the notebooks.

Can we split them into two separate jobs in the .travis.yml, or is Travis counting the entire time of all the CI targets?

philippjfr:
>Can we split them into two separate jobs in the .travis.yml, or is Travis counting the entire time of all the CI targets?

We can do that but that'll probably wait until we've set up the new architecture. Don't really want to make the travis.yml more complex only to then have to rewrite it all.

PR #3056, opened 2018-10-06
Simplify CI setup and environment

philippjfr:


stonebig:
pillow-5.3.0 is not ok ?

philippjfr:
We're seeing some issues with matplotlib gif rendering. For now this is just a test if this is really the cause.

philippjfr:
Doesn't appear to be the issue.

philippjfr:
The issue is probably an change in matplotlib 2.2.3 in that case.

philippjfr:
@jbednar @jlstevens I haven't been able to track down precisely what the cause of the gif export error was, pinning matplotlib and pillow to older versions individually did not work but now I'm pinning both of them it is working again. As such I'm going to merge this PR soon, but I'll quickly describe what I've done.

This PR is intended as a stepping stone towards slowly moving to @ceball fabulous new architecture. I've moved all the dependencies out of the .travis.yml into the environment.yml which has long been unused. The environment.yml is much more readable and makes it clear just how much we're pinning right now. We should try to slowly lift the pinned versions while keeping the tests passing. Once we can have our testing setup work on the latest releases the transition to the new architecture should be much easier and cleaner.

One thing to note here is that the environment.yml does not define a python version, which means that at least on travis the only way we are getting both py2 and py3 tests is that we are installing Miniconda and Miniconda3, which then determines the python version of the environment while the exact ``TRAVIS_PYTHON_VERSION`` env var is not respected. This should be safe in the near term but it is only a temporary solution until we switch over to the new setup.

jlstevens:
Just a quick comment to say I am happy with the changes Philipp made in this PR: we want to minimize travis.yml as much as possible!

PR #3066, opened 2018-10-09
Improved handling of the bokeh root model

philippjfr:
On bokeh plots the root model is required to set up comm callbacks. Currently this is handled in some pretty terrible ways, e.g. after subplots set up a comm callback any layout plot or similar that is wrapping the plot has to then search and replace the temporary root model id with the actual root model id. By enabling the ability for `BokehPlot` instances to store the model id these horrible operations are no longer necessary, which also allows external libraries such as ``panel`` to declare the root model and avoids them from also having to replicate code that searches all subplots to replace the root model id.


philippjfr:
Ready to review and merge.

PR #3074, opened 2018-10-10
Ensure toolbar=None does not disable tools

philippjfr:
As discussed in https://github.com/ioam/holoviews/issues/3006 setting toolbar=None should not disable all the tools, it should simply hide the toolbar. I've added a toolbar='disable' option instead.

- [x] Closes https://github.com/ioam/holoviews/issues/3006


jlstevens:
Looks good! Merging.

PR #3075, opened 2018-10-10
Fixed bug in HLine/VLine datetime handling

philippjfr:
Apparently there is no need to transform datetime types in the HLine/VLine plotting code and getting rid of this processing also fixes timezone issues.

- [x] Closes https://github.com/ioam/holoviews/issues/2857


jlstevens:
Looks good. Merging.

PR #3081, opened 2018-10-10
Ensure QuadMesh with xarray handles datetime range

philippjfr:
When using the xarray interface with a QuadMesh the range method would break on the datetimes.

- [x] Adds unit tests


jlstevens:
Looks good. I'll merge when the tests pass.

PR #3090, opened 2018-10-17
Handled exception on LinkedStream callbacks

philippjfr:
Ensure errors in Stream subscribers do not cause Stream metadata to stick around incorrectly.


jlstevens:
Looks good. Merging.

PR #3092, opened 2018-10-19
Add support for Polygons with holes

philippjfr:
In conjunction with the corresponding geoviews PR this represents a full protocol for round-tripping data between the dictionary format in holoviews and shapely geometries. Therefore this formally defines the conventions of our format relative to GEOS geometry definitions.

* A ``MultiInterface`` can hold a ``list`` of standard columnar data structures, the columnar data structures may be, numpy arrays, dictionaries of columns, (pandas/dask) dataframes and dictionaries of a geometry and other data.
* Each item in a ``MultiInterface`` ``list`` represents an individual geometry, where a geometry can be a LinearString, LinearRing, Polygon, MultiLineString or MultiPolygon.
* If an item in the list represents a MultiLineString or MultiPolygon then each polygon/line string should be separated by nans.
* In order to store holes for MultiPolygons the holes are deeply nested to unambiguously assign each hole to each polygon:
   - **1st.** The first level of nesting corresponds to the list of geometries
   - **2nd.** The second level corresponds to each Polygon in a MultiPolygon
   - **3rd.** The third level of nesting allows for multiple holes per Polygon
 
To demonstrate this let's take a few examples of polygon geometries with holes. In the simplest case we have a Polygon of a few coordinates with holes defined as a list of list of arrays:

```python
coords = [(1, 2), (2, 0), (3, 7)]
holes = [[[(1.5, 2), (2, 3), (1.6, 1.6)], [(2.1, 4.5), (2.5, 5), (2.3, 3.5)]]]
hv.Polygons([{('x', 'y'): coords, 'holes': holes}])
```

bokeh_plot

In the case of a ``MultiPolygon`` which contains the Polygon from above it becomes clearer. The coords now contain a nan separator and there are two separate lists of holes, one the same as above the second empty:

```python
coords = [(1, 2), (2, 0), (3, 7), (np.nan, np.nan), (6, 7), (7, 5), (3, 2)]
holes = [[[(1.5, 2), (2, 3), (1.6, 1.6)], [(2.1, 4.5), (2.5, 5), (2.3, 3.5)]], []]
hv.Polygons([{('x', 'y'): coords, 'holes': holes}])
```

bokeh_plot

To further illustrate this, we can split the ``MultiPolygon`` into two separate Polygons like this:

```
coords = [(1, 2), (2, 0), (3, 7)]
holes = [[[(1.5, 2), (2, 3), (1.6, 1.6)], [(2.1, 4.5), (2.5, 5), (2.3, 3.5)]]]
coords2 = [(6, 7), (7, 5), (3, 2)]
hv.Polygons([{('x', 'y'): coords, 'holes': holes, 'value': 1},
             {('x', 'y'): coords2, 'value': 2}], vdims='value')
```

bokeh_plot

producing almost the same plot but allowing us to set two distinct values for each polygon, when in the MultiPolygon case they would have had to share the same value.

This scheme can also be used to faithfully round-trip matplotlib geometries (e.g. Patches) to this format, which means we can finally handle the ``contours`` operation correctly.

- [x] Closes https://github.com/ioam/holoviews/issues/2017
- [x] Rewrite contours operation to convert matplotlib Patches correctly
- [x] Added [Geometry data user guide](https://anaconda.org/philippjfr/geometry_data/notebook)
- [x] Add unit tests

### DIFF
--- a/examples/user_guide/Geometry_Data.ipynb
+++ b/examples/user_guide/Geometry_Data.ipynb
@@ -1,0 +1,271 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In addition to the two main types of data, namely tabular/columnar and gridded data HoloViews also provide extensible interfaces to represent path geometry data. Specifically it has three main element types used to representing different types of geometries. In this section we will cover the HoloViews data model for representing different kinds of geometries.\n",
+    "\n",
+    "There are many different ways of representing path geometries but HoloViews' data model is oriented on GEOS geometry definitions and allows faithfully round-tripping data between its element types and GEOS geometry definitions such as ``LinearString``, ``Polygon``, ``MultiLineString`` and ``MultiPolygon`` geometries (even if this is not implemented in HoloViews itself). Since HoloViews interfaces are extensible many different formats for representing geometries could be supported (see [GeoViews](http://geoviews.org/user_guide/Geometries.html) for other representations) but here we will cover the native formats used by HoloViews to represent this data."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import numpy as np\n",
+    "import holoviews as hv\n",
+    "\n",
+    "hv.extension('bokeh')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Representing paths\n",
+    "\n",
+    "The ``Path`` element represents a collection of path geometries with associated values. Each path geometry may be split into sub-geometries on NaN-values and may be associated with scalar values or array values varying along its length. In analogy to GEOS geometry types a Path is a collection of LineString and MultiLineString geometries with associated values.\n",
+    "\n",
+    "While many different formats are accepted in theory, natively HoloViews provides the ``MultiInterface`` which allows representing paths as lists of regular columnar data objects including arrays, dataframes and dictionaries of column arrays and scalars. A simple path geometry may therefore be drawn using:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "hv.Path([{'x': [1, 2, 3, 4, 5], 'y': [0, 0, 1, 1, 2]}]).options(padding=0.1)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Here the dictionary of x- and y-coordinates could also be an NumPy array with two columns or a dataframe with 'x' and 'y' columns. Since the format supports lists any number of geometries may be drawn in this way. Additionally, it is also possible to associate a value with each path by declaring it as a value dimension:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "hv.Path([{'x': [1, 2, 3, 4, 5], 'y': [0, 0, 1, 1, 2], 'value': 0},\n",
+    "         {'x': [5, 4, 3, 2, 1], 'y': [2, 2, 1, 1, 0], 'value': 1}], vdims='value').options(padding=0.1, color_index='value')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "#### Multi-geometry"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Splitting the geometries in this way allows assigning separate values to each geometry, however often multiple geometries share the same value in which case it may be desirable to represent them as a multi-geometry by combining the coordinates and separating them by a NaN value:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "hv.Path([{'x': [1, 2, 3, 4, 5, np.nan, 5, 4, 3, 2, 1],\n",
+    "          'y': [0, 0, 1, 1, 2, np.nan, 2, 2, 1, 1, 0], 'value': 0}],\n",
+    "        vdims='value').options(padding=0.1, color_index='value')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "This represents a more efficient format particularly when there are very many small geometries with the same value.\n",
+    "\n",
+    "#### Scalar vs. continuously varying value dimensions\n",
+    "\n",
+    "Unlike ``Contours`` which are limited to representing iso-contours or isoclines, i.e. a function of two variables which describes a curve along which the function has a constant value, a ``Path`` element may also have continuously varying values along its path. Below we will declare a path with a value that varies along its path:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "a, b, delta = 3, 5, np.pi/2.\n",
+    "\n",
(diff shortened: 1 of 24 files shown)


jsignell:
Are you losing the mapping from polygon (section of multipolygon) to hole in that transformation?

jsignell:
nevermind I misread. I think that seems like a reasonable representation. It might be worth looking at geopandas though and trying to make that work well

philippjfr:
>nevermind I misread. I think that seems like a reasonable representation. It might be worth looking at geopandas though and trying to make that work well

Geopandas support will follow in geoviews, but there we don't have to worry about the representation since geopandas already represents holes.

jsignell:
> there we don't have to worry about the representation since geopandas already represents holes.

Oh I see, so we won't transform it to this style?

philippjfr:
>Oh I see, so we won't transform it to this style?

No, it should generally stay in its native format, i.e. inside shapely objects.

philippjfr:
Slight correction now that I've started on the geoviews representation, in geoviews paths and polygons are converted to lists of dictionaries containing shapely geometries, which simplifies projections massively.

holoviews/core/data/dictionary.py
```diff
@@ -84,32 +85,39 @@ def init(cls, eltype, data, kdims, vdims):
...
-                        for sd in d:
-                            unpacked.append((sd, np.array([], dtype=vals.dtype)))
-                    elif not vals.ndim == 2 and vals.shape[1] == len(d):
-                        raise ValueError("Values for %s dimensions did not have "
-                                         "the expected shape.")
-                    else:
-                        for i, sd in enumerate(d):
-                            unpacked.append((sd, vals[:, i]))
+
+        unpacked = []
```
philippjfr: Note that I simply unindented this section when I thought I'd have to complicate it further, there are no changes to ``DictInterface.init`` that need reviewing.

holoviews/core/data/dictionary.py
```diff
@@ -151,7 +159,12 @@ def isscalar(cls, dataset, dim):
         values = dataset.data[name]
         if isscalar(values):
             return True
-        unique = set(values) if values.dtype.kind == 'O' else np.unique(values)
+        if values.dtype.kind == 'O':
+            unique = set(values)
+        else:
+            unique = np.unique(values)
+            if (~util.isfinite(unique)).all():
```
philippjfr: A column of NaNs was not being detected as being scalar.

examples/user_guide/Geometry_Data.ipynb
```diff
@@ -0,0 +1,273 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Note that this conversion may be lossy if the converted format has no way of representing 'holes' or other data."
+   ]
+  }
+ ],
+ "metadata": {
```
jlstevens: Remember to clear out this metadata before this PR can be merged...

examples/user_guide/Geometry_Data.ipynb
```diff
@@ -0,0 +1,273 @@
...
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Representing paths\n",
+    "\n",
+    "The ``Path`` element represents a collection of path geometries with associated values. Each path geometry may be split into sub-geometries on NaN-values and may be associated with scalar values or array values varying along its length. In analogy to GEOS geometry types a Path is a collection of LineString and MultiLineString geometries with associated values.\n",
+    "\n",
+    "While many different formats are accepted in theory, natively HoloViews provides the ``MultiInterface`` which allows representing paths as lists of regular columnar data objects including arrays, dataframes and dictionaries of column arrays and scalars. A simple path geometry may therefore be drawn using:"
```
jlstevens: Are interfaces discussed elsewhere in the user guide? If not, I wouldn't mention `MultiInterface` explicitly, if so, I might want to point to where interfaces are discussed.

examples/user_guide/Geometry_Data.ipynb
```diff
@@ -0,0 +1,273 @@
...
+    "]\n",
+    "\n",
+    "hv.Polygons([{'x': xs, 'y': ys, 'holes': holes}]).options(padding=0.1)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "If a polygon has no holes at all the 'holes' key may be ommitted entirely:"
```
jlstevens: Maybe worth doing this first (no holes) before introducing holes? I.e a reminded of the `Polygon` constructor before holes were supported...

examples/user_guide/Geometry_Data.ipynb
```diff
@@ -0,0 +1,273 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "poly = hv.Polygons([\n",
+    "    {'x': xs, 'y': ys, 'holes': holes, 'value': 0},\n",
+    "    {'x': [4, 6, 6], 'y': [0, 2, 1], 'value': 1}\n",
+    "], vdims='value')\n",
+    "\n",
+    "polys = poly.split()\n",
```
jlstevens: Don't think it is worth introducing the `polys` handle here (just inline it to the `Layout`). Looks to me like the handle that is reused is just `poly`.

holoviews/plotting/mpl/util.py
```diff
@@ -185,3 +186,33 @@ def get_raster_array(image):
         else:
             data = np.flipud(data)
     return data
+
+def ring_coding(array):
```
jlstevens: Is this called `ring` because it is closed (or at least looks closed)? If so the docstring should mention this, if not, I'm not sure what 'ring' is referring to. If I remember right, matplotlib has explicit codes for declaring genuinely closed paths that I assume you've considered using...

holoviews/core/data/dictionary.py
```diff
@@ -358,5 +377,22 @@ def iloc(cls, dataset, index):
...
+        key = Polygons._hole_key
+        return key in dataset.data and isinstance(dataset.data[key], list)
+
+    @classmethod
+    def holes(cls, dataset):
+        from holoviews.element import Polygons
+        key = Polygons._hole_key
+        if key in dataset.data:
+            return [[[np.asarray(h) for h in hs] for hs in dataset.data[key]]]
+        else:
```
jlstevens: Why can't you use the super definition of `holes` if holes aren't defined/supported?

jlstevens: Unless the default implementation works for all interfaces except this one specifically?

philippjfr: Probably can do that.

jlstevens:
I've made a few comments but overall this PR looks good. This feature has been a long time coming!

holoviews/plotting/mpl/util.py
philippjfr: It generates the path codes for each of the exterior and interior paths (which are always rings in a polygon). Using CLOSEPOLY path codes is not necessary since PathPatch will automatically close any patches.

examples/user_guide/Geometry_Data.ipynb
philippjfr: Sure.

examples/user_guide/Geometry_Data.ipynb
philippjfr: Sounds good.

examples/user_guide/Geometry_Data.ipynb
philippjfr: Will do.

examples/user_guide/Geometry_Data.ipynb
philippjfr: Okay removing this.

examples/user_guide/Geometry_Data.ipynb
philippjfr: Actually I think saying that the format is exactly the same as the paths one is enough, otherwise it gets a bit repetitive.

jlstevens: Sure.

jlstevens:
The tests have now passed (finally!). All looks good to me.

PR #3096, opened 2018-10-23
Fixed bug setting integer number of contour levels

philippjfr:
Small bug I introduced in the recent geometry handling PR, for older versions of matplotlib an integer number of contours needs to be passed in as a positional argument.


jlstevens:
Thanks for the fix. The tests have now passed.

PR #3099, opened 2018-10-23
Implement MultiInterface groupby

philippjfr:
This PR implements groupby operations for the MultiInterface. A groupby on a MultiInterface can only be applied if the dimension being grouped on is a scalar value in each item of the data. This allows polygon type data indexed by a variable to be grouped by one of those index values. For example you might have a Dataset containing geometries for all countries indexed by continent. Using groupby you can group the polygons by continent:

```python
> ds
:Dataset   [lon,lat,continent] (country)

> ds.groupby('continent')
:HoloMap   [continent]
   :Dataset   [lon,lat]   (country)
```

- [x] Add unit tests

### DIFF
--- a/holoviews/core/data/multipath.py
+++ b/holoviews/core/data/multipath.py
@@ -1,8 +1,8 @@
 import numpy as np
 
+from .. import util
 from ..element import Element
 from ..ndmapping import NdMapping, item_check, sorted_context
-from ..util import max_range, get_param_values, unique_iterator, isscalar
 from .dictionary import DictInterface
 from .interface import Interface, DataError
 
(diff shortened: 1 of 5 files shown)


holoviews/core/data/multipath.py
```diff
@@ -293,7 +333,7 @@ def split(cls, dataset, start, end, datatype, **kwargs):
     def add_dimension(cls, dataset, dimension, dim_pos, values, vdim):
         if not len(dataset.data):
             return dataset.data
-        elif values is None or np.isscalar(values):
+        elif values is None or util.isscalar(values):
```
jlstevens: I thought these had all been replaced already...

jlstevens:
Looks good. Happy to merge once the tests pass.

holoviews/core/data/multipath.py
philippjfr: So did I.

PR #3101, opened 2018-10-24
Add size_index to graphs

ahuang11:
Try to address https://github.com/ioam/holoviews/issues/3097, but I'm not really sure what I'm doing; just copying a bunch of code from the other scripts.

Still useless too because I can't seem to figure out where it only takes the first three(?) columns and converts it to `['index', 'x', 'y']`(?) instead of `['index', 'x', 'y', 'test']` so size_index/color_index only works on either `'index'`, `'x'` or `'y'`... (I think if the three column restriction is removed, it'll help complement the fix to the image hover_col limit too(?) https://github.com/pyviz/geoviews/issues/242 besides changing `bounds=(1, None))`)

image

### DIFF
--- a/holoviews/element/graphs.py
+++ b/holoviews/element/graphs.py
@@ -82,7 +82,7 @@ class Nodes(Points):
     """
 
     kdims = param.List(default=[Dimension('x'), Dimension('y'),
-                                Dimension('index')], bounds=(3, 3))
+                                Dimension('index')], bounds=(3, None))
 
     group = param.String(default='Nodes', constant=True)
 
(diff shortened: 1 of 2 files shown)


philippjfr:
I'd hold off on this, https://github.com/ioam/holoviews/pull/2152 should already allow for this.

ahuang11:
Okay. Still wondering where in the code are the columns changed to ['x', 'y', 'index']

philippjfr:
That does sound like an issue, will have to look into it.

philippjfr:
I'll close since I'll be addressing any issues that crop up in the ops PR.

PR #3110, opened 2018-10-27
Remove the if statement

ahuang11:
Not sure if should just remove legend_cols and implement legend_orientation since I don't think bokeh supports legend cols yet?

image


philippjfr:
Personally I'm -1 on this, it's a major change in behavior. I'd vote for putting in the effort and adding proper support for legend_cols in bokeh.

philippjfr:
I'll close, maybe just bump https://github.com/bokeh/bokeh/issues/3880.

PR #3115, opened 2018-10-29
Relax bokeh pin to allow <1.1.0 and update to bokeh 1.0 on travis

philippjfr:


jlstevens:
This is a sensible change as we hope everyone will be using bokeh 1.0+ going forward. Merging.

PR #3127, opened 2018-11-03
Ensure methods accept both single dimension/spec and list of them

philippjfr:
In general we have ensured that methods that accept either dimensions or specs will accept both single items and lists of items but this has never been done very consistently.

- [x] Fixes https://github.com/ioam/holoviews/issues/3080


jlstevens:
Makes sense, merging.

PR #3130, opened 2018-11-04
added muted_alpha option for Vectorfield and for Spikes

ea42gh:
found myself wanting muted_alpha for both of these elements


philippjfr:
Thanks!

PR #3132, opened 2018-11-04
Ensure RGBPlot is not a ColorbarPlot

philippjfr:
Ensures that RGBPlot is separate from RasterPlot to ensure it does not inherit plot and style options related to colormapping.

- [x] Fixes https://github.com/ioam/holoviews/issues/2873


philippjfr:

PR #3139, opened 2018-11-05
 Handle updating of bokeh legends

philippjfr:
Previous legends were completely static in the bokeh backend, i.e. in a HoloMap it would display all legend items for all frames at once and in a DynamicMap they would simply not update. Additionally this PR fixes a bug where the addition of an annotation on a plot would cause a legend to appear.

Also includes various fixes for dynamically creating and reassigning plot instances to ensure that a) the cyclic index is consistent and that b) dynamically created plot instances are updated appropriately despite not being assigned the correct stream mappings.

- [x] Closes https://github.com/ioam/holoviews/issues/2611
- [x] Closes https://github.com/ioam/holoviews/issues/2927
- [x] Add unit tests

### DIFF
--- a/holoviews/plotting/bokeh/annotation.py
+++ b/holoviews/plotting/bokeh/annotation.py
@@ -17,13 +17,13 @@ except:
 from ...core.util import dimension_sanitizer, basestring
 from ...element import HLine
 from ..plot import GenericElementPlot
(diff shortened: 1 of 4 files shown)


philippjfr:
Going ahead and merging this, it's all changes in the plotting code which are hard to review anyway and I also want to ensure that this is tested for as long as possible before a release.

PR #3141, opened 2018-11-05
Skip Tap callbacks if outside axis bounds

philippjfr:
Unlike the PointerXY stream where we clip the pointer location if it goes outside the bounds to ensure that fast mouse movements are captured the tap callback should simply be skipped if the tap occurs outside the plot bounds.

- [x] Closes https://github.com/ioam/holoviews/issues/3135


philippjfr:
Another uncontroversial fix which will not affect tests so I'm going ahead and merging.

PR #3152, opened 2018-11-07
Ensure that contours operation is robust to curvilinear coord edges

philippjfr:
Due to recent changes the contours operation did not robustly deal with curvilinear quadmesh coordinate edges.

- [x] Add unit test


philippjfr:
Merging, straightforward bug fix along with a unit test.

PR #3153, opened 2018-11-07
Update dimensioned stream layout title

philippjfr:
When plotting a layout or grid containing dimensioned streams the layout or grid title does not currently update because it is not notified about changes in the stream values. This ensures that the plots subscribe to changes in any dimensioned streams.


philippjfr:
Relatively straightforward enhancement with associated unit tests so I'll go ahead and merge.

PR #3154, opened 2018-11-07
Add explicit xlabel/ylabel(/zlabel) plot options

philippjfr:
Overriding the axis labels is currently a fairly complex procedure because you have to redimension a bunch of elements. Particularly when you are working with a DynamicMap where you can't necessarily inspect the dimensions of returned elements it is, in some cases, not feasible at all. It also would add huge complexity to perform operations like this in ``hvplot``. Various users have therefore requested explicit ``xlabel``/``ylabel``(``/zlabel``) plot options, which are easy to add and much easier to control.

- [x] Closes https://github.com/ioam/holoviews/issues/2833

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1321,7 +1321,7 @@ class OverlayPlot(GenericOverlayPlot, LegendPlot):
                           'border', 'invert_xaxis', 'invert_yaxis', 'sizing_mode',
                           'title_format', 'legend_position', 'legend_offset',
                           'legend_cols', 'gridstyle', 'legend_muted', 'padding',
-                          'xlim', 'ylim', 'zlim']
+                          'xlabel', 'ylabel', 'xlim', 'ylim', 'zlim']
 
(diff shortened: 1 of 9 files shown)


jlstevens:
This approach is useful for an explicit override but it should *not* be something you keep repeating (that just means you've not declared your dimension right).

I'm fine with the code in this PR (and happy to merge) but I would like to know how we should communicate this. Essentially, 'you can use xlabel as a plot option but if you keep re-using the same xlabel you should declare your dimension properly instead'.

jlstevens:
Ok, I think I'll merge so we can add the docs in a subsequent PR. Here is the minimum I would expect for docs:

* In the User guide when people are introduced to how dimensions determine axis labels, we can add a pointer forwards to options to say 'if you need to override a label on a plot where changing the dimension temporarily is inappropriate, you can use the options system'.
* In the new section in the options user guide to a reverse pointer 'if you find yourself setting the same label frequently, declare the correct label at the dimension level'.

Otherwise, I am happy with the code here so I will merge.

PR #3156, opened 2018-11-07
Subclass pyviz_comms extension and add delete hook

philippjfr:
This PR subclasses the extension from pyviz_comms ensuring that independently of which extension you load all plots are cleaned up correctly when a cell is deleted, re-executed or cleared.

Should not be merged until pyviz_comms 0.7 containing the extension has been released.

- Depends on https://github.com/pyviz/pyviz_comms/pull/21


jlstevens:
The push build has passed. Merging.

PR #3158, opened 2018-11-07
Add mode to send bokeh updates as individual comm messages

philippjfr:
Currently when working in the notebook we batch bokeh change events and send them all in a big chunk. This is generally okay but if any of the events causes an error on the frontend none of the subsequent events are processed. This PR ensures that each event is sent and therefore processed separately. This allows certain changes to plots to be made without irrecoverably breaking a plot.

- [x] Fixes https://github.com/ioam/holoviews/issues/3125
- [x] Unblocks https://github.com/ioam/holoviews/pull/3072


philippjfr:
Merging, if we find any performance regression we can revert but since bokeh server uses the same approach this should be fine.

PR #3162, opened 2018-11-08
Fixed PermissionError when saving gif output on Windows

philippjfr:
NamedTemporaryFile causes issues when auto-deleting on windows, this fix is what is used by bokeh to ensure exporting works on windows and some basic testing by myself and by @smillies seems to confirm it works. That said in the long term we should set up CI tests on Windows.

- [x] Fixes https://github.com/ioam/holoviews/issues/3151

### DIFF
--- a/holoviews/plotting/mpl/renderer.py
+++ b/holoviews/plotting/mpl/renderer.py
@@ -1,3 +1,4 @@
+import os
 import sys
 import base64
 from io import BytesIO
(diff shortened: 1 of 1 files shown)


jlstevens:
Seems like an issue with `NamedTemporaryFile` to me which everyone has to work around. I'll merge when the tests are green (even if Travis isn't testing Windows).

PR #3163, opened 2018-11-08
Removing redundant call to `gaussian_kde`

ek-nath:
There are only two ways the procoess can go forth. Either `kde` is set correctly to the return value of `stats.gaussian_kde` or there's an exception and the function `return`s. The second call to `stats.gaussian_kde` seems redundant


philippjfr:
Looks good, merging.

PR #3168, opened 2018-11-10
Ensure changing CDS columns do not trigger errors

philippjfr:
In https://github.com/ioam/holoviews/pull/3158 I made a change to the way bokeh events are handled, which sent each bokeh model change as a separate event and thereby avoided an error triggered by one event breaking a plot irreparably because subsequent errors would be ignored. Unfortunately that change introduced a noticeable delay between a CDS data and color mapper ranges being updated, resulting in undesirable flicker on updates. It was also a bit of a hack around the actual issue because the underlying issue would still trigger errors. This PR reverts that change and fixes the underlying problem.

It does this through a fairly involved change, so I'll step through it slowly:

1. It detects whether a glyph dataspec references a CDS.data column which does not yet exist
2. If this is the case it triggers a new ModelChangedEvent which replaces the data with empty columns which do contain the new column(s) and the old column(s)
3. Once the glyph has been updated to reference the new column(s), it sends another event containing the actual data

### DIFF
--- a/holoviews/plotting/bokeh/bokehwidgets.js
+++ b/holoviews/plotting/bokeh/bokehwidgets.js
@@ -15,10 +15,7 @@ BokehScrubberWidget.prototype = Object.create(ScrubberWidget.prototype);
 var BokehMethods = {
   update_cache : function(){
     for (var index in this.frames) {
-      var data = this.frames[index];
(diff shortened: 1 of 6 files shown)


philippjfr:
Going to merge so I can continue working on other PRs that depend on this.

Issue #3095, opened 2018-10-22
Proposal replacement for opts magics


jlstevens:
Over the years we've had many proposals for how to specify options with the options system, many we have discussed and never implemented and many that we have. Now I am adding one more proposal that I *think* gets us 99% of where we want to be. :-)

We starting with the line and cell magics with multiple internal formats to represent the nesting, then we added ``__call__``,  then we renamed `__call__` to `opts`, then added `.options` (flat namespace).

 In addition we have the `hv.opts` utility which operates as a pure python replacement to the magics. It is safe to say that this is all quite confusing and we want one *good* way we can recommend and document (and eventually deprecate the other ways)

This all comes down to a fundamental problem with specifying nested structures to style nested, compositional objects. Even though we now encourage the use the `.options` method, we don't have a solution that satisfies all these goals:

1. Pure Python syntax. Obviously magics fail at this test.
2. Separation of presentation from content. Magics are good at this, with the options always found at the top of the cell. Currently, `.options` is often not as you need to refactor your code to apply `.options` at the right level.
3. Tab completion. Magics are good at this, `.options` is not.
4. Ideally something that lets us easily map only magics to the new approach (related to goal 2).

The closest thing to a recommendation right now is to declare a set of dictionaries at the top of the cell then applying `.options` appropriately. This satisfies goals 1 but fails at goal 3. Goal 4 fails as goal 2 is only partially satisfied as you need to factor out your code appropriately. On top of that, I consider it all quite ugly imho!

The current proposal is to extend `hv.opts` so it can be more a direct substitution for the magics. The aim is to improve our current recommendation (dictionaries + `.option` methods) with tab completion and compositionality so that you can just apply `.options` to a deeply nested object without having to refactor your code. In short, the proposal is based on these two ideas:

1. You can use `hv.opts` to get an object that enhances the dictionary approach (i.e for use in `.options`)
2. This object is queryable and compositional (could be the return value from the opts parser) and can be used by the `.options` method.

In its simplest form, instead of:

```python
curveopts = dict(color='green')
imageopts = dict(cmap='fire')
Image(...).options(**imageopts) * Curve(...).options(**curveopts) 
```

you could use:

```
curveopts = hv.opts(color='green')
imageopts = hv.opts(cmap='hsv')
Image(...).options(**imageopts) * Curve(...).options(**curveopts) 
```

(Note that this is only technically possible as the use of keywords is a distinct signature from what ``hv.opts`` currently supports.)

Why do this?

[...]

PR #3173, opened 2018-11-12
Replacement for cell magic with tab-completion

jlstevens:
Work in progress to address #3095. I'm mostly happy with this approach, illustrated here:

image

### Notes:

* This is *not* an exact replacement for `%%opts` unless you use `clone=False`. The magics set options on the displayed object, not a clone of it.
* This is still the line magic equivalent to `%opts Curve (color='k')`:

```
    opts("Curve (color='k')") # Or equivalently
    opts({'Curve': {'style': {'color':'k'}}})
```
* This is still the more *direct* cell magic equivalent (without tab-completion) to
   `%%opts Curve (color='k')`:
```
    curve = hv.Curve([1,2,3])
    opts("Curve (color='k')", curve) # Or equivalently
    opts({'Curve': {'style': {'color':'k'}}}, curve)
```
* `opts` can now tab-complete *all possible options* if you *only* use keywords, returning an `Option` object.
* `Option` objects have been enhanced with improved validation and reprs (and can now take something like `Curve.Group.Label` as a key.
* Switching the backend updates the completer objects (and the docstrings needed for tab-completion)
* The `opts` entries do not exist until a backend is loaded.
* There is a minor API change to `.options` as you can no longer set `backend` as a positional argument.

### TODO:

* [x] Check `set_current_backend` is used correctly throughout.
* [x] Fix setting the `opts` class docstring so it isn't clobbered.
* [x] Move the merge option utility out of the parser.
* [x] Unit tests.
* [x] Post an ad hoc script to demonstrate automated conversion from cell magics. **[Would be nice, doesn't have to be in this PR though]**
* [x] Update docs and new docs **[To be done in the next PR]**

### Decisions to make **[We can address these in the docs update PR]**:

* [x] Qualified imports always or can we make `opts` available in the namespace of examples? (I vote yes) 
* [x] Default name for the list of `Option` objects (I think `options` isn't bad but a bit long)
* [x] Improve the API of `Options` now it is more user facing than before? i.e make some methods private? Should we document any of the methods/properties at all? Are they useful (e.g the `cyclic` predicate?)


jbednar:
Looks great to me; thanks for seeing this through! Finally, we can have (nearly) everything we want!

jlstevens:
@philippjfr I've ticked off 'Check set_current_backend is used correctly throughout.' having checked that `set_current_backend` is used where I think it is necessary (see 34f2318, b857670).

I decided *not* to replace all the places where `current_backend` is set directly in the tests as that would result in a lot of relatively pointless changes: `set_current_backend` is only needed to run the hooks which the tests don't rely on (and the only hook is for building the completers on `hv.opts`).

philippjfr:
Sounds good.

philippjfr:
Sounds good, although I've not used hv.output myself. Unless you object it would also be good to fix the error message when applying options when no backend has been imported, e.g. when running this:

```
import holoviews as hv
hv.Curve([1, 2, 3]).options(xaxis=None)
```

you currently get this useless error:

```
~/development/holoviews/holoviews/core/options.py in options(cls, backend, val)
   1057         backend = cls.current_backend if backend is None else backend
   1058         if val is None:
-> 1059             return cls._options[backend]
   1060         else:
   1061             cls._options[backend] = val

KeyError: 'matplotlib'
```

Would be nice to put an informative error message in ``Store.options`` if no backend has been imported.

philippjfr:
Another thought I just had is that we are reasonably close to being able to validate the option *values* early. My ``op`` PR defines an interface to validate any style option for the bokeh backend using bokeh's property system like this:

```
from holoviews.plotting.bokeh.styles import validate
validate('line_color', 'red')
```

And once we implement https://github.com/ioam/param/issues/300 is merged it'll be easy to do the same for plot options. I think that could be really useful, even if it isn't enabled by default (at least at first).

philippjfr:
>Presumably we can then also not just validate the option values, but also tab complete them and show them in the help?

That only works for certain options, so at least as a first cut this seems like too much complexity for not that much gain.

jlstevens:
@philippjfr I added a unit test in 79950f6. I'm not entirely happy with it but maybe it will do?

jlstevens:
@philippjfr I would like to see this PR merged so we can start testing it (and you could try `op`/`dim` with it). I've punted a few tasks to the docs PR that will follow but for now I think it is ready for review/merge (assuming the tests go green).

philippjfr:
Was hoping to see a few more unit tests for the new accepted ``.options`` formats, but I'll merge if you want to add those later.

philippjfr:
That's my only review comment really.

jlstevens:
That is a good idea, I'll add a few more unit tests along those lines.

PR #3186, opened 2018-11-19
Fixed conversion of Pandas Timestamp to integer for Bokeh

### LINKED ISSUE #2857: For HLine/VLine time is in UTC
For HLine/VLine time is in UTC. Same date works fine with Points and Spikes.
image

### LINKED ISSUE #3167: VLine does not support Timestamps
In #3075, support for timestamps was removed.

In v1.10.7, the timestamp was incorrectly rendered due to timezones (as was documented), but the fix in #3075 actually just removed support for Timestamps entirely on VLine.

```
hv.VLine(pd.to_datetime('06:00'))
```

"works" in v1.10.7:
image

I get the following error message in v1.10.9:

image

edjubuh:
- Reverts #3075 which effectively drops support for datetime types entirely
- Use [`pandas.Timestamp.timestamp`](https://pandas.pydata.org/pandas-docs/version/0.23.4/generated/pandas.Timestamp.timestamp.html#pandas.Timestamp.timestamp) to convert to a POSIX timestamp, which will handle timezone conversions to fix #2857 and also support millisecond precision in Timestamps

Closes #3167

### DIFF
--- a/holoviews/plotting/bokeh/annotation.py
+++ b/holoviews/plotting/bokeh/annotation.py
@@ -14,13 +14,13 @@ except:
     arrow_end = {'->': NormalHead, '-[': OpenHead, '-|>': NormalHead,
                  '-': None}
 
-from ...core.util import dimension_sanitizer, basestring
+from ...core.util import datetime_types, dimension_sanitizer, basestring
 from ...element import HLine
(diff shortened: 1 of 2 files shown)


philippjfr:
Thanks for the fix, LGTM.

PR #3192, opened 2018-11-23
Implemented continuous style mapping for Paths

philippjfr:
Followup on the recent PR to support mapping styles. In that PR I had not yet implemented style mappings for continuously varying paths.

Here is an example of a continuously colormapped, alpha-mapped, and linewidth-mapped path:

screen shot 2018-11-23 at 3 01 30 am

- [x] Add unit tests
- [x] Update existing continuous colormapping example


jlstevens:
Great! I remember talking about implementing varying visual attributes along a path a *long* time ago so it is good to see it is finally possible.

Merging.

PR #3193, opened 2018-11-24
Unlimit the vdim bounds and add ability for additional vdim hover

ahuang11:
image

Let me know if there needs to be some changes (especially the l, b, dw, dh part; I only took a quick glance at it) or feel free to make changes directly!


ahuang11:
Any pointers on what needs to changed in hvplot?
image

philippjfr:
Thanks @ahuang11 I'll try to review and merge this tomorrow.

PR #3197, opened 2018-11-26
Implement datetime selections as strings and datetimes

philippjfr:
Closes https://github.com/ioam/holoviews/issues/886

### DIFF
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -419,6 +419,8 @@ class GridInterface(DictInterface):
 
     @classmethod
     def key_select_mask(cls, dataset, values, ind):
+        if util.pd and values.dtype.kind == 'M':
+            ind = util.parse_datetime_selection(ind)
         if isinstance(ind, tuple):
             ind = slice(*ind)
         if isinstance(ind, get_array_types()):
--- a/holoviews/core/data/interface.py
+++ b/holoviews/core/data/interface.py
@@ -283,30 +283,35 @@ class Interface(param.Parameterized):
         have been selected.
         """
         mask = np.ones(len(dataset), dtype=np.bool)
-        for dim, k in selection.items():
-            if isinstance(k, tuple):
-                k = slice(*k)
+        for dim, sel in selection.items():
+            if isinstance(sel, tuple):
+                sel = slice(*sel)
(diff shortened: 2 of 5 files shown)


poplarShift:
Great to see this coming along! Just tested your PR, ran into the following:

```
df = pd.DataFrame(dict(time=pd.date_range('2018-1-1', freq='1D', periods=10),
             x=np.arange(10), y=np.linspace(10,1,10)))

x = hv.Curve(df, vdims='x')
y = hv.Curve(df, vdims='y')

l = x*y

x['2018-1-1':'2018-1-3'] # works
y['2018-1-1':'2018-1-3'] # works
# x['2018-1-1':'2018-1-3'] + y['2018-1-1':'2018-1-3'] # fails
# l['2018-1-1':'2018-1-3'] # fails
```

poplarShift:
@philippjfr I've been chasing a few bugs in your PR. Would you mind if I pushed directly to this branch?

philippjfr:
Please do.

poplarShift:
Sorry for all the force pushing, I rolled back on a commit to rebase and use some recent datetime util of yours (only to hesitate once more). Could you take a look at 6dedb77? That piece clashes when values have been parsed in the meantime, such as datetimes, and that was the one that prevented layouts from being rendered. I don't think it's necessary either as the various get_extents methods and utils do the same thing, but please correct me.

philippjfr:
>That piece clashes when values have been parsed in the meantime, such as datetimes, and that was the one that prevented layouts from being rendered. I don't think it's necessary either as the various get_extents methods and utils do the same thing, but please correct me.

That stuff is the focus of multiple (very old) issues (see https://github.com/pyviz/holoviews/issues/181 and https://github.com/pyviz/holoviews/issues/566). The long and short of it is that I really would like to get rid of it, but the main problem is that to get rid of it consistently everywhere we would have to implement proper routines for slicing paths and polygons. I haven't thought about that for a long time and would say that maybe it's okay to simply drop it from Chart because slicing is implemented in a more appropriate way there.

poplarShift:
Ok, then I'd say this is ready for review. I've tested slicing a few more elements and overlays with various annotations and didn't come across anything obvious apart from the DictInterface issues mentioned in #3697.

PR #3198, opened 2018-11-26
Minor fixes for gridstyle and formatter options

philippjfr:
The formatter options weren't being propagated and the gridstyle visible option was not being processed correctly.


jlstevens:
Thanks for these fixes! I believe they address issues raised in #3188. Happy to merge once tests pass and once you've added any unit tests you think would help stop any regressions...

jlstevens:
All tests are green. Thanks!

PR #3201, opened 2018-11-26
Improve dim transform repr and add further operators

philippjfr:
This PR improves the repr of ``dim`` transform to ensure it is in most cases fully usable. It also adds support for further operators and numpy functions.

Here's a mixture of unary operators (``-``), binary operators (``*`` and ``+``), numpy ufuncs (``np.log``), and ndarray methods (``min``): 

```python
> np.log(-dim('A')*3+3).min()

np.log((-dim('A')*3)+3)).min()
```

And here's an example using one of some of the custom functions (``norm`` and ``categorize``):

```python
> (dim('y').norm()>0.5).categorize({False: 'red', True: 'blue'})

(dim('y').norm()>0.5).categorize(categories={False: 'red', True: 'blue'}, empty=None)
```

### DIFF
--- a/holoviews/util/transform.py
+++ b/holoviews/util/transform.py
@@ -11,22 +11,15 @@ from ..element import Graph
 
 
 def norm(values, min=None, max=None):
-    """
-    min-max normalizes to scale data into 0-1 range.
+    """min-max normalization to scale data into 0-1 range.
 
-    Arguments
-    ---------
-    values: np.ndarray
-       Array of values to be binned
-    min: float (optional)
-       Lower bound of normalization range
-    max: float (optional)
-       Upper bound of normalization range
+    Args:
+        values: Array of values to be normalized
+        min (float, optional): Lower bound of normalization range
+        max (float, optional): Upper bound of normalization range
 
-    Returns
-    -------
-    normalized: np.ndarray
-       Array of normalized values
+    Returns:
+        Array of normalized values
     """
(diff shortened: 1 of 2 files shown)


philippjfr:
@jlstevens Ready to review and merge.

holoviews/util/transform.py
```diff
@@ -304,15 +357,52 @@ def apply(self, dataset, flat=False, expanded=None, ranges={}, all_values=False)
...
+                    fn_name = self._custom_funcs[fn]
+                    format_string = prev+'.{fn}('
+                elif ufunc:
+                    fn_name = str(fn)[8:-2]
+                    if not (prev.startswith('dim') or prev.endswith(')')):
+                        format_string = '{fn}' + prev
+                    else:
+                        format_string = '{fn}(' + prev
+                    if fn_name in dir(np):
+                        format_string = 'np.'+format_string
```
jlstevens: Maybe this ('np.') should be set via a class attribute...

jlstevens:
Looks good! This will be very helpful for my helper script that uses the parser to suggest was of removing the magics (the old reprs where invalid/confusing).

Happy to merge though is there an option to avoid qualifying the namespace? I'm happy with the default but let's say someone used `import numpy` a lot instead of `import numpy as np` or used `from numpy import *`. I made a suggestion above as how this could be done in the code...

philippjfr:
Ready to merge.

PR #3206, opened 2018-11-27
Fixed handling of Link targets

philippjfr:
Since Links now use a weakrefdict to store handles on the source and target objects it is possible for the reference to be deleted before the plot is instantiated. In cases like that the plots should not attempt to instantiate the link callback.


jlstevens:
Makes sense. Happy to merge when the tests pass (I am assuming there are some unit tests for links).

philippjfr:
>(I am assuming there are some unit tests for links)

There are, I had considered adding new ones but didn't get the garbage collection to work consistently to actually test for the issue this is meant to prevent.

jlstevens:
Tests are now green. Merging.

PR #3208, opened 2018-11-27
Implement mul for overlay * holomap

philippjfr:
I'm a bit perplexed that this wasn't implemented already but it's a pretty straightforward addition.

- [x] Add unit test


jlstevens:
I'm also amazed this hasn't cropped up earlier!

PR #3211, opened 2018-11-27
Implement mul for overlay * holomap

jlstevens:
Attempting to fix the issues in #3208 and #3210 - there was a submodule update in those PRs that may have been intentional but may also have been accidental (explaining the issue).


holoviews/core/overlay.py
```diff
@@ -194,7 +194,7 @@ def dynamic_mul(*args, **kwargs):
             return other.clone(shared_data=False, callback=callback,
                                streams=[])
         elif not isinstance(other, ViewableElement):
-            return NotImplemented
+            raise NotImplementedError
```
jlstevens: I remember there is special semantics of `NotImplemented` so now I am wondering if that was the right thing.

jlstevens: I.e:

```suggestion
            return NotImplemented
```

philippjfr: ``return NotImplemented`` in ``__mul__`` force it to use ``__rmul__`` which is important to get the HoloMaps working.

jlstevens: Right, I figured that out and fixed it accordingly. Was I right to remove the submodule reference update? The tests are now green.

philippjfr: Yes, should be ready to merge.

PR #3215, opened 2018-11-28
Do not expand scalar style options

philippjfr:
As part of the style mapping PR I added some code that expands the ``color`` and ``alpha`` options to ``fill_color``/``fill_alpha`` and ``line_color``/``line_alpha``. This should however only apply for actual style mapping while currently it also applies to scalar options, which in some cases overrides the desired styling.

Before the fix:

bokeh_plot

With the fix:

bokeh_plot


jlstevens:
Thanks for the fix. Merging.

PR #3217, opened 2018-11-28
Stream and Link are inherited on clone

philippjfr:
A highly confusing issue has meant that whenever you attach a Stream or Link to an element and subsequently apply some method to it, e.g. ``.relabel``, ``.redim``, ``.options`` etc., it becomes unlinked. This behavior is extremely confusing and routinely catches even me out even though I'm fully aware of the issue.

In the  case of linking a DynamicMap we had already solved this issue via the ``link_inputs`` parameter on the ``Callable``. This solution is complementary to what is being proposed here, but this PR standardizes linking behavior by adding a ``link`` argument to all ``clone`` methods. This controls both the ``link_inputs`` behavior of a DynamicMap and controls whether an element remains linked. I haven't touched the way links on Stream and Link classes are actually stored as the current arrangement allows garbage collection to occur properly, therefore I've only changed how a linked Stream or a Link is looked up. The required changes are therefore fairly minimal, but I believe this will provide a huge improvement to usability.

There is some risk that a user might accidentally end up with multiple links in their plots, but I'm quite certain this would only occur if they use cloned elements in unorthodox ways and they always have the option to unlink. The benefit therefore hugely outweighs the risk because currently you have to structure your code in very awkward ways or know about things like ``.options(..., clone=False)`` if you want to ensure your element stays linked.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -648,13 +648,17 @@ class LabelledData(param.Parameterized):
             raise ValueError("Supplied label %r contains invalid characters." %
                              self.label)
 
-    def clone(self, data=None, shared_data=True, new_type=None, *args, **overrides):
+    def clone(self, data=None, shared_data=True, new_type=None, link=True,
+              *args, **overrides):
         """Clones the object, overriding data and parameters.
 
         Args:
             data: New data replacing the existing data
             shared_data (bool, optional): Whether to use existing data
             new_type (optional): Type to cast object to
+            link (bool, optional): Whether clone should be linked
+                Determines whether Streams and Links attached to
+                original object will be inherited.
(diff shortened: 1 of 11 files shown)


jlstevens:
I am also greatly in favor of this change and would be willing to argue that the old behavior was buggy: when you make a clone, you expect a new object but you also expect the new object to behave the same was as the object it was cloned from.

Losing the links and streams was an issue I knew about but it didn't really start biting me till I tried reorganizing how `.options` was used in our examples.

philippjfr:
I would favor merging this as soon as the next dev release is out.

jlstevens:
Now the alpha 11 is out, I agree that this should be merged so we can test it properly.

PR #3224, opened 2018-11-29
Consistently renamed to PyViz global object

philippjfr:
Completes the renaming from the HoloViews global object to PyViz.

Linked to https://github.com/pyviz/pyviz_comms/pull/24

### DIFF
--- a/holoviews/plotting/widgets/widgets.js
+++ b/holoviews/plotting/widgets/widgets.js
@@ -72,7 +72,7 @@ HoloViewsWidget.prototype.update = function(current){
 
 HoloViewsWidget.prototype.init_comms = function() {
   var that = this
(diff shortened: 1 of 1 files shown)


fmaussion:
not directly related: is the hv repository also going to move to the pyviz organization eventually?

philippjfr:
>not directly related: is the hv repository also going to move to the pyviz organization eventually?

Yes, that's long overdue. The plan was to split out the plotting components into their own repos (holoviews-bokeh, holoviews-matplotlib, holoviews-plotly) and then migrate everything across to pyviz.

jbednar:
I don't see why not move them first and be done with the move before having to worry about the split.

philippjfr:
>I don't see why not move them first and be done with the move before having to worry about the split.

I agree.

philippjfr:
The main annoyance will be updating various parts of our build and test infrastructure.

PR #3225, opened 2018-11-29
Fixed initialization of scrubber widget

philippjfr:
Fixes a regression introduced recently when I fixed initialization of the SelectionWidget.


philippjfr:
Simple fix and tests are passing so I'll merge.

PR #3229, opened 2018-11-30
Clip Pointer stream bounds

philippjfr:
The pointer and tap streams sometimes fire even when outside the plot area. We avoided this by clipping the values in a JS callback but the JS callback is never executed on bokeh server so this PR adds conditional code to ensure that the positions are clipped on bokeh server as well.

- [x] Fixes https://github.com/ioam/holoviews/issues/3145


jlstevens:
Good to fix things on bokeh server as well. Merging.

PR #3232, opened 2018-11-30
Skip empty stream callbacks

philippjfr:
This fix seems correct to me, i.e. I see no reason why an empty message received by a callback should trigger a stream event even if that stream is declared transient, but maybe I'm just missing something. The skipping and the exception for transient streams was made here https://github.com/ioam/holoviews/pull/1350 but I can't see any reference to the reasoning behind making an exception for transient streams.


philippjfr:
After discussion with @jlstevens we came to the conclusion this was the correct behavior.

PR #3234, opened 2018-12-01
Allow overlaying Layout with element or HoloMap

philippjfr:
This PR allows overlaying a Layout with an element, overlay or HoloMap.

- [x] Implements https://github.com/ioam/holoviews/issues/2596

### DIFF
--- a/holoviews/core/layout.py
+++ b/holoviews/core/layout.py
@@ -508,6 +508,14 @@ class Layout(ViewableTree):
         "Composes the Layout with another object returning a merged Layout."
         return Layout([self, other])
 
+    def __mul__(self, other, reverse=False):
+        from .spaces import HoloMap
+        if not isinstance(other, (ViewableElement, HoloMap)):
+            return NotImplemented
+        return Layout([other*v if reverse else v*other for v in self])
+
+    def __rmul__(self, other):
+        return self.__mul__(other, reverse=True)
 
 
 __all__ = list(set([_k for _k, _v in locals().items()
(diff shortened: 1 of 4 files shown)


jlstevens:
We considered this semantics a long time ago but decided explicitly against it. I know there are arguments for this behavior but I would like to know if you have any new arguments why this behavior should be supported. For one thing, this would be collated into a different structure than what you declared... `(layout  * element)` becomes a layout of overlays and not an overlay of layouts.

philippjfr:
>We considered this semantics a long time ago but decided explicitly against it.

I don't remember that and if you look at the linked issue we all seemed to agree that this change was sensible.

>For one thing, this would be collated into a different structure than what you declared... (layout * element) becomes a layout of overlays and not an overlay of layouts.

This is true of NdLayout and GridSpace too, but we have supported those for a long time.

jlstevens:
This is the partial support of `*` on layouts we discussed then. What is the error if you use `*` on two layouts? That needs a good error message because *that* is definitely something we do not want to support.

philippjfr:
>That needs a good error message because that is definitely something we do not want to support.

It's just the default error message that Python generates for objects that don't implement ``__mul__``, i.e.:

```
TypeError                                 Traceback (most recent call last)
<ipython-input-22-3deae7bba76b> in <module>()
----> 1 hv.Layout([hv.Curve([1, 2, 3])]) * hv.Layout([hv.Curve([1, 2, 3])])

TypeError: unsupported operand type(s) for *: 'Layout' and 'Layout'
```

jlstevens:
I'm not entirely comfortable with this feature (because the structure of what you get isn't what you declare) but there are certainly a few cases where it would be useful. 

I don't think distributing elements over a layout is an extremely common case (we haven't desperately needed this so far) so on balance I'm happy to make this a feature, even if it is one I personally don't expect to need or use very often!

PR #3235, opened 2018-12-01
Avoid erroring when aggregating on non-numeric columns

philippjfr:
- [x] Fixes https://github.com/ioam/holoviews/issues/2350
- [x] Adds unit test

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -648,16 +648,17 @@ class Dataset(Element):
                 vdims = self.vdims
             return self.clone([], kdims=kdims, vdims=vdims)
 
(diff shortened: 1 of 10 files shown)


jlstevens:
I am fine with this PR as long as you have thought of the sort of aggregation you could imagine on non-numeric data, specifically the sorts of stats you could compute relating to counts.

philippjfr:
That is a good point and I thought I'd have to add special handling for this but I just realized the current approach should work fine already. I'll just add an explicit unit test for this and then this can be merged.

philippjfr:
Added unit tests to ensure non-numeric aggregations are allowed on string types.

jlstevens:
Looks good. Merging.

PR #3240, opened 2018-12-02
Improved linking of axis ranges

philippjfr:
This PR makes two improvements:

1) Instead of having custom logic to compute custom axis labels for some plots it consistently computes the axis labels from dimensions. This also fixes various plots that did not implement the xlabel/ylabel options correctly.

2) It improves/fixes the logic which enables shared axes in bokeh, previously it was matching only on the label, which no longer works if a user defines a custom xlabel/ylabel and automatically links dimensions with identical labels but different names (which has forced us to use terrible Histogram/Distribution axis labels).

The new approach uses the bokeh tags property to define a tuple spec on the axis range, which other plots can then match on. Currently the spec consists of the (name, label) of the dimensions the axis range is drawn from. @jlstevens @jbednar Let me know if these semantics for linking axes make sense to you or if for example it should also match on the unit.

- [x] Add some docstrings
- [x] Add unit tests
- [x] Make Histogram/Distribution y-axis labels sane again (and fix https://github.com/ioam/holoviews/issues/3093)

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -519,7 +519,7 @@ class histogram(Operation):
     dimension = param.String(default=None, doc="""
       Along which dimension of the Element to compute the histogram.""")
 
-    frequency_label = param.String(default='{dim} Frequency', doc="""
+    frequency_label = param.String(default=None, doc="""
       Format string defining the label of the frequency dimension of the Histogram.""")
 
     groupby = param.ClassSelector(default=None, class_=(basestring, Dimension), doc="""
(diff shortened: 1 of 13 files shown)


jlstevens:
Matching on `(name, label)` of the *dimension* object and not on the displayed axis label does make sense to me. This is the behavior I would have expected so the main worry now is backwards compatibility.

philippjfr:
>This is the behavior I would have expected so the main worry now is backwards compatibility.

Personally I feel like the old behavior was at best buggy, so I'd be in favor of calling this a bug fix and therefore change the behavior in certain corner cases.

philippjfr:
I've added the semantics we discussed yesterday which also matches on the Dimension.unit as long as both dimensions define one.

jbednar:
I agree with all that.  If both dimensions define units and those don't match, it's not safe to link unless we are declaring units in an automatically convertible form, which we should do eventually but may never get to.

jlstevens:
I'm happy to merge once the tests are green.

jlstevens:
Tests have now passed. Merging.

PR #3249, opened 2018-12-02
Support linking DynamicMaps

philippjfr:
Reuses the logic used for finding stream sources to find Link sources.

- [x] Fixes https://github.com/ioam/holoviews/issues/3248
- [x] Adds unit test


jlstevens:
Looks good. Merging.

PR #3250, opened 2018-12-03
Tutorial typo fix (due to renaming?)

flothesof:
I've been going through this tutorial and spotted this.
I think this fixes a small typo in the "responding to events" tutorial: `crosshairs` must have been renamed to `markers` earlier, because at that point in the tutorial, no object with this name has been introduced.


philippjfr:
Thanks for catching this @flothesof.

PR #3257, opened 2018-12-05
Miscellaneous fixes to opts validation

jlstevens:
Fixes `opts.defaults` and enables cross-backend validation in the `hv.opts` completers.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -1311,10 +1311,78 @@ class Dimensioned(LabelledData):
 
         return self.opts(options, **kwargs)
 
-    def opts(self, options=None, backend=None, clone=True, **kwargs):
-        from ..util import opts
-        return opts.apply_option_types(self, options=options,
-                                       backend=backend, clone=clone, **kwargs)
+
+    def opts(self, *args, **kwargs):
+        """Applies nested options definition.
+
+        Applies options on an object or nested group of objects in a
+        flat format returning a new object with the options
+        applied. If the options are to be set directly on the object a
+        simple format may be used, e.g.:
+
+            obj.opts(cmap='viridis', show_title=False)
+
+        If the object is nested the options must be qualified using
+        a type[.group][.label] specification, e.g.:
+
+            obj.opts('Image', cmap='viridis', show_title=False)
+
+        or using:
+
+            obj.opts({'Image': dict(cmap='viridis', show_title=False)})
+
+        Identical to the .options method but returns the object by
+        default and not a clone.
+
+        Args:
+            *args: Sets of options to apply to object
+                Supports a number of formats including lists of Options
+                objects, a type[.group][.label] followed by a set of
+                keyword options to apply and a dictionary indexed by
+                type[.group][.label] specs.
+            backend (optional): Backend to apply options to
+                Defaults to current selected backend
+            clone (bool, optional): Whether to clone object
+                Options can be applied inplace with clone=False
+            **kwargs: Keywords of options
+                Set of options to apply to the object
+
+        For backwards compatibility, this method also supports the
+        option group semantics now offered by the
+        hv.opts.apply_options_type utility. This usage will be
+        deprecated and for more information see the apply_options_type
+        docstring.
+
+        Returns:
+            Returns the object or a clone with the options applied
+        """
+        apply_option_types = False
+        if len(args) > 0 and isinstance(args[0], dict):
+            apply_option_types = True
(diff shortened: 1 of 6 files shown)


holoviews/util/__init__.py
```diff
@@ -268,14 +268,15 @@ def _build_completer(cls, element, allowed):
         def fn(cls, spec=None, **kws):
             spec = element if spec is None else '%s.%s' % (element, spec)
             invalid = set(kws.keys()) - set(allowed)
+            prefix = None
             if invalid:
                 try:
-                    cls._options_error(list(invalid)[0], element,
-                                       Store.current_backend, allowed)
+                    cls._options_error(list(invalid)[0], element, None, allowed)
```
philippjfr: Does this support passing an explicit backend kwarg? The way it works for ``.options`` is that if you pass in a specific backend it becomes completely strict otherwise it's permissive about options which apply to one backend but not another. So if we want to support that you'd pop ``backend`` from kws above and then replace ``None`` here with the popped backend (defaulting to None).

jlstevens: Initially I thought it would be trivial to support this but in the end I needed to do a little refactoring in b43f60e to get this working.

jlstevens:
@philippjfr I renamed `expand_options` to `_expand_options`: I want to keep the public portion of the `opts` namespace very clear for users. Any objections?

holoviews/core/dimension.py
```diff
@@ -1311,10 +1311,78 @@ def __call__(self, options=None, **kwargs):
...
+
+        # By default do not clone in .opts method
+        clone = kwargs.pop('clone', None)
+        if clone is None:
+            kwargs['clone'] = False
+        else:
+            kwargs['clone'] = clone
+
+        if apply_option_types and util.config.future_deprecations:
+            param.main.warning("Dictionary signature of opts method deprecated. "
```
philippjfr: This warning does not seem particularly informative, what dictionary format? I'd probably be specific about it and have very specific warnings for each of the three conditions above since I don't think there is a clear way of describing all the nested formats at once.

philippjfr: Okay, fair enough, maybe just clarify "dictionary signature" a bit more, something like:

```
"Calling the .opts method with options broken down by options group (i.e. separate plot, style and norm groups) is deprecated. Use the .options method converting to the simplified format instead or use hv.opts.apply_option_types for backward compatibility."
```

jlstevens:
@philippjfr Ready to review/merge once the tests are green.

holoviews/util/__init__.py
```diff
@@ -95,6 +95,94 @@ def __call__(self, *args, **params):
 
             self._cellmagic(args[0], args[1])
 
+
+    @classmethod
+    def option_groups(cls, obj, options=None, backend=None, clone=True, **kwargs):
```
philippjfr: Didn't want to call this ``apply_option_groups`` or similar?

holoviews/core/dimension.py
```diff
@@ -1311,87 +1311,82 @@ def __call__(self, options=None, **kwargs):
...
-            **kwargs: Keywords of options by type
-                Applies options directly to the object by type
-                (e.g. 'plot', 'style', 'norm') specified as
-                dictionaries.
+            **kwargs: Keywords of options
+                Set of options to apply to the object
+
+        For backwards compatibility, this method also supports the
+        option group semantics now offered by the
+        hv.opts.apply_options_type utility. This usage will be
```
philippjfr: Called ``apply_options_type`` here?

philippjfr:
I haven't looked but aren't there changes needed for subclasses that define an ``opts`` method?

holoviews/core/dimension.py
jlstevens: Thanks.

holoviews/util/__init__.py
jlstevens: Ah yes, `apply_groups` I guess!

jlstevens:
> I haven't looked but aren't there changes needed for subclasses that define an opts method?

I'll double check that, but I think they all delegate most of the actual work via `super`.

holoviews/core/dimension.py
jlstevens: Fixed in  fb81455

holoviews/util/__init__.py
jlstevens: Fixed in fb81455, thanks.

PR #3260, opened 2018-12-06
Updated to param 2.0 API

philippjfr:
Methods updated:

- [x] debug
- [x] defaults
- [x] force_new_dynamic_value
- [ ] get_param_values
- [x] get_value_generator
- [x] inspect_value
- [x] message
- [ ] params
- [ ] pprint
- [x] print_param_defaults
- [x] print_param_values
- [x] script_repr
- [x] set_default
- [x] set_dynamic_time_fn
- [x] set_param
- [x] state_pop
- [x] state_push
- [x] verbose
- [x] warning

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -33,8 +33,8 @@ except ImportError:
     pd = None
 except Exception as e:
     pd = None
-    param.main.warning('Pandas interface failed to import with '
-                       'following error: %s' % e)
+    param.main.param.warning('Pandas interface failed to import with '
+                             'following error: %s' % e)
 
 try:
     from .xarray import XArrayInterface # noqa (Conditional API import)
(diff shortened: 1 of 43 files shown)


philippjfr:
I've changed all the methods that I think should definitely be moved onto the ``param`` subobject, leaving only ``get_param_values``, ``params`` and ``pprint`` which I think are the most useful methods. We should decide whether we want to simply move everything or if we want to leave these commonly useful methods or do something else entirely, e.g. by moving **and** renaming them.

jlstevens:
I quite like `obj.param.values()` but I'm not sure about `obj.param.objects()`.

jbednar:
Got anything better? :-) "values" is obvious; what to call the Parameter objects is much less obvious, because their name is "Parameter" but we have to distinguish it from the value, and "parameter" can mean "value"....

jlstevens:
The `Parameters` object could be callable. So you could use `obj.param('x')` to get the 'x' parameter...

jlstevens:
I'm going to go ahead and merge. I think it is important we make this change and test it as it is now - the remaining changes can go into a subsequent PR.

PR #3263, opened 2018-12-07
Allow Path elements with single item

philippjfr:
- [x] Add unit tests

### DIFF
--- a/holoviews/element/path.py
+++ b/holoviews/element/path.py
@@ -9,7 +9,7 @@ import numpy as np
 
 import param
 from ..core import Element2D, Dataset
-from ..core.data import MultiInterface
+from ..core.data import MultiInterface, datatypes
 from ..core.dimension import Dimension, asdim
 from ..core.util import config, disable_constant
 from .geom import Geometry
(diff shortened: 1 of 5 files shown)


holoviews/element/path.py
```diff
@@ -49,18 +49,18 @@ class Path(Geometry):
...
-                raise ValueError("Path x and y values must be the same length.")
-            data = [np.column_stack((x, y[:, i])) for i in range(y.shape[1])]
+            if y.ndim > 1:
+                if len(x) != y.shape[0]:
+                    raise ValueError("Path x and y values must be the same length.")
+                data = [np.column_stack((x, y[:, i])) for i in range(y.shape[1])]
         elif isinstance(data, list) and all(isinstance(path, Path) for path in data):
-            data = [p for path in data for p in path.data]
+            data = [p for path in data for p in
+                    (path.data if isinstance(path.data, list) else [path.data])]
```
jbednar: This might be clearer as:

```
path_data = path.data if isinstance(path.data, list) else [path.data]
data = [p for path in data for p in path_data]
```
But it's still hard to read all the "path"s and "data"s here....

philippjfr: That's not going to work, we're iterating over the items in each ``path.data``, i.e. you're iterating over the inner loop here which isn't defined by that point.

philippjfr:
I was skeptical about this at first but I've come to the conclusion that this is a good idea. Here's the benefits as I see them:

* ``hv.Path([(0, 1), (1, 2), ...])`` now produces something sensible not a list of length 1 paths, which caught me out a number of times
* It's now possible to unpack a groupby into a Path, so you can finally use ``.to`` with Paths.
* The analogy "Path is to Curve as Points is to Curve" now actually holds making our job explaining the relationship between elements a lot easier or to put it another way the semantics of elements are no longer entangled with their supported data formats.

jbednar:
Wow, that was almost too easy!

jlstevens:
In the end I am happy with this change. Initially, I was doubtful about extending the allowed constructors further but it does seems to be a decent win in terms of usability and consistency.

jbednar:
Yay!

philippjfr:
Ready to review.

philippjfr:
@jlstevens I'll likely merge this once the tests are passing so I can get on with other work that depends on this but please do review this if you get around to it and I'll open a followup PR to address your comments.

PR #3268, opened 2018-12-10
Refactored GraphPlot and fixed renderer ordering

philippjfr:
This PR refactors the ``GraphPlot._init_glyphs`` method so it is a bit less unwieldy and then adds a method which reorders the renderers on the plot depending on the defined ``draw_order``.

- [x] Adds unit tests

Before:

image 1

After:

bokeh_plot


jlstevens:
Thanks for the fix! Merging.

PR #3269, opened 2018-12-10
Update of bokeh reference gallery notebooks

jlstevens:
This PR will update the Bokeh reference gallery. Hopefully it can serve as a template for updating the matplotlib and bokeh reference galleries making those changes quicker to implement.

### DIFF
--- a/examples/reference/elements/bokeh/Curve.ipynb
+++ b/examples/reference/elements/bokeh/Curve.ipynb
@@ -73,7 +73,7 @@
    "source": [
     "overlay =hv.NdOverlay({interp: hv.Curve(points[::8]).opts(interpolation=interp, width=600)\n",
     "                       for interp in ['linear', 'steps-mid', 'steps-pre', 'steps-post']})\n",
-    "overlay.opts(opts.NdOverlay(legend_position='right'))"
+    "overlay.opts(legend_position='right')"
    ]
   },
   {
--- a/examples/reference/elements/bokeh/Curve.ipynb
+++ b/examples/reference/elements/bokeh/Curve.ipynb
@@ -21,6 +21,7 @@
    "source": [
     "import numpy as np\n",
     "import holoviews as hv\n",
+    "from holoviews import opts\n",
     "hv.extension('bokeh')"
    ]
   },
@@ -70,9 +71,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%opts Curve [width=600] NdOverlay [legend_position='right']\n",
-    "hv.NdOverlay({interp: hv.Curve(points[::8]).options(interpolation=interp)\n",
-    "              for interp in ['linear', 'steps-mid', 'steps-pre', 'steps-post']})"
+    "overlay =hv.NdOverlay({interp: hv.Curve(points[::8]).opts(interpolation=interp, width=600)\n",
+    "                       for interp in ['linear', 'steps-mid', 'steps-pre', 'steps-post']})\n",
+    "overlay.opts(opts.NdOverlay(legend_position='right'))"
    ]
   },
   {
--- a/examples/reference/elements/bokeh/Image.ipynb
+++ b/examples/reference/elements/bokeh/Image.ipynb
@@ -68,8 +68,8 @@
    "outputs": [],
    "source": [
     "opts.defaults(\n",
-    "    opts.Image(tools=['hover']),\n",
-    "    opts.Points (color='black', marker='x', size=20))"
+    "   opts.Image(tools=['hover']),\n",
+    "   opts.Points (color='black', marker='x', size=20))"
    ]
   },
   {
--- a/examples/reference/elements/bokeh/Image.ipynb
+++ b/examples/reference/elements/bokeh/Image.ipynb
@@ -21,6 +21,7 @@
    "source": [
     "import numpy as np\n",
     "import holoviews as hv\n",
+    "from holoviews import opts\n",
     "hv.extension('bokeh')"
    ]
   },
@@ -57,6 +58,24 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
+    "In recent versions of Bokeh, you can enable image hover to inspect the raw values:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "opts.defaults(\n",
+    "    opts.Image(tools=['hover']),\n",
+    "    opts.Points (color='black', marker='x', size=20))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
     "Slicing, sampling, etc. on an ``Image`` all operate in this continuous space, whereas the corresponding operations on a ``Raster`` work on the raw array coordinates."
    ]
   },
@@ -66,7 +85,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%opts Image [tools=['hover']]\n",
     "img + img[-0.5:0.5, -0.5:0.5]"
    ]
   },
@@ -83,7 +101,6 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%opts Points (color='black' marker='x' size=20)\n",
     "closest = img.closest((0.1,0.1))\n",
     "print('The value at position %s is %s' % (closest, img[0.1, 0.1]))\n",
     "img * hv.Points([img.closest((0.1,0.1))])"
--- a/examples/reference/elements/bokeh/ItemTable.ipynb
+++ b/examples/reference/elements/bokeh/ItemTable.ipynb
@@ -38,7 +38,7 @@
    "outputs": [],
    "source": [
     "itemtable = hv.ItemTable([('Age', 10), ('Weight',15), ('Height','0.8 meters')])\n",
-    "itemtable.opts(opts.ItemTable(height=60))"
+    "itemtable.opts(height=60)"
    ]
   },
   {
--- a/examples/reference/elements/bokeh/ItemTable.ipynb
+++ b/examples/reference/elements/bokeh/ItemTable.ipynb
@@ -20,6 +20,7 @@
    "outputs": [],
    "source": [
     "import holoviews as hv\n",
+    "from holoviews import opts\n",
     "hv.extension('bokeh')"
    ]
   },
@@ -36,8 +37,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%opts ItemTable [height=60]\n",
-    "hv.ItemTable([('Age', 10), ('Weight',15), ('Height','0.8 meters')])"
+    "itemtable = hv.ItemTable([('Age', 10), ('Weight',15), ('Height','0.8 meters')])\n",
+    "itemtable.opts(opts.ItemTable(height=60))"
    ]
   },
   {
--- a/examples/reference/elements/bokeh/Path.ipynb
+++ b/examples/reference/elements/bokeh/Path.ipynb
@@ -46,7 +46,8 @@
     "def lissajous(t, a, b, delta):\n",
     "    return (np.sin(a * t + delta), np.sin(b * t), t)\n",
     "\n",
-    "hv.Path([lissajous(lin, 3, 5, np.pi/2)]).opts(color='black', line_width=4)"
+    "path = hv.Path([lissajous(lin, 3, 5, np.pi/2)])\n",
+    "path.opts(color='black', line_width=4)"
    ]
   },
   {
@@ -63,8 +64,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "hv.Path([lissajous(lin, 3, 5, np.pi/2)], vdims='time').opts(\n",
-    "    cmap='hsv', color='time', line_width=4)"
+    "path = hv.Path([lissajous(lin, 3, 5, np.pi/2)], vdims='time')\n",
+    "path.opts(opts.Path(cmap='hsv', color='time', line_width=4))"
    ]
   },
   {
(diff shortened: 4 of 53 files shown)


examples/reference/elements/bokeh/Curve.ipynb
```diff
@@ -70,9 +71,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%opts Curve [width=600] NdOverlay [legend_position='right']\n",
-    "hv.NdOverlay({interp: hv.Curve(points[::8]).options(interpolation=interp)\n",
-    "              for interp in ['linear', 'steps-mid', 'steps-pre', 'steps-post']})"
+    "overlay =hv.NdOverlay({interp: hv.Curve(points[::8]).opts(interpolation=interp, width=600)\n",
+    "                       for interp in ['linear', 'steps-mid', 'steps-pre', 'steps-post']})\n",
+    "overlay.opts(opts.NdOverlay(legend_position='right'))"
```
philippjfr: This is a case where I would inline the keyword.

examples/reference/elements/bokeh/ItemTable.ipynb
```diff
@@ -36,8 +37,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%opts ItemTable [height=60]\n",
-    "hv.ItemTable([('Age', 10), ('Weight',15), ('Height','0.8 meters')])"
+    "itemtable = hv.ItemTable([('Age', 10), ('Weight',15), ('Height','0.8 meters')])\n",
+    "itemtable.opts(opts.ItemTable(height=60))"
```
philippjfr: I'd inline the keyword here.

examples/reference/elements/bokeh/Path.ipynb
```diff
@@ -89,8 +90,8 @@
     "adultL = [(-0.3, 0.4), (-0.3, 0.3), (-0.2, 0.3), (-0.2, 0.4),(-0.3, 0.4)]\n",
     "scene = hv.RGB.load_image('../assets/penguins.png')\n",
     "\n",
-    "(scene * hv.Path([adultL, adultR, baby]) * hv.Path([baby])).opts(\n",
-    "    opts.Path(line_width=4))"
+    "overlay = (scene * hv.Path([adultL, adultR, baby]) * hv.Path([baby]))\n",
+    "overlay.opts(opts.Path(line_width=4))"
```
philippjfr: Again, I'd inline.

examples/reference/elements/bokeh/Path.ipynb
```diff
@@ -110,7 +111,8 @@
     "paths = hv.Path((np.arange(N), np.random.rand(N, NLINES) + np.arange(NLINES)[np.newaxis, :]))\n",
     "paths2 = hv.Path((np.arange(N), np.random.rand(N, NLINES) + np.arange(NLINES)[np.newaxis, :]))\n",
     "\n",
-    "(paths * paths2).opts(width=600)"
+    "overlay = paths * paths2\n",
+    "overlay.opts(opts.Overlay(width=600))"
```
philippjfr: Same here.

examples/reference/elements/bokeh/Sankey.ipynb
```diff
@@ -37,14 +37,15 @@
...
+    "sankey = hv.Sankey([\n",
     "    ['A', 'X', 5],\n",
     "    ['A', 'Y', 7],\n",
     "    ['A', 'Z', 6],\n",
     "    ['B', 'X', 2],\n",
     "    ['B', 'Y', 9],\n",
     "    ['B', 'Z', 4]]\n",
-    ").options(width=600, height=400)"
+    ")\n",
+    "sankey.opts(opts.Sankey(width=600, height=400))"
```
philippjfr: This is two keywords, but I'd argue width/height is so common this should also be inlined.

jlstevens:
@jbednar @philippjfr The examples Philipp points out annoy me too but I opted to keep totally consistent. The reasons for doing what I did:

* Maximal consistency. If we can show one good way (or at least acceptable way) to *always* do things without judgement then I think we can avoid confusing users with what has been an abundance of choice.
* What if you only start with one or two keywords but want to add more later? This style offers tab-completion from the start.

Philipp's proposals are more readable and less verbose but also lack tab-completion.

So when should you inline keywords? What are the *exact* rules we can follow for the docs?

* Is it when keywords are deemed common enough? (e.g `width`, `height`, `color` and `cmap`?)
* Is it by number of keywords? E.g a maximum of 2 or 3?
* Is it by length of the line?

Imho even a single keyword should be tab-completable if it is obscure. I think I can agree on the 'common' keywords that can be inlined but if we do that I want an explicit list of the keywords that we deem common and can be inlined. And even then, how many should you inline before using the opts utility again (if only for validation)?

jlstevens:
After discussing this with @jbednar and @philippjfr, we can up with the following two rules for inlining:

Two Rules:

* If you only have a single keyword you can always choose to inline that keyword instead.
* If your keywords are a subset of the following and fit on one line, you can inline them.
    *height*, *width*, *color*, *cmap*, *fig_size*, *title* , *colorbar*, *alpha*, *line_width*, *size*, *marker*

There are two issues to tackle related to this:

* We want to make `title` an alias for `title_format` (`title` doesn't exist yet)
* There is a bug with `.opts` in the Bars notebook at the end (`.options` works though)
* We need to use backticks for single variable names due to a notebook bug.

examples/reference/elements/bokeh/Scatter.ipynb
```diff
@@ -37,10 +38,11 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%opts Scatter (color='k' marker='s' size=10)\n",
     "np.random.seed(42)\n",
     "coords = [(i, np.random.random()) for i in range(20)]\n",
-    "hv.Scatter(coords)"
+    "scatter = hv.Scatter(coords)\n",
+    "\n",
+    "scatter.opts(opts.Scatter(color='k', marker='s', size=10))"
```
philippjfr: All in our allowed list so can be inlined.

examples/reference/elements/bokeh/Table.ipynb
```diff
@@ -55,9 +56,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%opts Table [height=140]\n",
-    "hv.Table({'Gender':gender, 'Age':age, 'Weight':weight, 'Height':height},\n",
-    "         ['Gender', 'Age'],  ['Weight', 'Height'])"
+    "table = hv.Table({'Gender':gender, 'Age':age, 'Weight':weight, 'Height':height},\n",
+    "                 ['Gender', 'Age'],  ['Weight', 'Height'])\n",
+    "table.opts(opts.Table(height=140))"
```
philippjfr: Can be inlined.

examples/reference/elements/bokeh/Table.ipynb
```diff
@@ -73,8 +74,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%opts Table [height=140]\n",
-    "table = hv.Table((gender, age, weight, height), ['Gender', 'Age'], ['Weight', 'Height'])"
+    "table = hv.Table((gender, age, weight, height), ['Gender', 'Age'], ['Weight', 'Height'])\n",
+    "table.opts(opts.Table(height=140))"
```
philippjfr: Same here.

examples/reference/elements/bokeh/VectorField.ipynb
```diff
@@ -37,13 +38,13 @@
...
    "source": [
-    "%%opts VectorField [magnitude='Magnitude']\n",
     "x,y  = np.mgrid[-10:10,-10:10] * 0.25\n",
     "sine_rings  = np.sin(x**2+y**2)*np.pi+np.pi\n",
     "exp_falloff = 1/np.exp((x**2+y**2)/8)\n",
     "\n",
     "vector_data = (x, y, sine_rings, exp_falloff)\n",
-    "hv.VectorField(vector_data)"
+    "vectorfield = hv.VectorField(vector_data)\n",
+    "vectorfield.opts(opts.VectorField(magnitude='Magnitude'))"
```
philippjfr: Single option can be inlined.

jlstevens:
I've now gone through all the bokeh reference notebooks. Here are a couple of notes going through these examples:

* The gif for the Polyedit example is out of date.
* Selection1D_paired only works with `.options` not `.opts`

We should have a rule to say normalization option keywords should go at the end if present.

@philippjfr I think this is ready to review/merge.

examples/reference/elements/bokeh/Path.ipynb
philippjfr: Nevermind, I see it's a composite.

examples/reference/elements/bokeh/Points.ipynb
```diff
@@ -56,8 +59,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%opts Points (color='k' marker='+' size=10)\n",
-    "hv.Points(coords) + hv.Points(coords)[0.6:0.8,0.2:0.5]"
+    "layout = hv.Points(coords) + hv.Points(coords)[0.6:0.8,0.2:0.5]\n",
+    "layout.opts(opts.Points(color='k', marker='+', size=10))"
```
philippjfr: I'd reuse the points defined in the previous cell here, no need to style at all here then.

examples/reference/elements/bokeh/Scatter.ipynb
```diff
@@ -56,8 +58,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%opts Scatter (color='k' marker='s' size=10)\n",
-    "hv.Scatter(coords)[0:12] + hv.Scatter(coords)[12:20]"
+    "layout = hv.Scatter(coords)[0:12] + hv.Scatter(coords)[12:20]\n",
+    "layout.opts(opts.Scatter(color='k', marker='s', size=10))"
```
philippjfr: Again, would reuse the styled scatter from above, to avoid styling here altogether.

examples/reference/elements/bokeh/Spikes.ipynb
```diff
@@ -57,9 +58,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%opts Points (color='red')\n",
     "ys = np.random.rand(50)\n",
-    "hv.Points((xs, ys)) * hv.Spikes(xs)"
+    "overlay = hv.Points((xs, ys)) * hv.Spikes(xs)\n",
+    "overlay.opts(opts.Points(color='red'))"
```
philippjfr: This would fall under the "single element styled inline:

```python
points = hv.Points((xs, ys))
spikes = hv.Spikes(xs)
points.opts(color='red') * spikes
```

but I think that can go either way.

examples/reference/elements/bokeh/Spread.ipynb
```diff
@@ -75,10 +76,10 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "%%opts Spread (fill_color='indianred' fill_alpha=1)\n",
     "xs = np.linspace(0, np.pi*2, 20)\n",
-    "hv.Spread((xs, np.sin(xs), 0.1+np.random.rand(len(xs)), 0.1+np.random.rand(len(xs))),\n",
-    "          vdims=['y', 'yerrneg', 'yerrpos'])"
+    "spread = hv.Spread((xs, np.sin(xs), 0.1+np.random.rand(len(xs)), 0.1+np.random.rand(len(xs))),\n",
+    "                    vdims=['y', 'yerrneg', 'yerrpos'])\n",
+    "spread.opts(opts.Spread(fill_alpha=1, fill_color='indianred'))"
```
philippjfr: I'd inline two keywords.

jlstevens: Agreed.

examples/reference/streams/bokeh/BoundsX.ipynb
```diff
@@ -45,7 +45,7 @@
     "\n",
     "dmap = hv.DynamicMap(make_from_boundsx, streams=[streams.BoundsX(source=curve, boundsx=(0,0))])\n",
     "\n",
-    "curve.options(tools=['xbox_select']) + dmap"
+    "curve.opts(tools=['xbox_select']) + dmap"
```
philippjfr: You did use the single element rule yourself here :smile:

examples/reference/elements/bokeh/Image.ipynb
```diff
@@ -53,6 +54,24 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "opts.defaults(\n",
+    "    opts.Image(tools=['hover']),\n",
+    "    opts.Points (color='black', marker='x', size=20))"
```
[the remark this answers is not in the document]
jlstevens: Well spotted, thanks.

examples/reference/elements/bokeh/Points.ipynb
```diff
@@ -40,14 +41,16 @@
     "%%opts Points (color='k' marker='+' size=10)\n",
     "np.random.seed(12)\n",
     "coords = np.random.rand(50,2)\n",
-    "hv.Points(coords)"
+    "points = hv.Points(coords)\n",
+    "\n",
+    "points.opts(opts.Points(color='k', marker='+', size=10))"
```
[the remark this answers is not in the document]
philippjfr: In fact the exact same example in Scatter below is inlined.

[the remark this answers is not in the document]
jlstevens: Agreed. Three is borderline but they are common keywords which makes inlining fine.

examples/reference/elements/bokeh/Spikes.ipynb
jlstevens: I think I prefer your suggestion accepting the new rule.

jlstevens:
Before I forget what we discussed. Some other guidelines:

* Instead of only inlining a single keyword, you can inline up to three 'common' keywords. After that you don't *have* to use `opts` but should seriously consider whether you prefer the improved readability over validation and tab-completion. A keyword can always be considered common if it is mentioned in the immediately surrounding markdown text.

* You can inline a *single* simple `.opts` call in a composition if the composition is made up of handles. E.g this is okay `penguins * bounds.opts(color='orange', linewidth=6)`. The `.opts` call isn't simple if you use the `opts` utility.

PR #3275, opened 2018-12-11
Make matplotlib's BarPlot consistent with other backends

philippjfr:
Fixes bug in matplotlib BarPlot where Bars are not correctly aligned with ticks.

Broken:

download 21

Fixed:

download 20

### DIFF
--- a/examples/gallery/demos/matplotlib/bars_economic.ipynb
+++ b/examples/gallery/demos/matplotlib/bars_economic.ipynb
@@ -17,6 +17,8 @@
    "source": [
     "import pandas as pd\n",
     "import holoviews as hv\n",
+    "from holoviews import opts\n",
+    "\n",
     "hv.extension('matplotlib')"
    ]
   },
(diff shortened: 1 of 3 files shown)


philippjfr:
I couldn't put up with the horrendous API that the matplotlib BarPlot provided so for now I've simply locked it down so that the API matches plotly and bokeh plots.

philippjfr:
I think the notebook diffs say it all it's gone from:

```python
grouped = bars.relabel('Grouped').opts(category_index=5, color_by=['stack'], stack_index=1)
stacked = bars.relabel('Stacked').opts(stack_index=5)
grouped + stacked
```
to:

```
stacked = bars.relabel('Stacked')
grouped = bars.relabel('Grouped')
grouped + stacked.opts(stacked=True)
```

I maintain that the matplotlib BarPlot is the worst bit of code I've ever written.

jlstevens:
I am very happy to see the API made consistent. Personally, things like `category_index=5` never made much sense to me so I like the new version a *lot* more.

There's bound to be bugs but fixing the API is important enough that I'm not going to hesitate merging this now. I suspect these bugs will be the motivation for cleaning up the internal code for good.

PR #3278, opened 2018-12-11
Plotly dynamic map fixes

jonmmease:
In the process of working on `Selection1D` stream support to the plotly backend, I've hit a couple of issues with updating plotly figures inside a DynamicMap.

Setup:
```python
# Imports
import numpy as np
import holoviews as hv
import param
hv.extension('plotly')

# Data
data = np.random.multivariate_normal((0, 0), [[1, 0.1], [0.1, 1]], (100,))

# Stream
Inds = hv.streams.Stream.define('Inds', index=param.List(default=[]))
inds = Inds(index=[1, 2, 3])

# Points Element
points = hv.Points(data).options(color='gray')

# Dynamic map the generates a single point at mean of indexes provided by the stream
def mean_point(index):
    if index:
        means = data[index, :].mean(axis=0)
        return hv.Points([means]).options(size=12, color='red')
    else:
        return hv.Points([])
mean_sel = hv.DynamicMap(mean_point, kdims=[], streams=[inds])
```

**Case 1)** DynamicMap of single `hv.Points` element

```python
mean_sel
```
newplot 26

**Case 2)** DynamicMap of overlay of two `hv.Points` elements
```python
points * mean_sel
```
newplot 27

**Case 3)** DynamicMap of `Layout` of two `hv.Points` elements side-by-side
```python
points + mean_sel
```
newplot 28

Here the stream was initialized to the list `[1, 2, 3]` and all three cases display properly initially.  Now, update the stream with:

```python
inds.event(index=[4, 5, 6])
```

Prior to this PR, only the output of case 1 would update.  Case 2 and Case 3 remained unchanged. 

**Summary of changes:**
I'm operating at the edge of my understanding of HoloViews' internals here, so I'd appreciate any help on identifying if there are better ways to tackle these issues.

 - Case 2 was fixed by fe95392. Basically the overlay plot layers weren't being regenerated when the overlay plot's `generate_plot` method is called.
 - Case 3 was partially fixed by 19c1320.  A comm was not being created for the plotly LayoutPlot at all, so I pulled a few lines from the constructor of the Bokeh LayoutPlot to do this.  

The other problem with Case 3, and the part I understand the least, is that the subplots were not being regenerated when the LayoutPlot's `refresh` method was called.  I overrode `refresh` and added an explicit call to `initialize_plot` before calling the superclass method in 206b67d.  With this change the plots in case 3 do update when the stream is updated.  But a new problem is that the `mean_point` function ends up being called twice each time the stream is updated.

Thanks for any help on this 🙂  Once we get this worked out, I think I'm pretty close to having `Selection1D` stream support working!

### DIFF
--- a/holoviews/plotting/plotly/plot.py
+++ b/holoviews/plotting/plotly/plot.py
@@ -2,12 +2,14 @@ from __future__ import absolute_import, division, unicode_literals
 
 import param
 
+from holoviews.plotting.util import attach_streams
 from ...core import (OrderedDict, NdLayout, AdjointLayout, Empty,
                      HoloMap, GridSpace, GridMatrix)
 from ...element import Histogram
 from ...core.options import Store
 from ...core.util import wrap_tuple
-from ..plot import DimensionedPlot, GenericLayoutPlot, GenericCompositePlot
+from ..plot import DimensionedPlot, GenericLayoutPlot, GenericCompositePlot, \
+    GenericElementPlot
 from .util import figure_grid
 
 
@@ -46,6 +48,12 @@ class LayoutPlot(PlotlyPlot, GenericLayoutPlot):
     def __init__(self, layout, **params):
         super(LayoutPlot, self).__init__(layout, **params)
         self.layout, self.subplots, self.paths = self._init_layout(layout)
+
+        if self.top_level:
+            self.comm = self.init_comm()
+            self.traverse(lambda x: setattr(x, 'comm', self.comm))
+            self.traverse(lambda x: attach_streams(self, x.hmap, 2),
+                          [GenericElementPlot])
 
     def _get_size(self):
         rows, cols = self.layout.shape
--- a/holoviews/plotting/plotly/plot.py
+++ b/holoviews/plotting/plotly/plot.py
@@ -229,6 +229,10 @@ class LayoutPlot(PlotlyPlot, GenericLayoutPlot):
         self.handles['fig'] = fig
         return self.handles['fig']
 
+    def refresh(self, **kwargs):
+        # Initialize layout plot before executing standard refresh logic
+        self.initialize_plot()
+        super(LayoutPlot, self).refresh(**kwargs)
 
 
 class AdjointLayoutPlot(PlotlyPlot):
--- a/holoviews/plotting/plotly/element.py
+++ b/holoviews/plotting/plotly/element.py
@@ -431,6 +431,9 @@ class OverlayPlot(GenericOverlayPlot, ElementPlot):
     def generate_plot(self, key, ranges):
         element = self._get_frame(key)
 
+        # Compute subplot objects
+        self.subplots = self._create_subplots(ranges)
+
         ranges = self.compute_ranges(self.hmap, key, ranges)
         figure = None
         for okey, subplot in self.subplots.items():


philippjfr:
I've made some updates to make it consistent with other backends. It's not ideal because it duplicates code a bit but seems to work. It still calls the callback twice but I think we can deal with that later. Could you do some testing?

philippjfr:
I'm starting to think there's a general bug causing the two calls, still looking into that.

philippjfr:
Thanks, was about to push a fix.

philippjfr:
Pushed some further improvements. At some this stuff should all be factored out onto the baseclass.

holoviews/plotting/plot.py
```diff
@@ -1457,8 +1457,16 @@ def _get_frame(self, key):
...
+            clone = item.map(lambda x: x)
+
+            # Ensure that DynamicMaps in the cloned frame have
+            # identical callback inputs to allow memoization to work
+            for it1, it2 in zip(item.traverse(lambda x: x), clone.traverse(lambda x: x)):
+                if isinstance(it1, DynamicMap):
+                    with util.disable_constant(it2.callback):
+                        it2.callback.inputs = it1.callback.inputs
+            frame = clone.map(lambda x: get_plot_frame(x, key_map, cached=cached),
+                              [DynamicMap, HoloMap], clone=False)
```
philippjfr: This stuff is a bit ugly and should probably become a utility. The main problem is that ``.map`` is being abused a bit.

holoviews/plotting/plotly/plot.py
```diff
@@ -28,6 +31,31 @@ def state(self):
...
+        key = tuple(None if d in stream_params else k
+                    for d, k in zip(self.dimensions, key))
+        stream_key = wrap_tuple_streams(key, self.dimensions, self.streams)
+
+        # Update if not top-level, batched or an ElementPlot
+        if self.top_level:
+            self.update(stream_key)
+        else:
+            self.current_key = None
+            self.current_frame = None
```
philippjfr: Should probably just factor this bit out. I'll think about it and probably clean this up before merging.

jonmmease:
Thanks a lot for your help on this one @philippjfr, this working well in my testing now.  I can confirm that the callback is only being called once now and everything looks to be updating appropriately.

philippjfr:
I did have a question for you about the general architecture since it differs a fair bit from the bokeh and matplotlib backends. Basically in those backends we create the plot objects (i.e. matplotlib Figure/Axes and bokeh models) only once and then update them inplace, while the plotly backend resorts to rebuilding the figures from scratch. That certainly makes things a lot easier in the code but I'm wondering whether you think there is a lot of performance to be gained by modifying existing objects inplace? I'm guessing the only real gain we'd get is if we computing diffs, which would let us use ``Plotly.update`` or similar API on the frontend but then I don't really know enough about plotly yet.

jonmmease:
Now that the plotly figures are being represented as native Python `dict` objects, I don't expect there's much cost associated with rebuilding them from scratch each time, so I'd lean towards retaining the relative simplicity at this point.

That said, eventually I'd like to incorporate an optimization for sending diffs to the front end. For the common case where the number of traces in the figure hasn't changed I think it would be pretty straightforward to compute a per-trace diff on the Python side and then apply the diff using `Plotly.update` on the JavaScript side.  I think there are even some utilities in plotly.py that we could use for computing the recursive diffs.

I'd also like to work through how to take advantage of the binary serialization approach you use for the bokeh backend).

philippjfr:
We should probably meet to discuss things at some point. We are currently planning to start building holoviews' display machinery and widgets on top of panel and therefore basically render everything as bokeh models. If we go down that route we get binary serialization basically for free, because ``panel`` already implements a ``PlotlyPlot`` bokeh model which searches the figure for arrays and sticks them inside a bokeh ``ColumnDataSource`` which enables binary serialization, on the other end the json representation is then patched with the contents of the data sources.

jonmmease:
Thanks for the background, that's good to know and sounds like a nice approach. I'll hold off on any optimization work for the time being in that case.

I have a WIP branch that builds on this one and adds `Selection1D` stream support.  I'd like to get that in front of you soon (hopefully tomorrow) because I expect the approach I'm using may need to change somewhat along with the shift to `panel`.

jonmmease:
awesome, thanks for pulling this over the finish line @philippjfr!

PR #3279, opened 2018-12-11
Updates to the getting started guide notebooks

jlstevens:
This PR fixes a number of issues in the getting started guide notebooks but is primarily about getting rid of the magics and using `%%opts`. Most of the changes are in the customization and gridded data guides.

- [x] Update the gifs in the Live Data notebook.


examples/getting_started/2-Customization.ipynb
```diff
@@ -42,6 +42,7 @@
    "source": [
     "import pandas as pd\n",
     "import holoviews as hv\n",
+    "from holoviews import opts\n",
```
philippjfr: Commenting for the code three cells down, I think we should switch from ``group`` to ``label`` there if we're not using the group for anything:

```python
curve = hv.Curve(spike_train, 'milliseconds', 'Hertz', label='Firing Rate')
spikes = hv.Spikes(spike_train.sample(300), 'milliseconds', [], label='Spike Train')
curve + spikes
```

philippjfr:
I'm largely happy with the changes here but have two main comments:

1. I realize the customization notebook can't explain the recommendations about inlining vs. using the opts utility in detail but it should mention inlining for simple cases if only not to confuse users when they see it later.

2. The corresponding customization user guide is where the real meat is and I think that probably needs an extensive rewrite. That means I'm happy to keep the getting started guide short though.

jlstevens:
I agree with you on both counts though I don't want to spend much time talking about inlining as it will be expanded up in the user guide.

examples/getting_started/2-Customization.ipynb
jlstevens: Fixed in 550d9e61

examples/getting_started/3-Tabular_Datasets.ipynb
```diff
@@ -199,9 +204,12 @@
...
    "outputs": [],
    "source": [
-    "%%opts Curve [width=600]\n",
     "agg = ds.aggregate('Year', function=np.mean, spreadfn=np.std)\n",
-    "(hv.Curve(agg) * hv.ErrorBars(agg,vdims=['measles', 'measles_std']).iloc[::2]).redim.range(measles=(0, None))"
+    "errorbars = hv.ErrorBars(agg,vdims=['measles', 'measles_std']).iloc[::2]\n",
+    "overlay =  (hv.Curve(agg) * errorbars).redim.range(measles=(0, None))\n",
+    "\n",
+    "overlay.opts(\n",
+    "    opts.Curve(width=600))"
```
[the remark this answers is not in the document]
jlstevens: Fixed in dcca94b6

jlstevens:
@philippjfr I implemented you suggestion in 0ce29784, adding a short section about discovering options. 

Other than updating gifs (something that needs to happen outside this PR), I think this is now ready. That said, I would appreciate it if you and @jbednar could review this carefully. This is supposed to be people's very first contact with the options system!

examples/getting_started/2-Customization.ipynb
```diff
@@ -181,20 +174,32 @@
...
+    "dotted  = hv.Curve(spike_train, 'milliseconds', 'Hertz')\n",
+    "dotted.opts(dotted_options)"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "When using ``.options`` to apply options directly to an individual object we do not have to explicitly declare which object the options apply to, however often it is useful to set options on a composite object. In these cases the options can be declared as a dictionary of the type name and the options. The code below is therefore equivalent to the syntax we used above:"
+    "When working with elements, you don't have to use the `opts` utility. If you know the keywords you wish to use or have finalized a visualization and want to maximize readability for publication, you can inline the keywords instead:"
```
philippjfr: Not just with elements, ``opts`` makes sense when you want to customize only the top-level object, which in most cases will be true when styling elements but also frequently will work for setting the width/height of an overlay for instance. Not sure how best to explain that though. Would be interested to see what @jbednar thinks.

examples/getting_started/4-Gridded_Datasets.ipynb
```diff
@@ -85,38 +86,48 @@
...
+   "outputs": [],
    "source": [
-    "To see more details on working with different datatypes have a look at the [user guide](../user_guide/08-Gridded_Datasets.ipynb)."
+    "opts.defaults(\n",
+    "    opts.Spread(width=600),\n",
+    "    opts.GridSpace(shared_xaxis=True, shared_yaxis=True),\n",
+    "    opts.Image(cmap='viridis', width=400, height=400),\n",
+    "    opts.Path(color='white'),\n",
+    "    opts.Labels(text_color='white', text_font_size='8pt', text_align='left', text_baseline='bottom'),\n",
+    "    opts.Overlay(show_legend=False))"
```
philippjfr: Should we agree to list things in alphabetical order? If so it would be good to change this. Also while I agree defaults are often useful I worry that putting a huge block of options like this is intimidating and hard to understand since it's unclear what it will apply to.

examples/getting_started/5-Live_Data.ipynb
```diff
@@ -181,24 +195,18 @@
    "source": [
     "def interactive_clifford(a,b,c,d,x=0,y=0):\n",
     "    coords = clifford(a,b,c,d,x0=x,y0=y)\n",
-    "    return (hv.Curve(coords) * hv.Points(coords[0]) * hv.Curve(coords[:2], group='Init')\n",
-    "            * hv.Text(-0.75,1.35, 'x:{x:.2f} y:{y:.2f}'.format(x=coords[0][0],y=coords[0][1])))"
+    "    text = 'x:{x:.2f} y:{y:.2f}'.format(x=coords[0][0],y=coords[0][1])\n",
+    "    return (  hv.Points(coords)                                          # Attractor points\n",
+    "            * hv.Points(coords[0]).opts(color='black', size=10, alpha=1) # Starting point\n",
+    "            * hv.Curve(coords[:2], group='Init')                         # Line to first iteration\n",
+    "            * hv.Text(0,1.75, text))                                  # Starting position text"
```
philippjfr: Have to say, I really dislike this indentation style, I know you disagree but I find it much more readable to put the ``*`` at the end of a line.

examples/getting_started/4-Gridded_Datasets.ipynb
jlstevens: The current recommendation is alphabetical order with containers at the end (`Layout`, then `Overlay`). So in the above I need to swap the positions of `Path` and `Labels`.

examples/getting_started/5-Live_Data.ipynb
jlstevens: Sure. I have no strong opinion but I think I have a mild preference for your suggestion.

examples/getting_started/2-Customization.ipynb
jlstevens: I decided to gloss over that for the reasons you mention - it is hard to explain nicely. Happy to take suggestions.

examples/getting_started/4-Gridded_Datasets.ipynb
jlstevens: Actually, there are more elements mentioned here I need to sort.

jlstevens: The final ordering is:

```python
opts.defaults(
    opts.GridSpace(shared_xaxis=True, shared_yaxis=True),
    opts.Image(cmap='viridis', width=400, height=400),
    opts.Labels(text_color='white', text_font_size='8pt', text_align='left', text_baseline='bottom'),
    opts.Path(color='white'),
    opts.Spread(width=600),
    opts.Overlay(show_legend=False))
```

jlstevens: Fixed in 3126858

examples/getting_started/5-Live_Data.ipynb
jlstevens: Looking at it, I did this for this case due to the comments at the end of each line. Changed it anyway in d9d8adee

examples/getting_started/2-Customization.ipynb
jbednar: I'm ok with glossing over that distinction here and not mentioning it.

examples/getting_started/1-Introduction.ipynb
```diff
@@ -183,7 +184,7 @@
    "source": [
     "## Compositional  ``Overlays``\n",
     "\n",
-    "Once again, we can easily visualize this data with HoloViews by passing our array to [``hv.Image``](../reference/elements/bokeh/Image.ipynb) to create the ``image`` object.  This object has the spatial extent of the data declared as the ``bounds``, in terms of the corresponding range of latitudes and longitudes."
+    "Once again, we can easily visualize this data with HoloViews by passing our array to [``hv.Image``](../reference/elements/bokeh/Image.ipynb) to create the an object named ``image``.  This object has the spatial extent of the data declared as the ``bounds``, in terms of the corresponding range of latitudes and longitudes."
```
jbednar:
```suggestion
    "Once again, we can easily visualize this data with HoloViews by passing our array to [``hv.Image``](../reference/elements/bokeh/Image.ipynb) to create an object named ``image``.  This object has the spatial extent of the data declared as the ``bounds``, in terms of the corresponding range of latitudes and longitudes."
```

jlstevens:
@jbednar Your changes in 5d4cae4 look good!

jlstevens:
I've updated the GIFs in Live Data as well as one of the other outdated GIFs I saw in the bokeh streams example (with a minor change to that notebook).

doc/install.rst
```diff
@@ -53,7 +53,7 @@ for running unit tests, you can use::
 Between releases, development snapshots are made available on conda and
 can be installed using::
 
-  conda install -c ioam/label/dev holoviews
+  conda install -c pyviz/label/dev holoviews
```
philippjfr: I swear I've made this change multiple times now and it never stuck.

jlstevens:
@jbednar Thanks for the fixes and improvements. At this stage, I am happy with the updated getting started guides. Happy to merge if @philippjfr agrees.

examples/getting_started/1-Introduction.ipynb
```diff
@@ -200,7 +201,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "HoloViews supports ``numpy``, ``xarray``, ``iris``, and ``dask`` arrays when working with array data (see [Gridded Datasets](../user_guide/Gridded_Datasets.ipynb)).  We can also compose elements containing array data with those containing tabular data. To illustrate, let's pass our tabular station data to a [``Points``](../reference/elements/bokeh/Points.ipynb) element which is used to mark positions in two-dimensional space:"
+    "HoloViews supports ``numpy``, ``xarray``, ``iris``, and ``dask`` arrays when working with array data (see [Gridded Datasets](4-Gridded_Datasets.ipynb)).  We can also compose elements containing array data with those containing tabular data. To illustrate, let's pass our tabular station data to a [``Points``](../reference/elements/bokeh/Points.ipynb) element, which is used to mark positions in two-dimensional space:"
```
philippjfr: Let's drop iris here, it's moved to geoviews and seems pretty niche anyway.

jlstevens: Fine by me!

examples/getting_started/2-Customization.ipynb
```diff
@@ -16,13 +16,13 @@
     "HoloViews elements like the `Scatter` points illustrated in the [Introduction](1-Introduction.ipynb) contain two types of information:\n",
     "\n",
     "- **Your data**, in as close to its original form as possible, so that it can be analyzed and accessed as you see fit.\n",
-    "- **Metadata specifying what your data *is***, which allows HoloViews to construct a visual representation for it.\n",
+    "- **Metadata specifying what your data *is***, which allows HoloViews to construct an appropriate visual representation for it.\n",
```
philippjfr: Unbalanced ``*``s I think.

examples/getting_started/5-Live_Data.ipynb
```diff
@@ -207,10 +217,11 @@
...
    "outputs": [],
    "source": [
-    "%%opts Curve (line_width=0.03 color='blue') Points (color='red' size=10) Curve.Init (color='red' line_width=2)\n",
+    "from holoviews.streams import PointerXY\n",
+    "\n",
     "# When run live, this cell's output should match the behavior of the GIF below\n",
-    "dmap = hv.DynamicMap(interactive_clifford, kdims=['a','b','c','d'], streams=[PointerXY(x=0,y=0)])\n",
-    "dmap.redim.range(a=(-1.4,-1),b=(1.6,1.8),c=(1,1.5),d=(0.7,0.8))"
+    "iclifford = hv.DynamicMap(interactive_clifford, kdims=['a','b','c','d'], streams=[PointerXY(x=0,y=0)])\n",
+    "iclifford.redim.range(a=(-1.4,-1),b=(1.6,1.8),c=(1,1.5),d=(0.7,0.8), x=(-2,2), y=(-2,2))"
```
philippjfr: What's up with x padding here? I think mixing redim for the DynamicMap dimensions and to apply padding is confusing. I'd prefer using the padding plot option instead.

examples/getting_started/5-Live_Data.ipynb
```diff
@@ -236,7 +246,7 @@
...
+    "``DynamicMap`` and ``Streams`` allow specification of exciting, dynamic visualizations that let you build full-featured interactive applications and simulations with very little code (particularly when combined with a declarative widget library like [Panel](https://panel.pyviz.org)). The way these dynamic visualizations work is that HoloViews runs JavaScript in your browser, which then communicates with a running Python server process that may be running in the Jupyter notebook server or in the [Bokeh server](http://bokeh.pydata.org/en/latest/docs/user_guide/server.html). This Python process may be running locally on your machine or on some remote internet or local-network server. Regardless of where it is running, this Python process executes the callback you supply to ``DynamicMap``, allowing HoloViews to update your visualization whenever the parameters change.\n",
```
philippjfr: This section should probably link to the dashboard and bokeh deployment user guides.

examples/getting_started/5-Live_Data.ipynb
jbednar: To me it seems simpler to sneak in the x,y padding here rather than draw attention to it using the padding plot option.  It's not really padding, anyway, at least not just for visual appearance; it's to account for the fact that some values of the parameters end up with plots larger than others, and the initial one doesn't cover the full range expected across all parameters.

philippjfr:
My only other comment is a general one about ``.opts``, we argue that the inplace behavior helps you style things progressively, refining the representation. Having worked with it a bit in the last couple of days I have to express some reservations about it. 

The main issue with inplace styling is that when you apply an option and later decide you don't want it anymore inplace styling will persist it regardless unless you explicitly override the option. This makes sense but it will be confusing to a user since they can't easily inspect the options that are applied and resetting the options is a bit obscure (an empty ``.opts`` call). I think this means we need to devote at least a bit of explanation about applying options using inplace vs. clone operations and I think we should also prioritize adding an API to see the options that are applied.

Finally I would also like an explicit API to clear options, something like ``.opts(clear=True)``, since I think an empty options call is too implicit and actually clashes with the regular semantics of option setting which merges options rather than overriding them entirely.

None of this needs to hold up this PR but I think we should address these issues before release.

examples/getting_started/5-Live_Data.ipynb
philippjfr: I'd use ``xlim`` and ``ylim`` in that case but I'm okay with sneaking it in.

jlstevens:
Not just any old namespace conflict actually, I think `clear` could very easily be mistaken for a plot or style option. For this reason, I vote strong in favor of `opts.clear` as I think that is much clearer (no pun intended!)

philippjfr:
>(though I don't think options should be shown by default)

I think it probably should since you can simply ``print(obj)`` for the default repr.

>Implementing this would take a bit more work so I dunno if we want to delay the release for it.

I **think** it's worth delaying a release for, but then I also don't think it's much work. An initial ugly prototype took 5 minutes and adding a few lines of code to the existing pretty-printer. I guess the hardest problem, as always, is deciding on a format we all like.

jlstevens:
> I think it probably should since you can simply print(obj) for the default repr.

I can buy that argument: ``__str__`` would be the same as it is now and some pretty printer (either a utility, or maybe overloading params ``pprint`) could be configured to generate the textual representation in the notebook.

> I guess the hardest problem, as always, is deciding on a format we all like.

My default vote is to actually use the `Options` repr as is. Unlike elements which can have (huge) data, these objects have a relatively clear and working repr by default. The tricky thing is all the crazy indentation logic but maybe that problem is solved already (i.e just hook into what is there).

philippjfr:
>I don't disagree with any of that though I might suggest opts.clear(obj) instead of obj.opts(clear=True) which introduces another potential namespace conflict.

I'm fine with a ``opts.clear`` method.

jbednar:
To me it's more natural to specify clearing using the .opts() method than to require a global utility, but I agree that a keyword argument to .opts would be a namespace conflict, and I agree that clearing on an empty keyword list (or even an empty dict argument) is confusing.  .opts.clear() would be nicer, but I think .opts is just a normal method, so that won't work. Hmm.

examples/getting_started/1-Introduction.ipynb
jlstevens: Fixed in f2ce0219

examples/getting_started/2-Customization.ipynb
jlstevens: I think it is correct:

**Metadata specifying what your data *is***

philippjfr: Nevermind, missed the third closing ``*``

examples/getting_started/5-Live_Data.ipynb
jlstevens: Addressed in 9a68ceb7

jlstevens:
I have now implemented @philippjfr suggestions. I would recommend merging this PR now so I can work on the clearing/pretty printing API. Afterwards, we can revisit the customization notebook if we feel it is important enough to mention these new features.

philippjfr:
Agreed, I'll merge and if we decide to do so we can add small sections about the pretty printer and clear methods later.

jlstevens:
The suggestion of `obj.opts.clear()` is implemented in https://github.com/ioam/holoviews/pull/3292

PR #3283, opened 2018-12-12
Aliased 'title_format' parameter to 'title'

jlstevens:
Hopefully a simple PR. If the tests pass after the first commit I can update the tests and then the notebooks.

### DIFF
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -1079,7 +1079,7 @@ class GenericElementPlot(DimensionedPlot):
             if custom_title and self.title_format:
                 self.warning('Both title and title_format set. Using title_format parameter')
 
-            title = self.title_format if self.title_format else self.title
+            title = self.title if custom_title or self.title_format is None else self.title_format
             title_format = util.bytes_to_unicode(title)
             title = title_format.format(label=util.bytes_to_unicode(label),
                                         group=util.bytes_to_unicode(group),
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -207,9 +207,7 @@ class DimensionedPlot(Plot):
         The formatting string for the title of this plot, allows defining
         a label group separator and dimension labels.""")
 
-    title_format = param.String(default="{label} {group}\n{dimensions}", doc="""
-        The formatting string for the title of this plot, allows defining
-        a label group separator and dimension labels. Aliased to title.""")
+    title_format = param.String(default=None, doc="Alias for title.")
 
     normalize = param.Boolean(default=True, doc="""
         Whether to compute ranges across all Elements at this level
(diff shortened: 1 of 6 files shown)


holoviews/plotting/plot.py
```diff
@@ -1473,7 +1486,7 @@ def _format_title(self, key, dimensions=True, separator='\n'):
         type_name = type(self.layout).__name__
         group = util.bytes_to_unicode(layout.group if layout.group != type_name else '')
         label = util.bytes_to_unicode(layout.label)
-        title = util.bytes_to_unicode(self.title_format).format(label=label,
+        title = util.bytes_to_unicode(self.title).format(label=label,
                                                                 group=group,
```
philippjfr: Not major but indentation is off now.

holoviews/plotting/plot.py
```diff
@@ -243,6 +247,15 @@ def __init__(self, keys=None, dimensions=None, layout_dimensions=None,
...
         self._updated = False # Whether the plot should be marked as updated
 
+        # Setting up alias
+        if 'title' in params and 'title_format' in params:
+            if params['title'] != params['title_format']:
+                self.warning('The title and title_format parameters do not match. Using title.')
+        elif 'title_format' in params:
+            params['title'] = params['title_format']
+        elif 'title' in params:
+            params['title_format'] = params['title']
```
philippjfr: This is not the right place to do these checks since it can be set dynamically and title format would currently be ignored. Generally that means doing this where the parameter is actually used and not altering the actual parameter values on the plot.

holoviews/plotting/plot.py
```diff
@@ -1073,7 +1074,13 @@ def _format_title(self, key, dimensions=True, separator='\n'):
                 dim_title = self._frame_title(key, separator=separator)
             else:
                 dim_title = ''
-            title_format = util.bytes_to_unicode(self.title_format)
+
+            custom_title = (self.title != self.param.params('title').default)
+            if custom_title and self.title_format:
+                self.warning('Both title and title_format set. Using title_format parameter')
+
+            title = self.title_format if self.title_format else self.title
```
philippjfr: In other aliasing cases the recommended option takes precedence, I don't feel strongly about it though, so it's up you. The logic would be:

```suggestion
            title = self.title if custom_title or self.title_format is None else self.title_format
```

holoviews/plotting/plot.py
```diff
@@ -1473,10 +1480,17 @@ def _format_title(self, key, dimensions=True, separator='\n'):
...
-        title = util.bytes_to_unicode(self.title_format).format(label=label,
-                                                                group=group,
-                                                                type=type_name,
-                                                                dimensions=dim_title)
+
+
+        custom_title = (self.title != self.param.params('title').default)
+        if custom_title and self.title_format:
+            self.warning('Both title and title_format set. Using title parameter')
+        title_str = self.title_format if self.title_format else self.title
```
```suggestion
            title_str = self.title if custom_title or self.title_format is None else self.title_format
```

jlstevens:
I've committed @philippjfr's suggestions and I am happy to let @jbednar merge if he is happy with the updated precedence.

philippjfr:
Tests are passing so I'll merge.

PR #3285, opened 2018-12-13
Updated Dashboard user guide

philippjfr:
Updates dashboard user guide from using parambokeh to using panel.


jlstevens:
Thanks! Very happy with the updated way of doing things.

PR #3287, opened 2018-12-13
Added support for backends argument to opts.defaults

jlstevens:
In future this might helps us remove all direct uses of option trees.

### DIFF
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -189,18 +189,18 @@ class opts(param.ParameterizedFunction):
         return obj_handle
 
     @classmethod
-    def _process_magic(cls, options, strict):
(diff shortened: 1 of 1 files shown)


holoviews/util/__init__.py
```diff
@@ -215,21 +215,25 @@ def _cellmagic(cls, options, obj, strict=False):
...
+            StoreOptions.apply_customizations(options, Store.options(backend=backend))
 
 
     @classmethod
-    def defaults(cls, *options):
+    def defaults(cls, *options, **kwargs):
         """
         Set default options for a session, whether in a Python script or
         a Jupyter notebook.
         """
```
philippjfr: Would be nice to have a Google style docstring here documenting the backend kwarg.

PR #3289, opened 2018-12-14
Add options to PrettyPrinter

philippjfr:
Adds options to the ``PrettyPrinter``, very much a work in progress but the basics work. 

```
> p1 = hv.Points([1, 2, 3]).opts(color='red')
> c1 = hv.Curve([]).opts(alpha=0.3, xticks=[1, 2, 3])
> print(p1)

:Points   [x,y]
 Options(color='red')

> print(hv.HoloMap({0: c1*p1}))

:HoloMap   [Default]
   :Overlay
      .Curve.I  :Curve   [x]   (y)
       Options(alpha=0.3, xticks=[1, 2, 3])
      .Points.I :Points   [x,y]
       Options(color='red')

> print(p1 + (c1 * p1).opts(xaxis=None))

:Layout
   .Points.I  :Points   [x,y]
    Options(color='red')
   .Overlay.I :Overlay
    Options(xaxis=None)
      .Curve.I  :Curve   [x]   (y)
       Options(alpha=0.3, xticks=[1, 2, 3])
      .Points.I :Points   [x,y]
       Options(color='red')
```

To do:

- [x] Decide on nice formatting
- [x] Parameterize PrettyPrinter to allow toggling between showing options at all and whether to show default options
- [x] Add tests

### DIFF
--- a/holoviews/core/pprint.py
+++ b/holoviews/core/pprint.py
@@ -337,7 +337,7 @@ class PrettyPrinter(object):
         else:
             fst_lvl = level
 
-        if opts:
+        if opts and opts.kwargs:
             lines.append((fst_lvl, ' * ' + str(opts)))
         return (lvl, lines)
 
@@ -347,7 +347,7 @@ class PrettyPrinter(object):
         Return the information summary for an Element. This consists
         of the dotted name followed by an value dimension names.
         """
-        info =  cls.component_type(node)
+        info = cls.component_type(node)
         if len(node.kdims) >= 1:
             info += cls.tab + '[%s]' % ','.join(d.name for d in node.kdims)
         if value_dims and len(node.vdims) >= 1:
--- a/holoviews/core/pprint.py
+++ b/holoviews/core/pprint.py
@@ -338,7 +338,7 @@ class PrettyPrinter(object):
             fst_lvl = level
 
         if opts and opts.kwargs:
-            lines.append((fst_lvl, ' * ' + str(opts)))
+            lines.append((fst_lvl, ' ' + str(opts)))
         return (lvl, lines)
 
     @classmethod
--- a/holoviews/core/pprint.py
+++ b/holoviews/core/pprint.py
@@ -365,6 +365,8 @@ class PrettyPrinter(param.Parameterized):
 
     @bothmethod
     def option_info(cls_or_slf, node):
+        if not cls_or_slf.show_options:
+            return None
         from .options import Store, Options
         options = {}
         for g in ['plot', 'style', 'norm']:
--- a/holoviews/core/pprint.py
+++ b/holoviews/core/pprint.py
@@ -405,12 +405,11 @@ class PrettyPrinter(param.Parameterized):
         if cls_or_slf.show_options and opts and opts.kwargs:
             lines += [(level, l) for l in cls_or_slf.format_options(opts)]
 
-        additional_lines = []
         if len(node.data) == 0:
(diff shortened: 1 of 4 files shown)


jbednar:
Looks great!

philippjfr:
I just made a small edit. @jlstevens preferred one less indent, I'm not quite sure what I prefer.

holoviews/core/pprint.py
```diff
@@ -363,23 +379,17 @@ def adjointlayout_info(cls, node, siblings, level, value_dims):
...
         # .last has different semantics for GridSpace
         last = list(node.data.values())[-1]
-        if hasattr(last, 'children'):
-            additional_lines = cls.recurse(last, level=level)
-        # NdOverlays, GridSpace, Ndlayouts
-        elif last is not None and getattr(last, '_deep_indexable'):
-            level, additional_lines = cls.ndmapping_info(last, [], level, value_dims)
-        else:
-            _, additional_lines = cls.element_info(last, siblings, level, value_dims)
+        additional_lines = cls.recurse(last, level=level, value_dims=value_dims)
```
philippjfr: Need to double check this, may have to re-add:

```suggestion
        if last is not None and getattr(last, '_deep_indexable'):	
             level, additional_lines = cls.ndmapping_info(last, [], level, value_dims)
        else:
            additional_lines = cls.recurse(last, level=level, value_dims=value_dims)
```

jbednar:
I'm not sure what the alternatives are here.  Obviously having pprint be useless is not good, and so I vote to fix it.  But is the question whether, when fixing it, for pprint to show options or to only show the semantic information?  If the latter, I think my vote is for using pprint to show options by default, because the representation above clearly separates content from presentation.  It's true that the options could overwhelm the semantic bits, but as long as there is a way to turn them off it seems like I'd want the options information available by default.

philippjfr:
The question is about what API to provide, @jlstevens prefers to remove the ``.pprint`` method on all our objects along with all other methods inherited from Parameterized. Instead he suggested we introduce a new method on the ``opts`` utility. I find it more convenient to call a method on the object. The discussion/disagreement is quite analogous to our discussion about the opts clearing method where both you and I would prefer a method on the object itself while @jlstevens is also suggesting adding a method on the ``opts`` utility.

holoviews/core/pprint.py
```diff
@@ -255,132 +256,148 @@ def options_info(cls, plot_class, ansi=False, pattern=None):
...
-class PrettyPrinter(object):
+class PrettyPrinter(param.Parameterized):
     """
     The PrettyPrinter used to print all HoloView objects via the
-    pprint classmethod.
+    pprint method.
     """
 
+    show_defaults = param.Boolean(default=False, doc="""
+        Whether to show default options as part of the repr.""")
```
jbednar: Docstring should clarify what happens when show_options=False and show_defaults=True.  Does it show nothing (probably what most would expect) or show the full set of defaults without any regard for current settings (which would be useful but may be surprising)?

philippjfr: >Does it show nothing (probably what most would expect)

Right, it shows nothing, will add that once we've decided on the user level API.

philippjfr:
I believe @jlstevens is working on the user level API for this in a separate PR, so this can be merged after review.

jlstevens:
Format looks good to me and the tests are green. I'll merge now and of course we can work on another PR if there are more tweaks we want to make.

PR #3290, opened 2018-12-14
Document usage of Params stream

philippjfr:
This is the final bit of support for using parameterized methods which define dependencies as dynamic inputs. Here is a simple example, which uses a dynamic operation to change the width of a Curve.

```python
class App(param.Parameterized):
    frequency = param.Number(default=3, bounds=(1, 10))

    line_width = param.Number(default=3, bounds=(1, 10))
    
    @param.depends('frequency')
    def load_curve(self):
        return hv.Curve(np.sin(np.linspace(0, 6*self.frequency)))

    @param.depends('line_width')
    def apply_width(self, obj):
        return obj.opts(line_width=self.line_width)
    
    def view(self):
        dmap = hv.DynamicMap(self.load_curve)
        return hv.util.Dynamic(dmap, operation=self.apply_width)
```

Before this a user would have had to do this themselves, i.e. view would have to be either:

```python
def view(self):
   dmap = hv.DynamicMap(self.load_curve)
   return hv.util.Dynamic(dmap, operation=self.apply_width,
                          streams=[ParamMethod(self.apply_width)])
```

or:

```python
def view(self):
   dmap = hv.DynamicMap(self.load_curve)
   return hv.util.Dynamic(dmap, operation=self.apply_width,
                          streams=[Params(self, ['line_width'])])
```

I'll also use this PR to add documentation for the ``Params`` stream.

- [x] Add documentation
- [x] Add unit tests

### DIFF
--- a/examples/user_guide/12-Responding_to_Events.ipynb
+++ b/examples/user_guide/12-Responding_to_Events.ipynb
@@ -15,6 +15,8 @@
    "source": [
     "import numpy as np\n",
     "import holoviews as hv\n",
+    "from holoviews import opts\n",
+    "\n",
     "hv.extension('bokeh')"
    ]
   },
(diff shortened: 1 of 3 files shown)


philippjfr:
I've now added an example using the ``Params`` stream to the Responding to Events user guide so please review that. 

I've also refactored existing examples, I continually remind people that returning static elements from a DynamicMap callback is hugely suboptimal and should be factored out if possible. Reviewing the existing examples I now know why everyone is doing it. Even if it looks slightly more complex I don't think we should be setting a bad example in some of the very first examples.

philippjfr:
I didn't look very closely and just realized this was already implemented, therefore this PR only adds documentation and tests.

jlstevens:
The content looks good but I am increasingly convinced that we should have a param specific user guide. Param creeps into our examples all over the place and I think there should be a proper guide to param in a holoviews context.

philippjfr:
I don't disagree, however I don't think I can write that user guide before release. Could we agree to merge this as is and then move it out into its own guide at a later date?

jlstevens:
This PR includes some unrelated fixes (namely removing magics). For this reason and because the content added is fine I'll merge with the understanding that this content will be moved into a separate user guide about param soon.

Issue #3300, opened 2018-12-15
Opts method does not support literal signature


philippjfr:
We are now recommending people use ``.opts`` using the non-nested format. To make that conversion easier we should accept the same arguments as ``.options``, however currently it does not support the Python literal equivalent of the Options objects, e.g.:

```python
layout.opts({'Image': dict(width=600)})
```

PR #3301, opened 2018-12-15
Add support for literal opts format

philippjfr:
Add support for literal opts format to ``.opts`` method, e.g.:

```
(hv.Curve([1, 2, 3]) * hv.Scatter([1, 2, 3])).opts(
    {'Curve': dict(color='red'), 'Scatter': dict(size=10)})
```

If we are going to be recommending ``.opts`` everywhere we should support the same formats as ``.options`` did. Two users already raised their confusion over this. That doesn't mean we should recommend this now that we have the opts builders, but we should make transitioning as straightforward as possible.

- [x] Implements #3300

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -231,14 +231,18 @@ def deprecated_opts_signature(args, kwargs):
     Returns whether opts.apply_groups should be used (as a bool) and the
     corresponding options.
     """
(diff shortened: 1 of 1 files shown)


jlstevens:
I agree everything is currently meant to be staying backwards compatible. Merging.

PR #3304, opened 2018-12-16
Added support for static plotly export

philippjfr:
This was very easy to support and is necessary for us to create gallery thumbnails easily, so I will likely merge this shortly despite the feature freeze.

### DIFF
--- a/holoviews/plotting/plotly/renderer.py
+++ b/holoviews/plotting/plotly/renderer.py
@@ -1,5 +1,6 @@
 from __future__ import absolute_import, division, unicode_literals
 
+import base64
 import json
 
 import param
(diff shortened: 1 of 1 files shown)


holoviews/plotting/plotly/renderer.py
```diff
@@ -54,7 +55,7 @@ def __call__(self, obj, fmt='html', divuuid=None):
 
         if isinstance(plot, tuple(self.widgets.values())):
             return plot(), mime_types
-        elif fmt == 'html':
+        elif fmt in ('html', 'png', 'svg'):
             return self._figure_data(plot, divuuid=divuuid), mime_types
```
jonmmease: I got an error trying `hv.save` with this, and I *think* the issue is that we need to pass `fmt` into `self._figure_data` here. Otherwise it will return the script as a string

philippjfr: Yep, thanks noticed that too. Fixed in another PR.

PR #3305, opened 2018-12-16
Improved bokeh RGBPlot dtype handling

philippjfr:
This makes the RGB handling in the bokeh backend consistent with matplotlib. There has been some discussion that we should use the value dimension ranges to define the RGB range but a) I would find this very awkward to use and b) it'd be a large backward incompatible change. This PR instead fixes the ``int`` but not ``uint8`` case which was being handled like it was integers. Instead the plot now does what matplotlib does, which is to interpret floats in a 0-1 range and integers in a 0-255 range, anything outside these ranges is clipped.

- [x] Fixes https://github.com/ioam/holoviews/issues/3291

### DIFF
--- a/holoviews/plotting/bokeh/raster.py
+++ b/holoviews/plotting/bokeh/raster.py
@@ -101,13 +101,18 @@ class RGBPlot(ElementPlot):
         img = np.dstack([element.dimension_values(d, flat=False)
                          for d in element.vdims])
         if img.ndim == 3:
-            if img.shape[2] == 3: # alpha channel not included
-                alpha = np.ones(img.shape[:2])
-                if img.dtype.name == 'uint8':
-                    alpha = (alpha*255).astype('uint8')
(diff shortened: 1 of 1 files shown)


jbednar:
@philippjfr , what you describe above sounds reasonable, but from the diffs it doesn't look to me like the code does that.  Where is it cropping out values below zero?  Doing so is tricky because for signed 8-bit integers I don't know whether to crop below zero or to interpret them as unsigned; no idea if that comes up in practice.

philippjfr:
True, happy to add the below 0 clipping.

philippjfr:
Addressed in https://github.com/ioam/holoviews/pull/3310

jbednar:
Looks fine, as long as we're happy that signed 8-bit integers will lose values below zero, instead of interpreting those as 0..255.   I'm pretty sure that's correct, and if that's what mpl does then I'm fine with it. Thanks.

PR #3310, opened 2018-12-17
Improved RGB clipping logic

philippjfr:
As suggested in https://github.com/ioam/holoviews/pull/3305 this fixes the clipping logic in bokeh's RGBPlot. It is now identical to the logic used by matplotlib's imshow internally.


jlstevens:
Being as consistent between plotting extensions as possible is a good thing. Merging.

PR #3312, opened 2018-12-17
Active tool fix

philippjfr:
This is a two-fold fix for an issue with the ``active_tools`` option. First of all it changes the error when a tool listed in `active_tools` is not found to a warning and secondly it drops the styling applied to the stream source on streams which have a element property.


jlstevens:
Looks good. Ready to merge?

philippjfr:
Yes, ready to merge.

PR #3292, opened 2018-12-14
Made the opts method into a callable object

jlstevens:
The idea is to support `obj.opts.clear()` and `obj.opts.show()`.

PR #3313, opened 2018-12-17
Added examples of .opts.clear() and .opts.show()

jbednar:
Adds documentation for changes from #3292.

### DIFF
--- a/examples/getting_started/2-Customization.ipynb
+++ b/examples/getting_started/2-Customization.ipynb
@@ -213,7 +213,25 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


examples/getting_started/2-Customization.ipynb
```diff
@@ -276,7 +294,25 @@
...
+    "If we want to reset back to the original syling, we can call `.opts.clear()`:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "spikes.select(milliseconds=(2000,4000)).opts.clear()"
```
jlstevens: As `select` itself returns a clone, I think this example is a little confusing... `spikes` itself should still be styled after this.

jlstevens:
Other than one comment above this looks good and I am happy to merge.

examples/getting_started/2-Customization.ipynb
jbednar: Right; that's deliberate.

jbednar: Otherwise revisiting a cell with `spikes` will show spikes losing its styling, which would be complicated to explain.

jlstevens:
The `clear` example may cause a little confusion to people who know what is cloned and where (i.e more experienced users) but this guide is for total newcomers. On that basis, I am happy with these changes and I'll merge.

PR #3316, opened 2018-12-19
Selection1D and RangeXY callback support in plotly backend

jonmmease:
### Overview
This PR adds support the `Selection1D` and `RangeXY` streams in the plotly backend. Comments welcome, and this is obviously intended for consideration after the feature freeze 🙂 

I believe that it is working well, though I've only tested it manually on `Points` elements so far.

### Implementation
This implementation approach differs a bit from the Bokeh backend approach because plotly.py doesn't really have a notion of attaching a JavaScript callback to a plotly.js element.  Instead, the trace UID's are used to provide a mapping between traces and comms, and the renderer is responsible for producing all of the necessary callbacks to handle the active streams.

#### `Selection1D`
Here is an example of displaying a red point at the mean of the selected points
```python
import numpy as np
import holoviews as hv
from holoviews.streams import Stream, Selection1D, RangeXY
import param
hv.extension('plotly')

data = np.random.multivariate_normal((0, 0), [[1, 0.1], [0.1, 1]], (100,))
points = hv.Points(data).options(color='gray')

sel = Selection1D(source=points)
rangexy = RangeXY(source=points)

def mean_point(index):
    if index:
        means = data[index, :].mean(axis=0)
        pts1 = hv.Points([means]).options(size=12, color='red')
    else:
        pts1 = hv.Points([])
    
    return pts1

mean_sel = hv.DynamicMap(mean_point, kdims=[], streams=[sel])
points*mean_sel
```
selection1d

#### `RangeXY`
And here's an example of displaying a blue dot at the mean of all of the points in the `RangeXY` viewport.
```python
import numpy as np
import holoviews as hv
from holoviews.streams import Stream, Selection1D, RangeXY
import param
hv.extension('plotly')

data = np.random.multivariate_normal((0, 0), [[1, 0.1], [0.1, 1]], (100,))
points = hv.Points(data).options(color='gray')

sel = Selection1D(source=points)
rangexy = RangeXY(source=points)

def mean_point(x_range, y_range):

    if x_range and y_range:
        inds = np.logical_and(np.logical_and(x_range[0] <= data[:, 0], data[:, 0] <= x_range[1]),
                              np.logical_and(y_range[0] <= data[:, 1], data[:, 1] <= y_range[1]))
        x_mean = np.mean(data[inds, 0])
        y_mean = np.mean(data[inds, 1])
        
        print(x_mean, y_mean)
        pts2 = hv.Points([[x_mean, y_mean]]).options(size=12, color='blue')
    else:
        pts2 = hv.Points([])
    
    return pts2

mean_rangexy = hv.DynamicMap(mean_point, kdims=[], streams=[rangexy])
points*mean_rangexy
```
rangexy

#### Both `Selection1D` and `RangeXY`
Now with both streams active in the same overlay element

```python
import numpy as np
import holoviews as hv
from holoviews.streams import Stream, Selection1D, RangeXY
import param
hv.extension('plotly')

data = np.random.multivariate_normal((0, 0), [[1, 0.1], [0.1, 1]], (100,))
points = hv.Points(data).options(color='gray')

sel = Selection1D(source=points)
rangexy = RangeXY(source=points)

def mean_point(index, x_range, y_range):
    if index:
        means = data[index, :].mean(axis=0)
        pts1 = hv.Points([means]).options(size=12, color='red')
    else:
        pts1 = hv.Points([])

    if x_range and y_range:
        inds = np.logical_and(np.logical_and(x_range[0] <= data[:, 0], data[:, 0] <= x_range[1]),
                              np.logical_and(y_range[0] <= data[:, 1], data[:, 1] <= y_range[1]))
        x_mean = np.mean(data[inds, 0])
        y_mean = np.mean(data[inds, 1])
        
        print(x_mean, y_mean)
        pts2 = hv.Points([[x_mean, y_mean]]).options(size=12, color='blue')
    else:
        pts2 = hv.Points([])
    
    return pts1*pts2

mean_sel = hv.DynamicMap(mean_point, kdims=[], streams=[sel, rangexy])
points*mean_sel
```
selection1d_rangexy

And in a Layout (rather than Overlay) configuration
```python
points+mean_sel
```
selection1d_rangexy_layout

### Testing
There isn't any yet, do you have any suggestions here?  I have't really dug into the bokeh testing code to know if there's an existing approach to testing interactive callbacks.

### DIFF
--- a/holoviews/plotting/plotly/element.py
+++ b/holoviews/plotting/plotly/element.py
@@ -1,9 +1,11 @@
 from __future__ import absolute_import, division, unicode_literals
 
+from itertools import groupby
+
 import numpy as np
 import param
 
-from ...core import util
+from ...core import util, Stream
 from ...core.element import Element
 from ...core.spaces import DynamicMap
 from ...util.transform import dim
@@ -99,6 +101,55 @@ class ElementPlot(PlotlyPlot, GenericElementPlot):
     # Declare which styles cannot be mapped to a non-scalar dimension
     _nonvectorized_styles = []
 
+    def __init__(self, element, plot=None, **params):
+        super(ElementPlot, self).__init__(element, **params)
+        self.callbacks = self._construct_callbacks()
+        self.trace_uid = None
+
+    def _construct_callbacks(self):
+        """
+        Initializes any callbacks for streams which have defined
+        the plotted object as a source.
+        """
+        cb_classes = set()
+        registry = list(Stream.registry.items())
+        callbacks = Stream._callbacks['plotly']
+        for source in self.link_sources:
+            streams = [
+                s for src, streams in registry for s in streams
+                if src is source or (src._plot_id is not None and
+                                     src._plot_id == source._plot_id)]
+            cb_classes |= {(callbacks[type(stream)], stream) for stream in streams
+                           if type(stream) in callbacks and stream.linked
+                           and stream.source is not None}
+        cbs = []
+        sorted_cbs = sorted(cb_classes, key=lambda x: id(x[0]))
+        for cb, group in groupby(sorted_cbs, lambda x: x[0]):
+            cb_streams = [s for _, s in group]
+            cbs.append(cb(self, cb_streams, source))
+        return cbs
+
+    @property
+    def link_sources(self):
+        "Returns potential Link or Stream sources."
+        if isinstance(self, GenericOverlayPlot):
+            zorders = []
+        elif self.batched:
+            zorders = list(
+                range(self.zorder, self.zorder + len(self.hmap.last)))
+        else:
+            zorders = [self.zorder]
+
+        if isinstance(self, GenericOverlayPlot) and not self.batched:
+            sources = []
+        elif isinstance(self.hmap, DynamicMap):
+            sources = [o for i, inputs in self.stream_sources.items()
+                       for o in inputs if i in zorders]
+        else:
+            sources = [self.hmap.last]
+        return sources
+
+
     def initialize_plot(self, ranges=None):
         """
         Initializes a new plot object with the last available frame.
@@ -136,6 +187,15 @@ class ElementPlot(PlotlyPlot, GenericElementPlot):
         for i, d in enumerate(data):
             # Initialize graph
             graph = self.init_graph(d, opts, index=i)
+
+            if i == 0:
+                # Associate element with trace.uid property of the first
+                # plotly trace that is used to render the element. This is
+                # used to associate the element with the trace on the
+                # JavaScript side.
+                self.trace_uid = str(id(element))
+                graph['uid'] = self.trace_uid
+
             graphs.append(graph)
         self.handles['graphs'] = graphs
 
--- a/holoviews/plotting/plotly/renderer.py
+++ b/holoviews/plotting/plotly/renderer.py
@@ -24,9 +24,139 @@ $.each(data.data, function(i, obj) {{
   }});
 }});
 var plotly = window._Plotly || window.Plotly;
+
+// Restore axis range extents set by the user
+for (var key in data.layout) {{
+    if(key.slice(1, 5) === 'axis') {{
+        if (plot.layout[key] && plot.layout[key].range) {{
+            data.layout[key].range = plot.layout[key].range;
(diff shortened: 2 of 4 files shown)


jbednar:
Cool!

jbednar:
Is this now sufficient to support Datashader-backed Plotly-rendered HoloViews plots?

holoviews/plotting/plotly/element.py
```diff
@@ -99,6 +101,55 @@ class ElementPlot(PlotlyPlot, GenericElementPlot):
     # Declare which styles cannot be mapped to a non-scalar dimension
     _nonvectorized_styles = []
 
+    def __init__(self, element, plot=None, **params):
+        super(ElementPlot, self).__init__(element, **params)
+        self.callbacks = self._construct_callbacks()
+        self.trace_uid = None
+
+    def _construct_callbacks(self):
```
philippjfr: We should eventually factor this stuff out from the bokeh backend.

holoviews/plotting/plotly/renderer.py
```diff
@@ -24,11 +24,141 @@
...
+
+    # Based on JS_CALLBACK from pyviz_comms
+    return """
+var plotdiv = document.getElementById('{plot_id}');
+var plotly = window._Plotly || window.Plotly;
+plotly.newPlot(plotdiv, {data}, {layout}, {config}).then(function() {{
+    var elem = document.getElementById("{plot_id}.loading");
+    elem.parentNode.removeChild(elem);
+
+    function unique_events(events) {{
```
philippjfr: Could we just get most of this from ``pyviz_comms.JS_CALLBACK``?

jonmmease: We could if we refactor `pyviz_comms.JS_CALLBACK` a little. The utility functions are all the same, but this requires the initialization of a list of comms (`JS_CALLBACK` is for just 1) and we don't want to logic to construct `data` here.

jonmmease: Would you like me to try refactoring `pyviz_comms` a bit to reuse parts of `JS_CALLBACK` here?  That's fine with me, I just wasn't sure how tightly you want to link versions of `holoviews` and `pyviz_comms` (i.e. can `holoviews` require the newest version of `pyviz_comms` or does there need to be some fallback for older versions?).

jonmmease:
@philippjfr, I'm going to try to pick this one up again.  I think I'll take a look at refactoring `pyviz_comms.JS_CALLBACK` to see how much of the JavaScript code can be taken from there.  Please let me know if you have any other thoughts/concerns on the general approach here.  Thanks!

philippjfr:
We are currently debating how much of HoloViews we will end up rewriting on top of panel. This would basically give us comms for free by building on top of [Plotly bokeh model](https://pyviz-dev.github.io/panel/reference/panes/Plotly.html). We'd just have to extend that model to allow attaching callbacks to the rendered bokeh plot. I'd be very happy to hear/discuss that plan with you though. It is perhaps at least a little bit weird to reimplement plotly on top of bokeh but getting free comms, the ability to deploy the plots on bokeh server, binary transfers of data arrays and a unified rendering and deployment architecture is a pretty strong draw.

jonmmease:
Yeah, I can see how that approach would be appealing.  So HoloViews' own comms would go away then? (or be the same as panel's).  And I guess it might make sense then to use panel for widgets instead of the built-in HoloViews widgets.  I'll take a look at the implementation of the Plotly bokeh model (and yes, that is a bit of an odd sentence at this point :slightly_smiling_face:).  Are there any simple reference examples of models that support callbacks?

philippjfr:
>Are there any simple reference examples of models that support callbacks?

There are, have a look at the ``panel/models/vtk.ts`` and the way it syncs camera state. I also just had a look at the plotly.py's BaseFigureWidget, a lot of code could basically be reused in an equivalent bokeh model. There are definitely issues with that though, apart from being somewhat crazy, so a more minimal approach may be better.

philippjfr:
@jonmmease I assume this will all be superceded by the panel based code?

jonmmease:
Yes, I'll close this out.  The new one should be ready soon!

PR #3317, opened 2018-12-19
Fix missing import to get example to work

flothesof:
Hi,
I just went through the deploying with Bokeh tutorial and found that the above import was missing if you wanted to do standalone execution outside the notebook with `bokeh serve`.
This little fix allowed me to finally run the example.

Hope this helps others.
Best regards,
Florian


philippjfr:
Thank you!

PR #3318, opened 2018-12-19
Output cell magic replacement and minor tweaks

jlstevens:
Here is an example of the output cell magic replacement:

image

Here is an example of the updated options repr:

image

My main worry here is that the return value of `hv.output` has changed from returning the `obj` to returning `None` (needed to avoid double display). This means this change isn't backwards compatible but I don't think anyone was using this other than our automated conversion of magics to python syntax (e.g for bokeh server) behind the scenes.

It is also worth mentioning that anyone using `_` to get the value of the previously executed cell will now change if using this utility.

Finally, I should mention the small hack I added to avoid printing reprs in the IPython terminal. It isn't a big deal if we remove it as in most contexts, `IPython.display` won't be available.

### DIFF
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -509,6 +509,8 @@ class output(param.ParameterizedFunction):
 
             Store.output_settings.output(line=line, cell=obj, cell_runner=display_fn,
                                          help_prompt=help_prompt, **options)
+        elif obj is not None:
+            return obj
         else:
             Store.output_settings.output(line=line, help_prompt=help_prompt, **options)
 
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -505,10 +505,7 @@ class output(param.ParameterizedFunction):
                     from IPython.display import display
                 except:
                     return
-
-                # Small hack to avoid printing in IPython terminal
-                if get_ipython().__class__.__name__ != 'TerminalInteractiveShell': # noqa
-                    display(obj)
+                display(obj)
 
             Store.output_settings.output(line=line, cell=obj, cell_runner=display_fn,
                                          help_prompt=help_prompt, **options)
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -477,13 +477,7 @@ class output(param.ParameterizedFunction):
     are ignored.
     """
 
-    filename_warning = param.Boolean(default=True, doc="""
-       Whether to warn if the output utility is called on an object and
-       a filename is not given (in which case the utility has no
-       effect)""" )
-
     def __call__(self, *args, **options):
-        warn = options.pop('filename_warning', self.filename_warning)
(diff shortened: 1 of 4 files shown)


holoviews/util/__init__.py
```diff
@@ -502,15 +502,22 @@ def __call__(self, *args, **options):
...
-                def save_fn(obj, renderer): renderer.save(obj, options['filename'])
-                Store.output_settings.output(line=line, cell=obj, cell_runner=save_fn,
-                                             help_prompt=help_prompt, **options)
-            elif warn:
-                self.warning("hv.output not supplied a filename to export the "
-                             "given object. This call will have no effect." )
-            return obj
-        elif obj is not None:
-            return obj
+                if util.config.future_deprecations:
```
philippjfr: If it's a future deprecation you can't remove the actual code to save the output now.

holoviews/util/__init__.py
```diff
@@ -502,15 +502,22 @@ def __call__(self, *args, **options):
...
+                    from IPython.display import display
+                except:
+                    return
+
+                # Small hack to avoid printing in IPython terminal
+                if get_ipython().__class__.__name__ != 'TerminalInteractiveShell':
+                    display(obj)
+
+            Store.output_settings.output(line=line, cell=obj, cell_runner=display_fn,
+                                         help_prompt=help_prompt, **options)
```
philippjfr: I don't quite understand what makes this act like a cell magic, I assumed you'd have to apply the options, then display then restore the previous options.

holoviews/util/__init__.py
jlstevens: The `filename` argument still works via the display code path.

holoviews/util/__init__.py
jlstevens: The `cell_runner` function is executed in a context where the renderer has been temporarily modified.

philippjfr: Nevermind, I get it, cell_runner is the bit that makes it work.

holoviews/util/__init__.py
philippjfr: Ah right, it's now calling the actual codepath it would use for the cell magic.

holoviews/util/__init__.py
```diff
@@ -502,15 +502,22 @@ def __call__(self, *args, **options):
...
+
+            def display_fn(obj, renderer):
+                try:
+                    from IPython.display import display
+                except:
+                    return
+
+                # Small hack to avoid printing in IPython terminal
+                if get_ipython().__class__.__name__ != 'TerminalInteractiveShell':
+                    display(obj)
```
jbednar: I don't think we need to go out of our way to make things cleaner in an IPython terminal.  As long as it doesn't do something *too* crazy, I'd skip this check, because users wouldn't be using any of this in such a context normally.

PR #3320, opened 2018-12-19
Ensure custom trees are propagated for non-specified backend

philippjfr:


jlstevens:
I think this is the right fix and the tests are green. Merging.

PR #3325, opened 2018-12-20
Removed charwidth setting and added output.info method

jlstevens:
This screenshot illustrates the `output.show` method which is the replacement for using the `%output` magic without any arguments:

image

The charwidth setting is now removed as it was just confusing and unused cruft.

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -116,7 +116,7 @@ class Opts(object):
     def clear(self, clone=False):
         return self.obj.opts(clone=clone)
 
-    def show(self):
+    def info(self):
         pprinter = PrettyPrinter(show_options=True)
         print(pprinter.pprint(self.obj))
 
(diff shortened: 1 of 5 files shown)


jlstevens:
Maybe `output.show()` is confusing and it could be called `output.settings()`? Though maybe that might be expected to return the settings as a dict instead of just showing them?

jlstevens:
We decided both `opts` and `output` should call this `info`.

holoviews/core/options.py
```diff
@@ -116,7 +116,7 @@ def __call__(self, *args, **kwargs):
     def clear(self, clone=False):
         return self.obj.opts(clone=clone)
 
-    def show(self):
+    def info(self):
```
philippjfr: I'd like to expose whether to show defaults here and add a small docstring. I'd be happy to do so in another PR though.

jlstevens: Agreed. Good idea for a separate PR.

PR #3330, opened 2018-12-22
Fixed bug in looking up valid keywords

philippjfr:
Not every backend registers a plot and options for every element.


jlstevens:
Thanks.

PR #3334, opened 2018-12-23
Include license in wheels

toddrme2178:
The BSD license requires all copies of the software include the license text.


philippjfr:
Thanks!

PR #3341, opened 2018-12-24
Removed duplicated output in Custom Interactivity

jlstevens:
This user guide is currently the only one that has static versions of the expected output. For users running the user guides, this is quite confusing.

Although this PR removes the output, it does serve an important purpose for doc building so that the website is more useful. We haven't found a general solution to this yet but for now I think we should be consistent with what we have. The last commit good for doc building (including this output) is 409900f.

I hope we find a good solution to this issue that we can apply throughout.


jlstevens:
Let us make this the very last thing we merge before packaging.

PR #3362, opened 2019-01-01
Options to opts

ea42gh:
I went through the notebooks and modified `.options()` calls to `.opts()`

That leaves only `./examples/reference/features/bokeh/table_hooks_example.ipynb`
When I tried to run it, I got
```
WARNING:root:TablePlot01553: Plotting hook <function apply_format at 0x7fb51fda0d90> could not be applied:

 'plot'
```
which I did not sort out.

### DIFF
--- a/examples/reference/elements/bokeh/Graph.ipynb
+++ b/examples/reference/elements/bokeh/Graph.ipynb
@@ -93,7 +93,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "simple_graph.options(directed=True, arrowhead_length=0.05)"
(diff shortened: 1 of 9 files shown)


jlstevens:
I think I'll go ahead and merge (tests are failing for an unrelated reason).Then I'll run through these notebooks to make sure they are ok and grep for .options one more time...

jlstevens:
I just checked out all these notebooks and it all looks correct. Thanks again @ea42gh!

There is one issue with the matplotlib histogram notebook on current master (and I think the release). @philippjfr is this why we need a new bug fix release soon and why the tests have been failing?

PR #3367, opened 2019-01-04
Updated exporting and archiving user guide

jlstevens:
Updates this user guide to get rid of the `%%output` and `%%opts` magics and gets rid of some outdated material.

One more major edit I would do is to cut down on the text talking about reproducibility - it is probably better to just get to the point and show what you can do instead of just describing it in text.


jbednar:
Sounds good!  Yes, that section is wordy. Maybe instead of going into that depth, it could link to one of our papers about this topic.  Or maybe some of those points could be made more simply at the end of the document, once the examples have been introduced.

jlstevens:
> Or maybe some of those points could be made more simply at the end of the document, once the examples have been introduced.

I think that would be reasonable: people can easily skip over extra material that appears at the end after all the code concepts are introduced.

jlstevens:
I've move that large block of text to the bottom where people can read it if they are interested. Ready to review/merge.

jbednar:
Sounds good.

jlstevens:
I'm done with this PR now. Happy to merge?

PR #3370, opened 2019-01-07
Fix broken options in Game of Life demo app

flothesof:
Hi,
I've just tried to run the game of life demo app and noticed that some things had been changed due to options not working anymore.
The proposed pull request fixes the demo app in the following way:
- options are separated in plotting and style
- options are applied when creating the initial hv.Image object
- update was modified slightly to just return the img object that was modified in-place instead of creating a new one

Hope this helps
Best regards
Florian

### DIFF
--- a/examples/gallery/apps/bokeh/game_of_life.py
+++ b/examples/gallery/apps/bokeh/game_of_life.py
@@ -64,16 +64,16 @@ def update(pattern, counter, x, y):
         img.data[y:y+r,x:x+c] = pattern[::-1]
     else:
         img.data = step(img.data)
-    return hv.Image(img)
+    return img
 
 title = 'Game of Life - Tap to place pattern, Doubletap to clear'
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks @flothesof the contribution is really appreciated, however all the changes you're making here were deliberate and at least using the latest version the app works for me.

>options are separated in plotting and style

We have deliberately moved away from requiring plot and style options from being separated for simplifying usage.

>update was modified slightly to just return the img object that was modified in-place instead of creating a new one

I believe this was a necessary workaround to ensure that the plot updates.

Could you maybe report which part of the app is currently broken for you?

flothesof:
Hi @philippjfr. Sorry for the pull request. I should have first tried to upgrade my holoviews install because I was using an outdated version. As you said, the script works fine with the current version. Therefore I'm closing the PR.

PR #3371, opened 2019-01-07
Fixed bug updating Raster/Image/RGB alpha

philippjfr:
Fixes issue with alpha not being dynamically updated when changed.


jbednar:
Thanks!  I didn't realize there was a bug; I thought I just wasn't hooking it up correctly when I tried.

PR #3380, opened 2019-01-12
Fix jitter when axes inverted

ahuang11:
When testing https://github.com/ioam/holoviews/issues/3379, I realized if I turned off jitter, it showed but misaligned.
image

Then I tried just plotting scatter inverted with jitter, and noticed that jitter on inverted didn't work.
image

This PR fixes that for both scatter and points.
image

```
import holoviews as hv
from holoviews import dim
hv.extension('bokeh')

from bokeh.sampledata.autompg import autompg as df

title = "MPG by Cylinders and Data Source, Colored by Cylinders"
hv.BoxWhisker(df, 'cyl', 'mpg', label=title).opts(invert_axes=True) * \
hv.Scatter(df, ['cyl', 'mpg']).opts(invert_axes=True, jitter=0.5)

hv.Scatter(df, ['cyl', 'mpg']).opts(jitter=0.5) + \
hv.Scatter(df, ['cyl', 'mpg']).opts(invert_axes=True, jitter=0.5)

hv.Points(df, ['cyl', 'mpg']).opts(jitter=0.5) + \
hv.Points(df, ['cyl', 'mpg']).opts(invert_axes=True, jitter=0.5)
```

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -110,9 +110,12 @@ class PointPlot(LegendPlot, ColorbarPlot):
             style['angle'] = np.deg2rad(style['angle'])
 
         if self.jitter:
-            axrange = 'y_range' if self.invert_axes else 'x_range'
-            mapping['x'] = jitter(dims[xidx], self.jitter,
-                                  range=self.handles[axrange])
+            if self.invert_axes:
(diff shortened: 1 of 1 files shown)


philippjfr:
Looks like our test suite needs to be updated for pandas 0.24 support.

PR #3382, opened 2019-01-12
Add color bar limit opt and redim range if values are equal.

ahuang11:
Addresses https://github.com/ioam/holoviews/issues/3373 and https://github.com/ioam/holoviews/issues/3350 and isn't in the same PR as https://github.com/ioam/holoviews/pull/3381 because I wasn't sure if the extra computations are worth it for the warnings.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1416,8 +1416,9 @@ class ColorbarPlot(ElementPlot):
                 low, high = ranges[dim_name]['combined']
                 # Pad zero-range to avoid breaking colorbar
                 if low == high:
-                    low -= low / 10
-                    high += high / 10
+                    offset = self.default_span / 2
+                    low -= offset
+                    high += offset
             elif isinstance(eldim, dim):
                 low, high = np.nan, np.nan
             else:
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1324,6 +1324,10 @@ class ColorbarPlot(ElementPlot):
         Number of discrete colors to use when colormapping or a set of color
         intervals defining the range of values to map each color to.""")
 
+    clim = param.NumericTuple(default=(np.nan, np.nan), length=2, doc="""
+       User-specified colorbar axis range limits for the plot, as a tuple (low,high).
+       If specified, takes precedence over data and dimension ranges.""")
+
     colorbar = param.Boolean(default=False, doc="""
         Whether to display a colorbar.""")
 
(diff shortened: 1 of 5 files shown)


holoviews/plotting/bokeh/raster.py
```diff
@@ -83,6 +84,22 @@ def get_data(self, element, ranges, style):
             key = 'image' if i == 2 else dimension_sanitizer(vdim.name)
             data[key] = [img]
 
+            # check if there's an actual value (not np.nan)
+            nanmin = np.nanmin(img)
+            nanmax = np.nanmax(img)
+            if nanmin == nanmax and clim_undefined and self.colorbar:
```
philippjfr: Not sure this is the best behavior, I think it should simply add some default padding to the range.

holoviews/plotting/bokeh/element.py
```diff
@@ -1405,7 +1409,10 @@ def _get_colormapper(self, eldim, element, ranges, style, factors=None, colors=N
 
         ncolors = None if factors is None else len(factors)
         if eldim:
-            if dim_name in ranges:
+            # check if there's an actual value (not np.nan)
+            if sum((~np.isnan(v) for v in self.clim)):
```
philippjfr: Should use the `isfinite` utility here and below I think, that'll handle inf and None as well.

holoviews/plotting/bokeh/raster.py
philippjfr: Also don't think this is the correct place for this logic, it should be moved to the appropriate place on ColorbarPlot.

jbednar: Right; I think it's better to have an empty plot than to change the behavior or have a warning or error, so that it will behave well in streaming or batch situations where there may legitimately be no data in that particular instance.

ahuang11:
image

ahuang11:
image

holoviews/plotting/bokeh/element.py
```diff
@@ -1405,8 +1409,15 @@ def _get_colormapper(self, eldim, element, ranges, style, factors=None, colors=N
...
         if eldim:
-            if dim_name in ranges:
+            # check if there's an actual value (not np.nan)
+            if util.isfinite(self.clim).all():
+                low, high = self.clim
+            elif dim_name in ranges:
                 low, high = ranges[dim_name]['combined']
+                # Pad zero-range to avoid breaking colorbar
+                if low == high:
+                    low -= low / 10
```
philippjfr: This logic won't work if low and high are zero, maybe use the default_span instead:

```python
offset = self.default_span/2
```

philippjfr:
@ahuang11 I've made one more comment. If we want to get the tests passing you'll likely also have to rebase once https://github.com/ioam/holoviews/pull/3392 is merged. Some change in NumPy 1.16 is playing havoc with our python 2 tests.

philippjfr:
Thanks it's looking good, I'll handle the rebasing once the other PR is merged.

PR #3385, opened 2019-01-12
Add support for multi-level categorical axes for all Charts 

philippjfr:
This PR is meant to overhaul handling of categorical axes for all Chart elements. This is a backward incompatible change and the plan is to be able to toggle the new behavior on in the next release and eventually defaulting to the new behavior in the next release.

This PR implements the suggestion in https://github.com/ioam/holoviews/issues/1668, which allows all chart elements to support multi-level categorical axes. This means that a Chart element with more than one key dimension will always be interpreted as a multi-level categorical axis, which will make it possible to overlay existing charts on top of elements which already support this mechanism for multi-level categorical axes, e.g. as a straightforward example let's take a multi-level Bars plot:

```python
data = [
    ('A', 1, 2, 0.2), ('A', 2, 3, 0.3),
    ('B', 1, 4, 0.7), ('B', 2, 1, 0.5)]
hv.Bars(data, kdims=['a', 'b'], vdims=['c'])
```

bokeh_plot

Currently if we want to overlay points, errorbars, curves, etc. on top of this plot there would be no way of doing so. Once this PR is merged it will be possible to do this:

```python
data = [
    ('A', 1, 2, 0.2), ('A', 2, 3, 0.3),
    ('B', 1, 4, 0.7), ('B', 2, 1, 0.5)]
hv.Bars(data, kdims=['a', 'b'], vdims=['c']) * hv.ErrorBars(data, kdims=['a', 'b'], vdims=['c', 'error']) 
```

bokeh_plot

### DIFF
--- a/holoviews/element/chart.py
+++ b/holoviews/element/chart.py
@@ -240,7 +240,7 @@ class Spikes(Chart):
 
     group = param.String(default='Spikes', constant=True)
 
-    kdims = param.List(default=[Dimension('x')], bounds=(1, 1))
+    kdims = param.List(default=[Dimension('x')], bounds=(1, 3))
 
     vdims = param.List(default=[])
 
--- a/holoviews/plotting/bokeh/__init__.py
+++ b/holoviews/plotting/bokeh/__init__.py
@@ -31,9 +31,10 @@ from .annotation import (TextPlot, LineAnnotationPlot, SplinePlot,
 from ..plot import PlotSelector
 from .callbacks import Callback # noqa (API import)
 from .element import OverlayPlot, ElementPlot
-from .chart import (PointPlot, CurvePlot, SpreadPlot, ErrorPlot, HistogramPlot,
+from .chart import (CurvePlot, SpreadPlot, ErrorPlot, HistogramPlot,
                     SideHistogramPlot, BarPlot, SpikesPlot, SideSpikesPlot,
-                    AreaPlot, VectorFieldPlot)
+                    AreaPlot, ScatterPlot)
+from .geom import PointPlot, VectorFieldPlot
 from .graphs import GraphPlot, NodePlot, TriMeshPlot, ChordPlot
 from .heatmap import HeatMapPlot, RadialHeatMapPlot
 from .hex_tiles import HexTilesPlot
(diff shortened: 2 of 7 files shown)


philippjfr:
Putting this on hold for a while, I had large parts of this in my stash already and just wanted to open a PR for it since it came up in several discussions recently.

ahuang11:
Awesome! Curious if there will be support for extra hover columns in Points/Path?

eburling:
+1 for multi-level category support for `Scatter`. I'm hoping to overlay points on a `Violin` object.

holoviews/element/chart.py
```diff
@@ -240,7 +240,7 @@ class Spikes(Chart):
 
     group = param.String(default='Spikes', constant=True)
 
-    kdims = param.List(default=[Dimension('x')], bounds=(1, 1))
+    kdims = param.List(default=[Dimension('x')], bounds=(1, 3))
```
poplarShift: Does this mean categorical axes should generally only support at most 2 nested levels?

HWiese1980:
Any info on this PR? It's been almost a year since the last update...

rchurt:
Agree with HWiese1980, I need to use this now :)

HWiese1980:
Two years now. I don't see this happening anytime soon... Sad!

jbednar:
This change is part of HoloViews 2.0, which involves a lot of moving parts other than this one. Not sure if you guys are meaning to be offering some help in working towards that? :-) If so, let us know and we can help identify some tasks that will help get us there sooner...

HWiese1980:
Thank you for the feedback, @jbednar! If you have small tasks that don't require in-depth knowledge about internals of HoloViews, feel free to identify them and share them with us. I'm sure there will be people who're willing to help. I totally am, as far as I find the time next to my primary tasks.

pedropalb:
I recently tried to overlay a `Labels` element over a `Bars` with multi-level categories in `holoviews` v1.20 with `Bokeh` backed and got the error message "None of the available storage backends were able to support the supplied data format.".

Searching for solutions, I came across this PR. 
I suppose that this functionality is not yet supported by `Labels`.

@philippjfr and @jbednar, I would be very happy to contribute with the implementation of this functionality. I would just need an initial guidance about the structures and code parts I should look for and understand first.

In the meantime, is there any workaround I can use by now? Maybe switching to a backed other than `Bokeh`, or directly manipulating the underlying plot structure. I started to try the latter approach, but could not figure out how to get the actual positions of the bars from the `Bars` element to manually set them in the `Layers` element.

PR #3391, opened 2019-01-14
Handled (deprecated) empty __call__ case

jlstevens:
Although ``__call__`` on dimensioned objects is deprecated, the old behavior should be preserved for now.


jbednar:
I can confirm that this fixes the error condition (for bad code) reported in https://github.com/pyviz/pyviz/issues/145

PR #3392, opened 2019-01-15
Pin numpy to <1.16.0

philippjfr:
Travis currently failing due to a NumPy version bump, pinning it should fix it for now. Hopefully xarray will push out a compatible release soon.


philippjfr:
This has been fixed upstream in dask, once dask 1.1.0 is released on Friday we can unpin NumPy again.

PR #3402, opened 2019-01-15
Fix TypeError in string formatting expression

garibarba:
`TypeError: not enough arguments for format string` is raised when evaluating the expression.

I changed it the minimum so that it does not crash, although the error raised is not fully informative, since `not self.allowable(name)` excludes more than `self.disallowed`.


philippjfr:
Thank you!

PR #3405, opened 2019-01-16
Fix violin 'box'

Safrone:
Fix 'box' `inner` option for violin plots with singular data and add a test.


Safrone:
fixes https://github.com/ioam/holoviews/issues/3358 for 'box' option

philippjfr:
Thanks.

PR #3409, opened 2019-01-17
Fixes for Polygons plotting with holes

philippjfr:
The PolygonPlot implementation switched between the MultiPolygons and Patches glyphs depending on whether the polygons have holes or not. The problem with that is that in a dynamic plot or HoloMap the returned data format may therefore change but the glyph is only initialized once, so a MultiPolygons glyph could end up with the data format for the Patches glyph and vice versa. This PR ensures that Polygons are always plotted using the MultiPolygons glyph **unless** a PolyDraw or PolyEdit stream is attached to the plot, which do not yet support holes. This ensures backwards compatibility while also fixing the bug.

- [x] Adds unit tests


holoviews/core/data/__init__.py
```diff
@@ -284,7 +284,7 @@ def range(self, dim, data_range=True, dimension_range=True):
             return (None, None)
         elif all(util.isfinite(v) for v in dim.range) and dimension_range:
             return dim.range
-        elif dim in self.dimensions() and data_range and len(self):
+        elif dim in self.dimensions() and data_range and bool(self):
```
philippjfr: This is mainly an optimization, computing the length on a Polygons/Path can be very expensive, ``bool(self)`` is equivalent but a much quicker way to check whether the element is empty.

PR #3411, opened 2019-01-17
Updated pinning of bokeh and IPython

jlstevens:
Pins bokeh as 1.0.3 had an issue with broken hover (seems to be fixed in 1.0.4) and IPython due to the broken tab completion described in https://github.com/ipython/ipython/issues/11561.


jlstevens:
In the end there isn't a need to update bokeh. As the test data update is now unnecessary, I'll close this  PR and open a new one only pinning IPython.

PR #3419, opened 2019-01-18
Added entry point in conda.recipe for windows

jlstevens:
This should be the last PR for 1.11.1.


philippjfr:
Thanks for remembering to do this.

PR #3422, opened 2019-01-21
Added hook to clean up custom option trees

philippjfr:
This PR adds a dictionary of weakrefs to the Store which tracks all HoloViews objects which have had custom options applied to them. Whenever the ``id`` of an object changes or an object is garbage collected a callback then checks whether there are any remaining objects referencing the custom tree and deletes both the weakref and the custom options tree if it is no longer referenced.

- [x] Fixes https://github.com/ioam/holoviews/issues/3421
- [x] Add unit tests

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -658,17 +658,17 @@ class LabelledData(param.Parameterized):
         return self._id
 
     @id.setter
-    def id(self, id):
+    def id(self, opts_id):
         """Handles tracking and cleanup of custom ids."""
         old_id = self._id
-        self._id = id
+        self._id = opts_id
         if old_id is not None:
             cleanup_custom_options(old_id)
-        if id is not None:
-            if id not in Store._weakrefs:
-                Store._weakrefs[id] = []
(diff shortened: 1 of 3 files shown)


philippjfr:
There's something quite odd going on, a bunch of custom trees are created for ids which are never referenced by any object.

philippjfr:
Found the issue, when remapping custom trees it is possible that custom trees are created but never actually applied to anything. In my latest commit these trees are immediately cleaned up.

jlstevens:
Probably worth trying to get this PR in soon. My main comment when I looked at this PR earlier was that a try/except block could help ensure everything works even if the garbage collection step fails (maybe with a warning).

philippjfr:
>My main comment when I looked at this PR earlier was that a try/except block could help ensure everything works even if the garbage collection step fails (maybe with a warning).

That's already done, there's one weird issue I haven't been able to work out, which is why I didn't try to get this into 1.11.2.

jlstevens:
Ok, the changes should now be safe and I would like to merge sooner rather than later.

PR #3429, opened 2019-01-25
Allow defining hooks on backend load

philippjfr:
This PR allows downstream libraries such as geoviews to register a hook which is called when the extension is loaded. This will make it possible to defer loading of plotting modules until they are explicitly loaded (and therefore avoid issues with the matplotlib backend on some platforms).


jlstevens:
Nice and simple. Merging.

PR #3432, opened 2019-01-25
Improved attribute handling in Graph.from_networkx

philippjfr:
This PR improves Graph.from_networkx in a number of ways. First of all it makes sure that non-scalar attributes and indexes (e.g. lists, tuples or dictionaries) do not break the Graph. Secondly this PR will allow passing in a dictionary of node positions which is the canonical way to specify node positions in networkx. 

- [x] Handle non-scalar attributes
- [x] Handle dictionary node positions


philippjfr:
Merging to make sure the holoviews release is ready in time.

PR #3435, opened 2019-01-26
Fixed Overlay constructor bug

philippjfr:
The Overlay constructor was not rewritten when it was changed to inherit from ViewableTree. This meant that the constructor was called twice and the ``id`` used to associate options with an element was always set to None.

- [x] Fixes https://github.com/ioam/holoviews/issues/3434
- [ ] Adds unit tests


jlstevens:
Glad this was caught. Happy to merge when you tell me this PR is ready.

philippjfr:
Merging since this is blocking another PR.

PR #3440, opened 2019-01-29
Added a get method to Opts to get an Options object

jlstevens:
Prompted by a stackoverflow question I decide to go ahead and implement the `get` method on `Opts` we have previously discussed. It doesn't do anything fancy for container types but does let you specify the option type if you like e.g with `points.opts.get('style')`.

I haven't put too much thought into the name: `get` was the first thing that came to mind.

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -67,9 +67,15 @@ class Opts(object):
 
 
     def get(self, group=None, backend=None):
-        """
-        Returns an Options object, flattening across option groups if
-        group is None. Uses the current backend if backend is None.
+        """Returns the corresponding Options object.
+
+        Args:
+            group: The options group. Flattens across groups if None.
+            backend: Current backend if None otherwise chosen backend.
+
+        Returns:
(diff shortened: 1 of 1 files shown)


philippjfr:
Looks good, I'd also like to see an optional backend kwarg and proper Google style docstring.

jlstevens:
>  I'd also like to see an optional backend kwarg...

Sure.

> ... and proper Google style docstring.

I did consider it but saw all the other docstrings in that class weren't google style so I opted to be consistent. I guess I can update all the methods of that class to be google style - maybe the trick to update all the docstrings across HoloViews is to work class by class instead of file by file (or method by method).

philippjfr:
Looks good to me, I'll merge once tests pass.

jlstevens:
> I did consider it but saw all the other docstrings in that class weren't google style so I opted to be consistent. 

Little confused as the other docstrings were google style...I must have been looking at the wrong class when I said that!

philippjfr:
PR builds passed, merging.

PR #3444, opened 2019-01-30
Attempt updating to new build infrastructure

philippjfr:
This PR updates the test and build infrastructure in holoviews to use the pyct(dev) tools used by other packages. It makes a number of modifications to the travis tests:

* Splits flake/unit tests and regression/notebook tests into separate test builds. This simplifies the envs used for the different kinds of tests slightly, isolates the horrible regression testing code, and will make it easier to remove or replace it at a later stage.
* Enables build caching
* Adds new optional ``extras_require`` entries to specify the different test and build dependencies, the old 'recommended', 'examples' and 'extras' are unmodified.
* Switches appveyor to pyctdev based builds

To Do:

- [ ] Update ``holoviews`` command to use pyct, currently still uses ``holoviews --install-examples`` API.


ceball:
Does hv have any dependencies which can't be installed via pip (i.e. that require another package manager, such as conda)?

philippjfr:
Not any direct dependencies no. There are some optional dependencies (e.g. ffmpeg) that are much easier to install with conda, and currently listed that under ``extras_requires['recommended']``. If a pip user tried to install ``holoviews[recommended]`` would that now fail because it can't find ffmpeg?

philippjfr:
@jlstevens @jbednar Any thoughts or objections to what I've done here. The test builds are now working as I'd expect and I'm about to test tagging a dev release.

philippjfr:
I'm also tempted to add another build stage which is run only on tags that runs nbsmoke against at least a subset of user guides and gallery examples.

philippjfr:
I'll be experimenting with a few cron jobs to run some additional builds. The first one is adding a build with just the basic dependencies to avoid regressions https://github.com/ioam/holoviews/pull/3453 is merged.

philippjfr:
I've added an extra build stage which will only be run during daily cron jobs. Currently that just runs the unit tests without installing all dependencies.

ceball:
I think some of this can be simplified for holoviews (maybe you started from geoviews? has awkward dependencies to handle), but I'm _amazed_ you managed to do it! I'll try to propose simplifications after you've merged.

> someone has registered holoviews on test.pypi.org and there appears to be no way to figure out who owns it because the packages were deleted and it's therefore not indexed on https://test.pypi.org. Does anyone know who owns holoviews on test.pypi.org?

I was previously confused about this (I was sure it was one of you and you'd forgotten...if not, then I have no idea).

However, I realize now my account is the sole owner of several projects on testpypi; I'll transfer them to the pyviz account.

acc

philippjfr:
Thanks for doing that, presumably we should now update all the credentials in the Travis env vars? I've also finally figured out who owned test.pypi.org and put that under the pyviz account as well. So this PR should now be good to go.

jlstevens:
@philippjfr Thanks for doing all this! All the tests are currently passing and the best way to make sure is everything is the way we want it is to merge and to try it all out. Merging.

PR #3450, opened 2019-02-01
Compatibility with bokeh layout refactor

philippjfr:
This PR achieves compatibility with the layout refactor which was merged into bokeh dev and will land in bokeh 1.1.0. It will give us control over the actual plot frame dimensions (excluding the axes, colorbars, legends) and provide control over the plot aspect.

- [x] Adds support for ``aspect`` and ``data_aspect`` in bokeh (https://github.com/pyviz/holoviews/issues/335)
- [x] Fixes support for sizing modes in bokeh
- [x] Adds lower level control over reactive layouts using min/max_width/height and width/height_policy options
- [x] Ensure data aspect is maintained when updating plot
- [x] Ensure datetime and categorical axes warn if data_aspect set
- [x] Test on bokeh server
- [x] Check no events bounce on datashade
- [x] Update documentation
- [x] Update tests
- [x] Add tests

Fixes https://github.com/pyviz/holoviews/issues/3026
Fixes https://github.com/pyviz/holoviews/issues/2992

### DIFF
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -48,8 +48,12 @@ class BokehPlot(DimensionedPlot):
 
     sizing_mode = param.ObjectSelector(default='fixed',
         objects=['fixed', 'stretch_both', 'scale_width', 'scale_height',
-                 'scale_both'], doc="""
+                 'scale_both', 'auto'], doc="""
         How the item being displayed should size itself.
+
+        "auto" mode will use 'fixed' sizing when no aspect is defined
+        otherwise it will use 'scale_width' or 'scale_height' depending
+        on whether the aspect exceeds.
 
         "stretch_both" plots will resize to occupy all available
         space, even if this changes the aspect ratio of the element.
@@ -677,7 +681,6 @@ class GridPlot(CompositePlot, GenericCompositePlot):
     def _make_axes(self, plot):
         width, height = self.renderer.get_size(plot)
         x_axis, y_axis = None, None
-        kwargs = dict(sizing_mode=self.sizing_mode)
         keys = self.layout.keys(full_grid=True)
         if self.xaxis:
             flip = self.shared_xaxis
@@ -704,15 +707,15 @@ class GridPlot(CompositePlot, GenericCompositePlot):
                 r1, r2 = r2, r1
             if self.shared_yaxis:
                 r1, r2 = r1[::-1], r2[::-1]
-            plot = gridplot([r1, r2], **kwargs)
+            plot = gridplot([r1, r2])
         elif y_axis:
             models = [y_axis, plot]
             if self.shared_yaxis: models = models[::-1]
-            plot = Row(*models, **kwargs)
+            plot = Row(*models)
         elif x_axis:
             models = [plot, x_axis]
             if self.shared_xaxis: models = models[::-1]
-            plot = Column(*models, **kwargs)
+            plot = Column(*models)
         return plot
 
 
@@ -998,7 +1001,7 @@ class LayoutPlot(CompositePlot, GenericLayoutPlot):
         title = self._get_title_div(self.keys[-1])
         if title:
             self.handles['title'] = title
-            layout_plot = Column(title, layout_plot, **kwargs)
+            layout_plot = Column(title, layout_plot)
 
         self.handles['plot'] = layout_plot
         self.handles['plots'] = plots
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -40,37 +40,53 @@ class BokehPlot(DimensionedPlot):
     plotting interface for Bokeh based plots.
     """
 
-    width = param.Integer(default=300, doc="""
-        Width of the plot in pixels""")
+    css_classes = param.List(default=[], doc="""
+        CSS classes to apply to the layout.""")
 
-    height = param.Integer(default=300, doc="""
-        Height of the plot in pixels""")
+    sizing_mode = param.ObjectSelector(default=None, objects=[
+        'fixed', 'stretch_width', 'stretch_height', 'stretch_both',
+        'scale_width', 'scale_height', 'scale_both', None], doc="""
 
-    sizing_mode = param.ObjectSelector(default='fixed',
-        objects=['fixed', 'stretch_both', 'scale_width', 'scale_height',
-                 'scale_both', 'auto'], doc="""
-        How the item being displayed should size itself.
+        How the component should size itself.
 
-        "auto" mode will use 'fixed' sizing when no aspect is defined
-        otherwise it will use 'scale_width' or 'scale_height' depending
-        on whether the aspect exceeds.
+        This is a high-level setting for maintaining width and height
+        of the component. To gain more fine grained control over
+        sizing, use ``width_policy``, ``height_policy`` and
+        ``aspect_ratio`` instead (those take precedence over
+        ``sizing_mode``).
 
-        "stretch_both" plots will resize to occupy all available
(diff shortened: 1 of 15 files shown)


philippjfr:
The main thing to settle here is how we should handle the switch from widths/heights which include axes/colorbars/legends to frame widths which apply only to the plot itself. 
The original behavior makes it very easy to get multiple plots to align nicely so it still has uses and we don't want to mess with everyones existing layouts without at least some warning so I think there are two options I can think of:

1) Add an option to switch between frame size and overall size (and for 2.0 we could consider switching the default to the frame width/height)
2) Add separate frame_width/frame_height options which override the standard width/height

philippjfr:
I'm going to make an attempt at explaining the ``width``, ``height``, ``aspect`` and ``data_aspect`` behavior and explain how they interact with the plot **dimensions** and **ranges** particularly as it relates to responsive behavior and updating plots.

## width/height/frame_height/frame_width

width/height: Specifies the dimensions of the plot as a whole including legends/axes/colorbars/titles
frame_width/frame_height: Specifies the dimensions of just the plot frame

These options take absolute precedence over **everything** if a fixed height or width is specified by a user no other option including ``aspect``, ``data_aspect`` or ``responsive`` will have an effect (and will warn). A ``frame_width``/``frame_height`` will take precedence over a regular width/height.

### Aspect

The ``aspect`` specifies the ratio between the width and height dimensions of the plot. If specified this options takes absolute precedence over the **dimensions** of the plot but has no effect on the plot **ranges**. The only exception is ``aspect='equal'`` which is equivalent to ``data_aspect=1``. Aspect will only take effect if either or both and height and width are undefined.

### Data aspect

The ``data_aspect`` specifies the scaling between the x- and y-axis ranges. If specified this option will scale both the plot **ranges** and **dimensions** unless an ``aspect`` value overrides the plot dimensions.

### Responsive

In ``responsive`` mode the plot size rescales to the size of its container depending on the other options that were specified. ``responsive`` maps to bokeh's sizing mode as follows:

* ``stretch_both``: Stretches plot to maximum available width and height if no width, height or aspect are provided.
* ``stretch_width``: Stretches plot to maximum available width if no width and no aspect are provided
* ``stretch_height``: Stretches plot to maximum available height if no height and no aspect are provided
* ``scale_both``: Scales both width and height of plot while maximizing the width of the plot, if aspect is defined but no fixed width/height are. 
* ``scale_width``: Scales both width and height of plot while maximizing the width of the plot, if aspect is defined but no fixed width/height are and if ``responsive='width'`` (rarely used).
* ``scale_height``: Scales both width and height of plot while maximizing the width of the plot, if aspect is defined but no fixed width/height are and if ``responsive='width'`` (rarely used). 

There is one major caveat when combining a data aspect and responsive mode, in this case we cannot fully guarantee that a specified data aspect is maintained. This is because in responsive mode we don't have control over the plot frame dimensions just the overall plot aspect. We may be able to alleviate this by writing some JS code that maintains equal aspect in the plot ranges or by dynamically syncing the plot frame extents.

## Updating Plots

The other major issue here is maintaining a fixed aspect while a plot is updating. We do not want a plot resizing when you update it (in fact I could not even get that working), so the only way to achieve a fixed ``data_aspect`` is adjusting the ranges when new data comes in that has a different aspect. This works well but suffers from the same issue described above when responsive is enabled.

jbednar:
I know we discussed all of the above comments at some point, but my brain starts to hurt when trying to really think deeply about them now. That said, as far as I can see that all sounds like what we had discussed and seems like the best approach in each case.

jlstevens:
I agree with Jim that this sounds just like we discussed. From the description, it sounds like everything behaves the way I would hope, except maybe for the caveat about `data_aspect`.

ea42gh:
Feature request: plot_size option for GridSpace should support tuples

holoviews/plotting/bokeh/util.py
```diff
@@ -148,27 +156,166 @@ def compute_plot_size(plot):
...
+        return width, height
     elif isinstance(plot, (Plot, DataTable, Spacer)):
-        width, height = plot.width, plot.height
-    return width, height
+        return plot.width, plot.height
+    else:
+        return 0, 0
+
+
+def compute_layout_properties(
```
jlstevens: That is one long function indeed! Glad it is now split out from the rest.

holoviews/plotting/bokeh/plot.py
```diff
@@ -40,33 +38,8 @@ class BokehPlot(DimensionedPlot):
...
-        width available, while maintaining the original aspect ratio.
-
-        "scale_height" elements will responsively resize to fit to the
-        height available, while maintaining the original aspect ratio.
-
-        "scale_both" elements will responsively resize to for both the
-        width and height available, while maintaining the original
-        aspect ratio.""")
+    css_classes = param.List(default=[], doc="""
+        CSS classes to apply to the layout.""")
```
jlstevens: As far as I can tell there are no documented examples of how to use this plot option... (but maybe there is a diff I haven't expanded in this PR somewhere)

holoviews/plotting/bokeh/plot.py
```diff
@@ -463,6 +445,51 @@ class CompositePlot(BokehPlot):
     to such a plot.
     """
 
+    sizing_mode = param.ObjectSelector(default=None, objects=[
+        'fixed', 'stretch_width', 'stretch_height', 'stretch_both',
+        'scale_width', 'scale_height', 'scale_both', None], doc="""
```
jlstevens: I guess it is worth exposing the bokeh semantics even if users should rarely need to set this explicitly (in which case I would expect it to override everything else)

jlstevens: As this is CompositePlot I see this is only for containers...

holoviews/plotting/bokeh/util.py
jlstevens: I suspect some simplification could be done here but that doesn't have to hold up this PR...

holoviews/plotting/bokeh/util.py
```diff
@@ -148,27 +156,166 @@ def compute_plot_size(plot):
...
+                       '\'square\'.')
+
+    return ({'aspect_ratio': aspect_ratio,
+             'aspect_scale': aspect_scale,
+             'match_aspect': match_aspect,
+             'sizing_mode' : sizing_mode},
+            {'frame_width' : frame_width,
+             'frame_height': frame_height,
+             'plot_height' : height,
+             'plot_width'  : width})
```
jlstevens: It isn't clear to me from the docstring why two dictionaries of settings are being returned here...

jlstevens:
@philippjfr Looks great! Only had one minor request (docstring improvement) and other than that, I just want a way to remember to remove `-c bokeh/label/dev` once the new bokeh version is released.

Happy to merge once the small change is pushed (no need to wait on the tests this time).

holoviews/plotting/bokeh/plot.py
philippjfr: I'll add a small example to the bokeh guide, it's mostly useful for server apps I would think but adding CSS in a notebook is straightforward enough.

holoviews/plotting/bokeh/util.py
philippjfr: I'll add a docstring.

holoviews/plotting/bokeh/plot.py
philippjfr: Tbh, I'm probably going to remove it, I haven't been able to come up with a useful example. It makes sense for Row and Column containers once we add them but at the plot level messing with css just doesn't seem useful.

jlstevens:
Great thanks! Merging.

PR #3453, opened 2019-02-03
Ensure all tests pass without pandas, dask, xarray and scipy

philippjfr:
As the title says, ensuring the tests match mostly means ensuring that datetimes are handled more consistently. This also uncovered a number of pretty nasty bugs.

- Dropping a kdim on a non-scalar gridded dataset was broken without pandas
- The DictInterface always added an index to an 1D array (instead of respecting auto_indexable_1d)
- Datetime and Date times were not handled as scalars
- Without pandas but with dask the DaskInterface would still accept input but raise an exception when dask.dataframe is imported (fixes https://github.com/ioam/holoviews/issues/3430)
- Image.sample did not work at all without pandas
- Plot padding was broken for datetimes where the axis range was close to the date resolution

Otherwise I updated all tests to appropriately skip if a dependency was not available.

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -428,7 +428,9 @@ class Dataset(Element):
         data = self.interface.reindex(self, key_dims, val_dims)
         datatype = self.datatype
         if gridded and dropped:
-            datatype = [dt for dt in datatype if not self.interface.interfaces[dt].gridded]
+            interfaces = self.interface.interfaces
+            datatype = [dt for dt in datatype if not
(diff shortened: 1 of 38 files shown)


jlstevens:
Thanks for doing this. I'm not surprised that bugs were uncovered having tried to run the tests without all the dependencies in the past.

How do we make sure tests keep working properly using only core dependencies given we want to test as much as possible on travis (and therefore install as much as possible)?

philippjfr:
I'll add a cron job for this in the other PR.

Issue #3456, opened 2019-02-03
cellmagic output and options parsing problem


ea42gh:
```
#%%output backend='matplotlib'
def rectangle(x=0, y=0, width=.05, height=.05):
    return np.array([(x,y), (x+width, y), (x+width, y+height), (x, y+height)])

hv.Polygons(rectangle()).opts([ hv.opts.Polygons(color='indianred',line_alpha=.5, padding=2, backend='bokeh'),
                                hv.opts.Polygons(color='lightblue', padding=2, backend='matplotlib')])
```
works as expected.
Uncommenting the output backend cell magic results in
`WARNING:root:main: Option 'line_alpha' for Polygons type not valid for selected backend ('matplotlib'). Option only applies to following backends: ['bokeh']` in hv 1.11.2.post3+ged1cc85a3

philippjfr:
You also have to supply the backend to the opts method:

```python
hv.Polygons(rectangle()).opts(hv.opts.Polygons(color='indianred',line_alpha=.5, padding=2, backend='bokeh'), backend='bokeh').opts(hv.opts.Polygons(color='lightblue', padding=2, backend='matplotlib'), backend='matplotlib')
```

I'd be in favor of making that be less redundant but this is not a bug.

philippjfr:
I have an issue labelling something that is working as designed a bug even if it is poor API design.  

>I think hv.opts should store the backend in use when it was created (whether as an explicit backend= argument or simply because that's what the default backend was) so that the .opts() method can read that declaration and use it when determining how to apply the options.

I disagree here, my original opinion which I still stand by is that ``hv.opts`` should not be responsible for validation at all, which would get rid of the need for the backend kwarg. IMO the validation and the kwarg should only be introduced when the option is actually **applied**. I accept that I was overruled on this initially so I can live with that proposal, but I maintain that doing the validation twice was the wrong decision. The only other alternative I can suggest is that the utility simply validates against all loaded backends but I think in that case I actually prefer having Options objects store the specified backend. If we decide to go that route I'd expect this to work:

```python
object.opts(opts.Polygons(facecolor='red', backend='matplotlib'), opts.Polygons(fill_color='red', backend='bokeh'))
```

i.e. opts calls should allow specifying options for multiple backends at the same time.

ea42gh:
| I have an issue labelling something that is working as designed a bug even if it is poor API design.
agreed.

From my user perspective, I would want semantics that specify
[...]

PR #3463, opened 2019-02-05
More flexible cross-backend option application

jlstevens:
This PR is a first cut at fixing #3456 by tracking the backend specified:

image

Still work in progress: will need a lot of testing including new unit tests.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -1409,11 +1409,11 @@ class Dimensioned(LabelledData):
 
         from ..util import opts
         if options is None:
-            expanded_backends = [({}, backend)]
+            expanded_backends = [(backend, {})]
         elif isinstance(options, list): # List of Options objects
             expanded_backends = opts._expand_by_backend(options, backend)
         else:
-            expanded_backends = [(opts._expand_options(options, backend), backend)]
+            expanded_backends = [(backend, opts._expand_options(options, backend))]
 
         obj = self
         for backend, expanded  in expanded_backends:
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -1416,7 +1416,7 @@ class Dimensioned(LabelledData):
             expanded_backends = [(backend, opts._expand_options(options, backend))]
 
         obj = self
-        for backend, expanded  in expanded_backends:
+        for backend, expanded in expanded_backends:
             obj = obj.opts(expanded, backend=backend, clone=clone)
         return obj
 
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -1409,10 +1409,16 @@ class Dimensioned(LabelledData):
 
         from ..util import opts
         if options is None:
-            expanded = {}
+            expanded_backends = [({}, backend)]
+        elif isinstance(options, list): # List of Options objects
+            expanded_backends = opts._expand_by_backend(options, backend)
         else:
-            expanded = opts._expand_options(options, backend)
-        return self.opts(expanded, backend=backend, clone=clone)
+            expanded_backends = [(opts._expand_options(options, backend), backend)]
+
+        obj = self
+        for backend, expanded  in expanded_backends:
+            obj = obj.opts(expanded, backend=backend, clone=clone)
+        return obj
 
 
     def _repr_mimebundle_(self, include=None, exclude=None):
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -1932,7 +1932,7 @@ class StoreOptions(object):
         spec, compositor_applied = cls.expand_compositor_keys(options)
         custom_trees, id_mapping = cls.create_custom_trees(obj, spec)
         cls.update_backends(id_mapping, custom_trees, backend=backend)
-        
+
         # Propagate ids to the objects
         not_used = []
         for (match_id, new_id) in id_mapping:
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -583,7 +583,8 @@ class Options(param.Parameterized):
 
     _option_groups = ['style', 'plot', 'norm']
 
-    def __init__(self, key=None, allowed_keywords=[], merge_keywords=True, max_cycles=None, **kwargs):
+    def __init__(self, key=None, allowed_keywords=[], merge_keywords=True,
+                 max_cycles=None, backend=None, **kwargs):
 
         invalid_kws = []
         for kwarg in sorted(kwargs.keys()):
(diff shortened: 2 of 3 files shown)


holoviews/core/options.py
```diff
@@ -612,6 +613,7 @@ def __init__(self, key=None, allowed_keywords=[], merge_keywords=True, max_cycle
                             else Keywords(allowed_keywords))
         super(Options, self).__init__(allowed_keywords=allowed_keywords,
                                       merge_keywords=merge_keywords, key=key)
+        self.backend = backend
```
jbednar: I can't see all the logic involved here, but I think this should be something like `self.backend = backend if backend is not None else current_backend`.  That way the backend in use when these objects are created is bound to them so that it's clear which backend these options apply to.

holoviews/core/dimension.py
```diff
@@ -1409,10 +1409,16 @@ def options(self, *args, **kwargs):
...
+        elif isinstance(options, list): # List of Options objects
+            expanded_backends = opts._expand_by_backend(options, backend)
         else:
-            expanded = opts._expand_options(options, backend)
-        return self.opts(expanded, backend=backend, clone=clone)
+            expanded_backends = [(opts._expand_options(options, backend), backend)]
+
+        obj = self
+        for backend, expanded  in expanded_backends:
+            obj = obj.opts(expanded, backend=backend, clone=clone)
```
philippjfr: So maybe it's a bit more efficient if we only clone the first time and then set clone=False thereafter.

holoviews/core/options.py
philippjfr: Hmm, you can define these options backends before a backend is loaded and then apply them when it is loaded so it's not possible to rely on this being defined. Therefore I don't think it should be bound to the object here.

holoviews/util/__init__.py
```diff
@@ -244,6 +246,30 @@ def defaults(cls, *options, **kwargs):
...
+                opts_backend = Store.current_backend
+                obj.backend= opts_backend
+            else:
+                opts_backend = backend
+                obj.backend = opts_backend
+                used_fallback = True
+            groups[opts_backend].append(obj)
+
+        if backend and not used_fallback:
+            cls.param.warning('Fallback backend specified but not used.')
```
philippjfr: This warning should be more informative, it's not clear what a fallback backend is. 

```suggestion
            cls.param.warning("All supplied Options objects already define a backend, "
                              "backend override backend=%r will be ignored." % backend)
```

holoviews/core/options.py
jbednar: What will `Store.current_backend` be if there is no backend loaded?  Won't it (correctly) be None?

philippjfr: >What will `Store.current_backend` be if there is no backend loaded?

This actually needs looking at since at least under some conditions it's 'matplotlib' by default. 

However even if it was consistently None (which it should be), this would lead to unpredictable behavior. Options defined before a backend is loaded would default to whatever backend is used when applying the options while if a backend is loaded it immediately defaults to that. Seems like a recipe for confusion to me. I think if unspecified it should therefore default to whatever it is when it is applied, because at that point a backend has to be loaded.

jbednar: > under some conditions it's 'matplotlib' by default.

Definitely need to get rid of that!  I would hope people would be seeing validation errors if they start defining Bokeh options while the backend is defaulting to Matplotlib, so hopefully this would be a weird situation to get into in practice.

All of our examples consistently show loading an extension before doing anything, so I don't think we should be torturing the design to support people who don't do that.

jlstevens: I agree with @philippjfr on all counts here and that there should be no more historical 'matplotlib' by default (which I *hope* is only when you do `hv.extension()`).

> All of our examples consistently show loading an extension before doing anything, so I don't think we should be torturing the design to support people who don't do that.

I completely disagree. One of the goals of holoviews (at least for me) is to declaratively build objects in a very bare environment (e.g on an HPC cluster) that you can then export and load into a richer environment that does have plotting dependencies. Requiring backends to always exist first basically would break the boundary between holoviews-core and the plotting extensions.

holoviews/core/dimension.py
jlstevens: Happy to consider this as an optimization later on if it makes sense. At this time this PR is WIP and a first cut!

holoviews/util/__init__.py
jlstevens: Thanks for the suggestion. What I wrote was a WIP placeholder...

holoviews/core/dimension.py
philippjfr: Sure, seems like a simple optimization, simply set ``clone=False`` on the next line.

holoviews/core/dimension.py
```diff
@@ -1409,10 +1409,16 @@ def options(self, *args, **kwargs):
...
+            expanded_backends = [(backend, {})]
+        elif isinstance(options, list): # List of Options objects
+            expanded_backends = opts._expand_by_backend(options, backend)
         else:
-            expanded = opts._expand_options(options, backend)
-        return self.opts(expanded, backend=backend, clone=clone)
+            expanded_backends = [(backend, opts._expand_options(options, backend))]
+
+        obj = self
+        for backend, expanded  in expanded_backends:
```
philippjfr:
```suggestion
        for backend, expanded in expanded_backends:
```

jlstevens:
As mentioned in #3466, I consider this PR to be implementing a workaround that works with the `opts` utility but not the other ways of specifying options (primarily nested dictionaries). The proper approach is to add 'backend' as an option group, sitting alongside 'plot', 'style' and 'norm' options.

For this reason, I'm not entirely sure this PR should be merged. It might be useful to demonstrate that the approach can be made to work as a prototype but I think the more general mechanism should be used. For that reason, it might be better to make a new PR that supersedes this one. @philippjfr Any thoughts?

philippjfr:
Just to clarify when you talk about the nested dictionary formats do you also include the old formats or merely the dictionary format like this: ``{Element: {option: value}, ...}``?

philippjfr:
I do think it's worth exploring the backend as an options format.

jlstevens:
> Just to clarify when you talk about the nested dictionary formats...

I mean the format used in the 'Split into style , plot and norm options ' section of the [user guide](http://holoviews.org/user_guide/Applying_Customizations.html). Though that might need to be renamed to the 'Split into style , plot, norm and output options ' section. Currently I am only focusing on the backend but you could imagine allowing things normally passed to `hv.output` (e.g the format).

> I do think it's worth exploring the backend as an options format so I'd be fine with putting this on hold while we explore that option.

I'll do that then. This PR shows we can definitely do *something* but maybe we can do something more generally...

jlstevens:
I think I'll close this PR in favor of #3494 which introduces the 'output' option group and is therefore a more thorough approach to solving this problem.

PR #3483, opened 2019-02-11
Update website build to nbsite

philippjfr:
Makes use of ``nbsite.gallery`` (https://github.com/pyviz/nbsite/pull/106)


doc/conf.py
```diff
@@ -3,42 +3,65 @@
...
+pyviz_module = 'holoviews'
 description = 'Stop plotting your data - annotate your data and let it visualize itself.'
 
 # The version info for the project you're documenting, acts as replacement for
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
 # The short X.Y version.
-version = setup_args['version']
+version = 'v1.11.0' #setup_args['version']
```
jsignell: Is the version not accessible?

doc/conf.py
```diff
@@ -3,42 +3,65 @@
...
+                'elements',
+                'containers',
+                'streams',
+                'apps'
+            ],
+            'title': 'Reference Gallery',
+        }
+    },
+    'github_org': 'pyviz',
+    'github_project': 'holoviews'
```
jsignell: could these be a level higher and be used by generate-rst and build as well?

philippjfr: Probably.

doc/conf.py
philippjfr: I had trouble with it locally but will try to restore it before merging.

doc/conf.py
jsignell: Ok that would be a good goal for nbsite. Then we could just write a `doit build_docs` in pyctdev and wouldn't need to run `nbsite generate-rst` and `nbsite build` with all the flags everytime.

doc/conf.py
jsignell: This is what we do on our autover projects 

```python
import colorcet
version = release = colorcet.__version__
```

doc/conf.py
philippjfr: Should handle this in nbsite and then update it here. Will merge this as is for now.

jlstevens:
Looks good and I am happy to see this merged. It will be good once holoviews.org can be built entirely with nbsite (with a nice looking twitter feed that doesn't need manual tweaking!)

PR #3487, opened 2019-02-11
Fixed LICENSE year range

jlstevens:
Minor fix but mostly to check the test data updating infrastructure is now working after moving from ioam to pyviz.


philippjfr:
Thanks, all the infrastructure is working again.

PR #3488, opened 2019-02-11
Fix mistake in VectorField's angle conversion if invert_axes

poplarShift:
Title says it all.


philippjfr:
Thanks again, very much appreciated!

PR #3490, opened 2019-02-12
Fixed Param stream memoization

philippjfr:
This fixes a bug in `Param` and `Pipe` stream memoization. In order to support memoization when a hash is either very expensive to compute or when there is no parameter value to hash on we previously introduced a random hashkey which is returned when a Stream is triggering. This works fine in ensuring memoization is disabled for that particular call but any subsequent events triggered by another stream will receive a new hashkey, which disables memoization. Instead of using a random value this PR therefore introduces a `_memoize_counter` which is incremented whenever an event is triggered on these streams, which means subsequent events receive a consistent value to hash on.

### DIFF
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -157,9 +157,13 @@ class Stream(param.Parameterized):
                 subscriber(**dict(union))
 
         for stream in streams:
+            stream._trigger()
             with util.disable_constant(stream):
                 if stream.transient:
                     stream.reset()
+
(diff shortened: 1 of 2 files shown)


jlstevens:
Makes sense - a simple counter can always be unique but also nice and predictable. PR looks good, I'll merge when the tests are green and when you tell me it is ready.

philippjfr:
Ready to merge now.

PR #3491, opened 2019-02-12
Enable the use of xlim parameter for timeseries plots

henriqueribeiro:
Hey,

Since `xlim` parameter was a `NumericTuple` it was not possible to set it using dates for timeseries plot. Changing it to `Tuple` makes it more flexible.


philippjfr:
Thanks, very good point. Will merge once tests pass.

philippjfr:
Actually, would you mind changing ylim too?

henriqueribeiro:
Sure, no problem!

henriqueribeiro:
Also changed the zlim

PR #3494, opened 2019-02-13
Added an 'output' options group

jlstevens:
This PR builds on #3492 in order to tackle the suggestion in #3463: tracking backend in options should be done consistently with everything else by introducing an 'output' options group.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -1399,7 +1399,7 @@ class Dimensioned(LabelledData):
             if kwargs:
                 raise ValueError('Please specify a list of option objects, or kwargs, but not both')
             options = args[0]
-        elif args and kwargs:
+        elif args and [k for k in kwargs.keys() if k != 'backend']:
             raise ValueError("Options must be defined in one of two formats. "
                              "Either supply keywords defining the options for "
                              "the current object, e.g. obj.options(cmap='viridis'), "
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -18,7 +18,7 @@ import param
 import numpy as np
 
 from . import util
-from .options import Store, Opts, cleanup_custom_options
+from .options import Store, Opts, Options, cleanup_custom_options
 from .pprint import PrettyPrinter
 from .tree import AttrTree
 from .util import basestring, OrderedDict, bytes_to_unicode, unicode
@@ -1406,6 +1406,8 @@ class Dimensioned(LabelledData):
                              "or explicitly define the type, e.g. "
                              "obj.options({'Image': {'cmap': 'viridis'}}). "
                              "Supplying both formats is not supported.")
+        elif args and isinstance(args[0], dict):
+            options = [Options(spec, **kws) for spec,kws in args[0].items()]
         elif args:
             options = list(args)
         elif kwargs:
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -1388,7 +1388,7 @@ class Dimensioned(LabelledData):
         Returns:
             Returns the cloned object with the options applied
         """
-        backend = kwargs.pop('backend', None)
+        backend = kwargs.get('backend', None)
         clone = kwargs.pop('clone', True)
 
         if len(args) == 0 and len(kwargs)==0:
@@ -1413,11 +1413,16 @@ class Dimensioned(LabelledData):
 
         from ..util import opts
         if options is None:
-            expanded = {}
+            expanded_backends = [(backend, {})]
+        elif isinstance(options, list): # assuming a flat list of Options objects
+            expanded_backends = opts._expand_by_backend(options, backend)
         else:
-            expanded = opts._expand_options(options, backend)
-        return self.opts(expanded, backend=backend, clone=clone)
+            expanded_backends = [(backend, opts._expand_options(options, backend))]
 
+        obj = self
+        for backend, expanded in expanded_backends:
+            obj = obj.opts(expanded, backend=backend, clone=clone)
+        return obj
 
     def _repr_mimebundle_(self, include=None, exclude=None):
         """
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -1421,7 +1421,7 @@ class Dimensioned(LabelledData):
 
         obj = self
         for backend, expanded in expanded_backends:
-            obj = obj.opts(expanded, backend=backend, clone=clone)
+            obj = obj.opts._dispatch_opts(expanded, backend=backend, clone=clone)
         return obj
 
     def _repr_mimebundle_(self, include=None, exclude=None):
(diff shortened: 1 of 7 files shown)


jlstevens:
Note that since commit 226d65005, you can do this:

image

Deciding whether we want to use this flexibility to control renderers is a point for discussion. I just wanted to point out that this is now possible.

jlstevens:
I think this PR is now ready to review/merge.

holoviews/tests/core/testdimensioned.py
```diff
@@ -38,7 +38,8 @@ def register_custom(cls, objtype, backend, custom_plot=[], custom_style=[]):
         style_opts = Keywords(['style_opt1', 'style_opt2']+custom_style, name)
         plot_opts = Keywords(['plot_opt1', 'plot_opt2']+custom_plot, name)
         opt_groups = {'plot': Options(allowed_keywords=plot_opts),
-                      'style': Options(allowed_keywords=style_opts)}
+                      'style': Options(allowed_keywords=style_opts),
+                      'output':Options(allowed_keywords=['backend'])}
```
philippjfr:
```suggestion
                      'output': Options(allowed_keywords=['backend'])}
```

holoviews/core/util.py
```diff
@@ -242,7 +242,7 @@ def deprecated_opts_signature(args, kwargs):
     """
     from .options import Options
     groups = set(Options._option_groups)
-    opts = {kw for kw in kwargs if kw not in ('backend', 'clone')}
+    opts = {kw for kw in kwargs if kw !='clone'}
```
philippjfr:
```suggestion
    opts = {kw for kw in kwargs if kw != 'clone'}
```

holoviews/core/dimension.py
```diff
@@ -1388,7 +1388,7 @@ def options(self, *args, **kwargs):
         Returns:
             Returns the cloned object with the options applied
         """
-        backend = kwargs.pop('backend', None)
+        backend = kwargs.get('backend', None)
```
philippjfr: Now that the backend is still in kwargs won't the ``elif args and kwargs:`` branch execute and error?

holoviews/core/dimension.py
```diff
@@ -1406,18 +1406,25 @@ def options(self, *args, **kwargs):
                              "or explicitly define the type, e.g. "
                              "obj.options({'Image': {'cmap': 'viridis'}}). "
                              "Supplying both formats is not supported.")
+        elif args and isinstance(args[0], dict):
+            options = [Options(spec, **kws) for spec,kws in args[0].items()]
```
philippjfr: If there is more than one arg this should either warn or process all the args, silently dropping other args doesn't seem great.

holoviews/core/options.py
```diff
@@ -158,6 +158,18 @@ def __call__(self, *args, **kwargs):
         Returns:
             Returns the object or a clone with the options applied
         """
+        if self._mode is None:
+            apply_groups, _, _ = deprecated_opts_signature(args, kwargs)
+            if apply_groups and config.future_deprecations:
+                msg = ("Calling the .opts method with options broken down by options "
```
philippjfr: Confused this is in ``__call__`` but is warning about ``.opts``?

holoviews/core/dimension.py
jlstevens: I don't think there is any issue here - the point of the PR is that `backend` is now like any other kwarg. It is just that the value is needed here (if available) for `obj._dispatch_opts`.

holoviews/core/options.py
```diff
@@ -224,13 +236,6 @@ def _base_opts(self, *args, **kwargs):
 
         # By default do not clone in .opts method
         clone = kwargs.get('clone', None)
-
-        if apply_groups and config.future_deprecations:
-            msg = ("Calling the .opts method with options broken down by options "
-                   "group (i.e. separate plot, style and norm groups) is deprecated. "
-                   "Use the .options method converting to the simplified format "
-                   "instead or use hv.opts.apply_groups for backward compatibility.")
-            param.main.warning(msg)
```
philippjfr: Is there now any future warning for the deprecated formats? I know there was the issue with the recursive calls which meant it was always triggered when enabled but I had hoped there was a way around this.

holoviews/core/dimension.py
philippjfr: It will now error when you supply args and the backend as a kwarg as far as I can tell.

philippjfr: e.g. as far as I can tell ``.options(opts.Image(cmap='Reds'), backend='matplotlib')`` will get caught up in the ``elif args and kwargs`` branch and raise a ValueError. Maybe I'm missing something?

holoviews/core/dimension.py
jlstevens: Agreed though the use of `args[0]` wasn't introduced in this PR. I'll add the warnings as suggested.

holoviews/core/options.py
philippjfr: Nevermind, I realize this is the .opts accessor object.

holoviews/core/options.py
philippjfr: Nevermind, same confusion as the previous comment about ``__call__``. I can't visualize who calls what anymore, I'd really need a diagram to see how the different formats get converted and then pencil in the warnings at the appropriate stages.

holoviews/core/dimension.py
jlstevens: Unfortunately, it isn't simple to fix due to the way `opts` and `options` invoke each other. Valid signatures would issue such a warning.

holoviews/core/dimension.py
jlstevens: I don't think `.options` is supposed to accept `Option` objects afaik. Only `.opts` should.

holoviews/core/dimension.py
philippjfr: Not sure I follow, when would passing in multiple dictionaries as kwargs ever occur?

holoviews/core/dimension.py
philippjfr: I didn't think that was true, I thought they were basically equivalent except for clone and that ``.options`` doesn't allow the really old nested format. In any case though the point I'm making, the same problem applies to ``.opts({'Image': dict(cmap='Reds')}, backend='matplotlib')``

jlstevens: Addressed in f112875

philippjfr:
Looks good will merge once passing.

holoviews/core/dimension.py
jlstevens: I can probably warn for this one condition when dicts are involved but not for other cases where `args[0]` is used and potentially dropping other positional arguments.

jlstevens: Addressed in 9b9240bc1

PR #3495, opened 2019-02-14
Implement optimized TriMesh wireframe rendering

philippjfr:
Uses optimized datashader segments rendering for TriMesh wireframes and adds support for generating a wireframe using ``interpolation=None`` and ``aggregator='any'/'count'``.

Here's the performance profiling for the initial render (which includes the time converting the trimesh to a wireframe of paths and segments):

bokeh_plot

- [x] Add tests

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -703,10 +703,15 @@ class trimesh_rasterize(aggregate):
                 'vertices': verts}
 
     def _precompute_wireframe(self, element, agg):
-        simplexes = element.array([0, 1, 2, 0]).astype('int')
-        verts = element.nodes.array([0, 1])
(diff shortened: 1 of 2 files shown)


philippjfr:
@kcpevey Once this is merged and there's been a dev release you'll be able to speed up wiremesh plots by changing:

```
datashade(trimesh.edgepaths)
```

to:

```
datashade(trimesh.edgepaths, aggregator='any', interpolation=None)
```

jlstevens:
Looks good to me! Happy to see this merged.

philippjfr:
Okay, the long term plan here is that we will eventually add a Segments element and then expose that using a public ``TriMesh.wireframe`` property.

PR #3499, opened 2019-02-15
Always warn for deprecated call

jlstevens:
Looks like the last release still had the warning behind the [`util.config.future_deprecations` flag](https://github.com/pyviz/holoviews/blob/v1.11.2/holoviews/core/dimension.py#L453). This PR makes sure the warning is always issued so we can finally remove ``__call__`` immediately after the next release.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -306,9 +306,8 @@ class Dimension(param.Parameterized):
 
 
     def __call__(self, spec=None, **overrides):
-        if util.config.future_deprecations:
-            self.param.warning('Dimension.__call__ method has been deprecated, '
-                               'use the clone method instead.')
+        self.param.warning('Dimension.__call__ method has been deprecated, '
+                           'use the clone method instead.')
         return self.clone(spec=spec, **overrides)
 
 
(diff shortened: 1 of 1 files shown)


jlstevens:
Probably worth greping the travis logs for the warning in case it was missed anywhere internally (doubt it!)

philippjfr:
Just to double check, you are aware that this is ``Dimension.__call__`` and not ``Dimensioned.__call__`` right? This warning was indeed newly added in 1.11 afaik so enabling it in a micro release seems wrong.

jlstevens:
Oops yes and `util.config.warn_options_call` was `True` in [1.11](https://github.com/pyviz/holoviews/blob/v1.11.0/holoviews/core/util.py#L126).

4cf672903 restores `Dimension` and removes ``__call__` from `Dimensioned` entirely.

jlstevens:
Even if the tests go green, this PR doesn't need to be merged just yet as we will probably have another minor release first.

philippjfr:
I've restored the method and am now merging. The fact that we never actually enabled the warning means that we can't remove it yet.

PR #3530, opened 2019-02-26
Improved datetime handling in utils to avoid type mixing

### LINKED ISSUE #3528: Datetime axis problems in nested Layouts
```
import holoviews as hv
import numpy as np
import pandas as pd
import xarray as xr
hv.extension('bokeh')
dates = pd.date_range('2017-1-1', '2018-1-1', periods=242)
# dates = np.arange(242)

rr = xr.DataArray(np.random.rand(242,3),
                  coords=dict(Date=dates, Float=range(3)), 
                  dims=['Date', 'Float'],
                  name='r')

i = pd.DataFrame(dict(ice=range(242), Date=dates))
i2 = pd.DataFrame(dict(sic=np.linspace(0,300,242), Date=dates))

a = hv.Dataset(rr, kdims=['Date', 'Float'], vdims=hv.Dimension('r', range=(-.01, .05))
              ).to(hv.Scatter).overlay()
b = hv.Curve(i, 'Date', hv.Dimension('ice', range=(0,100)))
bb = hv.Scatter(i2, 'Date', 'sic')
l = a+b*bb
print(l)
l
```

gives a `TypeError`. The Layout is:
```
:Layout
   .NdOverlay.I :NdOverlay   [Float]
      :Scatter   [Date]   (r)
   .Overlay.I   :Overlay
      .Curve.I   :Curve   [Date]   (ice)
      .Scatter.I :Scatter   [Date]   (sic)
```

Replacing the `dates` coordinate by something numeric it works, as do the individual combinations `a+bb`, `a+bb`, `b*bb`.

poplarShift:
I came across two util functions that do not take into account the possibility that different datetime types may be present in an array, such as when comparing axis ranges in overlayed plots (how this can happen in the first place might be worth investigating in more detail at some point). E.g. combining both pandas and numpy datetimes in the same array ruins calls such as to `numpy.max` etc.

Fixes #3528

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -928,28 +928,42 @@ def find_range(values, soft_range=[]):
 def max_range(ranges, combined=True):
     """
     Computes the maximal lower and upper bounds from a list bounds.
+
+    Args:
+       ranges (list of tuples): A list of range tuples
+       combined (boolean, optional): Whether to combine bounds
+          Whether range should be computed on lower and upper bound
+          independently or both at once
+
+    Returns:
+       The maximum range as a single tuple
     """
     try:
         with warnings.catch_warnings():
             warnings.filterwarnings('ignore', r'All-NaN (slice|axis) encountered')
             values = [tuple(np.NaN if v is None else v for v in r) for r in ranges]
-            if pd and any(isinstance(v, datetime_types)
+            if pd and any(isinstance(v, datetime_types) and not isinstance(v, cftime_types)
                           for r in values for v in r):
-                values = [
-                    (pd.Timestamp(i).to_datetime64(),
-                     pd.Timestamp(j).to_datetime64())
-                    if isinstance(i, datetime_types)
-                    and isinstance(j, datetime_types)
-                    else (i, j) for i, j in values
-                ]
+                converted = []
+                for l, h in values:
+                    if isinstance(l, datetime_types) and isinstance(h, datetime_types):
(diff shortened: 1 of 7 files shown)


philippjfr:
Can you think of any tests you could add here? Otherwise I'm happy to merge this.

poplarShift:
Thanks, I'll think about that. Don't think we'd want to be too strict on what it should be able to handle either.

holoviews/core/util.py
```diff
@@ -933,8 +933,15 @@ def max_range(ranges, combined=True):
...
             values = [tuple(np.NaN if v is None else v for v in r) for r in ranges]
-            if pd and all(isinstance(v, pd.Timestamp) for r in values for v in r):
-                values = [(v1.to_datetime64(), v2.to_datetime64()) for v1, v2 in values]
+            if pd and any(isinstance(v, datetime_types)
+                          for r in values for v in r):
+                values = [
+                    (pd.Timestamp(i).to_datetime64(),
+                     pd.Timestamp(j).to_datetime64())
+                    if isinstance(i, datetime_types)
+                    and isinstance(j, datetime_types)
```
philippjfr: I also think this probably needs special handling for `cftime` types which have all kinds of additional complexities.

philippjfr: For now it's probably best to make this code ignore cftime types.

philippjfr:
Merging this despite notebook tests failing, the regression is due to changes in param which are fixed in https://github.com/pyviz/holoviews/pull/3474.

Issue #3498, opened 2019-02-15
Add a Multi-Segments element


philippjfr:
In recent versions of datashader there are much more efficient means of drawing fixed length segments, however we have to way of representing that data in HoloViews. I therefore propose we add a ``Segments`` element where the x-/y- coordinates are stored as key dimensions. This can also be used to render a TriMesh wireframe (indeed internally the `trimesh_rasterize` operation already does this). Once added I'd therefore create a public property on TriMesh which makes the wireframe available as a Segments element.

poplarShift:
+1!

You mean something like Bokeh's segment, so `kdims=['x0', 'x1', 'y0', 'y1']`?

philippjfr:
I haven't started yet no, we were thinking of a slight generalization of ``kdims=['x0', 'x1', 'y0', 'y1']`` which would allow multiple vertices per row, i.e. ``kdims=['x0', 'x1', 'x2', ..., 'y0', 'y1', 'y2', ...]``. That does stretch the definition of segment a bit, so maybe a different name would be more appropriate.

philippjfr:
Also should decide whether alternating (``['x0', 'y0', 'x1', ...]``) or consecutive (``['x0', 'x1', 'y0', ...]``) x/y dimensions are more intuitive.

poplarShift:
In that case I'd prefer alternating - it would be easier to modify the number of nodes, just clipping/pasting in one place instead of at least two per node.

philippjfr:
Segments were added.

PR #3532, opened 2019-02-27
Segments

poplarShift:
A quick first draft (sufficient for my own use case, which is why I went ahead and put something together). In particular, I haven't yet thought about how to best implement a variable number of kdims in the bokeh backend which is less important to me at this point but which @philippjfr mentioned should be implemented.

Addresses #3498

screen shot 2019-02-27 at 11 28 23 am

screen shot 2019-02-27 at 11 28 15 am

### DIFF
--- a/holoviews/element/chart.py
+++ b/holoviews/element/chart.py
@@ -5,7 +5,7 @@ from ..streams import BoundsXY
 from ..core import util
 from ..core import Dimension, Dataset, Element2D
 from ..core.data import GridInterface
-from .geom import Points, Geometry, VectorField # noqa: backward compatible import
+from .geom import Points, Geometry, Segments, VectorField # noqa: backward compatible import
 from .stats import BoxWhisker         # noqa: backward compatible import
 
 
@@ -368,14 +368,3 @@ class Area(Curve):
                                     new_type=Area)
             baseline = baseline + y
         return stacked
-
-class Segments(Geometry):
-    """
-    """
-    group = param.String(default='Segments', constant=True)
-
-    kdims = param.List(default=[Dimension('x0'), Dimension('y0'),
-                                Dimension('x1'), Dimension('y1')],
-                       bounds=(4, None), constant=True, doc="""
-        Segments represent lines in given by x- and y-
-        coordinates in a 2D space.""")
--- a/holoviews/element/chart.py
+++ b/holoviews/element/chart.py
@@ -5,7 +5,7 @@ from ..streams import BoundsXY
 from ..core import util
 from ..core import Dimension, Dataset, Element2D
 from ..core.data import GridInterface
-from .geom import Points, Geometry, Segments, VectorField # noqa: backward compatible import
+from .geom import Points, VectorField # noqa: backward compatible import
 from .stats import BoxWhisker         # noqa: backward compatible import
 
 
--- a/holoviews/element/chart.py
+++ b/holoviews/element/chart.py
@@ -97,7 +97,6 @@ class Scatter(Chart2dSelectionExpr, Chart):
     location along the x-axis while the first value dimension
     represents the location of the point along the y-axis.
     """
-
     group = param.String(default='Scatter', constant=True)
 
 
--- a/holoviews/element/chart.py
+++ b/holoviews/element/chart.py
@@ -5,7 +5,7 @@ from ..streams import BoundsXY
 from ..core import util
 from ..core import Dimension, Dataset, Element2D
 from ..core.data import GridInterface
-from .geom import Points, VectorField # noqa: backward compatible import
+from .geom import Points, Geometry, VectorField # noqa: backward compatible import
 from .stats import BoxWhisker         # noqa: backward compatible import
 
 
@@ -368,3 +368,14 @@ class Area(Curve):
                                     new_type=Area)
             baseline = baseline + y
         return stacked
+
+class Segments(Geometry):
+    """
+    """
+    group = param.String(default='Segments', constant=True)
+
+    kdims = param.List(default=[Dimension('x0'), Dimension('y0'),
+                                Dimension('x1'), Dimension('y1')],
+                       bounds=(4, None), constant=True, doc="""
+        Segments represent lines in given by x- and y-
+        coordinates in a 2D space.""")
--- a/holoviews/element/geom.py
+++ b/holoviews/element/geom.py
@@ -80,7 +80,7 @@ class Points(GeometrySelectionExpr, Geometry):
 
 class VectorField(GeometrySelectionExpr, Geometry):
     """
-    A VectorField represents a set of vectors in 2D spac with an
+    A VectorField represents a set of vectors in 2D space with an
     associated angle, as well as an optional magnitude and any number
     of other value dimensions. The angles are assumed to be defined in
     radians and by default the magnitude is assumed to be normalized
@@ -91,3 +91,16 @@ class VectorField(GeometrySelectionExpr, Geometry):
 
     vdims = param.List(default=[Dimension('Angle', cyclic=True, range=(0,2*np.pi)),
                                 Dimension('Magnitude')], bounds=(1, None))
+
+
+class Segments(Geometry):
+    """
+    Segments represent a collection of lines in 2D space.
+    """
+    group = param.String(default='Segments', constant=True)
+
+    kdims = param.List(default=[Dimension('x0'), Dimension('y0'),
+                                Dimension('x1'), Dimension('y1')],
+                       bounds=(4, None), constant=True, doc="""
+        Segments represent lines given by x- and y-
+        coordinates in 2D space.""")
--- a/holoviews/element/geom.py
+++ b/holoviews/element/geom.py
@@ -101,6 +101,6 @@ class Segments(Geometry):
 
     kdims = param.List(default=[Dimension('x0'), Dimension('y0'),
                                 Dimension('x1'), Dimension('y1')],
-                       bounds=(4, None), constant=True, doc="""
+                       bounds=(4, 4), constant=True, doc="""
         Segments represent lines given by x- and y-
         coordinates in 2D space.""")
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -1124,3 +1124,16 @@ class SegmentPlot(ColorbarPlot):
         data = {'x0': x0s, 'x1': x1s, 'y0': y0s, 'y1': y1s}
         mapping = dict(x0='x0', x1='x1', y0='y0', y1='y1')
         return (data, mapping, style)
+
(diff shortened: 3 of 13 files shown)


philippjfr:
Very neat, while I did suggest that we could/should support multiple segments per row that could be another element entirely as ``Segments`` definitely indicates to me that it's one segment per row.

poplarShift:
Indeed, that could be something like `Lines`? Although I'd be weary of code duplication.

holoviews/plotting/mpl/chart.py
```diff
@@ -1258,3 +1258,78 @@ class SideSpikesPlot(AdjoinedPlot, SpikesPlot):
...
+    """
+    # Deprecated parameters
+
+    color_index = param.ClassSelector(default=None, class_=(basestring, int),
+                                      allow_None=True, doc="""
+        Deprecated in favor of color style mapping, e.g. `color=dim('color')`""")
+
+    size_index = param.ClassSelector(default=None, class_=(basestring, int),
+                                     allow_None=True, doc="""
+        Deprecated in favor of size style mapping, e.g. `size=dim('size')`""")
```
philippjfr: Let's not add these deprecated parameters to new elements.

holoviews/plotting/bokeh/chart.py
```diff
@@ -1055,3 +1055,54 @@ def get_data(self, element, ranges, style):
...
+    """
+    # Deprecated parameters
+
+    color_index = param.ClassSelector(default=None, class_=(basestring, int),
+                                      allow_None=True, doc="""
+        Deprecated in favor of color style mapping, e.g. `color=dim('color')`""")
+
+    size_index = param.ClassSelector(default=None, class_=(basestring, int),
+                                     allow_None=True, doc="""
+        Deprecated in favor of size style mapping, e.g. `size=dim('size')`""")
```
philippjfr: Same comment as below, we shouldn't need these on new element plot classes.

holoviews/element/chart.py
```diff
@@ -281,3 +281,14 @@ def stack(cls, areas):
                                     new_type=Area)
             baseline = baseline + y
         return stacked
+
+class Segments(Geometry):
```
philippjfr: Let's move this definition to ``elements/geom.py``

poplarShift: Agreed.

holoviews/plotting/bokeh/chart.py
poplarShift: Agreed - I thought originally there might be compatibility issues with adapting old code to use the new Segments element.

holoviews/plotting/mpl/chart.py
poplarShift: Agreed, see above.

holoviews/plotting/bokeh/chart.py
poplarShift: Well, I should have tested before I replied - simply removing them does not work. So how about color_index etc. are deprecated at the same point in time?
```
~/coding/github/holoviews-dev/holoviews/plotting/bokeh/element.py in _get_color_data(self, element, ranges, style, name, factors, colors, int_categories)
   1498                         int_categories=False):
   1499         data, mapping = {}, {}
-> 1500         cdim = element.get_dimension(self.color_index)
   1501         color = style.get(name, None)
   1502         if cdim and ((isinstance(color, util.basestring) and color in element) or isinstance(color, dim)):

AttributeError: 'SegmentPlot' object has no attribute 'color_index'
```

philippjfr: You can skip the entire call to ``_get_color_data``.

poplarShift: Ah of course, I'd missed a line or two there.

poplarShift:
Ok, all your requested changes should have been pushed now.

holoviews/plotting/bokeh/chart.py
philippjfr: I'm looking forward to the day we can delete this mess from the other classes so it's nice to see how much neater new plot classes not burdened with this legacy can be.

philippjfr:
Thanks, before merging we still need reference gallery entries in ``examples/reference/element/bokeh/Segment.ipynb`` and ``examples/reference/element/matplotlib/Segment.ipynb``, but I'm happy to help out on those. They just need to describe what this element represents how to use it in a simple case and a more complex case.

Also if you can think of any examples in the main gallery that could be converted to Segments that might be nice too.

holoviews/element/chart.py
```diff
@@ -5,7 +5,7 @@
 from ..core import util
 from ..core import Dimension, Dataset, Element2D
 from ..core.data import GridInterface
-from .geom import Points, VectorField # noqa: backward compatible import
+from .geom import Points, Geometry, Segments, VectorField # noqa: backward compatible import
```
philippjfr: The others are just listed for backward compatibility.

```suggestion
from .geom import Points, VectorField # noqa: backward compatible import
```

holoviews/element/geom.py
```diff
@@ -50,3 +50,16 @@ class VectorField(Geometry):
...
+
+class Segments(Geometry):
+    """
+    Segments represent a collection of lines in 2D space.
+    """
+    group = param.String(default='Segments', constant=True)
+
+    kdims = param.List(default=[Dimension('x0'), Dimension('y0'),
+                                Dimension('x1'), Dimension('y1')],
+                       bounds=(4, None), constant=True, doc="""
```
philippjfr: I think we have decided to keep segments simple, so it should always have 4 dimensions:

```suggestion
                       bounds=(4, 4), constant=True, doc="""
```

holoviews/plotting/mpl/chart.py
```diff
@@ -1258,3 +1258,47 @@ class SideSpikesPlot(AdjoinedPlot, SpikesPlot):
...
+
+    def get_data(self, element, ranges, style):
+        # Get [x0, y0, x1, y1, ...]
+        x0idx, y0idx, x1idx, y1idx = (
+            (1, 0, 3, 2) if self.invert_axes else (0, 1, 2, 3)
+        )
+        # Compute segments
+        dims = element.dimensions()
+        data = [[(x0, y0), (x1, y1)] for x0, y0, x1, y1
+                in element.array([x0idx, y0idx, x1idx, y1idx])]
```
philippjfr: I think we need to unpack them all separate since ``.array`` may coerce types incorrectly.

holoviews/plotting/mpl/chart.py
```diff
@@ -1258,3 +1258,47 @@ class SideSpikesPlot(AdjoinedPlot, SpikesPlot):
...
+        return (data,), style, {'dimensions': dims}
+
+    def get_extents(self, element, ranges, range_type='combined'):
+        kdims = element.kdims
+        for kdim in [kdims[i].name for i in range(2)]:
+            new_range = {}
+            for r in ranges[kdim]:
+                new_range[r] = max_range([ranges[kd.name][r] for kd in kdims])
+            ranges[kdim] = new_range
+        return super(SegmentPlot, self).get_extents(element, ranges, range_type)
```
philippjfr: May have to implement this twice, we still don't have a good solution for element specific code that's shared across backends.

philippjfr:
Sorry for the two-stage review, I was short on time earlier.

holoviews/plotting/mpl/chart.py
philippjfr: In the long term if we want to add more geometry elements we should consider whether it'd be worth it to add "virtual" dimensions to elements which can return the x/y/(z) ranges rather than just the ranges of each column.

holoviews/element/chart.py
poplarShift: In that case we'd need a `from .geom import *` in `__init__.py`, unless I'm missing something.

holoviews/plotting/mpl/chart.py
poplarShift: You're thinking of `element.dimension_values()`; or `element.array([x0idx])` etc?

holoviews/plotting/mpl/chart.py
poplarShift: You mean port `get_extents()` to the bokeh backend?

holoviews/plotting/mpl/chart.py
philippjfr: Either works, the former is more canonical (although I've wanted a short API for it for a long time).

holoviews/element/chart.py
philippjfr: It should be imported into ``element/__init__.py``, there's no need to expose it in ``element/chart.py``.

holoviews/plotting/mpl/chart.py
philippjfr: Right.

holoviews/element/chart.py
poplarShift: Yes - but any sort of import from `.geom` has been missing from `element/__init__.py` so as far as I can see the import of VectorField has been relying on exposing it in `chart.py` first.

philippjfr: Sure, we should just add that.

holoviews/plotting/mpl/chart.py
poplarShift: So in fixing get_extent (turned out my implementation didn't do much), I found that in the bokeh backend, get_extent is being called twice for a simple Segments element - in mpl only once. Out of curiosity, any idea why?

poplarShift:
I added example example. I don't think there was any data in the holoviews repo that represented time periods so I just made some mock data, plus some fractal stuff to illustrate more advanced (?) styling.

holoviews/plotting/mpl/chart.py
poplarShift: Done

holoviews/element/chart.py
poplarShift: Done

holoviews/plotting/mpl/chart.py
philippjfr: I think in bokeh it calls it during initialization and then calls it again in `_update_plot`. Since the ranges are all cached that shouldn't be very expensive but we should probably look into fixing it at some point.

philippjfr:
Looking great. If you don't mind I'll work on a datashader operation which rasterizes segments efficiently and push it to the PR.

poplarShift:
Great, go ahead!

philippjfr:
Sorry this has sat around for so long, I'll try to get this in soon.

philippjfr:
Just rebased the PR and added datashading support. Just needs some tests now.

poplarShift:
Ah, great!

philippjfr:
Sorry this took so long!

poplarShift:
No worries, I'm just happy having contributed something!

PR #3535, opened 2019-03-01
Exposed option to set Sankey font size

philippjfr:
Fixes https://github.com/pyviz/holoviews/issues/3531


philippjfr:
Merging this despite notebook test failures, which are related to the new param version and fixed in https://github.com/pyviz/holoviews/pull/3474

PR #3545, opened 2019-03-07
Unified setup.py and meta.yaml

philippjfr:
Temporarily pins dask to avoid regression in 1.1.3, which we will investigate separately.


philippjfr:
Merging shortly to get the tests passing again.

PR #3553, opened 2019-03-11
Website: fix links in footer

ceball:


ceball:
I also happened to notice that in both http://holoviews.org/user_guide/Notebook_Magics.html and http://build.holoviews.org/user_guide/Notebook_Magics.html, the first link - to the 'applying customization' notebook - doesn't work. Customizations is plural when it shouldn't be, or the other way round. I'm not sure what it's supposed to be.

philippjfr:
Will merge, further fixes can happen in subsequent PRs.

PR #3554, opened 2019-03-12
Allow using apply method with method strings

philippjfr:
Implements the suggestion outlined in https://github.com/pyviz/datashader/pull/723. This allows the apply method to accept strings which refer to methods on the object, e.g.:

color_picker

sample_apply

After some discussion @jlstevens and I have decided to make ``.apply`` a namespace accessor with the following convenience methods:

* ``.apply.opts``
* ``.apply.select``
* ``.apply.reduce``
* ``.apply.aggregate``

This also means we can deprecate ``HoloMap.reduce`` and ``HoloMap.sample`` which are outliers in our API (which are inherited but are not actually usable on DynamicMap) and can be achieved by chaining apply and collapse methods.


philippjfr:
@jlstevens Ready to review.

holoviews/core/accessors.py
```diff
@@ -0,0 +1,463 @@
...
+                keyword argument, or the supplied function is a
+                parameterized method.
+            kwargs (dict, optional): Additional keyword arguments
+                Keyword arguments which will be supplied to the
+                function.
+
+        Returns:
+            A new object where the function was applied to all
+            contained (Nd)Overlay or Element objects.
+        """
```
jlstevens: Given that the only state used in `__call__` is `self._obj` I would consider pulling this out into a utility with a signature `apply(obj, function, streams=[], link_inputs=True, dynamic=None, **kwargs)`...then this method would just be `self.apply(obj, function, streams=[], link_inputs=True, dynamic=None, **kwargs)`.

This method is also a tad long but I think it is ok as I don't see an obvious way to break it up.

holoviews/core/accessors.py
```diff
@@ -0,0 +1,463 @@
...
+            if not all(samples):
+                raise ValueError('Applying a function to a DynamicMap '
+                                 'and setting dynamic=False is only '
+                                 'possible if key dimensions define '
+                                 'a discrete parameter space.')
+            return HoloMap(self._obj[samples]).apply(
+                function, streams, link_inputs, dynamic, **kwargs)
+
+        if isinstance(function, util.basestring):
+            args = kwargs.pop('_args', ())
```
jlstevens: Very minor suggestion, I might recommend renaming it to `_function_args` which makes it a little clearer what it is...

holoviews/core/accessors.py
```diff
@@ -0,0 +1,463 @@
...
+        """Applies a selection to all ViewableElement objects.
+
+        See :py:meth:`Dimensioned.opts` and :py:meth:`Apply.__call__`
+        for more information.
+        """
+        return self.__call__('select', **kwargs)
+
+
+
+class redim(object):
```
jlstevens: Any hard in capitalizing this class name to be consistent with the others? (i.e `Redim`)?

jlstevens: Maybe pickling issues...could leave an alias around I suppose but maybe it is no longer worth it...

holoviews/core/accessors.py
philippjfr: Could you explain why you'd prefer that? I don't think users should use it directly and ``redim`` and ``Opts`` accessors don't do this although they also just use ``self._obj`` as state.

holoviews/core/accessors.py
philippjfr: Sounds good.

holoviews/core/accessors.py
philippjfr: Right pickling issues was the reason I left it.

>maybe it is no longer worth it...

To be clear, do you mean renaming it is not worth it?

holoviews/core/accessors.py
philippjfr: The utility would also live in this file, so it doesn't help any from an organization standpoint either.

jlstevens: > ... although they also just use self._obj as state.

That isn't entirely true, unlike `Apply` they have `self.mode`. The reason I like utility functions is they are more functional - there is nowhere for state to hide outside the call signature. However, I accepts your argument that the other accessors don't do this and now that this code is in a new file anyway, I guess I take back that suggestion.

holoviews/core/accessors.py
philippjfr: I might go with ``_method_args`` as this is specifically meant for the ``apply.opts``, ``apply.select``, etc. accessors.

jlstevens:
Other than two minor suggestions (`_args` -> `_function_args` and `redim` -> `Redim`) I am very happy with the PR and pleased to see the accessors live in a separate file. Happy to merge once the renaming is done and the tests are green.

holoviews/core/accessors.py
jlstevens: Right. Basically if you think renaming `redim` to `Redim` is easy then it would be good to make things consistent. Otherwise, don't bother.

philippjfr: I think I'll file an issue about cleaning this up, for 2.0 we can break imports and I've already made a note to fix some of the inline imports.

jlstevens: Sounds good.

jlstevens:
Tests are passing now and I'll go ahead and merge. Excited to try it out!

PR #3556, opened 2019-03-13
add radius for node options for bokeh

reckoner:
Zooming in causes the node sizes to re-scale, thus defeating the purpose of the zoom. To allow node size to *increase* with zoom-in, Bokeh 
needs the `radius` of the `Circle` set. 
This is discussed here: https://stackoverflow.com/questions/41168347/python-bokeh-google-maps-dynamic-glyph-size-based-on-zoom


philippjfr:
I'll merge when tests pass.

PR #3575, opened 2019-03-24
Fixes when specifying both data_aspect and responsive modes

philippjfr:
Fixes a number of bugs in handling of aspect and responsive modes:

* Ensures that if data_aspect zoom tools are always locked
* Fixes inverted aspect_ratio when responsive=True
* Ensure that aspect is applied even if data_aspect is set


jlstevens:
Thanks, I just tested it and the horrible stretched layout is gone and it now works great!

PR #3580, opened 2019-03-26
Ensure Dimension clone copies all parameters

### LINKED ISSUE #3561: Using .redim.defaults overrides other Dimension attributes
As [discussed on StackOverflow](https://stackoverflow.com/questions/55092912/specifying-the-default-value-in-widgets-generated-by-pyviz-hvplot), it would be nice to be able to specify the default values for widgets created automatically by `hvplot`. 

I've got a 4D dataset and `hvplot` is nicely generating widgets for the dimensions I've specified in the `groupby`:
```python
import xarray as xr

ds = xr.open_dataset('http://gamone.whoi.edu/thredds/dodsC/coawst_4/use/fmrc/coawst_4_use_best.ncd')

rho_vars = ['salt','temp']

from cartopy import crs as ccrs
import hvplot.xarray
import holoviews as hv
from geoviews import tile_sources as gvts
import panel as pn

var_select = pn.widgets.Select(name='Variables:', options=rho_vars, 
                               value='temp')

crs = ccrs.PlateCarree()

def plot(var=None):
    base_map = gvts.OSM
    var = var or var_select.value
    mesh = ds[var][-24:,:,:].hvplot.quadmesh(x='lon_rho', y='lat_rho', rasterize=True, title=var,
                                    width=600, height=400, crs=crs,
                                    groupby=list(ds[var].dims[:-2]), cmap='jet')
    overlay = (base_map * mesh.opts(alpha=0.7)).opts(active_tools=['wheel_zoom', 'pan'])
    widgets = {dim: pn.widgets.Select for dim in ds[var].dims[:-2]}
    return pn.holoviews.HoloViews(overlay, widgets=widgets).layout

def on_var_select(event):
    var = event.obj.value
    dashboard[-1] = plot(var=var)

var_select.param.watch(on_var_select, parameter_names=['value']);

dashboard = pn.Column(var_select, plot(var_select.value))
```
which produces:
one

However, while `hvplot` defaults to the *first* value of the `options` in the `OrderedDict`, I would like to default to the *last* value.

If I look at the dashboard elements:
```python
    print(dashboard)

    Column
        [0] Select(name='Variables:', options=OrderedDict([('salt', ...]), value='temp')
        [1] Row
            [0] HoloViews(DynamicMap, widgets={'time': <class 'panel.wid...})
            [1] Column
                [0] Select(name='Forecast time f..., options=OrderedDict([('2019-03-09 ...]), value=numpy.datetime64('2019-03-...)
                [1] Select(name='S-coordinate a..., options=OrderedDict([('-0.96875', ...]), value=-0.96875)
```
I see that I can specify the `Select` widget value *post facto* like:
```python
dashboard[1][1][1].value = next(reversed(dashboard[1][1][1].options.values())) 
```
and that indeed works:
two

I would like to make a stand alone dashboard, however, that defaults to a last value when the user selects a new variable.  With the current code, when a new variable is selected, the default value is unfortunately again set to the first value. 

Is there a way to embed my *post facto* code to select the last value into my plot function, or is there another better approach to accomplish this?

The full notebook is here: https://nbviewer.jupyter.org/gist/rsignell-usgs/edb9eb572058c4fdee6708359d24539d

philippjfr:
The ``onlychanged=True`` check in param doesn't appear to be working at least in some cases. So using ``redim`` in certain cases drops Dimension parameters on clone. It's also not clear to me why ``onlychanged`` was there in the first place, there shouldn't be any harm in passing all of them along.

Fixes #3561

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -325,7 +325,7 @@ class Dimension(param.Parameterized):
         Returns:
             Cloned Dimension object
         """
-        settings = dict(self.get_param_values(onlychanged=True), **overrides)
+        settings = dict(self.get_param_values(), **overrides)
 
         if spec is None:
             spec = (self.name, overrides.get('label', self.label))
@@ -337,7 +337,6 @@ class Dimension(param.Parameterized):
                     'Using label as supplied by keyword ({!r}), ignoring '
                     'tuple value {!r}'.format(overrides['label'], spec[1]))
             spec = (spec[0],  overrides['label'])
-
         return self.__class__(spec, **{k:v for k,v in settings.items()
                                        if k not in ['name', 'label']})
 


jlstevens:
I'm ok with this change but it sounds like the real fix would be to address this in param. I don't see why it should only be the changed params either so I'll go ahead and merge.

PR #3581, opened 2019-03-26
add a environment variable HOLOVIEWS_INLINE_JS

basnijholt:
Like @philippjfr suggested on chat.

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -83,7 +83,9 @@ class notebook_extension(extension):
 
     logo = param.Boolean(default=True, doc="Toggles display of HoloViews logo")
 
-    inline = param.Boolean(default=True, doc="""
+    inline = param.Boolean(
+        default=os.environ.get('HOLOVIEWS_INLINE_JS', 'True').lower() == 'true',
+        doc="""
         Whether to inline JS and CSS resources. 
         If disabled, resources are loaded from CDN if one is available.""")
 


holoviews/ipython/__init__.py
```diff
@@ -83,7 +83,9 @@ class notebook_extension(extension):
 
     logo = param.Boolean(default=True, doc="Toggles display of HoloViews logo")
 
-    inline = param.Boolean(default=True, doc="""
+    inline = param.Boolean(
+        default=os.environ.get('HOLOVIEWS_INLINE_JS', 'True').lower() == 'true',
```
philippjfr: I don't quite follow how this would work: ``'true' != True``

philippjfr:
We've discussed making hv.config configurable via environment variables, so I think I'd prefer going ahead with that plan and moving this option onto ``hv.config``. @jlstevens Any opinions?

basnijholt:
all of the following would result in `inline=True`
```
export HOLOVIEWS_INLINE_JS="TRUE"
export HOLOVIEWS_INLINE_JS="true"
export HOLOVIEWS_INLINE_JS="True"
```
otherwise this would be `False`.

jlstevens:
I agree with @philippjfr - this should be controlled through `hv.config` and that would be the most sensible thing to control via environment variables.

philippjfr:
Need to decide how that will work. In panel I've used parameters and property pairs to make config options that are also controllable by environment variables, but I'm also okay if we take the simpler approach of simply setting the value on import like in this PR is fine too.

philippjfr:
This is no longer needed, we have switched back to inlining JS by default.

PR #3586, opened 2019-03-28
Unpin IPython and disable jedi completion by default

philippjfr:
Not supporting newer versions of IPython is really not feasible in the long run, so I've made the holoviews extension disable jedi tab-completion in IPython, if someone really wants jedi completion they can toggle it with ``hv.extension(allow_jedi_completion=True)``.

Fixes https://github.com/pyviz/holoviews/issues/3540

@jlstevens @jbednar Thoughts on this appreciated, users were getting offered old versions of holoviews by conda because conda prefers to install an old version of a new package rather than downgrade the version of an existing package.

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -100,6 +100,11 @@ class notebook_extension(extension):
         using the matplotlib backend) may be used. This may be useful to
         export figures to other formats such as PDF with nbconvert. """)
 
+    allow_jedi_completion = param.Boolean(default=False, doc="""
+       Whether to allow jedi tab-completion to be enabled in IPython.
+       Disabled by default because many HoloViews features rely on
+       tab-completion machinery not supported when using jedi.""")
+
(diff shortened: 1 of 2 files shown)


jlstevens:
Seems reasonable. Of course, I would prefer it if IPython weren't always broken.

philippjfr:
>I assume there is no prospect of this being fixed at that end...

It's a jedi issue I imagine so should probably be fixed there. It's certainly not getting fixed before hv 1.12.0 though, so for the time being I think this is the best we can do.

jlstevens:
Ok. I'll merge when the tests pass.

philippjfr:
@jlstevens Ready to merge.

stonebig:
new release is still planned for this week ?

philippjfr:
>new release is still planned for this week ?

@stonebig Hoping that's the case, waiting on bokeh though and I don't have direct control over that. It's certainly getting very close though.

PR #3593, opened 2019-04-02
Remove hardcoding of lightgray fill color to box and violin

ahuang11:
Setting it to Cycle() now makes overlaying color cycle as expected
image

Reverses https://github.com/pyviz/holoviews/pull/2975, but the cycling of these color is redundant of the x-axis 
image


jlstevens:
I'm happy to change things to use more sensible defaults - my only worry is this can cause backwards incompatible visual changes...

philippjfr:
>I'm happy to change things to use more sensible defaults - my only worry is this can cause backwards incompatible visual changes...

I'm okay considering the previous behavior a bug, since all elements should cycle.

PR #3602, opened 2019-04-03
Documented Glasbey colormaps

jbednar:


philippjfr:
I cancelled the builds since this is docs only, can I merge this?

jbednar:
Please do! Note that it does change some lines of code in holoviews/plotting/util.py, but it's hidden by a try/except and therefore very unlikely to break the build directly.  I also tested it locally with Colormaps.ipynb and it works as expected.

PR #3612, opened 2019-04-09
Allow ability to define style cycles for Draw tools 

philippjfr:
This PR allows adding style cycles to the PolyDraw stream (but this could be expanded to the other draw streams). As an example here we cycle over the fill_color and line_width of the drawn polygons:

```python
poly = hv.Polygons([[(2, 2), (5, 8), (8, 2)]])

poly_stream = streams.PolyDraw(source=poly, drag=True, show_vertices=True,
                               style_cycles={'fill_color': ['red', 'green', 'blue'],
                                             'line_width': [3, 5, 10]})

poly
```

style_cycles

- [x] Fixes https://github.com/pyviz/holoviews/issues/2694
- [x] Expand to BoxEdit and FreehandDraw tools
- [x] Add example in documentation

### DIFF
--- a/examples/reference/streams/bokeh/BoxEdit.ipynb
+++ b/examples/reference/streams/bokeh/BoxEdit.ipynb
@@ -42,7 +42,13 @@
     "\n",
     "**Delete box**\n",
     "\n",
-    "    Tap a box to select it then press BACKSPACE key while the mouse is within the plot area."
+    "    Tap a box to select it then press BACKSPACE key while the mouse is within the plot area.\n",
+    "    \n",
+    "### Properties\n",
+    "\n",
+    "* **``empty_value``**: Value to add to non-coordinate columns when adding new box\n",
+    "* **``num_objects``** (int): Maximum number of boxes to draw before deleting the oldest object\n",
+    "* **``styles``** (dict): Dictionary of style properties (e.g. line_color, line_width etc.) to apply to each box. If values are lists the values will cycle over the values."
    ]
   },
   {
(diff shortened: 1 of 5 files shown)


jbednar:
Looks great! Does it need to be `style_cycles`, or could it just be `styles`, which are cycled if they are cycles and otherwise just applied?  Just brainstorming; not sure which approach is better.

philippjfr:
>Looks great! Does it need to be style_cycles, or could it just be styles, which are cycled if they are cycles and otherwise just applied? Just brainstorming; not sure which approach is better.

It could just be styles I suppose, that is somewhat redundant since you can just set the style in the usual way but may be cleaner regardless.

philippjfr:
@jbednar @jlstevens Let me know which you prefer, I'd be happy with accepting both lists (as cycles) and scalar values, and then call it simply ``styles``.

jbednar:
I'd vote for `styles`.  Actually I vote for just respecting the usual style options set in the usual way and cycling if those are cycles, but I assume that's not feasible?

jlstevens:
I also vote for keeping the API as consistent as possible. Why can't `PolyDraw` support the common style options for polygons which can then be set with styles or constants?

philippjfr:
>Actually I vote for just respecting the usual style options set in the usual way and cycling if those are cycles, but I assume that's not feasible?

That's not consistent with the way cycles behave in HoloViews, styles cycle in an overlay but not within a single Polygons/Paths element.

>Why can't PolyDraw support the common style options for polygons which can then be set with styles or constants?

I'm not following what your suggesting here, that's what the ``styles`` suggestion is no?

jbednar:
> styles cycle in an overlay but not within a single Polygons/Paths element.

Ah, that's the key issue here.  Worth brainstorming if there is a way more general than this approach to achieve within-Element cycling.

philippjfr:
>if there is a way more general than this approach to achieve within-Element cycling.

That can now be done via `dim` style mappings but I don't see any way those could feasibly be implemented client-side.

philippjfr:
I'm going to go ahead and rename it to styles and support scalar values. Respecting cycles within an element would be highly inconsistent so I vote against that.

PR #3616, opened 2019-04-10
Add support for passing in parameter instances as streams

philippjfr:
Makes it possible to pass in a Parameter instance as a stream, which will internally be converted to a Params stream. Also adds some validation to ensure the same parameter is not subscribed to multiple times.

- [x] Adds tests

### DIFF
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -166,30 +166,41 @@ class Stream(param.Parameterized):
     def _on_trigger(self):
         """Called when a stream has been triggered"""
 
+
     @classmethod
     def _process_streams(cls, streams):
         """
         Processes a list of streams promoting Parameterized objects and
         methods to Param based streams.
         """
-        param_watch_support = util.param_version >= '1.8.0'
-        parameterizeds = [s.parameterized for s in streams if isinstance(s, Params)]
+        parameterizeds = defaultdict(list)
         valid, invalid = [], []
         for s in streams:
-            if not isinstance(s, Stream):
-                if isinstance(s, param.Parameterized) and param_watch_support:
-                    if s not in parameterizeds:
-                        s = Params(s)
-                    else:
-                        continue
-                elif util.is_param_method(s) and param_watch_support:
-                    if not hasattr(s, "_dinfo") or util.get_method_owner(s) in parameterizeds:
-                        continue
-                    else:
-                        s = ParamMethod(s)
-                else:
-                    invalid.append(s)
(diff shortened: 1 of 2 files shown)


holoviews/streams.py
```diff
@@ -166,30 +166,41 @@ def trigger(cls, streams):
...
+                pid = id(s.parameterized)
+                if pid in parameterizeds:
+                    overlap = (set(s.parameters) & set(parameterizeds[pid]))
+                    if overlap:
+                        pname = type(s.parameterized).__name__
+                        raise ValueError('Found multiple Params streams '
+                                         'subscribing to the %s parameter(s) '
+                                         'on the %s object. Ensure that '
+                                         'you only subscribe to a parameter '
+                                         'once.' % (list(overlap), pname))
```
jbednar: What would be the effect of simply ignoring the extra subscribers to a parameter?  Would that be safe? If so, could this error simply be skipped?  If not, maybe make a note about why that wouldn't work?

jlstevens:
Looks good. I suppose we don't have any ideas on how to remap parameter names?

One thought I had (probably not a good idea) would be to allow something like:

```
hv.DynamicMap(..., streams={'foo':bar.param.baz})
```

This isn't great as you can't really mix parameter instances and streams nicely. Alternatively:

```
hv.DynamicMap(..., streams=[position_stream, ('foo', bar.param.baz)])
```

But I'm not hot on introducing more tuple formats.

philippjfr:
>Looks good. I suppose we don't have any ideas on how to remap parameter names?

Yeah, it's unfortunate, but I don't want to rush things and come up with something we will hate later so I think this a safe first cut.

holoviews/streams.py
philippjfr: I suppose it's safe enough except for the fact that an event may get triggered multiple times. @jlstevens Any opinion here?

jlstevens:
Looks good! Much happier with the new warning.

jlstevens:
Tests are green. Merging.

holoviews/streams.py
jbednar: By ignoring I guess I didn't just mean skipping it over, I meant to remove all but one of the subscribers for that parameter.  I guess that's not feasible?

philippjfr: No you'd potentially be modifying user defined objects, which would be weird.

jbednar: Ok, sounds like a warning is best then.

PR #3636, opened 2019-04-18
HexTiles aggregator with multiple input vdims

poplarShift:
When exploring spatial data I often find myself regionalizing them before looking for e.g. temporal relationships, but currently HexTiles only permit aggregating column by column. In order to quickly look for regionally aggregated relationships, one would have to do an intermediate step of calculating that relationship after the binning into hex tiles, but before the final aggregation.

This is more of a draft, relying on pandas for the grouping, but generalizing shouldn't be a problem. I also wanted to leave the `.aggregate` method of individual elements/data interfaces alone because that would move it away from how the term is used by pandas and xarray.

Let me know what you think.

Screen Shot 2019-04-18 at 5 50 43 PM

### DIFF
--- a/holoviews/plotting/bokeh/hex_tiles.py
+++ b/holoviews/plotting/bokeh/hex_tiles.py
@@ -1,4 +1,5 @@
 from __future__ import absolute_import, division, unicode_literals
+from inspect import signature
 
 import param
 import numpy as np
@@ -25,8 +26,9 @@ class hex_binning(Operation):
     Should not be user facing as the returned element is not directly
     useable.
     """
-
     aggregator = param.Callable(default=np.size)
+
+    aggregate_input_vdims = param.List(default=[])
 
     gridsize = param.ClassSelector(default=50, class_=(int, tuple))
 
@@ -70,7 +72,7 @@ class hex_binning(Operation):
             vdims = ['Count']
         elif not element.vdims:
             raise ValueError('HexTiles aggregated by value must '
-                             'define a value dimensions.')
+                             'define a value dimension.')
         else:
             vdims = element.vdims
             values = tuple(element.dimension_values(vdim) for vdim in vdims)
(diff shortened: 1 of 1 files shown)


philippjfr:
IMO it should aggregate over all vdims in the input by default.

poplarShift:
You are suggesting to change `aggregate_input_vdims` so it defaults to passing all vdims into `aggregator`?

If I've understood current behaviour correctly, what it currently does by default is aggregating over all kdims, calculating aggregated values vdim by vdim. It does not allow mixing vdims in the aggregation function itself.

philippjfr:
I think I've probably misunderstood what you were trying to implement here. Ideally I would hope this could be expressed using `dim` transforms. Haven't worked out the spelling yet but if they don't currently allow it we should think about extending them in some way to allow it. I think as a rough spelling could relatively easily be made to work:

```
dim('time', regression, dim('air'))
``

but it's not very obvious.

**Comment by poplarShift** [comment]
Your formatting came out slightly off. To which parameter would you then pass `dim('time', regression, dim('air'))`?

I think the most intuitive API would be something like this:

```
opts.HexTiles(aggregator=regression(dim('z'), dim('w')))
```

which would also resemble the datashader API (`aggregator=dsh.mean('some_var')`) more closely.

In that case, however, these two currently permitted uses would have to be reconciled:

```
hv.Points(np.random.rand(100,3), ['x', 'y'], 'z').opts(opts.Points(color=dim('z')))

hv.HexTiles(np.random.rand(100,3), ['x', 'y'], 'z').opts(opts.HexTiles(aggregator=np.median))
```

by also allowing something like `aggregator=np.median(dim('some_var'))`.

**Comment by philippjfr** [comment]
I don't think there's any way to feasibly support:

```
regression(dim('z'), dim('w'))
```

This can work for numpy ufunc's because they have a special protocol but custom user functions will never have that.

poplarShift:
I see. I still haven't understood though where you would pass `dim('time', regression, dim('air'))` though. If it's into `aggregator`, that contradicts current behaviour, and if it's into something like the `current_input_vdims`, you end up passing the custom user function twice.

philippjfr:
>that contradicts current behaviour

I'm not following this, why does it contradict current behavior?

poplarShift:
The currently supported aggregation over one column only takes the function as the argument, not the `dim()`.

philippjfr:
If we accepted dim transforms like the horrible ``dim('time', regression, dim('air'))``, then ``np.median(dim('some_var'))`` would indeed be automatically supported too.

poplarShift:
It seems I chose my example badly. From what you said, if we supported `aggregator=dim('time', regression, dim('air'))`, then `aggregator=np.median(dim('some_var'))` would be supported, sure, but a) `aggregator=regression(dim('z'), dim('w'))` wouldn't because not a ufunc as you said,  and b) old behaviour is `aggregator=np.nanmedian`, not `nanmedian(dim('...'))`.

Anyway if the function has to be the second positional argument to `dim`, then I'm a bit at a loss of how to make the syntax clearer. From `dim('time', regression, dim('air'))` it is not at all clear which dimension gets piped in at which position in the call to `regression`. On the other hand, I'm not sure what would be so terrible about `opts.HexTiles(aggregator=regression, aggregator_args=[dim('a'), dim('b')])`, where `aggregator_args` defaults to the first vdim.

philippjfr:
I think we can probably fiddle with the `dim` API a bit to make things clearer, I don't expect many or any users would have constructed complex dim transforms like that. That said I can't really think of any spelling that would be particularly intuitive.

> and b) old behaviour is aggregator=np.nanmedian, not nanmedian(dim('...')).

This doesn't seem like a problem to me, as I said above it would simply default to using the first value dimension by default.

poplarShift:
Ok, then how about exposing it through a class method `dim.apply_func(regression, ['time', 'air'])` that internally translates to `dim('time', regression, dim('air'))`?

poplarShift:
Closing this for #4080

PR #3639, opened 2019-04-19
Fix missing data coercion for multipaths

poplarShift:
In some cases, when the dict interface has to convert e.g. an ndarray to a dict, this was not reflected in the data accessed by the element, leading to the interface failing to understand the data at a later point. The coercion is now explicit for all sub data types.

Required for #3640

### DIFF
--- a/holoviews/core/data/multipath.py
+++ b/holoviews/core/data/multipath.py
@@ -76,7 +76,7 @@ class MultiInterface(Interface):
 
 
     @classmethod
-    def _inner_dataset_template(cls, dataset, i):
+    def _inner_dataset_template(cls, dataset, i=0):
         """
         Returns a Dataset template used as a wrapper around the data
         contained within the multi-interface dataset.
--- a/holoviews/core/data/multipath.py
+++ b/holoviews/core/data/multipath.py
@@ -59,9 +59,9 @@ class MultiInterface(Interface):
             return
 
         from holoviews.element import Polygons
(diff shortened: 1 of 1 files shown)


philippjfr:
This was done for efficiency reasons, there should be no scenarios where this coercion is necessary. If there is that is a bug and should be fixed rather than paying the hugely huge penalty of sending hundreds or thousands of paths through the Interface constructor logic.

poplarShift:
You're right, I was underestimating that.

The problem that I had hit (there may be other ones) is here:
https://github.com/pyviz/holoviews/blob/9bf1270414d426b1194d1dfc58ff3c6975f26d53/holoviews/core/data/dictionary.py#L53 (lines 53-59)
in that the dict interface explicitly casts ndarrays to a dict, meaning that the coercion only happens for the template, and (IIRC) consequently the constructed Dataset ends up having ndarray data (https://github.com/pyviz/holoviews/blob/9bf1270414d426b1194d1dfc58ff3c6975f26d53/holoviews/core/data/multipath.py#L64) but being registered with the dict interface

philippjfr:
Ah, I know what the issue is, since the Bounds constructor is creating an array containing datetimes and floats, which the ArrayInterface does not allow the template falls back to a dictionary and errors. Fixed in https://github.com/pyviz/holoviews/pull/3662/commits/474fa2b180de3f675946da3bc3ed96462d89ac95.

I'll close this PR.

poplarShift:
Neat, thanks!

PR #3640, opened 2019-04-19
Allow nonnumeric tuples in Bounds

poplarShift:
Requires #3639 

The matplotlib backend still struggles (requiring floats), but the bokeh backend can now draw `Bounds` elements with datetimes or categorical axes too.

### DIFF
--- a/holoviews/element/path.py
+++ b/holoviews/element/path.py
@@ -476,7 +476,7 @@ class Bounds(BaseShape):
     used to compute the corresponding lbrt tuple.
     """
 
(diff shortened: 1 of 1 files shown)


philippjfr:
>The matplotlib backend still struggles (requiring floats), but the bokeh backend can now draw Bounds elements with datetimes or categorical axes too.

I don't quite recall where that transform lives but I think you should be able to use matplotlib's date2num utility to transform the dates to floats which can be displayed.

poplarShift:
Yes, though that would still preclude categorical axes. Though I'm not sure how much sense "Bounds" makes for categorical axes anyway.

philippjfr:
Superseded by #3662

PR #3644, opened 2019-04-22
Plotly 3.8 support

jonmmease:
Closes https://github.com/pyviz/holoviews/issues/3643

Two changes:
 - Change `plotly.offline.offline.utils` to `plotly.utils `
 - Only add legend options to traces that support them.  Plotly.js cleaned up the schema a bit to remove these entries from traces that don't use them, which resulted in some validation errors here.

These changes are backward compatible with plotly.py <3.8 so no need to change version constraints.


jbednar:
Thanks!

philippjfr:
Looks great, thanks!

PR #3648, opened 2019-04-23
De-duplicating stream contents by value

jlstevens:
If you change [this warning in streams.py](https://github.com/pyviz/holoviews/blob/master/holoviews/streams.py#L143) into a print (e.g `print('Parameter value clashes %s' % clashes)`) you will see that the warning is supposed to be issued for this example:

image

### Code example

```python
import holoviews as hv
import dask.dataframe as dd

hv.extension('bokeh')
```

```python
path = '../../datashader/examples/data/nyc_taxi_wide.parq'
df = dd.read_parquet(path).persist()

points = hv.Points(df, ['dropoff_x', 'dropoff_y'], ['dropoff_hour','pickup_hour'])
options = hv.opts.Image( data_aspect=1, responsive=True, logz=True,
                      tools=['hover'], colorbar=True)
rasterize(points).options(options)
```

On current master, that warning doesn't appear:

image

The fact the warning is swallowed is one problem and the fact that two sets of `height`, `width` and `scale` are being issued on initialization might be another. That said, the height, width and scale values being sent are consistent duplicates. This PR now only issues a clash warning if the keys clash *and* the values are different. Although we should also avoid sending duplicate stream contents, I think such a change makes sense regardless.

### DIFF
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -147,7 +147,7 @@ class Stream(param.Parameterized):
                 if len(values) > 1:
                     clashes.append((clash, values))
             if clashes:
-                msg = ', '.join(['%r has values %r' % (k,v) for k,v in clashes ])
+                msg = ', '.join(['%r has values %r' % (k, v) for k, v in clashes])
                 print('Parameter value clashes where %s' % msg )
 
         # Group subscribers by precedence while keeping the ordering
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -135,7 +135,7 @@ class Stream(param.Parameterized):
         needs to be called once.
         """
         # Union of stream contents
-        items = [stream.contents.items() for stream in streams]
+        items = [stream.contents.items() for stream in set(streams)]
         union = [kv for kvs in items for kv in kvs]
         klist = [k for k, _ in union]
         key_clashes = set([k for k in klist if klist.count(k) > 1])
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -138,9 +138,17 @@ class Stream(param.Parameterized):
         items = [stream.contents.items() for stream in streams]
         union = [kv for kvs in items for kv in kvs]
         klist = [k for k, _ in union]
-        clashes = set([k for k in klist if klist.count(k) > 1])
-        if clashes:
-            param.main.param.warning('Parameter name clashes for keys: %r' % clashes)
+        key_clashes = set([k for k in klist if klist.count(k) > 1])
+        if key_clashes:
+            clashes = []
+            dicts = [dict(kvs) for kvs in items]
+            for clash in key_clashes:
+                values = set(d[clash] for d in dicts if clash in d)
+                if len(values) > 1:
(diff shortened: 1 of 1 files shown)


holoviews/streams.py
```diff
@@ -138,9 +138,17 @@ def trigger(cls, streams):
...
+        key_clashes = set([k for k in klist if klist.count(k) > 1])
+        if key_clashes:
+            clashes = []
+            dicts = [dict(kvs) for kvs in items]
+            for clash in key_clashes:
+                values = set(d[clash] for d in dicts if clash in d)
+                if len(values) > 1:
+                    clashes.append((clash, values))
+            if clashes:
+                msg = ', '.join(['%r has values %r' % (k,v) for k,v in clashes ])
```
philippjfr:
```suggestion
                msg = ', '.join(['%r has values %r' % (k, v) for k, v in clashes])
```

holoviews/streams.py
```diff
@@ -138,9 +138,17 @@ def trigger(cls, streams):
...
+                print('Parameter value clashes where %s' % msg )
```
philippjfr:
```suggestion
                print('Parameter value clashes where %s' % msg)
```

philippjfr:
Makes sense to me, minor code formatting suggestions, otherwise happy to merge.

philippjfr:
Since you're out today I've decided to apply them myself and merge.

jlstevens:
Thanks! I also think this change makes sense regardless though I don't want us to forget about the other two issues here:

- Logging output not appearing in some circumstances.
- Why are there two sets of height/width/scale information appearing at once on initialization anyway?

Any thoughts on those mysteries?

philippjfr:
Hmm, actually the most obvious fix I would have made here is to de-duplicate the streams themselves before you deduplicate their stream values. I assume something is calling ``Stream.trigger([stream, stream])`` or similar (i.e. the same stream instance twice). So I'd prefer if you deduplicated the streams themselves rather than the keywords (although I won't object if you want to do both).

philippjfr:
>Logging output not appearing in some circumstances.

There are separate issues for this and I currently have no idea.

jlstevens:
> I assume something is calling Stream.trigger([stream, stream]) or similar (i.e. the same stream instance twice). So I'd prefer if you deduplicated the streams themselves rather than the keywords...

If this is indeed a duplication in a single `Stream.trigger` call then I agree. What I'll do is see if deduplicating at the `trigger` level fixes this independently in which case I think having both fixes together is best.

philippjfr:
>in which case I think having both fixes together is best.

I agree.

holoviews/streams.py
```diff
@@ -138,9 +138,17 @@ def trigger(cls, streams):
         items = [stream.contents.items() for stream in streams]
```
philippjfr:
```suggestion
        items = [stream.contents.items() for stream in set(streams)]
```

jlstevens:
I think this can now be merged once the tests pass...

PR #3652, opened 2019-04-25
One _box_stats function to use

ahuang11:
Part 1 of https://github.com/pyviz/holoviews/issues/3627

Not sure if I need to put it in utils since matplotlib seems to just use the built in boxwhisker/violin?

replicates https://holoviews.org/reference/elements/bokeh/BoxWhisker.html and https://holoviews.org/reference/elements/bokeh/Violin.html (excluding the randomness)
image
image


philippjfr:
Looks good to me.

Issue #3485, opened 2019-02-11
Using bokeh's Arrow to draw VectorField?


poplarShift:
So I found the drawing of the vectorfield arrows in the bokeh backend to be a bit awkward (mostly in looks and lack of styling options) and tried to replace it by bokeh's Arrow through a CompositeElementPlot. Problem is (which is maybe why you didn't do that in the first place...) that bokeh's `Figure` does not expose Arrow, which is an `annotation`, not a `marker` or `glyph` (if I understand correctly), and hence drawing arrows in holoviews has to pass by style options, not data (again, if my understanding is correct).

Question is, do you think that could be addressed on the Holoviews side, or would one have to make a PR on the bokeh side? I briefly skimmed some related discussions on here from when you were implementing VectorField, but that's a while ago and I couldn't find any information on this in particular.

philippjfr:
The main problem I think is that ``Arrow`` is not vectorized in any way. That means each individual arrow would have to be added as a separate annotation, which is not very scalable. After a few hundred arrows the plot would likely slow to a crawl. That's the main reason it was implemented the way it was, if we are aiming to improve on that it will likely take a PR in bokeh to either extend the Arrow annotation or write a new glyph entirely.

philippjfr:
Actually I think I'm wrong on this and Arrow is actually vectorized, which means we could try this approach.

poplarShift:
I am not sure I understand, how would that be implemented? Specifying `_plot_methods = dict(single='arrow')` in the VectorFieldPlot class just runs into the fact that Arrow isn't exposed in Figure. On the other hand, modifying bokeh's Arrow example passing lists for `x_start`  etc. (`p.add_layout(Arrow(end=OpenHead(line_color="firebrick", line_width=4),
                   x_start=[0,1], y_start=[0,1], x_end=[1,3], y_end=[0,5]))`) doesn't work either for me.

philippjfr:
Here's how to use a vectorized Arrow:

```python
p = figure(x_range=(0, 10), y_range=(0, 10))

cds = ColumnDataSource(data=dict(x_start=[0,1, 2], y_start=[0,1, 2], x_end=[1,3, 5], y_end=[0,5, 8]))
p.add_layout(Arrow(end=OpenHead(), source=cds, x_start='x_start', y_start='y_start', x_end='x_end', y_end='y_end'))

pn.Column(p)
```

poplarShift:
[...]

PR #3653, opened 2019-04-25
Using bokeh's Arrow to draw VectorField

poplarShift:
I realized I never made a PR. Addresses #3485.

This is only a first draft, I'm putting this here for future reference. Needs some tidying up in terms of styling options.

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -4,8 +4,14 @@ from collections import defaultdict
 
 import numpy as np
 import param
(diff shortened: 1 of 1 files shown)


philippjfr:
This is sadly not looking too promising, we'd be losing the ability to do colormapping, the size scaling isn't work very well for me, and various selection/non_selection options aren't working.

philippjfr:
Sadly I'll close this for now. There doesn't seem to be a good way forward atm. I won't delete the branch though.

PR #3658, opened 2019-04-27
add dimensioned streams to DynamicMap __mul__ clone

arabidopsis:
Adds any dimensioned streams to DynamicMap clone operation when overlay '*' operator is
invoked.

Without this, the operation converts a bounded DynamicMap (and therefore displayable) to an
unbounded one (and therefore fails to render).


philippjfr:
Makes sense to me and thanks for the tests!

PR #3688, opened 2019-05-03
Reuse PlotSize stream if it exists

philippjfr:
In order for data_aspect handling to work when a plot is updated dynamically a PlotSize stream is created to keep track of the screen size. However when the plot already has a PlotSize stream attached this is both superfluous and can cause parameter clashes in the Stream.trigger method. This PR simply reuses existing streams.


jlstevens:
Looks good to me and I do prefer it if we don't add (unnecessary) streams behind the scenes. Merging.

PR #3689, opened 2019-05-03
Fixed hv.help() for non-hv objects

jbednar:
hv.help() was meant to fall back to pydoc for anything it doesn't understand, but that has apparently been broken for some time (see below).  Looks like it was because `pydoc` was deleted from the module namespace, presumably to clean up our API.  To achieve the same thing while also working properly, did a local import of `pydoc` just where it was needed.

Screen Shot 2019-05-03 at 12 40 18 PM

### DIFF
--- a/holoviews/__init__.py
+++ b/holoviews/__init__.py
@@ -1,6 +1,6 @@
 
 from __future__ import print_function, absolute_import
-import os, pydoc, io
+import os, io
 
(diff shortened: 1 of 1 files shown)


jlstevens:
Thanks for the fix. Happy to merge when the tests pass (I restarted one, presumably failed due to a transient).

PR #3692, opened 2019-05-05
Add hover for other projs

ahuang11:
Fixes:
https://github.com/pyviz/geoviews/issues/251

image

Linked with:
https://github.com/pyviz/geoviews/pull/321


philippjfr:
Sorry I never got to this. I think we should simply override the method in geoviews rather than introducing projection related code here.

ahuang11:
Okay, I updated https://github.com/pyviz/geoviews/pull/321

PR #3694, opened 2019-05-06
DataLink when NaNs are present

### LINKED ISSUE #3690: DataLink fails if NaN present
When there is a NaN present in the column shared by the two objects passed to DataLink, it fails, claiming `ValueError: DataLink can only be applied if overlapping dimension values are equal`

```
import holoviews as hv
import numpy as np
hv.extension('bokeh')
from holoviews.plotting.links import DataLink

x = np.random.rand(100,3)

# no problem:
x[0,1] = np.nan

# makes it fail:
# x[0,0] = np.nan

df = pd.DataFrame(x, columns=['x', 'y', 'z'])

a = hv.Scatter(df, 'x', 'z')
b = hv.Scatter(df, 'x', 'y')
DataLink(a,b)
l=a+b
l
```

poplarShift:
Fixes #3690


philippjfr:
Perfect thanks (especially for the test)!

Issue #3685, opened 2019-05-02
dask requirement not handled properly (ImportError)


odoublewen:
sometime between releases holoviews releases 1.10.9 and 1.12.2., I started getting this import error if `dask` is in the namespace (in my case, it's being dragged in by [scikit-allel](https://github.com/cggh/scikit-allel)).

```
$ pipenv shell --three
$ pipenv install holoviews dask
$ python -c "import dask; import holoviews as hv; hv.HeatMap({'x':[1], 'y':[1], 'z':[1]})"
Traceback (most recent call last):
  ...
  File "/home/osolberg/temp/holoviews_test/.venv/lib/python3.6/site-packages/holoviews/core/data/interface.py", line 251, in initialize
    raise DataError(error)
holoviews.core.data.interface.DataError: None of the available storage backends were able to support the supplied data format. GridInterface raised following error:

 No module named 'toolz'

Dask array requirements are not installed.

Please either conda or pip install as follows:

  conda install dask                 # either conda install
  pip install dask[array] --upgrade  # or pip install
```

As the message suggests, I can get around the import error by installing  `dask[complete]`

The same command works if I don't import dask, i.e.:

```
python -c "import holoviews as hv; hv.HeatMap({'x':[1], 'y':[1], 'z':[1]})"
```

But this is a problem because I am using scikit-allel which uses dask.

It seems like holoviews doesn't **need** dask, but if holoviews **finds** dask in the namespace, then it will throw an error if it's not  `dask[complete]`

This is Python 3.6.5.  I can post a full Pipfile.lock if needed, but I think this should be pretty easy to reproduce.

philippjfr:
HoloViews does not use toolz directly so this seems like an issue upstream somewhere.

jbednar:
Datashader does, and HoloViews imports Datashader...

philippjfr:
Not by default so that seems unlikely to be at fault.

odoublewen:
The proximal cause is [this line](https://github.com/pyviz/holoviews/blob/master/holoviews/core/data/interface.py#L16)

```
if 'dask' in sys.modules:
        import dask.array as da
```

But it is not enough to test `if 'dask' in sys.modules` because [dask has several install variants](https://docs.dask.org/en/latest/install.html#pip), and the default variant does not provide `dask.array`.

The simplest solution would be to wrap the above import statement in a try-except....

philippjfr:
Thanks @odoublewen, I was not aware of all the install variants.

[...]

PR #3720, opened 2019-05-16
use dask_array_module() to get dask array, if possible

odoublewen:
Offered as a possible solution to #3685


odoublewen:
Closed in favor of #3719 
Also this PR would not work because it should says `array_types += (dask_array.Array, )`

PR #3727, opened 2019-05-20
Ensure RGB hover is handled

philippjfr:


emryslda:
@philippjfr I still have the problem. Can you see the RGB values instead?

philippjfr:
>@philippjfr I still have the problem. Can you see the RGB values instead?

You should see a single integer representing the int32 RGB value.

emryslda:
Right! Thank you very much! 
Sorry if I ask again but could be possible to implement the RGB visualization singularly? Pointing the single value It would be nice to see R G B separately, like for example   R 255 G 25 B 10. 

> > @philippjfr I still have the problem. Can you see the RGB values instead?
> 
> You should see a single integer representing the int32 RGB value.

emryslda:
@philippjfr  Sorry but could be possible to extract the int32 value from the RGB as a matrix of values (with the same x,y of the image i mean)?

cocoaaa:
@emryslda It may be helpful to convert the RGB element to an Image element:
```
%%opts Image [tools=['hover']]
hvimg = hvrgb.to(hv.Image) # hover tools gives x,y,r,g,b values
```
However, I'm not sure why the colormapping is different.. (@philippjfr)
image

cocoaaa:
@philippjfr The extents of the Image and RGB elements are also not correctly set to the limits of `x`, `y` grid.

jbednar:
An hv.Image element has only a single plane of data, and so it's inherently monochromatic, with colors created by colormapping from the single value dimension into RGB space. An hv.RGB element doesn't do any colormapping, having separate R, G, and B channels directly determining the colors seen. So if you convert an RGB element into an Image element, you're losing the three channels, resulting in a single monochromatic channel.  So converting to Image really isn't what you want here.  It would be great to have a PR on bokeh making RGB hovering more useful, though.

PR #3734, opened 2019-05-26
Fix setting edge colors for Chord graphs using an explicit mapping

schumann-tim:
and normalization (which is turned on by default).

Details:
The color map's vmin/vmax was computed based on factors of the whole range
whereas the palette only used the factors from the actually used values.

When using color normalization this lead to:
-- LineCollection normalizing via vmin/vmax computed above, reflecting the
   whole range.
-- ListedColormap is transforming the normalized color values provided by the
   LineCollection using its own size (N). However that size N was based on the
   actually used values and hence too small.

### DIFF
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -907,7 +907,13 @@ class ColorbarPlot(ElementPlot):
 
         if not isinstance(cmap, mpl_colors.Colormap):
             if isinstance(cmap, dict):
-                factors = util.unique_array(values)
+                # The palette needs to correspond to the map's limits (vmin/vmax). So use the same
+                # factors as in the map's clim computation above.
+                range_key = dim_range_key(vdim)
+                if range_key in ranges and 'factors' in ranges[range_key]:
+                    factors = ranges[range_key]['factors']
+                else:
+                    factors = util.unique_array(values)
                 palette = [cmap.get(f, colors.get('NaN', {'color': self._default_nan})['color'])
                            for f in factors]
             else:


philippjfr:
Thanks for the fix, it's very much appreciated.

philippjfr:
Presumably for the bokeh backend this was already handled correctly?

schumann-tim:
> Presumably for the bokeh backend this was already handled correctly?
I didn't try it out with bokeh; when I installed holoviews with conda, apparently a too-old version of bokeh was bundled. As I primarily wanted to generate the graphs as svgs I didn't follow up. I can check though.

Would a minimal example be useful for regression testing? (not sure how the current set-up of the project is).

philippjfr:
>Would a minimal example be useful for regression testing?

Definitely, you can look at the existing tests in holoviews/tests/plotting/matplotlib/testgraphplot.py ``TestMplChordPlot`` for some examples.

schumann-tim:
To keep the conversation going, here's a repro for the bug.
I'm unfortunately traveling without access to my computer. Will pick this up when i'm back in ~2 weeks.

import os
import pandas as pd
import holoviews as hv
from holoviews import opts, dim

hv.extension('matplotlib')

edges = [{'source': 0, 'target': 2, 'value': 168}, 
        {'source': 0, 'target': 3, 'value': 150}, 
        {'source': 1, 'target': 3, 'value': 50},
        {'source': 1, 'target': 2, 'value': 100}]

nodes = [{'index': 0, 'group': 'clonal_id', 'name': 'grey'},
        {'index': 1, 'group': 'clonal_id', 'name': 'red'},
        {'index': 2, 'group': 'hits', 'name': 'blue'},
        {'index': 3, 'group': 'hits', 'name': 'green'}]


# Edges reference by the source's node-id -- so include them as well.
color_map = {'grey':'grey', '0': 'grey',
             'red': 'red', '1': 'red',
             'blue': 'blue', '2': 'blue',
             'green': 'green', '3': 'green'}

links_df = pd.DataFrame(edges)
nodes_ds = hv.Dataset(pd.DataFrame(nodes), 'index')
graph = hv.Chord((links_df, nodes_ds), kdims=['source', 'target']).opts(
        opts.Chord(
            cmap=color_map, edge_cmap=color_map, 
            edge_color=(dim('source')).astype(str), label_index='name',
            node_color=dim('name').str(), normalize=True))

output_path =  '/tmp/repro.pdf'
hv.save(graph, output_path)

philippjfr:
Just going to merge.

PR #3736, opened 2019-05-27
datetime handling for spikes

poplarShift:
The previous implementation of get_data for Spikes in the matplotlib backend used the `element.array()` method, which already coerces datetime values to int in case of mixed data types. I fixed that and updated the datetime handling using recently added utils. 

The following now works, whereas before it didn't:

```
import pandas as pd
import numpy as np
import holoviews as hv
hv.extension('matplotlib')

e = hv.Spikes((pd.date_range('2100-1-1', '2100-10-10', periods=20), np.random.rand(20)), 'Date', 'c')
e.opts(xrotation=45)
```

Screen Shot 2019-05-27 at 3 20 57 PM

### DIFF
--- a/holoviews/plotting/mpl/chart.py
+++ b/holoviews/plotting/mpl/chart.py
@@ -996,7 +996,7 @@ class BarPlot(LegendPlot):
 
     def _create_bars(self, axis, element):
         # Get style and dimension information
-        values = self.values    
+        values = self.values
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks @poplarShift!

jlstevens:
Yes, thank you!

PR #3738, opened 2019-05-28
Fixed and unified links on Overlay reference page

hojo0590:
There was a wrong link on the plotly Overlay page which did not lead to the bokeh Overlay page.
Simultaneously I changed the order on the matplotlib Overlay page so that the links don't jump anymore when viewing the Overlay pages of the different plotting backends.


philippjfr:
Thanks for the fix, very much appreciated!

PR #3744, opened 2019-05-29
Add ability to use functions with param dependencies in DynamicMap

philippjfr:
Allows doing things like:

```python
slider = pn.widgets.FloatSlider()

@pn.depends(y=slider.param.value)
def test(y):
    return hv.Curve([0, 1, y])

pn.Column(slider, hv.DynamicMap(test))
```

- [x] Adds tests
- [x] Added docs

### DIFF
--- a/examples/user_guide/14-Data_Pipelines.ipynb
+++ b/examples/user_guide/14-Data_Pipelines.ipynb
@@ -112,7 +112,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The ``.apply`` method allows us to automatically build a pipeline given an object and some operation or function along with parameter instances passed in as keyword argument. Internally it will then build a Stream to ensure that whenever the parameter changes the plot is updated. To learn more about streams see the [Responding to Events](./12-Responding_to_Events.ipynb).\n",
+    "The ``.apply`` method allows us to automatically build a pipeline given an object and some operation or function along with parameter instances passed in as keyword arguments. Internally it will then build a Stream to ensure that whenever the parameter changes the plot is updated. To learn more about streams see the [Responding to Events](./12-Responding_to_Events.ipynb).\n",
     "\n",
     "This mechanism allows us to build powerful pipelines by linking parameters on a user defined class or even an external widget, e.g. here we import an ``IntSlider`` widget from [``panel``](https://pyviz.panel.org):"
    ]
@@ -132,7 +132,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Using the ``.apply`` method we can now apply the ``rolling`` operation to the DynamicMap and link the ``value`` parameter of the slider to the operations ``rolling_window`` parameter (this also works for simple functions):"
+    "Using the ``.apply`` method we can now apply the ``rolling`` operation to the DynamicMap and link the ``value`` parameter of the slider to the operation's ``rolling_window`` parameter (which also works for simple functions as will be shown below):"
    ]
   },
   {
@@ -150,7 +150,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The ``rolled_dmap`` is another DynamicMap which defines a simple two step pipeline, which calls the original callback when the ``symbol`` changes and reapplies the ``rolling`` operation when the slider value changes. Since the widget's value is now linked to the plot via a ``Stream`` we can display the widget and watch the plot update:"
+    "The ``rolled_dmap`` is another DynamicMap that defines a simple two-step pipeline, which calls the original callback when the ``symbol`` changes and reapplies the ``rolling`` operation when the slider value changes. Since the widget's value is now linked to the plot via a ``Stream`` we can display the widget and watch the plot update:"
    ]
   },
   {
--- a/examples/user_guide/14-Data_Pipelines.ipynb
+++ b/examples/user_guide/14-Data_Pipelines.ipynb
@@ -112,7 +112,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The ``.apply`` method allows us to automatically build a pipeline given an object and some operation or function along with parameter instances passed in as keyword arguments. Internally it will then build a Stream to ensure that whenever the parameter changes the plot is updated. To learn more about streams see the [Responding to Events](./12-Responding_to_Events.ipynb).\n",
+    "The ``.apply`` method allows us to automatically build a pipeline given an object and some operation or function along with parameter instances passed in as keyword argument. Internally it will then build a Stream to ensure that whenever the parameter changes the plot is updated. To learn more about streams see the [Responding to Events](./12-Responding_to_Events.ipynb).\n",
     "\n",
     "This mechanism allows us to build powerful pipelines by linking parameters on a user defined class or even an external widget, e.g. here we import an ``IntSlider`` widget from [``panel``](https://pyviz.panel.org):"
    ]
@@ -132,7 +132,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Using the ``.apply`` method we can now apply the ``rolling`` operation to the DynamicMap and link the ``value`` parameter of the slider to the operations ``rolling_window`` parameter. Using ``.apply`` we can even use simple functions, when using an operation we can also apply the operation directly (as shown in the comment below):"
+    "Using the ``.apply`` method we can now apply the ``rolling`` operation to the DynamicMap and link the ``value`` parameter of the slider to the operations ``rolling_window`` parameter (this also works for simple functions):"
    ]
   },
   {
@@ -142,8 +142,6 @@
    "outputs": [],
    "source": [
     "rolled_dmap = dmap.apply(rolling, rolling_window=slider.param.value)\n",
-    "\n",
-    "# rolled_dmap = rolling(dmap, rolling_window=slider.param.value)\n",
     "\n",
     "rolled_dmap"
    ]
@@ -188,7 +186,53 @@
    "source": [
     "We can chain operations like this indefinitely and attach parameters or explicit streams to each stage. By chaining we can watch our visualization update whenever we change a stream value anywhere in the pipeline and HoloViews will be smart about which parts of the pipeline are recomputed, which allows us to build complex visualizations very quickly.\n",
     "\n",
-    "In later guides we will see how we can combine HoloViews plots and Panel widgets into custom layouts allowing us to define complex dashboards. For more information on how to deploy bokeh apps from HoloViews and build dashboards see the [Deploying Bokeh Apps](./Deploying_Bokeh_Apps.ipynb) and [Dashboards](./17-Dashboards.ipynb) guides."
+    "The ``.apply`` method is also not limited to operations, we can just as easily apply a simple function to each object in the ``DynamicMap``. Here we define a function to compute the residual between the original ``dmap`` and the ``rolled_dmap``."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def residual_fn(overlay):\n",
+    "    # Get first and second Element in overlay\n",
+    "    el1, el2 = overlay.get(0), overlay.get(1)\n",
+    "\n",
+    "    # Get x-values and y-values of curves\n",
+    "    xvals  = el1.dimension_values(0)\n",
+    "    yvals  = el1.dimension_values(1)\n",
+    "    yvals2 = el2.dimension_values(1)\n",
+    "\n",
+    "    # Return new Element with subtracted y-values\n",
+    "    # and new label\n",
+    "    return el1.clone((xvals, yvals-yvals2),\n",
+    "                     vdims='Residual')"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "If we overlay the two DynamicMaps we can then dynamically broadcast this function to each of the overlays, producing a new DynamicMap which responds to both the symbol selector widget and the slider:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "residual = (dmap * rolled_dmap).apply(residual_fn)\n",
+    "\n",
+    "residual"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "In later guides we will see how we can combine HoloViews plots and Panel widgets into custom layouts allowing us to define complex dashboards. For more information on how to deploy bokeh apps from HoloViews and build dashboards see the [Deploying Bokeh Apps](./Deploying_Bokeh_Apps.ipynb) and [Dashboards](./17-Dashboards.ipynb) guides. To get a quick idea of what this might look like let's compose all the components we have no built:"
    ]
   }
  ],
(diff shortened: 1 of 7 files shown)


jlstevens:
Code looks fine but I'm starting to lose track of all the ways you can do things with param watchers, panel etc. I assume the alternative would be a class which is unnecessary for such a simple example?

I guess I am happy to merge but I would also like to see some docs detailing this (along with all the other tricks you can do with param).

philippjfr:
>I guess I am happy to merge but I would also like to see some docs detailing this style among all the other ways you can do things with param nowadays.

That is on my to-do list along with `.apply` docs, but I don't think I'll get there today and in the meantime I'd like to build examples on top of this.

examples/user_guide/14-Data_Pipelines.ipynb
```diff
@@ -138,12 +141,7 @@
...
    "source": [
-    "def rolled_data(symbol, rolling_window, **kwargs):\n",
-    "    curve = load_symbol(symbol)\n",
-    "    return rolling(curve, rolling_window=rolling_window)\n",
-    "    \n",
-    "rolled_dmap = hv.DynamicMap(rolled_data, kdims='Symbol',\n",
-    "                            streams=[stream]).redim.values(Symbol=stock_symbols)\n",
+    "rolled_dmap = dmap.apply(rolling, rolling_window=slider.param.value)\n",
     "\n",
     "rolled_dmap"
```
jlstevens: I would consider saying something about what calls to `.apply` returns - another DynamicMap that nests the original within the supplied function.

examples/user_guide/14-Data_Pipelines.ipynb
```diff
@@ -121,15 +123,16 @@
...
+    "\n",
+    "slider = pn.widgets.IntSlider(name='rolling_window', start=1, end=100, value=50)"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Now we can define a function that both loads the symbol and applies the ``rolling`` operation passing our ``rolling_window`` parameter to the operation:"
+    "Using the ``.apply`` method we can now apply the ``rolling`` operation to the DynamicMap and set the ``value`` parameter of the slider to supply the rolling_window value. Note that as long as the applied function is a HoloViews ``Operation`` we could also apply it directly but for pure functions we have to use ``.apply``."
```
jlstevens: > Note that as long as the applied function is a HoloViews ``Operation`` we could also apply it directly but for pure functions we have to use ``.apply``.

I find the phrasing here a little odd: I figured out what you mean but I had to re-read it a few times. I think the subject of the sentence should be the apply method and what it supports (HoloViews operations..but also regular functions). I don't have a concrete suggestion right now so I am happy to leave this as is if we can't think of some better phrasing...

examples/user_guide/14-Data_Pipelines.ipynb
```diff
@@ -105,14 +105,16 @@
...
-    "### Controlling operations via Streams"
+    "### Dynamically evaluating operations and parameters with ``.apply``"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "In the previous section we briefly mentioned that in addition to regular widgets ``DynamicMap`` also supports streams, which allow us to define custom events our ``DynamicMap`` should subscribe to. To learn more about streams see the [Responding to Events](./12-Responding_to_Events.ipynb). Here we will declare a stream that controls the rolling window:"
+    "The ``.apply`` method allows us to automatically build a pipeline given an object and some operation or function along with parameter instances passed in as keyword argument. Internally it will then build a Stream to ensure that whenever the parameter changes the plot is updated. To learn more about streams see the [Responding to Events](./12-Responding_to_Events.ipynb).\n",
```
philippjfr:
```suggestion
    "The ``.apply`` method allows us to automatically build a pipeline given an object and some operation or function along with parameter instances passed in as keyword arguments. Internally it will then build a Stream to ensure that whenever the parameter changes the plot is updated. To learn more about streams see the [Responding to Events](./12-Responding_to_Events.ipynb).\n",
```

examples/user_guide/14-Data_Pipelines.ipynb
```diff
@@ -177,37 +175,18 @@
...
     "\n",
-    "In later guides we will discover how to tie custom streams to custom widgets letting us easily control the stream values and making it trivial to define complex dashboards. ``paramNB`` is only one widget framework we could use: we could also choose ``paramBokeh`` to make use of bokeh widgets and deploy the dashboard on bokeh server, or we could manually link ``ipywidgets`` to our streams. For more information on how to deploy bokeh apps from HoloViews and build dashboards see the [Deploying Bokeh Apps](./Deploying_Bokeh_Apps.ipynb) and [Dashboards](./17-Dashboards.ipynb) guides."
+    "In later guides we will see how we can combine HoloViews plots and Panel widgets into custom layouts allowing us to define complex dashboards. For more information on how to deploy bokeh apps from HoloViews and build dashboards see the [Deploying Bokeh Apps](./Deploying_Bokeh_Apps.ipynb) and [Dashboards](./17-Dashboards.ipynb) guides."
```
jlstevens: As a general note, I would like to see an `apply` example where the keyword argument is a parameter on an explicitly declared `Stream` object. This doesn't have to be added in this PR (though I think you could make a `RollingWindow` stream example quite easily). One of the implications of `apply` that is worth pointing out is that you can do rebinding of stream parameter names e.g `foo.apply(fn, x=stream_instance.y)`.

examples/user_guide/17-Dashboards.ipynb
```diff
@@ -65,16 +65,16 @@
...
     "\n",
     "class StockExplorer(param.Parameterized):\n",
     "\n",
     "    rolling_window = param.Integer(default=10, bounds=(1, 365))\n",
     "    \n",
     "    symbol = param.ObjectSelector(default='AAPL', objects=stock_symbols)\n",
     "    \n",
-    "    variable = param.ObjectSelector(default='adj_close', objects=[\n",
-    "        'date', 'open', 'high', 'low', 'close', 'volume', 'adj_close'])\n",
+    "    variable = param.ObjectSelector(default='adj_close', objects=variables)\n",
```
jlstevens: In the markdown cell below the class definition (not in the diff) it says:

> To generate a set of widgets to control these parameters we can simply supply the explorer.param accessor...

As a general point, nothing about param has been properly introduced so I doubt that means much to the typical reader. Ideally there would be some great param docs to link to here in order to introduce param properly!

examples/user_guide/17-Dashboards.ipynb
```diff
@@ -117,15 +117,45 @@
...
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "symbol = pn.widgets.RadioButtonGroup(options=stock_symbols)\n",
+    "variable = pn.widgets.Select(options=variables)\n",
+    "window = pn.widgets.IntSlider(name='Rolling Window', value=10, start=1, end=365)\n",
```
jlstevens: `rolling_window` for consistency with the earlier parameter name?

examples/user_guide/17-Dashboards.ipynb
```diff
@@ -117,15 +117,45 @@
...
+    "def load_symbol_cb(symbol, variable):\n",
+    "    return load_symbol(symbol, variable)\n",
+    "\n",
+    "dmap = hv.DynamicMap(load_symbol_cb)\n",
+    "\n",
+    "smoothed = rolling(stock_dmap, rolling_window=window.param.value)\n",
     "\n",
-    "pn.Row(explorer.param, (smoothed * outliers).opts(width=600))"
+    "pn.Row(pn.WidgetBox('## Stock Explorer', symbol, variable, window), smoothed.opts(width=500))"
    ]
```
jlstevens: I am not sure that it is worth the effort to change, but I think that if we were starting this notebook from scratch we would introduce the conceptually simpler function approach *first* and then talk about classes and why you want to use them. Even if we decide to leave things in this order, I think a paragraph is necessary here to explain why you might want to use classes at all. Any thoughts @jbednar as to whether it is worth reworking the order in this PR?

examples/user_guide/17-Dashboards.ipynb
philippjfr: No the other one is a parameter name, this is effectively the label. Somewhat regret using name for this in panel widgets but stuck with it now.

jlstevens:
@philippjfr Looks good! 

My comments generally relate to expanding the docs in a few places. There is one question about reordering the presentation of the function/class approach but I have no opinion as to whether that should be part of this PR or not.

Another task that shouldn't be part of this PR is to expand the streams user guide to discuss these implicit parameter streams. There are now three types of stream in holoviews imho 1) explicitly defined custom stream objects and classes 2) linked streams 3) implicit parameter driven ones.

Happy to merge once these doc issues are addressed (or postponed for later PRs).

examples/user_guide/17-Dashboards.ipynb
jlstevens: Not quite sure what you mean. I'm a little confused as the handles `symbol` and `variable` here do match the `symbol` and `variable` parameters earlier in the class definition. Why does `window` need to be inconsistent?

examples/user_guide/14-Data_Pipelines.ipynb
philippjfr: >As a general note, I would like to see an apply example where the keyword argument is a parameter on an explicitly declared Stream object. This doesn't have to be added in this PR (though I think you could make a RollingWindow stream example quite easily)

I kind of don't see the point. Why wouldn't you just use an ordinary Parameterized? It's more explicit and imo conceptually simpler. Tbh at this point I'd personally never declare any non-linked stream class anymore, whether to reflect that in the documentation is a different question of course.

examples/user_guide/17-Dashboards.ipynb
philippjfr: @jbednar How are param docs coming along? 😄

examples/user_guide/17-Dashboards.ipynb
philippjfr: Ah, you mean the variable name, not the ``name`` parameter.

examples/user_guide/17-Dashboards.ipynb
philippjfr: This sentence above was my attempt at explaining the difference:

>This approach makes the link between the widgets and the computation very explicit at the cost of tying the widget and display code very closely together.

You're right that it probably deserves a larger discussion though.

examples/user_guide/14-Data_Pipelines.ipynb
jlstevens: > Tbh at this point I'd personally never declare any non-linked stream class anymore...

I think I can buy that argument in which case the streams documentation would have to talk about the parameter based approach instead, describing linked streams and not mention the explicitly declared custom stream classes.

examples/user_guide/17-Dashboards.ipynb
jbednar: Almost done; just a couple of steps more to do:

1. Start
2. Finish

examples/user_guide/17-Dashboards.ipynb
jbednar: I agree with reordering it, but that doesn't have to hold up this PR.

examples/user_guide/14-Data_Pipelines.ipynb
```diff
@@ -121,15 +123,16 @@
...
+    "\n",
+    "slider = pn.widgets.IntSlider(name='rolling_window', start=1, end=100, value=50)"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Now we can define a function that both loads the symbol and applies the ``rolling`` operation passing our ``rolling_window`` parameter to the operation:"
+    "Using the ``.apply`` method we can now apply the ``rolling`` operation to the DynamicMap and link the ``value`` parameter of the slider to the operations ``rolling_window`` parameter. Using ``.apply`` we can even use simple functions, when using an operation we can also apply the operation directly (as shown in the comment below):"
```
jbednar: I cannot parse the last sentence here; it seems to have two unrelated ideas.

examples/user_guide/14-Data_Pipelines.ipynb
jbednar:
```suggestion
    "The ``.apply`` method allows us to automatically build a pipeline given an object and some operation or function along with parameter instances passed in as keyword arguments. Internally it will then build a Stream to ensure that whenever the parameter changes the plot is updated. To learn more about streams see the [Responding to Events](./12-Responding_to_Events.ipynb).\n",
```

(I see that this was a deliberate change in the previous commit, but it doesn't *seem* correct -- should it be "a keyword argument" (doesn't seem to match) or "keyword arguments"?

examples/user_guide/14-Data_Pipelines.ipynb
```diff
@@ -121,15 +123,16 @@
...
+    "\n",
+    "slider = pn.widgets.IntSlider(name='rolling_window', start=1, end=100, value=50)"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Now we can define a function that both loads the symbol and applies the ``rolling`` operation passing our ``rolling_window`` parameter to the operation:"
+    "Using the ``.apply`` method we can now apply the ``rolling`` operation to the DynamicMap and link the ``value`` parameter of the slider to the operations ``rolling_window`` parameter (this also works for simple functions):"
```
jbednar:
```suggestion
    "Using the ``.apply`` method we can now apply the ``rolling`` operation to the DynamicMap and link the ``value`` parameter of the slider to the operation's ``rolling_window`` parameter (which also works for simple functions as will be shown below):"
```

examples/user_guide/14-Data_Pipelines.ipynb
```diff
@@ -152,7 +150,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Since we have a handle on the ``Stream`` we can now send events to it and watch the plot above update, let's start by setting the ``rolling_window=50``. "
+    "The ``rolled_dmap`` is another DynamicMap which defines a simple two step pipeline, which calls the original callback when the ``symbol`` changes and reapplies the ``rolling`` operation when the slider value changes. Since the widget's value is now linked to the plot via a ``Stream`` we can display the widget and watch the plot update:"
```
jbednar:
```suggestion
    "The ``rolled_dmap`` is another DynamicMap that defines a simple two-step pipeline, which calls the original callback when the ``symbol`` changes and reapplies the ``rolling`` operation when the slider value changes. Since the widget's value is now linked to the plot via a ``Stream`` we can display the widget and watch the plot update:"
```

examples/user_guide/14-Data_Pipelines.ipynb
```diff
@@ -177,19 +175,18 @@
...
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Since the ``rolling_stream`` instance we created is bound to both operations, triggering an event on the stream will trigger both the ``Curve`` and the ``Scatter`` of outliers to be updated:"
+    "We can chain operations like this indefinitely and attach parameters or explicit streams to each stage. By chaining we can watch our visualization update whenever we change a stream value anywhere in the pipeline and HoloViews will be smart about which parts of the pipeline are recomputed, which allows us to build complex visualizations very quickly.\n",
+    "\n",
+    "The ``.apply`` method is also not limited to operations, we can just as easily apply a simple function to each object in the ``DynamicMap``. Here we define a function to compute the residual between the original ``dmap`` and the ``rolled_dmap``."
```
jbednar:
```suggestion
    "The ``.apply`` method is also not limited to operations. We can just as easily apply a simple Python function to each object in the ``DynamicMap``. Here we define a function to compute the residual between the original ``dmap`` and the ``rolled_dmap``."
```

jbednar:
Looks good now, thanks!

PR #3755, opened 2019-06-04
Updated definition of box plot whiskers to match convention for Bokeh plots.

justinbois:
In #1926, I suggested using a definition of the whiskers length in box plots as follows:

The top whisker extends 1.5 times the IQR beyond the top of the box or to the maximal data point, whichever is smaller. An analogous definition holds for the bottom whisker. This is how the whisker is defined in some places, e.g., [here](https://www.stat.berkeley.edu/~brill/Papers/EDASage.pdf).

A vastly more widely adapted version uses this definition:
The top whisker extends to the maximum of the set of data points that are less than 1.5 times the IQR beyond the top of the box, with an analogous definition for the lower whisker.

This is what the [HoloViews docs say](http://holoviews.org/reference/elements/bokeh/BoxWhisker.html). This PR brings the Bokeh rendering of box and whisker plots.

My apologies for the confusion with #1926.

### DIFF
--- a/holoviews/plotting/bokeh/stats.py
+++ b/holoviews/plotting/bokeh/stats.py
@@ -137,8 +137,8 @@ class BoxWhiskerPlot(CompositeElementPlot, ColorbarPlot, LegendPlot):
             q1, q2, q3 = (np.percentile(vals, q=q)
                           for q in range(25, 100, 25))
             iqr = q3 - q1
-            upper = min(q3 + 1.5 * iqr, vals.max())
-            lower = max(q1 - 1.5 * iqr, vals.min())
+            upper = vals[vals <= top + 1.5*iqr].max()
+            lower = vals[vals >= bottom - 1.5*iqr].min()
         else:
             q1, q2, q3 = 0, 0, 0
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks @justinbois. Just to confirm, do you know which definition the inbuilt boxplot in matplotlib uses and whether that matches what you have done here?

philippjfr:
Here is the matplotlib definition:

>whis : float, sequence, or string (default = 1.5)
>As a float, determines the reach of the whiskers to the beyond the first and third quartiles. In other words, where IQR is the interquartile range (Q3-Q1), the upper whisker will extend to last datum less than Q3 + whisIQR). Similarly, the lower whisker will extend to the first datum greater than Q1 - whisIQR

Which reads to me to be identical to the definition you implemented here. I'll merge.

justinbois:
Thanks, @philippjfr , and sorry for the delay (been sick for the past several days). That definition also matches what is in the more general [HoloViews docs](http://holoviews.org/reference/elements/bokeh/BoxWhisker.html).

PR #3761, opened 2019-06-06
Allow automatic extent scaling on line annotations

### LINKED ISSUE #3759: Layout doesn't scale with VLines
Hello,

I'm using this example:

```
# Generate sample data
mu, sigma = 0, 0.1
s = np.random.normal(mu, sigma, 1000)

dist = hv.Histogram(np.histogram(s, bins=20))
s3min = hv.VLine(mu - sigma*3)
s3max = hv.VLine(mu + sigma*3)
s6min = hv.VLine(mu - sigma*6)
s6max = hv.VLine(mu + sigma*6)
layout = dist * s3min * s3max * s6min * s6max
layout
```

The layout doesn't extend to show all the lines. The X axis only scales to show the histogram. If i overlay another distribution, it does extend, but it's like it doesn't take VLines into account when calculating the bounds.

poplarShift:
Fixes #3759 

The following is now possible:

```
import holoviews as hv
hv.extension('bokeh')
import numpy as np

hv.VLine(-1).opts(apply_ranges=True, padding=.1) * hv.Scatter(np.random.rand(100,2))
```

Note that in the matplotlib backend ranges were apparently already applied before then.

### DIFF
--- a/holoviews/plotting/bokeh/annotation.py
+++ b/holoviews/plotting/bokeh/annotation.py
@@ -18,7 +18,7 @@ except:
 from bokeh.transform import dodge
 
 from ...core.util import datetime_types, dimension_sanitizer, basestring
-from ...element import HLine
+from ...element import HLine, VLine
 from ..plot import GenericElementPlot
 from .element import AnnotationPlot, ElementPlot, CompositeElementPlot, ColorbarPlot
 from .styles import text_properties, line_properties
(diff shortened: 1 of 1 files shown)


holoviews/plotting/bokeh/annotation.py
```diff
@@ -130,6 +130,8 @@ class LineAnnotationPlot(ElementPlot, AnnotationPlot):
 
     style_opts = line_properties + ['level']
 
+    apply_ranges = False
```
philippjfr: I think this should probably be a boolean parameter.

poplarShift: Ah, of course.

PR #3793, opened 2019-06-25
add package dependencies to nytaxi example

ltalirz:
when running the app, python will complain about missing direct imports
(dask, xarray, datashader), but the other missing dependencies (fastparquet
and python-snappy) are more difficult to track down, since the `bokeh serve`
command will simply show a white page with no exception on the
python console.

The dependencies listed in the PR are the ones needed in addition if
holoviews was installed via the recommended way:

 conda install -c pyviz holoviews bokeh


philippjfr:
Thanks @Iltalirz, much appreciated.

PR #3803, opened 2019-07-05
Enable cmap option in matplotlib HexTiles

poplarShift:
Now, the following is possible, whereas before it wasn't:

```
import holoviews as hv
hv.extension('matplotlib')
import numpy as np
hv.HexTiles(np.random.randn(10000,2)).opts(cmap='Greens', colorbar=True)
```

If there is any testing being done on cmap and the like, I'd be happy to implement a test if you'd like.


philippjfr:
Looks like the CI setup is currently broken. I'll try to look into that soon.

PR #3836, opened 2019-07-19
Replace HoloViews widgets with Panel

### LINKED ISSUE #84: Improving customizability and layout of widgets
Discussion in https://github.com/ioam/holoviews/issues/70 indicated that we should aim to improve the customizability of the widget layout. The original request by @basnijholt was:

> Is it also possible to determine where the slider will come? In a presentation the 'scrubber' orients itself nicely under the plots, although I don't actually want the buttons and the 'widgets' are on the right of my plots which make the plots too small. Any easy solutions for this?
### Comments:

philippjfr:

> The whole widget Javascript/HTML code isn't brilliant and the layout isn't customizable yet. Unfortunately neither of us is particularly experienced with JS. Once the IPython/Jupyter widget system is more stable we may also decide to start using their widget API again. What IPython doesn't yet offer is static widgets so we'll have to maintain our version at least for now. I'm the only one who is familiar with the widget jinja templates and won't have any time to work on it in the near future. If you know anything about JavaScript I'm happy to take you through it and you could have a shot at improving the layout and/or making the position customizable.

jlstevens:

> I agree with Philipp and I think our best approach is to see if anyone is succeeding in doing anything significantly better in this regard within the IPython Notebook. If they are, we should be able to adapt their code/approach for our purposes.
> 
> Unfortunately, neither Philipp or I are sufficiently well versed in Javascript right now to tackle a general approach for laying out the widgets right now - although I agree this would be a nice feature!
### Summary:

In summary there is currently very little customizability of the layout of the widgets. We are committed to improving the widgets until we or someone else can come up with a way to integrate static widgets with the IPython widgets API, however neither of us currently has time to devote to Javascript development. We would however be happy to provide input and guidance if someone wants to submit a pull request for improvements.
### Known issues:
- Widgets on the right don't leave enough space for large figures. 
- Widget layout behaves differently in IPython 2 vs 3 due to an updated dependency from Bootstrap 2 to 3.
- Both the Bootstrap 2 and 3 have issues when text label gets too long.
### Possibilities to consider:
- Implement Backbone.js based widgets indpendently from IPython
- Implement embedding within IPython widgets system (also Backbone.js based)
- Improve current widget layout

### LINKED ISSUE #449: Notebook archive should support interactive holomaps

### LINKED ISSUE #1677: HoloMap wrongly renders missing data
```python
def sine_curve(phase, freq):

phases      = [0, 1, 2]
frequencies = [0.5, 1.0, 1.5]
curves = {(p,f):sine_curve(p,f) for p in phases for f in frequencies}
del curves[0, 1.0]
hmap = hv.HoloMap(curves, kdims=['phase', 'frequency'])
hmap
```
for [0, 1.0] (which is missing) renders the same as for the one previously seen (here: [0, 1.5]) while it should render an empty plot.

Screenshots: 
- https://www.dropbox.com/s/v2r6jov9en8fupp/Screenshot%202017-07-04%2020.23.24.png?dl=0
- https://www.dropbox.com/s/6rrsg8101d27c68/Screenshot%202017-07-04%2020.23.36.png?dl=0

### LINKED ISSUE #2885: Saving scrubber to html broken (result is slider instead)
**When i run the following code in Spyder:**
import numpy as np
import holoviews as hv
hv.extension('bokeh')
np.random.seed(42)
coords = [(i, np.random.random()) for i in range(20)]
hv.Scatter(coords)

**The output is :"Out[45]: :Scatter   [x]   (y)**",nothing else is shown。。
I wonder if Holoviews can run properly in  Spyder(IDE) , but in jupyter notebook is OK.
Plesase help me, thanks.

### LINKED ISSUE #3027: holoviews dynamicmap widgets don't show sliders
Since our latest update on pangeo.pydata.org, holoviews dynamicmap widgets have been broken.

```
import holoviews as hv
import numpy as np
hv.extension('bokeh', width=100)

def fm_modulation(f_carrier=220, f_mod=220, mod_index=1, length=0.1, sampleRate=2000):
    sampleInc = 1.0/sampleRate
    x = np.arange(0,length, sampleInc)
    y = np.sin(2*np.pi*f_carrier*x + mod_index*np.sin(2*np.pi*f_mod*x))
    return hv.Curve((x, y), 'Time', 'Amplitude')

%%opts Curve [width=600]
f_carrier = np.linspace(20, 60, 3)
f_mod = np.linspace(20, 100, 5)

curve_dict = {(fc, fm): fm_modulation(fc, fm) for fc in f_carrier for fm in f_mod}

kdims = [hv.Dimension(('f_carrier', 'Carrier frequency'), default=40),
         hv.Dimension(('f_mod', 'Modulation frequency'), default=60)]
holomap = hv.HoloMap(curve_dict, kdims=kdims)
holomap
```

image

We are using holoviews version 1.10.7.

Any idea what is going on here?

### LINKED ISSUE #3212: Widgets are not escaping single quotes correctly
As the title says, single quotes in a widget dropdown item break the widgets. Minimal example:

```python
hv.HoloMap({c: hv.Curve([1, 2, 3]) for c in ["'", 'Some other name']}, 'Client Name')
```

### LINKED ISSUE #3868: multibyte characters in widgets are not working
desired result: 
スクリーンショット 2019-08-02 13 36 57
(Label in chart title is equal to one in selection box, and it is corresponding)

actual result:
スクリーンショット 2019-08-02 13 37 38

Now I'm facing two different problems:
1. Nothing happened when I select item in selection box
2. displayed character in selection box is garbled

To solve problem 1, I think you need to put `ensure_ascii=False` for `json.dumps` at L463 in `holoviews/plotting/widgets/__init__.py`

To solve second one, `static_template` at L64 in `holoviews/plotting/renderer.py` should be

```python
static_template = """
<html>
  <head>
    <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
    {css}
    {js}
  </head>
  <body>
    {html}
  </body>
</html>
"""
```

My code is here:

```python
import pandas
import holoviews
import bokeh
import hvplot.pandas

print('pandas', pandas.__version__)
print('holoviews', holoviews.__version__)
print('bokeh', bokeh.__version__)

holoviews.extension('bokeh')

df = pandas.DataFrame({
    'a': ['あ', 'あ', 'あ', 'あ', 'あ', 'い', 'い', 'い', 'い', 'う', 'う', 'う'],
    'b': [10, 11, 12, 11, 10, 20, 21, 22, 21, 31, 32, 31],
    'c': [0, 1, 2, 1, 0, 0, -1, -2, -1, 1, 2, 3]
})
line_plot_groupby = df.hvplot.line(x='c', y='b', groupby='a', dynamic=False)
renderer = holoviews.renderer('bokeh')
renderer.save(line_plot_groupby, 'line_plot_groupby')
```

pandas 0.25.0
holoviews 1.2.9.post67+g7dd50144
bokeh 1.3.1

philippjfr:
Replacing HoloViews widgets with Panel widgets.

Requires https://github.com/pyviz/panel/pull/543 to match the previous styling.

- [x] Check matplotlib
- [x] Check plotly
- [x] Check bokeh
- [x] Check fps works
- [x] Check periodic works
- [x] Check server works
- [x] Ensure widget json data export works (especially for docs)
- [x] Fix #449
- [x] Fixes #1677 
- [x] Fixes #2885
- [x] Add tests

Closes #3868
Closes #3212 
Closes #84
Closes #3027


jonmmease:
If I merge https://github.com/pyviz/holoviews/pull/3837 into this branch, and use the panel version from https://github.com/pyviz/panel/pull/529, I'm able to get a plotly HoloMap to update with the slider 🎉 

```python
import numpy as np
import holoviews as hv
import panel as pn
hv.extension("plotly")

def sine_curve(phase, freq):

curve_dict = {f:sine_curve(0,f) for f in frequencies}

hmap = hv.HoloMap(curve_dict, kdims='frequency')
hmap
```
Screen Shot 2019-07-24 at 8 01 22 AM

I do see an error in the JS console each time the slider is moved.

Screen Shot 2019-07-24 at 8 05 33 AM

But the plot does update successfully.

I was trying to get a better understanding of what the JS side is doing here, so I saved the HoloMap to a standalone HTML file with

```python
hv.save(hmap, 'hmap_plotly.html')
```

But in the resulting HTML file the slider updates don't update the plot.  This is also the case for the `bokeh` extension.

philippjfr:
>I do see an error in the JS console each time the slider is moved.

That should be fixed in bokeh 1.3.0, it's some minor bug to do with the slider tooltip.

philippjfr:
>I was trying to get a better understanding of what the JS side is doing here, so I saved the HoloMap to a standalone HTML file with

Still need to look at this, suspect save does not correctly use the panel embed features yet.

jonmmease:
Do you think this will be a somewhat long-running branch?  Would it make sense for me to PR the plotly updates into this branch?

philippjfr:
>Do you think this will be a somewhat long-running branch? Would it make sense for me to PR the plotly updates into this branch?

I think that probably makes sense. Can't say quite how long this will take and keeping the two PRs in sync seems like a pain.

jonmmease:
Sounds good. I'll make your suggested updates in the other branch, then rebase on this and open a new PR.

jonmmease:
Digging into the save issue a little. The resulting `Bokeh` state model is empty. And I think it starts with the fact that `panel.select(Widget)` returns empty on the `panel.pane.holoviews.HoloViews` object that `.save(embed=True)` is called on.

https://github.com/pyviz/panel/blob/b274c45252c54091dab4caf38a2b4b9a31d75c4b/panel/io/embed.py#L105-L106

`panel.widgets` is empty, but `panel.widgets_from_dimensions(panel.object)` returns the `DiscreteSlider` instance.

philippjfr:
Fixed the widget saving issue.

jonmmease:
capturing note by @philippjfr from https://github.com/pyviz/holoviews/pull/3851#issuecomment-514998547.

> I think at least for the simple matplotlib case we should probably not require panel...

This got me thinking, rather than just doing this for matplotlib maybe we should have a mode that allows non-dynamic non-widget figures to be rendered by their respective backend libraries.  In version 4, plotly.py has a pretty flexible rendering system and can display plots in a bunch of contexts that full-blown HoloViews can't (Colab, VSCode, etc.).  And of course matplotlib works everywhere, and I believe bokeh's support is pretty wide as well.

holoviews/plotting/renderer.py
```diff
@@ -449,72 +390,11 @@ class needed to render it with the current renderer.
...
-            js_html += comm_js
-        else:
-            js_html += '\n<script type="text/javascript">%s</script>' % comm_js
-
-        return unicode(js_html), unicode(css_html)
-
+        Deprecated: No longer needed
+        """
+        param.main.warning("Renderer.html_assets is deprecated as all "
+                           "JS and CSS dependencies are not handled by "
```
philippjfr:
```suggestion
                           "JS and CSS dependencies are now handled by "
```

jbednar:
I agree; sounds like a good baseline so that people can still use most HoloViews/hvPlot functionality anywhere.

philippjfr:
>In version 4, plotly.py has a pretty flexible rendering system and can display plots in a bunch of contexts that full-blown HoloViews can't (Colab, VSCode, etc.).

Last I checked we **can** render in VSCode and nteract now. Colab is an issue because they iframe the cell outputs which means the extension has to be loaded for each individual cell. I'd be okay leaving that for a later PR and doing the simplest thing in this PR.

holoviews/plotting/plot.py
```diff
@@ -42,6 +44,45 @@ class Plot(param.Parameterized):
...
+            return None
+
+    @property
+    def document(self):
+        return self._document
+
+    @document.setter
+    def document(self, doc):
+        if (doc and hasattr(doc, 'on_session_destroyed') and
+            self.root is self.handles.get('plot') and not hasattr(self, 'layout_dict')):
```
```
not hasattr(self, 'layout_dict')
```

Should fix this condition, it's really looking for AdjointLayoutPlots

holoviews/plotting/plot.py
```diff
@@ -83,6 +116,74 @@ def cleanup(self):
...
+        # Update if not top-level, batched or an ElementPlot
+        if not self.top_level or isinstance(self, GenericElementPlot):
+            self.update(key)
+
+
+    def push(self):
+        """
+        Pushes updated plot data via the Comm.
+        """
+        if self.renderer.mode == 'server':
```
philippjfr: This needs to handle the non-bokeh case for server.

holoviews/plotting/plot.py
jbednar: At least please document that that's what it's doing.

holoviews/tests/plotting/bokeh/testrenderer.py
```diff
@@ -1,45 +1,54 @@
...
+    pn = None
 
 
 class BokehRendererTest(ComparisonTestCase):
 
     def setUp(self):
-        if 'bokeh' not in Store.renderers:
-            raise SkipTest("Bokeh required to test widgets")
+        if 'bokeh' not in Store.renderers and pn is not None:
+            raise SkipTest("Bokeh and panel required to test bokeh renderer")
```
jbednar:
```suggestion
            raise SkipTest("Bokeh and Panel required to test 'bokeh' renderer")
```

holoviews/plotting/mpl/renderer.py
```diff
@@ -82,38 +82,12 @@ class MPLRenderer(Renderer):
     mode = param.ObjectSelector(default='default', objects=['default'])
 
 
-    mode_formats = {'fig':     {'default': ['png', 'svg', 'pdf', 'html', None, 'auto']},
-                    'holomap': {'default': ['widgets', 'scrubber', 'webm','mp4', 'gif',
-                                            'html', None, 'auto']}}
```
jlstevens: Have you checked the output magic still works? I suspect the old format might have been expected there...

holoviews/plotting/plotly/renderer.py
```diff
@@ -1,138 +1,54 @@
...
-var data = JSON.parse(msg);
-$.each(data.data, function(i, obj) {{
-  $.each(Object.keys(obj), function(j, key) {{
-    plot.data[i][key] = obj[key];
-  }});
-}});
-var plotly = window._Plotly || window.Plotly;
-plotly.relayout(plot, data.layout);
-plotly.redraw(plot);
-"""
```
jlstevens: Very happy to see this sort of inline JS go away!

holoviews/plotting/mpl/renderer.py
philippjfr: Yeah I made the change there too.

holoviews/util/settings.py
```diff
@@ -141,40 +141,46 @@ class OutputSettings(KeywordSettings):
...
-                            ('holomap'     , None),
-                            ('widgets'     , None),
-                            ('fps'         , None),
-                            ('max_frames'  , 500),
-                            ('size'        , None),
-                            ('dpi'         , None),
-                            ('filename'    , None),
-                            ('info'        , False),
-                            ('css'         , None)])
+    allowed = {'backend'       : list_backends(),
```
jlstevens: Ok, I think this addresses my worry about the output magic working - looks like it was updated here after all.

holoviews/plotting/renderer.py
```diff
@@ -134,43 +144,48 @@ class Renderer(Exporter):
...
-    widgets = {'scrubber': ScrubberWidget, 'widgets': SelectionWidget}
-
-    core_dependencies = {'jQueryUI': {'js': ['https://code.jquery.com/ui/1.10.4/jquery-ui.min.js'],
-                                      'css': ['https://code.jquery.com/ui/1.10.4/themes/smoothness/jquery-ui.css']}}
-
-    extra_dependencies = {'jQuery': {'js': ['https://code.jquery.com/jquery-2.1.4.min.js']},
-                          'underscore': {'js': ['https://cdnjs.cloudflare.com/ajax/libs/underscore.js/1.8.3/underscore-min.js']},
-                          'require': {'js': ['https://cdnjs.cloudflare.com/ajax/libs/require.js/2.1.20/require.min.js']},
-                          'bootstrap': {'css': ['https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/css/bootstrap.min.css']}}
-
```
jlstevens: Yay!

holoviews/plotting/renderer.py
```diff
@@ -13,11 +13,17 @@
 from contextlib import contextmanager
 
 import param
+
+from panel import config
+from panel.io.notebook import load_notebook, render_model, render_mimebundle
+from panel.pane import HoloViews
```
jlstevens: I would prefer to have `HoloViews` be qualified i.e used as `panel.pane.HoloViews` or be imported as `HoloViewsPane`....

holoviews/plotting/renderer.py
```diff
@@ -275,7 +286,7 @@ def _apply_post_render_hooks(self, data, obj, fmt):
         return data
 
 
-    def html(self, obj, fmt=None, css=None, **kwargs):
+    def html(self, obj, fmt=None, css=None, resources='cdn', **kwargs):
```
jlstevens: Any reason 'cdn' is lowercase here? Isn't it an acronym that is normally capitalized?

jlstevens:
Looks great! Very happy to see a load of nasty code (including inline JS) go away...

I have two very minor suggestions relating to the code but otherwise I am happy to merge. I suppose my only other thought is it would be good to record the sort of (hopefully minor!) changes people might expect to see the CHANGELOG/release notes...

holoviews/plotting/renderer.py
```diff
@@ -90,34 +92,42 @@ class Renderer(Exporter):
...
+                                objects=['default', 'server'], doc="""
+        Whether to render the object in regular or server mode. In server
+        mode a bokeh Document will be returned which can be served as a
+        bokeh server app. By default renders all output is rendered to HTML.""")
 
-    size=param.Integer(100, doc="""
+    size = param.Integer(100, doc="""
         The rendered size as a percentage size""")
 
+    widget_location = param.ObjectSelector(default=None, allow_None=True, objects=[
```
jlstevens: It would be nice to have a shorted name than `widget_location` but I don't currently have any better suggestions...

holoviews/plotting/renderer.py
philippjfr: Done

holoviews/plotting/renderer.py
philippjfr: Fixed.

jlstevens:
I won't wait for appveyor to go green as it is taking a while to start and all the other tests are green. I'm very happy to see 2000+ lines of old widget code removed and to let panel handle that side of HoloViews in a more maintainable and future proof way. Merging!

PR #3850, opened 2019-07-24
Replace plotly HTML rendering with Panel

jonmmease:
**Update:** For a more wholistic approach, see https://github.com/pyviz/holoviews/pull/3851

---

Supersedes https://github.com/pyviz/holoviews/pull/3837, based on https://github.com/pyviz/holoviews/pull/3836.

Rendering HoloViews objects with widgets was already working, this branch adds support simple elements without widgets using Panel.

One note, I'm getting a pyviz_comms error when calling `figure_pane._repr_mimebundle_()` from a non-IPython context (debugging in PyCharm).  I'm not sure if this method is supposed to work outside of IPython, so we might need to use something else here eventually.


jonmmease:
Closed in favor of https://github.com/pyviz/holoviews/pull/3851

PR #3851, opened 2019-07-24
Use Panel to generate HTML output

jonmmease:
In https://github.com/pyviz/holoviews/pull/3836, all HoloViews objects with widgets are displayed using Panel, but objects without widgets are still displayed using the legacy rendering paths.

This PR goes a step further and uses Panel to perform the HTML rendering of non-widget HoloViews objects as well (across all backends). The legacy code paths for rendering figures to static images have been preserved and simplified.

I've spot checked the rendering of objects with and without widgets across all three backends, as well as rendering to static images across all backends.  The tests suite will need some attention if we go this direction.

The commit log is pretty clean, so it might be helpful to look at the changes commit-by-commit.

@philippjfr, if this approach looks alright, then this would supersede https://github.com/pyviz/holoviews/pull/3850.  If this seems too large/risky of a change then I can keep building on https://github.com/pyviz/holoviews/pull/3850 instead.


philippjfr:
I think at least for the simple matplotlib case we should probably not require panel but otherwise I'm pretty happy with this since it simplifies things a lot.

jonmmease:
Great!

I'll take a look at removing the Panel dependency from the simple matplotlib case.  But this might complicate the logic a bit since we'll still need panel for the matplotlib widget case, so I'm not sure where to handle initializing panel (`pn.extension()`).  This might also require some back-end dependent logic in `_validate` (or for the matplotlib renderer to override this method).

jlstevens:
Happy to see this code go away! As pointed out above, the usage of `pn.extension` in the current context doesn't seem quite right but I'm sure that can be addressed.

philippjfr:
How would you feel about merging this into my PR directly? I'd rather test all the possible rendering  issues together rather than going through that twice.

jonmmease:
If you're ready for it that would be great.  I haven't done any more on it since my last push so this is a fine time to merge it in.

PR #3854, opened 2019-07-25
Interface improvements for dask and dtype handling

philippjfr:
* Adds an ``Interface.dtype`` method to inspect dtypes without loading data
* Adds Interface.values ``compute`` argument to access dask arrays without loading them into memory


philippjfr:
The changes here are all pretty straightforward so I'm going to merge so I can start on the issue I was actually hoping to fix, which is optimizing range computations.

PR #3860, opened 2019-07-29
Optimized axis range and factor finding

philippjfr:
- [x] Skips range finding if explicit dimension range is set
- [x] Avoid computing range on categorical axes
- [x] Respect Dimension.type if set to string

### DIFF
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -484,7 +484,10 @@ class DimensionedPlot(Plot):
             # Compute dimension normalization
             for el_dim in el.dimensions('ranges'):
                 if hasattr(el, 'interface'):
-                    dtype = el.interface.dtype(el, el_dim)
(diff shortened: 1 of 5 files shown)


holoviews/plotting/plot.py
```diff
@@ -483,16 +483,32 @@ def _compute_group_range(cls, group, elements, ranges):
 
             # Compute dimension normalization
             for el_dim in el.dimensions('ranges'):
-                if isinstance(el, Graph) and el_dim in el.kdims[:2]:
+                if hasattr(el, 'interface'):
+                    if isinstance(el, Graph) and el_dim in el.nodes.dimensions():
+                        dtype = el.nodes.interface.dtype(el.nodes, el_dim)
+                    elif isinstance(el, Contours) and el.level is not None:
+                        dtype = None # Deprecated level
```
```suggestion
                        dtype = np.array([el.level]).dtype # Remove when deprecating level
```

PR #3865, opened 2019-07-31
Propagate backend argument to Store.lookup_options

jonmmease:
Closes https://github.com/pyviz/panel/issues/566

The underlying issue is that the `backend` argument to `Store.lookup_options` was not passed along to the `closest` and `options` methods.  So the input backend was effectively ignored.


philippjfr:
Looks good to me, and thanks for the tests.

PR #3866, opened 2019-07-31
Plotly: Add Panel-based support for linked streams and dynamic updates

jonmmease:
## Overview
This PR supersedes https://github.com/pyviz/holoviews/pull/3316, building linked stream support on top of the HoloViews Panel integration started in the base branch (philippjrf/widgetsectomy).  Supported linked streams are:
 - `RangeXY`
 - `RangeX`
 - `RangeY`
 - `BoundsXY`
 - `BoundsX`
 - `BoundsY`
 - `Selection1d`

It also adds support for updating DynamicMaps using streams from Python.

@philippjfr @jlstevens @jbednar 

## Linked Streams Architecture
The core implementation of linked streams is in ff0eb73.  Here's an overview

Every Plotly `ElementPlot` now has a `trace_uid` property. This `trace_uid` is used as the `uid` property of the Plotly trace that will be associated with any linked stream attached to this plot's Element.

This plot element is an input to a new set of `PlotlyCallback` classes.  Each callback class is responsible for tracking all active instances if its own class type using a `WeakValueDictionary` from `trace_uid` to callback instance. (To reduce boilerplate the instance tracking logic is in the `MetaPlotlyCallback` metaclass).

Additionally, each callback class is responsible for dispatching callback events to the streams of all relevant callback instances.  This dispatching is done in the `update_streams_from_property_update` class method of each callback class.

When the plotly backend is activated, a backend switch hook is used to register a custom plotly pane construction function with the Panel `HoloViews` class.  This function is responsible for constructing the Plotly pane instance for the HoloViews pane. In addition, it also installs callbacks to call the `update_streams_from_property_update` methods when the appropriate callback parameter is updated.

## Dynamic Update Architecture
The core implementation of dynamic update support is in 6352793. Here's an overview.

The `id` property of `PlotlyPlot` instances is now set to a random UID.  This `id` is added as a top-level `_id` key in the plotly figure dictionary produced by the plot class. (this property is removed before the user would get it by calling `hv.render`, see 845f423baff59538679ff7f2918dfa9a939dd7d9).

The custom Plotly pane constructor function described above creates a mapping between the `_id` property in the figure dictionary and the associated plotly pane that is constructed.  This mapping is stored as a class level `WeakValueDictionary` in the `PlotlyRenderer` class.

The `PlotlyRenderer` class has a new `trigger_plot_pane` class method that inputs a plot id and a new figure dictionary, and it is responsible for updating the appropriate plotly pane to the updated value.  This method is called by the various `update_frame` methods across the `PlotlyPlot` class hierarchy.

## Other changes
The remaining commits are pretty small and self explanatory I think.  Let me know if you have any questions about them!

## Examples
The examples from https://github.com/pyviz/holoviews/pull/3316 still work as-is.


holoviews/plotting/plot.py
```diff
@@ -654,6 +654,51 @@ def __len__(self):
         return len(self.keys)
 
 
+class CallbackPlot(object):
```
philippjfr: Good idea to factor this out!

PR #3905, opened 2019-08-16
Plotly: Add Panel-based support for linked streams and dynamic updates

philippjfr:
Reopened #3866 against master

## Overview

This PR supersedes https://github.com/pyviz/holoviews/pull/3316, building linked stream support on top of the HoloViews Panel integration started in the base branch (philippjrf/widgetsectomy).  Supported linked streams are:
 - `RangeXY`
 - `RangeX`
 - `RangeY`
 - `BoundsXY`
 - `BoundsX`
 - `BoundsY`
 - `Selection1d`

It also adds support for updating DynamicMaps using streams from Python.

@philippjfr @jlstevens @jbednar 

## Linked Streams Architecture
The core implementation of linked streams is in ff0eb73.  Here's an overview

Every Plotly `ElementPlot` now has a `trace_uid` property. This `trace_uid` is used as the `uid` property of the Plotly trace that will be associated with any linked stream attached to this plot's Element.

This plot element is an input to a new set of `PlotlyCallback` classes.  Each callback class is responsible for tracking all active instances if its own class type using a `WeakValueDictionary` from `trace_uid` to callback instance. (To reduce boilerplate the instance tracking logic is in the `MetaPlotlyCallback` metaclass).

Additionally, each callback class is responsible for dispatching callback events to the streams of all relevant callback instances.  This dispatching is done in the `update_streams_from_property_update` class method of each callback class.

When the plotly backend is activated, a backend switch hook is used to register a custom plotly pane construction function with the Panel `HoloViews` class.  This function is responsible for constructing the Plotly pane instance for the HoloViews pane. In addition, it also installs callbacks to call the `update_streams_from_property_update` methods when the appropriate callback parameter is updated.

## Other changes
The remaining commits are pretty small and self explanatory I think.  Let me know if you have any questions about them!

## Examples
The examples from https://github.com/pyviz/holoviews/pull/3316 still work as-is.

### DIFF
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -21,7 +21,7 @@ from ...core.util import (
 from ..links import Link
 from ..plot import (
     DimensionedPlot, GenericCompositePlot, GenericLayoutPlot,
-    GenericElementPlot, GenericOverlayPlot, GenericAdjointLayoutPlot
+    GenericElementPlot, GenericOverlayPlot, GenericAdjointLayoutPlot,
     CallbackPlot
 )
 from ..util import attach_streams, displayable, collate
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -18,7 +18,6 @@ from ...core.options import SkipRendering
 from ...core.util import (
     basestring, cftime_to_timestamp, cftime_types, get_method_owner,
     unique_iterator, wrap_tuple, wrap_tuple_streams, _STANDARD_CALENDARS)
-
 from ..links import Link
 from ..plot import (
     DimensionedPlot, GenericCompositePlot, GenericLayoutPlot,
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -18,11 +18,12 @@ from ...core.options import SkipRendering
 from ...core.util import (
     basestring, cftime_to_timestamp, cftime_types, get_method_owner,
     unique_iterator, wrap_tuple, wrap_tuple_streams, _STANDARD_CALENDARS)
-from ...streams import Stream
+
 from ..links import Link
 from ..plot import (
     DimensionedPlot, GenericCompositePlot, GenericLayoutPlot,
     GenericElementPlot, GenericOverlayPlot, GenericAdjointLayoutPlot
+    CallbackPlot
 )
 from ..util import attach_streams, displayable, collate
 from .callbacks import LinkCallback
@@ -31,7 +32,7 @@ from .util import (
     empty_plot, decode_bytes, theme_attr_json, cds_column_replace)
 
 
-class BokehPlot(DimensionedPlot):
+class BokehPlot(DimensionedPlot, CallbackPlot):
     """
     Plotting baseclass for the Bokeh backends, implementing the basic
     plotting interface for Bokeh based plots.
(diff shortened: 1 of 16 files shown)


philippjfr:
@jonmmease I rebased this PR because I forgot to merge it against the PR before merging. It was **very** painful and I may have messed this up, so could I ask you to test it?

holoviews/plotting/plotly/renderer.py
```diff
@@ -68,3 +112,18 @@ def load_nb(cls, inline=True):
         """
         import panel.models.plotly # noqa
         cls._loaded = True
+
+
+
+    @classmethod
+    def trigger_plot_pane(cls, plot_id, fig_dict):
```
philippjfr: In the latest version panel adds itself to `Plot._panel`, I think we just have to propagate this down to all child plots to avoid the ``_plot_panes`` registry entirely.

holoviews/plotting/bokeh/plot.py
```diff
@@ -18,11 +18,12 @@
 from ...core.util import (
     basestring, cftime_to_timestamp, cftime_types, get_method_owner,
     unique_iterator, wrap_tuple, wrap_tuple_streams, _STANDARD_CALENDARS)
-from ...streams import Stream
+
```
philippjfr:
```suggestion
```

holoviews/plotting/plotly/element.py
```diff
@@ -100,6 +102,13 @@ class ElementPlot(PlotlyPlot, GenericElementPlot):
     # Declare which styles cannot be mapped to a non-scalar dimension
     _nonvectorized_styles = []
 
+    def __init__(self, element, plot=None, **params):
+        super(ElementPlot, self).__init__(element, **params)
+        self.trace_uid = str(uuid.uuid4())
+        self.static = len(self.hmap) == 1 and len(self.keys) == len(self.hmap)
```
philippjfr: We should have this information somewhere tbh, I'll look into it.

jonmmease:
Wow, thanks for taking on this rebase @philippjfr.  Yes, I'll review and test soon!

holoviews/plotting/plotly/plot.py
```diff
@@ -45,8 +65,9 @@ def initialize_plot(self, ranges=None):
 
 
     def update_frame(self, key, ranges=None):
-        return self.generate_plot(key, ranges)
-
+        plot = self.generate_plot(key, ranges)
+        PlotlyRenderer.trigger_plot_pane(self.id, self.state)
```
philippjfr: Does this have to happen here or is the triggering that happens in ``.push`` enough? One thing I just realized is that the current ``push`` approach probably doesn't work for Layouts and I need to propagate the ``Plot._pane`` down to all child plots anyway.

holoviews/plotting/plotly/callbacks.py
```diff
@@ -0,0 +1,215 @@
...
+from param.parameterized import add_metaclass
+
+from ...streams import (
+    Stream, Selection1D, RangeXY, RangeX, RangeY, BoundsXY, BoundsX, BoundsY
+)
+
+from .util import _trace_to_subplot
+
+
+class MetaPlotlyCallback(type):
```
jlstevens: As a quick comment, I would prefer the name `PlotlyCallbackMetaClass` because the name alone just suggests it is meta, not specifically a metaclass.

holoviews/plotting/plotly/renderer.py
```diff
@@ -3,12 +3,35 @@
...
 with param.logging_level('CRITICAL'):
     import plotly.graph_objs as go
 
 from ..renderer import Renderer, MIME_TYPES, HTML_TAGS
 from ...core.options import Store
 from ...core import HoloMap
+from .callbacks import callbacks
+
+
+def _PlotlyHoloviews(fig_dict):
```
jlstevens: I find the naming convention here confusing, it is a `def` and therefore a function, but it is capitalized like a class. It is also underscored which is not a convention generally used in HoloViews for classes. In addition, a more descriptive name would be good...

philippjfr: I think we should be underscoring more things that are completely private and shouldn't be used in any other context, so I vote to keep the underscore. I'll rename to ``_PlotlyHoloviewsPane`` and add a docstring.

holoviews/plotting/plotly/callbacks.py
philippjfr: No objection.

holoviews/plotting/plotly/renderer.py
jbednar: The underscore seems fine, but why is it capitalized?

holoviews/plotting/plotly/plot.py
```diff
@@ -10,18 +11,32 @@
...
+class PlotlyPlot(DimensionedPlot, CallbackPlot):
 
     backend = 'plotly'
 
     width = param.Integer(default=400)
 
     height = param.Integer(default=400)
 
+    def __init__(self, *args, **kwargs):
+        # The Plotly backend doesn't use a comm directly because it's dynamic
```
jbednar:
```suggestion
        # The Plotly backend doesn't use a comm directly because its dynamic
```

holoviews/plotting/plotly/renderer.py
philippjfr: Because it slightly modifies the behavior of the regular `HoloViews` pane.

philippjfr: i.e. it is an alternative class constructor.

jlstevens: Even if you keep the underscore and capitalization, I still think a more meaningful name would be good.

philippjfr: Thats why I added Pane to the name.

PR #3907, opened 2019-08-16
Add extend for plot

ahuang11:
The main reason why I want to set this manually is so if I have an animation and the min/max goes out of range, it doesn't awkwardly make rectangular edges into triangular and adjust the plot position

### DIFF
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -663,10 +663,6 @@ class ColorbarPlot(ElementPlot):
        User-specified colorbar axis range limits for the plot, as a tuple (low,high).
        If specified, takes precedence over data and dimension ranges.""")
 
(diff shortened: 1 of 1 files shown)


philippjfr:
HoloViews master isn't broken it just requires the latest dev release of panel now.

ahuang11:
image

holoviews/plotting/mpl/element.py
```diff
@@ -698,13 +698,17 @@ class ColorbarPlot(ElementPlot):
     symmetric = param.Boolean(default=False, doc="""
         Whether to make the colormap symmetric around zero.""")
 
+    extend = param.ObjectSelector(
```
philippjfr: I think this should probably be `cbar_extend` or `colorbar_extend` since ``extend`` could mean a number of things.

PR #3913, opened 2019-08-17
Add cformatter for bokeh

ahuang11:
image

```
import xarray as xr
import hvplot.xarray

import holoviews as hv

hv.extension('bokeh')

da = xr.tutorial.open_dataset('air_temperature')['air'].isel(time=0)
da.hvplot('lon', 'lat').opts(clabel='red', cformatter='%.5f')
```


philippjfr:
Thanks @ahuang11!

PR #3919, opened 2019-08-20
Add LabelledData.dataset property

jonmmease:
## Motivation
This PR was motivated by the automatic selection proposal in https://github.com/pyviz/holoviews/issues/3842.  The goal that this PR addresses is to provide a way for  elements to maintain a reference to all of the dimensions in the dataset that they were created from.

I believe this approach accomplishes this goal in a fairly general way, and so I expect it to be useful for other purposes as well.

@jlstevens @philippjfr 

## Overview
A new `.dataset` property is added to the `LabelledData` class (the class that defines the `.data` property).  This property should always be either `None` or an instance of a `Dataset` class (not an instance of a subclass of `Dataset`).

## Rules for the dataset property
The rules that the new dataset property follows are best described by the tests in the new test file at `holoviews/tests/core/testdatasetproperty.py`.  Here are some notes:

 - The `dataset` property generally references the same `.data` object as the element and it generally includes dimensions for all of the columns in `.data`.
 - When an element is created directly from the constructor, the `dataset` property is simply a new `Dataset` object wrapping the element's `data`.
- When elements are created from other elements, the goal is to propagate the `dataset` property from the initial element. Some examples where the dataset property propagates:
    - `element.to`
    - `element.clone`
    - `element.reindex`
    - `element.iloc`
    - `element.select`
    -  `Element(element)` (passing one element to an element constructor)

## Special rules for `Histogram`
This PR includes special handling for the `Histogram` element.  This element stores, as its `data` property, the bin edges and frequencies/counts of the histogram.  It does not retain a reference back to any dataset that may have been used to create it.  In this way, it is a bit inconsistent with other related elements, like `Distribution`, which do store the raw data in the `.data` property.

For the `Histogram` element, the `dataset` property stores a reference to the `Dataset` that it was created from (or `None` if the `Histogram` is constructed with predetermined edges and frequencies).  This makes it possible to recreate the histogram from the `dataset` property, or to create a histogram with only a subset of the original data (This is the linked data selection usecase).

If we decided to someday change `Histogram` to work like `Distribution`, then `Histogram` would no longer be a special case.

## TODO
 - [x] Flow dataset through `Distribution` and `Bivariate` elements

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -505,7 +505,13 @@ class LabelledData(param.Parameterized):
         if self._dataset is None:
             # Create a default Dataset to wrap input data
             try:
-                dataset = Dataset(self.data)
+                kdims = list(params.get('kdims', []))
+                vdims = list(params.get('vdims', []))
+                dims = kdims + vdims
+                dataset = Dataset(
+                    self.data,
+                    kdims=dims if dims else None
+                )
                 if len(dataset.dimensions()) == 0:
                     # No dimensions could be auto-detected in data
                     raise DataError("No dimensions detected")
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -490,8 +490,13 @@ class LabelledData(param.Parameterized):
         self.data = data
 
         # Handle initializing the dataset property.
+        input_dataset = params.pop('dataset', None)
         if type(self) is Dataset:
             self._dataset = self
+        elif input_dataset is not None:
+            # Clone dimension info from input dataset with reference to new
+            # data. This way we keep the metadata for all of the dimensions.
+            self._dataset = input_dataset.clone(data=self.data)
         else:
             # Create a default Dataset to wrap input data
             self._dataset = Dataset(self.data)
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -505,7 +505,11 @@ class LabelledData(param.Parameterized):
         if self._dataset is None:
             # Create a default Dataset to wrap input data
             try:
-                self._dataset = Dataset(self.data)
+                dataset = Dataset(self.data)
+                if len(dataset.dimensions()) == 0:
+                    # No dimensions could be auto-detected in data
+                    raise DataError("No dimensions detected")
+                self._dataset = dataset
             except DataError:
                 # Data not supported by any storage backend. leave _dataset as
                 # None
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -490,21 +490,26 @@ class LabelledData(param.Parameterized):
         self.data = data
 
         # Handle initializing the dataset property.
+        self._dataset = None
         input_dataset = params.pop('dataset', None)
         if type(self) is Dataset:
             self._dataset = self
         elif input_dataset is not None:
             # Clone dimension info from input dataset with reference to new
             # data. This way we keep the metadata for all of the dimensions.
-            self._dataset = input_dataset.clone(data=self.data)
-        else:
+            try:
+                self._dataset = input_dataset.clone(data=self.data)
+            except DataError:
+                # Dataset not compatible with input data
+                pass
+        if self._dataset is None:
             # Create a default Dataset to wrap input data
             try:
                 self._dataset = Dataset(self.data)
             except DataError:
                 # Data not supported by any storage backend. leave _dataset as
                 # None
-                self._dataset = None
+                pass
 
         self._id = None
         self.id = id
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -486,7 +486,7 @@ class LabelledData(param.Parameterized):
         This class also has an id instance attribute, which
         may be set to associate some custom options with the object.
         """
-        from . import Dataset
+        from . import Dataset, DataError
         self.data = data
 
         # Handle initializing the dataset property.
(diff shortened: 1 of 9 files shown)


holoviews/core/dimension.py
```diff
@@ -486,7 +486,41 @@ def __init__(self, data, id=None, plot_id=None, **params):
...
+            except DataError:
+                # Dataset not compatible with input data
+                pass
+        if self._dataset is None:
+            # Create a default Dataset to wrap input data
+            try:
+                kdims = list(params.get('kdims', []))
+                vdims = list(params.get('vdims', []))
+                dims = kdims + vdims
+                dataset = Dataset(
```
philippjfr: Just wondering whether this will have performance implications. Generally I wouldn't expect it to be expensive to create a Dataset but there might be a bit of overhead. Since we already have some guarantees that the data is in a valid format maybe we can bypass all the initialization code and just provide it with the Interface, data and dimensions.

holoviews/core/dimension.py
```diff
@@ -486,7 +486,41 @@ def __init__(self, data, id=None, plot_id=None, **params):
...
+            try:
+                self._dataset = input_dataset.clone(data=self.data)
+            except DataError:
+                # Dataset not compatible with input data
+                pass
+        if self._dataset is None:
+            # Create a default Dataset to wrap input data
+            try:
+                kdims = list(params.get('kdims', []))
+                vdims = list(params.get('vdims', []))
```
philippjfr: Should this be self.kdims and self.vdims?

philippjfr:
Just so I fully understand the proposal. Should all operations try to keep the original Dataset, e.g. what about operations like ``univariate_kde``, ``bivariate_kde``, ``contours``? Also I'm not quite clear when an internal operation should keep hold of the original data or reference the new data. What would be kind of nice is if along with the data we could keep a ``dim`` expressions (for select/groupby operations) or operation instances (for operations) which let you replay the whole data provenance starting with original dataset.

holoviews/core/dimension.py
jonmmease: Maybe, I hadn't tracked through whether those would always be set by this point in the object construction process.

philippjfr: Oh right, probably not and by this point ``params`` would always contain the appropriate kdims and vdims I think.

philippjfr: So nevermind.

jonmmease:
> Should all operations try to keep the original Dataset

eventually, yes I think so.  We could extend https://github.com/pyviz/holoviews/pull/3921 to store off the operation kwargs in all elements that are produced by operations.  I haven't considered what it would look like to store operation instances instead of the kwargs, but maybe that is a better approach.

> Also I'm not quite clear when an internal operation should keep hold of the original data or reference the new data

My current thinking is that goal is that:
 - `operations` should produce elements with the full input dataset and the operation arguments needed to reconstruct the element.
 - filtering transformations (`select`, `iloc`, `__getitem__`) should include a reference to the parent dataset that has been filtered just like the element.  So there would be the expectation that `dataset.to(...).select(...).datset` would equal `dataset.select(...)`.
 - For grouping operations that produce a `HoloMap`, the elements of the HoloMap would each have a `.dataset` property that includes only those rows from the source dataset that belong to that group.
 - Any operations/transforms not covered/supported yet can always include `None` as the dataset (or set the dataset to only the `kdims` `vdims` the element knows about).  So I think we should be able to add support for more situations later without breaking compatibility.

> What would be kind of nice is if along with the data we could keep a dim expressions (for select/groupby operations) or operation instances (for operations) which let you replay the whole data provenance starting with original dataset.

This is an interesting idea.  I think this is more than would be needed for the selection linking application I have in mind, but may be useful for other use-cases.  It sounds like this would be pretty close to defining a full lazy version of HoloViews (even if we still perform the operations strictly, there would be a sort of lazy version on the side).

Huh, if we had that I suppose `.data`/`.dimensions_values()` could even be evaluated lazily.

jonmmease:
@poplarShift Yes, I think this handles the case you're talking about. The `.dataset` property is always a `Dataset` instance (*not* a subclass).  That said, if you have time to play with this branch and see if it behaves how you would expect, that would be a great help!

jlstevens:
This PR looks good to me! Like Philipp, I'm wondering whether or not this relates to a more complete provenance mechanism but that would be outside of the scope for what is needed here.

Looking at the code, it looks like `.dataset` is a property with a getter but no setter. My only other question is whether there is a valid use for a setter? E.g maybe you want to set `.dataset` to `None` explicitly for some reason or maybe you created an element some other way and want to specify the originating `Dataset` instance after the face?

That is just something extra for you to think about and consider..otherwise I am happy to merge the PR as is.

jonmmease:
>  Looking at the code, it looks like .dataset is a property with a getter but no setter. My only other question is whether there is a valid use for a setter? E.g maybe you want to set .dataset to None explicitly for some reason or maybe you created an element some other way and want to specify the originating Dataset instance after the fact?

One way to do this is to pass `dataset=` as a kwarg to the element constructor. I *think* this already works, but I'll double check that it flows through everywhere.  I personally like the idea of being able to override this in the constructor, but not to be able to mutate it after construction.  Does that sounds like a reasonable compromise?

jlstevens:
> Does that sounds like a reasonable compromise?

That does seem reasonable, yes.

jonmmease:
I just reviewed the code again and yes, passing the `dataset` kwarg to the element constructors is how the `dataset` property is propagated internally, so a user/library could set this property themselves if it is known.

Merging!

PR #3920, opened 2019-08-21
Add dim-expression support in Dataset.select 

jonmmease:
This PR adds a new `selection_expr` argument to the `Dataset.select` method. This argument may be set to a `holoviews.dim` expression as a way to select rows from a dataset using symbolic expressions.

For example:
```python
import holoviews as hv
import pandas as pd
from holoviews import dim
import dask.dataframe as dd

df = pd.DataFrame({
    'a': [1, 1, 3, 3, 2, 2, 0, 0],
    'b': [10, 20, 30, 40, 10, 20, 30, 40],
    'c': ['A', 'A', 'B', 'B', 'C', 'C', 'D', 'D'],
    'd': [-1, -2, -3, -4, -5, -6, -7, -8]
})

ds = hv.Dataset(df)

print(ds.data)
```
```
   a   b  c  d
0  1  10  A -1
1  1  20  A -2
2  3  30  B -3
3  3  40  B -4
4  2  10  C -5
5  2  20  C -6
6  0  30  D -7
7  0  40  D -8
```
```python
print(ds.select(selection_expr=(dim('b') >= 20) & (dim('a') % 2 == 0)).data)
```
```
   a   b  c  d
5  2  20  C -6
6  0  30  D -7
7  0  40  D -8
```

As long as the expressions are compatible with Dask DataFrames, this also works for `Dataset`s backed by Dask dataframes.

```python
ddf = dd.from_pandas(df, npartitions=2)
dds = ds.clone(data=ddf)
print(dds.select(selection_expr=(dim('b') >= 30)).data)
```
```
Dask DataFrame Structure:
                   a      b       c      d
npartitions=2                             
0              int64  int64  object  int64
4                ...    ...     ...    ...
7                ...    ...     ...    ...
Dask Name: getitem, 8 tasks
```
```python
print(dds.select(selection_expr=(dim('b') >= 30)).data.compute())
```
```
   a   b  c  d
2  3  30  B -3
3  3  40  B -4
6  0  30  D -7
7  0  40  D -8
```

Not all expressions currently work with Dask data structures. This is something I'd like to improve, but I think that should be in a separate PR.

@jlstevens @philippjfr

### DIFF
--- a/holoviews/plotting/util.py
+++ b/holoviews/plotting/util.py
@@ -477,7 +477,7 @@ def initialize_unbounded(obj, dimensions, key):
     """
     select = dict(zip([d.name for d in dimensions], key))
     try:
-        obj.select([DynamicMap], **select)
+        obj.select(selection_specs=[DynamicMap], **select)
     except KeyError:
         pass
 
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -335,7 +335,7 @@ class Dataset(Element):
         return self.clone(data, **dimensions)
 
 
-    def select(self, selection_specs=None, **selection):
+    def select(self, selection_specs=None, selection_expr=None, **selection):
         """Applies selection by dimension name
 
         Applies a selection along the dimensions of the object using
@@ -360,11 +360,18 @@ class Dataset(Element):
 
             ds.select(x=[0, 1, 2])
 
+        * predicate expression: A holoviews.dim expression, e.g.:
+
+            from holoviews import dim
+            ds.select(selection_expr=dim('x') % 2 == 0)
+
         Args:
             selection_specs: List of specs to match on
                 A list of types, functions, or type[.group][.label]
                 strings specifying which objects to apply the
                 selection on.
+            selection_expr: holoviews.dim predicate expression
+                specifying selection.
             **selection: Dictionary declaring selections by dimension
                 Selections can be scalar values, tuple ranges, lists
                 of discrete values and boolean arrays
@@ -378,10 +385,21 @@ class Dataset(Element):
         selection = {dim: sel for dim, sel in selection.items()
                      if dim in self.dimensions()+['selection_mask']}
         if (selection_specs and not any(self.matches(sp) for sp in selection_specs)
-            or not selection):
+                or (not selection and not selection_expr)):
             return self
 
-        data = self.interface.select(self, **selection)
+        # Handle selection dim expression
+        if selection_expr is not None:
+            mask = selection_expr.apply(self, compute=False, keep_index=True)
+            dataset = self[mask]
+        else:
+            dataset = self
+
+        # Handle selection kwargs
+        if selection:
+            data = dataset.interface.select(dataset, **selection)
+        else:
+            data = dataset.data
 
         if np.isscalar(data):
             return data
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -453,7 +453,7 @@ class Dataset(Element):
                object.
         """
         slices = util.process_ellipses(self, slices, vdim_selection=True)
-        if isinstance(slices, np.ndarray) and slices.dtype.kind == 'b':
+        if getattr(getattr(slices, 'dtype', None), 'kind', None) == 'b':
             if not len(slices) == len(self):
                 raise IndexError("Boolean index must match length of sliced object")
             return self.clone(self.select(selection_mask=slices))
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -335,7 +335,7 @@ class Dataset(Element):
         return self.clone(data, **dimensions)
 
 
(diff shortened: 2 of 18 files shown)


jonmmease:
The current (without this PR) API signature for `select` looks like this:
```python
def select(self, selection_specs=None, **selection):
    ...
```
So right now, selection are only specified as kwargs, and the first positional argument is treated as a selection specification that I believe configures the nested application of selections.

I went with `selection_expr` since there was already `selection_specs` and `selection_mask` in the API. One reason to err on the side of being wordy here is that this name lives in the same "namespace" as the dimensions names being selected.  So right now, if there is a dimensions named `selection_specs` it's not possible to reference it in a standard kwarg-based selection.

jbednar:
Ok, that's a good reason for having a longer name, but isn't `selection_expr` just another kind of `selection_spec`, which would let it be used as a positional argument?

philippjfr:
No selectiob_spec specifies which object the selection applies to, selection_expr applies to the actual selected objects.

jbednar:
Ah, ok; I had to go back and look at examples to see what was up here; I was only remembering part of how to use `.select()`.  As I see it, `ds.select(b=20)` works fine already due to kwargs processing, but we would also want to do things like `ds.select(b=>20)`, which wouldn't be properly evaluated by Python.  That's what `dim` is for, so can't we get `ds.select(dim('b') >= 20)` to work, without requiring the awkward and verbose `ds.select(selection_expr=(dim('b') >= 20))`?

To make that work, it seems like all we'd need to do is to make `selection_expr` be the first argument in the signature, rather than `selection_specs`.  That would break compatibility for anyone who had used `selection_specs` as  a positional argument, but maybe (a) that's rare (I hope!) and (b) could we detect someone trying to pass a `selection_spec` to `selection_expr` and raise an error?  Having this work smoothly and cleanly seems important to me for making it easily usable in practice...

jonmmease:
I'd be happier having `selection_expr` as the first argument for the usability reasons you mentioned.  

One thing that we could do would be to make `selection_expr` the first argument, but send the value along the `selection_spec` codepath if the `selection_expr` object isn't a `dim` expression. Though this is probably too much magic.

@philippjfr @jlstevens Do you have strong feelings about the compatibility/convenience trade-off here?

jbednar:
I think that's too much magic to happen silently, but it seems reasonable if accompanied by a warning message saying that people need to use the `selection_specs` keyword argument instead.

jonmmease:
Alright, in 6c1937a I moved `selection_expr` to the first positional arg and raise an exception referring users to `selection_specs` if the argument is not  `dim` expression.

The first example above may now be written as:

```python
ds.select((dim('b') >= 20) & (dim('a') % 2 == 0))
```

philippjfr:
>Do you have strong feelings about the compatibility/convenience trade-off here?

I think usage of ``selection_specs`` is pretty rare tbh, selections already only apply if they match any dimensions. I think there are one or two places in the plotting code that use it. I wonder if we could check if the first argument is a ``dim`` expression and if not issue a warning about specs no longer being the first argument. I'd also want to rename ``selection_specs`` -> ``specs`` (aliasing for now maybe?) and then simply call ``selection_expr`` ``expr``.

jbednar:
> I'd also want to rename ``selection_specs`` -> ``specs`` (aliasing for now maybe?) and then simply call ``selection_expr`` ``expr``.

But with the simpler names, won't it conflict with possible dimension names as Jon mentioned?  Seems likely there will be data sources with columns named `expr` and possibly `specs`; `selection_specs` and `selection_expr` seem much less likely.

philippjfr:
Oh right, missed that and yes that's a good point.

jlstevens:
Reading the above before looking at the code, I am happy to see that all the  suggestions I was thinking of were proposed and accepted. `selection_expr` is indeed ugly but it also needs to avoid clashes with dimension names so making it the first positional argument makes sense to me. 

I agree that usage of `selection_specs` is pretty rare and I am happy with any reasonable way of either warning about potential mistakes or detecting the type of the argument and processing it accordingly. Now I'm going to look through the diff but so far it all looks good!

holoviews/plotting/util.py
```diff
@@ -477,7 +477,7 @@ def initialize_unbounded(obj, dimensions, key):
     """
     select = dict(zip([d.name for d in dimensions], key))
     try:
-        obj.select([DynamicMap], **select)
+        obj.select(selection_specs=[DynamicMap], **select)
```
jlstevens: I count only four times where `selection_specs` had to be specified as a keyword instead of by position! If that is how often it was used that positional argument in our own codebase, I am pretty certain users barely used it (if at all).

jlstevens:
Looks good to me!

jbednar:
It looks like the appveyor failures are independent of this PR?  Seems like they need to be addressed in any case. Otherwise this seems mergeable.

jonmmease:
Thanks for the feedback all. Merging!

PR #3921, opened 2019-08-22
Added operation_kwargs Histogram property

jonmmease:
## Overview
For `Histogram` elements created using the `holoviews.operation.histogram` operation, this new property holds a dict of kwargs that can be used with the operation to recreate the `Histogram` from the same data set.

Or it will contain `None`, if the Histogram was created directly using the Histogram constructor.

## Motivation
Having this information available makes it possible to create histograms with a subset of the original dataset that share the same properties as an original `Histogram`.  For example:

```python
import numpy as np
import holoviews as hv
from holoviews import opts
hv.extension('plotly')

np.random.seed(1)
points = hv.Points(np.random.randn(100,2))

hist1 = hv.operation.histogram(points, cumulative=True, num_bins=10, dynamic=False, normed=False)
hist1.operation_kwargs
```
```
{'cumulative': True,
 'dynamic': False,
 'normed': False,
 'bins': [-1.857981864446752,
  -1.419351107321437,
  -0.980720350196122,
  -0.542089593070807,
  -0.10345883594549199,
  0.335171921179823,
  0.7738026783051382,
  1.212433435430453,
  1.651064192555768,
  2.0896949496810833,
  2.528325706806398]}
```

```python
hist1
```
newplot

```python
hist2 = hv.operation.histogram(points.select(y=(0, None), x=(0, None)), **hist1.operation_kwargs)
hist1 * hist2
```
newplot-1

This will be especially convenient when combined with https://github.com/pyviz/holoviews/pull/3919, because the original dataset will be available as the `hist1.dataset` property.

TODO:
 - [x] Presever operation_kwargs through `clone`.


jbednar:
Sounds good!

jonmmease:
I made `operation_kwargs` a private property of the `Histogram` element so that we're not committing to a particular API at this point.  I opened https://github.com/pyviz/holoviews/issues/3941 to discuss what we might want to do with the API in the future.

Merging!

PR #3948, opened 2019-09-07
Add `visible` style option for simple plotly/bokeh plots.

jonmmease:
Setting `visible` to `False` will fully hide the individual element trace.

This option is very useful for the linked selection effort...

### DIFF
--- a/holoviews/plotting/bokeh/annotation.py
+++ b/holoviews/plotting/bokeh/annotation.py
@@ -28,7 +28,7 @@ from .util import date_to_integer
 
 class TextPlot(ElementPlot, AnnotationPlot):
 
-    style_opts = text_properties+['color', 'angle']
+    style_opts = text_properties+['color', 'angle', 'visible']
     _plot_methods = dict(single='text', batched='text')
 
     def get_data(self, element, ranges, style):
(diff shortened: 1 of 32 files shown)


jbednar:
Is there a chance to unify this API with Bokeh?

jonmmease:
> Is there a chance to unify this API with Bokeh?

I'm not sure. Does anyone know if Bokeh plot elements have an equivalent property to hide everything?  For the selection work I'm using the `opacity` style options to do this.

jbednar:
I'm not an expert at it, but Bokeh has a variety of related options like muted_alpha, muted_color, selection_glyph.visible, nonselection_glyph.visible, and so on; @philippjfr may be able to chat about it on Monday at see where things line up between the two backends.

philippjfr:
>I'm not sure. Does anyone know if Bokeh plot elements have an equivalent property to hide everything?

Most bokeh elements indeed have a visible property and we should expose it.

jonmmease:
> Most bokeh elements indeed have a visible property and we should expose it.

Ok, thanks. I'll take a look

jonmmease:
Alright, I added the `visible` style property to all of the simple (non-composite) Bokeh element plots.  Basically, for all of the cases where all that is needed is to pass the property along to the single Bokeh glyph representing the element.  I didn't do anything with the composite Bokeh elements to this point, as these will take some actual logic changes.

Sound alright for now?

philippjfr:
>Sound alright for now?

Yes, absolutely.

jbednar:
Great!

PR #3960, opened 2019-09-13
Updated cross-selector example to use Panel

philippjfr:


jbednar:
That's a lot better!  I assume similar treatment is needed for the other Applications:

image

PR #3961, opened 2019-09-15
Update ErrorBars to have horizontal errorbar

DancingQuanta:
Up until now, the `ErrorBars` element only support conveying errors along y-axis.
This PR will update the element to allow selecting between x- and y-axis for the errorbars.
The main user facing change will be addition of `horizontal` param. Setting this param `True` will set the errors to be along x-axis instead of y-axis.

EDIT:
Objectives

- [x] Element Errorbars keyword argument 'horizontal' and tests
- [x] Update plotting classes to use `horizontal` parameter
  - [x] Matplotlib
  - [x] Bokeh
  - [x] Ploty
- [x] Update guides
  - [x] Matplotlib
  - [x] Bokeh
  - [x] Ploty

### DIFF
--- a/holoviews/element/chart.py
+++ b/holoviews/element/chart.py
@@ -117,8 +117,16 @@ class ErrorBars(Chart2dSelectionExpr, Chart):
     """
     ErrorBars is a Chart element representing error bars in a 1D
     coordinate system where the key dimension corresponds to the
-    location along the x-axis and the value dimensions define the
-    location along the y-axis and the symmetric or assymetric spread.
+    location along the x-axis and the first value dimension 
+    corresponds to the location along the y-axis and one or two 
+    extra value dimensions corresponding to the symmetric or 
+    asymetric errors either along x-axis or y-axis. If two value
+    dimensions are given, then the last value dimension will be 
+    taken as symmetric errors. If three value dimensions are given 
+    then the last two value dimensions will be taken as negative and
+    positive errors. By default the errors are defined along y-axis.
+    A parameter `horizontal`, when set `True`, will define the errors
+    along the x-axis.
     """
     group = param.String(default='ErrorBars', constant=True, doc="""
         A string describing the quantity measured by the ErrorBars
@@ -127,6 +135,8 @@ class ErrorBars(Chart2dSelectionExpr, Chart):
     vdims = param.List(default=[Dimension('y'), Dimension('yerror')],
                        bounds=(1, None), constant=True)
 
+    horizontal = param.Boolean(default=False, doc="""
+        Whether the errors are along y-axis (vertical) or x-axis.""")
 
     def range(self, dim, data_range=True, dimension_range=True):
         """Return the lower and upper bounds of values along dimension.
@@ -144,10 +154,11 @@ class ErrorBars(Chart2dSelectionExpr, Chart):
         Returns:
             Tuple containing the lower and upper bound
         """
+        dim_with_err = 0 if self.horizontal else 1
         didx = self.get_dimension_index(dim)
         dim = self.get_dimension(dim)
-        if didx == 1 and data_range and len(self):
-            mean = self.dimension_values(1)
+        if didx == dim_with_err and data_range and len(self):
+            mean = self.dimension_values(didx)
             neg_error = self.dimension_values(2)
             if len(self.dimensions()) > 3:
                 pos_error = self.dimension_values(3)
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -510,20 +510,21 @@ class ErrorPlot(ColorbarPlot):
         if self.static_source:
             return {}, mapping, style
 
-        base = element.dimension_values(0)
-        ys = element.dimension_values(1)
-        if len(element.vdims) > 2:
-            neg, pos = (element.dimension_values(vd) for vd in element.vdims[1:3])
-            lower, upper = ys-neg, ys+pos
-        else:
-            err = element.dimension_values(2)
-            lower, upper = ys-err, ys+err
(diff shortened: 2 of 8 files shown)


DancingQuanta:
Should we have a check for x-axis categorical data? Errorbars only works with numerical data. I don't think anyone will try using the Errorbars with categorical data on x-axis but accidents may happen. Without this check, users will get errors with non-numeric data in numeric operations which might be hard to debug.

philippjfr:
>Should we have a check for x-axis categorical data? Errorbars only works with numerical data.

We could (and probably should) make it work with categorical data but that doesn't have to be this PR.

holoviews/element/chart.py
```diff
@@ -117,8 +117,15 @@ class ErrorBars(Chart2dSelectionExpr, Chart):
     """
     ErrorBars is a Chart element representing error bars in a 1D
     coordinate system where the key dimension corresponds to the
-    location along the x-axis and the value dimensions define the
-    location along the y-axis and the symmetric or assymetric spread.
+    location along the x-axis and a value dimension define the
+    location along the y-axis and one or two extra value dimensions
+    define the symmetric or asymetric spread either along x-axis or
+    y-axis. If two values dimensions are given, then the last value
+    diemnsion will be taken as symmetric errors. If three value
```
philippjfr:
```suggestion
    dimension will be taken as symmetric errors. If three value
```

DancingQuanta:
I believe that as an Element class, the horizontal errorbars work well. However, I am unable to figure out  how to communicate the intention that the errorbars is to be horizontal to the plotting classes. I have included the same param `horizontal` to both element and plotting classes. But the param in the plotting classes appear not to change whether the keyword is included or not.
Please try out the matplotlib notebook.

philippjfr:
>I believe that as an Element class, the horizontal errorbars work well. However, I am unable to figure out how to communicate the intention that the errorbars is to be horizontal to the plotting classes. I have included the same param horizontal to both element and plotting classes. But the param in the plotting classes appear not to change whether the keyword is included or not.

The plotting class has access to the element during plotting, so I don't think you need to duplicate the parameter.

DancingQuanta:
If I remove out the `horizontal` param, it just throw an error that `self.horizontal` is undefined.

DancingQuanta:
i have done much I could.
The only issue is now is that the plotting classes do not take any information about `horizontal` parameter from Element Errorbars. The main complaint is `ErrorPlot have no attribute horizontal`.
Any advice where do I go from here?

holoviews/plotting/bokeh/chart.py
```diff
@@ -510,20 +510,27 @@ def get_data(self, element, ranges, style):
...
+        xs = element.dimension_values(0)
         ys = element.dimension_values(1)
-        if len(element.vdims) > 2:
-            neg, pos = (element.dimension_values(vd) for vd in element.vdims[1:3])
-            lower, upper = ys-neg, ys+pos
+        neg_error = element.dimension_values(2)
+        pos_idx = 3 if len(element.dimensions()) > 3 else 2
+        pos_error = element.dimension_values(pos_idx)
+
+        if self.horizontal:
```
philippjfr: `element.horizontal`

holoviews/plotting/mpl/chart.py
```diff
@@ -176,18 +172,18 @@ def update_handles(self, key, axis, element, ranges, style):
...
-        samples = len(xs)
-        pos_error = element.dimension_values(3) if len(element.dimensions()) > 3 else neg_error
-        if self.invert_axes:
-            bxs, bys = ys - neg_error, xs
-            txs, tys = ys + pos_error, xs
-            new_arrays = [np.array([[bxs[i], xs[i]], [txs[i], xs[i]]])
-                          for i in range(samples)]
+        pos_idx = 3 if len(element.dimensions()) > 3 else 2
+        pos_error = element.dimension_values(pos_idx)
+        if self.horizontal:
```
philippjfr:
```suggestion
        if element.horizontal:
```

holoviews/plotting/plotly/chart.py
```diff
@@ -150,13 +150,14 @@ class ErrorBarsPlot(ChartPlot, ColorbarPlot):
 
     def get_data(self, element, ranges, style):
         x, y = ('y', 'x') if self.invert_axes else ('x', 'y')
+        error_k = 'error_' + x if self.horizontal else 'error_' + y
```
philippjfr:
```suggestion
        error_k = 'error_' + x if element.horizontal else 'error_' + y
```

DancingQuanta: Ah, so what it is for!

DancingQuanta:
I am all done here. I have tested with `invert_axes` which shows good behaviour in plots.
@philippjfr thank you for your guidance. This PR can be merged.

philippjfr:
I'm happy to merge this, but there's one thing I wanted to confirm. Does this also work for categorical axes?

I'm also okay with adding the `horizontal` parameter to the element itself and not a plot option because it's semantic information about the data not merely how to visualize it.

DancingQuanta:
Good question about categorical axes. I did try an experiment where I set `horizontal=True` once and checking the CI tests.  There numerous errors in tests involving categorical data. Most of the errors involved trying to compute lower and upper points of the errorbars from the mean and error. In this case the mean dimension is the x-axis which is the categorical axis. So basically arithmetic operation between strings and numbers was attempted here and failed expectedly.
There may need to be a guard against this, perhaps at the initialisation of the element checking for the type of the data in the supplied dimensions.

However, I have not tested the cases where we have numerical data with errors (y, yerr) against categorical data (x). I will try to test this with one of the example notebooks with categorical data.

philippjfr:
I think it's fine tbh, going to go ahead and merge.

PR #3951, opened 2019-09-11
Add link_selections function for automated linked selections

jonmmease:
## Overview
This PR adds a `holoviews.selection.link_selections` function that is an initial implementation of the automatic linked selection approach proposed in https://github.com/pyviz/holoviews/issues/3842.

This function can be used to transform HoloViews objects into new objects with linked selections enabled.

### Dependencies
This PR depends on the following open PRs, and will be rebased once these are merged.
 - https://github.com/pyviz/holoviews/pull/3924
 - https://github.com/pyviz/holoviews/pull/3927

## Example notebook
For examples, see the notebook at https://anaconda.org/jonmmease/link_selections_pr/notebook.

Please play around with this notebook to get a feel for the workflow and capabilities and report back on how things look @jbednar @philippjfr @jlstevens.

## Highlights
Linked `Points` plots with Bokeh backend:
points_layout

---
Linked `Points`, `Histogram`, `Scatter3d`, and `Table` layout with Plotly backend:
plotly_selections

---
Linked `Violin`, `Points` and `Bivariate` overlay, and `Distribution` layout with Bokeh backend:
bivariate_selections

---
Selections on datashaded elements
datashader_selection

---
Linked plots across notebook output cells, and updating selection color after display:
cross_cell_linking

## Extensions
There is a lot more functionality that can be built on this foundation. Here are some directions:

 - The `link_selections` function only stores the latest selection. But the `_base_link_selections` base class is more general, and other subclasses can be added for cross filtering, and glue-style multi-color brushing.
    - For cross-filtering, I really like the 4-color approach used in https://plotly.github.io/plotly.js-crossfilter.js/.  I personally find this easier to make sense of compared to the 2-color approach used in http://crossfilter.github.io/crossfilter/ where the "filtered" data is fully removed from the plot.
    - For multi-colored brushing, the inspiration is the glue project http://docs.glueviz.org/en/stable/getting_started/index.html?highlight=subsets#defining-subsets.  This would require a small Panel GUI interface for choosing the current subset color and the subset mode (union, intersection, or subtraction).
 - Add `Bounds` element to display the selected region after a selection is completed. I think this will largely be a requirement for the cross-filter mode, otherwise it would be hard to make sense of the union of the selected regions across plots.
 - Selections could be displayed on grid elements (e.g. vector field, raster) if we had an `element.where` operation (https://github.com/pyviz/holoviews/issues/3933).
[...]

PR #3967, opened 2019-09-18
Add pipeline property to track data lineage

jonmmease:
## Overview
This PR adds a new `pipeline` property to the `Dataset` class.  This property holds a list of `(function, args, kwargs)` tuples that represent the sequence of operations needed to transform the `Dataset` stored in the `dataset` property into an element equal to current element.

It also adds a new `execute_pipeline` method that can evaluate this sequence of functions on an input dataset.  This makes it possible to reproduce the same sequence of operations on a new Dataset.

### dataset property
The `dataset` property was added to the `LabelledData` class in https://github.com/pyviz/holoviews/pull/3919. This PR moves the `dataset` property down to the `Dataset` class, so there is no longer a `dataset` property on, for example, the `Layout` class.  This reduces the scope of where `dataset` and `pipeline` need to be correct / consistent.

### Histogram _operation_kwargs
This PR removes all special cases associated with Histogram elements.  So the `Histogram._operation_kwargs` property added in https://github.com/pyviz/holoviews/pull/3921 has been removed.

### select all dims
In https://github.com/pyviz/holoviews/pull/3924, the `select` method is updated to consider all dimensions in the `Dataset` stored in the element's `dataset` property.  This PR does not do this, and instead provides the `execute_pipeline` method as a more powerful alternative to acheiving the same goal. See examples below.

### link_selections
This PR will become a more powerful foundation for the automatic linked selection support being added in https://github.com/pyviz/holoviews/pull/3951

## Example 1: Points
Create a sample 3-dimensional dataset. x and y are independently drawn from the standard normal distribution and r is calculated to be the radius of each point from the origin.
```python
import numpy as np
import pandas as pd
import holoviews as hv
from holoviews import dim
from holoviews.operation.datashader import rasterize, datashade, dynspread 
hv.extension('bokeh')

np.random.seed(1)
df = pd.DataFrame(np.random.randn(100, 2), columns=['x', 'y'])

# Add radius column
df['r'] = (df.x ** 2 + df.y ** 2) ** 0.5

ds = hv.Dataset(df)
points = ds.to.points(kdims=['x', 'y'], groupby=[])
points
```
bokeh_plot-1

Display the pipeline for the new `points`  element
```python
points.pipeline
```
```
[(holoviews.core.data.Dataset, [], {}),
 (holoviews.element.geom.Points,  [],  {'label': '', 'kdims': [Dimension('x'), Dimension('y')], 'vdims': []})]
```
Next, create a new points element by running `execute_pipline` on a subset of the dataset stored in `points.dataset`.  Note that it would not be possible to compute this subset using `points.select` directly because it involves the `r` dimensions which is not a key or value dimension of `points`.

```python
points * points.execute_pipeline(points.dataset.select(x=(0, None), r=(0, 1.5))) 
```
bokeh_plot-2

## Example 2: Datashade
Create an `RGB` image element from `points` using the `datashade` and `dynspread` operations with `dynamic=False`.

```python
points_rgb = dynspread(datashade(points, dynamic=False), dynamic=False, threshold=0.9)
points_rgb
```
bokeh_plot-3

Display the pipeline for `points_rgb`

```python
points_rgb.pipeline
```
```
[(holoviews.core.data.Dataset, [], {}),
 (holoviews.element.geom.Points,  [],  {'label': '', 'kdims': [Dimension('x'), Dimension('y')], 'vdims': []}),
 (datashade(...),  [],  {'dynamic': False}),
 (dynspread(...),  [],  {'dynamic': False, 'threshold': 0.9})]
```
Next, compute a new `RGB` element by calling the `execute_pipeline` method with a subset of the original dataset.  Note that this is a selection that was not possible using the approach in https://github.com/pyviz/holoviews/pull/3924.

```python
points_rgb + points_rgb.execute_pipeline(points_rgb.dataset.select(x=(0, None), r=(0, 1.5)))
```
Screenshot_20190918_054413

## Example 3: Histogram
Next, repeat the same process using a `Histogram` element created from `points`.

```python
hist1 = hv.operation.histogram(points, num_bins=10, dynamic=False, normed=False)
hist1
```
bokeh_plot-4

Display pipeline
```python
hist1.pipeline
```
```
[(holoviews.core.data.Dataset, [], {}),
 (holoviews.element.geom.Points,  [],  {'label': '', 'kdims': [Dimension('x'), Dimension('y')], 'vdims': []}),
 (histogram(...),  [],  {'num_bins': 10, 'dynamic': False, 'normed': False})]
```
Create new `Histogram` element with `execute_pipeline`

```python
hist2 = hist1.execute_pipeline(hist1.dataset.select(x=(0, None), r=(0, 1.5))) 
hist1 * hist2
```
bokeh_plot-5

## Example 4: Custom aggregation
In this example, create a `Bars` element from the result of aggregating an original `Dataset`.

```python
df = pd.DataFrame({'a': [1, 2, 3, 4, 5, 6],
                   'b': [2, 1, 3, 0, 10, 4],
                   'c': [0, 0, 0, 1, 1, 1]
                  })
ds = hv.Dataset(df, kdims=['c'], vdims=['a', 'b'])
bars = ds.aggregate('c', function=np.sum).to(hv.Bars)
bars
```
bokeh_plot-6

pipeline
```python
bars.pipeline
```
```
[(holoviews.core.data.Dataset,
  [],
  {'kdims': [Dimension('c')], 'vdims': [Dimension('a'), Dimension('b')]}),
 (<function holoviews.core.data.Dataset.aggregate(...)>,
  ['c'],
  {'function': <function numpy.sum(...)>}),
 (holoviews.element.chart.Bars,
  [],
  {'label': '',
   'kdims': [Dimension('c')],
   'vdims': [Dimension('a'), Dimension('b')]})]
```
Create a new `Bars` element on a subset of the original dataset using `execute_pipeline`

```python
bars * bars.execute_pipeline(bars.dataset.select(b=(3, None)))
```
bokeh_plot-7

### DIFF
--- a/holoviews/core/accessors.py
+++ b/holoviews/core/accessors.py
@@ -58,7 +58,6 @@ class AccessorPipelineMeta(type):
                                 init_op, call_op
                             ],
                             output_type=type(result),
-                            group=result.group
                         )
                     elif isinstance(result, MultiDimensionalMapping):
                         for key, element in result.items():
@@ -72,7 +71,6 @@ class AccessorPipelineMeta(type):
                                     init_op, call_op, getitem_op
                                 ],
                                 output_type=type(result),
-                                group=element.group
                             )
             finally:
                 if not in_method:
--- a/holoviews/core/accessors.py
+++ b/holoviews/core/accessors.py
@@ -4,6 +4,7 @@ Module for accessor objects for viewable HoloViews objects.
 from __future__ import absolute_import, unicode_literals
 
 from collections import OrderedDict
+import copy
 
 import param
 from param.parameterized import add_metaclass
@@ -30,6 +31,7 @@ class AccessorPipelineMeta(type):
                 # Wrapped object doesn't support the pipeline property
                 return __call__(*args, **kwargs)
 
+            inst_pipeline = copy.copy(inst._obj. _pipeline)
             in_method = inst._obj._in_method
             if not in_method:
                 inst._obj._in_method = True
@@ -39,13 +41,13 @@ class AccessorPipelineMeta(type):
             if not in_method:
                 mode = getattr(inst, 'mode', None)
                 if isinstance(result, Dataset):
-                    result._pipeline = inst._obj._pipeline + [
+                    result._pipeline = inst_pipeline + [
                         (type(inst), [], {'mode': mode}),
                         (__call__, list(args[1:]), kwargs)
                     ]
                 elif isinstance(result, MultiDimensionalMapping):
                     for key, element in result.items():
-                        element._pipeline = inst._obj._pipeline + [
+                        element._pipeline = inst_pipeline + [
                             (type(inst), [], {'mode': mode}),
                             (__call__, list(args[1:]), kwargs),
                             (getattr(type(result), '__getitem__'), [key], {})
--- a/holoviews/core/accessors.py
+++ b/holoviews/core/accessors.py
@@ -23,31 +23,31 @@ class AccessorPipelineMeta(type):
 
     @classmethod
     def pipelined(mcs, __call__):
-        def pipelined_call(*a, **k):
+        def pipelined_call(*args, **kwargs):
             from .data import Dataset, MultiDimensionalMapping
-            inst = a[0]
+            inst = args[0]
             if not hasattr(inst._obj, '_pipeline'):
                 # Wrapped object doesn't support the pipeline property
-                return __call__(*a, **k)
+                return __call__(*args, **kwargs)
 
             in_method = inst._obj._in_method
             if not in_method:
                 inst._obj._in_method = True
 
-            result = __call__(*a, **k)
+            result = __call__(*args, **kwargs)
 
             if not in_method:
                 mode = getattr(inst, 'mode', None)
                 if isinstance(result, Dataset):
                     result._pipeline = inst._obj._pipeline + [
                         (type(inst), [], {'mode': mode}),
-                        (__call__, list(a[1:]), k)
+                        (__call__, list(args[1:]), kwargs)
                     ]
                 elif isinstance(result, MultiDimensionalMapping):
                     for key, element in result.items():
                         element._pipeline = inst._obj._pipeline + [
                             (type(inst), [], {'mode': mode}),
-                            (__call__, list(a[1:]), k),
+                            (__call__, list(args[1:]), kwargs),
                             (getattr(type(result), '__getitem__'), [key], {})
                         ]
                 inst._obj._in_method = False
--- a/holoviews/core/accessors.py
+++ b/holoviews/core/accessors.py
@@ -162,7 +162,7 @@ class Apply(object):
                     mapped.append((k, new_val))
             return self._obj.clone(mapped, link=link_inputs)
 
-        
+
     def aggregate(self, dimensions=None, function=None, spreadfn=None, **kwargs):
         """Applies a aggregate function to all ViewableElements.
 
@@ -226,7 +226,7 @@ class Redim(object):
             list: List of dimensions with replacements applied
         """
         from .dimension import Dimension
-        
+
         replaced = []
         for d in dimensions:
             if d.name in overrides:
--- a/holoviews/core/accessors.py
+++ b/holoviews/core/accessors.py
@@ -26,6 +26,7 @@ class AccessorPipelineMeta(type):
     @classmethod
     def pipelined(mcs, __call__):
         def pipelined_call(*args, **kwargs):
+            from ..operation.element import method as method_op, factory
             from .data import Dataset, MultiDimensionalMapping
             inst = args[0]
             if not hasattr(inst._obj, '_pipeline'):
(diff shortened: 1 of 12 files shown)


philippjfr:
This is pretty much exactly what I expected when we discussed this so I'm very happy to see it seems to have worked. The ``_in_method`` flag is also what I imagined and it should handle nested method calls but I'm wondering I haven't yet spotted how it works for ``.apply(operation, ...)`` calls for instance, does the operation get added twice in that case?

jonmmease:
Yeah, thanks for working through the design with me!

In terms of `apply`, since this is an accessor (not a method) it doesn't cause `_in_method` to be set.
```python
points.apply(hv.operation.histogram).pipeline
```
```
[(holoviews.core.data.Dataset, [], {}),
 (holoviews.element.geom.Points,
  [],
  {'label': '', 'kdims': [Dimension('x'), Dimension('y')], 'vdims': []}),
 (histogram(...),
  [],
  {'dynamic': False})]
```

But that does remind me that I should add some tests for `apply`.

And, there might be a hole here if the think `apply` calls is not already an `operation`.  I'll take a look.

jonmmease:
> And, there might be a hole here if the thing apply calls is not already an operation. I'll take a look.

No, I don't think this is a problem. We only need to update the pipeline if the function passed to apply returns a `Dataset` object, and to do this the function would call out to an operation or call a method on the object.

```python
points.apply(lambda p: p.select(x=(0, None))).pipeline
```
```
[(holoviews.core.data.Dataset, [], {}),
 (holoviews.element.geom.Points,
  [],
  {'label': '', 'kdims': [Dimension('x'), Dimension('y')], 'vdims': []}),
 (<function holoviews.core.data.Dataset.select(self, selection_expr=None, selection_specs=None, **selection)>,
  [],
  {'x': (0, None)})]
```

Hmm, there are also the `opts` and `redim` accessors. Do you think these should be captured in the pipeline?

philippjfr:
>We only need to update the pipeline if the function passed to apply returns a Dataset object, and to do this the function would call out to an operation or call a method on the object.

I frequently write functions that take an object compute something from it and then repack a new Dataset, e.g. here's an apply functions I just wrote for a dashboard I'm writing:

```
    def get_table(ds):
        arr = ds.array()
        weights = list(zip(stocks.columns, arr[0, 2:])) if len(arr) else []
        return hv.Table(weights, 'Stock', 'Weight').opts(editable=True)
```

>Hmm, there are also the opts and redim accessors

``redim`` should definitely be captured since without it the pipeline might be invalid. I have no strong opinion on opts but for completeness sake I guess we should do it.

jonmmease:
Ok, yeah. That's a good point regarding the apply function constructing a brand new object. So this will need to be captured separately from the `PipelineMeta` metaclass. Which is fine.

I'll work on `apply`, `redim`, and `opts`, next.  Let me know if any other cases come to mind.  Right now the following are covered:

 - Dataset methods
 - Operation subclasses
 - iloc and ndloc accessors

jonmmease:
In 702e531 I added a new meta class to support pipelines in `apply`, `redim`, and `opts` accessors.

```python
points.apply(
    lambda p: hv.Points(p.select(x=(0, None)).data)
).redim.label(x="The X Dim").opts(color='green').pipeline
```
```
[(holoviews.core.data.Dataset, [], {}),
 (holoviews.element.geom.Points,  [],  {'label': '', 'kdims': [Dimension('x'), Dimension('y')], 'vdims': []}),
 (holoviews.core.accessors.Apply, [], {'mode': None}),
 (<function holoviews.core.accessors.Apply.__call__(...)>,  [<function __main__.<lambda>(p)>],  {}),
 (holoviews.core.accessors.Redim, [], {'mode': 'dataset'}),
 (<function holoviews.core.accessors.Redim.__call__(...)>,  [None],  {'x': {'label': 'The X Dim'}}),
 (holoviews.core.accessors.Opts, [], {'mode': None}),
 (<function holoviews.core.accessors.Opts.__call__(...)>,  [],  {'color': 'green'})]
```

johnzzzzzzz:
Jon,
This feature looks great!
I am personally interested in model-view-control links between different hvplot diagrams. 
It looks like a view's (hv.Points) pipeline could be used to update a view, when the model selection (DataSet.select) changes. 
Could you show an example of that working using bokeh backend?
For example 1, if the two Points Elements where in a Layout instead of an Overlay. Then if the ds Dataset had rows selected, these rows would be selected in both Elements. That is all the selected points would be displayed in the first Points, but only the rows selected and then passed through the pipeline would be displayed on the second Points.

jonmmease:
Hi @johnzzzzzzz,

Have you seen https://github.com/pyviz/holoviews/pull/3951? This is work towards creating a workflow to automatically link selections between HoloViews elements (including those produced by hvplot).  The next iteration of that PR is going to build on top of this pipeline work.

johnzzzzzzz:
I am excited about #3951 and would like to help create test cases. 
I will try to figure out how to clone a branch that includes this and #3951 
This feature may also help [panel 604](https://github.com/pyviz/panel/issues/604) not supporting linked Elements

jbednar:
I'm excited too. Would it be possible for `obj.pipeline` to work as it does above (returning the list) while `obj.pipeline()` does what is currently invoked with `obj.execute_pipeline()`?  Having words like `execute` in a function call slightly annoys me, because every function call executes something, so it seems sufficient to convey "calling" with the standard Python `()` call syntax alone.  But on a quick glance at the property I can't tell if overloading it in that way would work, so I'm just proposing it here if it's possible.

jonmmease:
> Would it be possible for obj.pipeline to work as it does above (returning the list) while obj.pipeline()

I don't *think* so, unless the thing returned by `obj.pipeline` isn't a standard Python list.  We could return some object of our own that both represents the pipeline and evaluates it with `__call__`.  But I'm not sure how intuitive this would be for users.

I'm definitely open to renaming `execute_pipeline` though!

jbednar:
That's what I suspected.  It would be easy to have something that prints like a list while being callable, but I agree that it's nicer to have it simply _be_ a list when it's returned as a value.  I don't have any suggestions for a better name, then.

philippjfr:
Looks good! I don't actually much like the ``group`` handling of operations. I think we should set the group default to None in most cases and then skip setting the `group` is it is None, e.g. in ``chain``

```python
return processed if self.p.group is None else processed.clone(group=self.p.group) 
```

jonmmease:
> Looks good!

Thanks! It is nice for pipeline to be a standard `chain` operation.

> I don't actually much like the group handling of operations. I think we should set the group default to None in most cases and then skip setting the group is it is None

That sounds good to me. I made the change in the `chain` operation in 8dd49c940416b880a762c37145db361fce96091d.

I *think* this PR is in pretty good shape now. Thanks for taking a look, and let me know if anything else comes to mind that we should do before merging.

philippjfr:
Happy to see this merged. I'll give @jlstevens a chance to review though.

philippjfr:
Okay, since he's on PTO for the foreseeable future I'm going to go ahead and merge.

jonmmease:
Thanks!

PR #3975, opened 2019-09-19
Resolve dependent functions in Dynamic operations

philippjfr:
This makes it possible to provide a function which declares param dependencies in ``.apply`` and in operations, e.g.:

```python
@pn.depends(field.param.value, agg_fn.param.value)
def aggregator(field, agg_fn):
    field = None if field == "counts" else field
    return agg_fn(field)

datashade(..., agg=aggregator)
# or
obj.apply(datashade, agg=aggregator)
```

This complements the ability to use functions with dependencies as ``DynamicMap`` callbacks.

Cc: @jbednar

### DIFF
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -921,6 +921,11 @@ class Dynamic(param.ParameterizedFunction):
         for k, v in self.p.kwargs.items():
             if util.is_param_method(v):
                 v = v()
+            elif isinstance(v, FunctionType) and hasattr(v, '_dinfo'):
+                deps = v._dinfo
+                args = (getattr(p.owner, p.name) for p in deps.get('dependencies', []))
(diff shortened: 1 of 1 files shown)


jbednar:
Excellent!  I can't test it because my new laptop doesn't have Xcode yet and I'm having Apple ID problems, but it certainly looks like it would work...

jbednar:
Ok, I've tested it and it works perfectly!  You can see the result at https://anaconda.org/jbednar/dashboard_barewidgets , comparing it to the Param-based version at http://datashader.org/dashboard.html .   The main remaining awkwardness in that notebook is the constant `.param.value` invocations, which would be addressed by https://github.com/pyviz/holoviews/issues/3976 and changes already in Panel master that will be in 0.7.0.

PR #3977, opened 2019-09-20
Correctly resample colorcet colormaps

philippjfr:
Fixes  https://github.com/pyviz/holoviews/issues/3964

Cc: @jsignell @jbednar


jbednar:
Thanks! Instead of hardcoding "glasbey" here, maybe it would be more general and extensible to use the database of metadata that we already have about the colormaps? (`hv.plotting.util.list_cmaps("Categorical")`)

PR #3980, opened 2019-09-20
Ensure Dynamic utility subscribes to dependent function

philippjfr:
Fixes issue with #3975 

While dependent functions were resolved without this, the DynamicMap did not actually subscribe changes in the dependencies and therefore wouldn't trigger an update until something else changed.


jbednar:
Thanks! I hadn't noticed any problem in my quick testing, so I'm glad you caught this!

PR #3982, opened 2019-09-21
Add cuDFInterface to work with cuDF GPU dataframes and cupy support for XArrayInterface

philippjfr:
This PR adds a new data interface to allow HoloViews to work directly with cuDF GPU dataframes.

- [x] values
- [x] range
- [x] dframe
- [x] select
- [x] groupby (look at optimizing) 
- [x] aggregate
- [x] iloc
- [x] add_dimension
- [x] sort
- [x] sample
- [x] concat

Datashader support:

- [x] Points/Scatter
- [x] Curve/Path/Area
- [x] QuadMesh: rectilinear and curvilinear

Other things to do:

- [x] Implement cuDatashader support once integrated with datashader itself

One major thing to figure out is how we will set up CI tests for the GPU.

### DIFF
--- a/holoviews/core/data/cudf.py
+++ b/holoviews/core/data/cudf.py
@@ -121,7 +121,6 @@ class cuDFInterface(PandasInterface):
         return data, {'kdims':kdims, 'vdims':vdims}, {}
 
 
-
     @classmethod
     def range(cls, dataset, dimension):
         column = dataset.data[dataset.get_dimension(dimension, strict=True).name]
@@ -239,7 +238,8 @@ class cuDFInterface(PandasInterface):
             selection_mask = cls.select_mask(dataset, selection)
 
         indexed = cls.indexed(dataset, selection)
-        df = df[selection_mask]
+        if selection_mask is not None:
+            df = df[selection_mask]
         if indexed and len(df) == 1 and len(dataset.vdims) == 1:
             return df[dataset.vdims[0].name].iloc[0]
         return df
@@ -266,14 +266,18 @@ class cuDFInterface(PandasInterface):
         vdims = dataset.dimensions('value', label='name')
         reindexed = data[cols+vdims]
         agg = function.__name__
-        if agg in ('amin', 'amax'):
-            agg = agg[1:]
-        if not hasattr(data, agg):
-            raise ValueError('%s aggregation is not supported on cudf DataFrame.' % agg)
         if len(dimensions):
+            agg_map = {'amin': 'min', 'amax': 'max'}
+            agg = agg_map.get(agg, agg)
             grouped = reindexed.groupby(cols, sort=False)
+            if not hasattr(grouped, agg):
+                raise ValueError('%s aggregation is not supported on cudf DataFrame.' % agg)
             df = getattr(grouped, agg)().reset_index()
         else:
+            agg_map = {'amin': 'min', 'amax': 'max', 'size': 'count'}
+            agg = agg_map.get(agg, agg)
+            if not hasattr(reindexed, agg):
+                raise ValueError('%s aggregation is not supported on cudf DataFrame.' % agg)
             agg = getattr(reindexed, agg)()
             data = dict(((col, [v]) for col, v in zip(agg.index, agg.to_array())))
             df = util.pd.DataFrame(data, columns=list(agg.index))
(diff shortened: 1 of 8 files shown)


holoviews/core/data/cudf.py
```diff
@@ -0,0 +1,283 @@
...
+from ...dimension import dimension_name
+from ..element import Element
+from ...ndmapping import NdMapping, item_check, sorted_context
+from .interface import DataError, Interface
+from .pandas import PandasInterface
+
+
+class cuDFInterface(PandasInterface):
+    """
+    The cuDFInterface allows a Dataset objects to wrap a cuDF
```
jbednar:
```suggestion
    The cuDFInterface allows a Dataset object to wrap a cuDF
```

holoviews/core/data/cudf.py
```diff
@@ -0,0 +1,283 @@
...
+    DataFrame object. Using cuDF allows working with columnar
+    data on a GPU. Most operation leave the data in GPU memory,
+    however to plot the data it has to be loaded into memory.
```
jbednar:
```suggestion
    but to plot the data it has to be loaded into CPU memory.
```

holoviews/core/data/cudf.py
```diff
@@ -0,0 +1,283 @@
...
+
+    The cuDFInterface covers almost the complete API exposed
+    by the PandasInterface with two notable exceptions:
+
+    1) Sorting is not supported and any attempt at sorting will
+       be ignored with an warning.
+    2) cuDF does not easily support adding a new column to an existing
+       dataframe unless it is a scalar, add_dimension will therefore
```
jbednar:
```suggestion
       dataframe unless it is a scalar; add_dimension will therefore
```

holoviews/core/data/cudf.py
```diff
@@ -0,0 +1,283 @@
...
+       error when supplied a non-scalar value.
+    3) Not all functions can be easily applied to a dask dataframe so
```
jbednar:
```suggestion
    3) Not all functions can be easily applied to a cuDF dataframe so
```

?

holoviews/core/data/cudf.py
```diff
@@ -0,0 +1,283 @@
...
+        dropped = []
+        for vd in vdims:
+            if vd not in df.columns:
+                dropped.append(vd)
+        return df, dropped
+
+
+    @classmethod
+    def sort(cls, dataset, by=[], reverse=False):
+        raise NotImplementedError('Sorting is not supported by cudf DataFrames.')
```
jbednar: Earlier it said sorting will be ignored with a warning, but this doesn't look like a warning...

philippjfr: Yeah need to change this, dask also just warns.

jonmmease:
> Implement GPU optimized histogram operation (in this PR)

I think we could do this with `cupy.histogram` with something like this:

```python
import cupy
import cudf
cdf = cudf.from_pandas(...)
ca = cupy.array(cdf["col1"].to_gpu_array(), copy=False)
hist, bin_edges = cupy.histogram(ca)
```

We'll probably also want to add a cupy interface for grid data eventually as well, so I'd be in favor of requiring that both `cudf` and `cupy` be installed for any GPU support.

philippjfr:
>We'll probably also want to add a cupy interface for grid data eventually as well

This PR now also supports cupy backed xarrays using a two line change.

philippjfr:
Okay, this PR is now almost ready. I've added tests, which won't run on Travis of course but can be run locally. I'll add some additional tests for datashader integration.

There are one major class of tests that are failing which is groupby/aggregation tests. It seems like cuDF groupby(..., sort=False) results in reverse sorting which seems quite odd. I'll keep looking at that but in the worst case I'll disable those tests until the issue has been fixed upstream.

holoviews/core/data/cudf.py
```diff
@@ -0,0 +1,340 @@
...
+from ..ndmapping import NdMapping, item_check, sorted_context
+from .interface import DataError, Interface
+from .pandas import PandasInterface
+
+
+class cuDFInterface(PandasInterface):
+    """
+    The cuDFInterface allows a Dataset objects to wrap a cuDF
+    DataFrame object. Using cuDF allows working with columnar
+    data on a GPU. Most operation leave the data in GPU memory,
```
jbednar:
```suggestion
    data on a GPU. Most operations leave the data in GPU memory,
```

PR #3984, opened 2019-09-22
Remove topics notebooks

philippjfr:
Since the existing topics are now built on example.pyviz.org I've decided to remove them from the HoloViews repo.

Closes https://github.com/pyviz/holoviews/issues/1638


jbednar:
Closing them seems fine, but it seems like they still ought to be linked from the hv docs.

philippjfr:
Any suggestion on where?

philippjfr:
The gallery I suppose.

jbednar:
Our original idea with examples.pyviz.org is that it would serve as a place to capture and publish examples in a well-defined and stable location that could then be referenced by any project illustrated by that example.  So I assume they can just appear in the HoloViews gallery like any other HoloViews examples...

philippjfr:
Okay, I'm going ahead with this PR. The topics aren't linked currently so adding them is a separate issue.

PR #3985, opened 2019-09-23
Enabled deprecation warnings for all future deprecations

philippjfr:


jlstevens:
Thanks! Some of those warnings have been ready for a very long time so it is good to make sure everyone sees them now and eventually apply the deprecations.

PR #4007, opened 2019-09-28
pprint_value_unit func for printing value and unit

DancingQuanta:
The printing of dimension value with unit is not consistent throughout,
eg missing space between unit and value. This is due to lack of
consistency in logic within construction of the value with unit string.
This PR adds a new function `pprint_value_unit` to `Dimension` based on
`pprint_value_string` that prints just a value and unit, if exists,
with a space between them.
This function replaces the `pprint_value` along with checks for unit.
This concentrates the logic within one place and this will improve 
maintainability

### DIFF
--- a/holoviews/tests/core/testndmapping.py
+++ b/holoviews/tests/core/testndmapping.py
@@ -26,7 +26,9 @@ class DimensionTest(ComparisonTestCase):
     def test_dimension_pprint(self):
         dim = Dimension('Test dimension', cyclic=True, type=float, unit='Twilight zones')
         self.assertEqual(dim.pprint_value_string(3.23451), 'Test dimension: 3.2345 Twilight zones')
-        self.assertEqual(dim.pprint_value_string(4.23441),  'Test dimension: 4.2344 Twilight zones')
+        self.assertEqual(dim.pprint_value_string(4.23441), 'Test dimension: 4.2344 Twilight zones')
+        self.assertEqual(dim.pprint_value(3.23451, print_unit=True), '3.2345 Twilight zones')
+        self.assertEqual(dim.pprint_value(4.23441, print_unit=True), '4.2344 Twilight zones')
 
 
 class NdIndexableMappingTest(ComparisonTestCase):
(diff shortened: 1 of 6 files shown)


philippjfr:
Sorry I left this sit for so long (I thought I'd left a comment but it seems I never hit submit). Instead of a new method I think I'd prefer an argument to the existing function.

holoviews/tests/core/testndmapping.py
```diff
@@ -26,7 +26,9 @@ def test_dimension_clone(self):
     def test_dimension_pprint(self):
         dim = Dimension('Test dimension', cyclic=True, type=float, unit='Twilight zones')
         self.assertEqual(dim.pprint_value_string(3.23451), 'Test dimension: 3.2345 Twilight zones')
-        self.assertEqual(dim.pprint_value_string(4.23441),  'Test dimension: 4.2344 Twilight zones')
+        self.assertEqual(dim.pprint_value_string(4.23441), 'Test dimension: 4.2344 Twilight zones')
+        self.assertEqual(dim.pprint_valueg(3.23451, print_unit=True), '3.2345 Twilight zones')
```
philippjfr:
```suggestion
        self.assertEqual(dim.pprint_values(3.23451, print_unit=True), '3.2345 Twilight zones')
```

holoviews/tests/core/testndmapping.py
```diff
@@ -26,7 +26,9 @@ def test_dimension_clone(self):
...
+        self.assertEqual(dim.pprint_valueg(4.23441, print_unit=True), '4.2344 Twilight zones')
```
philippjfr:
```suggestion
        self.assertEqual(dim.pprint_value(4.23441, print_unit=True), '4.2344 Twilight zones')
```

philippjfr:
Two typos in the tests, otherwise I'm happy to merge.

philippjfr:
@DancingQuanta Any chance you could apply my suggestions? Just need to click "Commit suggestion".

DancingQuanta:
Hi @philippjfr, I have added your suggestions and rebased on current master.

philippjfr:
Much appreciated!

PR #4035, opened 2019-10-07
Fix Image element bounds check

philippjfr:


jsignell:
Nice! Any way to test?

PR #4039, opened 2019-10-07
Add option to rotate second categorical axis tick labels

philippjfr:
Fixes https://github.com/pyviz/holoviews/issues/2515

@jbednar Any suggestion for a better name than `x_group_rotation`?


holoviews/plotting/bokeh/element.py
```diff
@@ -174,6 +174,9 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         default=None, class_=(util.basestring, TickFormatter, FunctionType), doc="""
         Formatter for ticks along the x-axis.""")
 
+    x2rotation = param.Integer(default=None, bounds=(0, 360), doc="""
+        Rotation angle of the xticks.""")
```
jbednar: Rotation angle of _which_ xticks?

jbednar:
> Any suggestion for a better name than x_group_rotation?

It seems like this is rotating the label for a group of objects on the x axis.  If so, why x in particular?  Shouldn't this be supported for both x and y axes?

In any case, `x2rotation` doesn't convey anything to me, nor does `x_group_rotation`; neither are clear they are about rotating a label. It seems like this is a rarely needed option, so the name can be long.  What about `x_group_label_rotation` (and `y_group_label_rotation`)?

philippjfr:
>It seems like this is rotating the label for a group of objects on the x axis. If so, why x in particular? Shouldn't this be supported for both x and y axes?

Technically we only support two-level categorical axes on the x-axis but you **can** use invert_axes to get it to appear on the y-axis so you're right about this sadly.

> It seems like this is a rarely needed option, so the name can be long. What about x_group_label_rotation (and y_group_label_rotation)?

I'd prefer if it was at least somewhat consistent with the existing ``x/yrotation`` options.

jbednar:
`xrotation` and `yrotation` are much more commonly needed, so it makes sense that those are shorter and less precise names (i.e. that they are not `xtick_label_rotation` and `ytick_label_rotation`, which is presumably what they mean).  Ok, treating those names as fixed, how about `group_xrotation`?

jbednar:
`xrotation_toplevel`?  `outer_xrotation`?  `xrotation_level2`?

PR #4044, opened 2019-10-08
support `link_selections` with Dask DataFrames

jonmmease:
Adds support for using the new `link_selections` transformation on elements backed by Dask DataFrames.

There were two problems before:
 1. The `keep_index` and `compute` args were not being passed along to subexpressions (fixed in 
c99c0cf)
 2. The `digitize` and `isin` numpy functions are not implemented as Pandas or Dask Series methods, so Dask Series were getting converted into numpy arrays.  I added custom wrappers for these functions in e38f68f.

Closes https://github.com/pyviz/holoviews/issues/4034

TODO:
 - [x] Tests: We should add tests across all dim transformations to make sure they preserve dask Series as Series when appropriate.

### DIFF
--- a/holoviews/util/transform.py
+++ b/holoviews/util/transform.py
@@ -130,7 +130,7 @@ def categorize(values, categories, default=None):
         cats.append(cat)
     result = np.asarray(cats)
     # Convert unicode to object type like pandas does
-    if result.dtype.kind == 'U':
+    if result.dtype.kind in ['U', 'S']:
         result = result.astype('object')
     return result
 
--- a/holoviews/util/transform.py
+++ b/holoviews/util/transform.py
@@ -34,7 +34,7 @@ def norm(values, min=None, max=None):
 def lognorm(values, min=None, max=None):
     """Unity-based normalization on log scale.
        Apply the same transformation as matplotlib.colors.LogNorm
-    
+
     Args:
         values: Array of values to be normalized
         min (float, optional): Lower bound of normalization range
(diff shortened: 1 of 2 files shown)


holoviews/util/transform.py
```diff
@@ -375,7 +412,15 @@ def apply(
...
             for arg in args:
                 if isinstance(arg, dim):
-                    arg = arg.apply(dataset, flat, expanded, ranges, all_values)
+                    arg = arg.apply(
+                        dataset,
+                        flat,
+                        expanded,
+                        ranges,
+                        all_values,
+                        keep_index,
```
philippjfr: Ah, think I missed this, which is why my attempts failed.

holoviews/util/transform.py
```diff
@@ -106,6 +106,36 @@ def categorize(values, categories, default=None):
     return np.asarray(cats)
 
 
+def _maybe_map(numpy_fn):
```
philippjfr: Was wondering whether we should do this for any numpy function that's not implemented but I think for now sticking with just isin and digitize is safest.

jonmmease:
Ok, all dim expressions are now tested against Dask DataFrames. I had to fix Dask support for a few of the operations along the way, but I think things are in pretty good shape now.

philippjfr:
Amazing, thanks so much @jonmmease!

philippjfr:
Some minor python 2 issues in the tests.

jonmmease:
Python 2 issues fixed

PR #4049, opened 2019-10-08
Correctly detect depends fn dependencies

philippjfr:


jbednar:
Works for me now; thanks so much!

PR #4061, opened 2019-10-16
Fixed title formatter in GenericElementPlot

poplarShift:
Before: 
Screen Shot 2019-10-16 at 2 38 29 PM

After:
Screen Shot 2019-10-16 at 2 46 15 PM

 This said, I would also like to point out that the `_format_title` methods of `GenericElementPlot` and `GenericCompositePlot` look pretty similar, and both classes directly derive from `DimensionedPlot`, yet when both `title` and `title_format` are set, one falls back to `title`, the other one to `title_format`. I don't know if there was a special reason why the two classes have separate implementations but I think the code could potentially be clearer.

```
from holoviews import opts
import holoviews as hv
hv.extension('bokeh')
# hv.extension('matplotlib')

l = hv.NdLayout({'FOO': hv.Scatter([], group='ONE'), 'BAR': hv.Scatter([], label='TWO')}, kdims='MYDIM').opts(
    opts.Scatter(title_format='this {label} is {group} my {dimensions} formatter')
)
l
```

### DIFF
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -488,7 +488,7 @@ class DimensionedPlot(Plot):
 
     def _format_title(self, key, dimensions=True, separator='\n'):
         if self.title_format and util.config.future_deprecations:
-            self.param.warning('title_format is deprecated. Use title')
+            self.param.warning('title_format is deprecated. Please use title instead')
 
         label, group, type_name, dim_title = self._format_title_components(
             key, dimensions=True, separator='\n'
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -486,6 +486,39 @@ class DimensionedPlot(Plot):
         return util.bytes_to_unicode(separator.join(g for g in groups if g))
 
 
+    def _format_title(self, key, dimensions=True, separator='\n'):
+        label, group, type_name, dim_title = self._format_title_components(
+            key, dimensions=True, separator='\n'
+        )
+
+        custom_title = (self.title != self.param.params('title').default)
+        if custom_title and self.title_format:
+            self.warning('Both title and title_format set. Using title parameter')
+        title_str = (
+            self.title if custom_title or self.title_format is None
+            else self.title_format
+        )
+
+        title = util.bytes_to_unicode(title_str).format(
+            label=util.bytes_to_unicode(label),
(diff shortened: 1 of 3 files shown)


DancingQuanta:
This is great! I was about to use `title_format` until I saw this. Is this case already covered in test suite?

poplarShift:
Hm, I didn't check actually. (I wanted to wait for feedback on whether we should simplify the `title`/`title_format` arguments while we're at it but I guess that can be done later too.)

philippjfr:
So the idea was that `title` would eventually replace `title_format` entirely and they were just aliases. Therefore `title` should at least for now take precedence. So if we could combine the two implementations in some form and have `title` take precedence in both cases that would be great.

poplarShift:
Does it make sense to move the `_format_title` method entirely into `DimensionedPlot`?

poplarShift:
Done @philippjfr. MWE to play around with it:

```
from holoviews import opts
import holoviews as hv
hv.extension('bokeh')

options = [
    opts.Scatter(title='this {label} is {group} my {dimensions} formatter', frame_width=400),
    opts.NdLayout(title='LAYOUT TITLE: this {label} is {group} my {dimensions} formatter'),
    opts.NdOverlay(title='OVERLAY TITLE: this {label} is {group} my {dimensions} formatter')
]
foo = hv.Scatter([], group='ONE', label='first')
bar = hv.Scatter([], group='TWO', label='second')
l = hv.NdLayout({'FOO': foo, 'BAR': bar}, kdims='MYDIM').opts(*options).cols(1).relabel(group='spam', label='eggs')
l
```

philippjfr:
Thanks! Looking **much** better to me. Any chance of adding some tests?

philippjfr:
Oh I'm also happy to start adding a deprecation warning for ``title_format``, maybe leaving it behind a `hv.config.future_deprecations` flag for now.

poplarShift:
I'll look at the tests later. I guess we'd need some for a simple Element, a Layout, and an Overlay, and for all backends? (The code *should* not depend on backend so one could get away with only one I guess.)

philippjfr:
>(The code should not depend on backend so one could get away with only one I guess.)

That seems fine.

poplarShift:
I've added tests for the bokeh backend.

Issue #3819, opened 2019-07-12
PointDraw missing `styles`


ahuang11:
Seems a bit inconsistent that BoxEdit and PolyDraw both have the styles kwd, but not PointDraw

philippjfr:
That would be greatly appreciated! Let me know if you need any pointers. The basics are that you need to look at hv.streams.PolyDraw and copy the definition to hv.stream.PointDraw. Next you need to look at PolyDrawCallback in holoviews/plotting/bokeh/callbacks.py and implement the same styles handling for PointDraw.

srp3003:
I have raised pull request #4063, however I am pretty sure I am missing something. Are there any tests for PointDraw that I can run first?

PR #4063, opened 2019-10-16
Added styles for PointDraw (#3819)

srp3003:


holoviews/streams.py
```diff
@@ -1069,12 +1069,17 @@ class PointDraw(CDSStream):
     num_objects: int
         The number of polygons that can be drawn before overwriting
         the oldest polygon.
+
+     styles: dict
+    A dictionary specifying lists of styles to cycle over whenever
+    a new Point glyph is drawn.
```
philippjfr: Indentation here seems off.

philippjfr:
Thanks a lot! This looks good, testing this properly would require setting up selenium tests which we currently don't have but definitely *should* have. For now I'm happy to merge when the indentation issues are fixed.

Issue #4071, opened 2019-10-25
Suggestions for improving the Spikes element


jlstevens:
Here is a list of improvements that we would like to make to the `Spikes` element:

1. The `spike_length` plot option (when explicitly supplied) should override the second vdim (if present) in terms of controlling the spike heights.  This is necessary if you want extra vdims (e.g for hover information) but don't need the spike heights to represent anything.
2. The `position` plot option would be better named `offset`. We could alias these two and eventually deprecate the former name.
3. The `position/offset` plot options should support the `dim` object.

I'll be filing a simple PR to address the first point shortly.

PR #4072, opened 2019-10-25
Allow spike_length to override second Spikes vdim behavior

jlstevens:
Simple PR that addresses the first suggestion mentioned in #4071. 

Specifically, the spike_length plot option (when explicitly supplied) should override the second vdim (if present) in terms of controlling the spike heights. This is necessary if you want extra vdims (e.g for hover information) but don't need the spike heights to represent anything.

The other two suggestions are also nice but I think they ought to be addressed in separate PRs.

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -79,6 +79,22 @@ def cleanup_custom_options(id, weakref=None):
                         'an unreferenced orphan tree may persist in '
                         'memory' % (e, id))
 
+def lookup_options(obj, group, backend):
+    plot_class = None
+    try:
+        plot_class = Store.renderers[backend].plotting_class(obj)
+        style_opts = plot_class.style_opts
+    except SkipRendering:
+        style_opts = None
+
+    node = Store.lookup_options(backend, obj, group)
+    if group == 'style' and style_opts is not None:
+        return node.filtered(style_opts)
+    elif group == 'plot' and plot_class:
+        return node.filtered(list(plot_class.params().keys()))
+    else:
+        return node
+
 
 class SkipRendering(Exception):
     """
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -80,6 +80,10 @@ def cleanup_custom_options(id, weakref=None):
                         'memory' % (e, id))
 
 def lookup_options(obj, group, backend):
+    """
+    Given a HoloViews object, a plot option group (e.g 'style') and
+    backend, return the corresponding Options object.
+    """
     plot_class = None
     try:
         plot_class = Store.renderers[backend].plotting_class(obj)
--- a/holoviews/plotting/mpl/chart.py
+++ b/holoviews/plotting/mpl/chart.py
@@ -1139,14 +1139,31 @@ class SpikesPlot(PathPlot, ColorbarPlot):
         ax.add_collection(line_segments)
         return {'artist': line_segments}
 
+    def _get_axis_dims(self, element):
+        if 'spike_length' in self.lookup_options(element, 'plot').options:
+            return  [element.dimensions()[0], None, None]
+        return super(SpikesPlot, self)._get_axis_dims(element)
+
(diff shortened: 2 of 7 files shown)


philippjfr:
The changes all look fine, here are the other changes needed before this can be merged:

- [ ] Apply the same changes to the matplotlib implementation
- [ ] Find some way of doing the same thing for spike datashader rasterization

The logic here looks like this:

```
        if element.vdims: # Perhaps check the plot option here?!
            x, y = element.dimensions()
            if not self.p.y_range:
                y0, y1 = element.range(1)
                if y0 >= 0:
                    default = (0, y1)
                elif y1 <= 0:
                    default = (y0, 0)
                else:
                    default = (y0, y1)
            else:
                default = None
        else:
            x, y = element.kdims[0], None
            default = (0, 1) # This should probably also respect the spike_length and position/offset
```

jlstevens:
@philippjfr Just noting that the matplotlib implementation doesn't seem to be called the `_get_axis_dims` method which means the vdim is still being shown on the y-axis when it should be disabled.

philippjfr:
>@philippjfr Just noting that the matplotlib implementation doesn't seem to be called the _get_axis_dims method which means the vdim is still being shown on the y-axis when it should be disabled.

Please open an issue.

philippjfr:
Actually it seems like currently matplotlib has a completely different mechanism here, SpikePlot.get_data returns this:

```
        return (clean_spikes,), style, {'dimensions': dims}
```

For now change the return here to be what you expect and remove the _get_axis_dims but still file the issue so we can generalize this at some point and use the same mechanism across backends.

jlstevens:
Filed the issue as requested (#4073) and updated `SpikePlot.get_data` as you proposed.

holoviews/plotting/mpl/chart.py
```diff
@@ -1209,9 +1222,16 @@ def get_data(self, element, ranges, style):
             style['array'] = element.dimension_values(cdim)
             self._norm_kwargs(element, ranges, style, cdim)
 
+        if 'spike_length' in self.lookup_options(element, 'plot').options:
```
philippjfr:
```suggestion
        if 'spike_length' in opts:
```

jlstevens:
@philippjfr Travis is now passing. I know there are other things to fix with spikes (e.g renaming `position` to `offset`, documenting this change etc) but I think this is the end of the code changes I intend to make in this PR (the aliasing of the parameter can be in a different PR). As such, I think it is ready for a code review.

holoviews/core/options.py
```diff
@@ -79,6 +79,22 @@ def cleanup_custom_options(id, weakref=None):
                         'an unreferenced orphan tree may persist in '
                         'memory' % (e, id))
 
+def lookup_options(obj, group, backend):
```
philippjfr: A docstring here would be nice.

philippjfr:
Maybe add a test to `holoviews/tests/plotting/bokeh/testspikesplot.py` that tests overriding the vdim with a spike_length by adapting this existing test:

```python
    def test_spikes_datetime_kdim_hover(self):
        points = Spikes([(dt.datetime(2017, 1, 1), 1)], 'x', 'y').options(tools=['hover'])
        plot = bokeh_renderer.get_plot(points)
        cds = plot.handles['cds']
        self.assertEqual(cds.data['x'].astype('datetime64'), np.array([1483228800000000000]))
        hover = plot.handles['hover']
        self.assertEqual(hover.tooltips, [('x', '@{x_dt_strings}'), ('y', '@{y}')])
```

jlstevens:
@philippjfr Added the suggested unit test and docstring. Ready to merge?

holoviews/core/options.py
philippjfr: Ideally the docstring would conform to our docstring spec.

philippjfr:
Looks good, thanks for fixing it!

Issue #3932, opened 2019-08-27
Implement .transform method to apply dim transforms across containers


poplarShift:
I think it could be very much worthwhile to implement data aggregation using `dim` transforms from the ground up (i.e. in each interface). This would give us any sort of multidimensional transforms on HexTiles such as proposed in #3636 etc. for free.

The API would not be changed, but the following would become possible:
```
from holoviews import dim, opts
import numpy as np
import holoviews as hv
hv.extension('bokeh')

xx = np.arange(20).reshape(5, 4)
xx2 = np.c_[xx[:, :2], np.arange(20, 50, 3).reshape(5, 2)]
x = np.r_[xx, xx2]

kdims = ['x', 'y']
vdims = ['z', 'a']
e = hv.Points(np.r_[x, x], kdims, vdims).opts(color='z', colorbar=True)
ds = hv.Dataset(e)

def myfunc(z, a):
    return np.sum(z)+np.sum(a)

dim_transform = dim('z', myfunc, dim('a'))
print(e)
print(e.aggregate(function=dim_transform).data)
```
yielding (note how the vdims `z`, `a` are condensed into one vdim `z` by the summation
```
:Points   [x,y]   (z)
[[  0   1  96]
 [  4   5 136]
 [  8   9 176]
 [ 12  13 216]
 [ 16  17 256]]
```

I'm happy to work on this but wanted to hear what your thoughts are first.

philippjfr:
This is an interesting proposal, my main concern is that aggregate generally retains all vdims and it's not clear how it would know to drop 'a' but keep 'z' in your example. Originally I had envisioned adding another method for this and was hoping to call it ``apply`` but that is now reserved for something else.

philippjfr:
The spelling for apply would have been:

```
points.apply(z=dim_transform)
```

i.e. you'd tell it to create a new dimension or overwrite an existing one using the keyword argument.

poplarShift:
I like that API and I think we could build on that.

How about we call it `.transform`? Then `.aggregate` with a dim transform would be roughly equivalent to `.groupby(kdims).transform(...)`.

# Proposal

## API

`.transform(v=dim_transform)` just returns the applied dim transform.
`.groupby(...).transform(v=dim_transform)` would return the dim transform applied to each part of the resulting container.

## Further issues

[...]

PR #4080, opened 2019-10-31
Multi-dimensional dim transforms on data sets

poplarShift:
Addresses #3932 and https://github.com/holoviz/holoviews/issues/237

Supersedes #3636.

Related to #3790.

In this PR, we have the possibility to apply arbitrary dim transforms with multiple output values to Datasets, taking into account correct insertion of dimensions. The name of the new method used here is `.transform`, which provides two ways of specifying transforms. You either supply tuples of dimensions and dim_transforms as args or as kwargs and the method will apply these transforms and either replace the existing dimensions or add the newly added dimensions as new value dimensions. 

The upside of all of this is that we get complex statistical aggregation for free - see below for an example with hex bins that compute trends within each bin.

List of changes
- [x] Add Dataset.transform
- [x] Implement Interface.assign
- [x] Dataset.aggregate now handles dim transforms
- [x] tests

### Setup

```python
import xarray as xr
from holoviews import Dataset
import numpy as np
import pandas as pd
import holoviews as hv
from holoviews import dim, opts
```

### Multi-dimensional dim transforms and aggregations

```python
df = pd.DataFrame(dict(
    x=np.array(range(7))%2,
    y=np.array(range(7))%3,
    u=np.array(range(7))%4,
    v=np.array(range(7))%5
))
ds = Dataset(df, ['x', 'y'], ['u', 'v'])
nds = ds.groupby(['x'])

# scalar output
tf1 = dim('u', lambda u, v: np.sum(u) + np.sum(v), dim('v'))
# tuple of arrays
tf2 = dim('u', lambda u, v: (u, np.mean(v)), dim('v'))
```

```python
print(ds.data.head())
```

x|y|u|v
---|---|---|---
0|0|0|0
1|1|1|1
0|2|2|2
1|0|3|3
0|1|0|4

```python
print(ds.transform(w=tf1).data)

# same as:
# ds.transform(('w', tf1)).data
```

x|y|u|v|w
---|---|---|---|---
0|0|0|0|20
1|1|1|1|20
0|2|2|2|20
1|0|3|3|20
0|1|0|4|20
1|2|1|0|20
0|0|2|1|20

```python
print(ds.transform((('a', 'b'), tf2), drop=True).data)
```

a|b
---|---
0|1.5714285714285714
1|1.5714285714285714
2|1.5714285714285714
3|1.5714285714285714

```python
print(ds.aggregate('y', w=tf1).data)
```

y|w
---|---
0|9
1|6
2|5

### Example: Complex hex binning operations

```python
hv.extension('bokeh')
xds = xr.tutorial.open_dataset('air_temperature').sel(time=slice(None, '2013-1-5'))
df = xds.to_dataframe().reset_index()

def regression(x1, x2):
    x1 = pd.to_numeric(x1)/1e9/86400.
    p = np.polyfit(x1, x2, 1)
    return p[0], p[1]

tf = dim('time', regression, dim('air'))
ds = hv.Dataset(df, ['lon', 'lat'], ['time', 'air'])

e = hv.HexTiles(ds)
e.opts(gridsize=10, aggregator=(('trend', 'offset'), tf),
       color=dim('trend'),
       scale=dim('offset').norm(),
       colorbar=True, width=600)
```
image

### DIFF
--- a/holoviews/core/accessors.py
+++ b/holoviews/core/accessors.py
@@ -258,9 +258,26 @@ class Apply(object):
         See :py:meth:`Dataset.transform` and :py:meth:`Apply.__call__`
         for more information.
         """
+        from ..streams import Params
+        streams = []
+        params = {}
+        for _, arg in list(args)+list(kwargs.items()):
+            for op in arg.ops:
+                op_args = list(op['args'])+list(op['kwargs'].items())
+                for op_arg in op_args:
+                    if 'panel' in sys.modules:
+                        from panel.widgets.base import Widget
+                        if isinstance(op_arg, Widget):
+                            op_arg = op_arg.param.value
+                    if (isinstance(op_arg, param.Parameter) and
+                        isinstance(op_arg.owner, param.Parameterized)):
+                        params[op_arg.name+str(id(op))] = op_arg
+        streams += Params.from_params(params, watch_only=True)
         kwargs['_method_args'] = args
         kwargs['per_element'] = True
+        kwargs['streams'] = streams
         return self.__call__('transform', **kwargs)
+
 
 @add_metaclass(AccessorPipelineMeta)
 class Redim(object):
--- a/holoviews/core/accessors.py
+++ b/holoviews/core/accessors.py
@@ -221,6 +221,14 @@ class Apply(object):
         See :py:meth:`Dimensioned.opts` and :py:meth:`Apply.__call__`
         for more information.
         """
+        from ..util.transform import dim
+        from ..streams import Params
+        params = {}
+        for arg in kwargs.values():
+            if isinstance(arg, dim):
+                params.update(arg.params)
+        streams = Params.from_params(params, watch_only=True)
+        kwargs['streams'] = kwargs.get('streams', []) + streams
         kwargs['_method_args'] = args
         return self.__call__('opts', **kwargs)
 
@@ -258,24 +266,16 @@ class Apply(object):
         See :py:meth:`Dataset.transform` and :py:meth:`Apply.__call__`
         for more information.
         """
+        from ..util.transform import dim
         from ..streams import Params
-        streams = []
         params = {}
         for _, arg in list(args)+list(kwargs.items()):
-            for op in arg.ops:
-                op_args = list(op['args'])+list(op['kwargs'].items())
-                for op_arg in op_args:
-                    if 'panel' in sys.modules:
-                        from panel.widgets.base import Widget
-                        if isinstance(op_arg, Widget):
-                            op_arg = op_arg.param.value
-                    if (isinstance(op_arg, param.Parameter) and
-                        isinstance(op_arg.owner, param.Parameterized)):
-                        params[op_arg.name+str(id(op))] = op_arg
-        streams += Params.from_params(params, watch_only=True)
+            if isinstance(arg, dim):
+                params.update(arg.params)
+        streams = Params.from_params(params, watch_only=True)
+        kwargs['streams'] = kwargs.get('streams', []) + streams
         kwargs['_method_args'] = args
         kwargs['per_element'] = True
-        kwargs['streams'] = streams
         return self.__call__('transform', **kwargs)
 
 
--- a/holoviews/core/accessors.py
+++ b/holoviews/core/accessors.py
@@ -205,16 +205,25 @@ class Apply(object):
                     mapped.append((k, new_val))
             return self._obj.clone(mapped, link=link_inputs)
 
-
     def aggregate(self, dimensions=None, function=None, spreadfn=None, **kwargs):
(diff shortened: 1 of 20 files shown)


philippjfr:
Sorry I never reviewed this. I'd very much like to get this into the release, so I'm going to take it over.

holoviews/core/dimension.py
```diff
@@ -703,6 +710,34 @@ def map(self, map_fn, specs=None, clone=True):
         else:
             return map_fn(self) if applies else self
 
+    def transform(self, *args, drop=False, drop_duplicate_data=True, **kwargs):
```
philippjfr: -1 on adding this, we can now do this using `.apply`. I'll therefore suggest we add `.apply.transform`.

holoviews/core/data/__init__.py
```diff
@@ -890,6 +952,70 @@ def load_subset(*args):
...
+        dim_transform=None,
+        drop=False,
+        drop_duplicate_data=True,
+        **kwargs
+    ):
+        """
+        Transforms the Dataset according to a dimension transform.
+
+        Args:
+            signature: Specify output arguments as a list of strings
```
philippjfr: I'm having a tough time understanding the signature argument. I'm guessing it allows a single transform which returns multiple dicts but I'd like to come up with a better API here, e.g. instead of:

`obj.transform(('a', 'b'), transform)`

I'd probably allow:

`obj.transform((('a', 'b'), transform))`

and allow supplying multiple such transforms using *args.

holoviews/core/dimension.py
jbednar: Ok by me.

holoviews/core/data/__init__.py
jbednar: Seems ok to me.

holoviews/core/dimension.py
poplarShift: Sounds good!

holoviews/core/data/__init__.py
poplarShift: Your suggestion looks good!

poplarShift:
Sorry for not replying earlier. I currently don't have a lot of spare time but I'm still interested in getting this in.

philippjfr:
@poplarShift I have a potentially **much** simpler implementation here, but one thing I haven't figured out is the drop_duplicate_data keyword argument? Why is that needed here?

poplarShift:
I see you already went ahead and got rid of it.  I like the solution with `.assign`!

philippjfr:
>I see you already went ahead and got rid of it. I like the solution with .assign!

I did, but is there a good reason why I shouldn't have?

holoviews/core/accessors.py
```diff
@@ -253,6 +252,32 @@ def select(self, **kwargs):
         """
         return self.__call__('select', **kwargs)
 
+    def transform(self, *args, **kwargs):
+        """Applies transforms to all Datasets.
```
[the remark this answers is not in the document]
philippjfr: This is on the Apply accessor and therefore broadcasts the apply to all leaf nodes.

holoviews/core/data/dictionary.py
```diff
@@ -270,6 +270,13 @@ def values(cls, dataset, dim, expanded=True, flat=True, compute=True, keep_index
         return values
 
 
+    @classmethod
+    def assign(cls, dataset, new_data):
```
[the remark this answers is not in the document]
philippjfr: Interface methods do not currently have docstrings. I'll consistently add those at some point but since they aren't really public we've not prioritized that yet.

holoviews/core/accessors.py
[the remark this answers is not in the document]
jbednar: A docstring should fundamentally document what this method actually does, not how it is used by other things.  Such other usage and context can be mentioned _later_ in a docstring, but it's not the definition of what this method does, as the first line of the docstring is meant to be.  It sounds to me like the Apply accessor applies this transform to all Datasets, not that this method itself does? But maybe I'm misunderstanding what you mean here.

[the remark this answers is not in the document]
philippjfr: >A docstring should fundamentally document what this method actually does, not how it is used by other things.

This is a user facing method which allows them to apply a transform on all Datasets on the object they using the method on, so I'm having trouble understanding the distinction you're trying to make here.

poplarShift:
@philippjfr Awesome!

Sorry for being a bit slow with replying these days. I don't remember the exact reasons for the drop_duplicate_data kwarg, but I think they were specific to my implementation.

Also thanks for seeing this through, I'm super excited about using this straight from the source  instead of my monkey patched code snippets! As @jbednar said this is indeed a major step up for workflow design.

Issue #4083, opened 2019-11-04
Allow defining Stream and Link callback transforms


philippjfr:
Currently we define a number of Stream and Link Callbacks which have to be subclassed in GeoViews to add support for projecting the data. This results in some pretty ugly code that has to subclass all the callbacks and add a single line to apply the projection transform. Therefore it would be nice if both types of callback simply allowed adding transforms to these callbacks, which could be registered when importing geoviews (or whatever other library wants to define a transform).

PR #4084, opened 2019-11-05
Add Annotators

philippjfr:
- [x] Add `Annotator` baseclass
- [x] Add `PointAnnotator`
- [x] Add `PathAnnotator`
- [x] Add `PolyAnnotator`
- [ ] Add `BoxAnnotator`
- [x] Add `annotate` function
- [x] Document `annotate` function
- [x] Add tests for `Annotator` and `annotate`
- [x] Allow declaring transforms for Stream callbacks (#4083)
- [x] Add SelectionLink and VertexTableLink

### DIFF
--- a/holoviews/annotators.py
+++ b/holoviews/annotators.py
@@ -116,7 +116,7 @@ class annotate(param.ParameterizedFunction):
         return (Overlay(layers).collate() + tables).opts(sizing_mode='stretch_width')
 
     def __call__(self, element, **params):
-        for eltype, annotator_type in self._annotators_types.items():
+        for eltype, annotator_type in self._annotator_types.items():
             if isinstance(element, eltype):
                 break
             else:
@@ -206,7 +206,7 @@ class Annotator(PaneBase):
         self.param.watch(self._update, self._triggers)
         self.layout[:] = [self.plot, self.editor]
 
-    @param.depends('annotations', 'object', 'opts')
+    @param.depends('annotations', 'object', 'default_opts')
     def _get_plot(self):
         return self._process_element(self.object)
 
@@ -448,7 +448,7 @@ class PointAnnotator(Annotator):
 
 
 # Register Annotators
-annotate.update([
+annotate._annotator_types.update([
     (Polygons, PolyAnnotator),
     (Path, PathAnnotator),
     (Points, PointAnnotator)
--- a/holoviews/annotators.py
+++ b/holoviews/annotators.py
@@ -1,15 +1,18 @@
 from __future__ import absolute_import
 
+import sys
+
 import param
 
 from panel.param import ParamMethod
+from panel.pane import PaneBase
 from panel.layout import Row, Tabs
 from panel.util import param_name
 
 from .core import Element, Overlay
 from .element import Path, Polygons, Points, Table
 from .plotting.links import VertexTableLink, DataLink, SelectionLink
-from .streams import PolyDraw, PolyEdit, Selection1D, PointDraw
+from .streams import BoxEdit, PolyDraw, PolyEdit, Selection1D, PointDraw
 
 
 def preprocess(function, current=[]):
@@ -107,7 +110,7 @@ class AnnotationManager(param.Parameterized):
 
 
 
-class Annotator(param.Parameterized):
+class Annotator(PaneBase):
     """
     An Annotator allows drawing, editing and annotating a specific
     type of element. Each Annotator consists of the `plot` to draw and
@@ -119,7 +122,7 @@ class Annotator(param.Parameterized):
     annotations = param.ClassSelector(default=[], class_=(dict, list), doc="""
         Annotations to associate with each object.""")
 
-    element = param.ClassSelector(class_=Element, doc="""
+    object = param.ClassSelector(class_=Element, doc="""
         The Element to edit and annotate.""")
 
     num_objects = param.Integer(default=None, bounds=(0, None), doc="""
@@ -144,57 +147,68 @@ class Annotator(param.Parameterized):
     # Allows patching on custom behavior
     _extra_opts = {}
 
+    priority = 0.7
+
+    @classmethod
+    def applies(cls, obj):
+        if 'holoviews' not in sys.modules:
+            return False
+        return isinstance(obj, cls.param.object.class_)
+
+    def select(self, selector=None):
+        return self.layout.select(selector)
+
     @property
     def _element_type(self):
-        return self.param.element.class_
+        return self.param.object.class_
 
     @property
-    def _element_name(self):
+    def _object_name(self):
         return self._element_type.__name__
 
-    @param.depends('element')
-    def _get_plot(self):
-        return self.element
-
-    def __init__(self, element=None, **params):
-        super(Annotator, self).__init__(**params)
+    def __init__(self, object=None, **params):
+        super(Annotator, self).__init__(object, **params)
         self._tables = []
         self.editor = Tabs()
         self._selection = Selection1D()
-        self._initialize(element)
+        self._initialize(object)
         self.plot = ParamMethod(self._get_plot)
-        self._layout = Row(self.plot, self.editor, sizing_mode='stretch_width')
+        self.layout[:] = [self.plot, self.editor]
 
-    @param.depends('annotations', 'element', 'num_objects', 'opts', 'table_opts', watch=True)
+    def _get_model(self, doc, root=None, parent=None, comm=None):
+        return self.layout._get_model(doc, root, parent, comm)
+
+    @param.depends('annotations', 'object', 'num_objects', 'opts', 'table_opts', watch=True)
     @preprocess
-    def _initialize(self, element=None):
+    def _initialize(self, object=None):
         """
-        Initializes the element ready for annotation.
+        Initializes the object ready for annotation.
         """
-        element = self.element if element is None else element
-        self._init_element(element)
+        object = self.object if object is None else object
+        self._init_element(object)
         self._init_table()
-        self._selection.source = self.element
+        self._selection.source = self.object
         self.editor[:] = self._tables
         self._stream.add_subscriber(self._update_element)
 
+    @param.depends('object')
+    def _get_plot(self):
+        return self.object
+
     def _update_element(self, data=None):
         with param.discard_events(self):
-            self.element = self._stream.element
+            self.object = self._stream.element
 
     def _table_data(self):
         """
         Returns data used to initialize the table.
         """
-        element = self.element
+        object = self.object
         for transform in self.table_transforms:
-            element = transform(element)
-        return element
+            object = transform(object)
+        return object
 
-    def _repr_mimebundle_(self, include=None, exclude=None):
-        return self._layout._repr_mimebundle_(include, exclude)
-
-    def _init_element(self, element):
+    def _init_element(self, object):
         """
         Subclasses should implement this method.
         """
@@ -207,7 +221,7 @@ class Annotator(param.Parameterized):
     @property
     def selected(self):
         """
-        Subclasses should return a new element containing currently selected objects.
+        Subclasses should return a new object containing currently selected objects.
         """
 
 
@@ -220,8 +234,8 @@ class PathAnnotator(Annotator):
     edit_vertices = param.Boolean(default=True, doc="""
         Whether to add tool to edit vertices.""")
 
-    element = param.ClassSelector(class_=Path, doc="""
-        Path element to edit and annotate.""")
+    object = param.ClassSelector(class_=Path, doc="""
+        Path object to edit and annotate.""")
 
     show_vertices = param.Boolean(default=True, doc="""
         Whether to show vertices when drawing the Path.""")
@@ -270,18 +284,18 @@ class PathAnnotator(Annotator):
         tools = [tool() for tool in self._tools]
         opts = dict(tools=tools, color_index=None, **self.opts)
         opts.update(self._extra_opts)
-        self.element = element.options(**opts)
+        self.object = element.options(**opts)
 
     def _init_table(self):
         name = param_name(self.name)
         self._stream = PolyDraw(
-            source=self.element, data={}, num_objects=self.num_objects,
+            source=self.object, data={}, num_objects=self.num_objects,
             show_vertices=self.show_vertices, tooltip='%s Tool' % name,
             vertex_style=self.vertex_style
         )
         if self.edit_vertices:
             self._vertex_stream = PolyEdit(
-                source=self.element, tooltip='%s Edit Tool' % name,
+                source=self.object, tooltip='%s Edit Tool' % name,
                 vertex_style=self.vertex_style,
             )
 
@@ -290,11 +304,11 @@ class PathAnnotator(Annotator):
         table_data = {a: [d.dimension_values(a, expanded=False)[0] for d in table_data]
                       for a in annotations}
         self._table = Table(table_data, annotations, []).opts(**self.table_opts)
-        self._link = DataLink(self.element, self._table)
+        self._link = DataLink(self.object, self._table)
         self._vertex_table = Table(
-            [], self.element.kdims, list(self.vertex_annotations)
+            [], self.object.kdims, list(self.vertex_annotations)
         ).opts(**self.table_opts)
-        self._vertex_link = self._vertex_table_link(self.element, self._vertex_table)
+        self._vertex_link = self._vertex_table_link(self.object, self._vertex_table)
         self._tables = [
             ('%s' % name, self._table),
             ('%s Vertices' % name, self._vertex_table)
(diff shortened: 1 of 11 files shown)


jbednar:
- [ ] Add RangeAnnotator (for x and y axis ranges (assuming colorbar axis ranges are too much to ask)?

jlstevens:
After discussing this PR with @philippjfr, here are my thoughts.

Firstly, even a very basic prototype that shows that the API here could apply to the plotly backend would go a long way to justifying annotators at the HoloViews level. @philippjfr said he believes that a basic POC wouldn't be very hard to come up with. Then we could try to see how many of the annotators could be made to work properly with plotly.

Here are some other more concrete suggestions I made regarding this PR:

1) I dislike the `opts` argument when constructing an annotator. It seems to me that the element can be styled either before it is passed to the annotator or afterwards via the `.element` attribute.

2) Right now, annotators are using a panel `Row` in the background. I think the *displayed representation* of an annotator should be the same as ``annotator.element + annotator.editor``. Even better would be if an annotator is just a utility that simply *returns* ```element + editor```. In other words, ideally we wouldn't need a handle on an annotator 'object' at all. I think this can be made to work by implementing point 5 below so I'll come back to this point.

3) I'm not a fan of `annotator.editor` as the whole point of annotators is to edit elements visually. What `editor` is the editable *textual* representation. Perhaps it could be ``annotator.entries`` or some other name?

4) Here I show a progression in my thinking, starting from the current `AnnotationManager` to how I would like it to work:

   i) My initial reaction is that  `AnnotationManager` is really `AnnotatorOverlay` as you can add annotators as well as regular elements. That naturally suggests that ``__mul__`` should be supported where ``__mul__`` involving any annotators becomes an `AnnotatorOverlay` instead of a regular `Overlay`. Just like with `Overlay`, users wouldn't *have* to explicitly use that class unless they wanted to.

   ii) This in turn concerns me as then we'll have three types of `Overlay`: `Overlay`, `NdOverlay` and `AnnotatorOverlay`. `NdOverlay` has keys and a dimension, so the constructor is different but I think the ideal situation would be if `Overlay` just supported annotators.

   iii) So what is the thing going into the `Overlay`? This becomes the sticking point if you want to avoid new 'annotator' objects in the hierarchy and if you want a single annotator to just be ``element + editor`` (which I do). I don't think there is a good way to do this so I decided I do *not* want to support ``__mul__`` on any new object type.

   iv) Instead I think `AnnotatorOverlay` should *not* interact with ``__mul__`` in any way. Instead, like other annotators, it should return ``overlay + editor`` where the first part of the layout overlays the components correctly and the `editor` composes the tables correctly. The signature would be ``AnnotatorOverlay(obj1, obj2, ...)``  (or equivalent) where the objects are the output of individual annotator functions (i.e `element` + `editor`) or single elements in which case they are just included in the final overlay on the left.

In general, I would like to follow the approach used in the new linked selections feature where no new elements/containers are introduced: you simply pass an existing layout through a function to introduce the desired linking behavior. This brings me to my last point that there is a an even bigger generalization that I think we should be making here. 

What I am realizing is that linked selections and annotators are similar concepts. These are utilities (specifically `ParameterizedFunction`s) that we are offering users that result in composite HoloViews objects with special behaviors. In the case of linked selections, you are taking a layout containing elements derived from a `Dataset` and applying selection linking. Here, with annotators, you are taking an element and giving it a table that is linked to the edits you make as you annotate.

Semantically, the concept here is that they are both `Coordinators` (horrible name though!): i.e utilities that either take HoloViews elements and coordinates them, or creating new, coordinated elements (tables in the case of annotators). I think there are shared concepts (and hopefully machinery, e.g the `pipeline`) that would help make it clear that both annotators and linked selections are two instances of an even more general approach for adding special user interactions easily in HoloViews.

jbednar:
Thanks for these interesting ideas!

philippjfr:
I'll go through your comments 1 by 1, as we already discussed them I mostly agree but there are some subtleties that were lost in discussion

1.The reason for the opts argument is that the element is recreated in certain scenarios, e.g. when setting a new .element  on the annotator, in those cases it can be annoying to reapply the options. We could consider copying the options when that happens, I'll have to prototype this to see how well it works.

2. I'm against this, the annotator needs a handle to be easy to include in a panel pipeline, have the ability to update it with a new element and access the current selections. I have no objection to easily make a regular layout available in some form.


3. I prefer editor to entries but wouldn't be opposed to a different name.

4. I agree with some of your thinking here but disagree with your overall conclusion that this thing should be a function not an object. As I mentioned before a lot of the functionality required of these objects only makes sense in the context of having objects. Just to make sure I follow your suggestion, the annotators would be functions which return a layout of the element and the table, while the manager would simply compose these objects together? I think maybe I can explore this behaviour while also supporting a mode where you make an annotator instance which retains some of the current functionality (the linked selections function also support something similar).

I'll reserve final judgement until I've prototyped some of this and am happy with the usability.

philippjfr:
Part of the problem here is that the annotations and annotation manager need to be more dynamic than a simple HoloViews object can be which is why some of these requirements are fairly hard to satisfy without losing some of the functionality, e.g. if the manager just composes the objects on the input into a new HoloViews layout the resulting object wont link to the original annotators anymore. It may be possible to make all the linking work that way but again I'll have to try it first and make some significant changes to the way streams and links are set up to be sure.

jlstevens:
In short, I think that if annotator objects are to be justified as independent things, I need to know how they fit in this diagram:

image

Right now, I think they offer useful functionality but don't really fit in with the rest of HoloViews. We naturally feel they should support `*` to compose annotators but they are kinda like specialized layouts.

I don't know what the solution is right now. As you say some prototyping will help us figure out if this concept can fit more naturally with the rest of HoloViews.

philippjfr:
So after playing around and thinking about it more I've now come to the conclusion that many of the suggestions made above aren't really workable. The reason `Annotators` need to be objects not functions (even if parameterized) is that:

1. They need to be responsive to changes in the parameters so they can be reused, e.g. in a pipeline. You need to be able to swap out the element you are annotated for instance.   
2. You need to be able to listen to parameters on the object. More specifically when you update the drawing or add attributes in the table(s) you need to be able to watch for these changes and access the annotated data.

That brings me to the second point, Annotators are not really following any existing HoloViews paradigm, they are responsive to changes in their parameters, they wrap a HoloViews object and build additional functionality around it, they are part plot, part table, which means they are effectively a layout. So in the end they are much more like a Panel object than they are a simple HoloViews object and adding some of the affordances like `+` or `*` to them is bound to result in lots of confusion as lots of edge cases aren't handled or other assumption you can make about HoloViews objects won't hold.

So overall I think they should be Panel objects, something I already discussed with @jlstevens. The sticking point then is still where they live, I really don't think they should live in Panel. Panel is a general library with good support for HoloViews but that doesn't mean Panel should gain significant plotting specific functionality and documentation about Annotators is absolutely out of place in the Panel user guide. At the same time @jlstevens seems to be objecting to adding Panel based objects to HoloViews itself, particularly if the functionality doesn't work across backends and at the moment I don't see how to support any useable annotator with any backend except bokeh as neither draw tools nor editable tables are easily available from plotly or matplotlib. I also do not have the bandwidth to work on these, so I don't really know where that leaves me.

jbednar:
Some thoughts on the above:

- I think Annotators are part of a very general, recurring theme in both HoloViews and Panel, which includes pn.interact(), hv.HoloMap, hv.DynamicMap, and hv.link_selections, all of which involve things that are laid out adjacent to each other (and thus independent in some ways) but are linked to each other (and thus need coordinating between the individually laid out elements, whether it's a bank of widgets that control either the interact() output, the DynamicMap contents, or a HoloMap, or whether it's a set of interlinked selections across elements, or a table that is tied to another separately plotted Element) -- in each case there are separately laid out items that are automatically linked and coordinated, either by Panel or by HoloViews. This is obviously an important type of capability, and one that we should work hard to ensure is supported well and fully generally so that we can handle other similar cases as they emerge.

- I think we all agree that annotators are not a concept specific to Bokeh, and so I don't think it's in any way crucial to provide a Plotly or other backend initially. I would think that adding an editable table type to Plotly could enable quite a lot of annotating functionality for Plotly+HoloViews without a major investiment, but I don't think any other code needs to wait on that -- plotly-backend support for annotators can be added if and when such editing capability exists in Plotly or another backend. 

- I strongly feel that annotators do not belong in Panel, regardless of whether they use Panel objects. Conceptually, annotators allow working with and constructing HoloViews Elements, and Panel is fully independent of HoloViews and should never be made to tie closely to HoloViews. Panel needs to stay agnostic about plotting libraries and their particular affordances.

- I really want annotators to be available properly soon, because they are the only practical way I've seen to work with Bokeh drawing tools, and I'll be needing to train people on how to use the drawing tools for labeling data in early January 2020. I don't want to create any more new libraries, and so if annotators don't fit in Panel, they need to go _somewhere_, and given that they are ways of working with HoloViews Elements I believe they should go in HoloViews.

- I agree with Jean-Luc that it would be great if Annotators would fit into the HoloViews object hierarchy; if all other things are equal, it is very natural that they do so. But I disagree with Jean-Luc that they _must_ fit into the object hierarchy; HoloViews is already built on various Python libraries (including Panel, but also including Numpy and Pandas) and already has methods and functions that return items not in that object hierarchy, such as numbers, arrays, and lists; I don't think it's inconceivable to have some other functions provide results that are not in the hierarchy (such as Panel objects). So I consider it _highly desirable_ that annotators fit into the current HoloViews object hierarchy, but not at the expense of failing to have somewhere we can provide suitably capable annotators, and not at the expense of having to cram them into an unrelated library like Panel. 

- I am still not fully clear on exactly what is lost by having the annotator objects be HoloViews layouts. @philippjfr mentions setting a new .element on the annotator, potentially as part of a Pipeline, but I'm not sure when that's useful -- what exactly does having that ability allow, that wouldn't otherwise be allowed?  When does one need to do that? When will an already-existing annotator need to have the element replaced?

philippjfr:
>I am still not fully clear on exactly what is lost by having the annotator objects be HoloViews layouts. @philippjfr mentions setting a new .element on the annotator, potentially as part of a Pipeline, but I'm not sure when that's useful -- what exactly does having that ability allow, that wouldn't otherwise be allowed?

This is very much secondary, it's very useful for reusing an Annotator in a pipeline where you might go back and forth and reuse an existing annotator by feeding it an element from a previous stage. The much more important functionality is the ability to easily read out the the annotations and selections made in the annotator. To look up the current data on an annotator layout would look something like:

```python
layout.Table.I.streams[0].element.dframe()
# vs.
annotator.object.dframe()
```

and that is the simple case, to combine the vertex and polygon labels you'd need to combine data from two separate streams manually. If you additionally want to access the currently *selected* objects it gets even more complex. Having an object to make the data accessible in a straightforward way that lets you integrate it into an overall workflow is the reason Annotators exist and HoloViews objects just don't cut it.

philippjfr:
Okay I just wanted to expand the complex case, let's say I want to annotate a Polygons element with some labels per polygon and some values per vertex and then get access to a new element which includes those annotations for all currently selected elements, using an Annotator object that looks something like this:

```python
annotator = PolyAnnotator(polygon, annotations=['Label'], vertex_annotations=['Value'])
# Do some annotating and selecting
element_with_all_annotations_and_current_selection = annotator.object 
```

If however we just return a HoloViews layout we need to get handles on the underlying streams and then merge the data by ourselves:

```python
annotations = ['Label']
vertex_annotations = ['Value']
draw = PolyDraw(polygon, annotations=annotations, vertex_annotations=vertex_annotations)
selection = Selection1D()
layout = annotate(polygon, stream=draw, selection=selection,
                  annotations=annotations, vertex_annotations=vertex_annotations)

# Do some annotating

element = draw.element
if (element.interface.datatype == 'multitabular' and
    element.data and isinstance(element.data[0], dict)):
    for path in element.data:
        for col in annotations:
            if len(path[col]):
                path[col] = path[col][0]

data = [p for i, p in enumerate(element.split()) if i in selection.index]
element_with_all_annotations_and_current_selection = element.clone(data)
```

philippjfr:
Just to summarize our final discussion and what I will actually be implementing here. Achieving all the different requirements I laid out in my previous comments is quite a challenge, particularly in a purely HoloViews based framework. 

What we did decide is that from a HoloViews perspective annotation should be like other operations (and somewhat similar to linked selections). This means that ``annotate`` is a ParameterizedFunction which returns a HoloViews layout containing the editable plot and the table(s) which allow editing the element, this addresses point 1 and 2 from above. Additionally the ParameterizedFunction can be instantiated explicitly to keep a handle on the object, which allows accessing the annotated element and the current selection addressing point 3 from above. The HoloViews API therefore looks like this:

```python
annotator = annotate.instance()
layout = annotator(hv.Points([]), annotations=['Label'])

# Access the individual components
layout.Points.I + layout.Annotator.I

# Access the edited/annotated data
annotator.element

# Access the annotated/edited data and current selection
annotator.selected
```

For now I have not changed the underlying implementation of the Annotators which is still built on Param and Panel. I have however completely eliminated the AnnotationManager. The Annotator classes are an implementation detail from a HoloViews perspective and will not be documented explicitly (at least in HoloViews itself). These components can still be used directly if you need to have annotation capabilities in a Panel pipeline or other Param based workflow. The difference here is that it is fully dynamic, i.e. you can drop it in the pipeline and update the object with new elements fed in from a previous stage. This cannot be achieved using just HoloViews due to limitations on dynamic linking which are at least **currently** quite fundamental to the way HoloViews plots are updated.

jonmmease:
I'm not sure if this is helpful in terms of deciding where annotators go, but I have another example in mind of a Panel layout that is specifically for use with HoloViews objects.

I would like to eventually support more advanced multi-color data brushing workflows (inspired by glue: http://docs.glueviz.org/en/stable/getting_started/index.html#defining-subsets) using the `link_selections` approach.   To make these features easy to use, I would want to have a small Panel-based control panel for doing things like selecting the active brush color and specifying whether the next selection should apply an intersection, union, or subtraction operation.

I was imagining something like
```python
brusher = brush_selections.instance()
brushed_layout = brusher(layout)

# display layout
brushed_layout
```

```python
# display control panel
brusher.control_panel
```

philippjfr:
I suspect this kind of thing might come up more often in future so we should generally consider how to integrate Panel and HoloViews in a way that makes sense and doesn't lead to confusion.

jlstevens:
@jonmmease That API looks good (and fits in with the approach we have been discussing) but as Philipp says, we still need to find a way to make HoloViews work with panel in a clear and consistent way.

philippjfr:
I've removed BoxAnnotator for now, it really requires adding a Rects or Boxes element to HoloViews.

holoviews/core/data/interface.py
```diff
@@ -93,7 +93,8 @@ def _perform_getitem(cls, dataset, index):
         rows, cols = index
         if rows is Ellipsis:
             rows = slice(None)
-        data = dataset.interface.iloc(dataset.dataset, (rows, cols))
+
+        data = dataset.interface.iloc(dataset, (rows, cols))
```
philippjfr: Had to fix this to get my tests passing (should have been a new PR sorry).

holoviews/element/path.py
```diff
@@ -124,7 +124,9 @@ def split(self, start=None, end=None, datatype=None, **kwargs):
         to select a subset of paths.
         """
         if not self.interface.multi:
-            if datatype == 'array':
+            if not len(self):
+                return []
```
philippjfr: Another bug fix required to get annotations working.

jlstevens:
One quick comment for now...it might be nice to enable the hover tool for `.annotated` because then you would easily see the extra data that was added.

holoviews/annotators.py
```diff
@@ -0,0 +1,454 @@
...
+                                       'padding': 0.1}, doc="""
+        Opts to apply to the element.""")
+
+    object = param.ClassSelector(class_=Element, doc="""
+        The Element to edit and annotate.""")
+
+    num_objects = param.Integer(default=None, bounds=(0, None), doc="""
+        The maximum number of objects to draw.""")
+
+
```
jlstevens: Double spacing here.

philippjfr:
>it might be nice to enable the hover tool for .annotated because then you would easily see the extra data that was added.

That would mean adding bokeh specific options somewhere.

jlstevens:
> That would mean adding bokeh specific options somewhere.

Good point. You could set the option if you detect that bokeh is the current backend though (probably via `Store`). It is a nicety at that point and I'll leave it up to you.

jbednar:
I'd favor turning on hover by default for the Bokeh implementation whether or not you make it configurable.

jlstevens:
As appveyor tests aren't working right now and travis is passing, I'll go ahead and merge. There is more work to be done (e.g improvements to the docs) but I think the API is about as reasonable as we can make it for the functionality offered.

Merging.

PR #4087, opened 2019-11-06
Ensure plot minimum span causes RangeXY dependent operations to recompute

jbednar:
Fixes https://github.com/pyviz/holoviews/issues/4086 .


jbednar:
Thanks. A lot more complex than my version, but seems to work the same for the test cases I had.

philippjfr:
>Thanks. A lot more complex than my version, but seems to work the same for the test cases I had.

True, but it will now correctly respect the `default_span` plot option. In theory I should also trigger a recompute if `padding` is enabled but I might leave that for later.

PR #4088, opened 2019-11-06
Control sort from .overlay/.layout

DancingQuanta:
Currently, the elements of a kdim subjected to `.overlay` or `.layout` are sorted by default.

This PR is to explore how to let a user disable sorting through these methods.
The proposed API is
```
obj.overlay('kdim', sort=False')
```

### DIFF
--- a/holoviews/core/ndmapping.py
+++ b/holoviews/core/ndmapping.py
@@ -298,8 +298,9 @@ class MultiDimensionalMapping(Dimensioned):
         group_type = group_type if group_type else type(self)
         dimensions = [self.get_dimension(d, strict=True) for d in dimensions]
         with item_check(False):
+            sort = kwargs.pop('sort', True)
             return util.ndmapping_groupby(self, dimensions, container_type,
-                                          group_type, sort=True, **kwargs)
+                                          group_type, sort=sort, **kwargs)
 
 
     def add_dimension(self, dimension, dim_pos, dim_val, vdim=False, **kwargs):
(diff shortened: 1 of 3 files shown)


DancingQuanta:
I think I have identified three issues here and dealt with them.
* Test `test_idxmapping_groupby_unsorted` used incorrect data to test against where a test list is sorted instead of unsorted
* `ndmapping_groupby.groupby_pandas` does not pass `sort` to `group_type`, `container_type` or `pd.DataFrame.groupby`. In all three cases, `sort=True` by default.
* `MultiDimensional.groupby`  passes `sort=True` to `ndmapping_groupby` which does not make any use of `sort` passed to `gropuby`.

I tried to resolve above issues and made sure to proprogate `sort` to newer `MultiDimensional` objects. I wanted to add `sort` to `get_param_values` however this conflicts with `sort` in function arguments and I have no idea how to prioritize one over other.

DancingQuanta:
Is `groupby_python` still needed? The test suite include pandas dependency and so this function `groupby_python` never get tested.

DancingQuanta:
The `overlay' and 'layout' methods are now able to pass `sort` to the `groupby` after these changes in the PR.
@philippjfr Can you review ? This PR ended up fixing bugs in more than 2 years old code and so i am not sure if I have introduced any behaviour changes beyond what I am aiming for with `overlay` and `layout`.

jlstevens:
The functionality seems fine to me!

At a glance, the code also seems fine and I don't think there are backwards compatibility issues as `sort` should just override the behavior. That said, @philippjfr might spot some issues that I haven't.

philippjfr:
Thanks for contributing @DancingQuanta, much appreciated.

PR #4105, opened 2019-11-25
Passing extra parameters through rasterize operation

jlstevens:
This PR is *not* ready right now: the goal is to get a discussion going about how exactly we can dispatch parameters from a generic parameterized function to delegated parameterized functions.

The code path is tricky but here are the key points of what I am trying to achieve together with the various caveats:

1. `Operation` is given the class attribute `_allow_extra_keywords=False` by default. This preserves the current behavior for most operations. Classes that need to dispatch keywords to parameters can override this to `True` (which all classes will then inherit). This sets the corresponding argument in `param.ParamOverrides`.

2. I can't use param to warn here as the warning is determined by the set of parameterized functions that can be dispatched over. It would be nice to generalize but with this approach HoloViews, not param is doing the warning. 

3. This means I have to also silence unknown parameter warnings in the `instance` method which get used when `rasterize` is called. This is why I am filtering with `{k:v for k,v in params.items() if k in self_or_cls.param}`.

4. The `_process` method involves a loop over 'transforms' (rasterizers) and then over the layers of any overlay passed in. I have to filter out unknown parameters but track which one could get used by one of the rasterizers *whether or not there is an element present for it to apply to*. It might be nicer to warn about parameters that don't actually apply to a specific overlay but it is also reasonable to keep `rasterize` general in what it accepts in case the contents of overlays change.

5. The `process_element` method clobbers `self.p` when it gets used with a new `ParamOverrides` which means `.extra_keywords()` doesn't track the extra keywords properly. This is why it is commented out for now: it seems we need a way to merge/update into an existing `ParamOverrides` object.

As you can see this is a little hairier than I hoped and it isn't entirely obvious how this could be abstracted to work at the param level in a nice way. The current code shows an outline of what we might want to do but right now it is more of a pointer to the various problems!

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -111,8 +111,8 @@ class ResamplingOperation(LinkableOperation):
 
     @bothmethod
     def instance(self_or_cls,**params):
-        inst = super(ResamplingOperation, self_or_cls).instance(**
-                                        {k:v for k,v in params.items() if k in self_or_cls.param})
+        filtered = {k:v for k,v in params.items() if k in self_or_cls.param}
+        inst = super(ResamplingOperation, self_or_cls).instance(**filtered)
         inst._precomputed = {}
         return inst
 
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -47,6 +47,7 @@ class LinkableOperation(Operation):
         interactive, e.g. when trying to display an interactive plot a
         second time.""")
 
+    _allow_extra_keywords=True
 
 class ResamplingOperation(LinkableOperation):
     """
@@ -110,7 +111,8 @@ class ResamplingOperation(LinkableOperation):
 
     @bothmethod
     def instance(self_or_cls,**params):
-        inst = super(ResamplingOperation, self_or_cls).instance(**params)
+        inst = super(ResamplingOperation, self_or_cls).instance(**
+                                        {k:v for k,v in params.items() if k in self_or_cls.param})
         inst._precomputed = {}
         return inst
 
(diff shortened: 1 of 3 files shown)


philippjfr:
>The process_element method clobbers self.p when it gets used with a new ParamOverrides which means .extra_keywords() doesn't track the extra keywords properly. This is why it is commented out for now: it seems we need a way to merge/update into an existing ParamOverrides object.

In the long term we should add a proper implementation of .update to ParamOverrides, for now something like this should work:

```python
if self._allow_extra_keywords:
    self.p._extra_keywords.update(self.p._extract_extra_keywords(params))
    params = {k: v for k, v in params.items() if p not in self.p._extra_keywords}
self.p.update(params)
self.p._check_params(params)
```

philippjfr:
>also reasonable to keep rasterize general in what it accepts in case the contents of overlays change.

Agreed, I'm fine with these semantics.

jbednar:
How does that implementation differ from a proper one?

jlstevens:
Travis tests are now passing after a rebase and a commit based on Philipp's suggestion. Obviously more tests will be needed and this still feels like a bunch of hackery.

@jbednar @philippjfr In particular, even if I update the `update` method of `ParamOverrides`, I will still have an ugly `if hasattr(self, 'p')` branch which I don't like. Any other ideas?

holoviews/operation/datashader.py
```diff
@@ -1128,15 +1130,29 @@ class rasterize(AggregationOperation):
                    (Path, aggregate)]
 
     def _process(self, element, key=None):
+        # Potentially needs traverse to find element types first?
+        all_allowed_kws = set()
+        all_supplied_kws = set()
         for predicate, transform in self._transforms:
-            op_params = dict({k: v for k, v in self.p.items()
+            op_params = dict({k: v for k, v in (list(self.p.items())+list(self.p.extra_keywords().items()))
```
philippjfr:
```suggestion
            op_params = dict({k: v for k, v in self.p.items()
```

jlstevens:
I've added the key unit tests I wanted so once again ready to review/merge.

philippjfr:
Looks good to me!

PR #4106, opened 2019-11-27
Sankey: toggle node sorting in ascending breadth.

mgsnuno:
```python
import pandas as pd
import holoviews as hv

hv.extension("bokeh")

sankey = hv.Sankey(
    pd.DataFrame(
        [
            ["B", "X", 2],
            ["A", "X", 5],
            ["A", "Y", 7],
            ["A", "Z", 6],
            ["B", "Y", 9],
            ["B", "Z", 4],
            ["Y", "F", 4],
            ["Y", "C", 12],
            ["Z", "C", 5],
            ["Z", "F", 5],
            ["X", "F", 7],
        ],
        columns=["source", "target", "value"],
    )
)
sankey.opts(
    width=600,
    height=300,
    edge_color="target",
    node_color="source",
    color_index=0,
    node_sort=True,  #default
```
sorted:
sorted
not sorted:
not_sorted

This is important when comparing Sankey plots for the same dataset while filtering several options. If the sorting is handled by the order of the dataframe passed to Sankey (hence by the user) then nodes not getting sorted allows a straightforward comparison of the plots.

### DIFF
--- a/holoviews/element/sankey.py
+++ b/holoviews/element/sankey.py
@@ -33,6 +33,9 @@ class _layout_sankey(Operation):
     iterations = param.Integer(default=32, doc="""
         Number of iterations to run the layout algorithm.""")
 
+    node_sort = param.Boolean(default=True, doc="""
+        Sort nodes in ascending breadth.""")
+
     def _process(self, element, key=None):
         nodes, edges, graph = self.layout(element, **self.p)
         params = get_param_values(element)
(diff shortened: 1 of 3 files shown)


mgsnuno:
@philippjfr correct me if I'm wrong but these changes are not causing the failed test.

philippjfr:
No they aren't you are right, our appveyor setup broke at some point. I'll take a look at that separately. The changes here look good to me, thanks for contributing!

Issue #4116, opened 2019-12-03
glaciers app broken on master


jonmmease:
I haven't been able to track it down, but there seems to be a problem with the glaciers demo (https://github.com/pyviz-topics/examples/tree/master/glaciers) on master (as of 93294f38c0cad0e9bbaf24952d98e15a2f55aa42). It works fine with latest release (1.12.7).

The notebook displays the following warning (which is not present with 1.12.7):
![Screenshot_20191203_084646](https://user-images.githubusercontent.com/15064365/70056430-72eccf80-15a9-11ea-805f-21f932a3d284.png)

And the datashader overlay isn't showing up

![Screenshot_20191203_084817](https://user-images.githubusercontent.com/15064365/70056563-a9c2e580-15a9-11ea-9250-7dc732d61538.png)

jbednar:
@jlstevens , that warning sounds related to your rasterize changes.

jbednar:
The warnings are caused by `datashade` inheriting from both `rasterize` and `shade` while sharing the same arguments; previously HoloViews and Param had been silent about unused arguments, but now the fact that `shade` is consuming `cmap` is not known by `rasterize`.  @jlstevens will look into what to do about that, but in the meantime, the warning can be avoided by replacing `datashade(elem, ..., cmap=X)` with `shade(rasterize(elem, ...), cmap=X`.  I don't _think_ this explains anything about why the data is not showing up, though.

jlstevens:
@jonmmease I can reproduce the warning but not the lack of datashaded output on current HoloViews master (1.13.0a13.post12+g605ba06d5):

![image](https://user-images.githubusercontent.com/890576/70352659-41863500-1831-11ea-8a79-03047bcfcc58.png)

jlstevens:
Actually, I can reproduce when it comes to the final panel dashboard:

![image](https://user-images.githubusercontent.com/890576/70352810-9aee6400-1831-11ea-811a-21124ac10297.png)

Odd that the output no longer appears!

jlstevens:
@jonmmease #4121 should fix the warning and @philippjfr has a fix for the datashaded output not showing up.

jonmmease:
great, thanks!

jlstevens:
@jonmmease My parameter warning fix has been merged and the annotators PR (which contains the fix for the datashader output not showing up) has also been merged. I think this PR can be closed now but maybe you would first like could confirm that this issue has now been addressed?

[...]

PR #4119, opened 2019-12-05
Add crossfilter mode to link_selections

jonmmease:
## Overview
This PR introduces a new `mode` argument to the `link_selections` function that was added in https://github.com/holoviz/holoviews/pull/4044.

The default value, `"overwrite"`, matches the existing behavior where each selection replaces the prior selection across all elements.

In the new `"crossfilter"` mode, each selection replaces any existing selection on the current element, and selections across all elements are combined with an intersection operation.

To make the crossfilter state easier to interpret, the selected region for each element on which a selection has been performed is displayed in a lighter shade of the selection color. For box-select on scatter-like 2D elements, the selection region is displayed as a `Bounds` element.  For histogram elements, the selected region is displayed as an overlayed histogram indicating which bars have been selected.

See the GIFs below for examples of what this looks like.  Do these selection region indicators make sense to others?

Additionally, this PR uses the `PlotReset` stream to reset all selections when the plot is reset.  This works for the bokeh backend, but the `PlotReset` stream is not yet implemented for the Plotly backend.

## Examples
Example notebook at https://anaconda.org/jonmmease/cross_filter_link_selections_pr
### Bokeh scatter/histogram example
crossfilter1

### Plotly scatter/histogram/scatter3d/table example
crossfilter2

### Bokeh violin/bivariate/distribution/histogram example
crossfilter3

### Bokeh datashader example
crossfilter4

## Glaciers Example
Note: I believe this functionality is enough to rewrite the glaciers app once https://github.com/holoviz/holoviews/issues/4116 is fixed.


jbednar:
Cool!  Note that the behavior here is a bit different from in the Glaciers app; for Glaciers subsequent selections always act on the already selected subset, even if done in the same element as previously selected.  E.g. if selected on the map view, a subsequent selection on the map view selects the intersection between the current selection and the new selection; it doesn't replace the old selection.

What's implemented here seems like it would often be more useful, particularly as it displays what the old selection is, but maybe that is an additional mode option?

jonmmease:
> Note that the behavior here is a bit different from in the Glaciers app

Ahh, ok, thanks for pointing that out.  That behavior is actually a little easier to implement since you don't need to track the source of the selections, but I personally find it to be a bit disorienting.

> but maybe that is an additional mode option?

That's a possibility.  A full generalization here would be to have separate options for the within element operation (overwrite, intersection, union) and the cross element operation (overwrite, intersection, union).  

But, to do this well would take a bit of rework I think.  Displaying the selected region in the presence of union/intersection operations would require some shapely-style operations to merge polygons (which we would probably want to implement in `spatialpandas`).  And, we may want to define a higher-level selection object. Right now, selections are represented as dim expressions. But to support efficiently applying geometry operations on selections, I think it would be better for selections to be geometry-like objects that know how to produce dim expressions.

All that said, if we want to merge this `mode="crossfilter"` support now, we would still be a able to support this API in the context of the enhancements above.  At that point `mode` would become a high-level specification of common interaction modes that would be implemented in terms of these lower-level constructs. So maybe `mode="crossfilter"` would be equivalent to something like `within_op="overwrite", between_op="intersection"`.

jbednar:
> Displaying the selected region in the presence of union/intersection operations would require some shapely-style operations to merge polygons

I'm inclined to be less ambitious, and would be happy with either not showing the selection bounding box in the `within_op="intersection"` case, or just showing all the selected regions (letting the user see that the actual selection is the intersection of the multiple bounding boxes shown).  And in general being able to turn off showing the selection box would be useful in some cases.

jonmmease:
Ok, I've replaced `mode` with three orthogonal args: `element_op`, `cross_element_op`, and `show_regions`.  The example notebook above has been updated with the new syntax.

New default: show_regions=True, cross_element_op="intersect", element_op="overwrite"
crossfilter1

element_op="intersect", cross_element_op="intersect"
crossfilter2

element_op="intersect", cross_element_op="intersect", show_regions=False
crossfilter3

element_op="difference", cross_element_op="overwrite"
crossfilter4

jbednar:
Fabulous! Is there anything preventing the Glaciers app from being rewritten to use this now?

jonmmease:
> Is there anything preventing the Glaciers app from being rewritten to use this now?

I don't think so, I'll try it again later today or this evening.  in the process, I'll confirm that https://github.com/holoviz/holoviews/issues/4116 is fixed on master.

jonmmease:
I ran into some issues while porting the glaciers example. Here are PRs that address these: https://github.com/holoviz/holoviews/pull/4137, https://github.com/holoviz/geoviews/pull/414.

With these PRs, I can get pretty close!

jonmmease:
Here is a port of the glaciers example to this PR (combined with https://github.com/holoviz/holoviews/pull/4137, https://github.com/holoviz/geoviews/pull/414): https://anaconda.org/jonmmease/glaciers_link_selections

It's a lot shorter and simpler I think!

The main difference is the change from displaying the glaciers on the map with `rasterize` and a custom color scale, to using `datashade` and letting `link_selections` control the color scale.  Also, right now `link_selections` only uses `BoundsXY` and `box_select` when dealing with Histograms, whereas the current glaciers example uses `BoundsX`/`xbox_select` for the histograms.  I think this would eventually make sense as an option `link_selections`.

Here's some example usage:
glaciers_link_selections

jbednar:
Yes, that's much, much simpler!  BoundsX and BoundsY are definitely clearer when they are appropriate, such as in this case, so it would be nice to support that.

@fmaussion, can you check out the new notebook and see what you think?

fmaussion:
> @fmaussion, can you check out the new notebook and see what you think?

I can't test it right now, but if I understand well this allows to make selections withing selections, which is nice! I'm happy to have a look at the updated app on https://github.com/pyviz-demos/glaciers when it is adapted

jbednar:
Here the interesting part is not the behavior, but the code -- it should now be possible to make something like your original dashboard with almost no code. So the task is to look at the behavior, but also the code and how they relate. I'll put up a running version once the code is merged and available in a dev release.

philippjfr:
@jonmmease, is this ready to review and merge?

jonmmease:
@jbednar asked me to take a look at having selections preserve the original color(scale) of elements and I haven't had a chance to look at that yet.

If we want to change this behavior now, then we should probably hold off. If we want to keep this approach and add something new later then I think it's ready for review.

jonmmease:
I've made several updates:

 - The default behavior is now for the selected element layer to have the original color of the element and for the unselected layer to be gray. A specific selection color is still available as well.
 - The initial state is now for everything to be selected, so you don't see the gray background until a selection is performed.
 - Added support for elements that are have a `cmap` sytle property. This was needed to support the result of `rasterize`.
 - The selection region, when enabled, is now always a darker shade of the background color. This is a lot simpler, and I think it looks better, especially since there is not necesarily a single selection color anymore.

## Basic
crossfilter1

## Plotly
crossfilter2

## Distributions
crossfilter3

## Datashader
crossfilter4

## Glaciers
glaciers_link_selections

This example can now keep the original approach of rasterizing glaciers with a custom colormap and colorbar.

jonmmease:
Looking into the failing tests...

jbednar:
Looks fabulous!  People can now enable cross filtering with essentially zero changes to their original plotting code and styling, giving really useful functionality with no cost!  

My vote is for the unselected regions to be about half as visible as they are now, i.e. for a white background for them to be a much lighter gray. Right now I find the unselected regions to be a bit distracting; I think they should literally fade into the background a bit, as they are for comparison and context not for close study. Maybe make what is currently dark gray be the same color as what is currently medium gray, and make what is currently medium gray be half as dark as it is how? @philippjfr may also have opinions here.

jbednar:
@philippjfr, can you think about how this relates to gridmatrix (e.g. http://holoviews.org/gallery/demos/bokeh/iris_density_grid.html)? Does the interlinking in gridmatrix now reduce to simply calling this method?

jbednar:
BTW, in the Datashader example above, the selected Datashader plot appears to be drawn on top of the selection outline?

image

Presumably it should be the other way around?

philippjfr:
Gridmatrix interlinking is all client side and therefore comes for free anyway.

jbednar:
> Gridmatrix interlinking is all client side and therefore comes for free anyway.

We still have to maintain that code, but of course we'll do so since it provides client-side (exportable) functionality, which is complementary. What I'm asking is whether (apart from client-side or server-side considerations) the gridmatrix functionality is now a subset of this general crossfiltering functionality? If so that makes things easier to talk and reason about. (And of course to someday hope for full client-side cross-filtering support for the cases where all data is in the client.)

philippjfr:
There is no code associated with that, we get it for free from bokeh. I'm not sure how to answer your question though. The difference is that bokeh exposes selection/nonselection options to control the styling of each. Additionally the UI is different in that you can apply multiple selections using shift select. So I wouldn't say they are subsets.

jonmmease:
@jbednar 
In https://github.com/holoviz/holoviews/pull/4119/commits/6e1064e66560eed8ccff633161a23e8e4c8af8cd I updated the ordering of the selected regions so that the box outline is drawn on top.

In https://github.com/holoviz/holoviews/pull/4119/commits/6f8c969a63f85680fdf0f937be220487dd51c08c I lightened the shades of gray used for the unselected layers.

Here's how things look now:
Screenshot_20200101_105031

@philippjfr, when you have a chance, could you take a look at https://github.com/holoviz/holoviews/pull/4119/commits/ac4025ba6a99eac9e3c220bce6e38f35e14a7f47?  My goal here was to ensure that `rasterize.instance(**kwargs)(element)` is always equivalent to `rasterize(element, **kwargs)`.  This wasn't always true before, because in the former case the `kwargs` did not get propagated to the rasterization implementations.

jbednar:
Thanks! The darker gray still looks too dark to me; I'd want it to visually group with the lighter gray, because they are both actually the background and not the main point of the plot anymore.  I.e. it should be distinguishable from the lighter gray, but not dramatically different, so that together the grey bits fade into the background, while the foreground is quite clearly distinct.  The coloration would need to be adjusted for the map plot in the Glaciers example, where fading into the background requires darker colors and not lighter ones, but that's an unusual case and it's fine for it to require custom styling when that occurs (i.e. when one pane in a layout has a dark background and the rest don't).

jonmmease:
As things stand right now, it's not possible to configure separate unselected colors per plot or plot type. So the gray for the selected histogram bars matches the gray for the selection boxes. I think it would work fine to have the histogram color lighter than the box color, but I can't picture a clear API at the moment for configuring separate colors per subplot.

jonmmease:
In https://github.com/holoviz/holoviews/pull/4119/commits/442eea8b5b9f1bb4264977963a06384a422bb7b8 I updated things so that selected histogram bars are just slightly darker than the unselected bars, while keeping the selected boxes a darker shade.

I also gave the box lines width 1 (instead of 2) to hopefully be a bit less distracting.  Here is the updated example. Note that the bottom left histogram was created with a white line color.

Screenshot_20200102_104312

philippjfr:
That looks good to me. Should we also expose the alpha of the selected and unselected elements?

jonmmease:
> Should we also expose the alpha of the selected and unselected elements?

Yeah, I suppose we could. Though I think I'd prefer not to block this PR on that.

jbednar:
That looks good to me too, thanks!  I'm happy with it, but just as an idea -- would it make sense to have a fixed shade of gray for the background bits and use alpha changes between the selected and unselected ones rather than different shades of gray, so that they would automatically blend into the background (becoming darker when the background is dark, and lighter when the background is light)?  That might alleviate or eliminate the issue with the map pane for the Glaciers example.

jonmmease:
> would it make sense to have a fixed shade of gray for the background bits and use alpha changes between the selected and unselected ones rather than different shades of gray, so that they would automatically blend into the background...?

Yeah, that's a good idea and certainly something worth experimenting with.

philippjfr:
Would you want to merge this now and follow up with that later? Hoping to release hv 1.13.0 containing all of this work early next week.

jonmmease:
> Would you want to merge this now and follow up with that later?

That would be my preference since I don't think I'll be able to do much more on this before next week.

philippjfr:
Great, maybe I'll have some time to look at it too. I'd definitely like to start writing a user guide around this stuff.

holoviews/operation/datashader.py
```diff
@@ -1334,12 +1334,12 @@ class rasterize(AggregationOperation):
...
     aggregator = param.ClassSelector(class_=(ds.reductions.Reduction, basestring),
-                                     default=None)
+                                     default='default')
 
     interpolation = param.ObjectSelector(
-        default='bilinear', objects=['linear', 'nearest', 'bilinear', None, False], doc="""
+        default='default', objects=['default', 'linear', 'nearest', 'bilinear', None, False], doc="""
         The interpolation method to apply during rasterization.
-        Defaults to linear interpolation and None and False are aliases
+        Default depends on element type, None and False are aliases
```
jbednar:
```suggestion
        Default depends on element type; None and False are aliases
```

Is that really true? It seems to require the string 'default' now, not None or False.

holoviews/operation/datashader.py
```diff
@@ -1362,15 +1362,23 @@ class rasterize(AggregationOperation):
                    (Points, aggregate),
                    (Curve, aggregate),
                    (Path, aggregate),
-                   (type(None), shade) # To handles parameters of datashade 
+                   (type(None), shade) # To handles parameters of datashade
```
jbednar:
```suggestion
                   (type(None), shade) # To handle parameters of datashade
```

holoviews/plotting/bokeh/chart.py
```diff
@@ -400,7 +400,7 @@ class HistogramPlot(ColorbarPlot):
 
     _nonvectorized_styles = ['line_dash', 'visible']
 
-    selection_display = BokehOverlaySelectionDisplay()
+    selection_display = BokehOverlaySelectionDisplay(color_prop=['color', 'fill_color'])
```
jbednar: Is this where we'd try varying alpha?

holoviews/plotting/bokeh/selection.py
```diff
@@ -23,8 +23,44 @@ def alpha_opts(alpha):
...
+        if element_name != "Histogram":
+            # Darken unselected color
+            region_color = linear_gradient(unselected_color, "#000000", 9)[3]
+            options["color"] = region_color
+            options["line_width"] = 1
+        else:
+            # Darken unselected color slightly
+            region_color = linear_gradient(unselected_color, "#000000", 9)[1]
+            options["fill_color"] = region_color
+            options["color"] = region_color
```
jbednar: And then we'd manipulate alpha here?

holoviews/plotting/plotly/selection.py
```diff
@@ -20,7 +20,45 @@ def _build_element_layer(
...
+        options = {}
+
+        if element_name != "Histogram":
+            # Darken unselected color
+            region_color = linear_gradient(unselected_color, "#000000", 9)[3]
+            if "line_width" in allowed_keywords:
+                options["line_width"] = 1
+        else:
+            # Darken unselected color slightly
+            region_color = linear_gradient(unselected_color, "#000000", 9)[1]
```
jbednar: And also adjust alpha here?  Probably should pull these out into a function (maybe one that can be replaced as a parameter) for computing the new color given some arguments.

holoviews/selection.py
```diff
@@ -170,8 +220,26 @@ def _expr_stream_updated(self, hvobj, selection_expr, bbox):
 
 class link_selections(_base_link_selections):
     selection_expr = param.Parameter(default=None)
-    unselected_color = param.Color(default="#99a6b2")  # LightSlateGray - 65%
-    selected_color = param.Color(default="#DC143C")  # Crimson
+    unselected_color = param.Color(default="#e6e9ec")
+    selected_color = param.Color(default=None, allow_None=True)
```
jbednar: I think these parameters need docstrings. E.g. how is None interpreted for selected_color?

holoviews/util/transform.py
```diff
@@ -223,6 +223,7 @@ def __round__(self, ndigits=None):
     # Unary operators
     def __neg__(self): return dim(self, operator.neg)
     def __not__(self): return dim(self, operator.not_)
+    def __invert__(self): return dim(self, operator.inv)
```
jbednar: Thanks. Are there any other operators missing?

jonmmease: No, I don't think there any other other missing.

holoviews/plotting/bokeh/chart.py
jonmmease: Yeah, I think we'd need to list out all of the alpha properties that need to be overridden here

holoviews/plotting/bokeh/selection.py
jonmmease: Yes, I think so

holoviews/selection.py
jonmmease: Done

jbednar:
Ok, I'm very happy with it.  I don't think I know enough about how Bokeh handles alpha to mess with that, but if @philippjfr wants to try that approach, he can.  Otherwise, seems ready to me!

holoviews/element/geom.py
```diff
@@ -63,8 +64,27 @@ def _get_selection_expr_for_stream_value(self, **kwargs):
                     (dim(ydim) >= y0) & (dim(ydim) <= y1)
             )
 
-            return selection_expr, bbox
-        return None, None
+            # region_element = Bounds(kwargs['bounds'])
+            x0, y0, x1, y1 = kwargs['bounds']
+            region_element = Curve(([x0, x1, x1, x0, x0], [y0, y0, y1, y1, y0]))
```
philippjfr: Why did you switch to a Curve element here? I think now that it's merged I'd probably swap this to `Rectangles`.

jlstevens:
I've done a superficial pass over this PR and the functionality seems to be very nice. One thing that jumps out at me is that I don't like the names `cross_element_op` and `element_op`. Here is what is supported:

```python
    cross_element_op = param.Selector(
        ['overwrite', 'intersect'], default='intersect', doc="""
        Set operation to use to combine selections across different elements"""
    )
    element_op = param.Selector(
        ['overwrite', 'intersect', 'union', 'difference'], default='overwrite', doc="""
        Set operation to use to combine successive selections on the same element"""
    )
```

To me, there are two concepts here: the 'selection mode' which can be stateless (`element_op='overwrite'`) or stateful (the other options) when repeatedly applying a selection of the same thing. The other concept is the actual *cross* filtering, which is the  `cross_element_op`.


It is possible that I don't fully understand how these parameters work, but I propose the names `selection_mode` and `filter_mode` to replace `cross_element_op` and `element_op`. Does that make sense @jonmmease @philippjfr ? I think it makes more sense to me at least!

holoviews/core/element.py
```diff
@@ -378,6 +378,8 @@ class Element3D(Element2D):
 
     __abstract = True
 
+    _selection_streams = ()
```
jlstevens: I'm a little concerned about elements having more state attached to them that needs to reasoned about (e.g for pickling?)

jonmmease: `_selection_streams` is an immutable class-level property that indicates which stream types can be used on the element to generate selections, so I don't think it should be an issue here for pickling.

jlstevens: Looking at [line 296](https://github.com/holoviz/holoviews/blob/26a0afb7c0b58c625b48769644a836c3195a07a3/holoviews/selection.py#L296) of `selection.py` I see the named tuple being built:

```
return _SelectionStreams(
            colors_stream=colors_stream,
            exprs_stream=exprs_stream,
            cmap_streams=cmap_streams,
        )
```

Where I see that `exprs_stream = _Exprs(exprs=[True, None])` and `_Exprs` is defined using `Stream.define`. That would make it an instance of a stream (i.e stateful) as opposed to a just a type (i.e just the stream class). 

If `_selection_streams` really does contains the types and not the instances (which doesn't seem to be the case unless I am misunderstanding things!) then it should be called something like `_selection_stream_types`.

jlstevens: It is true that the named tuple is immutable, but the things inside it are not which means it becomes the responsibility of the code using `_selection_streams` not to mutate anything. I haven't checked whether this is the case yet but it could be that these streams are not actually used directly anywhere but are used to create new objects elsewhere....

jonmmease: > If _selection_streams really does contains the types and not the instances (which doesn't seem to be the case unless I am misunderstanding things!) then it should be called something like _selection_stream_types.

Yes, it should be named `_selection_stream_types`. I'll make that change.

philippjfr: I'm also confused now, as @jlstevens points out doesn't this line make an instance: https://github.com/holoviz/holoviews/blob/26a0afb7c0b58c625b48769644a836c3195a07a3/holoviews/selection.py#L280

jonmmease: The named tuple instances created in that `_build_selection_streams` method are only stored by the `link_selections` `Operation` object.  The `Element` class-level tuple currently named `_selection_streams` (which should be named `_selection_stream_types`) is what stores a tuple of `LinkedStream` types, not any of the named tuple types defined in `selection.py`.

Does that help at all?

philippjfr: Ah, yes, still digging through the code. Thanks for explaining it. Let me know if you want me to take over the PR since you indicated you won't have more time until later next week.

philippjfr:
Do we need a new param release for https://github.com/holoviz/param/pull/367?

holoviews/core/element.py
jlstevens: Yes, thanks for explaining! I think it will be fine after the renaming...

jbednar:
I'm happy for a new Param release to be tagged, but was not sure if it was needed or not.

jonmmease:
> Let me know if you want me to take over the PR since you indicated you won't have more time until later next week

Thanks @philippjfr, yeah that would be great if you're up for it!

> Do we need a new param release for holoviz/param#367?

This PR doesn't directly depend on that param change, but without that change this PR won't preserve the original colorscale for `datashade`/`rasterize` operations.

philippjfr:
Introduced two small regressions here:

- [ ] The bounding regions sometimes appear filled
- [ ] The unselected region on datashader does not appear correctly (likely due to the way I'm applying DynamicMap pipelines)

holoviews/plotting/bokeh/chart.py
philippjfr: @jbednar What was the discussion about alpha about? Is it something I should tackle now?

holoviews/core/spaces.py
```diff
@@ -625,6 +625,10 @@ class Callable(param.Parameterized):
          The list of inputs the callable function is wrapping. Used
          to allow deep access to streams in chained Callables.""")
 
+    operation_kwargs = param.Dict(default={}, constant=True, doc="""
+        Potential dynamic keyword arguments associated with the
```
jbednar: I find the word "potential" confusing here.  Does it mean something like:

```
       Optional keyword arguments to be supplied to the operation if it needs to be applied again.
```

holoviews/plotting/bokeh/chart.py
jbednar: @philippjfr , the discussion of alpha is [above](https://github.com/holoviz/holoviews/pull/4119#issuecomment-570268274), but I'll summarize the status as I know it. The  underlying idea here is that we want to "mute" the unselected glphs so that they fade more into the background, letting people focus on the selected ones.  Right now, link_selections is doing that by darkening the unselected color, which I believe is appropriate only for dark backgrounds (making unselected items more similar to the background); it should do the opposite for light backgrounds. But even if we had a switch in link_selections between darkening and lightening, it wouldn't work for the glaciers example, where the map plot has a dark background and the rest are light. Hence my proposal to decrease the opacity of unselected glyphs instead of changing their color. I believe that the result would be to make the unselected glyphs fade into the background in all cases, but I haven't tried it. To try it, you'd look at the various places where "alpha" is mentioned in the comments on this PR.

holoviews/selection.py
```diff
@@ -215,6 +217,11 @@ class link_selections(_base_link_selections):
         Determines how to combine selections across different
         elements.""")
 
+    index_cols = param.List(default=None, doc="""
+        If provided selection switches to index mode where all queries
+        are expressed solely in terms discrete values along the
```
jbednar:
```suggestion
        are expressed solely in terms of discrete values along the
```

holoviews/selection.py
```diff
@@ -215,6 +217,11 @@ class link_selections(_base_link_selections):
...
+        index_cols.""")
```
jbednar: Is this accurate?

```suggestion
        index_cols.  All Elements given to link_selections must define the index_cols, either as explicit dimensions or by sharing an underlying Dataset that defines them.""")
```

holoviews/streams.py
```diff
@@ -763,31 +763,33 @@ def __init__(self, parameterized, parameters=None, watch=True, **params):
...
-            raise ValueError("""
-The source of SelectionExpr must be an instance of an Element subclass,
-or a DynamicMap that returns such an instance
-            Received value of type {typ}: {val}""".format(
-            typ=type(source), val=source
-        ))
+            raise ValueError(
+                "The source of SelectionExpr must be an instance of an "
+                "Element subclass or a DynamicMap that returns such an "
+                "instance Received value of type {typ}: {val}".format(
```
jbednar:
```suggestion
                "instance. Received value of type {typ}: {val}".format(
```

philippjfr:
@jonmmease Two questions about the plotly support:

1) How do you reset selections once you've applied them?
2) There seems to be a regression which is causing only the first selection to get applied. Any idea where that might be coming from?

philippjfr:
@jonmmease So I looked at the Panel implementation and it seems that when a plotly plot is rerendered the events aren't set up again. If I set [this condition](https://github.com/holoviz/panel/blob/master/panel/models/plotly.ts#L154) to true it all works again but that presumably has other annoying side-effects.

philippjfr:
Okay, ignore me, I've found a quick fix and pushed it [here](https://github.com/holoviz/panel/pull/1109).

jbednar:
I've updated Linked_Selections.ipynb to focus on real data.  It's a bit heavier than I would like, but there's a lot to explain!  Remaining issues:

- [ ] The bokeh toolbar bug https://github.com/bokeh/bokeh/issues/9699 is very annoying
- [ ] I'm pretty sure I saw the Plotly plot working at first, but I don't think I touched that code and don't see it working now.
- [ ] There are issues with Run All, in part because of switching the backend to Plotly, but maybe for other reasons too
- [ ] In the Panel app, should show_regions affect already drawn regions? It does not, and I would have expected it to.
- [ ] There should be some Datashader example(s)
- [ ] I also still can't quite see why all elements with appropriate axes aren't supported, for range selections. Can't one select a range on any plot, regardless of what it's showing (e.g. heatmaps, images, even violin (one axis only)?

This is all really cool and useful, in any case!

jonmmease:
Just a heads up, 376b790161ebcce3c2d966712b63b3912333b6f1 seems to have caused something to break in the Histogram codepath.

The first `link_selections` cell in https://anaconda.org/jonmmease/cross_filter_link_selections_pr/notebook results in the following stack trace.

```
AttributeError                            Traceback (most recent call last)
<ipython-input-5-8eadbeb83fa6> in <module>
      2 # lnk_sels = link_selections.instance(show_regions=True, selected_color=None)# "#004400")
      3 lnk_sels = link_selections.instance()# "#004400")
----> 4 lnk_sels(layout1)

~/PyDev/repos/holoviews/holoviews/selection.py in __call__(self, hvobj, **kwargs)
     92 
     93         # Perform transform
---> 94         hvobj_selection = self._selection_transform(hvobj.clone(link=False))
     95 
     96         return hvobj_selection

~/PyDev/repos/holoviews/holoviews/selection.py in _selection_transform(self, hvobj, operations)
    142         elif isinstance(hvobj, (Layout, Overlay, NdOverlay, GridSpace)):
    143             data = OrderedDict([(k, self._selection_transform(v, operations))
--> 144                                  for k, v in hvobj.items()])
    145             new_hvobj = hvobj.clone(data)
    146             if hasattr(new_hvobj, 'collate'):

~/PyDev/repos/holoviews/holoviews/selection.py in <listcomp>(.0)
    142         elif isinstance(hvobj, (Layout, Overlay, NdOverlay, GridSpace)):
    143             data = OrderedDict([(k, self._selection_transform(v, operations))
--> 144                                  for k, v in hvobj.items()])
    145             new_hvobj = hvobj.clone(data)
    146             if hasattr(new_hvobj, 'collate'):

~/PyDev/repos/holoviews/holoviews/selection.py in _selection_transform(self, hvobj, operations)
    137                 return chart.selection_display.build_selection(
    138                     self._selection_streams, element, operations,
--> 139                     self._region_streams.get(element, None),
    140                 )
    141             return hvobj

~/PyDev/repos/holoviews/holoviews/selection.py in build_selection(self, selection_streams, hvobj, operations, region_stream)
    442             streams = [region_stream, selection_streams.style_stream]
    443             region = hvobj.clone(link=False).apply(update_region, streams)
--> 444             if getattr(hvobj, '_selection_dims') == 1:
    445                 layers.insert(0, region)
    446             elif isinstance(hvobj, Histogram):

AttributeError: 'Histogram' object has no attribute '_selection_dims'
```

philippjfr:
Thanks, easy fix.

jonmmease:
Thanks for the great work picking this up @philippjfr!  I'm working on getting caught up on your updates, but I'm getting an error in the "Brushing_and_Linking" example notebook in the first `link_selections` cell.

```python
link_selections(layout).cols(6)
```
```
AttributeError                            Traceback (most recent call last)
~/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/IPython/core/formatters.py in __call__(self, obj, include, exclude)
    968 
    969             if method is not None:
--> 970                 return method(include=include, exclude=exclude)
    971             return None
    972         else:

~/PyDev/repos/holoviews/holoviews/core/dimension.py in _repr_mimebundle_(self, include, exclude)
   1306         combined and returned.
   1307         """
-> 1308         return Store.render(self)
   1309 
   1310 

~/PyDev/repos/holoviews/holoviews/core/options.py in render(cls, obj)
   1393         data, metadata = {}, {}
   1394         for hook in hooks:
-> 1395             ret = hook(obj)
   1396             if ret is None:
   1397                 continue

~/PyDev/repos/holoviews/holoviews/ipython/display_hooks.py in pprint_display(obj)
    280     if not ip.display_formatter.formatters['text/plain'].pprint:
    281         return None
--> 282     return display(obj, raw_output=True)
    283 
    284 

~/PyDev/repos/holoviews/holoviews/ipython/display_hooks.py in display(obj, raw_output, **kwargs)
    253     elif isinstance(obj, (Layout, NdLayout, AdjointLayout)):
    254         with option_state(obj):
--> 255             output = layout_display(obj)
    256     elif isinstance(obj, (HoloMap, DynamicMap)):
    257         with option_state(obj):

~/PyDev/repos/holoviews/holoviews/ipython/display_hooks.py in wrapped(element)
    144         try:
    145             max_frames = OutputSettings.options['max_frames']
--> 146             mimebundle = fn(element, max_frames=max_frames)
    147             if mimebundle is None:
    148                 return {}, {}

~/PyDev/repos/holoviews/holoviews/ipython/display_hooks.py in layout_display(layout, max_frames)
    218         return None
    219 
--> 220     return render(layout)
    221 
    222 

~/PyDev/repos/holoviews/holoviews/ipython/display_hooks.py in render(obj, **kwargs)
     66         renderer = renderer.instance(fig='png')
     67 
---> 68     return renderer.components(obj, **kwargs)
     69 
     70 

~/PyDev/repos/holoviews/holoviews/plotting/renderer.py in components(self, obj, fmt, comm, **kwargs)
    361             doc = Document()
    362             with config.set(embed=embed):
--> 363                 model = plot.layout._render_model(doc, comm)
    364             return render_model(model, comm) if embed else render_mimebundle(model, doc, comm)
    365         else:

~/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/panel/viewable.py in _render_model(self, doc, comm)
    263         if comm is None:
    264             comm = state._comm_manager.get_server_comm()
--> 265         model = self.get_root(doc, comm)
    266 
    267         if config.embed:

~/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/panel/viewable.py in get_root(self, doc, comm)
    432         """
    433         doc = doc or _curdoc()
--> 434         root = self._get_model(doc, comm=comm)
    435         self._preprocess(root)
    436         ref = root.ref['id']

~/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/panel/layout.py in _get_model(self, doc, root, parent, comm)
    117         if root is None:
    118             root = model
--> 119         objects = self._get_objects(model, [], doc, root, comm)
    120         props = dict(self._init_properties(), objects=objects)
    121         model.update(**self._process_param_change(props))

~/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/panel/layout.py in _get_objects(self, model, old_objects, doc, root, comm)
    109                 child, _ = pane._models[root.ref['id']]
    110             else:
--> 111                 child = pane._get_model(doc, root, model, comm)
    112             new_models.append(child)
    113         return new_models

~/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/panel/pane/holoviews.py in _get_model(self, doc, root, parent, comm)
    234             else:
    235                 plot = self._render(doc, comm, root)
--> 236             plot.pane = self
    237             backend = plot.renderer.backend
    238             if hasattr(plot.renderer, 'get_plot_state'):

~/PyDev/repos/holoviews/holoviews/plotting/plot.py in pane(self, pane)
    129     @pane.setter
    130     def pane(self, pane):
--> 131         if (config.console_output != 'disable' and self.root and
    132             self.root.ref['id'] not in state._handles and
    133             isinstance(self.comm, JupyterComm)):

AttributeError: '_config' object has no attribute 'console_output'

:Layout
   .Points.I   :DynamicMap   []
      :Overlay
         .Points.I     :Points   [cyl,accel]   (mpg,cyl,displ,hp,weight,accel,yr)
         .Points.II    :Points   [cyl,accel]   (mpg,cyl,displ,hp,weight,accel,yr)
         .Rectangles.I :Rectangles   [x0,y0,x1,y1]
   .Points.II  :DynamicMap   []
      :Overlay
         .Points.I     :Points   [displ,accel]   (mpg,cyl,displ,hp,weight,accel,yr)
         .Points.II    :Points   [displ,accel]   (mpg,cyl,displ,hp,weight,accel,yr)
         .Rectangles.I :Rectangles   [x0,y0,x1,y1]
   .Points.III :DynamicMap   []
      :Overlay
         .Points.I     :Points   [hp,accel]   (mpg,cyl,displ,hp,weight,accel,yr)
         .Points.II    :Points   [hp,accel]   (mpg,cyl,displ,hp,weight,accel,yr)
         .Rectangles.I :Rectangles   [x0,y0,x1,y1]
   .Points.IV  :DynamicMap   []
      :Overlay
         .Points.I     :Points   [mpg,accel]   (mpg,cyl,displ,hp,weight,accel,yr)
         .Points.II    :Points   [mpg,accel]   (mpg,cyl,displ,hp,weight,accel,yr)
         .Rectangles.I :Rectangles   [x0,y0,x1,y1]
   .Points.V   :DynamicMap   []
      :Overlay
         .Points.I     :Points   [weight,accel]   (mpg,cyl,displ,hp,weight,accel,yr)
         .Points.II    :Points   [weight,accel]   (mpg,cyl,displ,hp,weight,accel,yr)
         .Rectangles.I :Rectangles   [x0,y0,x1,y1]
   .Points.VI  :DynamicMap   []
      :Overlay
         .Points.I     :Points   [yr,accel]   (mpg,cyl,displ,hp,weight,accel,yr)
         .Points.II    :Points   [yr,accel]   (mpg,cyl,displ,hp,weight,accel,yr)
         .Rectangles.I :Rectangles   [x0,y0,x1,y1]
```
I'm on bokeh 1.4.0 in case that's relevant.

jonmmease:
Looks like this was introduced in e532d49be54b9eddee41ebda5ec362a555682da2

philippjfr:
Which version of panel do you have? I'll be pinning Panel 0.8 as minimum which should fix the issue.

philippjfr:
Which version of panel do you have? I'll be pinning Panel 0.8 as minimum which should fix the issue.

jonmmease:
Thanks, that helps. Looks like I got something corrupted in my environment, but once I get that worked out I assume that will take care of it.

jonmmease:
Thanks, that helps. Looks like I got something corrupted in my environment, but once I get that worked out I assume that will take care of it.

jonmmease:
Thanks, that helps. Looks like I got something corrupted in my environment, but once I get that worked out I assume that will take care of it.

jonmmease:
Hmm, I just built a new environment, pulling holoviz dependencies from pyviz/label/dev and I'm getting a new error:

```
AttributeError                            Traceback (most recent call last)
~/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/IPython/core/formatters.py in __call__(self, obj, include, exclude)
    968 
    969             if method is not None:
--> 970                 return method(include=include, exclude=exclude)
    971             return None
    972         else:

~/PyDev/repos/holoviews/holoviews/core/dimension.py in _repr_mimebundle_(self, include, exclude)
   1305         combined and returned.
   1306         """
-> 1307         return Store.render(self)
   1308 
   1309 

~/PyDev/repos/holoviews/holoviews/core/options.py in render(cls, obj)
   1386         data, metadata = {}, {}
   1387         for hook in hooks:
-> 1388             ret = hook(obj)
   1389             if ret is None:
   1390                 continue

~/PyDev/repos/holoviews/holoviews/ipython/display_hooks.py in pprint_display(obj)
    280     if not ip.display_formatter.formatters['text/plain'].pprint:
    281         return None
--> 282     return display(obj, raw_output=True)
    283 
    284 

~/PyDev/repos/holoviews/holoviews/ipython/display_hooks.py in display(obj, raw_output, **kwargs)
    253     elif isinstance(obj, (Layout, NdLayout, AdjointLayout)):
    254         with option_state(obj):
--> 255             output = layout_display(obj)
    256     elif isinstance(obj, (HoloMap, DynamicMap)):
    257         with option_state(obj):

~/PyDev/repos/holoviews/holoviews/ipython/display_hooks.py in wrapped(element)
    144         try:
    145             max_frames = OutputSettings.options['max_frames']
--> 146             mimebundle = fn(element, max_frames=max_frames)
    147             if mimebundle is None:
    148                 return {}, {}

~/PyDev/repos/holoviews/holoviews/ipython/display_hooks.py in layout_display(layout, max_frames)
    218         return None
    219 
--> 220     return render(layout)
    221 
    222 

~/PyDev/repos/holoviews/holoviews/ipython/display_hooks.py in render(obj, **kwargs)
     66         renderer = renderer.instance(fig='png')
     67 
---> 68     return renderer.components(obj, **kwargs)
     69 
     70 

~/PyDev/repos/holoviews/holoviews/plotting/renderer.py in components(self, obj, fmt, comm, **kwargs)
    361             doc = Document()
    362             with config.set(embed=embed):
--> 363                 model = plot.layout._render_model(doc, comm)
    364             return render_model(model, comm) if embed else render_mimebundle(model, doc, comm)
    365         else:

~/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/panel/viewable.py in _render_model(self, doc, comm)
    355         if comm is None:
    356             comm = state._comm_manager.get_server_comm()
--> 357         model = self.get_root(doc, comm)
    358 
    359         if config.embed:

~/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/panel/viewable.py in get_root(self, doc, comm)
    526         """
    527         doc = doc or _curdoc()
--> 528         root = self._get_model(doc, comm=comm)
    529         self._preprocess(root)
    530         ref = root.ref['id']

~/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/panel/layout.py in _get_model(self, doc, root, parent, comm)
    134         if root is None:
    135             root = model
--> 136         objects = self._get_objects(model, [], doc, root, comm)
    137         props = dict(self._init_properties(), objects=objects)
    138         model.update(**self._process_param_change(props))

~/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/panel/layout.py in _get_objects(self, model, old_objects, doc, root, comm)
    124             else:
    125                 try:
--> 126                     child = pane._get_model(doc, root, model, comm)
    127                 except RerenderError:
    128                     return self._get_objects(model, current_objects[:i], doc, root, comm)

~/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/panel/pane/holoviews.py in _get_model(self, doc, root, parent, comm)
    227             plot = self._render(doc, comm, root)
    228 
--> 229         plot.pane = self
    230         backend = plot.renderer.backend
    231         if hasattr(plot.renderer, 'get_plot_state'):

~/PyDev/repos/holoviews/holoviews/plotting/plot.py in pane(self, pane)
    139             for plot in self.subplots.values():
    140                 if plot is not None:
--> 141                     plot.pane = pane
    142                 for cb in getattr(plot, 'callbacks', []):
    143                     cb.comm._on_error = partial(pane._on_error, self.root)

~/PyDev/repos/holoviews/holoviews/plotting/plot.py in pane(self, pane)
    130     def pane(self, pane):
    131         if (config.console_output != 'disable' and
--> 132             self.root.ref['id'] not in state._handles and
    133             isinstance(self.comm, JupyterComm)):
    134             handle = display(display_id=uuid.uuid4().hex)

AttributeError: 'NoneType' object has no attribute 'ref'
```
Here's the environment
```
# Name                    Version                   Build  Channel
_libgcc_mutex             0.1                        main  
asn1crypto                1.3.0                    py37_0  
attrs                     19.3.0                     py_0  
backcall                  0.1.0                    py37_0  
blas                      1.0                         mkl  
bleach                    3.1.0                    py37_0  
bokeh                     1.4.0                    py37_0  
ca-certificates           2020.1.1                      0  
certifi                   2019.11.28               py37_0  
cffi                      1.14.0           py37h2e261b9_0  
chardet                   3.0.4                 py37_1003  
cryptography              2.8              py37h1ba5d50_0  
cycler                    0.10.0                   py37_0  
dbus                      1.13.12              h746ee38_0  
decorator                 4.4.1                      py_0  
defusedxml                0.6.0                      py_0  
entrypoints               0.3                      py37_0  
expat                     2.2.6                he6710b0_0  
fontconfig                2.13.0               h9420a91_0  
freetype                  2.9.1                h8a8886c_1  
glib                      2.63.1               h5a9c865_0  
gmp                       6.1.2                h6c8ec71_1  
gst-plugins-base          1.14.0               hbbd80ab_1  
gstreamer                 1.14.0               hb453b48_1  
holoviews                 1.13.0a22.post72+ge532d49be           dev_0    <develop>
icu                       58.2                 h9c2bf20_1  
idna                      2.8                      py37_0  
importlib_metadata        1.5.0                    py37_0  
intel-openmp              2020.0                      166  
ipykernel                 5.1.4            py37h39e3cac_0  
ipython                   7.12.0           py37h5ca1d4c_0  
ipython_genutils          0.2.0                    py37_0  
ipywidgets                7.5.1                      py_0  
jedi                      0.16.0                   py37_0  
jinja2                    2.11.1                     py_0  
jpeg                      9b                   h024ee3a_2  
jsonschema                3.2.0                    py37_0  
jupyter                   1.0.0                    py37_7  
jupyter_client            5.3.4                    py37_0  
jupyter_console           6.1.0                      py_0  
jupyter_core              4.6.1                    py37_0  
kiwisolver                1.1.0            py37he6710b0_0  
ld_impl_linux-64          2.33.1               h53a641e_7  
libedit                   3.1.20181209         hc058e9b_0  
libffi                    3.2.1                hd88cf55_4  
libgcc-ng                 9.1.0                hdf63c60_0  
libgfortran-ng            7.3.0                hdf63c60_0  
libpng                    1.6.37               hbc83047_0  
libsodium                 1.0.16               h1bed415_0  
libstdcxx-ng              9.1.0                hdf63c60_0  
libtiff                   4.1.0                h2733197_0  
libuuid                   1.0.3                h1bed415_2  
libxcb                    1.13                 h1bed415_1  
libxml2                   2.9.9                hea5a465_1  
markdown                  3.1.1                    py37_0  
markupsafe                1.1.1            py37h7b6447c_0  
matplotlib                3.1.3                    py37_0  
matplotlib-base           3.1.3            py37hef1b27d_0  
mistune                   0.8.4            py37h7b6447c_0  
mkl                       2020.0                      166  
mkl-service               2.3.0            py37he904b0f_0  
mkl_fft                   1.0.15           py37ha843d7b_0  
mkl_random                1.1.0            py37hd6b4f25_0  
nbconvert                 5.6.1                    py37_0  
nbformat                  5.0.4                      py_0  
ncurses                   6.2                  he6710b0_0  
notebook                  6.0.3                    py37_0  
numpy                     1.18.1           py37h4f9e942_0  
numpy-base                1.18.1           py37hde5b4d6_1  
olefile                   0.46                     py37_0  
openssl                   1.1.1d               h7b6447c_4  
packaging                 20.1                       py_0  
pandas                    1.0.1            py37h0573a6f_0  
pandoc                    2.2.3.2                       0  
pandocfilters             1.4.2                    py37_1  
panel                     0.8.1a1                    py_0    pyviz/label/dev
param                     1.10.0a2                   py_0    pyviz/label/dev
parso                     0.6.1                      py_0  
pcre                      8.43                 he6710b0_0  
pexpect                   4.8.0                    py37_0  
pickleshare               0.7.5                    py37_0  
pillow                    7.0.0            py37hb39fc2d_0  
pip                       20.0.2                   py37_1  
prometheus_client         0.7.1                      py_0  
prompt_toolkit            3.0.3                      py_0  
ptyprocess                0.6.0                    py37_0  
pycparser                 2.19                     py37_0  
pyct                      0.4.6                      py_0    pyviz/label/dev
pyct-core                 0.4.6                      py_0    pyviz/label/dev
pygments                  2.5.2                      py_0  
pyopenssl                 19.1.0                   py37_0  
pyparsing                 2.4.6                      py_0  
pyqt                      5.9.2            py37h05f1152_2  
pyrsistent                0.15.7           py37h7b6447c_0  
pysocks                   1.7.1                    py37_0  
python                    3.7.6                h0371630_2  
python-dateutil           2.8.1                      py_0  
pytz                      2019.3                     py_0  
pyviz_comms               0.7.3                      py_0    pyviz/label/dev
pyyaml                    5.3              py37h7b6447c_0  
pyzmq                     18.1.1           py37he6710b0_0  
qt                        5.9.7                h5867ecd_1  
qtconsole                 4.6.0                      py_1  
readline                  7.0                  h7b6447c_5  
requests                  2.22.0                   py37_1  
send2trash                1.5.0                    py37_0  
setuptools                45.2.0                   py37_0  
sip                       4.19.8           py37hf484d3e_0  
six                       1.14.0                   py37_0  
sqlite                    3.31.1               h7b6447c_0  
terminado                 0.8.3                    py37_0  
testpath                  0.4.4                      py_0  
tk                        8.6.8                hbc83047_0  
tornado                   6.0.3            py37h7b6447c_3  
tqdm                      4.42.1                     py_0  
traitlets                 4.3.3                    py37_0  
urllib3                   1.25.8                   py37_0  
wcwidth                   0.1.8                      py_0  
webencodings              0.5.1                    py37_1  
wheel                     0.34.2                   py37_0  
widgetsnbextension        3.5.1                    py37_0  
xz                        5.2.4                h14c3975_4  
yaml                      0.1.7                had09818_2  
zeromq                    4.3.1                he6710b0_3  
zipp                      2.2.0                      py_0  
zlib                      1.2.11               h7b6447c_3  
zstd                      1.3.7                h0b5b093_0  
```
Any thoughts?

philippjfr:
Are you sure you have pulled the latest? Locally there is an explicit `and self.root` check in there which prevents your error.

```
    def pane(self, pane):
        if (config.console_output != 'disable' and self.root and
            self.root.ref['id'] not in state._handles and
```

jonmmease:
Yeah, I had fallen behind. It's working now. Thanks!

holoviews/core/util.py
```diff
@@ -1106,6 +1106,20 @@ def unique_iterator(seq):
             yield item
 
 
+def lzip(*args):
+    """
+    zip function which returns a list.
```
jbednar:
```suggestion
    zip function that returns a list.
```

holoviews/streams.py
```diff
@@ -993,6 +1009,24 @@ class BoundsXY(LinkedStream):
         Bounds defined as (left, bottom, right, top) tuple.""")
 
 
+class SelectionXY(BoundsXY):
+    """
+    A stream representing the selection along the x-axis and y-axis.
+    Unlike a BoundsXY stream this stream return range or categorical
```
jbednar:
```suggestion
    Unlike a BoundsXY stream, this stream returns range or categorical
```

holoviews/selection.py
```diff
@@ -169,15 +170,62 @@ def _expr_stream_updated(self, hvobj, selection_expr, bbox):
...
+    multiple objects.
+    """
+
+    cross_filter_mode = param.Selector(
+        ['overwrite', 'intersect'], default='intersect', doc="""
+        Determines how to combine selections across different
+        elements.""")
+
+    index_cols = param.List(default=None, doc="""
+        If provided selection switches to index mode where all queries
```
jbednar:
```suggestion
        If provided, selection switches to index mode where all queries
```

jonmmease:
@philippjfr I was trying out my original PR notebook from https://anaconda.org/jonmmease/cross_filter_link_selections_pr/notebook and selections are not being displayed for the first example.  I think it has something to do with the Dataset have `redim.label` and `redim.units` applied because when I remove these everything works again.

philippjfr:
Okay, thanks, I'll look into it.

jonmmease:
There's something a little odd going on in the datashader example from that notebook. It looks like the colormap of the left-most subplot is getting applied to the unselected layer in the middle subplot.

In the gif below I set `unselected_alpha` to 0.5 to make it more clear.

```python
lnk_sel = link_selections.instance(unselected_alpha=0.5)
lnk_sel(
    datashade(points1_big, cmap='green')
    + datashade(points2_big)
    + hist_big)
```
holoviews_selection

philippjfr:
Very odd indeed, not quite sure how the operation kwargs can leak into another plot like that but I'll dig into it.

jonmmease:
This is great @philippjfr, I've tested and read through as much as I could and other then the two issues I noted above, I'm really happy with it.  Thanks again for taking over the PR and pushing it so much further!

jbednar:
Yay!

PR #4120, opened 2019-12-06
Add spatialpandas interface and polygon rasterization

philippjfr:
- [x] Add spatialpandas interface
- [x] Add canonical conversion for all geometry data formats
- [x] Add support for rasterizing geometries 
- [x] Add docs
- [x] Add tests

### DIFF
--- a/examples/user_guide/Geometry_Data.ipynb
+++ b/examples/user_guide/Geometry_Data.ipynb
@@ -6,7 +6,7 @@
    "source": [
     "In addition to the two main types of data, namely tabular/columnar and gridded data HoloViews also provide extensible interfaces to represent path geometry data. Specifically it has three main element types used to representing different types of geometries. In this section we will cover the HoloViews data model for representing different kinds of geometries.\n",
     "\n",
-    "There are many different ways of representing path geometries but HoloViews' data model is oriented on GEOS geometry definitions and allows faithfully round-tripping data between its element types and GEOS geometry definitions such as ``LinearString``, ``Polygon``, ``MultiLineString`` and ``MultiPolygon`` geometries (even if this is not implemented in HoloViews itself). Since HoloViews interfaces are extensible many different formats for representing geometries could be supported (see [GeoViews](http://geoviews.org/user_guide/Geometries.html) for other representations) but here we will cover the native formats used by HoloViews to represent this data."
+    "There are many different ways of representing path geometries but HoloViews' data model is oriented on GEOS geometry definitions and allows faithfully round-tripping data between its element types and GEOS geometry definitions such as ``LinearString``, ``Polygon``, ``MultiLineString`` and ``MultiPolygon`` geometries (even if this is not implemented in HoloViews itself). HoloViews defines a dictionary based format for the geometries but also supports [spatialpandas](https://github.com/holoviz/spatialpandas), which is a highly optimized implementation similar to [geopandas](https://github.com/geopandas/geopandas/) but without the heavy geo-dependencies such as shapely and fiona. [GeoViews](https://geoviews.org/user_guide/Geometries.html) supports both geopandas and raw shapely geometries directly."
    ]
   },
   {
@@ -19,7 +19,7 @@
     "import holoviews as hv\n",
     "from holoviews import opts\n",
     "\n",
-    "hv.extension('bokeh')\n",
+    "hv.extension('matplotlib')\n",
     "\n",
     "opts.defaults(opts.Path(padding=0.1), opts.Polygons(padding=0.1))"
    ]
@@ -41,14 +41,16 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "hv.Path({'x': [1, 2, 3, 4, 5], 'y': [0, 0, 1, 1, 2]})"
+    "hv.Path({'x': [1, 2, 3, 4, 5], 'y': [0, 0, 1, 1, 2]}, ['x', 'y'])"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Here the dictionary of x- and y-coordinates could also be an NumPy array with two columns or a dataframe with 'x' and 'y' columns. To draw multiple paths the data-structures can be wrapped in a list. Additionally, it is also possible to associate a value with each path by declaring it as a value dimension:"
+    "Here the dictionary of x- and y-coordinates could also be an NumPy array with two columns or a dataframe with 'x' and 'y' columns.\n",
+    "\n",
+    "To draw multiple paths the data-structures can be wrapped in a list. Additionally, it is also possible to associate a value with each path by declaring it as a value dimension:"
    ]
   },
   {
@@ -57,8 +59,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "hv.Path([{'x': [1, 2, 3, 4, 5], 'y': [0, 0, 1, 1, 2], 'value': 0},\n",
-    "         {'x': [5, 4, 3, 2, 1], 'y': [2, 2, 1, 1, 0], 'value': 1}], vdims='value').opts(color='value')"
+    "p = hv.Path([{'x': [1, 2, 3, 4, 5], 'y': [0, 0, 1, 1, 2], 'value': 0},\n",
+    "             {'x': [5, 4, 3, 2, 1], 'y': [2, 2, 1, 1, 0], 'value': 1}], vdims='value').opts(color='value')\n",
+    "p"
    ]
   },
   {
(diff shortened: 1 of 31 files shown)


philippjfr:
Handling geometries consistently turned out to be a pretty hard problem taking me most of last week and most of today. Some of the features here:

* `.iloc` will now work consistently throughout (meaning that Selection1D will work nicely)
* `.groupby` is now handled consistently throughout
* `.sort` is now implemented allowing us to sort by any dimension other than a geometry dimension
* `.values` can now be used consistently to access both the geometry and value arrays (previously you had to split to access them)
* Polygon and Ring geometries now always return closed rings from `.values`
* All geometry interfaces can now be converted between one another meaning that they will even roundtrip nicely
* Behavior for the native dictionary format is now well defined for different geometry types
* Shared testing infrastructure for all geometry interfaces including MultiInterface (and all its subtypes), SpatialPandasInterface, GeoPandasInterface (in geoviews)

philippjfr:
Some of the Path plotting code looks pretty terrible now but it should all become much cleaner once I pick up #3986 again.

philippjfr:
This is ready to review.

philippjfr:
Thanks for the initial review @jlstevens, I've now made sure that only actual geometry interfaces are supported by Path types and reverted the API change on `.split`.

examples/user_guide/Geometry_Data.ipynb
```diff
@@ -19,7 +19,7 @@
     "import holoviews as hv\n",
     "from holoviews import opts\n",
     "\n",
-    "hv.extension('bokeh')\n",
+    "hv.extension('matplotlib')\n",
```
jlstevens: Why the change in the loaded plotting extension?

philippjfr: Accidental.

philippjfr:
```suggestion
    "hv.extension('bokeh')\n",
```

examples/user_guide/Geometry_Data.ipynb
```diff
@@ -237,7 +241,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "poly.split(datatype='array')"
+    "poly.split(datatype='columns')"
```
jlstevens: Does the markdown above now need to be updated? Currently it is:

> Using the ``datatype`` argument the data may instead be returned in the desired format, e.g. a list of arrays:

philippjfr: Ah, that was on my todo list but seems I forgot.

philippjfr: I generally wouldn't recommend using array here since you can't represent the additional holes  that way. Hence why I switched.

holoviews/element/comparison.py
```diff
@@ -487,8 +487,6 @@ def compare_paths(cls, el1, el2, msg='Path'):
         paths2 = el2.split()
         if len(paths1) != len(paths2):
             raise cls.failureException("%s objects do not have a matching number of paths." % msg)
-        for p1, p2 in zip(paths1, paths2):
-            cls.compare_dataset(p1, p2, '%s data' % msg)
```
jlstevens: Should there be an equivalent check or is it ok to remove this comparison?

holoviews/element/path.py
```diff
@@ -80,20 +79,7 @@ def __init__(self, data, kdims=None, vdims=None, **params):
...
-            (isinstance(data, list) and not len(data) == 0 and all(
-                isinstance(d, tuple) and all(isscalar(v) for v in d)
-                for d in data))):
-            datatype = [dt for dt in datatype if dt != 'multitabular']
-        elif isinstance(data, list) and 'multitabular' not in datatype:
-            datatype = datatype + ['multitabular']
-
-        super(Path, self).__init__(data, kdims=kdims, vdims=vdims,
-                                   datatype=datatype, **params)
+        super(Path, self).__init__(data, kdims=kdims, vdims=vdims, **params)
```
jlstevens: Good to see this constructor simplified!

holoviews/element/comparison.py
philippjfr: It's already comparing the data so this is just doing it a second time.

philippjfr: Really this is more of a test that the interfaces implement split correctly which isn't the job of a comparison.

holoviews/tests/operation/testdatashader.py
```diff
@@ -19,6 +19,14 @@
 except:
     raise SkipTest('Datashader not available')
 
+try:
+    import spatialpandas
+except:
+    spatialpandas = None
+
+sp_skip = skipIf(spatialpandas is None, "SpatialPandas not available")
```
jlstevens: I just about guessed that `sp_skip` meant 'spatial pandas skip' because of the context of this PR but otherwise I think it wouldn't be a bad idea to be explicit and call it `spatialpandas_skip`.

holoviews/tests/operation/testdatashader.py
```diff
@@ -68,7 +78,7 @@ def test_aggregate_points_categorical(self):
     def test_aggregate_points_categorical_zero_range(self):
         points = Points([(0.2, 0.3, 'A'), (0.4, 0.7, 'B'), (0, 0.99, 'C')], vdims='z')
         img = aggregate(points, dynamic=False,  x_range=(0, 0), y_range=(0, 1),
-                        aggregator=ds.count_cat('z'))
+                        aggregator=ds.count_cat('z'), height=2)
```
jlstevens: Why is there a new `height` argument in a test about aggregating points?

holoviews/tests/operation/testdatashader.py
```diff
@@ -149,7 +159,7 @@ def test_aggregate_ndoverlay_count_cat_datetimes_microsecond_timebase(self):
 
     def test_aggregate_dt_xaxis_constant_yaxis(self):
         df = pd.DataFrame({'y': np.ones(100)}, index=pd.date_range('1980-01-01', periods=100, freq='1T'))
-        img = rasterize(Curve(df), dynamic=False)
+        img = rasterize(Curve(df), dynamic=False, width=3)
```
jlstevens: Similar question to the one above about `height` being introduced.

examples/user_guide/Geometry_Data.ipynb
jlstevens: I'm fine with the switch as long as the explanation matches.

holoviews/tests/operation/testdatashader.py
philippjfr: Something was very wrong with the comparisons and I can't quite explain it. I updated comparisons to avoid flattening the array, so it is now comparing 2D arrays, which provides a much better error message. After doing that the test (correctly) started complaining that one array was (0, 400) and the other (0, 3), obviously when you flatten the arrays both are shape `(0,)` so it didn't error.

holoviews/element/comparison.py
jlstevens: I suppose so though I don't see much harm in this extra check (unless it is really slow or something)

philippjfr: Sure, I'll re-enable.

holoviews/tests/operation/testdatashader.py
philippjfr: I'll update it.

holoviews/tests/operation/testdatashader.py
jlstevens: Ah. Glad you caught that then!

holoviews/tests/plotting/matplotlib/testutils.py
```diff
@@ -30,7 +30,7 @@ def test_polygon_to_path_patches(self):
             (1.5, 2), (2, 3), (1.6, 1.6), (1.5, 2),
             (2.1, 4.5), (2.5, 5), (2.3, 3.5), (2.1, 4.5)]))
         self.assertEqual(paths[0][0].get_path().codes, np.array([1, 2, 2, 79, 1, 2, 2, 79, 1, 2, 2, 79], dtype='uint8'))
-        self.assertEqual(paths[0][1].get_path().vertices, np.array([(6, 7), (7, 5), (3, 2), (6, 7)]))
+        self.assertEqual(paths[0][1].get_path().vertices, np.array([(3, 2), (7, 5), (6, 7),  (3, 2),]))
```
jlstevens: I assume this existing test was updated due to some bug that was fixed? i.e was the original test incorrect?

jlstevens:
@philippjfr Other than some comments above (most of which you have addressed already), I am happy to merge this PR now.

holoviews/tests/plotting/matplotlib/testutils.py
philippjfr: Spatialpandas will reorient vertices to be in the expected order, so to ensure that tests will work across all interfaces I need to make sure that the test data is oriented correctly. Otherwise spatialpandas will have a different orientation to other interfaces on comparison.

jlstevens: Ok, that makes sense. Thanks for the explanation.

jlstevens:
Looks good. Thanks!

PR #4121, opened 2019-12-06
Fix incorrect parameter warnings when using datashade

jlstevens:
I believe this might be the simplest approach to address the superfluous warnings when using `datashade` as described in #4116.


jlstevens:
Note that the diff is misleadingly large - other than swapping two class definitions around in the file, the only change is this line:

```
(type(None), shade) # To handles parameters of datashade 
```

philippjfr:
Thanks, kinda hacky but should work fine.

PR #4126, opened 2019-12-12
Fix XArrayInterface.aggregate handling of **kwargs

nluetts:
The `aggregate` method of a `holoviews.Dataset` object that was prepared from a `xarray.DataArray` does not pass extra keyword arguments to the aggregation function.

Adding `**kwargs` to the underlying call to xarray's `reduce()` function in the `XArrayInterface.aggregate()` method fixed the issue for me.


philippjfr:
Thanks for the fix @nluetts

PR #4131, opened 2019-12-18
Update 04-Style_Mapping.ipynb

Jacob-Barhak:
Changed the first example to show the case where kdims are not x,y . This may help beginner readers to use data sets that are not defined by x,y names.

### DIFF
--- a/examples/user_guide/04-Style_Mapping.ipynb
+++ b/examples/user_guide/04-Style_Mapping.ipynb
@@ -26,7 +26,7 @@
    "source": [
     "One of the major benefits of HoloViews is the fact that Elements are simple, declarative wrappers around your data, with clearly defined semantics describing how the dimensions of the data map to the screen. Usually the key dimensions (kdims) and value dimensions map to coordinates of the plot axes and/or the colormapped intensity. However there are a huge number of ways to augment the visual representation of an element by mapping dimensions to visual attributes. In this section we will explore how we can declare such mappings including complex transforms specified by so called ``dim`` objects.\n",
     "\n",
-    "To illustrate this point let us create a set of three points with x/y-coordinates and alpha, color, marker and size values and then map each of those value dimensions to a visual attribute by name:"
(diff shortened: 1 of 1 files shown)


examples/user_guide/04-Style_Mapping.ipynb
```diff
@@ -680,9 +682,22 @@
   }
  ],
  "metadata": {
+  "kernelspec": {
+   "display_name": "Python 2",
```
philippjfr: You really should upgrade to Python3 😄 

Also ordinarily we clear this metadata because it can confuse other peoples notebook sessions. I'll push to your PR to remove it.

Jacob-Barhak:
You lost me Philipp,

You asked to clear the notebook before and I did so before submitting the
PR. I cleared the kernel of the notebook before submitting it and I figured
using Python 2 makes things more compatible.

There are some issues with Python 3 when I compile some of my
presentations, so I still prefer python 2 environment - unfortunately, it
still produces better results with Holoviz in many cases. I can report a
problem with URL resolution, yet those are outside the scope of Holoviz. So
I have good reasons to use Python 2 - it is still better for stability in
some cases.

What do you want to remove now?

I can easily compile it in Python 3 environment if this is what you ask
for. Just please be clear on actions to be taken.

             Jacob

philippjfr:
>You asked to clear the notebook before and I did so before submitting the PR

I'm talking about the notebook metadata, but please don't worry about I'll push to your PR.

>There are some issues with Python 3 when I compile some of my
presentations, so I still prefer python 2 environment - unfortunately, it
still produces better results with Holoviz in many cases.

None of the developers of HoloViz tools use python2 at this point and the next releases of HoloViews and Panel will be the last to support Python2. Bokeh master has already dropped support for it. So I'd strong advise to upgrade asap.

Jacob-Barhak:
Thanks Philipp,

There are good reasons to upgrade to Python 3 - I see them yet I do have
some python 2 environments and some python 3 environments.

However, developers pushing to move to 3 neglect legacy code, which comes
with a price.

In the case of Holoviz, it makes sense since its newer technology, as long
as the transition is gradual enough it makes sense.

I will use python 3 for future PRs.

Thanks.

           Jacob

PR #4137, opened 2019-12-19
Dataset propagation improvements

jonmmease:
https://github.com/holoviz/holoviews/pull/4137/commits/c5adc794e3ecc04ebc6721b0765ef4520938a59e: Cloning a `Dataset` object  with `ds.clone(dataset=None)` should not propagate the `dataset` property to the cloned  object.  That was the behavior initially, but was changed (I think unintentionally) in 83da274a080e99553f531eb591c3983746c12acf. @philippjfr, does this patch look alright?

Added a simple test to lock in the behavior for the future.

https://github.com/holoviz/holoviews/pull/4137/commits/c65395b6fbcc8865f096798e1b5902bc77d99f6f:  Add `Operation._propagate_dataset` class flag to control whether to automatically propagate dataset/pipeline through an operation.

For context, these are both small fixes/improvements needed to get the glaciers demo working with `link_selections`.

### DIFF
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -130,8 +130,8 @@ class Operation(param.ParameterizedFunction):
         for hook in self._postprocess_hooks:
             ret = hook(self, ret, **kwargs)
 
-        if (self._propagate_dataset and
-                isinstance(ret, Dataset) and isinstance(element, Dataset)):
+        if (self._propagate_dataset and isinstance(ret, Dataset)
+                and isinstance(element, Dataset)):
             ret._dataset = element.dataset.clone()
             ret._pipeline = element_pipeline.instance(
                 operations=element_pipeline.operations + [
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -70,6 +70,10 @@ class Operation(param.ParameterizedFunction):
     _postprocess_hooks = []
     _allow_extra_keywords=False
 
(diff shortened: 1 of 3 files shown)


holoviews/core/operation.py
```diff
@@ -126,7 +130,8 @@ def _apply(self, element, key=None):
         for hook in self._postprocess_hooks:
             ret = hook(self, ret, **kwargs)
 
-        if isinstance(ret, Dataset) and isinstance(element, Dataset):
+        if (self._propagate_dataset and
+                isinstance(ret, Dataset) and isinstance(element, Dataset)):
```
philippjfr: Odd indentation.

philippjfr:
>Cloning a Dataset object with ds.clone(dataset=None) should not propagate the dataset property to the cloned object.

Sorry about that, I rushed those changes a little bit since I was giving a talk which included linked selections.

One very minor comment otherwise this looks fine.

holoviews/core/operation.py
jonmmease: Updated to follow PEP8 rules, indentation still the same I think, but looks less awkward now.  Another alternative would be:
```python
if (
    self._propagate_dataset and
    isinstance(ret, Dataset) and
    isinstance(element, Dataset)
):
    ...
```

jonmmease:
No worries! Thanks for taking a look. I updated the `if` indentation, let me know if you'd rather have another alternative.

PR #4146, opened 2020-01-02
Add Rectangles element

philippjfr:
A Rect element is a vectorized geometry element which is somewhat similar to Segments in that it has four kdims, `x0`, `y0`, `x1`, `y1`, representing the corners of the rectangle. It's a versatile element which can be used to draw a variety of different plots, e.g.:

Screen Shot 2020-01-03 at 4 49 15 PM

It is also better suited towards usage with the `BoxEdit` tool, so this PR refactors the corresponding stream to accept and return `Boxes` and adds a `BoxAnnotator` which allows annotating a plot with rectangles/boxes.

- [x] Update BoxEdit stream
- [x] Add bokeh, matplotlib and plotly implementations
- [x] Add Boxes reference examples
- [x] Add BoxAnnotator and added example to user guide
- [ ] Add tests


jbednar:
It's great to see this.  I'd favor Boxes or Rectangles, as Rects doesn't sound like public API to me.

holoviews/element/geom.py
```diff
@@ -104,3 +104,18 @@ class Segments(Geometry):
                        bounds=(4, 4), constant=True, doc="""
         Segments represent lines given by x- and y-
         coordinates in 2D space.""")
+
+
+class Rects(Geometry):
+    """
+    Rects represent a collection of rectangles in 2D space.
```
jbednar:
```suggestion
    Rects represent a collection of axis-aligned rectangles in 2D space.
```

(right?)

philippjfr:
Okay, renamed to Boxes. I also did some refactoring, I was getting sick of duplicating `get_extents` methods all over the place so I've added `holoviews/plotting/mixins.py` which declares various mixins which can provide code which is shared across multiple plotting backends.

jlstevens:
Sorry to propose another rename, but I think `Rectangles` sounds more like a geometry (which reflects where this element lives)...

jlstevens:
Discussing with @philippjfr, this PR can be part of a larger plan where we offer `Boxes` and `Ellipses` (vectorized) in annotations, start deprecating `Box` and `Ellipse` and leave `Bounds` (unvectorized) as an infrequently used annotation.

Any thoughts @jbednar?

jbednar:
I'm happy with Rectangles.

holoviews/plotting/mixins.py
```diff
@@ -0,0 +1,103 @@
+from __future__ import absolute_import, division, unicode_literals
```
jlstevens: It would have been nice to add `mixins` in a separate PR to make it easier to see what is related to the new element. Too late now though!

jbednar:
Do the non-vectorized versions have some advantage I'm not aware of?  If not, I'd prefer to hide all the non-vectorized versions (leaving them supported, but not presented in the docs or examples) and only focus on the vectorized ones.

jlstevens:
> If not, I'd prefer to hide all the non-vectorized versions (leaving them supported, but not presented in the docs or examples) and only focus on the vectorized ones.

Agreed. That can be part of the process of slowly deprecating the non-vectorized versions (first by removing reference to them in the docs, then adding warnings etc)

jlstevens:
@philippjfr The `Rectangles` stuff looks good to me. 

Ironically, I wish the Mixins weren't mixed in with this PR as I don't know how I feel about them yet!

philippjfr:
I've opened an issue [here](https://github.com/holoviz/holoviews/issues/4152) to factor out the mixin classes I introduced here into a function based approach which is probably cleaner on balance (but also considerably more effort).

jlstevens:
Travis is now green. Merging!

Issue #4144, opened 2020-01-01
Typo in Annotators.ipynb


kcpevey:
Super excited to see Annotators making their way into Holoviews!   
  
I was trying out the latest dev release 1.13.0a16 and I noticed a typo in the last cell of the example notebook `Annotators.ipynb`. To be clear, the works as expected (creates a poly annotator), but the labels are misleading (its labeled as path and should be poly)
  
The code is:  
```
poly_annotator = annotate.instance()
path_layout = path_annotator(hv.Polygons([]), annotations=['Label'])

annotate.compose(hv.element.tiles.Wikipedia(), point_layout, path_layout)
```
It should be:
```
poly_annotator = annotate.instance()
poly_layout = poly_annotator(hv.Polygons([]), annotations=['Label'])

annotate.compose(hv.element.tiles.Wikipedia(), point_layout, poly_layout)
```

kcpevey:
I can open a PR if that would be helpful.

jbednar:
Yes please!

kcpevey:
@jbednar As you know, I'm great at breaking things. I found bugs with poly_annotators. Is that worth reporting or is this still a WIP?

kcpevey:
@jbednar or @philippjfr Can you give me access to create a branch and open a PR?

philippjfr:
Done.

PR #4147, opened 2020-01-02
typo fix, path->poly

kcpevey:
minor typo fix to the last cell in the notebook. The poly_annotator was labeled as path annotator.


kcpevey:
Resolves https://github.com/holoviz/holoviews/issues/4144

philippjfr:
Thanks Kim! Btw you can still use the old Panel based API if you need to but for integration with HoloViews we added an API that's more in line with the way HoloViews works.

Issue #4107, opened 2019-11-27
Option to make all fontsizes in your plot with the same scale larger (like in seaborn)


SandervandenOord:
It would be great in Holoviews if you would have a nice, easy and fast way to make your plot more readable without having to specify every fontsize for every part of the plot separately.

Seaborn has the option to scale all your fontsizes (such as title, xlabel, ylabel, xticks etc) with the same factor, like this:

```python 
import seaborn as sns
sns.set(font_scale=2.0)
```
This will not set all fontsizes to the same size, but rather multiply every separate fontsize with the same factor.

Currently you have to do in Holoviews:
```python
my_plot.opts(fontsize={'title': 20, 'yticks': 14, 'xticks': 14, 'xlabel': 16})
````
This is great if you want to define your plot precisely, but not if you want to rapidly change your plot.

(solution could be something like ```my_plot.opts(fontscale=2.0)``` or ```hv.output(fontscale=200)```, but these are just suggestions.)

ahuang11:
Turns out it's harder to implement than expected since holoviews doesn't have the default fontsizes available in `self` (https://github.com/holoviz/holoviews/blob/master/holoviews/plotting/plot.py#L512 simply returns an empty dict if None) and I think to get the defaults, you need to either hardcode the defaults or render the figure and painstakingly retrieve the attributes (p.title.text_fontsize, p.xlabel.axis_label_text_font_size, etc)

One workaround for bokeh is doing `fontsize={'ticks': '200%', 'labels': '200%'}`

philippjfr:
I think we'll probably have to do this sadly, each backend will have to implement a method that returns the defaults.

jbednar:
Having done so, wouldn't we then be able to implement a font_scale property that would multiply all those values by a factor?  I can't quite see how to integrate that with the options system, though.

philippjfr:
>Having done so, wouldn't we then be able to implement a font_scale property that would multiply all those values by a factor, for the current backend (loop through the defaults, and multiply each by the factor)?

I think that's exactly what's being proposed here. I think the idea would be that it would live at the hv.output level just like the size scaling option, so you could do it globally.

jbednar:
The hv.output level sounds good.  Your previous wording sounded like you were giving up on having a global scaling option in favor of only a dictionary of defaults, but it now it sounds like we're all on the same page; the dictionary would be a way to implement the global scaling. Sounds good!

ahuang11:
[...]

PR #4160, opened 2020-01-06
Implement fontscaling

philippjfr:
Fixes https://github.com/holoviz/holoviews/issues/4107

Here's 1.4x fontscaling for bokeh and matplotlib:

Screen Shot 2020-01-06 at 2 25 56 PM

Screen Shot 2020-01-06 at 2 32 05 PM

### DIFF
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -35,7 +35,7 @@ from .util import (get_dynamic_mode, initialize_unbounded, dim_axis_label,
                    attach_streams, traverse_setter, get_nested_streams,
                    compute_overlayable_zorders, get_nested_plot_frame,
                    split_dmap_overlay, get_axis_padding, get_range,
-                   get_minimum_span, get_plot_frame)
+                   get_minimum_span, get_plot_frame, scale_fontsize)
 
 
 class Plot(param.Parameterized):
@@ -514,7 +514,7 @@ class DimensionedPlot(Plot):
 
     def _get_fontsize_defaults(self):
         """
-        Must returns default fontsize for the following objects:
+        Should returns default fontsize for the following keywords:
 
             * ticks
             * minor_ticks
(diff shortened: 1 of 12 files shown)


holoviews/plotting/plot.py
```diff
@@ -508,29 +512,76 @@ def _format_title_components(self, key, dimensions=True, separator='\n'):
         return (self.label, self.group, type(self).__name__, '')
 
 
+    def _get_fontsize_defaults(self):
+        """
+        Must returns default fontsize for the following objects:
```
jbednar:
```suggestion
        Must return default fontsize for the following objects:
```

philippjfr:
>Could this at some point be part of an even higher-level property that acts like fig_dpi , scaling both fonts and all line, dot, and other glyph sizes to allow increasing the resolution while keeping the plot otherwise the same?

Yes, I like the idea, seaborn has something similar, it provides plotting contexts with various scaling factors, which scale font sizes, linewidths, mark sizes and tick sizes:

```
scaling = dict(paper=.8, notebook=1, talk=1.5, poster=2)
```

jlstevens:
Looks good! I like the concept and I think this will be useful in the general case.

The only thing I am not entirely sure about is whether you should scale when an explicit override is specified. I think of scaling as a quick and easy way to get font sizes roughly where you want them without specifying absolute values. If absolute values are set, I feel those should be respected without applying font scaling. @jbednar @philippjfr Thoughts?

philippjfr:
Just to record my thinking, I can see both arguments. The way I see it fontscaling is an option you might apply when exporting to different formats, e.g. for a talk or poster, and so it might make sense to scale the overrides too in that case, but I can definitely see that you might expect your override to be respected too. In the case of seaborn the fontscale option is global and modifies the defaults only, so I lean towards changing it to match that behavior but I'll let @jbednar chime in.

jbednar:
Same here -- I can't decide which is more important: being able to specify an override that is always true regardless of overall size (e.g. to have a watermark or annotation that is always the same size in pixels regardless of scaling, like the map tile attributions in Bokeh), or to be able to confidently rescale the entire thing in one call, overrides and all.  I think I lean towards the latter (fontscale controls overrides as well) as being the least surprising, but would be happy with either behavior.

jlstevens:
@jbednar As non of us seems to have a strong opinion, it sounds like we are leaving this up to you,  @philippjfr.

philippjfr:
Okay, after thinking about it more I think scaling everything consistently is the least surprising behavior as well so I'll leave things as they are.

SandervandenOord:
Just did some checks on plots I use.
```my_plot.opts(fontscale=1.5)``` works nicely for my plots :)

Issue #4166, opened 2020-01-07
Confusing behavior for integer Annotator initial values


jbednar:
Using the latest HoloViews master (hopefully soon to be released, the first example in Annotators.ipynb uses an Element with integer coordinates: `points = hv.Points([(0, 0), (1, 1), (2, 2)])`.  The behavior for that example is quite confusing: the points can be dragged to an arbitrary location in the x,y plane, but the locations remain integers:

![image](https://user-images.githubusercontent.com/1695496/71935311-16fb0500-316c-11ea-82d7-23f076aa0333.png)

If the type is truly meant to stay integral, I would expect that the location of the point should "snap" to the integer coordinates after dragging, to make the display consistent with the Python data structure.  In any case, this is a very confusing example, and the behavior should either be acknowledged or updated not to use integer coordinates for the initial locations.

![image](https://user-images.githubusercontent.com/1695496/71935105-9c31ea00-316b-11ea-9b77-9aa3de9a9616.png)

BTW, if updating that example, it would be nice not to show NaN, and it would also be nice if hover would show the label values so that people could verify that the right label is associated with the right point.

philippjfr:
>If the type is truly meant to stay integral, I would expect that the location of the point should "snap" to the integer coordinates after dragging

It's problematic because it's only displayed as an integer because the table uses type inference to instantiate an IntFormatter on initialization. Internally the data is actually stored as float it just gets rendered as an int. I think changing the example is the only reasonable approach apart from introducing a ton of special handling on the JS end that actually looks at the table formatters to perform casting.

>it would be nice not to show NaN

What empty cells get displayed as also depends on the formatter, if the formatter is numeric (which is required to enter numeric values) then it will always render empty cells as NaNs, so I really have no control here.

>it would also be nice if hover would show the label values so that people could verify that the right label is associated with the right point.

That definitely seems like a good idea.

jbednar:
Ok, changing the example to avoid this confusion seems like a reasonable approach.

> What empty cells get displayed as also depends on the formatter, if the formatter is numeric (which is required to enter numeric values) then it will always render empty cells as NaNs, so I really have no control here.

Hmm; can a default value be supplied for that column?  

Separately, seems like empty values should display as empty space, which is what Pandas, Excel, and just about every other table that I know of does...

jbednar:
Hmm; empty values _do_ display as empty space for values after the table is created, just not for points that existed from the start:

[...]

PR #4167, opened 2020-01-08
Expanded and clarified hv annotator docs

jbednar:
Some changes worth noting/explaining:
- I found it very confusing to introduce the annotator without .instance() first, then have to say it's unusable. Given that doing so is a very strange thing, I opted to use .instance right away, even if it's mysterious.
- To emphasize that annotate is the underlying call and annotator is the thing you'll actually use, I removed the import of annotate and systematically referred to it hv.annotate instead. I think that helps keep the two things straight; one is an object you can work with in your namespace (annotator); the one provided by the HoloViews library is just for getting something you actually then use, not something to call or work with directly. That's the hope, anyway!
- Changed integer coordinates to floats where it was necessary to avoid issue #4166.
- Added Scatter to the list of supported elements (but there's no Scatter example; should say in words how to make one).
- Commented-out the last cell, which is very confusing on a statically rendered notebook, as the output no longer matches the cell code when reading through it.  Should now be ok.

Ready to merge, but there are things left to do (presumably in a later PR, as this one should be merged to make the docs be usable now):
- [x] Should explain better what to do for types not supported yet (cast to supported type, write new hv code, add Bokeh tools, ?)
- [x] Annotating paths/polygons needs much better docs; it's really hard to figure out how to use it right now. I didn't try to update that section, but not because it doesn't need it!


philippjfr:
All sounds sensible.

>I found it very confusing to introduce the annotator without .instance() first, then have to say it's unusable. Given that doing so is a very strange thing, I opted to use .instance right away, even if it's mysterious.

That was done in response to a review comment on the original PR, I do agree.

>Annotating paths/polygons needs much better docs; it's really hard to figure out how to use it right now. I didn't try to update that section, but not because it doesn't need it!

I didn't want to repeat the explanations from the PolyDraw/PolyEdit stream reference docs but maybe I should.

philippjfr:
I've pushed some additional fixes/changes to this PR:

- Renamed BoxAnnotator->RectangleAnnotator
- Selecting a Path/Polygon in the first table now updates the vertex table
- Exposed the `empty_value` as a parameter on Annotators
- Expanded docs on using the PolyDraw/PolyEdit tools in the Annotator docs

philippjfr:
Lastly I added a `CurveEdit` stream and `CurveAnnotator`.

PR #4174, opened 2020-01-09
Simplified packaging

philippjfr:
* Get's rid of the hacky package bundling and uses the MANIFEST in the way we use across projects.
* Uses setuptools `find_packages` which reduces the risk of forgetting to add a newly added subpackage (which has happened a few times)
 * Adds pyproject.toml to ensure that `pip install -e .` works consistently (and should allow GitHub to extract the necessary information from the setup.py)
* Gets rid of param embed_version, we don't use that in any other package and no one has ever complained

I'm also hoping that this will finally allow GitHub to extract the information it needs from the setup.py to add the `Used By` button at the top of the repo.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -2,15 +2,12 @@ include README.rst
 include LICENSE.txt
 include CHANGELOG.md
 include holoviews/.version
-
-recursive-include tests *
-
-recursive-exclude * __pycache__
-recursive-exclude * *.py[co]
-
-recursive-include doc Makefile make.bat *.conf *.css *.html *.ico *.ipynb *.js *.json *.npy *.png *.py *.rst *.svg
-prune doc/builder
-prune doc/nbpublisher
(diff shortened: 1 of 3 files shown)


jbednar:
Looks great to me!

ceball:
I think this will also fix one of the problems mentioned in https://github.com/pyviz-dev/pyct/issues/22: examples and package/examples will now have the same content in the sdist, so anyone using the sdist plus the `holoviews --install-examples` command will get the bokeh apps.

jlstevens:
Happy to see `embed_version` go and I *think* it is barely used. It shouldn't be needed when pyproject.toml is used (for any recent enough pip) though I'm wondering whether `autover` should be listed there too?

Otherwise, it looks fine to me.

philippjfr:
All our other packages obtain autover from param>=1.7.0 so that should work here too. I'll merge and tag a release to test all this.

Issue #2128, opened 2017-11-16
Relax categorical coordinates for heatmap?


jbednar:
Right now, heatmaps treat their coordinates as categorical, which only works well for small sizes because all labels are always shown for categoricals (since the ordering is not necessarily meaningful, unlike a numerical axis).  For larger heatmaps, having categorical values generates unreadable axes:

![image](https://user-images.githubusercontent.com/1695496/32877039-18e6d17a-ca65-11e7-81d8-d03d053ba5ae.png)

One option is for heatmap to support non-categorical dimensions, so that Bokeh's usual better-performing axis numbering algorithm can be used.  Another option is to add [hover tool support to images in Bokeh](https://github.com/bokeh/bokeh/issues/984), which would let HeatMap be reserved for the small-number categorical case, while Image could be used for cases like the above. In this case, however, one of the axes is truly categorical, while the other one is numerical, so allowing Heatmaps to have non-categorical dimensions  seems like the only general solution.

jbednar:
Note that the WSJ original appears to be using a numeric year axis, as requested here:

![image](https://user-images.githubusercontent.com/1695496/32878887-741617ce-ca6d-11e7-832f-e5be2a328a5c.png)

TomBugnon:
Any progress or workaround on this?

NumesSanguis:
I would call this a bug, because `hv.help(hv.HeatMap)` shows that there is a `xticks` and `yticks`, but even if you write `xticks=None`, it is ignored.

Based on this example (in Jupyter Notebook): # https://stackoverflow.com/questions/38701969/dynamic-spectrum-using-plotly
(vignesh.wav: https://github.com/MTG/essentia-audio/raw/65d8eded3d8145df74625b0c60e3d605c21fa860/recorded/vignesh.wav)

```
import numpy as np
import holoviews as hv
hv.extension('bokeh')

import numpy as np

from scipy.io import wavfile # scipy library to read wav files
AudioName = "vignesh.wav" # Audio File
fs, Audiodata = wavfile.read(AudioName)
Audiodata = Audiodata / (2.**15) # Normalized between [-1,1]

 #Spectrogram
from scipy import signal
N = 512 #Number of point in the fft
w = signal.blackman(N)
freqs, bins, Pxx = signal.spectrogram(Audiodata, fs, window=w, nfft=N)

heatmap = hv.HeatMap((bins, freqs, 10*np.log10(Pxx)))
[...]

PR #4180, opened 2020-01-11
Add support for non-categorical HeatMap

philippjfr:
Fixes https://github.com/holoviz/holoviews/issues/2128

### DIFF
--- a/examples/reference/elements/bokeh/HeatMap.ipynb
+++ b/examples/reference/elements/bokeh/HeatMap.ipynb
@@ -43,8 +43,9 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "data = [(chr(65+i), chr(97+j),  i*j) for i in range(5) for j in range(5) if i!=j]\n",
-    "hv.HeatMap(data).sort()"
+    "data = [(i, chr(97+j),  i*j) for i in range(5) for j in range(5) if i!=j]\n",
+    "hm = hv.HeatMap(data).sort()\n",
+    "hm.opts(xticks=None)"
    ]
   },
   {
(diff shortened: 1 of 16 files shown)


jbednar:
Whoa, you're just going to throw this out here without any fanfare?  This is big news! What exactly has been done here?  Can we now have one numerical and one categorical axis?  Can categorical axis labels now be skipped when xticks is set?  Tell us more!

philippjfr:
>Whoa, you're just going to throw this out here without any fanfare?

Still in progress :)

>Can we now have one numerical and one categorical axis?

Yes.

>Can categorical axis labels now be skipped when xticks is set?

Sadly no, might have a go at allowing that in bokeh though.

philippjfr:
@jbednar I'm never quite decided about integers, sometimes you want to treat them as categorical other times you don't. Any opinion?

philippjfr:
Here's the measles example in matplotlib and bokeh:

Screen Shot 2020-01-11 at 6 45 55 PM
Screen Shot 2020-01-11 at 6 46 26 PM

philippjfr:
Okay I'm happy with this, I've disabled the xmarks and ymarks properties for regular HeatMaps and only support them when radial=True. They were quite confusing, very hairy to implement especially now that axes can be categorical OR numeric, and imo not all that useful. For a radial axis they basically replace grid marks and so are somewhat useful.

jbednar:
That all sounds great.  I'd vote for treating integers as a numeric axis, since there is no ambiguity in labeling when skipping integer labels and numeric integer axes already support such skipping.  That example doesn't show how integer examples are handled right now; if the curve/error-bar plot were removed what would the axis look like?

philippjfr:
>if the curve/error-bar plot were removed what would the axis look like?

The same.

Issue #1090, opened 2017-01-30
feature request: scale extents a bit more than the absolute fit


basnijholt:
Right now:
```
hv.Scatter([[-1,-1], [-1, 1], [0, 1], [1,1]])
```
gives points that you cannot even see:
![download](https://cloud.githubusercontent.com/assets/6897215/22422998/d3aa8376-e6ef-11e6-8af0-5ec3f31eb454.png)

Preferably the plot would take the extent and make it 10% bigger or so, like:
```
hv.Scatter([[-1,-1], [-1, 1], [0, 1], [1,1]])[-1.1:1.1, -1.1:1.1]
```
which does show the points:
![download 1](https://cloud.githubusercontent.com/assets/6897215/22423044/057fdee6-e6f0-11e6-8e50-dae943e109df.png)

philippjfr:
Agreed, this is something we should have added a long time ago. All our plots should automatically pad the axis ranges by a few percent. The question has always been how precisely the padding is calculated. Is it a fixed 10% or is there special handling around zero? Other plotting libraries have had this discussion (see https://github.com/matplotlib/matplotlib/issues/4891) and have generally settled on a fixed 4-5% padding along each axis:

>The easiest way to handle this is to simply ensure that the axis limits are slightly larger than necessary to cover all the data, and then leave it at that. It avoids the really nasty cases, avoids surprises, and if people want something fancier than they almost always have to tweak the limits by hand anyway. As further evidence that this works, this is what R does and it basically just works -- no-one complains.

>Specifically, base R calculates default axis limits by: taking the min and max of the data, and then setting the limits 4% further beyond each of those. ggplot2 uses a similar algorithm, but uses 5% by default. (See here, semantics of expand= is (multiplicative expansion, additive expansion).)

If we agree with both matplotlib and ggplot2 that's what we can use by default as well. It's trivial to do and I strongly believe this belongs in 1.7 as part of changing our global styling defaults (https://github.com/ioam/holoviews/issues/823).

jbednar:
I agree.

jlstevens:
I think this is a good idea and I agree that we should use someone else's magic value for this parameter (and stick to it). My preference for conventions would be to follow that matplotlib does, then bokeh and then ggplot2.

philippjfr:
They all do the same thing 5% padding on each end of an axis.

jbednar:
5% it is, then!

philippjfr:
[...]

PR #4182, opened 2020-01-14
Enable padding on all non-raster elements

philippjfr:
- [x] Fixes https://github.com/holoviz/holoviews/issues/1090
- [x] Set padding default to 0.1, default for Area/Curve/Spread to (0, 0.1) and default for raster-like elements to 0
- [x] Update tests
- [x] Remove explicit padding from examples
- [x] Fixed shared axes in plotly
- [x] Fixed BarPlot get_extents issues
- [x] Add config.no_padding flag


jbednar:
Non raster  elements don't seem to share a base class?

philippjfr:
No, they generally don't share any implementation and/or parameters. Might make sense to introduce one for this PR though.

philippjfr:
Once [this build](https://travis-ci.org/holoviz/holoviews/builds/636660562) is done we'll have to carefully check [the doc build](http://build.holoviews.org/) for any undesirable behavior.

holoviews/plotting/bokeh/chart.py
```diff
@@ -188,6 +188,8 @@ class VectorFieldPlot(ColorbarPlot):
         distance between vectors, this can be disabled with the
         rescale_lengths option.""")
 
+    padding = param.ClassSelector(default=0.05, class_=(int, float, tuple))
```
jbednar: `class_=(int, float, tuple))` is probably common enough in HoloViews code that it should be an actual parameter class, something like NumberOrTuple, which would let the Parameter implement checking that the tuple is numeric.  Here a non-numeric tuple would presumably not be caught until much later in the code, and there's no clear user-visible declaration that it would even need to be numeric.  Not something for this PR, necessarily, but in general it seems like HoloViews probably needs more custom Parameter types to achieve better error handling in cases like this.

philippjfr: >HoloViews probably needs more custom Parameter types to achieve better error handling in cases like this.

Yes, it's a huge shame that HoloViews does not define any custom parameters...pretty embarrassing tbh given that it's almost certainly the largest param based project.

jbednar: Yup.

philippjfr:
>Violin gets cut off vertically for Bokeh, which might be more about an auto-ranging issue than a padding issue (as actual data gets cut off)presumably?)

This is intentional and controlled by the `cut` option.

jbednar:
> This is intentional and controlled by the cut option, without it outliers can make the plot useless.

In that case is there a good reason why the Matplotlib and Bokeh Violin defaults are so different? The Matplotlib version doesn't cut off by default for the reference examples, but the Bokeh and to some extent Plotly versions do.

philippjfr:
>In that case is there a good reason why the Matplotlib and Bokeh Violin defaults are so different?

Yes, the matplotlib version uses the highly limited violin implementation that's shipped with matplotlib. Would be nice to replace that but it's a ton of work.

jbednar:
I'll stop mentioning it now, but I do think the Matplotlib defaults are better for Violin, which I guess means I suggest using a higher cut threshold.

Issue #4193, opened 2020-01-18
Use __signature__ to generate opts object signatures


philippjfr:
Currently we use a docstring trick to generate custom backend dependent signatures for the `hv.opts.[object]` utilities. This approach is no longer supported by IPython in recent versions and is also pretty hacky (and ugly to look at in a docstring since all the keywords appear there). Instead we should use the ``__signature__`` object (see [PEP362](https://www.python.org/dev/peps/pep-0362/)).

![Screen Shot 2020-01-18 at 9 51 33 PM](https://user-images.githubusercontent.com/1550771/72670351-b8236e80-3a3c-11ea-87a9-1077571a698a.png)

philippjfr:
Would like to do this as soon as 1.13.0 is out and we are dropping Python <3.5 support.

philippjfr:
Actually I think I'd like to see this in 1.13.0 already and use it if Python > 3.3, then right after the release we can drop the old codepath.

jlstevens:
I'm looking into this now. The docstring approach was used in param a fair bit (and I agree it is hacky but it was the only thing we could support at the time) so I'll coordinate there as well.

PR #4199, opened 2020-01-21
Tab completion via __signature__

jlstevens:
This PR aims to address #4193. 

Note that the docstring approach is used for 2.x only (i.e 2.7 which will be dropped this year) and ``__signature__`` is used for all 3.x. This means things won't work in 3.0, 3.1, 3.2 as ``__signature__`` was introduced in 3.3. I don't believe this matters as holoviews doesn't support those versions anyway.

Currently work in progress:

- [x] Tab completion for `opts`
- [x] Tab completion for the extension.
- [ ] Address other uses of ``__doc__`` for tab completion.

The bigger issue is that most of our parameterized objects used param to implement the tab completion and updating param in a general way is trickier. Thankfully, I think that tab completing options is the critical thing for holoviews and that needs special handling anyway.

### DIFF
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -493,10 +493,10 @@ class opts(param.ParameterizedFunction):
         if sys.version_info.major == 2:
             builder.__doc__ = '{element}({kws})'.format(element=element, kws=kws)
         else:
-            signature = Signature([Parameter('cls', Parameter.POSITIONAL_ONLY)]
-                                  + ([Parameter('spec', Parameter.KEYWORD_ONLY)]
-                                     + [Parameter(kw, Parameter.KEYWORD_ONLY)
-                                        for kw in sorted_kw_set]))
+            signature = Signature([Parameter('cls', Parameter.POSITIONAL_ONLY),
+                                   Parameter('spec', Parameter.POSITIONAL_OR_KEYWORD)]
+                                  + [Parameter(kw, Parameter.KEYWORD_ONLY)
+                                     for kw in sorted_kw_set])
             builder.__signature__ = signature
         return classmethod(builder)
 
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -2,6 +2,7 @@ import os, sys, inspect, shutil
 
 from collections import defaultdict
 from types import FunctionType
+from inspect import Parameter, Signature
 
 try:
     from pathlib import Path
@@ -484,10 +485,17 @@ class opts(param.ParameterizedFunction):
             return Options(spec, **kws)
 
         filtered_keywords = [k for k in completions if k not in cls._no_completion]
-        kws = ', '.join('{opt}=None'.format(opt=opt) for opt in sorted(filtered_keywords))
-        builder.__doc__ = '{element}({kws})'.format(element=element, kws=kws)
+        sorted_kw_set = sorted(set(filtered_keywords))
+        kws = ', '.join('{opt}=None'.format(opt=opt) for opt in sorted_kw_set)
+        if sys.version_info.major == 2:
+            builder.__doc__ = '{element}({kws})'.format(element=element, kws=kws)
+        else:
+            signature = Signature([Parameter('cls', Parameter.POSITIONAL_ONLY)]
+                                  + ([Parameter('spec', Parameter.KEYWORD_ONLY)]
+                                     + [Parameter(kw, Parameter.KEYWORD_ONLY)
+                                        for kw in sorted_kw_set]))
+            builder.__signature__ = signature
         return classmethod(builder)
-
 
     @classmethod
     def _element_keywords(cls, backend, elements=None):
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -89,6 +89,9 @@ class opts(param.ParameterizedFunction):
        to strict (default), any invalid keywords are simply skipped. If
        strict, invalid keywords prevent the options being applied.""")
 
+    def __init__(self, *args, **kwargs): # Needed for opts specific __signature__
+        super(opts, self).__init__(*args, **kwargs)
+
     def __call__(self, *args, **params):
         if not params and not args:
             return Options()
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -2,7 +2,7 @@ import os, sys, inspect, shutil
 
 from collections import defaultdict
 from types import FunctionType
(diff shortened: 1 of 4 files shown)


jlstevens:
I've decided to delete the ``__doc__`` wrangling on the notebook extension: it is just the regular parameter completion that should be supported by all parameterized objects at the param level. It was supposed to also complete potential plotting backends but that never worked as the backends are known until the extension is actually executed.

jlstevens:
Other places that need to be addressed in holoviews:

- [ ] Output utility
- [ ] ``opt.__call__`` (i.e when no specific element is specified)

But there is also this in the notebook archiver which I am not sure is worth handling:

```python
        keywords = ['%s=%s' % (k, v.__class__.__name__)
                    for k, v in self.param.objects().items()]
        self.auto.__func__.__doc__ = 'auto(enabled=Boolean, %s)' % ', '.join(keywords)
```

It just tab completed the `auto` keyword so it doesn't seem that useful to me...

holoviews/util/__init__.py
```diff
@@ -484,11 +485,18 @@ def builder(cls, spec=None, **kws):
...
         filtered_keywords = [k for k in completions if k not in cls._no_completion]
-        kws = ', '.join('{opt}=None'.format(opt=opt) for opt in sorted(filtered_keywords))
-        builder.__doc__ = '{element}({kws})'.format(element=element, kws=kws)
+        sorted_kw_set = sorted(set(filtered_keywords))
+        kws = ', '.join('{opt}=None'.format(opt=opt) for opt in sorted_kw_set)
+        if sys.version_info.major == 2:
+            builder.__doc__ = '{element}({kws})'.format(element=element, kws=kws)
+        else:
+            signature = Signature([Parameter('cls', Parameter.POSITIONAL_ONLY)]
+                                  + ([Parameter('spec', Parameter.KEYWORD_ONLY)]
```
philippjfr: Sorry if this review is premature but shouldn't it be `Parameter('spec', Parameter.POSITIONAL_OR_KEYWORD)`. Also not that it matters much but why separate lists for `cls` and `spec`?

jlstevens: Both good suggestions that I'll implement in the cleanup pass once I finish getting the tab completion working elsewhere.

jlstevens: Addressed in 91b6359

jlstevens:
@philippjfr Ready for review. Unit test are passing except for a clone issue in `holoviews.tests.test_annotators.TestPointAnnotator`.

holoviews/util/__init__.py
```diff
@@ -525,9 +536,17 @@ def _update_backend(cls, backend):
...
-        old_doc = cls.__original_docstring__.replace('params(strict=Boolean, name=String)','')
-        cls.__doc__ = '\n    opts({kws})'.format(kws=kws) + old_doc
+        sorted_kw_set = sorted(set(filtered_keywords))
+        if sys.version_info.major == 2:
+            kws = ', '.join('{opt}=None'.format(opt=opt) for opt in sorted_kw_set)
+            old_doc = cls.__original_docstring__.replace(
+                'params(strict=Boolean, name=String)','')
+            cls.__doc__ = '\n    opts({kws})'.format(kws=kws) + old_doc
+        else:
+            signature = Signature([Parameter('args', Parameter.VAR_POSITIONAL)]
```
philippjfr: No `self` needed in this case?

holoviews/util/settings.py
```diff
@@ -234,8 +235,17 @@ def _generate_docstring(cls):
...
+            doc_signature = '\noutput(%s)\n' % ', '.join('%s=None' % kw for kw in keywords)
+            return '\n'.join([doc_signature] + intro + descriptions)
+        else:
+            return '\n'.join(intro + descriptions)
+
+    @classmethod
+    def _generate_signature(cls):
+        keywords = ['backend', 'fig', 'holomap', 'widgets', 'fps', 'max_frames',
+                    'size', 'dpi', 'filename', 'info', 'css', 'widget_location']
+        return  Signature([Parameter(kw, Parameter.KEYWORD_ONLY) for kw in keywords])
```
philippjfr:
```suggestion
        return Signature([Parameter(kw, Parameter.KEYWORD_ONLY) for kw in keywords])
```

holoviews/util/__init__.py
```diff
@@ -484,11 +488,18 @@ def builder(cls, spec=None, **kws):
             return Options(spec, **kws)
 
         filtered_keywords = [k for k in completions if k not in cls._no_completion]
-        kws = ', '.join('{opt}=None'.format(opt=opt) for opt in sorted(filtered_keywords))
-        builder.__doc__ = '{element}({kws})'.format(element=element, kws=kws)
+        sorted_kw_set = sorted(set(filtered_keywords))
+        kws = ', '.join('{opt}=None'.format(opt=opt) for opt in sorted_kw_set)
```
philippjfr: Doesn't matter much but why not move this into the py2 block?

philippjfr:
Mostly minor comments. I'll look into the unit test failure in a bit.

holoviews/util/__init__.py
jlstevens: Fixed in dc3c44401

holoviews/util/__init__.py
jlstevens: Good point! I went back and checked and it seems the tab completion stopped working due to the lack of the positional arguments *other* than `cls`/`self`. In other words, if there are positional arguments and the signature doesn't specify any, the tab completion stops working. I've now simplified things in dc3c44401.

philippjfr: So I'm confused, `cls` and `self` are or aren't needed?

philippjfr: I get it now, nevermind.

holoviews/util/__init__.py
```diff
@@ -2,6 +2,7 @@
 
 from collections import defaultdict
 from types import FunctionType
+from inspect import Parameter, Signature
```
philippjfr: You'll have to inline the import for Python 2 tests to pass:

```
  File "/home/travis/build/holoviz/holoviews/holoviews/__init__.py", line 11, in <module>
    from . import util                                       # noqa (API import)
  File "/home/travis/build/holoviz/holoviews/holoviews/util/__init__.py", line 5, in <module>
    from inspect import Parameter, Signature
ImportError: cannot import name Parameter
```

jlstevens: Thanks. Fixed in 0a808f9.

PR #4201, opened 2020-01-23
Warning when using methods outside param namespace

jlstevens:
Currently toggling on warnings but once they are all fixed I think it might be worth raising errors instead. My plan is to use travis to find out where things need updating.

### DIFF
--- a/holoviews/element/comparison.py
+++ b/holoviews/element/comparison.py
@@ -312,8 +312,8 @@ class Comparison(ComparisonInterface):
                                        % (set(dim1_params.keys()), set(dim2_params.keys())))
 
         for k in dim1_params.keys():
-            if (dim1.param[k].__class__.__name__ == 'Callable'
-                and dim2.param[k].__class__.__name__ == 'Callable'):
+            if (dim1.param.objects('existing')[k].__class__.__name__ == 'Callable'
+                and dim2.param.objects('existing')[k].__class__.__name__ == 'Callable'):
                 continue
             try:  # This is needed as two lists are not compared by contents using ==
                 cls.assertEqual(dim1_params[k], dim2_params[k], msg=None)
--- a/holoviews/element/comparison.py
+++ b/holoviews/element/comparison.py
@@ -312,8 +312,8 @@ class Comparison(ComparisonInterface):
                                        % (set(dim1_params.keys()), set(dim2_params.keys())))
 
         for k in dim1_params.keys():
-            if (dim1.params(k).__class__.__name__ == 'Callable'
-                and dim2.params(k).__class__.__name__ == 'Callable'):
+            if (dim1.param[k].__class__.__name__ == 'Callable'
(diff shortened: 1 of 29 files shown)


jlstevens:
Seems like we hadn't done much of a pass on this update before (as I had thought). There is a lot that needs updating (e.g see the two PRs referenced above) and `PlotSelector` in particular is a pain (possibly fixed by aliasing `.params` to `.param` there). Then there are lots of messages like this:

> .WARNING:param.CurvePlot214092: The Parameterized instance has instance parameters created using new-style param APIs, which are incompatible with .params. Use the new more explicit APIs on the .param accessor to query parameter instances.To query all parameter instances use .param.objects with the option to return either class or instance parameter objects. Alternatively use .param[name] indexing to access a specific parameter object by name.

I'll have to go through the unit test files one by one to weed out old usage...

@philippjfr As this will end up as a big PR that touches a lot of files, I hope there aren't going to be any big merge conflicts!

philippjfr:
>As this will end up as a big PR that touches a lot of files, I hope there aren't going to be any big merge conflicts!

Not worried tbh, if there's merge conflicts so be it.

jlstevens:
@philippjfr I've fixed the datashader unit test as requested. The other unit test failures need fixes elsewhere that are not related to this PR. If you are happy with the changes I would appreciate if you could review/merge.

holoviews/plotting/bokeh/util.py
```diff
@@ -587,7 +587,7 @@ def get_tab_title(key, frame, overlay):
             title = []
             if frame.label:
                 title.append(frame.label)
-                if frame.group != frame.params('group').default:
+                if frame.group != frame.param['group'].default:
```
philippjfr: Annoying I know but this avoids making an instance parameter when it's not needed.

```suggestion
                if frame.group != frame.param.objects('existing)['group'].default:
```

holoviews/element/comparison.py
```diff
@@ -312,8 +312,8 @@ def compare_dimensions(cls, dim1, dim2, msg=None):
                                        % (set(dim1_params.keys()), set(dim2_params.keys())))
 
         for k in dim1_params.keys():
-            if (dim1.params(k).__class__.__name__ == 'Callable'
-                and dim2.params(k).__class__.__name__ == 'Callable'):
+            if (dim1.param[k].__class__.__name__ == 'Callable'
```
philippjfr:
```suggestion
            if (dim1.param.objects('existing')[k].__class__.__name__ == 'Callable'
```

holoviews/element/comparison.py
```diff
@@ -312,8 +312,8 @@ def compare_dimensions(cls, dim1, dim2, msg=None):
...
+                and dim2.param[k].__class__.__name__ == 'Callable'):
```
philippjfr:
```suggestion
                and dim2.param.objects('existing')[k].__class__.__name__ == 'Callable'):
```

PR #4202, opened 2020-01-23
Ensure dynamic parameter values are evaluated before being passed to param

philippjfr:
Fixes issue introduced by https://github.com/holoviz/param/pull/367


philippjfr:
Note needed now that https://github.com/holoviz/param/pull/367 has been [reverted](https://github.com/holoviz/param/pull/374). I'll use this PR to address the issue that the param change was introduced for in the first place.

PR #4278, opened 2020-03-10
Improve TriMesh wiremesh rendering

philippjfr:
Switches to rendering datashaded TriMesh wireframe using Canvas.line(axis=1) and also allows rendering trimeshes where the vertex indices are not straightforward integer indices into the node array.


jbednar:
How does it improve it?

philippjfr:
The PR is not ready....

PR #4286, opened 2020-03-11
Handle Image types in spread operations

philippjfr:
Once datashader supports simple (non-rgb) arrays in spread operations this will ensure they are supported.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1520,20 +1520,31 @@ class SpreadingOperation(LinkableOperation):
         """Apply the spread function using the indicated parameters."""
         raise NotImplementedError
 
-    def _process(self, element, key=None):
-        if not isinstance(element, RGB):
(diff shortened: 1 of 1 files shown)


holoviews/operation/datashader.py
```diff
@@ -1520,20 +1520,31 @@ def _apply_spreading(self, array):
...
+        return tf.Image(self.uint8_to_uint32(rgbarray.astype('uint8')))
+
+    def _process(self, element, key=None):
+        if isinstance(element, RGB):
+            rgb = element.rgb
+            data = self._preprocess_rgb(rgb)
+        elif isinstance(element, Image):
+            data = element.clone(datatype=['xarray']).data[element.vdims[0].name]
+        else:
+            raise ValueError('spreading can only be applied to Image or RGB Elements.')
```
jbednar: What about spreading on categorical rasterized data (which I believe is an NdOverlay of Image)? Is that already handled separately so that the operation is applied per Image?

Issue #1107, opened 2017-02-08
Please make holoviews.operation.ElementOperation compatible with NdOverlays and NdLayouts


drs251:
A starting point might be the `__call__`method in this Gist: [https://gist.github.com/drs251/8218814fa3360b3301abc5df3bdb896f](https://gist.github.com/drs251/8218814fa3360b3301abc5df3bdb896f)

philippjfr:
NdLayouts are now supported, dealing with NdOverlays requires more thoughts and should be resolved for 2.0.

PR #4288, opened 2020-03-12
Add flag to declare Operations that should be applied per element

philippjfr:
This adds a `_per_element` flag to the Operation baseclass and enables that flag for select subclasses. In 2.0 we can consider flipping that switch and applying it by default as suggested by https://github.com/holoviz/holoviews/issues/1107


jbednar:
Looks good. Should _per_element=False be specified for those operations as well, so that they don't change behavior when the default changes?

philippjfr:
Can do that in the follow-on PR, it's more important for downstream packages (e.g. geoviews).

PR #4295, opened 2020-03-17
V1.13 doc fixes

philippjfr:
* Fixes bug in Slope
* Fixes Bokeh compatibility issues with Linked Stream callbacks including undeclared vars and usage of require to import bokeh modules
* Updates all stream reference notebooks

ToDo:

- [x] Fix error in http://dev.holoviews.org/user_guide/Linked_Brushing.html
- [x] Fix bokeh Bars, CurveEdit, PointerXY thumbnails in ref gallery http://dev.holoviews.org/reference/index.html
- [x] Fix plotly BoxWhisker thumbnail in ref gallery: http://dev.holoviews.org/reference/index.html
- [x] Fix matplotlib Bars thumbnail in ref gallery
- [x] Fix error in http://dev.holoviews.org/user_guide/Tabular_Datasets.html
- [x] Fix warning (WARNING:param.DynamicMap04364: Setting non-parameter attribute link_inputs=False using a mechanism intended only for parameters
) in http://dev.holoviews.org/user_guide/Custom_Interactivity.html
- [x] Fix error in http://dev.holoviews.org/user_guide/Large_Data.html


jbednar:
I love how dynamic maps now work reasonably on the website; that's so much clearer!
- Issues in the build (suppressing all subsequent plots, which I thus haven't checked):
   * [ ] image
   * [ ] image
   * [ ] image
   * [ ] image
   * [ ] image
- Output that seems broken or could be improved:
   * [ ] I'm not sure if this is an error or just me being confused, but I don't see why there is a one-item selector here: image
   * [ ] Seems like an error (plot empty and title not matching slider) when dragging slider to the value "3": image
   * [ ] Size of datashaded elements seems to have increased, making it not comparable to the non-datashaded ones any more (and thus harder to line up and compare):
image
   * [ ] Annoying spacing bug:
image
- Wording, clarity, or organization issues:
   * [ ] Text seems out of date; should mention that sliders do work sometimes (just not other dynamicmaps)? image
   * [ ] Are we still agreed on `The default plotting extension for HoloViews until a 2.0 release is Matplotlib when HoloViews will start defaulting to Bokeh`? I might rather remove the default than change the default.
   * [ ] Probably should explain shared_axes and how to link and not link axes on http://dev.holoviews.org/user_guide/Composing_Elements.html , i.e. when people learn to put things in the same layout.
   * [ ] Seems like various newer functionality like .apply (and its submethods) should be introduced in http://dev.holoviews.org/user_guide/Transforming_Elements.html . They seem to be in http://dev.holoviews.org/user_guide/Data_Pipelines.html , but I'm not sure why that's a separate notebook; the connection isn't clear to me.
   * [ ] I don't think link_selections should be buried in http://dev.holoviews.org/user_guide/Linked_Brushing.html ; although it's new, it's not an advanced technique. I think it should be introduced very early in getting_started as a major motivator for what HoloViews can let you do when you've declared the appropriate metadata for your data.
   * [ ] In general, I think that a lot of the power added in this release won't be discovered by people, because it's buried in notebook 24 of 33 of the user guide.  Should we have a section in the getting-started guide that succinctly advertises some of that power in a tangible way, and then let people dive into the user guide once they want more info or to mess with it?  I don't think very many people will discover the most amazing features of holoviews from the current structure.
- [ ] Maybe remove the Gitter link from the top bar?  I think people should mainly be heading to Discourse and Git.

philippjfr:
> Custom Interactivity Text seems out of date; should mention that sliders do work sometimes (just not other dynamicmaps)?

No, linked streams cannot be captured in this way, the message is correct.

philippjfr:
>In general, I think that a lot of the power added in this release won't be discovered by people, because it's buried in notebook 24 of 33 of the user guide. Should we have a section in the getting-started guide that succinctly advertises some of that power in a tangible way, and then let people dive into the user guide once they want more info or to mess with it? I don't think very many people will discover the most amazing features of holoviews from the current structure.

I'd very much like to throw out the entire getting started guide and have a single overview notebook which links to the user guides.

jbednar:
> > Custom Interactivity Text seems out of date; should mention that sliders do work sometimes (just not other dynamicmaps)?
> No, linked streams cannot be captured in this way, the message is correct.

The message may be correct about its contents, but it doesn't sound correct on its own.  It says "When viewed statically, DynamicMaps will only show the first available Element", but that's definitely not true about the holoviz.org site as a whole any more, right? Maybe the message on this particular page should be "When viewed statically, the DynamicMaps in this page will typically show only the first available Element"?

philippjfr:
Merging for now, will pick it up in another PR.

Issue #4302, opened 2020-03-19
Layouts are not correctly saved (includes a suggested fix)


Guillemdb:
#### ALL software version info
Current master branch of the project.

#### Description of expected behavior and the observed behavior

When using the `bokeh` backend the panel object cannot be exported as a `.png` file because the `fmt` parameter is ignored when saving the underlying `panel.layout`.

Also, when using the `matplotlib` backend  `holoviews.save` fails throwing the following error
```python
AttributeError                            Traceback (most recent call last)
<ipython-input-4-6297fc239759> in <module>
----> 1 holoviews.save(summary.current_plot, "test_2_plot", fmt="auto")

~/forks/holoviews/holoviews/util/__init__.py in save(obj, filename, fmt, backend, resources, **kwargs)
    784         if formats[-1] in supported:
    785             filename = '.'.join(formats[:-1])
--> 786     return renderer_obj.save(obj, filename, fmt=fmt, resources=resources)
    787 
    788 

~/forks/holoviews/holoviews/plotting/renderer.py in save(self_or_cls, obj, basename, fmt, key, info, options, resources, **kwargs)
    520 
    521         with StoreOptions.options(obj, options, **kwargs):
--> 522             plot, fmt = self_or_cls._validate(obj, fmt)
    523         if isinstance(plot, Viewable):
    524             from bokeh.resources import CDN, INLINE, Resources

~/forks/holoviews/holoviews/plotting/renderer.py in _validate(self, obj, fmt, **kwargs)
    261         holomap_formats = self.mode_formats['holomap']
    262 
--> 263         holomaps = obj.traverse(lambda x: x, [HoloMap])
    264         dynamic = any(isinstance(m, DynamicMap) for m in holomaps)
    265 

AttributeError: 'NoneType' object has no attribute 'traverse'

```

But I am not sure if this is a bug or a feature.

#### Complete, minimal, self-contained example code that reproduces the issue

You can see how the `fmt` parameter is not processed when saving the layout here:

https://github.com/holoviz/holoviews/blob/69e931a5039fc5cb45d0ae8ae653464770c20d30/holoviews/plotting/renderer.py#L532

Doing something like this would solve the problem.

```python
 if fmt not in ["html", "png"]:
    raise ValueError("Format for exporting panel not supported: %s" % fmt)
basename = basename if fmt != "png" else "%s.png" % basename
plot.layout.save(basename, embed=True, resources=resources)
```
[...]

PR #4304, opened 2020-03-19
Fix hv.save for png images

Guillemdb:
Closes https://github.com/holoviz/holoviews/issues/4302

Calling `holoviews.save` on panels ignores the parameter `fmt` and outputs all the layout objects as `html` files.

This PR updates the name of the file passed to `plot.layout.save` to include the file extension when  the selected format is `png`. This allows panel to also export the panels as `png` images.


Guillemdb:
AFAIK panel only supports `html` and `png` formats, so I added a check for the format to be value. If you want invalid formats to be handled different please tell me and I will update it.

I am not familiar with the tests of the package. But If you point me to the place where you are testing the function I modified I can add a test to assert the `ValueError` is raised. 

I can also change the message of the error if you have a better suggestion.

philippjfr:
To be clear the intention was that the format would be specified as part of the filename and fmt would mostly be used to switch between different widget types, i.e. widgets or scrubber.

philippjfr:
I think `hv.save` should probably check the filename extension and then validate that against the supplied format but that can be a separate PR.

holoviews/plotting/renderer.py
```diff
@@ -529,6 +528,9 @@ def save(self_or_cls, obj, basename, fmt='auto', key={}, info={},
                 resources = CDN
             elif resources.lower() == 'inline':
                 resources = INLINE
+            if fmt not in ["html", "png", "auto"]:
+                raise ValueError("Format for exporting panel not supported: %s" % fmt)
+            basename = basename if fmt == "auto" else "%s.%s" % (basename, fmt)
```
philippjfr: Suspect this should be something like this:

```suggestion
            fmt = 'html' if fmt == 'auto' else fmt
            if basename.split('.')[-1] not in ('html', 'png'):
                basename = "%s.%s" % (basename, fmt)
```

Guillemdb: `basename` is already stripped of the extension before being passed to `Renderer.save `.
Do you want me to add the if clause anyway?

If you want me to make explicit that "auto" corresponds to "html" I could also do something like this:
```python
fmt = 'html' if fmt == 'auto' else fmt
basename = "%s.%s" % (basename, fmt)
```

Guillemdb:
> To be clear the intention was that the format would be specified as part of the filename and fmt would mostly be used to switch between different widget types, i.e. widgets or scrubber.

I have been using this wrong all the time :sweat_smile:, I will modify the code of my libraries to reflect this. Thanks for the tip!

philippjfr:
>I have been using this wrong all the time 😅, I will modify the code of my libraries to reflect this. Thanks for the tip!

That's likely more of a failure of our docs, if you have suggestions on how and where to clarify this I'd be very open to make those changes.

Guillemdb:
> I think `hv.save` should probably check the filename extension and then validate that against the supplied format but that can be a separate PR.

Should I remove the `ValueError`?

If you send me a link with the valid formats (just to be sure I am not missing anything) I can make another PR with that.

philippjfr:
>Should I remove the ValueError?

I think the ValueError is fine tbh, that's all the formats that bokeh can export directly. Anything else is already handled in a different way.

Guillemdb:
> That's likely more of a failure of our docs, if you have suggestions on how and where to clarify this I'd be very open to make those changes.

I will think about it and make an informed suggestion (I hope in the form of a PR)

Issue #4305, opened 2020-03-20
missing modules when install with conda 


gandroz:
When installing holoviews with conda, I noticed missing modules like `annotate` class in `annotators.py` module.

I'm working on Ubuntu 18.04 and did the following steps to install:

```
$> conda create -n dev python=3  # will create a virtual env with python 3.8
$> conda activate dev
$> conda install -c pyviz holoviews bokeh
$> conda install -c conda-forge jupyterlab nodejs numpy
$> jupyter labextension install @pyviz/jupyterlab_pyviz
```

Then, I launched a jupyter lab instance and tried to reproduce the `Annotator.ipynb` notebook

![image](https://user-images.githubusercontent.com/38106881/77126123-65efd100-6a1e-11ea-887f-cea17aecfc23.png)

From the virtual environment, I can confirm that the module is not here:

![image](https://user-images.githubusercontent.com/38106881/77125954-e530d500-6a1d-11ea-850b-95128c3368ae.png)

And here are the first lines of the `__init__.py` file inside the package:

```
from __future__ import print_function, absolute_import
import os, io

import numpy as np # noqa (API import)
import param


__version__ = str(param.version.Version(fpath=__file__, archive_commit="$Format:%h$",
                                        reponame="holoviews"))

from . import util # noqa (API import)
from .core import archive, config                        # noqa (API import)
from .core.boundingregion import BoundingBox             # noqa (API import)
from .core.dimension import OrderedDict, Dimension       # noqa (API import)
from .core.element import Element, Collator              # noqa (API import)
from .core.layout import (Layout, NdLayout, Empty,       # noqa (API import)
                          AdjointLayout)
from .core.ndmapping import NdMapping                    # noqa (API import)
from .core.options import (Options, Store, Cycle,        # noqa (API import)
                           Palette, StoreOptions)
from .core.overlay import Overlay, NdOverlay             # noqa (API import)
from .core.spaces import (HoloMap, Callable, DynamicMap, # noqa (API import)
                          GridSpace, GridMatrix)

from .operation import Operation                         # noqa (API import)
from .element import *                                   # noqa (API import)
from .element import __all__ as elements_list
from .util import (extension, renderer, output, opts,    # noqa (API import)
                   render, save)
from .util.transform import dim                          # noqa (API import)
```

which misses the line
```
from .annotators import annotate                         # noqa (API import)
```

Furthermore, I tried to install with `pip` instead, but with `pip install 'holoviews[all]', I got the error 

```
[...]

PR #4307, opened 2020-03-20
Website: fix canonical link

ceball:
See comments in https://github.com/holoviz/holoviews/issues/4305

Note, I skimmed nbsite + theme code only _really briefly_, so this might not be correct. It might even break something else. But if it's easy to try on a dev site build, might be worth it (unless someone inspects the situation)...


philippjfr:
Thank you!

PR #4320, opened 2020-03-24
Explicitly namespace dim expressions

philippjfr:
In previous PRs we allowed dim expressions to provide access to any method or accessor. This provides zero validation against typos and is also quite brittle. Instead we have decided that we will provide two additional `dim` objects to explicitly use specific namespaces, the default `dim` effectively mirrors the NumPy namespace, while `df_dim` provides access to the DataFrame-like namespaces, and `xr_dim` to provide access to the XArray namespace. Instead of use these directly we will allow users to access the namespace by using `dim(...).df` and `dim(...).xr`. By default these will only work specifically on objects that can be coerced to conform to this namespace, i.e. a `df_dim` will only work with columnar (or geometry) data, an `xr_dim` will work only on gridded data. The `df_dim` can serve this double purpose since coercion can convert any geometry dataset to either spatialpandas or geopandas which do conform to the dataframe API.

The benefits of namespacing the objects are:

- Tab-completion of the API
- Validation of the API (no more typos accidentally making it through)
- Clear semantics
- Ability to use namespace on compatible objects through coercion


jlstevens:
I think this is a very important distinction to make that will help avoid confusion. As I understand it, `dim` will support methods/behaviors regardless of the underlying data format while the other two will indicate how they are specialized. I would just like to propose `pdim` and `xdim` as potential names for the sake of brevity. Your suggestions are more explicit but also a bit more clunky to use...

philippjfr:
Not a fan of `pdim` or `xdim`, I don't think `df_dim` or `xr_dim` is too clunky but I'd be okay ommitting the underscore for `dfdim` and `xrdim`.

jbednar:
I think the problem with pdim and xdim is that they look like vdims and kdims, i.e., seem to be about new types of dimensions an element can have. Omitting the underscore seems fine.

jlstevens:
> I think the problem with pdim and xdim is that they look like vdims and kdims

Good point. Sounds like Philipp's names without the underscore is our preferred choice right now.

jbednar:
How about `dim().pd.` ( or `dim().df.` ) and `dim().xr.`? I think the semantics of that are clearer than `dfdim` and `xrdim`, i.e., it's about accessing the dim() as a Pandas or an Xarray object, or about making a Pandas or an Xarray call on that object.

jlstevens:
I think I briefly suggested something like `dim().df` and `dim().xr` in the discussion yesterday. I'm ok with this (and might slightly prefer it in fact). Up to Philipp!

holoviews/util/transform.py
```diff
@@ -233,15 +240,51 @@ def __init__(self, obj, *args, **kwargs):
...
+            return self.ops[-1]['fn']
+
+    def __call__(self, *args, **kwargs):
+        if (not self.ops or not isinstance(self.ops[-1]['fn'], basestring) or
+            'accessor' not in self.ops[-1]['kwargs']):
+            raise ValueError("Cannot use __call__ method on dim expression "
+                             "which is not an accessor. Ensure that you only "
+                             "call a dim expression, which was created by "
+                             "accessing an attribute that does not exist "
+                             "on an existing dim expression.")
```
philippjfr: Need to update this error.

holoviews/util/transform.py
```diff
@@ -675,3 +759,75 @@ def __repr__(self):
...
+    """
+
+    namespace = 'xarray'
+
+    def __init__(self, obj, *args, **kwargs):
+        try:
+            import xarray as xr
+        except ImportError:
+            raise ImportError("XArray could not be imported, xr_dim "
+                              "requires the pandas namespace to be "
```
philippjfr:
```suggestion
                              "requires the xarray namespace to be "
```

holoviews/util/transform.py
philippjfr:
```suggestion
            fn = self.ops[-1]['fn']
            raise ValueError("Cannot use __call__ method on %r expression. "
                             "__call__ is only supported on attributes accessed via the"
                             "namespace (e.g. dim().df or dim().xr")
```

holoviews/util/transform.py
```diff
@@ -675,3 +759,75 @@ def __repr__(self):
...
+                              "available.")
```
philippjfr:
```suggestion
            raise ImportError("XArray could not be imported, dim().xr "
                              "requires the xarray to be available.")
```

examples/user_guide/11-Transforming_Elements.ipynb
```diff
@@ -134,7 +134,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "quantile_expr = hv.dim('air').quantile(0.95, dim='time')\n",
+    "quantile_expr = hv.xr_dim('air').quantile(0.95, dim='time')\n",
```
jbednar:
```suggestion
    "quantile_expr = hv.dim('air').xr.quantile(0.95, dim='time')\n",
```

examples/user_guide/11-Transforming_Elements.ipynb
```diff
@@ -106,7 +106,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "This allows almost any mathematical transformation to be expressed and applied on a `Dataset` in a deferred way. The `dim` expression supports all the  standard mathematical operators, NumPy array methods but may also be used to apply methods which exist only on the datatype the data as stored as. Let us for example load an XArray Dataset, which has a number of custom methods to perform complex computations on the data, e.g. the quantile method:"
+    "This allows almost any mathematical transformation to be expressed and applied on a `Dataset` in a deferred way. The regular `dim` expression supports all the  standard mathematical operators and NumPy array methods. However if we want to use methods which exist only on specific datatypes we can use the `.df` or `.xr`, which provide the ability to make use of (pandas) dataframe and xarray namespaces respectively. Let us for example load an XArray Dataset, which has a number of custom methods to perform complex computations on the data, e.g. the quantile method:"
```
jbednar:
```suggestion
    "This allows almost any mathematical transformation to be expressed and applied on a `Dataset` in a deferred way. The regular `dim` expression supports all the  standard mathematical operators and NumPy array methods. However if we want to use methods which exist only on specific datatypes we can invoke them using `.df` or `.xr`, which let you make (pandas) dataframe and xarray API (method and accessor) calls respectively. Let us for example load an XArray Dataset, which has a number of custom methods to perform complex computations on the data, e.g. the quantile method:"
```

examples/user_guide/11-Transforming_Elements.ipynb
```diff
@@ -191,7 +191,7 @@
     "\n",
     "q = pn.widgets.FloatSlider(name='quantile')\n",
     "\n",
-    "quantile_expr = hv.dim('air').quantile(q, dim='time')\n",
+    "quantile_expr = hv.xr_dim('air').quantile(q, dim='time')\n",
```
jbednar:
```suggestion
    "quantile_expr = hv.dim('air').xr.quantile(q, dim='time')\n",
```

jlstevens:
Looks good to me. I've directly given @philippjfr a few minor items of feedback to consider but otherwise I am very happy with this change!

jlstevens:
Travis push build passed and pr build only failed due to a transient. Very happy with this PR and it was very worthwhile to get into this release! Merging.

examples/user_guide/11-Transforming_Elements.ipynb
```diff
@@ -65,7 +65,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "expr = np.sin(hv.dim('x')*10+5)\n",
+    "expr = np.sin(hv.dim('x').df*10+5)\n",
```
jbednar: Is this .df intentional?  It does not seem to be needed here...

philippjfr: Damn, I had thought you had caught that in your suggestions so didn't look further.

jbednar: And I thought _you_ were changing that one.  :-) No big deal, just checking...

jbednar: I pushed a fix to master.

PR #4335, opened 2020-03-30
rcparam datapath deprecated in mpl 3.2.1

poplarShift:
See https://github.com/holoviz/geoviews/issues/452

This whole code to handle deprecated rcparams is now being duplicated in renderer.py and plot.py, should probably factor that out somehow?

https://matplotlib.org/3.2.1/api/prev_api_changes/api_changes_3.2.0.html#the-datapath-rcparam


philippjfr:
Much appreciated. I'm not sure what's up with travis and the appveyor tests are flakey as hell so I'm merging.

philippjfr:
>This whole code to handle deprecated rcparams is now being duplicated in renderer.py and plot.py, should probably factor that out somehow

Yes, want to make an issue?

PR #4358, opened 2020-04-03
REL+BUG: Fix panel requirement on Python 3.5

scottclowe:
The current release of Holoviews still claims to support Python 3.5.
However, it can not be installed on Python 3.5 with its current installation requirements. The reason for this is an anomaly in the release history of panel. We work around this problem by specifying a maximum version requirement of panel for holoviews when installing on Python 3.5.
More details of the problem are given below.

The new major release of bokeh, v2.0.0, dropped support for Python 3.5. This new python version requirement was specified in its setup.py requirements.
To support the new bokeh API, panel was changed with what was effectively a new major release in v0.9.0. This version of panel is only for bokeh>=2.0.0, and this requirement was specified in its setup.py requirements. However, panel did not change its Python version requirements to >=3.6 until version 0.9.4.
This means that whenever one tries to install the latest available version of panel on Python 3.5, pip will attempt to install version 0.9.3, and then fail because it is not possible to support the bokeh>=2.0.0 requirement on Python 3.5. Consequently, all packages which require panel and continue to support Python 3.5 need to include an extra specification for the panel version on Python <3.6.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -11,7 +11,9 @@ import pyct.build
 
 setup_args = {}
 install_requires = ['param>=1.9.3,<2.0', 'numpy>=1.0', 'pyviz_comms>=0.7.3',
-                    'panel>=0.7.0', 'pandas']
+                    "panel>=0.7.0,<0.9.0; python_version<'3.6'",
+                    "panel>=0.7.0; python_version>='3.6'",
+                    'pandas']
 
 extras_require = {}
 
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -29,7 +29,7 @@ requirements:
     {% for dep in sdata.get('install_requires',{}) %}
     {% if "python_version<" not in dep %}
     {% set dep_parts = dep.split(';') %}
-    - {{ dep_parts[0] }}
+    - {{ dep_parts[0]|trim }}
     {% endif %}
     {% endfor %}
     {% for dep in sdata['extras_require']['recommended'] %}
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -27,11 +27,11 @@ requirements:
   run:
     - python {{ sdata['python_requires'] }}
     {% for dep in sdata.get('install_requires',{}) %}
-    {% if "panel" not in dep %}
-    - {{ dep }}
+    {% if "python_version<" not in dep %}
+    {% set dep_parts = dep.split(';') %}
+    - {{ dep_parts[0] }}
     {% endif %}
     {% endfor %}
-    - panel >=0.8
     {% for dep in sdata['extras_require']['recommended'] %}
     - {{ dep }}
     {% endfor %}
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -29,9 +29,9 @@ requirements:
     {% for dep in sdata.get('install_requires',{}) %}
     {% if "panel" not in dep %}
     - {{ dep }}
-	{% endif %}
+    {% endif %}
     {% endfor %}
-    - panel >=0.7
+    - panel >=0.8
     {% for dep in sdata['extras_require']['recommended'] %}
     - {{ dep }}
     {% endfor %}
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -22,7 +22,10 @@ requirements:
   build:
     - python {{ sdata['python_requires'] }}
     {% for dep in sdata['extras_require']['build'] %}
-    - {{ dep }}
+    {% if "python_version<" not in dep %}
+    {% set dep_parts = dep.split(';') %}
+    - {{ dep_parts[0]|trim }}
+    {% endif %}
     {% endfor %}
   run:
     - python {{ sdata['python_requires'] }}
@@ -33,7 +36,10 @@ requirements:
     {% endif %}
     {% endfor %}
     {% for dep in sdata['extras_require']['recommended'] %}
-    - {{ dep }}
+    {% if "python_version<" not in dep %}
+    {% set dep_parts = dep.split(';') %}
+    - {{ dep_parts[0]|trim }}
+    {% endif %}
     {% endfor %}
 
 test:
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -27,8 +27,11 @@ requirements:
   run:
     - python {{ sdata['python_requires'] }}
     {% for dep in sdata.get('install_requires',{}) %}
+    {% if "panel" not in dep %}
     - {{ dep }}
+	{% endif %}
     {% endfor %}
+    - panel >=0.7
     {% for dep in sdata['extras_require']['recommended'] %}
     - {{ dep }}
     {% endfor %}


setup.py
```diff
@@ -11,7 +11,9 @@
 
 setup_args = {}
 install_requires = ['param>=1.9.3,<2.0', 'numpy>=1.0', 'pyviz_comms>=0.7.3',
-                    'panel>=0.7.0', 'pandas']
+                    "panel>=0.7.0,<0.9.0; python_version<'3.6.0'",
+                    "panel>=0.7.0; python_version>='3.6.0'",
```
maximlt, line 15: Shouldn't it be?
```suggestion
                    "panel>=0.9.0; python_version>='3.6.0'",
```

scottclowe: No, because holoviews still only requires a minimum panel version of 0.7.0. We're only adding a maximum panel version to the requirements (and it only applies on Python 3.5).

scottclowe:
This bug was seen in a [cron job for fissa's ci tests](https://travis-ci.org/github/rochefort-lab/fissa/jobs/669402304).
I traced it back and was able to fix the problem by specifying a maximum version of python [(commit)](https://github.com/scottclowe/fissa/commit/bf70d2b22c58d5b8cb586b0d4f53e2700b2475e5). Thought it would be best to propagate the fix upstream to holoviews too.

philippjfr:
Thanks Scott, hope you're doing well! The problem with specifying it this way is that I don't know how this will work with our conda builds. The `conda.recipe/meta.yaml` use the `setup.py` as their source and I'm not sure they will support this syntax.

scottclowe:
I'm doing well thanks, Phil. Hope all is good with you!

This is certainly the right way to specify this from the perspective of setup.py and pip.
https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-platform-specific-dependencies
https://www.python.org/dev/peps/pep-0508/#environment-markers

I am not familiar with conda recipes. Presumably that is why the CI builds are failing as well? Is there a workaround for this?

Of course ideally this wouldn't be necessary if pip would automatically detect a valid combination of package versions, but sadly that's not the case. It seems like relying on packages to always declare their requirement versions exactly correctly is the most fragile part of the python packaging infrastructure.

To prevent this problem arising again, it would be better to test on at least the oldest and newest python2 and python3 versions which the package is supposed to support. I notice that the unit test current run for Python 2.7 and 3.6. Could this be changed to Python 2.7, 3.5 and 3.8 instead? Similarly, panel should also test on the oldest and newest versions of Python which it supports (3.6 and 3.8).

philippjfr:
Sorry for having this sit for so long.

>It seems like relying on packages to always declare their requirement versions exactly correctly is the most fragile part of the python packaging infrastructure.

Certainly and as packages are transitioning to a Python3 only codebase this is currently even worse. Part of the problem here I think is that both Bokeh  2.0.x and Panel 0.9.x technically require Python>3.5, so in some sense this is an upstream problem. We are also about to drop support for Python < 3.6 in HoloViews so I don't know how much sense it makes to add Python 3.5 to the CI matrix at this point, although we should maintain some backported fixes for Python 2 branches for a little while. We should definitely add Python 3.8 tests in any case. 

I'll have a go at making your changes here work by adding some additional logic into the conda recipe templating to drop the python version spec after the semi-colon.

scottclowe:
> Sorry for having this sit for so long.

No worries. Packaging problems are the worst! Any mistakes you make are permanent.

> Part of the problem here I think is that both Bokeh 2.0.x and Panel 0.9.x technically require Python>3.5, so in some sense this is an upstream problem.

Indeed! It is panel's fault because panel defacto dropped support for Python 3.5 in 0.9.0, but didn't update its `python_requires` until 0.9.4. If you could retroactively fix the panel's `python_requires` metadata on pypi for 0.9.0-0.9.3, that would be the better solution to the problem. Unfortunately that isn't possible.

To be clear with respect to #4384, bokeh in of itself isn't the problem because its `python_requires` was always accurate. Although the symptom shown when you try to pip install holoviews on Python 3.5 is `ERROR: Could not find a version that satisfies the requirement bokeh>=2.0.0.`, you don't need to specify a bokeh requirement to resolve the issue. That's actually just a cascade from a problem with panel packaging (panel being a primary dependency, and bokeh a secondary dependency). Hence specifying `panel<0.9.0`, as per the PR, is sufficient and I think a more appropriate solution.

> We are also about to drop support for Python < 3.6 in HoloViews so I don't know how much sense it makes to add Python 3.5 to the CI matrix at this point

If it's a lot of work to change the CI matrix that's fair. Just make sure to manually check releases can install on Python 3.5 (from when this fix is in) until you do drop support!

> I'll have a go at making your changes here work by adding some additional logic into the conda recipe templating to drop the python version spec after the semi-colon.

Sounds good.

philippjfr:
> It is panel's fault because panel defacto dropped support for Python 3.5 in 0.9.0, but didn't update its python_requires until 0.9.4.

😞 

>If you could retroactively fix the panel's python_requires metadata on pypi for 0.9.0-0.9.3, that would be the better solution to the problem.

Yeah I wish, conda at least allows retroactive patches to be applied to builds.

philippjfr:
Looks like our CI setup isn't too happy about the syntax. I'll have to try to figure something out.

scottclowe:
Where is the command `develop_install` defined? https://github.com/holoviz/holoviews/blob/master/.travis.yml#L68

I don't think it's doing what you think its doing at the install step, because the travis build is not using conda.recipe/meta.yaml at the step it is breaking on.

- I [changed](https://github.com/scottclowe/holoviews/commit/2bd91d7684307176964221b886870c374e099dc5) the `{{ deps }}` variable output to say `foo` instead, and the [build still broke the same way](https://travis-ci.org/github/scottclowe/holoviews/jobs/677163222)
- I [wiped](https://github.com/scottclowe/holoviews/commit/30880e7951fe279ecf4b96faabe3d7a903dac126) the contents of conda.recipe/meta.yaml and it [still broke the same way](https://travis-ci.org/github/scottclowe/holoviews/jobs/677165495)

philippjfr:
>I don't think it's doing what you think its doing at the install step, because the travis build is not using conda.recipe/meta.yaml at the step it is breaking on.

It doesn't use the meta.yaml in the install step no, but it is using it when building conda packages.

>Where is the command develop_install defined? 

It's defined in [pyctdev](https://github.com/pyviz-dev/pyctdev/) so yes fixing the meta.yaml isn't sufficient. pyctdev also uses the specified dependencies to install from conda and since they are not valid conda version specs they break.

scottclowe:
> > I don't think it's doing what you think its doing at the install step, because the travis build is not using conda.recipe/meta.yaml at the step it is breaking on.
> 
> It doesn't use the meta.yaml in the install step no, but it is using it when building conda packages.
>
> > Where is the command develop_install defined?
> 
> It's defined in [pyctdev](https://github.com/pyviz-dev/pyctdev/) so yes fixing the meta.yaml isn't sufficient. pyctdev also uses the specified dependencies to install from conda and since they are not valid conda version specs they break.

Okay, I take it back! You definitely know what it is doing. I have no idea though (can't see where develop_install is defined, even now I'm looking in pyctdev!). Sounds like handling conda builds from setup.py requirements containing environment markers needs to be addressed in pyctdev in a similar way to meta.yaml, so I'll just leave it with you. :)

I added some more commits which generalise the Jinja code to work with arbitrary environment markers.

philippjfr:
> I have no idea though (can't see where develop_install is defined, even now I'm looking in pyctdev!)

Yeah, pyctdev is not the most intuitive piece of software, mostly because the processes it tries to automate are a real pain. Let's turn this issue into an Edinburgh reunion and ask @ceball if he has any ideas on expressing these pip version specs without breaking the conda based workflow.

scottclowe:
@philippjfr You can't update versions on PyPI, but you can yank them. Maybe you should just yank panel 0.9.0, 0.9.1, 0.9.2 from PyPI instead, so pip doesn't try to install them on python 3.5. As there are no breaking changes or compatibility changes in 0.9.3, nobody should need 0.9.0--0.9.2 (and if they do, you can still install a yanked version from pypi if you specify an exact requirement with ==).

philippjfr:
Best solution I've heard so far, I know that was only just added, is yanking already implemented?

ceball:
Hi, sorry, I haven't been following hv issues and just noticed this by chance.

>> You can't update versions on PyPI, but you can yank them
> I know that was only just added, is yanking already implemented?

pip has known about yanked packages since sometime in the middle of last year, I think (and yes, as a maintainer, you can yank from pypi already).

ceball:
I'll respond to some points while I'm here, but note there is nothing concrete/productive below...

> It seems like relying on packages to always declare their requirement versions exactly correctly is the most fragile part of the python packaging infrastructure.

I learned a joke at work: "you must be new here" :)

I don't think you can rely on there being valid metadata (particularly dependencies) on pypi in general (i.e. not only for holoviz); I have found it's safer to manage that side of things myself instead of trusting package developers.

When pip's new dependency resolver is turned on, I guess everything everywhere will completely break at first, and then things might get better from then on, because package authors will have to start getting it right...

> If it's a lot of work to change the CI matrix that's fair. Just make sure to manually check releases can install on Python 3.5 (from when this fix is in) until you do drop support!

In an ideal world, we would do that (I think we should do it!)...but focus of limited resources is on the current leading edge, and also on using conda rather than pip, so if it matters to you, you probably have to do that yourself.

> any ideas on expressing these pip version specs without breaking the conda based workflow.

The current approach was copied from bokeh (https://github.com/bokeh/bokeh/blob/master/setup.py#L86). The next version of pyctdev (currently unreleased for 2 years or whatever) takes a different approach. And the plan is to use something like dephell.

One of the main points of pyctdev was to make package authors treat dependency expression as something as important as the code itself (by making it easy, and having tests of those dependencies happen for generated packages in various different environments). However, it completely failed to do that, and I am living at the opposite extreme now of completely ignoring anything developers ever say about what their code depends on, because it's all lies :)

scottclowe:
> Best solution I've heard so far, I know that was only just added, is yanking already implemented?

Yes, I checked and I can yank versions of a package I'm maintainer on. (I didn't yank any, but the button to do it exists.)

Issue #4359, opened 2020-04-03
link_selections to support lasso


afonit:
Link selections are amazing especially when combined with datashader.

My do a lot of geographic datashader plots.  The areas we want to select don't usually fit within the Box Select.

This Feature Request is to add in the capability for **datashader link_selections to support Lasso**

For example, on this page: http://holoviews.org/user_guide/Linked_Brushing.html

There is this example:
![image](https://user-images.githubusercontent.com/1798497/78384304-f21a0080-75a7-11ea-9036-f0c2575a172c.png)

In addition to the box select it would be great to have lasso select as some projects may require more delicate/drawn shapes to be used rather than the square box select.

jbednar:
Yes, lasso select would be fabulous, but it's quite tricky to support it. So that you see how it works, when you use a box select tool for a linked selection, an expression is generated like this:

`((dim('x') >= 0.1) & (dim('x') <= 5.5)) & (dim('y') >= 1.6)) & (dim('y') <= 9.8))`

This expression matches the box that you drew with the box-select tool, and it is then used to do a query on the original dataset, creating a new dataset consisting only of those points that fall into the specified range of x and y. Basically, all datapoints with x or y not in the indicated ranges are filtered out, and then the resulting smaller set of data points is datashaded.

To do the same thing for a lasso select tool would require constructing an expression for the lasso shape that could be used for such a query, which is much more difficult than for a square axis-aligned rectangle as in box select. What we would do is to first calculate the bounding box of the lassoed region, filter on that to reject most of the points, and then iterate through the remaining data points to test whether they fit into a polygonal approximation to the lasso, or maybe a rasterized (pixelated) version of the lasso, further rejecting those datapoints that are within the bounding box but not within the lassoed area.  We can certainly do that, but it's very tricky, and not something we're currently planning to attempt. 

For our own purposes, what we are doing instead is a poor approximation to rasterizing the lasso, which is to enable the "union" mode for selections, then draw multiple rectangular regions that together add up to the non-rectangular region we want to select:

![image](https://user-images.githubusercontent.com/1695496/78388762-2ccf6900-75a7-11ea-8a9b-23ce3885ee83.png)

philippjfr:
To make this happen we need to handle it for both columnar and gridded data, in theory we have everything to make that happen:

* `Gridded`: In theory this could be as simple as `rasterize(hv.Polygons([lasso_polygon]), target=gridded_dataset)`
[...]

PR #4362, opened 2020-04-04
Lasso linked select

philippjfr:
Implements #4359 

- [x] Support tabular data, i.e. Points, Scatter (requires spatialpandas or shapely)
- [x] Support gridded/rectilinear data, i.e. Image/QuadMesh (requires datashader)
- [x] Support curvilinear data, i.e. QuadMesh (requires spatialpandas or shapely)
- [x] Support Rectangles and Segments (requires shapely)
- [x] Support Path, Contours, Polygons (requires shapely)
- [x] Fix linked selections on complex dynamicmaps
- [x] Fix region indicators for dynamic plots
- [x] Update docs
- [x] Fix tests
- [x] Add new tests

lasso_example

gridded_lasso

qmesh_lasso

box_lasso

poly_lasso


holoviews/element/selection.py
```diff
@@ -36,6 +36,47 @@ def _get_selection_expr_for_stream_value(self, **kwargs):
...
+            raise ImportError("Lasso selection on gridded data requires "
+                              "datashader to be available.")
+        xs, ys = xvals[0], yvals[:, 0]
+        target = Image((xs, ys, np.empty(ys.shape+xs.shape)))
+        poly = Polygons([geometry])
+        mask = rasterize(poly, target=target, dynamic=False, aggregator='any')
+        return mask.dimension_values(2, flat=False)
+    else:
+        mask = spatial_select_columnar(xvals.flatten(), yvals.flatten(), geometry)
+        return mask.reshape(xvals.shape)
```
jbednar, line 56: From this implementation it looks like lasso selection on curvilinear quadmeshes is going to be slow? If so the docs should warn that. I'm curious why it can't use datashader for both rectilinear and curvilinear grids, since datashader supports both.

holoviews/element/selection.py
```diff
@@ -36,6 +36,47 @@ def _get_selection_expr_for_stream_value(self, **kwargs):
...
+def spatial_select_gridded(xvals, yvals, geometry):
+    rectilinear = (np.diff(xvals, axis=0) == 0).all()
+    if rectilinear:
+        from .raster import Image
+        from .path import Polygons
+        try:
+            from ..operation.datashader import rasterize
+        except ImportError:
+            raise ImportError("Lasso selection on gridded data requires "
+                              "datashader to be available.")
```
jbednar, line 48: Nice use of Datashader for this purpose! Can you not fall through to the spatial_select_columnar case when datashader is not available?

holoviews/element/selection.py
philippjfr: We actually apply the rasterization on the lasso polygon so what I was worried about was applying the resulting mask to the quadmesh but actually there's no issue with that either.  So yes we can optimize this.

holoviews/element/selection.py
philippjfr: Sure, can do, although for large rasters this will be slow.

jbednar: Right.  I'm not sure if it makes sense to warn in such a case; on the one hand without a warning people might not realize that there's an easy way to make it go much faster, but on the other hand, if people genuinely can't install datashader or don't want to or have a small dataset, having such a warning will be annoying.

holoviews/element/selection.py
philippjfr: Actually no that isn't valid, datashader can only generate a curvilinear grid from the lasso polygon which won't correctly mask the curvilinear grid.

jbednar: You mean, datashader can only generate a _rectilinear_ (and indeed equally spaced, and not just recilinear, i.e. raster) grid?  Good point; it only supports raster output grids.

philippjfr: Yes, that's right fixed.

philippjfr:
Took me basically all weekend but I also used this PR to fix issue I had encountered when applying this to various hvPlot based DynamicMaps. The issue was mainly related to the fact that it was recording and replaying both dynamic operations by traversing the DynamicMap graph and the pipeline which is accumulated as operations are applied. This could cause all kinds of weirdness, therefore we now use the `pipeline` as the single source of truth for all transforms that are applied.

This should not cause any actual issues since the pipeline is a far better and easier to manipulate record of all the transforms that were applied to an object.

holoviews/element/selection.py
philippjfr: Tbh, spatialpandas/shapely are more difficult to install and this isn't worth the extra complexity and possible user confusion.

philippjfr:
Merging since there are people waiting on a dev build which includes this work.

holoviews/util/__init__.py
```diff
@@ -870,6 +874,14 @@ class Dynamic(param.ParameterizedFunction):
          corresponding visualization should update this stream with
          range changes originating from the newly generated axes.""")
 
+    link_dataset = param.Boolean(default=True, doc="""
+         Determines whether the output of the operation should inherit
+         the .dataset property of the input to the operation. Helpful
+         for tracking data providence for user supplied functions,
+         which do not make use of the clone method. Should be disabled
+         for operations where the output is not derived from the input
+         and instead depends on some external state.""")
```
jlstevens, line 883: > Should be disabled for operations where the output is not derived from the input and instead depends on some external state.

Are there any common examples that could be listed here?

jlstevens:
I know the PR is merged now but I had a look anyway. The functionality is very nice and I only have one comment regarding documentation of the `link_dataset` parameter on `Dynamic`.

PR #4364, opened 2020-04-07
Only mask value dimensions

philippjfr:
This ensures that linked selection masking is only applied to value dimensions. Key dimensions will often not support nan masking so this ensures weird type conversions do not happen and in the cuDF case ensures that there are no type errors.

### DIFF
--- a/holoviews/core/data/dictionary.py
+++ b/holoviews/core/data/dictionary.py
@@ -230,7 +230,7 @@ class DictInterface(Interface):
         for vd in dataset.vdims:
             new_array = np.copy(dataset.data[vd.name])
             new_array[mask] = mask_value
-            masked[k] = new_array
+            masked[vd.name] = new_array
         return masked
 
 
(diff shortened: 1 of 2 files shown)


jbednar:
But what about when someone selects in a vdim on one plot that's a kdim on another, or vice versa? Will the selection not be respected?

philippjfr:
It has to make a copy to mask anything.

holoviews/core/data/dictionary.py
```diff
@@ -226,9 +226,9 @@ def concat(cls, datasets, dimensions, vdims):
...
     @classmethod
     def mask(cls, dataset, mask, mask_value=np.nan):
-        masked = OrderedDict()
-        for k, v in dataset.data.items():
-            new_array = np.copy(dataset.data[k])
+        masked = OrderedDict(dataset.data)
+        for vd in dataset.vdims:
+            new_array = np.copy(dataset.data[vd.name])
             new_array[mask] = mask_value
             masked[k] = new_array
```
philippjfr, line 233:
```suggestion
            masked[vd.name] = new_array
```

PR #4383, opened 2020-04-16
Handle log colormapper zero lower bound

### LINKED ISSUE #4361: cmap arguments not accepted as tuples
Originally posted this in [discourse](https://discourse.holoviz.org/t/cmap-arguments-not-accepted-as-tuples/417), @philippjfr said to open it as an issue.

Briefly: holoviews plots will fail if `cmap` is given a tuple argument, but will be fine if converted to a list. (This behavior is particularly not ideal when many `bokeh.palettes` options are tuples.) Failure comes from `holoviews.plotting.utils.process_cmap()`.

Suggest accepting tuples as arguments to `cmap`.

### LINKED ISSUE #4376: Handle log colormapping change in bokeh 2.0
Example of the issue as reported by K.-Michael Aye on gitter:

image

Bokeh 2.x no longer lets the log colormapper handle zeros. I understand the goal of this change was to address this issue https://github.com/bokeh/bokeh/issues/8061 (also see the discussion here https://github.com/bokeh/bokeh/issues/8724). This problem can be more easily avoided when using floats as you can use NaN to avoid the log zero issues.

Unfortunately, this fix will cause a lot of confusion as illustrated by the screenshot above. After discussing this with @jbednar, we listed the following ways we could address this:

1. Document this issue and update how to use `clim` appropriately, checking and updating all our notebooks and handling all the confusion caused to users who miss the updated docs.

2. Maintain backwards compatibility somehow in bokeh 2.0, stating that our `logz` semantics match Bokeh > 2.0 and offer a new option for the better behavior (the old behavior resulted in incorrect colorbars). I'm not sure how this could be achieved technically if the old behavior is not available in bokeh 2.x anymore...

3. Somehow support a masking approach at the element level (i.e `Image` at least, and that could be generalized to other elements?). Seems like a whole lot of work...

4. Address this for the most common use case which will be for any datashading operation that returns zero integers (e.g `count`). To do this you have to avoid generating integers in the first place as NaN isn't a valid integer: the idea would be to cast integer arrays to float and replace all zeros with NaNs. 

If it isn't already obvious, I am in favor of option 4 (and maybe option 2, if it is possible) as it would fix the problem in the majority of cases without having to update all the notebooks and inform all affected users. Although I don't think this is ideal (I would prefer to keep count dtypes as integer) this transformation doesn't violate datashader semantics where NaN and zero are equivalent. Maybe we could have an option for the operations (default off) to skip this transformation for people who want to preserve integer types?

Unfortunately, I don't see any easy solutions and nothing that doesn't involve documenting this change in behavior in Bokeh 2.0 somewhere. What are your thoughts @philippjfr ?

philippjfr:
Ensures compatibility with changes to log colormapping and change to tuple palettes in bokeh 2.0.x 

Fixes #4361 
Fixes #4376

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -1265,6 +1265,12 @@ def is_number(obj):
     else: return False
 
 
+def is_int(obj, int_like=False):
+    real_int = isinstance(obj, int) or getattr(getattr(obj, 'dtype'), 'kind') in 'ui'
+    if real_int or (int_like and hasattr(obj, 'is_integer') and obj.is_integer()):
+        return True
+    return False
+
 class ProgressIndicator(param.Parameterized):
     """
     Baseclass for any ProgressIndicator that indicates progress
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1615,7 +1615,7 @@ class ColorbarPlot(ElementPlot):
         An explicit override of the color bar label, if set takes precedence
         over the title key in colorbar_opts.""")
 
-    clim = param.NumericTuple(default=(np.nan, np.nan), length=2, doc="""
+    clim = param.Tuple(default=(np.nan, np.nan), length=2, doc="""
        User-specified colorbar axis range limits for the plot, as a tuple (low,high).
        If specified, takes precedence over data and dimension ranges.""")
 
@@ -1723,8 +1723,16 @@ class ColorbarPlot(ElementPlot):
         ncolors = None if factors is None else len(factors)
         if eldim:
             # check if there's an actual value (not np.nan)
-            if dim_name in ranges:
+            if all(util.isfinite(cl) for cl in self.clim):
+                low, high = self.clim
+            elif dim_name in ranges:
                 low, high = ranges[dim_name]['combined']
+                dlow, dhigh = ranges[dim_name]['data']
+                if (util.is_int(low, int_like=True) and
+                    util.is_int(high, int_like=True) and
+                    util.is_int(dlow) and
+                    util.is_int(dhigh)):
+                    low, high = int(low), int(high)
             elif isinstance(eldim, dim):
                 low, high = np.nan, np.nan
             else:
@@ -1732,10 +1740,8 @@ class ColorbarPlot(ElementPlot):
             if self.symmetric:
                 sym_max = max(abs(low), high)
                 low, high = -sym_max, sym_max
-            if util.isfinite(self.clim[0]):
-                low = self.clim[0]
-            if util.isfinite(self.clim[1]):
-                high = self.clim[1]
+            low = self.clim[0] if util.isfinite(self.clim[0]) else low
+            high = self.clim[1] if util.isfinite(self.clim[1]) else high
         else:
             low, high = None, None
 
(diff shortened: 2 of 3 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -1832,7 +1838,20 @@ def _get_color_data(self, element, ranges, style, name='color', factors=None, co
...
+                colormapper = LogColorMapper
+                if util.is_int(low) and util.is_int(high) and low == 0:
+                    low = 1
+                    if 'min' not in colors:
+                        colors['min'] = 'rgba(0, 0, 0, 0)'
+                elif util.is_number(low) and low == 0:
+                    self.param.warning(
+                        "Log color mapper lower bound is 0, ensure you "
+                        "set a non-zero lower bound on the color dimension "
+                        "or using the `clim` option."
```
jlstevens, line 1851: As a slight tweak, the lower bound isn't *supposed* to be zero but that is our heuristic/guess. So maybe something like:

> 'Unspecified log color mapper lower bound defaulting to 0, ....'

jlstevens:
Looks good to me! I think with an augmented warning message, this can be merged.

holoviews/plotting/bokeh/element.py
```diff
@@ -1832,7 +1838,20 @@ def _get_color_data(self, element, ranges, style, name='color', factors=None, co
 
     def _get_cmapper_opts(self, low, high, factors, colors):
         if factors is None:
-            colormapper = LogColorMapper if self.logz else LinearColorMapper
+            if self.logz:
+                colormapper = LogColorMapper
+                if util.is_int(low) and util.is_int(high) and low == 0:
+                    low = 1
+                    if 'min' not in colors:
+                        colors['min'] = 'rgba(0, 0, 0, 0)'
```
jbednar, line 1846: At first I thought this would be a black color, which wouldn't always be appropriate, but now I see that alpha is zero as well, so it's transparent.  Maybe add a comment to clarify:

```suggestion
                        # Make integer 0 be transparent
                        colors['min'] = 'rgba(0, 0, 0, 0)'
```

holoviews/plotting/bokeh/element.py
```diff
@@ -1832,7 +1838,20 @@ def _get_color_data(self, element, ranges, style, name='color', factors=None, co
...
+                elif util.is_number(low) and low == 0:
+                    self.param.warning(
+                        "Log color mapper lower bound is 0, ensure you "
+                        "set a non-zero lower bound on the color dimension "
+                        "or using the `clim` option."
+                    )
```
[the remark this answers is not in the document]
philippjfr: Good point.

holoviews/core/util.py
```diff
@@ -1265,6 +1265,13 @@ def is_number(obj):
     else: return False
 
 
+def is_int(obj, int_like=False):
```
philippjfr, line 1268:
```suggestion
def is_int(obj, int_like=False):
    """
    Checks for int types including the native Python type and NumPy-like objects

    Args:
        obj: Object to check for integer type
        int_like (boolean): Check for float types with integer value

    Returns:
         Boolean indicating whether the supplied value is of integer type.
"""
    
    
```

holoviews/plotting/bokeh/element.py
philippjfr: There are no assumptions being made and in all likelihood everything will be white since there is no particularly principled way to pick a lower bound.

Issue #4390, opened 2020-04-22
Pin the binder image to a specific release


maximlt:
Hi!

I tried a few times to launch the binder environment during the last couple of weeks. When the docker image isn't available (for various reasons mentioned [here](https://mybinder.readthedocs.io/en/latest/faq.html#what-factors-influence-how-long-it-takes-a-binder-session-to-start)), binder takes a pretty long time to build and deploy it, close to 10 minutes. This is pretty long and could discourage people.

One thing that causes rebuilding the image is that as of now binder reflects the current state of the **master** branch. So any new commit to master will force binder to rebuild an image when someone clicks on the launch button. One way to avoid tracking master could be to select a **specific release**. A tag can be provided to the binder link and a version can be pinned in the ``environment.yml`` file that binder uses to build the docker image. We could now pin it to ``1.13.2`` (I haven't found a way to tell binder to select the latest release, too bad).

An appropriate badge generated from [here](https://mybinder.readthedocs.io/en/latest/howto/badges.html) could also be displayed:

[...]

PR #4400, opened 2020-04-28
Pin binder to v1.13.2

maximlt:
As suggested in https://github.com/holoviz/holoviews/issues/4390 this PR pins both the conda environment and the repo to the last release, v1.13.2, so that binder doesn't have to build an image again every time there is a commit to master.

The files required by binder have been moved to a ``binder`` directory that it can detect.


rsignell-usgs:
@philippjfr and @maximlt, would another approach here would be to have the binder environment in a separate "binder" branch that only contains the binder environment?

PR #4404, opened 2020-05-01
Add ipywidgets and VSCode rendering support

philippjfr:


holoviews/ipython/__init__.py
```diff
@@ -83,6 +83,11 @@ class notebook_extension(extension):
 
     logo = param.Boolean(default=True, doc="Toggles display of HoloViews logo")
 
+    comms = param.ObjectSelector(
+        default='default', objects=['default', 'ipywidgets', 'vscode'], doc="""
+        Whether to render output in Jupyter with the default Jupyter
+        extension or use the jupyter_bokeh ipywidget model.""")
```
jbednar, line 89: Should the docstring mention the vscode option?

philippjfr: Not sure I want to merge this yet, but it we want to add this hacky workaround then yes. Otherwise we can just wait until VSCode adds ipywidgets 8.0 support and remove the VSCode option.

MarcSkovMadsen:
One vote for having vs code support ASAP.

holoviews/ipython/__init__.py
philippjfr: Comment from JL: Don't need this parameter

Issue #3944, opened 2019-09-04
Allow reset (clearing) of auto archive


jhermann:
# Use-case / Problem statement
As a notebook user, if you repeatedly use "Run All", the archive grows with each iteration. It would be nice if you could clear remnants of older executions on activation.

Right now, this is only possible via `hv.archive._files.clear()` after the `hv.archive.auto` call.

## Proposed solution
Add a public `clear()` method to the `FileArchive` class, delegating that call to the `_files` attribute (see above).

As a convenience, maybe also add a new `clear=True` kwparam (default: `False`) to the `auto()` call that triggers the new method.

jlstevens:
This is a very reasonable feature request that I support. It would also be a nice simple fix that would be suitable for a user contributed PR so I'll tag this with 'good first issue'. :-)

Yura52:
Looks like the issue can be closed, since the corresponding PR is merged.

jhermann:
That related PR is #4410

PR #4410, opened 2020-05-05
Allow clearing of auto archive

petros1999:
As new contributors, we would appreciate any suggestions or guidelines in order to make this pr succesfull. Thanks in advance

### DIFF
--- a/holoviews/core/io.py
+++ b/holoviews/core/io.py
@@ -854,3 +854,10 @@ class FileArchive(Archive):
     def listing(self):
         "Return a list of filename entries currently in the archive"
         return ['.'.join([f,ext]) if ext else f for (f,ext) in self._files.keys()]
+
+    def clear(self):
+        "Clears the file archive"
+        self._files.clear()
+        return self._files
+    
+
--- a/holoviews/core/io.py
+++ b/holoviews/core/io.py
@@ -858,6 +858,6 @@ class FileArchive(Archive):
     def clear(self):
         "Clears the file archive"
         self._files.clear()
-        return self._files
+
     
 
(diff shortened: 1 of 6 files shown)


petros1999:
This pull request refers to issue #3944 
We have added a clear() method in FileArchive class of io.py and the corresponding test case.
Also we have added an if statement in auto() method of NotebookArchive class in which we call the clear() method.

holoviews/tests/core/testarchives.py
```diff
@@ -124,3 +124,14 @@ def test_filearchive_json_single_file(self):
...
+    """
+    A test case that examines the clear() method of
+    FileArchives of io.py
+    """
+    def test_filearchive_clear_file(self):
+        export_name = "archive_for_clear"
+        archive = FileArchive(export_name=export_name, pack=False)
+        archive.add(self.image1)
+        archive.add(self.image2)
+        self.assertEqual(None, FileArchive.clear(archive))
```
philippjfr, line 137: Not sure what this is testing, shouldn't this be something like:

```suggestion
        archive.clear()
        self.assertEqual(archive._files, {})
```

holoviews/core/io.py
```diff
@@ -854,3 +854,10 @@ def contents(self, maxlen=70):
     def listing(self):
         "Return a list of filename entries currently in the archive"
         return ['.'.join([f,ext]) if ext else f for (f,ext) in self._files.keys()]
+
+    def clear(self):
+        "Clears the file archive"
+        return self._files.clear()
```
philippjfr, line 860: Don't think `dict.clear()` returns anything

```suggestion
        self._files.clear()
```

philippjfr:
Much appreciated, thanks for contributing!

Issue #3751, opened 2019-05-31
Remove toolbar from png


jherreral:
When a plot is exported to png using `hv.save()` and Bokeh renderer the toolbar is plotted by default. This is also true if `hv.render()` and `bokeh.io.export_png()` are used. On the other side, when exporting to svg using `hv.render()` and `bokeh.io.export_svgs()` the toolbar is left out by default.
There are quick workarounds like `hv.save(myPlot.options(toolbar=None),'myExample.png')` that make the change inplace, but I think that it should be left out by default like in svg.

Using `hv.save(myPlot,'myDirtyPlot.png)`
_Why would a png have a toolbar?_
![myDirtyPlot](https://user-images.githubusercontent.com/33136201/58712571-d2fe8100-838e-11e9-86ff-3d504e391e22.png)

Using `hv.save(myPlot.options(toolbar=None),'myNeatPlot.png')`
_This is what I expected!_
![myNeatPlot](https://user-images.githubusercontent.com/33136201/58712561-cda13680-838e-11e9-8f05-3edc41115ceb.png)

jbednar:
I agree; PNGs only need toolbars on the Bokeh site, for docs about Bokeh; otherwise I think they should be gone by default.

PR #4421, opened 2020-05-14
Add obj.potions(toolbar=None) to render and save method

petros1999:
This pull request refers to issue #3751

Every comment or guideline in order to make the pr successful is more than welcome. Thanks in advance

### DIFF
--- a/holoviews/core/io.py
+++ b/holoviews/core/io.py
@@ -854,3 +854,10 @@ class FileArchive(Archive):
     def listing(self):
         "Return a list of filename entries currently in the archive"
         return ['.'.join([f,ext]) if ext else f for (f,ext) in self._files.keys()]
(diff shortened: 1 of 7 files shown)


holoviews/core/io.py
```diff
@@ -854,3 +854,10 @@ def contents(self, maxlen=70):
     def listing(self):
         "Return a list of filename entries currently in the archive"
         return ['.'.join([f,ext]) if ext else f for (f,ext) in self._files.keys()]
+
+    def clear(self):
+        "Clears the file archive"
+        self._files.clear()
```
philippjfr, line 860: You seem to be basing your PRs off an commit, which keeps reintroducing these changes, which makes it harder to review your PRs.

PR #4430, opened 2020-05-20
Clarify data types accepted by Points

jbednar:
This may be a local fix to a global problem, but it's difficult to tell which types of data are accepted by HoloViews elements, so I thought it would be useful to illustrate various accepted formats for at least one basic Element type, i.e. Points.  The previous example uses np.random to make a random multidimensional array, which is convenient but is quite opaque -- what does each column and row in that array represent? How would I format my own data to use with this Element?

I've kept the existing examples mostly as-is, but added a new bit explicitly listing the x, y, size, and color variables and showing how to supply those to Points:

image

Issues:

- [ ] Is this the right place to put this information?  If not, where, and we'll need to link to it from here.
- [ ] Is this a relatively complete list of the data formats that are reasonable to provide? It doesn't need to be exhaustive, but should include the most typical or common ways that we expect users to provide data.
- [ ] Pending approval of this change, I have not updated the Matplotlib and Plotly notebooks in the same way yet.

Also note that I updated the colormap, because the previous colormap ("fire") was mapping the brightest values to white (the background color of the page), making them disappear:

image

as-is (apart from updating the colormap so

### DIFF
--- a/examples/reference/elements/bokeh/Points.ipynb
+++ b/examples/reference/elements/bokeh/Points.ipynb
@@ -24,6 +24,7 @@
    "source": [
     "import numpy as np\n",
     "import holoviews as hv\n",
+    "import pandas as pd\n",
     "from holoviews import opts, dim\n",
     "hv.extension('bokeh')"
    ]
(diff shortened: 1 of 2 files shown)


jbednar:
I'm also wondering why I had to specify `, vdims=['z', 'size']` for the dictionary but not for the dataframe. Aren't the dimension names available in the same way already in both cases? For whatever reason, the dataframe works fine without the explicit vdims but the dictionary does not:

```
WARNING:param.PointPlot03040: Specified size dim transform dim('size')*20 
could not be applied, as not all dimensions could be resolved.
```

philippjfr:
>Aren't the dimension names available in the same way already in both cases?

DataFrames have column order, dictionaries do not (until Python 3.6).

jbednar:
How does ordering affect finding `size` by name?

philippjfr:
The lack of ordering means we cannot automatically deduce a consistent dimension order so you have to declare dimensions explicitly.

PR #4437, opened 2020-05-25
Linked selection improvements

philippjfr:
- Fixes empty lasso selects causing errors during selection


philippjfr:
Dropping this for now and merging just the lasso fix:

- Removes special handling of Histogram linked selections, the idea was nice but:
  * Selecting on the bin height is not that useful
  * np.digitize on the input on every selection is fairly expensive
  * It is very complex and brittle code
  * The digitize operation is not supported by a cuDF

PR #4463, opened 2020-06-08
Implement support for setting custom options on plot components

philippjfr:
A long time ago we introduced hooks to allow users to apply customizations for options we haven't exposed on the plot classes themselves. However writing custom hooks is cumbersome and quite verbose. This PR implements a suggestions I've been mulling for a long time (I haven't been able to find the issue where I laid out this proposal). Basically it allows declaring a dictionary with accessor specification for setting custom options on the plot components, e.g. something like this:

```
{'colorbar.title_text_color': 'red'}
```

Looks up the colorbar model on the plot and then sets the `title_text_color' to 'red'. This approach should work for arbitrarily nested objects. The user just needs some knowledge of the attributes of the underlying plotting library (which is also needed for hooks) but here we can at least in theory apply some fuzzy-matching to give good suggestions. 

- [x] Add tests
- [x] Implement for matplotlib and plotly backends
- [x] Implement some fuzzy matching to give helpful suggestions when the spec is slightly mis-specified.
- [x] Add documentation.


philippjfr:
>This might provide an easy way for options to be added over time. E.g. hv.declare_option(hv.Points, 'colorbar_title_color', 'colorbar.title_text_color'), then hv.Points(...).opts(colorbar_title_color="red")

I like that idea a lot. I did consider whether if you fully qualify it could just support it by default, i.e. `colorbar.title_text_color` could be set as `colorbar_title_text_color` but that would make early validation impossible I suppose.

jonmmease:
For the Plotly backend, it's not as clear cut what to use as the root component(s) for the dot notation.  For the `Scatter` element for example, many of the style properties live under `scatter.marker`  (like `color`) in the plotly schema, but some live directly under `scatter` (like `selectedpoints`). Other properties (like dragmode above) live under the `figure.layout` hierarchy.

These are things I'd like to clean up eventually, but it runs directly into the larger tension of whether to try to unify style option names across backends.  For someone used to the plotly API, it would be really nice if all of the `Scatter` style property names matched their path relative to the plotly `scatter` trace, with underscores separating nested levels. This would mean using `marker_color` instead of `color`.  This wouldn't match bokeh and matplotlib, but it would make it much easier to document and validate. The `hv.help` output could directly reference where in the plotly documentation to look for descriptions of the options (e.g. https://plotly.com/python/reference/scatter/). And we could automate the lookup of the appropriate plotly.py validator if we wanted to delegate property validation to plotly.py.

If we went this direction and were willing to change the plotly style names in a HoloViews 2.0 to be more consistent with the plotly schema, then I think this proposal could work for the Plotly backend as well.

The alternative is the keep style options as-is and work towards unifying them across the plotting backends. This is the case where I'd especially want https://github.com/holoviz/holoviews/issues/4638 as a fallback.

philippjfr:
Note there isn't necessarily any one root element, the dot notation starts by inspecting the `.handles` on the plot class where we can put anything and establish convenient conventions that way.

jbednar:
Right; it sounds like this could be just what I was envisioning in https://github.com/holoviz/holoviews/issues/4638 , i.e., a concise way of specifying options on some hierarchy of objects, for whatever hierarchy is available on a given plotting backend, which is particularly valuable if it also has a way to map those hierarchical specifications into flat option names.

jonmmease:
> Note there isn't necessarily any one root element, the dot notation starts by inspecting the .handles on the plot class where we can put anything and establish convenient conventions that way.

Oh, ok. So for the plotly `ScatterPlot` we could add references to multiple anchor points in the same figure dictionary to `handles`.

```python
self.hanldes["figure"] = fig
self.hanldes["layout"] = fig["layout"]
self.hanldes["trace"] = fig["data"][0]
self.hanldes["marker"] = fig["data"][0]["marker"]
self.hanldes["xaxis"] = fig["layout"]["xaxis"]
self.hanldes["yaxis"] = fig["layout"]["yaxis"]
```

So this way, the custom property string could start with `figure`, `layout`, `trace`, `marker`, `xaxis`, or `yaxis`.

```python
scatter.opts(trace_visible=True, marker_color="green", layout_dragmode="pan", xaxis_tickfont_family="courier")
```

I'd be happy with this from the plotly side. These anchor points could be documented in `hv.help` making it pretty easy to cross reference the appropriate plotly documentation section.

I'm wondering if we could work toward the following split:
 - All properties with names that we want to be consistent across backends, and all properties that HoloViews does special  handling for (e.g. dim expression support, colormap construction, etc.) are moved to plot options.
 - All style options have names that systematically match the corresponding backend (e.g. with underscores separating nested levels as above), and HoloViews passes them directly through to the backend without validation or conversions.

To me, this would be a lot less confusing than having the style options consist of a mix of options that are passed through with matching names in the backend, and options that are have special handling by HoloViews and/or names that are designed to be consistent across backends.

jbednar:
Sounds like a good plan, though I don't know what backwards compatibility issues that would raise.

jlstevens:
I'm not sure how I feel about this as a plot option given that there is no obvious way to generalize this idea to plotly and matplotlib. The name should also be something like `custom_setters` imho which describes the idea more accurately imho.

philippjfr:
>I'm not sure how I feel about this as a plot option given that there is no obvious way to generalize this idea to plotly and matplotlib.

That's exactly what we're discussing above. There absolutely are ways of generalizing this to both plotly and matplotlib but you're right it's not totally straightforward.

>At first glace, I don't like the idea of something like hv.declare_option(hv.Points, 'colorbar_title_color', 'colorbar.title_text_color') automatically (as this will hugely explode the number of options) but I might support it instead of custom_opts because then these would appear as style options rather than plot options which I think is more appropriate.

I'm also wary about this, doing this automatically and exploding the option space in this way will definitely make the tab-completions less useful for instance.

jlstevens:
> I'm also wary about this, doing this automatically and exploding the option space in this way will definitely make the tab-completions less useful for instance.

I can't remember exactly how the tab completion works but I think we could allow these keywords without suggesting them. This makes the proposal for automatic expansion more palatable to me modulo the possibility of namespace clashes.

jlstevens:
> To me, this would be a lot less confusing than having the style options consist of a mix of options that are passed through with matching names in the backend, and options that are have special handling by HoloViews and/or names that are designed to be consistent across backends.

That was always the idea of plot options versus style options but unfortunately the boundary isn't always as hard as we would like. In some cases plot options simply can't be easily made consistent across backends and sometimes style options need some special handling before they are passed through to the plotting library. What you suggest has always been our goal but being totally strict about it for the ambiguous cases would be more confusing than helpful imho...

philippjfr:
In practical terms style options have always been things that control the visual styling of a particular glyph/artist/trace, while plot options have applied to everything else axes/colorbars/legends/titles etc. However it is also true that for plot option we've generally tried to maintain compatibility between backends (as much as possible), so this definitely kind of straddles the line of the two since most of these options would be about axes/colorbars/legends etc. but they are also definitely backend specific.

jonmmease:
> In practical terms style options have always been things that control the visual styling of a particular glyph/artist/trace, while plot options have applied to everything else axes/colorbars/legends/titles

I guess my core difficulty is that I just don't grok the difference here. `kdims` and `vdims` are annotations about the data and, to me, everything else is visual / styling. How to encode data by position (e.g. logx=True, invert_axes=True) isn't fundamentally different from encoding the same data by color or marker size.

I'd argue that the marker color/size encoding is more related to the core nature of the plot than current plot options like the padding around the colorbar.

The distinction of whether options are uniform across backends is a very practical distinction.

jbednar:
I haven't previously been pushing for _automatically_ registering these extra options, just to make it easy for a user to register some.  But thinking about it in terms of root elements for each backend, I wonder if there is a good middle ground: register a finite set of root elements for each backend, and then accept options automatically and without validation, but only for those root elements. I.e., validate up to the first underscore, and leave the rest to the backend.  That way we wouldn't lose validation in general, while also supporting the full set of options available in each backend.  Of course, if we could query the backend for the actual available options, as is possible for Bokeh and may also be possible for other backends, then we could provide validation and tab completion even in that case as well.

jlstevens:
Discussing this with @jbednar , @philippjfr  and @jonmmease  I think I'm fine with the PR as is, if it can be generalized to matplotlib and plotly as well with `custom_opts` renamed to something more suitable.

I suggest `config` as this style is very much like the matplotlib configs/rcfiles. This plot option would take precedence over other options (e.g `opacity` here would override `alpha` for plotly). It would then be a very direct way for users to make use of the API corresponding to the selected plotting backend.

Having the option for users to add such options to the allowed keywords seems fragile to me as that code would be having a side effect potentially far from where the keyword is used (e.g someone copies and pastes a cell at the bottom of a notebook using a keyword enabled at the top of the notebook). Doing such expansion automatically seems problematic due to namespace clashes and the reduction in utility of tab-completion.

jonmmease:
I would suggest that the option name include `backend` to make it more clear that these are options are specific to a particular backend. Maybe `backend_opts` or `backend_config`?

But other than that I'm happy with it as well and I could work on the Plotly backend portion.  For the plotly backend, I think I could actually implement early validation if that was desirable.

Wouldn't need to be this PR necessarily, but the remaining question for me is discoverability.  Could we add a backend-specific section to `hv.help` that documents what root elements are available for the and directs people to some external resource where they can find the available options? For example, `hv.help` on a `Scatter` element with the plotly backend would include something like:

```
root elements
 - figure (https://plotly.com/python/reference/)
 - trace (https://plotly.com/python/reference/scatter/)
 - marker (https://plotly.com/python/reference/scatter/#scatter-marker)
 - layout (https://plotly.com/python/reference/layout/)
 - xaxis (https://plotly.com/python/reference/layout/xaxis/)
 - yaxis (https://plotly.com/python/reference/layout/yaxis/)
```


One last thing. Some plotly elements include lists, so it would be nice to support list indexing in the strings as well. e.g.

```python
{'layout.images[0].opacity': 0.5}
```

jbednar:
> I would suggest that the option name include backend to make it more clear that these are options are specific to a particular backend. Maybe backend_opts or backend_config?

Hmm.  "backend_" makes it wordy without contributing much, especially since people always get confused about what constitutes a backend (given that most people think of Bokeh and Plotly as _frontends_, in that they are what the users interact with directly). 

What about the actual backend name as a prefix instead? I.e. `bokeh_config={'colorbar.title.color':'red'}`, to directly state that these are bokeh configuration options (and similarly for "plotly_config" and "matplotlib_config" or "mpl_config")? `bokeh_opts` would be shorter but is confusing with the hv options system, so I'd avoid that.  `bokeh_settings` seems clearer than `config` to me, but it's longer, so maybe not that.

> One last thing. Some plotly elements include lists, so it would be nice to support list indexing in the strings as well. e.g.

I guess attribute access covers dictionary-like access already, for the only reasonable case (string keys), and thus we don't need to consider supporting dictionaries too? 
```
{'layout["toolbar"].num_colors': 2}
```

jlstevens:
The names `plotly_config`, `matplotlib_config` and `mpl_config` are all fine by me if we don't want it to just be `config`.

philippjfr:
Also happy with `bokeh_config`, `mpl_config` and `plotly_config`.

philippjfr:
>I guess attribute access covers dictionary-like access already, for the only reasonable case (string keys), and thus we don't need to consider supporting dictionaries too?

Yes, that was the idea, I'd definitely like to avoid having strings in strings and the hassles that causes.

jonmmease:
> Also happy with bokeh_config, mpl_config and plotly_config

Works for me!

philippjfr:
@jonmmease @jbednar @jlstevens should we aim to get this into the release?

jonmmease:
> @jonmmease @jbednar @jlstevens should we aim to get this into the release?

I wasn't picturing we would for this release, given release date goal.

jbednar:
Up to you, @philippjfr ; if it's feasible to implement something you're happy with, I'd find a way to review it in time.

ahuang11:
Okay I implemented MPL support and tests. I was wondering if plotly is also a requirement?
```
import holoviews as hv

hv.extension("matplotlib")

heat_map = hv.HeatMap([(1, 2, 3), (2, 3, 4), (3, 4, 5)]).opts(
    colorbar=True,
    custom_opts={
        "colorbar.set_label": "Testing",
        "colorbar.set_ticks": [3.5, 5],
        "colorbar.ticklabels": ["A", "B"],
        "ax.spines.right.visible": False,
        "ax.spines.top.visible": False,
    },
)
heat_map
```
image


Bokeh also works:
```
import holoviews as hv

hv.extension("bokeh")

heat_map = hv.HeatMap([(1, 2, 3), (2, 3, 4), (3, 4, 5)]).opts(
    colorbar=True,
    custom_opts={
        "colorbar.title": "Testing",
        "colorbar.ticks": [3.5, 5],
    },
)
heat_map
```
image

Here's the fuzzy search
`WARNING:param.HeatMapPlot02515: Could not find 'ticks' property on ColorBar model. Ensure the custom option spec 'colorbar.ticks' you provided references a valid attribute on the specified model. Similar options include ['ticker']`


Also, I wanted to ask, where should I put this in docs?

philippjfr:
This is great @ahuang11! I'd add docs to https://holoviews.org/user_guide/Plotting_with_Bokeh.html and https://holoviews.org/user_guide/Plotting_with_Matplotlib.html for now. Once we re-organize these into how-to guides it'll be a lot more discoverable.

Plotly would definitely be nice but doesn't seem like a must have.

philippjfr:
Presumably the aim was to index into some dictionary or list subobject, e.g. `'legend[0].font_size': '10pt'` or something. Agree `backend_opts` seems more appropriate.

ahuang11:
I couldn't find an equivalent example for bokeh to get item, but for matplotlib:
image


```python
import holoviews as hv

hv.extension("matplotlib")

a = hv.Curve([1, 2, 3], label="a")
b = hv.Curve([1, 4, 9], label="b")
plot = (a * b).opts(
    show_legend=True,
    backend_opts={
        "legend.get_texts()[0].fontsize": 35,
    }
)
plot
```

holoviews/plotting/bokeh/tiles.py
```diff
@@ -24,7 +24,7 @@ def get_extents(self, element, ranges, range_type='combined'):
 
     def get_data(self, element, ranges, style):
         if not isinstance(element.data, (str, dict)):
-            SkipRendering("WMTS element data must be a URL string or "
+            SkipRendering("WMTS element data must be a URL string, dictionary, or"
```
philippjfr, line 27:
```suggestion
            SkipRendering("WMTS element data must be a URL string, dictionary, or "
```

holoviews/plotting/plot.py
```diff
@@ -1493,6 +1501,92 @@ def _format_title_components(self, key, dimensions=True, separator='\n'):
...
 
+    def _parse_backend_opt(self, opt, plot, model_accessor_aliases):
+        """
+        Parses a custom option of the form 'model.accessor.option'
+        and returns the corresponding model and accessor.
+        """
+        accessors = opt.split('.')
+        if len(accessors) < 2:
+            self.param.warning("Custom option %r expects at least "
+                                "two accessors separated by '.'" % opt)
```
philippjfr, line 1512: Nit

```suggestion
            self.param.warning(f"Custom option {opt!r} expects at least "
                               "two accessors separated by '.'")
```

holoviews/plotting/mpl/element.py
```diff
@@ -240,6 +244,38 @@ def _finalize_ticks(self, axis, dimensions, xticks, yticks, zticks):
...
+        model_accessor_aliases = {
+            "figure": "fig",
+            "axes": "axis",
+            "ax": "axis",
+            "colorbar": "cbar",
+        }
+
+        for opt, val in self.backend_opts.items():
+            parsed_opt = self._parse_backend_opt(
+                opt, plot, model_accessor_aliases)
```
philippjfr, line 259: Why limit this to the `model_accessor_aliases`, can we pass in `dict(self.handles, **model_accessor_aliases)` instead?

philippjfr: Nvm, I see the handling below.

holoviews/plotting/plot.py
```diff
@@ -1493,6 +1501,92 @@ def _format_title_components(self, key, dimensions=True, separator='\n'):
...
+                                "figure object.".format(model_accessor,
+                                                    type(self).__name__,
+                                                    opt, list(self.handles.keys())))
+            return
+
+        for acc in accessors[1:-1]:
+
+            # the logic handles resolving something like:
+            # legend.get_texts()[0].set_fontsize
+            if '[' in acc and acc.endswith(']'):
```
philippjfr, line 1538: What if there are multiple getitems? Seems rare but possible in theory.

holoviews/plotting/plot.py
```diff
@@ -1493,6 +1501,92 @@ def _format_title_components(self, key, dimensions=True, separator='\n'):
...
+
+        # convert alias to handle key (figure -> fig)
+        model_accessor = model_accessor_aliases.get(model_accessor) or model_accessor
+
+        if model_accessor in self.handles:
+            model = self.handles[model_accessor]
+        elif hasattr(plot, model_accessor):
+            model = getattr(plot, model_accessor)
+        else:
+            self.param.warning("{} model could not be resolved on {} plot. "
```
philippjfr, line 1525: I think let's make sure we use f-strings going forward.

holoviews/plotting/plot.py
```diff
@@ -1493,6 +1501,92 @@ def _format_title_components(self, key, dimensions=True, separator='\n'):
...
+
+        for acc in accessors[1:-1]:
+
+            # the logic handles resolving something like:
+            # legend.get_texts()[0].set_fontsize
+            if '[' in acc and acc.endswith(']'):
+                getitem_index = acc.index('[')
+                getitem_spec = acc[getitem_index+1:-1]
+                try:
+                    getitem_acc = literal_eval(getitem_spec)
```
philippjfr, line 1542: This won't handle slices (which is probably okay).

holoviews/plotting/plot.py
```diff
@@ -1493,6 +1501,92 @@ def _format_title_components(self, key, dimensions=True, separator='\n'):
...
+            else:
+                getitem_acc = None
+
+            if "(" in acc and ")" in acc:
+                method_ini_index = acc.index("(")
+                method_end_index = acc.index(")")
+                method_spec = acc[method_ini_index + 1:method_end_index]
+                try:
+                    if method_spec:
+                        method_args = literal_eval(method_spec)
```
philippjfr, line 1559: This won't handle kwargs (which may also be okay).

philippjfr (commented):
Left a few comments. The parser does have a few limitations which are okay but maybe it's worth validating against them.

holoviews/plotting/plot.py
ahuang11: Added support for `"legend.get_texts()[3,4].fontsize": 388,`

holoviews/plotting/plot.py
ahuang11: "legend.get_texts()[1:2].fontsize": 388,

holoviews/plotting/plot.py
ahuang11: Added support, but couldn't think of a practical example. Let me know if you think of one.

examples/user_guide/Plotting_with_Matplotlib.ipynb
```diff
@@ -52,6 +52,91 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "plot = hv.Curve([1, 2, 3], label=\"a\") * hv.Curve([1, 4, 9], label=\"b\")\n",
+    "plot.opts(\n",
+    "    show_legend=True,\n",
+    "    backend_opts={\"legend.get_texts()[0].fontsize\": 18}\n",
```
ahuang11, line 127:
```suggestion
    "    backend_opts={\"legend.get_texts()[0:2].fontsize\": 18}\n",
```

holoviews/plotting/plot.py
```diff
@@ -1493,6 +1501,128 @@ def _format_title_components(self, key, dimensions=True, separator='\n'):
...
+        model_accessor = model_accessor_aliases.get(model_accessor) or model_accessor
+
+        if model_accessor in self.handles:
+            model = self.handles[model_accessor]
+        elif hasattr(plot, model_accessor):
+            model = getattr(plot, model_accessor)
+        else:
+            self.param.warning(
+                f"{model_accessor} model could not be resolved "
+                f"on {type(self).__name__} plot. "
```
ahuang11, line 1527:
```suggestion
                f"on {type(self).__name__!r} plot. "
```

holoviews/plotting/plot.py
```diff
@@ -1493,6 +1501,128 @@ def _format_title_components(self, key, dimensions=True, separator='\n'):
...
+                f"Ensure the {opt!r} custom option spec "
+                f"references a valid model in the "
+                f"plot.handles {list(self.handles.keys())} or on the underlying "
```
ahuang11, line 1530:
```suggestion
                f"plot.handles {list(self.handles.keys())!r} or on the underlying "
```

holoviews/plotting/plot.py
```diff
@@ -1493,6 +1501,128 @@ def _format_title_components(self, key, dimensions=True, separator='\n'):
...
+                        model = [model.__getitem__(i) for i in getitem_acc]
+                acc = acc[method_end_index:]
+
+            if acc == "" or model is None:
+                continue
+
+            if not hasattr(model, acc):
+                self.param.warning(
+                    f"Could not resolve {acc!r} attribute on "
+                    f"{type(model).__name__} model. Ensure the "
```
ahuang11, line 1614:
```suggestion
                    f"{type(model).__name__!r} model. Ensure the "
```

ahuang11:
Thanks for the review! I've added support for multiple indices, slices, and kwargs, but I couldn't immediately think of a practical example for kwargs (there probably is some use case somewhere).

holoviews/plotting/plot.py
Hoxbro: How does it handle something like `"legend.get_texts()[3][4].fontsize"`?

ahuang11: Before adding even more complexity, can you think of a valid use case for that?

I don't think this is valid:
legend.get_texts()[1][2].fontsize

legend.get_texts() => ["a label", "b label"]
[1] => "a label"
[2] => "l"
fontsize => raise AttributeError (because l is `str`)

ahuang11: It will likely fail

Hoxbro:
@ahuang11, Python 3.8 tests on Ubuntu tests are failing, can you take a look at it? So we can get this PR merged.

ahuang11:
Thinking about how to address this since I don't have a ubuntu machine.

Wondering what version of matplotlib is being used.

```
FAILED holoviews/tests/plotting/matplotlib/test_elementplot.py::TestElementPlot::test_element_backend_opts - AssertionError: ['', ''] != ['A', 'B']
FAILED holoviews/tests/plotting/matplotlib/test_elementplot.py::TestElementPlot::test_element_backend_opts_alias - AssertionError: ['', ''] != ['A', 'B']
```

Hoxbro:
Ir is available in the capture environment section in holoviz-dev/holoviz_tasks/install@v0.1a15

ahuang11:
Thanks; seems like the matplotlib is a bit old...
Will see if I can reproduce this on my Mac

```
  matplotlib                3.3.2                         0    conda-forge
  matplotlib-base           3.3.2            py38h5c7f4ab_1    conda-forge
  matplotlib-inline         0.1.6              pyhd8ed1ab_0    conda-forge
```

ahuang11:
Okay, it's fixed now!

Hoxbro:
I think people are going to love this feature. 

Thank you, @ahuang11, for finishing this PR!

PR #4491, opened 2020-06-22
Pass inline argument to load_nb

### LINKED ISSUE #4490: hv.extension('bokeh', inline=False) doesn't load javascript from CDN

* Holoviews 1.13.2
* jupyterlab 2.1.0
* bokeh 2.0.1
* panel 0.9.5

To reduce the size of the notebooks, I use `holoviews.extension('bokeh', inline=False)`, but the size of the notebook doesn't change.

```python
import holoviews as hv
hv.extension('bokeh', inline=False)
```

and then check the size of the notebook.

I found how to fix this:

https://github.com/holoviz/holoviews/blob/0693a07f2af1095bca84be9c9a8a2503d1ded9ab/holoviews/ipython/__init__.py#L188

change the line to `Renderer.load_nb(inline=p.inline)`.

ruoyu0088:
fix #4490

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -185,7 +185,7 @@ class notebook_extension(extension):
 
         for r in [r for r in resources if r != 'holoviews']:
             Store.renderers[r].load_nb(inline=p.inline)
-        Renderer.load_nb()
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks so much @ruoyu0088!

PR #4506, opened 2020-07-02
Standardizing build procedures

kebowen730:
This adds support for building dev and release websites with commit messages containing `website_dev` and `website_release` respectively, and building building from tags using `website_release` and `website_dev`, as well as tags of the format `v1.3.1a4` for dev, and `v1.3.1A4` for release.


philippjfr:
We are in the process of switching to GH Actions now so I don't see much point in merging this anymore.

Issue #4512, opened 2020-07-08
MultiDimensionalMapping.info doesn't print anything


ruoyu0088:
holoviews 1.13.3

The following code should print some information about the empty HoloMap, but it output nothing.

```
import holoviews as hv
hv.HoloMap().info
```

The source code of https://github.com/holoviz/holoviews/blob/master/holoviews/core/ndmapping.py#L465 doesn't call `print()`.

PR #4513, opened 2020-07-08
add print() to the info property

ruoyu0088:
fix https://github.com/holoviz/holoviews/issues/4512#issue-652965036


holoviews/core/ndmapping.py
```diff
@@ -486,6 +486,7 @@ def info(self):
                 if d.value_format:
                     dmin, dmax = d.value_format(dmin), d.value_format(dmax)
                 info_str += '\t %s: %s...%s \n' % (d.pprint_label, dmin, dmax)
+        print(info_str)
```
jbednar, line 489: I know the docstring says "print", but I think what was actually intended is "return", which then prints in Jupyter or on the commandline but can also be used in other ways:

```suggestion
        return info_str
```

My guess is that `return info_str` was deleted accidentally at some point, and you're the first to notice, so thanks for pointing it out!

PR #4517, opened 2020-07-10
Add an IbisInterface

philippjfr:
Adds an IbisInterface built on the [Ibis project](https://docs.ibis-project.org/index.html) which will let us interface directly with various database backends.

So far this is a basic skeleton which we can fill in bit by bit. The first item for discussion is what object a HoloViews `Dataset` should wrap. To me it seems most appropriate for it to wrap any `Expr` object which represents some expression on a Database, e.g. the selection of an individual Table or any other operation. The question then is what types of expressions should be supported, e.g. can or should we support a `ColumnExpr` or must the basic unit always represent a table.

### Core methods

- [x] `init`: Basic constructor
- [x] `range`: Computes the range along a column  
- [x] `values`: Returns all values along a column (or only unique values)
- [x] `length`: Returns number of rows in the table
- [x] `nonzero`: Whether there are any rows in the table
- [x] `dtype`: Return the dtype of the column
- [x] `isscalar`: Checks whether column value has a single unique value
- [x] `select`:  Select given boolean mask or given scalar, range or list selection(s)
- [x] `groupby`: Returns one or more Datasets grouped by the values along one or more columns 
- [x] `redim`: Rename columns in the table
- [x] `iloc`: Integer indexing of rows and columns
- [x] `sort`: Sort by one or more columns ascending/descending
- [x] `aggregate`: Aggregate along one or more columns with given function (e.g. mean, min, max)
- [x] `sample`: A list of selections

### Write methods (these methods write back to the data)

- [x] `add_dimension`: Add a column to the table (can this be supported?)
- [ ] `mask`: Mask out values given a boolean mask
- [x] `assign`: Add or overwrite one or more columns

### Conversion methods

- [x] `array`: Return a NumPy array of requested columns

## Tests Status

**Passing**: 73
**Error**: 34
**Fail**: 0
**Skip**: 16

Cc: @kcpevey @tonyfast

### DIFF
--- a/holoviews/tests/core/data/testibisinterface.py
+++ b/holoviews/tests/core/data/testibisinterface.py
@@ -20,7 +20,7 @@ from .base import HeterogeneousColumnTests, ScalarColumnTests, InterfaceTests
 
 
 def create_temp_db(df, name, index=False):
-    file_obj = NamedTemporaryFile()
+    file_obj = NamedTemporaryFile(dir='.')
     con = sqlite3.Connection(file_obj.name)
     df.to_sql(name, con, index=index)
     return sqlite.connect(file_obj.name)
@@ -34,13 +34,7 @@ class IbisDatasetTest(HeterogeneousColumnTests, ScalarColumnTests, InterfaceTest
     datatype = "ibis"
     data_type = (ibis.expr.types.Expr,)
 
-    __test__ = True
-
-    def setUp(self):
-        self.init_column_data()
-        self.init_grid_data()
-        self.init_data()
-
+    __test__ = Truesqlite3.OperationalError: unable to open database file
     def tearDown(self):
         pass
 
--- a/holoviews/tests/core/data/testibisinterface.py
+++ b/holoviews/tests/core/data/testibisinterface.py
@@ -1,5 +1,4 @@
 import sqlite3
-import pathlib
 from unittest import SkipTest
 
 from tempfile import NamedTemporaryFile
--- a/holoviews/tests/core/data/testibisinterface.py
+++ b/holoviews/tests/core/data/testibisinterface.py
@@ -16,6 +16,7 @@ import numpy as np
 import pandas as pd
 
 from holoviews.core.data import Dataset
+from holoviews.core.spaces import HoloMap
 
 from .base import HeterogeneousColumnTests, ScalarColumnTests, InterfaceTests
 
@@ -89,8 +90,10 @@ class IbisDatasetTest(HeterogeneousColumnTests, ScalarColumnTests, InterfaceTest
         self.dataset_hm = Dataset(hm_db.table('hm'), kdims=['x'], vdims=['y'])
         self.dataset_hm_alias = Dataset(hm_db.table('hm'), kdims=[('x', 'X')], vdims=[('y', 'Y')])
 
-
     def test_dataset_array_init_hm(self):
+        raise SkipTest("Not supported")
+
+    def test_dataset_dict_dim_not_found_raises_on_scalar(self):
         raise SkipTest("Not supported")
 
     def test_dataset_array_init_hm_tuple_dims(self):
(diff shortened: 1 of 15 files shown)


MarcSkovMadsen:
I have been thinking of HoloViews/ Panel and Ibis lately on the back of a blog by quantstack.

I cannot read from the above the why. So i hope you will write a bit about the why or vision.

For me the vision would be to make it very easy to use a wide range of data sources within HoloViz. Similarly to how easy it is to use BI tools like Tableau. 

And maybe this interface could be used to only load the data needed to display in the plot or do the aggregation on the backend server.

For me unfortunately Ibis does not support sql server or snowflake which means i cannot use it in my Daily Work.

jbednar:
Yes, the vision is to be able to instantiate a HoloViews element where the "data" is lazy, just a query and not the actual data, with the data fetched dynamically when it is actually needed for plotting or analysis.  That does indeed set HoloViews up to do some more of the things that BI tools do. 

If you have any funding, I would imagine that [Quansight](https://www.quansight.com) would add support to Ibis for sql server, and maybe snowflake...

kcpevey:
@MarcSkovMadsen Adding additional support in Ibis is definitely something we at Quansight would be open to discussing.

tonyfast:
There is some WIP on a microsoft sql server interface https://github.com/ibis-project/ibis/pull/1997

philippjfr:
This is starting to look good. I've got a few comments but will hold off until you tell me it's ready.

tonyfast:
I'd love to take some comments soon. I'm nearing the end of the skeleton and starting to understand the model better.

Is my PR what is breaking the travis tests?

philippjfr:
I'll do a high-level review today then. While I do think this PR is
probably failing tests, the tests are also generally broken today. I'll
merge a PR to get the tests running again and then you can rebase.

philippjfr:
Will have a go at this now.

philippjfr:
Huge amount of progress, 73 tests passing and 34 errors.

philippjfr:
Okay the only real problems remaining:

* iloc doesn't work
* groupby and aggregate do not preserve ordering (this is okay, just needs to be documented since the same is true for cuDFs).

tonyfast:
Any ideas why the appveyor ci is getting mad at the sqlite connections? I can't seem to find a solution.

tonyfast:
Also, I'm not sure about the pyenv issue failing on Travis.

tonyfast:
I don't think it is possible to avoid sorting for some ibis backends. what should be done in this case?

philippjfr:
That's fine, we just have to document it and update the test. The cuDF interface has the same problem currently.

dharhas:
Looks like `travis_wait 30 doit develop_install $CHANS_DEV -o $HV_REQUIREMENTS` is failing, If I understand right this is waiting 30 mins and then dying so 30 mins is isn't long enough for this command to execute. Not sure why this command is taking greater than 30 mins though.

philippjfr:
>Looks like travis_wait 30 doit develop_install $CHANS_DEV -o $HV_REQUIREMENTS is failing, If I understand right this is waiting 30 mins and then dying so 30 mins is isn't long enough for this command to execute. Not sure why this command is taking greater than 30 mins though.

I'll rebase, I fixed that one master. Pulling from conda-forge made it very slow.

goanpeca:
>I'll rebase, I fixed that one master. Pulling from conda-forge made it very slow.

Hi @philippjfr, I am looking with @tonyfast into the failing tests. Had to reset a commit.

philippjfr:
I had dropped conda-forge because it was taking forever to solve in the Py2 build. Will see if we can get around that somehow, otherwise let's just increase the timeout. Looks like we're still getting the NoWindowOp errors, still the wrong version of Ibis?

tonyfast:
naw. i'm doing something wrong. i had to add ibis builds from conda-forge cause i couldn't get it from anywhere the other channels. I've got to keep trying things otherwise I'll likely have to duck some of the iloc tests until a new version of ibis shows up.

philippjfr:
>'ll likely have to duck some of the iloc tests until a new version of ibis shows up.

Totally fine, I'd suggest adding a decorator that checks for a specific version:

```
from unittest import skipIf
ibis_skip = skipIf(ibis is None or ibis_version < '...', 'Ibis not available or wrong version')
```

kcpevey:
The sister PR to add rowid support in Ibis has been merged https://github.com/ibis-project/ibis/pull/2345

tonyfast:
im getting timeouts in py27 on travis and appveyer. i extended the wait time on travis, but that still 🤮.  any suggestions about to get around this?

kcpevey:
@philippjfr `ibis-framework` is not available for py 2.7. Are there other backends with this issue? Can we skip those CI tests and document that it's not available in 2.7?

philippjfr:
>Are there other backends with this issue? Can we skip those CI tests and document that it's not available in 2.7?

No other interfaces with that issue but it's really no problem, if a Python2 user can't install ibis then they also can't try to pass an Ibis table to HoloViews. Let's just put a SkipTest check for py2 at the top of the test file and be done with it.

philippjfr:
Btw, I'll probably release 1.13.4 today or tomorrow and then merge this so it can be part of the 1.14.0 release.

philippjfr:
Still haven't released 1.13.4 since there seem to be a few regressions but this PR is firmly in my court so don't feel like you need to keep following up here. Let's aim to get this into a dev release by next week though.

kcpevey:
I see that mask is still unchecked. What is the plan for that? Does it need to be resolved before this is merged?

kcpevey:
@tonyfast Can you do that?

philippjfr:
I've added a mechanism for caching to this PR. I still want to iterate a bit on it but given that you have a demo to build I thought I'd get it in and we can improve it later.

philippjfr:
Okay, instead of performing any kind of automatic caching I've decided to finally implement `Dataset.persist` and `Dataset.compute` methods. We now have sufficient interfaces that support lazy data that this makes sense, e.g. the dask, grid, xarray, and now the ibis interfaces can each meaningfully implement these. Just to be clear the meaning of the two is this:

- `compute`: Convert to an in-memory data format, e.g. dask.DataFrame->pandas.DataFrame or dask.array backed xarray -> numpy.array backed xarray.
- `persist`: Persist the data to memory but keep the data format the same. Since Ibis does not itself have a concept of `persist`, it simply caches the computed data on the `Dataset._cached` attribute.

For Ibis a full persistence implementation would also allow chaining operations on top of an already persisted dataset and simply apply the chained operations on the next persist call. This should be fairly straightforward to implement using the existing `Dataset.pipeline` functionality but for now this implementation simply discards the persisted cache when you chain further operations.

philippjfr:
In theory I'd expect Ibis' `materialize` to be somewhat similar to persisting but in practice I've found it provides basically no speedup.

kcpevey:
@philippjfr Is there anything @ppwadhwa can work on to wrap this up?

philippjfr:
No, I think this is pretty much there now.

holoviews/tests/core/data/testibisinterface.py
```diff
@@ -0,0 +1,290 @@
+import sqlite3
+from unittest import SkipTest
+
+from tempfile import NamedTemporaryFile
+
+try:
```
xmnlab, line 6: on ibis master branch, now all the backends are inside ibis.backends https://github.com/ibis-project/ibis/tree/master/ibis/backends. probably in the next few days the new ibis release will be available.

philippjfr:
Thanks for all your efforts here @kcpevey and @tonyfast! I'm merging now and hoping that we can get some docs in before release.

Issue #4515, opened 2020-07-09
hv.save() doesn't include toolbar in HTML in 1.13.3, but does include it in 1.13.2


Dr-Irv:
#### ALL software version info
holoviews 1.13.3  (by using `conda install holoviews=1.13.3`)

#### Description of expected behavior and the observed behavior

I have a Holoviews graphic that I save to an HTML file (and also end up serving via a server).  With Holoviews 1.13.2, the toolbar is there.  With Holoviews 1.13.3, there is no toolbar, so you can't zoom, reset, save, etc.

I had a code working and hours figuring out why upgrading holoviews caused the problem!  Here is a simple example.

#### Complete, minimal, self-contained example code that reproduces the issue

Create a conda environment as follows:
```
conda create --name hvtest python=3.7
conda install holoviews=1.13.3 numpy
conda activate hvtest
```

Using this code: 
```python
import numpy as np
import holoviews as hv
hv.extension("bokeh")

xs = np.linspace(0, np.pi*4, 40)
fname = "hv" + hv.__version__ + ".html"
g = hv.Text(4,0,fname) + hv.Area((xs, np.sin(xs)))

hv.save(g, fname)
```
it will create the file `hv1.13.3.html`.  When you open that file in a browser, it does not have the toolbar.

But if you then do
```
conda install holoviews=1.13.2
```
and run the code, the file `hv1.13.2.html` _does_ have the toolbar.

#### Stack traceback and/or browser JavaScript console output

Not applicable

#### Screenshots or screencasts of the bug in action
Here are the two HTML files side by side.  The one on the left is from Holoviews 1.13.2.  The one on the right is from Holoviews 1.13.3.

![image](https://user-images.githubusercontent.com/15113894/87089265-2c78e680-c204-11ea-85f0-7cbbbdb07850.png)

drossaop:
I can confirm that I'm seeing this as well after an upgrade to 1.13.3.

philippjfr:
This was intentional but probably should have been reserved for 1.14.0. You can restore the old behavior with:

```python
hv.save(g, fname, toolbar=True)
```

Dr-Irv:
@philippjfr It sounds like the documentation needs to be updated to reflect that!

philippjfr:
>@philippjfr It sounds like the documentation needs to be updated to reflect that!

Absolutely, PR welcome but I've assigned it to the next milestone in any case.

Dr-Irv:
[...]

PR #4518, opened 2020-07-14
make toolbar=True the default in save()

Dr-Irv:
Reference issue #4515 .  Making the toolbar default to `True` to correspond to past behavior.

Documentation issue is that the docs for 1.13.3 are not published on the web site, so presumably if they were to get rebuilt the option would show up there.

### DIFF
--- a/doc/releases.rst
+++ b/doc/releases.rst
@@ -14,6 +14,13 @@ Release notes
 Version 1.13
 ~~~~~~~~~~~~
 
+Version 1.13.4
+**************
+
+Bug fixes:
+* Fix regression in ``save`` method with toolbars
+  (`#4518 <https://github.com/holoviz/holoviews/pull/4518>`_)
+
 Version 1.13.3
 **************
 
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -755,7 +755,7 @@ class extension(_pyviz_extension):
         cls._backend_hooks[backend].append(callback)
 
 
-def save(obj, filename, fmt='auto', backend=None, resources='cdn', toolbar=None, title=None, **kwargs):
+def save(obj, filename, fmt='auto', backend=None, resources='cdn', toolbar=True, title=None, **kwargs):
     """
     Saves the supplied object to file.
 
@@ -811,7 +811,7 @@ def save(obj, filename, fmt='auto', backend=None, resources='cdn', toolbar=None,
                              title=title)
 
 
-def render(obj, backend=None, toolbar=None, **kwargs):
+def render(obj, backend=None, **kwargs):
     """
     Renders the HoloViews object to the corresponding object in the
     specified backend, e.g. a Matplotlib or Bokeh figure.
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -755,7 +755,7 @@ class extension(_pyviz_extension):
         cls._backend_hooks[backend].append(callback)
 
 
-def save(obj, filename, fmt='auto', backend=None, resources='cdn', toolbar=True, title=None, **kwargs):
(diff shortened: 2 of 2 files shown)


philippjfr:
>Documentation issue is that the docs for 1.13.3 are not published on the web site

This is not true (e.g. you can see the release notes for 1.13.3 here: http://holoviews.org/releases.html), I think we just neglected to mention this in the docs.

I also don't think the default should be true as a picture of a toolbar is not useful even if it is useful for docs or other scenarios. I'm sorry about the regression in a micro-release which shouldn't have happened but I'm -1 on reverting to the old behavior. Cc: @jbednar Any opinion?

philippjfr:
Also thanks for the PR, I appreciate the effort even though we disagree on the change.

Dr-Irv:
> This is not true (e.g. you can see the release notes for 1.13.3 here: http://holoviews.org/releases.html), I think we just neglected to mention this in the docs.

See this PR #4519 - if you go to the docs for the `util` module, the `toolbar` option is not shown for `save()`

Dr-Irv:
> I also don't think the default should be true as a picture of a toolbar is not useful even if it is useful for docs or other scenarios. I'm sorry about the regression in a micro-release which shouldn't have happened but I'm -1 on reverting to the old behavior. Cc: @jbednar Any opinion?

But it's NOT just for a picture!  I use it to serve up the HTML that allows a user to interact with the visualization.  The examples that are provided for deploying with Flask and Bokeh use this as well.  I think you should keep the OLD default behavior and if someone needs to have it for a PNG, they can turn the toolbar off.

philippjfr:
Oh sorry in that case I'd consider this a bug. The intended behaviour was to only enable this for PNG exports not for static HTML renders.

Dr-Irv:
> Oh sorry in that case I'd consider this a bug. The intended behaviour was to only enable this for PNG exports not for static HTML renders.

So how do you want to handle?  Accept this PR as is, and create a new issue relative to making the behavior dependent on the type, and a follow-up PR?  Or fix that "bug" in this PR by testing for the 'PNG' format?  IMHO, I think that documenting that behavior could be confusing, so you're better off just having the `toolbar` argument, and if someone has a PNG, they just set `toolbar=False` .

jbednar:
As suggested in #4515, I vote for having the toolbar suppressed by default for PNG, SVG, or other static image formats, and enabled by default for "live" formats like HTML.  I think that's what nearly everyone will want, so that not many people will ever need to end up reading the docs for how to enable the opposite behavior.

Dr-Irv:
> As suggested in #4515, I vote for having the toolbar suppressed by default for PNG, SVG, or other static image formats, and enabled by default for "live" formats like HTML. I think that's what nearly everyone will want, so that not many people will ever need to end up reading the docs for how to enable the opposite behavior.

So I could modify this PR to have that behavior, and then should we remove the `toolbar` parameter to `save()` ?  If not, then I could implement the following behavior:

- `toolbar=None` : default behavior based on the `fmt` parameter
- `toolbar=True` : toolbar appears, independent of `fmt`
- `toolbar=False`: no toolbar, independent of `fmt`

philippjfr:
That seems good, I'd go with something like this for the condition:

```python
if not toolbar and backend == 'bokeh' and (fmt == 'png' or (isinstance(filename, str) and filename.endswith('png'))):
    obj = obj.opts(toolbar=None, backend='bokeh', clone=True)
```

philippjfr:
This also seems like a fairly annoying regression after you've pointed out that this indeed affects all static `save` output so I'd favor a 1.13.4 release sooner rather than later.

philippjfr:
Thanks for being patient with me despite being confused about the actual issue @Dr-Irv and thanks for the PR!

PR #4522, opened 2020-07-17
Remove special instructions for IPython 5.0

hamogu:
The installation section still contained a paragraph on working around a specific issue in the notebook 5.0. That issue was fixed in 5.1 and the notebook is actually on version > 6 now, so it seems that that note is no longer needed.

This particular notebook workaround is confusing and I personally don't like tinkering with the defaults if I can avoid it because I never know how they impact my other projects or how these changes are carried forward in upgrades. Since this issue seems to be relevant only for a very specific notebook version that is now a thing of the past, I suggest to remove that paragraph.
(See links in the removed text to stackoverflow and the related notebook issues that show that the issue was resolved.)

### DIFF
--- a/examples/user_guide/Installing_and_Configuring.ipynb
+++ b/examples/user_guide/Installing_and_Configuring.ipynb
@@ -51,26 +51,6 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "## Jupyter notebook configuration\n",
-    "\n",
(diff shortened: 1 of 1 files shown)


jbednar:
Personally, I left that setting on for years after moving on from version 5.0, and I never noticed it causing problems, but sure, it seems like most people should have a newer version now and we can remove the instructions.

PR #4527, opened 2020-07-22
Hover mode

ahuang11:
https://github.com/holoviz/hvplot/issues/423 I just realized the issue was about `hover='vline'` not `tools=['vline']` but related

ezgif-5-bacfb7339e4a


philippjfr:
Thanks @ahuang11, seems like a reasonable addition.

jlstevens:
Sorry for not chiming in earlier but thanks for the PR! Looks good to me.

Issue #4530, opened 2020-07-23
Proposal: Transform/Derived Stream


jonmmease:
### Overview
As a step towards https://github.com/holoviz/holoviews/issues/4433#issuecomment-633154495, which is itself a step towards HoloViews+Dash support, I propose the introduction of a new stream type, tentatively called `Transform`.

### Motivation
For Dash support, the core feature that is needed is the ability to convert a dynamic stateful HoloViews object (e.g. layout of Dynamic Maps as produced by `datashade` and `link_selections`) into a pure function that inputs values that are usually provided by streams, and outputs a simple (non-dynamic) HoloViews Element.

To build this function, it will be necessary to traverse the dynamic HoloViews object (in a manner similar to `link_selections`) and extract the function portion of `DynamicMap`s. Then it will be necessary to look at the stream inputs to the `DynamicMap`s and replace them with the function signature.

For Dash's use, it's important to be able to trace streams all the way back to the interactive streams that are produced by plot interactions.  One important case where this is not currently practical is the `SelectionExpr` stream.

This stream wraps an element that supports selections. It registers a callback with the element it wraps, and then in that callback is calls `self.event(...)` to push the change to observers of the `SelectionExpr` stream.

The problem is that when traversing a `DynamicMap` that inputs a `SelectionExpr` stream, it's not possible to get references to the interaction streams that actually trigger the `SelectionExpr` to update (e.g. the `BoundsXY` stream of a `Scatter` element).

### Proposal
I propose the addition of a new `Transform` stream type that consists of:
 1. A collection of input streams
 2. Specification of the `param` values that the stream produces
 3. A pure function that inputs the stream values of the input streams (1) and returns the stream values of the stream itself (2).

The `SelectionExpr` stream can then be re-implemented as a `Transform` stream subclass. This would make it possible to traverse through the `SelectionExpr` stream to the interactive stream values.

I expect this would be a useful in other contexts as well.  We may want a different name since the base `Stream` class already has a `transform` method for subclasses to use to rename and update the values of the streams params computed by the parent class.

If this generally sounds like a good idea, I'd like to work on implementing this soon. We can talk about, and iterate on, API details once I get a draft PR together.

PR #4532, opened 2020-07-24
Add new Derived stream class

jonmmease:
closes https://github.com/holoviz/holoviews/issues/4530

This PR introduces a new `Derived` stream class that generalizes the concept of having a `Stream` that produces values based on one or more other input streams.  

The `SelectionExpr` stream had done this in a somewhat ad-hoc way, but with this PR `SelectionExpr` is now a subclass of the new `Derived` stream.

The existing `SelectionExpr` tests pass with minimal updates. I also added a new set of tests for `Derived` streams directly in `holoviews/tests/teststreams.py`.  This is a good place to look at for a simple example of what a `Derived` stream subclass looks like.

Let me know if you have any questions or concerns about the approach or API. Thanks!

### DIFF
--- a/holoviews/tests/teststreams.py
+++ b/holoviews/tests/teststreams.py
@@ -880,16 +880,16 @@ class TestBufferDataFrameStream(ComparisonTestCase):
 class Sum(Derived):
     v = param.Number()
 
-    def __init__(self, val_streams, base=0):
+    def __init__(self, val_streams, exclusive=False, base=0):
         self.base = base
-        super(Sum, self).__init__(input_streams=val_streams)
+        super(Sum, self).__init__(input_streams=val_streams, exclusive=exclusive)
 
     @property
     def constants(self):
         return dict(base=self.base)
 
     @classmethod
-    def transform_function(cls, stream_values, trigger_index, constants):
+    def transform_function(cls, stream_values, constants):
         v = sum([val["v"] for val in stream_values if val["v"]])
         return dict(v=v + constants['base'])
 
@@ -908,17 +908,14 @@ class TestDerivedStream(ComparisonTestCase):
 
         # Check outputs
         self.assertEqual(s0.v, 3.0)
-        self.assertIsNone(s0.trigger_index)
 
         # Update v0
         v0.event(v=7.0)
         self.assertEqual(s0.v, 9.0)
-        self.assertEqual(s0.trigger_index, 0)
 
         # Update v1
         v1.event(v=-8.0)
         self.assertEqual(s0.v, -1.0)
-        self.assertEqual(s0.trigger_index, 1)
 
     def test_nested_derived_stream(self):
         v0 = Val(v=1.0)
@@ -935,12 +932,10 @@ class TestDerivedStream(ComparisonTestCase):
         # Update top-level value
         v2.event(v=8.0)
         self.assertEqual(s0.v, 13.0)
-        self.assertEqual(s0.trigger_index, 1)
 
         # Update nested value
         v1.event(v=5.0,)
         self.assertEqual(s0.v, 14.0)
-        self.assertEqual(s0.trigger_index, 0)
 
     def test_derived_stream_constants(self):
         v0 = Val(v=1.0)
@@ -956,7 +951,27 @@ class TestDerivedStream(ComparisonTestCase):
         # Update value
         v2.event(v=8.0)
         self.assertEqual(s0.v, 113.0)
-        self.assertEqual(s0.trigger_index, 2)
+
+    def test_exclusive_derived_stream(self):
+        # Define input streams
+        v0 = Val()
+        v1 = Val(v=2.0)
+
+        # Build exclusive Sum stream
+        # In this case, all streams except the most recently updated will be reset on
+        # update
+        s0 = Sum([v0, v1], exclusive=True)
+
+        # Check outputs
+        self.assertEqual(s0.v, 2.0)
+
+        # Update v0
+        v0.event(v=7.0)
+        self.assertEqual(s0.v, 7.0)
+
+        # Update v1
+        v1.event(v=-8.0)
+        self.assertEqual(s0.v, -8.0)
 
 
 class TestExprSelectionStream(ComparisonTestCase):
--- a/holoviews/tests/teststreams.py
+++ b/holoviews/tests/teststreams.py
@@ -878,7 +878,7 @@ class TestBufferDataFrameStream(ComparisonTestCase):
 
 
 class Sum(Derived):
-    v = param.Number()
+    v = param.Number(constant=True)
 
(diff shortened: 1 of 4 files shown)


holoviews/tests/teststreams.py
```diff
@@ -889,14 +971,14 @@ def test_selection_expr_stream_scatter_points(self):
             expr_stream = SelectionExpr(element)
 
             # Check stream properties
-            self.assertEqual(len(expr_stream._source_streams), 2)
-            self.assertIsInstance(expr_stream._source_streams[0], SelectionXY)
-            self.assertIsInstance(expr_stream._source_streams[1], Lasso)
+            self.assertEqual(len(expr_stream.input_streams), 2)
```
jbednar, line 974: Is there a particular reason to switch from `source` to `input` here?  Does that help make the terminology more consistent?

jonmmease: Here was my thinking. For the `SelectionExpr` stream, it happened to be the case that all of the input streams were always attached to a single `source` element.  The `Derived` stream doesn't assume this, all it knows is that it should calculate a value based on a list of streams that come from somewhere else.  So I guess it seemed to me that using "source" here was no longer accurate.

jbednar: Thanks; makes sense to me!

jonmmease:
The AppVeyor tests are failing due to a bunch a bunch of failures in `spatialpandas` that I think are pyarrow 1.0 incompatibilities.  Though it looks like spatialpandas pins the pyarrow dependency to `<1` (https://github.com/holoviz/spatialpandas/blob/master/setup.py#L27), so I'm not sure how this is happening.

Travis tests aren't even being run I don't think:
```
$ pip install pyctdev && doit miniconda_install && pip uninstall -y doit pyctdev

pyenv: version `3.7' is not installed (set by PYENV_VERSION environment variable)

The command "pip install pyctdev && doit miniconda_install && pip uninstall -y doit pyctdev" failed and exited with 1 during .
```
Does this mean anything to anyone else?

jbednar:
spatialpandas master pins to <1.0 to avoid those issues, but it hasn't been released yet, so you're presumably still getting an older spatialpandas here. Pinning pyarrow here seems reasonable to me; @jlstevens was looking at releasing spatialpandas.

jbednar:
Oh, and this particular immediate issue is presumably that Travis is providing 3.6 or 3.8, while the rest of the setup is assuming 3.7.  That happens every time Travis updates their default images, I think, and we have to adapt. I forget the details, but again @jlstevens will presumably remember.

philippjfr:
The pyenv stuff can be fixed by switching back to `language: python` in the travis.yml.

philippjfr:
No objection to pinning pyarrow in the test dependencies in setup.py for now.

jonmmease:
I went ahead and got spatialpandas working with pyarrow 1.0 in https://github.com/holoviz/spatialpandas/pull/46.  So hopefully this will take care of the issue here once that package is released.

philippjfr:
Looking good, I'll review and hopefully merge this evening.

holoviews/tests/teststreams.py
philippjfr: Makes sense to me too.

jonmmease:
Actually, I'm working on the `uncollate` transform and I may need to revisit this a little. The `trigger_index` idea doesn't really fit with the stateless goal, so I may need to think on how to remove this.

As you have time, I'd appreciate feedback on https://github.com/holoviz/holoviews/pull/4534 and I'll update this PR once I've gotten further with `uncollate`.  Thanks!

jonmmease:
Ok, I removed `trigger_index` and added an `exclusive` option to `Derived` that, when set, will clear all streams except the most recently updated stream.

jlstevens:
I know this is WIP and I am only raising a naming issue, but I don't think `uncollate` is a real word. :-)

The antonym I probably liked the most from [this list](https://www.thesaurus.com/browse/collate) is probably `separate` but I would like to hear any other suggestions!

jbednar:
It may not be the best word, but it's a word people use: https://www.support.xerox.com/en-us/article/en/x_wc7245_en-O12395 

I think it's useful to have the visual semantic link to `collate`, so that they can be understood as duals. As long as that's true (i.e. that `collate(uncollate(x)) = x`, at least roughly), a different name would need to be quite compelling before it would unseat `uncollate`.  Conversely, if it's not really the dual of `collate`, then `uncollate` is confusing and shouldn't be used...

jonmmease:
> I think it's useful to have the visual semantic link to collate, so that they can be understood as duals. As long as that's true (i.e. that collate(uncollate(x)) = x, at least roughly), a different name would need to be quite compelling before it would unseat uncollate

Yeah, I believe these should be very nearly duals. I'll document more when that PR is ready. 

@jlstevens, just to clarify, `uncollate` doesn't show up in this PR. I'll have another PR soon that introduces it. Thanks!

philippjfr:
@jonmmease Could you rebase both PRs? CI is working again on master.

jonmmease:
Updated with master, and things have gotten further!

In AppVeyor, seeing this error
```
  File "C:\Miniconda36-x64\envs\test\lib\distutils\version.py", line 337, in _cmp
    if self.version < other.version:
TypeError: '<' not supported between instances of 'str' and 'int'
```

with this in the stack trace:
```
C:\Miniconda36-x64\envs\test\lib\site-packages\panel\pane\holoviews.py in _render(self, doc, comm, root)
    278 
    279         kwargs = {'margin': self.margin}
--> 280         if backend == 'bokeh' or LooseVersion(str(hv.__version__)) >= str('1.13.0'):
    281             kwargs['doc'] = doc
    282             kwargs['root'] = root
C:\Miniconda36-x64\envs\test\lib\distutils\version.py in __ge__(self, other)
```

Based on https://bugs.python.org/issue14894, looks like this is an issue with comparing version numbers with different number parts that mix numbers and letters. e.g. this also raises the error `LooseVersion("22.7-r1") < LooseVersion("22.7.3-r1")`.

philippjfr:
Let's fix appveyor separately, as long as Travis passes I'm happy to merge. Only a very minor issue there:

```
ERROR: holoviews.tests.teststreams.test_all_stream_parameters_constant
  File "/home/travis/miniconda/envs/test-environment/lib/python3.6/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/home/travis/build/holoviz/holoviews/holoviews/tests/teststreams.py", line 28, in test_all_stream_parameters_constant
    % (name, stream_cls.__name__))
TypeError: Parameter v of stream Sum not declared constant
```

jonmmease:
ping @philippjfr. Travis is green now. Thanks for taking the time to look this over!

philippjfr:
Thank you for working on this!

Issue #4433, opened 2020-05-23
Session stateless workflow for HoloViews


jonmmease:
This issue is the result of an offline discussion on what it would take to make it possible to serve interactive HoloViews objects in Plotly Dash applications.

The translation of Dash callback data (e.g. from Zoom or Box Selection events) into the inputs to HoloViews interactive streams (e.g. `RangeXY` and `BoundsXY`) is relatively straightforward.

The more complex problem is the mismatch in state models between HoloViews and Dash.  

### HoloViews state model
The current HoloViews architecture is optimized for responsiveness.

Conceptually, you can decompose a rendered HoloViews object into:
  1. A plot object in the backend library (Plotly or Bokeh)
  2. A collection of interactive streams. The inputs to these streams come from callbacks on the plot object for things like zoom and selection events.

The relatively expensive operation of transforming a HoloViews object into a plot object is performed only once. After this initial "render", the plot object is efficiently mutated in response to updates to the stream values.

### Dash state model
Dash's architecture is optimized for horizontal scalability.

Conceptually, you can decompose a Dash application into:
 1. A tree of components (e.g. dropdowns, sliders, figures, etc.), where each component exposes a set of properties (e.g. the slider value).
 2. A collection of pure functions that each input a collection of component property values and output a different collection of component property values.

This approach is more scalable than the current HoloViews architecture because the server does not need to store the current session state for every simultaneous user of the app. It is less responsive than the current HoloViews architecture because the server can't make small efficient updates to the prior state of a figure in response to a callback. Instead, it must fully reconstruct the figure in response to changes in the property values that the figure depends on.

### Questions for discussion
 1. Is the current state model a fundamental characteristic of HoloViews, or is it (at least conceptually) an implementation detail?
 2. If it's an implementation detail, what would it take to support a "session stateless" paradigm as well?

My intuition is that the current state model is an implementation detail, but one that is pretty fully baked in due to the fact that this aligns with the Bokeh+Panel state model.

### Potential paradigm.
As I've thought more on this, I've come up with one potential approach.  I'm not sure if it could be made efficient enough to be practical, but I don't think it require a major change to how HoloViews works internally.

Suppose we perform the relatively expensive transformation from HoloViews element to plot+streams once during application initialization.  Could we develop an efficient clone operation, one that would clone both the plot object and the associated streams (wiring up the cloned streams to reference the cloned plot object)? 

[...]

PR #4534, opened 2020-07-25
Support positional stream args in DynamicMap callback

jonmmease:
This PR is a step toward supporting #4433 using a `DynamicMap` transformation.  The idea is described in https://github.com/holoviz/holoviews/issues/4433#issuecomment-663715053.

## Background
The callback function for a `DynamicMap` is currently called by passing the kdim values as positional arguments, followed by stream parameter values as keyword arguments.  For example:

```python
import holoviews as hv
XY = Stream.define('XY', x=0,y=0)
xy_stream = XY()

def callback_fn(i_value, y, x):
    ...

dmap = hv.DynamicMap(callback_fn, kdims=["i"], streams=[xy_stream])
```
Here the `i_value` argument is the `i` key dimension that is called positionally (named `i_value` to emphasize that it's called by position not by name), and the `x` and `y` arguments are the stream parameters and they are called as keyword arguments.  So here the `x` and `y` argument names must match the param names of the `xy_stream` stream.

This approach is convenient when writing callback functions by hand, when there are only a few stream parameters and the stream parameter names don't conflict. But it's less convenience when building a callback function programmatically that will input many streams with potentially overlapping parameter names. This will be the situation when building the transformation in #4433.

## Description
This PR adds a `positional_stream_args` parameter to the `DynamicMap` class that, when set to `True`, will configure the `DynamicMap` to pass stream values to the callback function positionally.  The stream arguments will follow the kdim arguments (if any), the value of each stream argument is a `dict` containing key-value pairs for each param in the stream, and the order of the stream arguments is determined by the order of the streams in the `streams` parameter list.

Here's an example of what this looks like:

```python
import holoviews as hv
XY = Stream.define('XY', x=0,y=0)
xy_stream0 = XY()
xy_stream1 = XY()

def callback_fn(i_value, s0, s1):
    x0, y0 = s0['x'], s0['y']
    x1, y1 = s1['x'], s1['y']
    ...

dmap = hv.DynamicMap(
    callback_fn, kdims=["i"], streams=[xy_stream0, yx_stream1], positional_stream_args=True
)
```

When `positional_stream_args=True` there is no possibility of stream parameter name clashes.

Of course, the default is `False` to match the current behavior.

Does this sound reasonable @jlstevens @philippjfr ?

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -873,6 +873,14 @@ class DynamicMap(HoloMap):
        cache where the least recently used item is overwritten once
        the cache is full.""")
 
+    positional_stream_args = param.Boolean(default=False, constant=True, doc="""
+       If False, stream parameters are passed to the callback as keyword arguments.
+       If True, stream parameters are passed to callback as positional arguments.
+       Each positional argument is a dict containing the contents of a stream.
+       The positional stream arguments follow the positional arguments for each kdim,
+       and they are ordered to match the order of the DynamicMap's streams list. 
+    """)
+
     def __init__(self, callback, initial_items=None, streams=None, **params):
         streams = (streams or [])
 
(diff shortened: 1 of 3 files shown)


philippjfr:
The changes look fine to me, but I'll try to fix the tests before merging.

jonmmease:
Ok, updated with master and tests are passing. Anything else come to mind for this PR @philippjfr?

jonmmease:
@philippjfr alright if we merge this today?

philippjfr:
Yes, looks good to me.

jonmmease:
Thanks a lot @philippjfr!

PR #4537, opened 2020-07-28
fix bug when the default value of Dimension is 0

### LINKED ISSUE #4536: DynamicMap doesn't use Dimension's default value if the value is 0
Holoviews 1.13.3

Here is the code:

```python
import holoviews as hv
hv.extension('bokeh')

def f(a, b):
    x = [0, 1]
    y = [a, b]
    return hv.Curve((x, y))

dm = hv.DynamicMap(f, kdims=[
    hv.Dimension('a', range=(-3.0, 3.0), default=0.0),
    hv.Dimension('b', range=(-3.0, 3.0), default=0.0)
])

dm
```

the output plot is:

image

the value of a and b is 0, but the y axis of the line is -3. It uses the start value of the Dimension's range.

ruoyu0088:
fix #4536

Use the default value of Dimension if it is not None.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -951,7 +951,7 @@ class DynamicMap(HoloMap):
         for kdim in self.kdims:
             if str(kdim) in stream_params:
                 key.append(None)
-            elif kdim.default:
+            elif kdim.default is not None:
                 key.append(kdim.default)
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks, much appreciated!

PR #4544, opened 2020-08-08
Ensure operations do not recursively accumulate pipelines

philippjfr:
Major, major bug causing Operations to recursively accumulate operations in a pipeline which would progressively degrade anything that would then replay that pipeline (e.g. linked selections). On many linked_selections examples this provides a 2x speedup on initial rendering and then a progressively increasing speedup over the buggy version every time the pipeline is replayed, e.g. for a datashaded plot with linked selections applied you would get 10x slowdowns after just a few seconds of zooming around.

### DIFF
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -129,13 +129,20 @@ class Operation(param.ParameterizedFunction):
 
 
         element_pipeline = getattr(element, '_pipeline', None)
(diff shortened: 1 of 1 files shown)


jbednar:
I added emojis to visualize my combination of joy that you found this and sadness that it existed. :-)

PR #4547, opened 2020-08-09
Add ability to listen to dataset linked_selection

philippjfr:
Linked selections are powerful and can easily be applied to HoloViews elements. However in many cases we may want to subscribe to changes in a linked selection from outside of HoloViews. This PR makes this possible by providing a new method that returns a instance parameter which will contain a reference to the current selection applied to a dataset. By combining this with the internal linked_selection cache we can make sure that if set up correctly this does not result in additional computation. As an example we can do something like this:

```python
@pn.depends(ls.selection_param(df))
def count(selected_df):
    return f'## {len(selected_df)}/{len(df)} selected'
```

Here we create a callback that depends on the current selection of the df and reports the fraction of datapoints that have been selected.

### DIFF
--- a/holoviews/selection.py
+++ b/holoviews/selection.py
@@ -5,12 +5,13 @@ import param
 
 from param.parameterized import bothmethod
 
+from .core.data import Dataset
 from .core.dimension import OrderedDict
 from .core.element import Element, Layout
 from .core.options import CallbackError, Store
 from .core.overlay import NdOverlay, Overlay
 from .core.spaces import GridSpace
-from .streams import SelectionExpr, PlotReset, Stream, SelectMode
+from .streams import SelectionExpr, Pipe, PlotReset, Stream, SelectMode
 from .util import DynamicMap
 from .util.transform import dim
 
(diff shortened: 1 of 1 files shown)


jbednar:
Looks good! This seems very convenient.  Are there cases where one might want to depend on the selection itself (e.g. to apply it to some other data somehow?), rather than on the selected data as here?

philippjfr:
>Are there cases where one might want to depend on the selection itself (e.g. to apply it to some other data somehow?), rather than on the selected data as here?

Sure but you can already do that with `@param.depends(ls.param.selection_expr)` but there may be a usecase for depending on the boolean mask, e.g. to compute a count that'll be more efficient than actually applying the selection and returning the data.

jonmmease:
It might not cause I problem, but the usage of `cache={}` function arguments throughout this PR make be a bit nervous and I was wondering if you had thought through it further than I have.

If the default empty dict value is chosen, and then modified in the function by adding to the cache, then the this default argument will be modified for all future use of the function.  It might not cause a real issue, but I'd be worried about ending up with the cached values stored in default function argument values.

philippjfr:
>If the default empty dict value is chosen, and then modified in the function by adding to the cache, then the this default argument will be modified for all future use of the function. It might not cause a real issue, but I'd be worried about ending up with the cached values stored in default function argument values.

I'd be happy with the safer thing of using a default of `None` and then skipping the caching (or simply using `cache = {} if cache is None else cache`). I think all uses of the method do pass in an explicit cache in right now though.

PR #4551, opened 2020-08-14
Add decollate transformation

jonmmease:
This PR closes https://github.com/holoviz/holoviews/issues/4433.

The goal of this issue was to develop a workflow for using dynamic HoloViews objects in a stateless multi-threaded context like Dash.  The approach that I proposed was to develop a transformation of an arbitrary dynamic HoloViews object into a pure function that inputs all of the required stream values as input argument and returns a non-dynamic HoloViews object.

It turns out that the transformation needed here is the inverse of the current `collate` transformation.  Per the docstring:
```
        Collation allows unpacking DynamicMaps which return Layout,
        NdLayout or GridSpace objects into a single such object
        containing DynamicMaps. Assumes that the items in the layout
        or grid that is returned do not change.
```
While not described in the docstring, collate also processes the streams of the input DynamicMap and pushes them down to the individual DynamicMaps of the resulting container according to the `dmap.callback.stream_mapping` property.

This PR introduces an ~`uncollate`~ `decollate` transformation that is the inverse of `DynamicMap.collate`. Here is the docstring description of ~`Layout.uncollate`~ `Layout.decollate`:
```
        Decollation allows packing a Layout of DynamicMaps into a single DynamicMap
        that returns a Layout of simple (non-dynamic) elements. All nested streams are
        lifted to the resulting DynamicMap, and are available in the `streams`
        property.  The `callback` property of the resulting DynamicMap is a pure,
        stateless function of the stream values. To avoid stream parameter name
        conflicts, the resulting DynamicMap is configured with
        positional_stream_args=True, and the callback function accepts stream values
        as positional dict arguments.
```

Decollation populates the `dmap.callback.stream_mapping` property so that top-level streams can be associated to the correct subplot by the interactive plotting backends. It also lifts key dimensions of the nested objects to the resulting DynamicMap.  

## Example:
Here is a full non-trivial example. Create a two plot layout where:
 - First subplot is a DynamicMap with two kdims and no streams
 - Second subplot is a DynamicMap with one kdim, wrapped in datashade and spread operations.

```python
import holoviews as hv
from holoviews.operation.datashader import datashade, spread

import numpy as np
hv.extension('bokeh')

phases = [0, 0.25, 0.5, 0.75, 1.0]

def sine_curve(phase, freq):
    return hv.Scatter((xvals, [np.sin(phase+freq*x) for x in xvals]))

def cos_curve(phase):
    freq = 0.5
    return hv.Scatter((xvals, [np.cos(phase+freq*x) for x in xvals]))

dmap0 = hv.DynamicMap(sine_curve, kdims=['phase', 'frequency']).redim.range(phase=(0, 3), frequency=(0.5, 1.25))
dmap1 = hv.DynamicMap(cos_curve, kdims=['phase']).redim.range(phase=(0, 3))

layout = dmap0 + spread(datashade(dmap1))
layout
```
Screenshot_20200814_125707

```python
print(layout)
```
```
:Layout
   .DynamicMap.I  :DynamicMap   [phase,frequency]
      :Scatter   [x]   (y)
   .DynamicMap.II :DynamicMap   [phase]
      :RGB   [x,y]   (R,G,B,A)
```

Now decollate layout. The resulting object looks and behaves the same way as the input. It has interactive streams for `PlotSize` and `RangeXY` so that datashader zoom/pan interactions work as they do on the input.
```python
layout_decollated = layout.decollate()
layout_decollated
```
Screenshot_20200814_125707

If we print the structure of `layout_decollated` we see that is is very different from the input. The object is a DynamicMap that returns a `Layout` of non-dynamic elements. Also note that the union of all of the key-dimensions are granted to the resulting DynamicMap.

```python
print(layout_decollated)
```
```
:DynamicMap   [phase,frequency]
   :Layout
      .Scatter.I :Scatter   [x]   (y)
      .RGB.I     :RGB   [x,y]   (R,G,B,A)
```

The resulting `DynamicMap` has streams corresponding to those introduced by the `datashade` operation on the input subplot.
```python
print(layout_decollated.streams)
```
```
[PlotSize(height=245,scale=1.0,width=238), RangeXY(x_range=None,y_range=None)]
```

## Other inputs
`decollate` can also be run on `Overlay`, `GridSpace`, and `NDOverlay` containers containing DynamicMaps.

## Implementation notes
The implementation of `decollate` involves recursively traversing the input object and building a symbolic expression tree for it. streams and kdims are extracted from the expression tree and replaced by the kdim-index and stream-index respectively.  The callback for the resulting `DynamicMap` inputs the kdim and stream values and uses these to evaluate the expression tree.

[Greenspun's tenth rule](https://en.wikipedia.org/wiki/Greenspun%27s_tenth_rule) constantly came to mind while working on this :smiley:

### DIFF
--- a/holoviews/core/decollate.py
+++ b/holoviews/core/decollate.py
@@ -140,7 +140,7 @@ def to_expr_extract_streams(
             else:
                 # Add new stream
                 stream_index = StreamIndex(index=len(streams))
-                cloned_stream = type(hvobj)(**hvobj.contents)
+                cloned_stream = hvobj.clone()
                 original_streams.append(hvobj)
                 streams.append(cloned_stream)
                 if container_key is not None:
--- a/holoviews/core/decollate.py
+++ b/holoviews/core/decollate.py
@@ -229,9 +229,9 @@ def expr_to_fn_of_stream_contents(expr, nkdims):
     return expr_fn
 
 
-def uncollate(hvobj):
+def decollate(hvobj):
     """
-    Uncollate transforms a potentially nested dynamic HoloViews object into single
+    Decollate transforms a potentially nested dynamic HoloViews object into single
     DynamicMap that returns a non-dynamic HoloViews object. All nested streams in the
     input object are copied and attached to the resulting DynamicMap.
 
--- a/holoviews/core/decollate.py
+++ b/holoviews/core/decollate.py
@@ -100,22 +100,6 @@ def to_expr_extract_streams(
             fn = hvobj.callback.callable
             args.extend(kdim_args)
 
-        # Process streams
-        def get_stream_index(stream):
-            # Compute stream index
-            if stream in original_streams:
-                # Reuse index to existing stream
-                stream_index = StreamIndex(index=original_streams.index(stream))
-            else:
-                # Add new stream
-                stream_index = StreamIndex(index=len(streams))
-                cloned_stream = type(stream)(**stream.contents)
-                original_streams.append(stream)
-                streams.append(cloned_stream)
-                if container_key is not None:
-                    stream_mapping.setdefault(container_key, []).append(cloned_stream)
-            return stream_index
-
         for dm_stream in dm_streams:
             stream_arg = to_expr_extract_streams(
                 dm_stream, kdims, streams,  original_streams,
--- a/holoviews/core/decollate.py
+++ b/holoviews/core/decollate.py
@@ -9,7 +9,7 @@ from ..streams import Stream, Derived
 
 from ..plotting.util import initialize_dynamic
 
-HoloViewsExpr = namedtuple("HoloviewsExpr", ["fn", "args", "kwargs"])
+Expr = namedtuple("HoloviewsExpr", ["fn", "args", "kwargs"])
 StreamIndex = namedtuple("StreamIndex", ["index"])
 KDimIndex = namedtuple("KDim", ["index"])
 
@@ -127,9 +127,9 @@ def to_expr_extract_streams(
                 kwargs.append(stream_arg)
 
         if expand_kwargs:
-            expr = HoloViewsExpr(fn, args, kwargs)
+            expr = Expr(fn, args, kwargs)
         else:
-            expr = HoloViewsExpr(fn, args, [{"kwargs": HoloViewsExpr(dict, [], kwargs)}])
+            expr = Expr(fn, args, [{"kwargs": Expr(dict, [], kwargs)}])
         return expr
 
     elif isinstance(hvobj, Stream):
@@ -144,7 +144,7 @@ def to_expr_extract_streams(
                     )
                 )
             constants = hvobj.constants
-            return HoloViewsExpr(
+            return Expr(
                 stream_arg_fn, [stream_indexes, constants], []
             )
         else:
@@ -188,7 +188,7 @@ def to_expr_extract_streams(
             data_expr = ViewableTree._process_items(data_expr)
 
         kwargs = [{"data": data_expr}]
-        return HoloViewsExpr(fn, args, kwargs)
+        return Expr(fn, args, kwargs)
     elif isinstance(hvobj, Element):
         return hvobj.clone(link=False)
     else:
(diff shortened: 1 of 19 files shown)


jonmmease:
Renamed to decollate.  This should be ready for review.

jlstevens:
Thanks! I'll take a look shortly.

holoviews/tests/teststreams.py
```diff
@@ -893,7 +893,9 @@ def transform_function(cls, stream_values, constants):
         v = sum([val["v"] for val in stream_values if val["v"]])
         return dict(v=v + constants['base'])
 
-Val = Stream.define("Val", v=0.0)
+
+class Val(Stream):
+    v = param.Number(constant=True)
```
jlstevens, line 898: I'm fine with this either way, but shouldn't the `Stream.define` version be equivalent?

holoviews/core/decollate.py
```diff
@@ -0,0 +1,255 @@
...
+from .. import (
+    Layout, DynamicMap, Element, Callable, Overlay, GridSpace, NdOverlay, HoloMap
+)
+from . import ViewableTree
+from collections import namedtuple
+from ..streams import Stream, Derived
+
+from ..plotting.util import initialize_dynamic
+
+HoloViewsExpr = namedtuple("HoloviewsExpr", ["fn", "args", "kwargs"])
```
jlstevens, line 12: Given that everything here is HoloViews, why not just name it `Expr`? :-)

holoviews/core/decollate.py
```diff
@@ -0,0 +1,255 @@
...
+StreamIndex = namedtuple("StreamIndex", ["index"])
+KDimIndex = namedtuple("KDim", ["index"])
+
+
+def to_expr_extract_streams(
```
jlstevens, line 17: Bit of a monster function but that seems unavoidable given what it is trying to do. As it is simply building an expression, it might be nice to have one or two tests of just this function (which can also help make what it does more concrete).

jlstevens:
Looks good to me! My comments are fairly minor and I am happy to merge if @philippjfr agrees. The only thing I really would like considered is renaming `HoloViewsExpr` to just `Expr`...

holoviews/core/decollate.py
jonmmease: sure!

holoviews/tests/teststreams.py
jonmmease: The reason I changed that was that I was cloning streams using:

```python
cloned_stream = type(stream)(**stream.contents)
```

And this wasn't working for streams build with `Stream.define`. Is there a better way to clone streams? Maybe we should look at adding a `Stream.clone` method.

jonmmease: Actually, this does work. Not sure what I was running into during development.  In https://github.com/holoviz/holoviews/pull/4551/commits/711f082e81b13e17824e054c55ac994b71047156 I reverted some of these to use `Stream.define` so that there is a mix of both in the tests.

holoviews/core/decollate.py
jonmmease: Done in https://github.com/holoviz/holoviews/pull/4551/commits/1e0aeb67d938e7342ef7bca3f948833f4c5cdd4e

jonmmease:
Thanks @jlstevens, I renamed `HoloViewsExpr` to `Expr` but I'll wait for @philippjfr input on whether to add some lower-level tests for the `to_expr_extract_streams` function itself.  I think the functionality is covered pretty well by the `decollate` tests, but I'm happy to add some lower level tests as well.

jlstevens:
That was just a suggestion and not something I am insisting on: happy to see this merged as is!

jonmmease:
Thanks @jlstevens!

Got some weird CI errors, so I merged in master and pushing again :crossed_fingers:

jonmmease:
Not sure what's going on with the tests. But the same things are failing here as on master (https://github.com/holoviz/holoviews/commit/1a366bcb0440a8fdb50e583f4acefa83dcc87d31).

jlstevens:
Looks like a simple flake error: I've pushed a commit that I think should fix the tests.

philippjfr:
Python 2 builds are still timing out but that's a separate issue. Going to merge.

jlstevens:
Great to see this merged!

@philippjfr I was just looking at this and it didn't look like a time out issue to me: https://travis-ci.org/github/holoviz/holoviews/jobs/721832764#L1752 (`KeyError: u'glyph_renderer'` in the `test_active_tools_draw_stream` unit test)

philippjfr:
That's a transient error. Never quite figured out what was causing it.

jonmmease:
Thanks!

PR #4554, opened 2020-08-15
Add History stream

jonmmease:
## Overview
This PR adds a `History` stream type. `History` stream instances wrap another stream and collect a history of the values taken on by that stream. This list of states is as the `values` property

## Motivation
A follow-on PR will refactor `link_selections` to be based on `Derived` (https://github.com/holoviz/holoviews/pull/4532) and `History` streams, making it possible for `uncollate` (See https://github.com/holoviz/holoviews/pull/4551) to extract the linked streams from result of `link_selections`.

But, this is a simple and generally useful stream type, so I wanted to split it out into its own PR.

## Example Usage
```python
Val = Stream.define("Val", v=0.0)
history = History(val)
prihnt(history.contents)
```
```
{"values": [{"v": 1.0}]}
```
```python
# Perform a few updates on val stream
val.event(v=2.0)
val.event(v=3.0)
print(history.contents)
```
```
{"values": [{"v": 1.0}, {"v": 2.0}, {"v": 3.0}]}
```
```python
history.clear_history()
print(history.contents)
```
```
{"values": []}
```

### DIFF
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -855,6 +855,41 @@ class Derived(Stream):
         self._unregister_input_streams()
 
 
+class History(Stream):
+    """
+    A Stream that maintains a history of the values of a single input stream
+    """
+    values = param.List(constant=True, doc="""
+        List containing the historical values of the input stream""")
+
+    def __init__(self, input_stream, **params):
+        super(History, self).__init__(**params)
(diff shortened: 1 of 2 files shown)


jonmmease:
It's ready on my end. Thanks @jlstevens

Issue #4356, opened 2020-04-02
legend_opts similar to colorbar_opts?


ahuang11:
I wanted to get a transparent legend, but I had to use hooks. Not too big of a deal, but I was wondering how feasible/easy it would be to implement something like colorbar_opts for legend_opts.

```
import holoviews as hv
hv.extension('bokeh')

def _transparent_legend(plot, element):
    p = plot.state
    p.legend.background_fill_alpha = 0

(hv.Curve([0, 1, 2], label='a') * hv.Curve([5, 6, 7], label='b')).opts(
    hooks=[_transparent_legend])
```

Desired:
```
import holoviews as hv
hv.extension('bokeh')

(hv.Curve([0, 1, 2], label='a') * hv.Curve([5, 6, 7], label='b')).opts(
    legend_opts={'background_fill_alpha': 0})
```

philippjfr:
This was merged.

PR #4558, opened 2020-08-17
legend_opts implemented

ahuang11:


ahuang11:
https://github.com/holoviz/holoviews/issues/4356

PR #4562, opened 2020-08-19
Updated spreading operation to support aggregate arrays

jlstevens:
Minor changes to the spreading operation in order to support aggregate arrays (https://github.com/holoviz/datashader/pull/771). This change does have some important implications:

1. This won't be compatible with datashader releases prior to the PR references above. I could do a version check to turn `None` into an operator accepted by earlier datashader versions (maybe even use 'source'?)
2. The value of `None` maps to 'over' for RGBs and 'add' for arrays in datashader. This breaks backwards compatibility as the current default is 'source'. 

If I implement the version check idea to support older datashader, then the incompatibility would be for newer datashader versions only (where aggregate spreading is supported). I don't know if that makes things more or less confusing overall...

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1511,10 +1511,11 @@ class SpreadingOperation(LinkableOperation):
     to make sparse plots more visible.
     """
 
-    how = param.ObjectSelector(default='source',
-            objects=['source', 'over', 'saturate', 'add'], doc="""
+    how = param.ObjectSelector(default=None,
+            objects=[None, 'source', 'over', 'saturate', 'add', 'max', 'min'], doc="""
         The name of the compositing operator to use when combining
-        pixels.""")
+        pixels. Default of None uses 'over' operator for RGB elements
+        and 'add' operator for aggregate arrays.""")
 
     shape = param.ObjectSelector(default='circle', objects=['circle', 'square'],
                                  doc="""


jlstevens:
The tests are failing due to the use of `None` so I'll map it back to 'source' for now for previous datashader versions.

holoviews/operation/datashader.py
```diff
@@ -1584,7 +1585,9 @@ class spread(SpreadingOperation):
         Number of pixels to spread on all sides.""")
 
     def _apply_spreading(self, array):
-        return tf.spread(array, px=self.p.px, how=self.p.how, shape=self.p.shape)
+        replace_none_how = ds_version <= '0.11.1' and (self.p.how is None)
+        how = 'source' if replace_none_how else self.p.how
```
jbednar, line 1589: That's some pretty convoluted logic. What about:

```
def _calculate_how_from_default(how):
    """Handle changes in values accepted for `how` across Datashader versions"""
    if ds_version <= '0.11.1' and (how is None):
         how = 'source'
    return how

how = _calculate_how_from_default(self.p.how)
```

jlstevens: Personally, I would rather not introduce such a trivial function. Happy to let @philippjfr be the tie-breaker...

philippjfr: Don't want to just change the default depending on the version?

jlstevens: That would also work and might be simpler. I'll try that idea, thanks!

jlstevens:
@philippjfr @jbednar Are we happy with breaking backwards compatibility for RGBs for newer datashader versions? Essentially the operator 'source' will be switched for 'over'...

jbednar:
I think so, because anyone who notices an issue should be told to switch to agg-based spreading anyway.  Neither `source` nor `over` on RGB will work as well as `add` on aggs, so it should prompt people to choose the better approach if they notice an issue, and if they don't notice an issue (much more likely, as `source` and `over` work the same for isolated points anyway), then there's no problem in the first place. So I'm happy to see this merged as is.

Issue #3500, opened 2019-02-16
Changing default HoloViews colormap


ahuang11:
Since the default cmap's highest value is white, it's hard to tell if the white values are the max or NaN without hover (and most colormaps used do not use gray?)

![image](https://user-images.githubusercontent.com/15331990/52905688-749fff00-31f2-11e9-84f3-567251b3b045.png)

poplarShift:
+1, this makes sense to me. Either that or change the default cmap?

jbednar:
I thought we already had an issue about various problems with the default colormap, but I was unable to find it by searching for "colormap" or "default" or "hot".  From what I remember, there are many reasons to change HoloViews' default "hot" colormap, given that the default page color is typically white for Jupyter notebooks:

1. "hot" is highly nonuniform perceptually as described at [colorcet.pyviz.org](colorcet.pyviz.org).
2. The peak value of the colormap is white, making it impossible to distinguish a NaN area from a peak value
3. The highest data values are those most similar to the page color, when for intuitive understanding of a sequentially ordered set of numeric values the highest values should be maximally different from the page color. (That way, isolated patches will be more visible the higher their value, but currently they have _less_ contrast as their values increase.)
4. (Less crucial) The lowest value in the colormap is black, which can be ambiguous with other data elements in the plot that default to black.

Possible improvements:

- It has been proposed to use "fire" instead, which addresses issue 1, but it still has the other three problems. To illustrate problems 2 and 3, with "fire" on a white background you get sharp perceptual discontinuities between low values and NaN values; it only really works on a black background:
![image](https://user-images.githubusercontent.com/1695496/53128464-7f2c0280-352a-11e9-813b-99fe6ad5a912.png)
![image](https://user-images.githubusercontent.com/1695496/53128484-88b56a80-352a-11e9-9a98-9648ea5399a1.png)
- The `linear_kry_5_98_c75` colormap in [colorcet](http://colorcet.pyviz.org/user_guide/index.html) is a version of fire that doesn't top out at white, avoiding problem 2, but it still has problems 3 and 4. 
- A reversed version of `linear_kry_5_98_c75` would avoid problems 1, 2, and 3, but it still has 4 (now at the top end) and also is a bit odd, as it uses colors from black-body radiation (like "hot" and "fire") but in an unintuitive order
[...]

Issue #2487, opened 2018-03-27
transparent option for rasterize?


rsignell-usgs:
When I use `datashade` I get the transparent background that lets me see the WMTS tiles

![2018-03-27_7-57-13](https://user-images.githubusercontent.com/1872600/37970747-3b013d0a-31a2-11e8-8da4-a8375a1ca6d7.png)

but when I use `rasterize`, I get grey instead of transparent 
![2018-03-27_7-57-58](https://user-images.githubusercontent.com/1872600/37970752-3e095686-31a2-11e8-9c29-8e0d49317530.png)

and I can't figure out how to fix it.  

I did try to figure this out by googling and looking in docs and code, but failed... :disappointed:

ocefpaf:
Same here, is there an option to set the gray to transparent?

jbednar:
`datashade` is just a shorthand for `rasterize` followed by `shade`.  `shade` uses Datashader to do the colormapping from within Python (server side), while the default is to use Bokeh to do the colormapping in JavaScript (client side).  So you can presumably make it work by doing `shade(rasterize(...))`, but I suppose you're asking for the Bokeh/JS-based colormapping to work the same as the datashader/Python colormapping, which I agree seems appropriate. I believe that doing so wasn't previously possible due to a problem with Bokeh's handling of special color values, but now that's been fixed in Bokeh (https://github.com/bokeh/bokeh/pull/6756) maybe we need to declare that "transparent" is the color to use by default for NaN? @philippjfr?

philippjfr:
A transparent background is actually the default when using GeoViews, i.e. if you were to switch to ``gv.TriMesh`` on the input to rasterize it would return a ``gv.Image`` (as long as you are using master of hv and gv), which uses a transparent color for NaN values. This was a deliberate choice because when plotting non-geographic data you often do want to see the nan values, but I'd be happy to have this discussion again.

To override the default NaN color to be transparent you can set ``clipping_colors={'NaN': (0, 0, 0, 0)}``. This will be covered properly in the new "Styling Plots" user guide being added in https://github.com/ioam/holoviews/pull/2483.

jbednar:
I would think that the default should be transparent whether or not using GeoViews; is there a reason it shouldn't be?

ocefpaf:
> This was a deliberate choice because when plotting non-geographic data you often do want to see the nan values

Makes sense.

[...]

PR #4567, opened 2020-08-25
Implement cnorm option

philippjfr:
Implements `cnorm` option to switch between 'linear', 'log' and 'eqhist'.

Screen Shot 2020-08-25 at 4 09 15 PM

- [x] Add tests
- [x] Update docs
- [x] Handle clashes with `logz` option


jlstevens:
Summarizing a discussion between @jbednar @philippjfr and myself:

- We can (eventually) deprecate `clipping_colors` with options for minimum, maximum and nan colors.
- These have clear semantics but won't ever apply when auto-ranging as only with explicit ranges can you be above/below the highest/lowest value (really the semantics of 'over' and 'under' are clearer than 'max' and 'min')
- For rasterized output, specifically with the 'count' aggregator we can set the dimension range to `(1, None)` where we have defined `1` as the lowest true count because `0` really is missing data (this is datashader semantics).
- We can set default options for image/raster types in HoloViews such that values under/over the declared range are transparent (instead of extending the colormap which is Bokeh's default behavior). 
- Note, we wouldn't set these defaults for any other elements (which would continue to follow Bokeh semantics) and we already set the nan color to transparent for image types in HoloViews (overriding Bokeh's default which is gray).
- The naming is up for debate with suggestions including 1) following bokeh's `low_color`, `high_color` and `nan_color` 2) `max_color`, `min_color` and `nan_color` or 3) `cunder`,  `cover` and `cnan`.
- Personally I would like some variant of the third option where the concept of being above/below the max/min is clear with the `c` prefix to group with the other colormap options like `clim`, `cmap`, `cformatter` etc. Another variant of this might be `cbelow`, `cabove` and `cnan`.
- The value of `None` would mean (follow the defaults) and we would keep the string 'transparent' for transparent pixels. That way, someone who wants the Bokeh default behavior for an image could use `image.opts(cbelow=None)` or similar.

Open issue:

* One problem we have is that 'fire' is our default datashader colormap which has a white peak value. If you have transparent pixels, you will show the plot background which is typically white. This looks quite horrible! Two options would be to 1) set the bgcolor to something other than white for images (not a great solution!) or 2) use a different colormap than fire which we have discussed in the past but would break backwards compatibility.

Other than option naming and the issue with the 'fire' colormap, I think this sounds like a good plan!

jbednar:
Sounds good. Yes, I think there's already an open issue for dealing with `fire`; it's really not a suitable colormap for the default white-background pages, particularly not when missing values are possible.  I also prefer a variant of option 3 for the naming, because "low"/"high" and "max"/"min" are ambiguous or misleading; the specified behavior does not occur _at_ the low or min value, it occurs below or under the min value.  So names like `below` or `under` are more accurate and easier to reason about; it's clearer that you're defining what happens to values _below_ or _under_ the range being colormapped, not what happens _at_ that lower bound. Given that matplotlib uses `over` and `under` for this purpose, I'd favor those terms, but note that unfortunately `cover` is going to be read as "cover" (cuvver, not "c"-over), by any new person. So I guess I slightly favor "cbelow" and "cabove", ugly as those are.

jlstevens:
Looking for alternatives, the only other one I'll throw out there is `chigher` and `clower`...

jlstevens:
One issue where colormaps is discussed is https://github.com/holoviz/holoviews/issues/2487. Here is the relevant bit:

> image

jbednar:
Sure, I have no preference between chigher/clower and cabove/cbelow.  Yes, https://github.com/holoviz/holoviews/issues/2487#issuecomment-376626529 is the prior issue about colormaps, which remains unaddressed.

philippjfr:
You have the strongest opinions here so would you mind coming up with a proposal for the colormap changes? Then we can veto if we hate it 😆

jbednar:
No, I just want to complain! :-) Sigh.  I'll try to think of something suitable, but please ping me if I haven't done so in a reasonable time.

jlstevens:
I thought a bit about this issue and here is the best suggestion I have right now:

1. Pick a colormap that is suitable for use with transparency and make it available in HoloViews.
2. Add an `hv.config` option that when set, enables this new colormap as the default for the datashader shading operations and for Images (off by default). We could also target the Images affected using the group/label mechanism...
3. Have a page describing the issue that we can be linked to and update the pages where we need the better colormap to use `hv.config` to switch to the new cmap.
4. Mention this option in the release notes and state that we will switch away from fire sometime in future.
5. Eventually switch away from fire by toggling the `hv.config` default.

Don't love any of this but that seems the best approach to me right now...

holoviews/plotting/bokeh/__init__.py
```diff
@@ -191,6 +191,8 @@ def colormap_generator(palette):
 
 # Rasters
 options.Image = Options('style', cmap=dflt_cmap)
+options.Image = Options('plot', clipping_colors={'min': 'transparent',
+                                                 'max': 'transparent'})
```
jbednar, line 195: Why is 'max' transparent?  I don't think we want to avoid seeing the maximum pixels; seems like those should be in another color sometimes (e.g. red for a colormap not including red), but it's hard to think of a use for them being transparent.

jlstevens: I think we discussed this already: we want anything out of the colormap bounds to show up with symmetry at both ends i.e mix and max. I am happy to change it again if there is a good reason for max to be treated different from min...

jlstevens:
After discussion with @jbednar, @philippjfr we came to an agreement that `clipping_colors` is currently being abused for the purpose of making zero counts transparent, resulting in needless discussion about the min and max clipping colors.

The new suggestion is to remove setting a dimension range at all and to be explicit and clear about what we want to express: a special value that indicates missing/no data. This isn't needed for floats which naturally have a no data value (NaN) but is needed for integer data which has no such special value. Zero counts mean no data for the datashader count aggregator but there is [plenty of data in the wild](https://geozoneblog.wordpress.com/2013/02/26/troubles-tiff-no-data-values-floating-point/) that use special 'nodata' values such as -32767 or -9999. 

Therefore, if you *need* transparency for a count aggregate after using `rasterize`, you would typically do something like:

```
rasterize(element).opts(cnorm='eqhist', nodata=0)
```

I think this is clear, explicit with no more mysterious implicit behavior or abuse of clipping options.

jbednar:
Sounds good! We'll have to decide what to about integers vs. floats for `nodata`, given that float comparison is unreliable. I think there are two plausible approaches: (1) don't support floats as `nodata` values, and silently skip applying `nodata` to float arrays, or (2) accept floats as `nodata` values and apply `nodata` to float arrays, but in our docs warn that only some floats are likely to be useful (including NaN, which we'd have to check for explicitly, and up to 7-digit integers; see https://en.wikipedia.org/wiki/Single-precision_floating-point_format).

I can see arguments for either approach. For (1), if it can be achieved, only applying the `nodata` translation to integer arrays would mean that users (or we) could set this option to 0 and never mess with it. Not sure if `rasterize` sets a unique HoloViews group and label by default, but if it does, seems like the option could be enabled quite specifically for Datashader output and almost never interfere with anything else (just signed ints, if those occur). But it would require detecting integer arrays and handling them specially.

For (2), I can easily imagine data originally integer showing up in workflows as floats, and given that short integers work fine as floats, it may be more convenient to simply ignore whether it's float or not and accept e.g. `nodata=-9999` or  `nodata=-9999.0` and not make any distinction.

I don't think I have much preference, except to _not_ prefer option (3) of not allowing floats while also not selectively applying `nodata` to integers, which seems like a restriction without a benefit.

jlstevens:
I think there is an option (that makes sense to me) that isn't 1) or 2):

3) Declare the plot option `nodata = param.Integer(None, allow_None=True)` so only integer values can be specified. Then whether we apply it to float arrays is orthogonal to that i.e `nodata=1` is always allowed but `nodata=1.1` isn't. Given that I don't see any reason *not* to apply it to float arrays, assuming the float/int comparison is sensible.

I have no opinion on that but `nodata = param.Integer(None, allow_None=True)` as the plot option declaration is what makes sense to me.

jbednar:
My only opinion about your option 3 is that it doesn't let people declare the obvious `nodata = np.nan`; instead we have to tell people that `nodata = None` means use `np.nan` for floats and nothing for integers, and that there's no way to turn off `np.nan` handling?

jlstevens:
That is right - I'm of the opinion that as float have a dedicated NaN value, there is no need for `nodata` for floats (the semantics is clearly defined without needing magic values). Any thoughts @philippjfr ?

jlstevens:
@jbednar @philippjfr Regarding the default datashader colormap issue, why not just use the blueish colormap that is already the default for `hvplot`? If we are breaking backwards compatibility of colormaps in HoloViews, might as well make it consistent with something else.

Unless the default for hvplot is also problematic?

philippjfr:
For reference here's the colormap in various scenarios:

bokeh_plot - 2020-09-21T122532 514

bokeh_plot - 2020-09-21T122556 128

bokeh_plot - 2020-09-21T122559 228

jlstevens:
What about some raster with transparency overlaid on a tile source?

jbednar:
Jean-Luc and I discussed this issue today and came to the conclusion that any changes to the colormap are independent of this PR in its current state. Now that we have decided that rasterize won't generate transparent Images for the default count aggregation, then most of the colormap problems go away, because fire is already a suitable colormap for cases with no transparency. With the default `nodata=None`, the plot is actually independent of the page background color, and so it can continue to use fire; ambiguity only happens for `nodata=0`, which a user is only likely to add when they want to overlay the plot onto something like a web tile server, and many of those may already be dark enough that fire is again appropriate.  

To summarize, with the defaults (nodata=None, cmap=Fire) rasterize should work well (no ambiguity of peak color vs. background showing through, no sharp edge between white background and weak data points).  The defaults will _not_ allow overlaying the results on a web tile source or other data, but the user must explicitly enable `nodata=0` in that case and can at the same time select a suitable colormap for whatever they are overlaying on, if it's not something dark.

jbednar:
Note that there will still be issues for non-uint aggregations (not Count or Any), where NaN will automatically be transparent and a fire colormap may be inappropriate on a white background. Even so, in that case users will very often want to select their own colormap anyway, e.g. a diverging one if values can be both positive and negative, as they can be for many non-Count aggregations.  So I'm not particularly concerned about what may need to be adjusted once people are using a non-uint aggregation; once they mess with agg then they will also often need to set cmap appropriately.

jlstevens:
Discussing this again with @jbednar, we decided that there is probably no need to change the default colormap right now: before the `nodata` option (I'll be pushing some commits soon) the idea was transparency for missing values by default. This would have made the need for a better colormap more urgent, but now that `nodata` will need to be supplied only in cases where it is needed (namely integer data where you are overlaying on top of something) this is no longer the case.

We also agreed that *not* using transparency unless requested (or made explicit with NaNs for the float case) actually makes things more predictable as you cannot reason about the background color people use (ofter white, but many people also like dark themes).

jlstevens:
One thing to point out about `nodata`: it is a shortcut to replace a value with `NaN` which means you get the following behavior with hover:

image

image

Ideally, the hover would show the original value (i.e 0) but Bokeh doesn't support transparency masks afaik. This also means in the float case, the `nodata` values map to `NaN` which will show up in hover the same way as the original `NaN` values (if any).

jlstevens:
I suppose for the integer case, you could smooth over the hover issue with a custom hover tool that knows that `NaN` really is the `nodata` value. If we did this, it might be a reason to disallow `nodata` for the float case?

philippjfr:
>One thing to point out about nodata: it is a shortcut to replace a value with NaN which means you get the following behavior with hover:

You could use a custom hover formatter here.

jlstevens:
> You could use a custom hover formatter here.

That is exactly what I was suggesting in my previous comment. Do you think we *should*?

philippjfr:
Ah sorry, missed that. It's not a huge amount of effort so seems worth doing.

holoviews/plotting/mpl/raster.py
```diff
@@ -14,6 +14,9 @@
 
 class RasterBasePlot(ElementPlot):
 
+    nodata = param.Integer(default=None,
+                           doc="Missing data (NaN) value for integer data", allow_None=True)
```
jbednar, line 18:
```suggestion
    nodata = param.Integer(default=None, doc="""
        Missing data (NaN) value for integer data""")
```

(no n

holoviews/plotting/mpl/raster.py
```diff
@@ -82,6 +85,12 @@ def get_data(self, element, ranges, style):
         style['extent'] = [l, r, b, t]
         style['origin'] = 'upper'
 
+        plot_opts = element.opts.get('plot', 'matplotlib')
+        nodata = plot_opts.kwargs.get('nodata')
+        if nodata is not None:
+            data = data if (data.dtype.kind  == 'f') else data.astype(np.float64)
+            data[data == nodata] = np.NaN
```
jbednar, line 92: I thought you weren't going to apply `nodata` to float arrays? Or are you just not going to suggest applying it?

holoviews/plotting/mpl/raster.py
```diff
@@ -126,6 +135,9 @@ def update_handles(self, key, axis, element, ranges, style):
 
 class QuadMeshPlot(ColorbarPlot):
 
+    nodata = param.Integer(default=None,
+                           doc="Missing data (NaN) value for integer data", allow_None=True)
```
jbednar, line 139:
```suggestion
    nodata = param.Integer(default=None, doc=""
        Missing data (NaN) value for integer data""")
```"

File: holoviews/plotting/mpl/raster.py [hunk 5]
```diff
@@ -141,6 +153,13 @@ class QuadMeshPlot(ColorbarPlot):
     def get_data(self, element, ranges, style):
         zdata = element.dimension_values(2, flat=False)
         data = np.ma.array(zdata, mask=np.logical_not(np.isfinite(zdata)))
+
+        plot_opts = element.opts.get('plot', 'matplotlib')
+        nodata = plot_opts.kwargs.get('nodata')
+        if nodata is not None:
+            data = data if (data.dtype.kind  == 'f') else data.astype(np.float64)
+            data[data == nodata] = np.NaN
```
jbednar, line 161: Same as above; docstring indicates that `nodata` applies only to integer data, but here it's applied to all data.

holoviews/plotting/mpl/raster.py
jlstevens: We decided to simply ignore floats. Implemented in f146bd3

holoviews/plotting/mpl/raster.py
jlstevens: Resolved in f146bd3

jlstevens:
I removed the dimension range declaration in ef56e50 as I think that will get more unit tests passing. These range declaration are not *wrong* but I'm wondering if they are worth introducing in this PR if that means having to update tests when these declarations are no longer needed for the immediate goals of using `rasterize` easily...

holoviews/plotting/bokeh/raster.py
```diff
@@ -15,6 +15,9 @@
 
 class RasterPlot(ColorbarPlot):
 
+    nodata = param.Integer(default=None, doc="""
+        Missing data (NaN) value for integer data""")
```
jbednar, line 19:
```suggestion
        Optional missing-data value for integer data.
        If non-None, data with this value will be replaced with NaN so that it is transparent when plotted.""")
```

holoviews/plotting/mpl/element.py
```diff
@@ -683,6 +685,9 @@ class ColorbarPlot(ElementPlot):
         Number of discrete colors to use when colormapping or a set of color
         intervals defining the range of values to map each color to.""")
 
+    cnorm = param.ObjectSelector(default='linear', objects=['linear', 'log', 'eqhist'], doc="""
+        Color normalization applied during colormapping.""")
```
jbednar, line 689:
```suggestion
        Color normalization to be applied during colormapping.""")
```

holoviews/plotting/bokeh/element.py
```diff
@@ -1634,17 +1634,20 @@ class ColorbarPlot(ElementPlot):
...
-       User-specified colorbar axis range limits for the plot, as a tuple (low,high).
-       If specified, takes precedence over data and dimension ranges.""")
+        User-specified colorbar axis range limits for the plot, as a tuple (low,high).
+        If specified, takes precedence over data and dimension ranges.""")
 
-    cformatter = param.ClassSelector(
-        default=None, class_=(util.basestring, TickFormatter, FunctionType), doc="""
-        Formatter for ticks along the colorbar axis.""")
+    cnorm = param.ObjectSelector(default='linear', objects=['linear', 'log', 'eqhist'], doc="""
+        Color normalization applied during colormapping.""")
```
jbednar, line 1650:
```suggestion
        Color normalization to be applied during colormapping.""")
```

holoviews/plotting/mpl/raster.py
```diff
@@ -14,6 +14,9 @@
 
 class RasterBasePlot(ElementPlot):
 
+    nodata = param.Integer(default=None, doc="""
+        Missing data (NaN) value for integer data""")
```
jbednar, line 18:
```suggestion
        Optional missing-data value for integer data.
        If non-None, data with this value will be replaced with NaN so that it is transparent when plotted.""")
```

holoviews/plotting/mpl/raster.py
```diff
@@ -126,6 +135,9 @@ def update_handles(self, key, axis, element, ranges, style):
 
 class QuadMeshPlot(ColorbarPlot):
 
+    nodata = param.Integer(default=None, doc="""
+        Missing data (NaN) value for integer data""")
```
jbednar, line 139:
```suggestion
        Optional missing-data value for integer data.
        If non-None, data with this value will be replaced with NaN so that it is transparent when plotted.""")
```

holoviews/plotting/plotly/raster.py
```diff
@@ -13,6 +13,9 @@ class RasterPlot(ColorbarPlot):
 
     padding = param.ClassSelector(default=0, class_=(int, float, tuple))
 
+    nodata = param.Integer(default=None, doc="""
+        Missing data (NaN) value for integer data""")
```
jbednar, line 17:
```suggestion
        Optional missing-data value for integer data.
        If non-None, data with this value will be replaced with NaN so that it is transparent when plotted.""")
```

holoviews/plotting/plotly/raster.py
```diff
@@ -99,6 +109,9 @@ def get_data(self, element, ranges, style):
 
 class QuadMeshPlot(RasterPlot):
 
+    nodata = param.Integer(default=None, doc="""
+        Missing data (NaN) value for integer data""")
```
jbednar, line 113:
```suggestion
        Optional missing-data value for integer data.
        If non-None, data with this value will be replaced with NaN so that it is transparent when plotted.""")
```

holoviews/plotting/plotly/raster.py
```diff
@@ -39,6 +42,13 @@ def get_data(self, element, ranges, style):
         if self.invert_axes:
             x0, y0, dx, dy = y0, x0, dy, dx
             array = array.T
+
+        plot_opts = element.opts.get('plot', 'plotly')
+        nodata = plot_opts.kwargs.get('nodata')
+        if nodata is not None and (array.dtype.kind  == 'i'):
+            array = array.astype(np.float64)
+            array[array == nodata] = np.NaN
```
jbednar, line 50: This chunk of code occurs so many times in this PR that it really seems to call out for centralizing it somewhere:


```
def apply_nodata(opts, data):
    "Replace `nodata` value in data with NaN, if specified in opts"
    nodata = opts.kwargs.get('nodata')
    if nodata is not None and (data.dtype.kind  == 'i'):
        data = data.astype(np.float64)
        data[data == nodata] = np.NaN
        return data

array = apply_nodata(element.opts.get('plot', 'plotly'), array)
```

holoviews/plotting/bokeh/element.py
```diff
@@ -1634,17 +1634,20 @@ class ColorbarPlot(ElementPlot):
         Number of discrete colors to use when colormapping or a set of color
         intervals defining the range of values to map each color to.""")
 
+    cformatter = param.ClassSelector(
+        default=None, class_=(util.basestring, TickFormatter, FunctionType), doc="""
+        Formatter for ticks along the colorbar axis.""")
+
     clabel = param.String(default=None, doc="""
         An explicit override of the color bar label, if set takes precedence
```
jbednar, line 1642:
```suggestion
        An explicit override of the color bar label. If set, takes precedence
```

jbednar (approved):
Looks good after minor suggestions, thanks!  I only skimmed EqHistNormalize, which I think has been evaluated already.

holoviews/plotting/bokeh/element.py
```diff
@@ -1634,17 +1634,20 @@ class ColorbarPlot(ElementPlot):
...
     clim = param.Tuple(default=(np.nan, np.nan), length=2, doc="""
-       User-specified colorbar axis range limits for the plot, as a tuple (low,high).
-       If specified, takes precedence over data and dimension ranges.""")
+        User-specified colorbar axis range limits for the plot, as a tuple (low,high).
+        If specified, takes precedence over data and dimension ranges.""")
 
-    cformatter = param.ClassSelector(
-        default=None, class_=(util.basestring, TickFormatter, FunctionType), doc="""
-        Formatter for ticks along the colorbar axis.""")
+    cnorm = param.ObjectSelector(default='linear', objects=['linear', 'log', 'eqhist'], doc="""
```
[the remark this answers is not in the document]
jlstevens: I half agree...it is an annoying extra character!

holoviews/plotting/bokeh/raster.py
jlstevens: Add this suggestion to the batch even though it isn't actually true and will need updating again: see my previous screenshot with `'NaN':'green'` where `NaN` *isn't* transparent.

holoviews/plotting/plotly/raster.py
jlstevens: I considered this but I don't know how helpful it really is: there isn't that much duplication *per backend*. Each time this occurs, you go from 5 lines (where it is very easy to see what it does) to one line where you would need to look up what something like `apply_nodata(element.opts.get('plot', 'plotly'), array)` means, Note there is also a `return data` missing at the end of the suggested function...

Not strongly opposed so I am happy to factor out a utility if @philippjfr thinks it makes sense...

jlstevens:
I already wrote this in a comment (which is now hidden):

All the docstrings will need updating again: see my previous screenshot shows  `'NaN':'green'` where `NaN` isn't transparent. 
Maybe something like `it is transparent (by default) when plotted` instead of `it is transparent when plotted` which isn't necessarily true!

holoviews/plotting/bokeh/element.py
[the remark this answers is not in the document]
jlstevens: Alternatively put, is it worth making the specification longer for HoloViews and hvPlot (the way we want to do things going forward) due to the history of the thing we are moving away from (server side eqhist within datashader)?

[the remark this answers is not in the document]
jbednar: Pros for `eq_hist`:
- Matches datashader, so that in e.g. hvPlot and HoloViews docs we can say `eq_hist` consistently to refer to the equivalent operation whether it's done by `shade` or by Bokeh.
- Gives a hint to pronunciation and meaning. In a list with "linear" and "log", the default pronuciation of "eqhist" would be something like "eckhist", which would just be weird. `eq_hist` makes it clearer that it's an abbreviation of two words, and then people can guess or at least more easily remember what those two words are.

Pros for `eqhist`:
- One character shorter.

Seems like a clear choice to me :-), but I'll let @philippjfr cast the tiebreaking vote.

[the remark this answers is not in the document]
philippjfr: +1 for `eq_hist`

holoviews/plotting/plotly/raster.py
jbednar: Happy to let Philipp be the tiebreaker here as well, but I disagree that it's _very_ easy to see what it does.  At first glance it's a chunk of code that my eyes skip over, not a semantically meaningful action whose intent is obvious when skimming. `apply_nodata` is clearer to me in its intent, even in skimming.  If you study the code line by line, yes, it's easy enough to follow, but that does take work! And then if one wanted to update this code (e.g. if we decided it should apply to floats, or to change the float type), it's very likely that would happen only on one backend or element type, in part because the code isn't identical in the many different spots it appears, due to the name of `data` changing everywhere.  

Anyway, smells bad to me as it is, but @philippjfr can make the call.

philippjfr: +1 for factoring it out to `plotting/util.py`

jbednar:
> Maybe something like 'it is transparent (by default) when plotted' instead of 'it is transparent when plotted' which isn't necessarily true!

Sure, as you wish.  I didn't want to make it too verbose, and on its own it's accurate because transparency is the default, but if you think that's clearer, do add '(by default)'.

holoviews/plotting/bokeh/element.py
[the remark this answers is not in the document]
jlstevens: Last suggestion before I give up and use `eq_hist`: why not just `eq`? Where it is `eq_hist` or `eq`, a newcomer needs to look up what it means regardless.

That would mean 'equalized' where the fact it is 'histogram equalization' isn't particularly important (which would also suggest the name should be `hist_eq`!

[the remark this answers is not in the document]
jbednar: Such a struggle over a few characters! 🙂 'eq' seems unsearchable, so I don't prefer that, plus it doesn't match datashader.

holoviews/plotting/plotly/raster.py
jlstevens: Addressed in 2c1e1e0

holoviews/plotting/bokeh/element.py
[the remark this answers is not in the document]
jlstevens: I renamed 'eqhist' to 'eq_hist' in 204ac6c. I don't like that string much but I'm happy for this to be merged as long as you have considered possible alternatives.

jbednar (approved):
Looks great, once tests pass!

jlstevens:
Rebased to get the Travis tests green leaving only the Appveyor tests with errors that don't look related to this PR. I would like to see this merged now and Appveyor address separately..

jlstevens:
I've fixed the merge conflict though that may break the Travis tests that were passing in the last commit. I recommend merging now...

philippjfr:
This PR still needs tests and a section about cnorm in the docs, probably in the style mapping user guide.

jlstevens:
If merging this PR isn't holding anything up for @jbednar (wrt the datashader docs) then yes, I can add tests/docs here. Otherwise I am happy to add those in a follow up PR.

jbednar:
Please do add them here; it's not holding me up.

jlstevens:
Here is the example I came up with for the style mapping user guide:

image

I think this demonstrates the modes though it is a shame that the colorbar labels for `eq_hist` are so cramped (unavoidable really). Note that I also update the `logz` description as follows:

`logz: Enable logarithmic color scale (Will be deprecated in future in favor of cnorm='log')`

I hope that is true!

jlstevens:
In addition, I think this definition is useful to reproduce the result of `shade` exactly:

```python
import matplotlib as mpl
from matplotlib import cm
shade_cmap = mpl.colors.LinearSegmentedColormap.from_list("shade_cmap", ["lightblue", "darkblue"])
```

image


Without the colorbar for an undistorted comparison:

image

This means the minimal *exact* reproduction of `datashade` with `rasterize` is now:

```
.opts(cnorm='eq_hist', nodata=0, cmap=shade_cmap)
```

jlstevens:
@philippjfr @jbednar I've added unit tests and docs, updating the style mapping and large data user guides.

Other than deciding whether to include `shade_cmap` (which I am in favor of), this PR is ready for review/merge.

holoviews/plotting/bokeh/raster.py
```diff
@@ -201,6 +208,11 @@ def get_data(self, element, ranges, style):
 
 class QuadMeshPlot(ColorbarPlot):
 
+    nodata = param.Integer(default=None, doc="""
```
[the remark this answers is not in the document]
jlstevens: First time I've heard about this rule but sure, I can do that.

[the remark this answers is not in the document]
philippjfr: Too late, already done it :)

jbednar:
I've pushed some updates to the user guide, but I don't think they go far enough. There is now a section stating that rasterize is preferred, but we still use datashade in the initial examples! We should reserve datashade just for the cases where it makes sense, which currently is really just for categorical color mixing.  I can do some more work on it later, but have to sign off for today.

jlstevens:
Summarizing the discussion I had with @jbednar about trying to minimize the effort needed by users to apply datashader to their elements.

### Ongoing Issues

1. Running `rasterize` with the default aggregator (i.e 'count') results in `Image` elements with a `'Count'` vdim regardless of the input element (e.g could be `Points` with a vdim, `Path` without etc). This means that you will get axiswise normalization in a layout across elements where it doesn't make sense (e.g. a rasterized `Points` with a `Path`). This is a pitfall which might not have a technical solution but should at least be documented. Previously this wasn't an issue as datashaded `RGBs` couldn't be normalized together anyway.

2. The 'fire' colormap is particularly problematic after rasterizing sparse elements (e.g `Points`) where you need to use `nodata`. In particular, this colormap has white in it which generally causes problems with the typical white background of plots. However for `Image` types which are genuine raster data (and not mostly empty), 'fire' isn' a bad choice.

3. Assuming the colormap default is okay, you still need to set the `cnorm` and `nodata` options. We have already discussed this and I still think being explicit about `cnorm` is a good thing. Whether `nodata` should be explicitly specified is less clear and it might be nice to have it automatically applied when it makes sense (i.e for count aggregates in particular).

### Possible solutions/improvements

1. Change the default colormap e.g to `kbc_r` to match `hvplot` . Pros: simple solution, improves consistency with `hvplot`. Cons: Breaks backwards compatibility, 'fire' is actually fine for `Image` elements.

2. Apply styling to the return values from `rasterize`. Pros: easy to implement. Cons: this is something we want to avoid with operations that should be about the semantics of the data, not about the aesthetics of how it is displayed.

3. If `Dimension` has a `nodata` parameter, then `rasterize` could set that, avoiding the need for the plot option after rasterizing. Pros: simplifies the options needed Cons: needs to be implemented but otherwise seems like a good idea to me!

4. One new suggestion is we could have a trivial subclass of `Image` called `Rasterized` which can then be given whatever default options we want. The `rasterize` operation would simply return this element which is simply an image by another name (and rasterized data *does* have somewhat different semantics than `Image` as the data is then explicitly a raster approximation to something else). Doesn't seem entirely crazy to me and might solve quite a few of our issues here. Pros: increasing flexibility Cons: new element type to document.

### Discussion points

* I still think `cnorm='eq_hist'` should be explicitly specified by the user. Colorbars are not on by default (rightly so in my opinion) and implicit histogram equalization can be very misleading. For this reason I don't really want the output of `rasterize` to look exactly like the output of `datashade` by default.

Any other thoughts @jbednar, @philippjfr ?

philippjfr:
Only logging on for a bit, just wanted to chime in and say that I'm strongly against adding a new element. There's no semantic difference between an Image and whatever the new element would be so it would simply be introduced as a crutch to allow changing the defaults. While that may seem okay the backwards compatibility breakage is huge and will introduce massively non-obvious issues to other peoples code because something like `.opts(opts.Image(cmap='viridis'))` would no longer match. So I'd actually prefer option 2. over 4.

One other option I'd mention is that we could consider allowing `Store.transfer_options` on operations, enable that option for rasterize, **and** change the default colormap for points/scatter (where 'fire' is a really bad idea). This would ensure that rasterized Points/Scatter specifically inherit the defaults from the image that's being rasterized.

Additionally I'd also be happy to implement option 3 this week. I don't think it'll be particularly difficult.

jlstevens:
> ...  and will introduce massively non-obvious issues to other peoples code because something like `.opts(opts.Image(cmap='viridis'))`. 

I think this convinces me that it is a bad idea though I do think that an image output by rasterization does have some extra semantics. Though probably not enough to justify a new element.

> One other option I'd mention is that we could consider allowing Store.transfer_options on operations, enable that option for rasterize, and change the default colormap for points/scatter (where 'fire' is a really bad idea).

Grepping suggests that `transfer_options` is only used as a `Compositor` parameter right now. I'm not entirely sure about using it for operations especially here - I can't tell if it does all the necessary filtering to map options between different element types (the docstring suggests it is for transferring options for all backends between objects, but presumably of the same type).

That said, transferring `cmap` *specifically* might make sense and is worth considering as an alternative.

philippjfr:
>(the docstring suggests it is for transferring options for all backends between objects, but presumably of the same type).

It does filter and transfers between element types. It's primarily used to transfer options on things like Bivariate to the Contours that are spit out after the compositor transform is applied.

jlstevens:
Thanks for clarifying.

I will say that operations that try to *preserve* options from input to outputs seem less objectionable to me than ones that try to set things. The `transfer_options` suggestion does have some appealing properties:

1. Doesn't affect normal `Image` elements where 'fire' is ok i.e fewer unnecessary backwards compatibility breaking changes.
2. Would fix the cases where 'fire' is *not* a good choice.
3. Doesn't introduce new option state i.e options on the input simply get preserved on the output.
4. Helps makes the transition to datashading easier for customized elements without going crazy by mapping simple options across (e.g alpha, cmap) as opposed to trying to tackle really difficult problems (e.g mapping bokeh point size to spread sizes).
5. Could be a boolean parameter which you can set to False if you want operations that don't touch options (i.e the current behavior).

My only worry is whether there are cases where transferring trying to transfer all options across have unintended negative affects. My other concern is that this might be tackling *half* the problem of seamlessly datashading elements while preserving the style, leaving the other half of the problem incomplete and very difficult to address.

jbednar:
> There's no semantic difference between an Image and whatever the new element would be

This is not to argue for solution 4 in particular, but I disagree. The Image element is one of the few I could be said to have designed, and as laid out in my [Continuous_Coordinates](http://holoviews.org/user_guide/Continuous_Coordinates.html) guide, my intended semantics for an Image type is as a gridded sampling/approximation to an underlying continuous field, such as temperature or brightness or some other continuous quantity. The potential Rasterized element would not be an approximation to a continuous field, it would be an approximation to a large number of discrete points, lines, paths, shapes, etc. Only if it's passed a raster for regridding would I say that the semantics of what rasterize() returns are the same as the usual Image semantics.

That said, I'm not necessarily arguing for a Rasterize element, because Elements don't inherit all their options from superclasses in a straightforward way, and the backwards compatibility implications of this approach seem problematic.

> One other option I'd mention is that we could consider allowing Store.transfer_options on operations, enable that option for rasterize, and change the default colormap for points/scatter (where 'fire' is a really bad idea). This would ensure that rasterized Points/Scatter specifically inherit the defaults from the image that's being rasterized.

Do you mean, inherit from the _Element_ _type_ that's being rasterized?  Otherwise I'm not sure what you mean; it's not an image being rasterized, usually.

> Additionally I'd also be happy to implement option 3 this week. 

That would be great, thanks!


> I will say that operations that try to preserve options from input to outputs seem less objectionable to me than ones that try to set things. 

Agreed. But `cmap` in particular is a tricky one, because there's some argument that if a user has set an explicit color on e.g. a Points element, they would expect the Datashaded plot to have that particular color, but then the color would be subject to overplotting because no color differences are available to show count values. We don't necessarily know whether people used `rasterize` because of speed or memory issues, or whether they particularly wanted the overplotting-avoidance that a multicolor colormap gives with datashader, and so it's not a no-brainer to simply accept the single color of that Element in this case. 

I guess we could, though: Datashader's shade function actually uses single colors just fine, keeping the color constant and modulating the alpha channel based on the intensity. Hmm; that's an interesting possibility -- respect the color of whatever we are rasterizing, but then modulate alpha, and then users can overlay the results with at least some useful properties. Hmm...


> The transfer_options suggestion does have some appealing properties:

I have _always_ wanted meaningful options to get transferred down to the result of datashade and rasterize. There are backwards compatibility implications of doing that at this late stage, and we are of course under pressure to preserve stability, but I think every new user of datashade or rasterize runs into this issue of options not transferring. So if it's practical, I am all for it!

> mapping bokeh point size to spread sizes

Yeah, that's a tough one. Maybe punt on that one until we have spatially extended points we can rasterize, not just hacky spreading tricks.

> Could be a boolean parameter which you can set to False if you want operations that don't touch options (i.e the current behavior).

Agreed.

> One worry I have is whether there are cases where transferring trying to transfer all options across have unintended negative affects. My other concern is that this might be tackling half the problem of seamlessly datashading elements while preserving the style, leaving the other half of the problem incomplete and very difficult to address. The options that can just be copied across are the easy ones...it is all the settings that need complex transformations that are hard to match.

Agreed. And unfortunately cmap and point size/linewidth are both the most obvious ones to want to transfer, and arguably the most complicated ones to do so, due to the very different models of color and size used by Datashader and normal plotting libraries.

jlstevens:
Thanks @philippjfr for adding the `nodata` support to `Dimension`. I do think it is worth keeping as a plot option to make it easy to override. That said, is `element.redim.nodata` now a thing?

philippjfr:
>That said, is element.redim.nodata now a thing?

Not yet, could easily add it though.

jbednar:
I've pushed some updates to the Style Mapping and Large Data notebooks. Large Data still needs work, but it's getting a lot closer. Remaining issues I can think of:

# Custom colormap
 
For the colormap, I've been trying Large Data with one of these two options:

```
from matplotlib.colors import LinearSegmentedColormap
shade_cmap = LinearSegmentedColormap.from_list("shade_cmap", ["lightblue", "darkblue"])
opts.defaults(opts.Image(cmap=shade_cmap))

from colorcet import kbc
opts.defaults(opts.Image(cmap=kbc[::-1]))
```

To-do:
- [ ] Pick Image colormap default globally
- [ ] Decide if rasterize should in some way have a separate cmap default
- [ ] Force a specific default in the Large Data notebook if we want something different from the defauls

# cnorm

- [ ] Large Data needs a discussion of the implications and usage of cnorm, but it's hard to talk about that until we pick a colormap, as it's hard to describe differences without mentioning specific colors. 


# Shared xy axes

In some cases (probably only a few of those needed) I've declared Paths to have separate u,v axes from Points, so that they are no longer normalized together to share axes when in the same Layout. The Points and Paths plots are meant to be entirely different visualizations, not sharing spatial axes, which seems easy enough to achieve in the examples if we take care to do so.

- [ ] Check if any other plots need separate xy axes.

# Shared Count axis

Even if the x,y axes are separate, all rasterized plots end up with a z (color) Dimension of "Count". As a result, they all get normalized together, which can make some plots become very dim, clustered near the bottom of the colormap range. To fix this I've added `.redim(Count="pcount")` on one of the Path plots when it's displayed next to a Points plot, but this seems like it will be a big practical issue. Just because we use a Count aggregator in two rasterize calls, it doesn't seem to me that we would expect the color axes to be shared; they are rasterizations of what can be entirely independent quantities, united only in their size and not their semantics. So I'm not at all sure it's appropriate for them to share a Dimension.

Could we rename their dimensions to make them better reflect that they are rasterizations of _different_ data? E.g. have rasterize create a Dimension like `Count of x,y` or `Count of Points` or something else that would more often be unique when the plots are different and the same only when there is a good reason to think that the quantities are comparable? Otherwise users will never be able to rasterize more than one Element in a Layout without explicitly redim-ing the Counts.

- [ ] Decide what to do about shared Count axes

# Updates for nodata

- [ ] Remove `nodata` nearly everywhere it is used, as it's no longer needed in most cases
- [ ] Add a discussion of when to use `nodata`

# Misc

- [ ] Add a custom hover formatter (see above)?
- [ ] Define element.redim.nodata?
- [ ] Need to add colorbar examples as soon as bokeh=2.3.0dev7 is available on NPM (Mateusz is working on it)

holoviews/plotting/mpl/__init__.py
```diff
@@ -221,18 +221,20 @@ def grid_selector(grid):
...
 dflt_cmap = 'fire'
+dflt_chart_cmap = 'kbc_r'
+
 # Default option definitions
 # Note: *No*short aliases here! e.g use 'facecolor' instead of 'fc'
 
 # Charts
 options.Curve = Options('style', color=Cycle(), linewidth=2)
-options.Scatter = Options('style', color=Cycle(), marker='o', cmap=dflt_cmap)
+options.Scatter = Options('style', color=Cycle(), marker='o', cmap=dflt_chart_cmap)
```
jlstevens, line 231: What about `Path`?

jlstevens: @jbednar apparently Path already has viridis as a default colormap. Maybe it should be consistent?

jlstevens:
Commit 10ffb26 addresses https://github.com/holoviz/holoviews/issues/4692

jlstevens:
@jbednar Do you really think `hv.output(backend='bokeh')` is necessary at the end of the Large Data notebook?

I think it is just confusing in a documentation context.

holoviews/plotting/mpl/__init__.py
jbednar: Path should be consistent.

jbednar:
This is looking great! There are still lots of loose ends to tidy up, though.

- [ ] Setting kbc_r as the default colormap, as described in #3500. I tried simulating that here using 
```
from colorcet import kbc
kbc_r = kbc[::-1]
opts.defaults(opts.Image(cmap=kbc_r))
```
but that's not working for rasterize(), presumably because the cmap is now transferred from other Element defaults? Need the cmap made to be kbc_r globally in a separate PR and then rebase this one, presumably. At that point need to check that all the colormaps are now kbc_r, and fix any that aren't; unless there's some special reason, no other colormaps should be needed in 14-Large_Data.

- [ ] The changes to the Count dimension name make it work much better, thanks! But I agree with Philipp that the string formatting should match what Histogram does already, i.e. not 
"Count over x_y" but something like "x_y Count".

- [ ] In the `datashaded.opts(title='datashade') + rasterized.opts(title='rasterized equivalent')` plot, why are the colors different? I see Philipp and Jean-Luc have discussed this, but I don't see that as having been resolved.

- [ ] In the Spreading section, the plots with xlim and ylim specified that are meant to show zoomed in bits with spreading visible, actually only show a few blown-up pixels. Somehow the image is not refreshing to show the actual rasterized viewport.

- [ ] I put in a new cell in the last section "Element types supported for Datashading" showing what the output of `rasterize` is, which should replace what's currently shown for `datashade` once it's working. It's using the fire colormap, which is working very poorly since high data disappears, but presumably this will be fixed when the default colormap is set?

- [ ] Also, I couldn't see how to make the non-datashaded Texas choropleth use the same Glasbey colormap for comparison; presumably there's a dimension that needs specifying? Somehow it defaults to Viridis right now. Conversely, a `rasterize()`d choropleth shows up empty (all white), presumably because the active areas are all marked white, and it needs to change either to Viridis (or ideally Glasbey) so it works properly.

- [ ] The "Element types supported for Datashading needs to include an example of Rectangles, which are now supported. I've listed it in text, but it would be nice to show it as well.

- [ ] At least in some cases, it looks like spread and dynspread don't work on DynamicMaps?  Try changing `dynamic=False` to `dynamic=True` in the last cell where it appears; it complains that the output is neither an Image nor an RGB (which of course it isn't; it's a DynamicMap). This seems like a big problem, as DynamicMaps are the default output of `rasterize()`. Yet I don't see this same problem in other cells; maybe it's specific to the mpl backend?

- [ ] Need to document `cnorm` in 14-Large_Data.

- [ ] 14-Large_Data depends on Datashader 0.11.2a5, which is a dev version of unreleased code. Datashader needs a real release, but that's held up right now due to CUDA being broken in master. Seems like we can release hv without ds, but the docs will now be very confusing to people.

> @jbednar Do you really think hv.output(backend='bokeh') is necessary at the end of the Large Data notebook?

Yes, because the notebook won't work cell-by-cell otherwise, but I've added a comment on it to explain that.

holoviews/core/data/interface.py
```diff
@@ -359,7 +359,17 @@ def dtype(cls, dataset, dimension):
...
+    def replace_value(cls, data, nodata):
+        """
+        Replace `nodata` value in data with NaN
+        """
+        data = data.astype('float64')
+        mask = data != nodata
+        if hasattr(data, 'where'):
+            return data.where(mask, np.NaN)
+        return np.where(mask, data, np.NaN)
+
```
jlstevens, line 372: I guess this duplicates `apply_nodata._replace_value` in `plotting.util` a bit but that is probably ok...

holoviews/core/operation.py
```diff
@@ -178,6 +178,9 @@ def process_element(self, element, key, **params):
         The process_element method allows a single element to be
         operated on given an externally supplied key.
         """
+        if self._per_element and not isinstance(element, Element):
+            return element.clone({k: self.process_element(el, key, **params)
+                                  for k, el in element.items()})
```
jlstevens, line 183: To clarify, does this mean `_per_element` simply wasn't respected for `DynamicMaps`?

jlstevens:
> 14-Large_Data depends on Datashader 0.11.2a5, which is a dev version of unreleased code. Datashader needs a real release, but that's held up right now due to CUDA being broken in master. Seems like we can release hv without ds, but the docs will now be very confusing to people.

I think the docs should work with 0.11 as datashader 0.11.2 isn't going to be out in time for this release. We can keep a copy of the notebook ready so that as soon as 0.11.2 *is* released, we can have a minor HoloViews release at which point we can make those updated docs live. Until that happens, I think we can *mention* that spreading on rasterize will be supported in 0.11.2 but we should use the `datashade` operation so that everything keeps working.

philippjfr:
One thing though, I thought the next datashader release is 0.12, not 0.11.2 given that it includes various features.

jlstevens:
> One thing though, I thought the next datashader release is 0.12, not 0.11.2 given that it includes various features.

I agree though the 0.11.2 alphas are already released and support aggregate spreading. 0.12 would have made more sense to be as well.

philippjfr:
>I agree though the 0.11.2 alphas are already released and support aggregate spreading

alphas don't bind us in any way. It absolutely should still be released as 0.12 since it's not just aggregate spreading but also anti-aliasing.

jbednar:
More to-do:

Jim:
- [ ] File issue for .hist() overriding cnorm='eq_hist'

Jean-Luc:
- [ ] Change default colormap to kbc_r uniformly
- [ ] Add config option to restore previous default cmap behavior
- [ ] Change normalization to cnorm in shade() and datashade(), supporting normalization but warning if both are supplied (and use cnorm in that case)

Philipp:
- [ ] Rebase this PR on the one with cmap name string support
- [ ] Back out putting cmap on transfer_options?
- [ ] element.redim.nodata

jlstevens:
@jbednar When you pull you will want to change `normalization` to `cnorm` in the notebook to avoid the new warnings.

holoviews/plotting/mpl/__init__.py
```diff
@@ -219,6 +220,14 @@ def grid_selector(grid):
 register_cmap("fire", cmap=fire_cmap)
 register_cmap("fire_r", cmap=fire_r_cmap)
 
+def mpl_cm(name,colorlist):
+    cm = LinearSegmentedColormap.from_list(name, colorlist, N=len(colorlist))
+    register_cmap(name, cmap=cm)
+
+mpl_cm('kbc_r',list(reversed(kbc)))
```
jbednar, line 227: Not sure if there's a reason to define mpl_cm here; is it ever reused?

jbednar: Seems like this should just follow the pattern that fire_cmap and fire_r_cmap follow above, or vice versa.

jlstevens: Addressed in c72f83b

holoviews/operation/datashader.py
```diff
@@ -1264,9 +1264,13 @@ def _process(self, element, key=None):
         kdims = element.kdims
 
         overrides = dict(self.p.items())
-        if 'normalization' in overrides and config.future_deprecations:
-            self.param.warning("Shading 'normalization' parameter deprecated, "
-                               "use 'cnorm' parameter instead'")
+        if 'normalization' in overrides:
+            if 'cnorm' in overrides:
+                self.param.warning("Cannot supply both 'normalization' and 'cnorm' for shading; "
+                                   "use 'cnorm' instead'")
```
jlstevens, line 1270: What do you mean 'cannot'...this is just a warning so something will happen and one will take precedence!

jlstevens: In addition, it doesn't seem like an error if both specify the same value (e.g 'eq_hist`) even if it is redundant. A warning makes sense here but I don't think the wording is correct.

jlstevens: Maybe "Both the 'cnorm' and 'normalization' keywords specified; 'cnorm' value taking precedence over deprecated 'normalization' option"

jbednar:
Remaining to-do:

- [ ] HeatMap on each of the different backends is using acmap of RdBu or RdYlBu_r instead of kbc_r. I think this should change, as those two colormaps differ, are not perceptually uniform, and assume symmetrically divergent data which isn't a valid assumption. 
- [ ] Jim to finish up his local Large Data and Styling changes and push, either in this PR if done in time or separately.

Remaining issues not expected to be addressed in this PR, but worth noting for later:

- [ ] It would be nice to have a custom hover formatter so that a count of 0 doesn't show up as NaN on hover.
- [ ] We should still consider whether the default aggregator for Path and Curve should be `any`; for isolated lines `any` will work much better, but for large numbers of overlaid paths or curves it won't. We can worry about that independently of this PR.
- [ ] We might want to put `shade_cmap = mpl.colors.LinearSegmentedColormap.from_list("shade_cmap", ["lightblue", "darkblue"])` somewhere in our colormap definitions so that people can more easily port `datashade()` calls to use `rasterize()` or to replicate plots first done in Datashader's own API.
- [ ] We could consider mapping a single color from Curve, Path, Points, etc. down to `cmap` using Datashader's support for mapping single-colors as values using the alpha channel. That way if someone overlays red, blue, yellow, etc. lines and calls `rasterize`, the resulting images will still have the same base color per element. But we don't have time to evaluate this idea now.
- [ ] We have not addressed the idea of transferring options from Elements to their rasterized (or potentially datashaded) equivalents. There's a mechanism for it, but someone would need to work out all the implications, and the current mechanism of `transfer_options` doesn't allow partial mappings, transformed mappings, etc. that would be needed in practice.

jlstevens:
Tests are passing. Time to hit merge on this rather large and long running PR!

PR #4572, opened 2020-08-29
Link selections streams refactor

jonmmease:
## Description of Changes
This PR refactors the `link_selections` transformation to fully rely on chained streams converting linked stream interactions (BoundsXY and Lasso) into linked selection interactions.

This is done by extracting much of `link_selections` logic into two new stream classes:
 - `SelectionExprSequence` represents the selection produced by a sequence of selection operations on a single plot (e.g. when selection_mode is `"intersect"`).  `SelectionExprSequence` is a `Derived` stream subclass that uses as input stream a `History` stream wrapping a `SelectionExpr` stream.  It is configured by a `mode` property that corresponds to the `link_selections.selection_mode` parameter.  It produces a `selection_expr` and `region_element` params that represent the series of selections on the plot, combined according to the current `mode`.

 - `CrossFilterSet` represents the combined selection across all linked plots. `CrossFilterSet` is a `Derived` stream subclass that uses as input one `SelectionExprSequence` stream per linked plot. It produces a combined `selection_expr` param. It is supports a `mode` argument that corresponds to the `link_selections.cross_filter_mode` param.

The `SelectionDisplay` for each linked plot now gets the combined `selection_expr` from the `CrossFilterSet` stream, and it gets the region element from the `SelectionExprSequence` stream associated with the plot.

## Motivation
The `link_selections` class still registers callbacks on it's own params to update values of non-interactive streams (e.g. setting the `selected_color` param pushes an update to the internal `_Cmap` stream), but the linked stream updates are handled using `Derived` and `History` streams.  This makes it possible to introspect the Dynamic object produced by `link_selections` and find the linked streams that are driving it.  Along with the new `decollate` transformation logic, this will make it possible to add support for displaying selection linked plots using Dash.

## Testing

The current `link_selections` test suite has been updated, and I ran through the documentation notebook and everything looks like it's working the same way to me. But it would be good for someone else to run through that notebook and play with it by hand as well to check for regressions.

Thanks!


jonmmease:
In 3492f0a, 64b6f9c, and 4547b36 I added a few Plotly-specific changes related to linked selections, but not directly needed by the core `link_selections` refactor here.

Happy to move these to a separate PR if that would make it easier to review.

philippjfr:
@jonmmease Could you rebase? At that point I'd love to merge this asap so I can start on a further refactor to start reusing some of the linked selection logic for something else.

jonmmease:
> Could you rebase?

Definitely, I'll do it today. Thanks!

jonmmease:
I merged and updated. If you have a strong preference I can work through a proper rebase.  link selections tests still pass, but I haven't done any manual QA post merge. Could you give your later changes a try on this branch? The things I noticed that happened since I started this are the bokeh tool for choosing the selection mode, and the caching logic.  Thanks!

philippjfr:
Seems fine, but I'll review properly shortly. Is there a tool in Plotly to change selection modes?

jonmmease:
> Seems fine, but I'll review properly shortly. Is there a tool in Plotly to change selection modes?

Not really. At the JavaScript level it's possible to add custom buttons so it might be possible to do something at that level, but it would need to be routed through Panel/Dash first.

philippjfr:
Thanks, appreciate it.

PR #4579, opened 2020-08-31
Fix aspect if width/height has been constrained

philippjfr:
If a user supplies min/max width and/or height then the initially computed aspect may be wrong because we can't actually work out what width/height bokeh computes until it is rendered. Therefore we allow the PlotSize stream to trigger an initial event which sets the correct aspect by adjusting the axis limits after initial rendering.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -793,7 +793,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         fixed_width = (self.frame_width or options.get('width'))
         fixed_height = (self.frame_height or options.get('height'))
         constrained_width = options.get('min_width') or options.get('max_width')
-        constrained_width = options.get('min_height') or options.get('max_height')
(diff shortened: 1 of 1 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -789,6 +792,8 @@ def _update_ranges(self, element, ranges):
         options = self._traverse_options(element, 'plot', ['width', 'height'], defaults=False)
         fixed_width = (self.frame_width or options.get('width'))
         fixed_height = (self.frame_height or options.get('height'))
+        constrained_width = options.get('min_width') or options.get('max_width')
+        constrained_width = options.get('min_height') or options.get('max_height')
```
kcpevey, line 796: I think you mean constrained_height?

philippjfr: Thanks :)

philippjfr:
```suggestion
        constrained_height = options.get('min_height') or options.get('max_height')
```

Issue #4581, opened 2020-08-31
unpickling dim object causes RecursionError


jonmmease:
When unpickling a `holoviews.util.transform.dim` object, a `RecursionError` is raised.

```python
from holoviews.util.transform import dim
import pickle
pickle.loads(pickle.dumps(dim("a")))
```
```
...
    ns = self._ns
  File "/home/jmmease/PyDev/repos/holoviews/holoviews/util/transform.py", line 271, in __getattr__
    if attr in dir(self):
  File "/home/jmmease/PyDev/repos/holoviews/holoviews/util/transform.py", line 277, in __dir__
    ns = self._ns
  File "/home/jmmease/PyDev/repos/holoviews/holoviews/util/transform.py", line 271, in __getattr__
    if attr in dir(self):
RecursionError: maximum recursion depth exceeded
```

PR on the way

PR #4582, opened 2020-08-31
Add __getstate__ / __setstate__ method to dim to fix RecursionError

jonmmease:
Closes https://github.com/holoviz/holoviews/issues/4581

The `Recursion` error was being caused by bouncing between the `__dir__` and `__getattr__` methods while unpickling.  This adds `__getstate__`/`__setstate__` methods to explicitly implement pickling.


philippjfr:
Thanks, looks good.

PR #4604, opened 2020-09-11
Allow toggling link_selections mode from bokeh

philippjfr:
Allow toggling the `selection_mode` on `link_selections` from the context menu in the bokeh toolbar:

Screen Shot 2020-09-11 at 12 08 14 PM


jlstevens:
Very nice!

Issue #3949, opened 2019-09-08
Missing user guide link to "Plotting with plotly"?


ceball:
http://holoviews.org/user_guide/index.html (and github master) has a link to "Plotting with plotly" (http://holoviews.org/user_guide/Plotting_with_Plotly.html) under "Supplementary guides", but it goes nowhere. I would have fixed it, but I couldn't see the intended target anywhere.

jbednar:
That's a good point!  I don't see it either, nor any record of it.  Odd that it's never been pointed out before!  Of course, if we had one, it would be very out of date by now, as the Plotly backend is now much more fleshed out.  @jonmmease, any interest in making such a guide to parallel http://holoviews.org/user_guide/Plotting_with_Bokeh.html and http://holoviews.org/user_guide/Plotting_with_Matplotlib.html?

PR #4605, opened 2020-09-11
Add Plotly Dash support

jonmmease:
This PR adds support for displaying dynamic HoloViews objects using Plotly Dash.

The new logic is localized to the single `holoviews/plotting/plotly/dash.py` file which provides a new `holoviews_to_dash` function.

A new section has been added to all of the dynamic Plotly reference guide notebooks that includes the code for displaying the example in a Dash app.  This is a good place to look at the example usage for the `holoviews_to_dash` function.

This PR is dependent on some of the fixes in https://github.com/holoviz/holoviews/pull/4572, and so it should be rebased on master after #4572 is merged.


pytest.ini
```diff
@@ -0,0 +1,2 @@
+[pytest]
+addopts = -p no:dash
```
jonmmease, line 2: This was added to disable the dash pytest plugin, which was causing this error (https://travis-ci.org/github/holoviz/holoviews/jobs/731249907):

```
pluggy.manager.PluginValidationError: Plugin 'dash' could not be loaded: (dash-table 4.4.1 (/home/travis/miniconda/envs/test-environment/lib/python2.7/site-packages), Requirement.parse('dash-table==4.10.1'))!
```

jonmmease:
I rebased on master and this PR is ready for review.

examples/reference/containers/plotly/DynamicMap.ipynb
```diff
@@ -74,6 +75,41 @@
...
+    "     components.store]\n",
+    ")\n",
+    "\n",
+    "if __name__ == '__main__':\n",
+    "    app.run_server(debug=True)\n",
+    "```"
+   ],
+   "metadata": {
+    "collapsed": false,
+    "pycharm": {
```
jlstevens, line 108: Wonder if this metadata is why GitHub isn't able to preview these notebooks? I recommend just clearing it out...

jlstevens:
Looks great! 

Not much to say other than I would consider naming `holoviews_to_dash` to simple `to_dash` - anything imported from holoviews is implicitly all about holoviews :-)

examples/reference/containers/plotly/DynamicMap.ipynb
```diff
@@ -74,6 +75,41 @@
...
+   "source": [
+    "This example can also be displayed using [Plotly Dash](https://dash.plotly.com/):\n",
+    "\n",
+    "```python\n",
+    "from holoviews.plotting.plotly.dash import holoviews_to_dash\n",
+    "import dash\n",
+    "import dash_html_components as html\n",
+    "\n",
+    "# Create App\n",
+    "external_stylesheets = ['https://codepen.io/chriddyp/pen/bWLwgP.css']\n",
```
jbednar, line 89: It would be good to avoid depending on code in anyone's personal accounts; does it really need the .css decoration? Or could the file be moved somewhere more permanent?

examples/reference/streams/plotly/Bounds.ipynb
```diff
@@ -67,6 +68,37 @@
...
+    "external_stylesheets = ['https://codepen.io/chriddyp/pen/bWLwgP.css']\n",
+    "app = dash.Dash(__name__, external_stylesheets=external_stylesheets)\n",
+    "\n",
+    "# Dash display\n",
+    "components = holoviews_to_dash(\n",
+    "    app, [layout], reset_button=True\n",
+    ")\n",
+    "\n",
+    "app.layout = html.Div(\n",
+    "    [components.graphs[0], components.resets[0], components.store]\n",
```
jbednar, line 92: I don't suppose there's a way to eliminate this step?  Is it not appropriate for `to_dash` to modify `app` to do that already, at least as a default (with the default overridden in some examples that need special widgets)? 

Overall, this is a lot of code to be adding to each of the examples, when most of it seems boilerplate.  Can we not instead just make a single page in the user guide that shows one Panel example, one Voila example, and one Dash example (plus the caveat that Dash will only work for the Plotly backend), and then have all the reference notebooks (for all backends) link back to that page? That seems easier for us to maintain plus throws less code at the user to try to follow (assuming any given user will click on many of these pages).

pytest.ini
jbednar: Would it be better to add dash-table to the testing envt so that the plugin doesn't have to be disabled?

holoviews/plotting/plotly/dash.py
```diff
@@ -0,0 +1,548 @@
...
+                stream_callbacks[cb.output_id] = cb
+                populate_stream_callback_graph(stream_callbacks, [stream.input_stream])
+
+
+def encode_store_data(store_data):
+    """
+    Encode store_data dict into a JSON serializable dict
+
+    This is currently done by pickling store_data and converting to a base64 encoded
+    string. If HoloViews support JSON serialization in the future, this method could
```
jbednar, line 222:
```suggestion
    string. If HoloViews supports JSON serialization in the future, this method could
```

holoviews/plotting/plotly/dash.py
```diff
@@ -0,0 +1,548 @@
...
+        # Update store_data with interactive stream values
+        for fig_ind, fig_dict in enumerate(initial_fig_dicts):
+            graph_id = graph_ids[fig_ind]
+            # plotly_stream_types
+            for plotly_stream_type, uid_to_streams_for_type in uid_to_stream_ids.items():
+                panel_prop = plotly_stream_type.callback_property
+                if panel_prop == "selected_data":
+                    if graph_id + ".selectedData" in triggered_prop_ids:
+                        # Only update selectedData values that just changed.
+                        # This way we don't the the may have been cleared in the
```
jbednar, line 478: ???

jbednar (commented):
Looks great!  I have a few small suggestions, and generally a wish for consolidating the Dash examples to avoid bloating the Reference gallery (see comments), but it will be great to have Dash support!  

I think I've seen a demo of Dash supporting other libraries like Matplotlib and Bokeh, though I don't know how solid that was. I wonder if the mpl backend for HoloViews could be supported with Dash fairly easily, because it wouldn't need to be very ambitious (just displaying an image). Bokeh would be more tricky, I assume.

jonmmease:
> Not much to say other than I would consider naming `holoviews_to_dash` to simply `to_dash`

Sure, works for me

pytest.ini
jonmmease: I did try that, but the version on the main anaconda channel is too old (`4.4.1`). We'd have to pull `4.10.*` from conda-forge, and it doesn't look like any packages are pulled from conda-forge right now.

examples/reference/containers/plotly/DynamicMap.ipynb
jonmmease: Yeah, good point. I'll clear it out if we decide to keep these notebook changes

examples/reference/containers/plotly/DynamicMap.ipynb
jonmmease: Ahh, yeah, I'll pull that out if we keep the examples in these notebooks.

examples/reference/streams/plotly/Bounds.ipynb
jonmmease: > I don't suppose there's a way to eliminate this step? Is it not appropriate for to_dash to modify app to do that already, at least as a default (with the default overridden in some examples that need special widgets)?

I don't think so, the idea is that user should place them in the layout of a larger Dash app that includes other Dash components.

I'll make some comments about the docs down below

jonmmease:
Regarding the documentation. I'm currently working on a documentation page that will be added to the main Dash docs with a structure similar to https://dash.plotly.com/layout.  I've waffled on this a little bit, but I'm currently leaning towards including Dash versions of these 6 interactive Plotly examples in that documentation page.  The nice thing about having them there is that they will be fully interactive in the documentation.

So the simplest path forward may be for me to revert the notebook changes in this PR, then after the Dash documentation is published (which will require these changes to be released in HoloViews), add links from these pages to the corresponding example in the Dash docs.

How does that sound?  The shared Panel/Voila/Dash page makes sense, but I guess I'm not volunteering to write it for this PR :upside_down_face:

jlstevens:
Linking to plotly docs from the reference notebooks seems reasonable to me. My other suggestion is that instead of putting code into those notebooks and only having links, a section could be added to the [Plotting with Plotly](http://holoviews.org/user_guide/Plotting_with_Plotly.html) user guide but unfortunately it looks there was a problem with doc building :-(

image

jonmmease:
> Linking to plotly docs from the reference notebooks seems reasonable to me. 

Ok, I'll go ahead and revert the notebook changes from this PR and then come back to it after the Dash docs are up.

> ... Plotting with Plotly ...

Looks like there isn't an rst file for *Plotting with Plotly*

Screenshot_20200930_051430

Do you think this existed at one point? Or maybe was never written?

jlstevens:
> Do you think this existed at one point? Or maybe was never written?

Good question! As far as I remember, those `.rst` files need to be explicitly added and I see no reason why such a file would ever have been removed. Could you add one for plotly in this PR? Following the Bokeh example, it should just be:

```rst
Plotting with Plotly
____________________

.. notebook:: holoviews ../../examples/user_guide/Plotting_with_Plotly.ipynb
    :offset: 1
```

jonmmease:
### Updates
 - `holoviews_to_dash` renamed to `to_dash` in https://github.com/holoviz/holoviews/pull/4605/commits/dd373a95fc8e215f4deadf679021a5ceb9b10b29
 - Notebook changes reverted in https://github.com/holoviz/holoviews/pull/4605/commits/59a93814e6185bb0253de8b55052f2a03c2edab3. I'll add links to Dash docs once they are live.

> Could you add one for plotly in this PR? Following the Bokeh example, it should just be ...

Looks like a target `Plotting_with_Plotly.ipynb` notebook was never written. This was raised in https://github.com/holoviz/holoviews/issues/3949.  I should write this soon, but I don't think it will be in this PR.

setup.py
```diff
@@ -67,7 +68,7 @@
     'mock',
     'flake8 ==3.6.0',
     'coveralls',
-    'path.py', 
+    'path.py',
```
jonmmease, line 71: whoops, my editor :hocho: the trailing whitespace here automatically, I can revert if anyone cares that it's not related to this PR.

philippjfr:
That's right, a `Plotting_with_Plotly` guide was never actually written.

holoviews/plotting/plotly/dash.py
```diff
@@ -0,0 +1,558 @@
...
+
+    # Add kdim sliders
+    kdim_uuids = []
+    for kdim_name, (kdim_label, kdim_range) in all_kdims.items():
+        slider_uuid = str(uuid.uuid4())
+        slider_id = kdim_name + "-" + slider_uuid
+        slider_label_id = kdim_name + "-label-" + slider_uuid
+        kdim_uuids.append(slider_uuid)
+
+        html_lable = html.Label(id=slider_label_id, children=kdim_label)
```
nicolaskruchten, line 398: typo: `html_lable` ;)

jonmmease: Thanks! https://github.com/holoviz/holoviews/pull/4605/commits/772b9ed2dd30c2b47beda4e1eea3b7309a2d81b2

jonmmease:
@kcpevey, is this Ubuntu Python 3.6 build failure a known issue?  I don't see any error messages related to the tests that I recognize. Thanks!

kcpevey:
@jonmmease that is not a known failure. I thought perhaps it was a one-off thing so I reran the workflows and now there are 3 failures across different versions/architectures. Based on the error, I don't know what's causing that.

philippjfr:
Trying to fix up those flakey tests now.

jonmmease:
Thanks @philippjfr!

PR #4611, opened 2020-09-15
Params stream returns True/False for triggering Action

### LINKED ISSUE #4610: Params stream does not correctly support transient
Demo code and outcome:

```python
import param
import panel as pn
import holoviews as hv
hv.extension('bokeh')

class Foo(param.Parameterized):
    
    a = param.Action(default=lambda x: x.param.trigger('a'))
    b = param.Action(default=lambda x: x.param.trigger('b'))
    
    def _callback(self, *args, **kwargs):
        print(args, kwargs)
        return hv.Curve([1,2,3])
    
    def view(self):
        pstream = hv.streams.Params(self, 
                                    parameters=['a', 'b'], 
                                    transient=True)
            
        return pn.Row(self.param,
                      hv.DynamicMap(self._callback,  streams=[pstream]))
                       
Foo().view()
```

image

Clicking on the 'A' and 'B' buttons have the same result in the console:

image

I would expect `transient=True` to set the value to `None` for the button that isn't clicked, allowing me to know which button was clicked.

philippjfr:
Closes #4610 

- [x] Updates test
- [ ] Add docs

### DIFF
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -757,10 +757,19 @@ class Params(Stream):
     def contents(self):
         if self._watch_only:
             return {}
-        filtered = {(p.owner, p.name): getattr(p.owner, p.name) for p in self.parameters}
-        return {self._rename.get((o, n), n): v for (o, n), v in filtered.items()
-                if self._rename.get((o, n), True) is not None}
-
+        contents = {}
+        for p in self.parameters:
+            key = (p.owner, p.name)
+            if key in self._rename and self._rename[key] is None:
+                continue
+            name = self._rename.get(key, p.name)
+            if isinstance(p, param.Action):
+                value = any(p.owner in (event.obj, event.cls) and p.name == event.name
+                            for event in self._events)
+            else:
+                value = getattr(p.owner, p.name)
+            contents[name] = value
+        return contents
 
 
 class ParamMethod(Params):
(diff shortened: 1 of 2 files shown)


jlstevens:
Looks good though I also pushed a PR (#4612) to do the same thing which I now closed. I think that PR is clearer about what it is trying to do but it doesn't handle renaming. So I would simply add a comment to the new code in `contents` to explain that this is about handling Action parameters in a special way.

jlstevens:
Note that this PR aims to address #4610.

jlstevens:
I see two unit tests that need updating:

```
FAIL: test_param_stream_action 
..
AssertionError: [{'action': True}] != [{'action': <function TestParamsStream.setUp.<locals>.InnerAction.<lambda> at 0x7fc8725c3488>}]


FAIL: test_param_stream_memoization 
...
AssertionError: [{'action': True, 'x': 0}] != [{'action': <function TestParamsStream.setUp.<locals>.InnerAction.<lambda> at 0x7fc88c1a1ea0>, 'x': 0}]
```

jlstevens:
I checked out this PR and it is doing what I would expect:

image

The fact the panel buttons don't look right is a separate issue...

philippjfr:
Ready to merge if we agree on the behavior.

jlstevens:
Just a note to say we are considering a new param Parameter type that might make this PR redundant. I'll link to the relevant param PR as soon as it exists.

jlstevens:
I think we are agreed that this should be addressed with a new `Event` parameter type and not with special handling at the HoloViews level. This PR was useful for deciding what behavior we would like to see but I don't think we want this code merged. Closing.

PR #4612, opened 2020-09-15
Added special handling to Params for Action parameters

jlstevens:
Work in progress to address #4610. 

- [ ] Handle renaming properly
- [ ] Fix/add tests
- [ ] Make sure we agree on this policy


jlstevens:
Philipp opened a PR addressing the same thing a few minutes before I pushed this #4611 which handles renaming.

PR #4624, opened 2020-09-21
CuDF Series is not iterable

bryevdv:
@philippjfr This is a fix for https://github.com/holoviz/hvplot/issues/510 (which should probably be copied to this repo)

Summary: `.unique()` returns a CuDF Series, which is not iterable unless copied to host. 

No tests as of yet (looking in to GPU CI options)


jlstevens:
Looks good to me!

philippjfr:
Let's go ahead and merge and then find a solution for GPU CI later. Before a 1.13.5 I or someone else should run the test suite against latest cuDF again.

PR #4634, opened 2020-09-28
Improve handling of categorical dtypes on empty arrays

philippjfr:
An empty array of string types was not treated correctly for categorical colormapping for various reasons:

- The dtype was dropped in the range calculation
- An explicit colormapping is treated as ground truth for factors and forces categorical colormapping


jlstevens:
Thanks! 

This addresses the problem I was having combining `PointDraw` with `Link/LinkCallback`.

PR #4639, opened 2020-09-30
Use PIL for RGB.load_image

philippjfr:
Switches to using PIL for `RGB.load_image` and adds proper docstring. `plt.imread` leaked memory the way it was used and using PIL directly eliminates an unnecessary dependency on matplotlib here.

### DIFF
--- a/holoviews/element/raster.py
+++ b/holoviews/element/raster.py
@@ -634,7 +634,7 @@ class RGB(Image):
             raise ImportError("RGB.load_image requires PIL (or Pillow).")
 
         with open(filename, 'rb') as f:
-            data = np.array(Image.open(f.read()))
+            data = np.array(Image.open(f))
 
         if array:
             return data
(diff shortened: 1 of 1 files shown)


kcpevey:
You beat me to this. Thanks!

jlstevens:
PIL historically has not been a dependency of HoloViews but it has been an indirect dependency, at least when `datashader` is [used](https://github.com/holoviz/datashader/blob/master/setup.py#L21). If datashader isn't installed, you could have a `PIL` import error but that isn't any worse than a matplotlib import error if you are just using Bokeh frankly.

I'm fine with this though I note `PIL` seems to be imported in the plotly backend:

```
import PIL.Image
        except ImportError:
            raise VersionError("""\
Rendering RGB elements with the plotly backend requires the Pillow package""")
```

The question then is whether we want to make PIL into a direct dependency...

philippjfr:
`plt.imread` uses PIL internally so this really doesn't change anything but drop the unnecessary dependency on matplotlib. Both matplotlib and bokeh depend on PIL as well afaik.

PR #4641, opened 2020-10-02
Added selected parameter to Bokeh PathPlot

jlstevens:
Looks like `PathPlot` was forgotten when the `selected` parameter was added to the Bokeh backend. This PR aims addresses this small oversight.


philippjfr:
Looks good, thanks!

Issue #4642, opened 2020-10-13
Linked Brushing Tap not working


WesleyTheGeolien:
Thanks for contacting us! Please read and follow these instructions carefully, then delete this introductory text to keep your issue easy to read. Note that the issue tracker is NOT the place for usage questions and technical assistance; post those at [Discourse](https://discourse.holoviz.org) instead. Issues without the required information below may be closed immediately.

#### ALL software version info
Python: 3.8.5
Holoviews: 1.13.4
Bokeh: 2.2.1
Jupyter Notebook Server: 6.0.3
JupyterLab: 1.2.0
Firefox: 81.0.1 (64-bit)
Chrome: 87.0.4280.11 (Official Build) dev (64-bit)
OS: Fedora release 32 (Thirty Two)

#### Description of expected behavior and the observed behavior
The tap selection whilst using linked brushing does not seem to be working (Tap does not do anything). I would expect the tap selection to work the same as the box or lasso select by clicking once on a data point. I have tried simple tap, double tap, holding on to left click, I have also varied the type of selection

#### Complete, minimal, self-contained example code that reproduces the issue
I observe this behaviour in the linked brushing demo notebook: https://holoviews.org/user_guide/Linked_Brushing.html / https://raw.githubusercontent.com/pyviz/holoviews/master/examples/user_guide/Linked_Brushing.ipynb when working on the chlorplet of the USA in the index based selections chapter.

```
import holoviews as hv
from holoviews.selection import link_selections

from bokeh.sampledata.us_counties import data as counties
from bokeh.sampledata.unemployment import data as unemployment

counties = [dict(county, Unemployment=unemployment[cid])
            for cid, county in counties.items()
            if county["state"] == "tx"]

choropleth = hv.Polygons(counties, ['lons', 'lats'], [('detailed name', 'County'), 'Unemployment'])
hist = choropleth.hist('Unemployment', adjoin=False, normed=False)

linked_choropleth = link_selections(choropleth + hist, index_cols=['detailed name'])

linked_choropleth.opts(
    hv.opts.Polygons(tools=['hover', 'tap', 'box_select'], xaxis=None, yaxis=None,
                   show_grid=False, show_frame=False, width=500, height=500,
                   color='Unemployment', colorbar=True, line_color='white'),
    hv.opts.Histogram(width=500, height=500)
)

```

philippjfr:
[...]

PR #4643, opened 2020-10-13
Implement index based selection in linked brushing

philippjfr:
There appears to have been a regression where `index_cols` was no longer being passed through so that didn't work at all (have to add some tests here). However additionally this also adds support for using the tap tool alongside the lasso and box select tools when `index_cols` are provided.

Implements https://github.com/holoviz/holoviews/issues/4642

ToDo:

- [ ] Write tests for index_cols examples
- [x] Solve problem of both the Selection1D and Lasso/BoundsXY stream triggering on box/lasso select. If we already have the index computed in JS we don't have to do the reverse mapping, from box/lasso -> index selection in Python but currently it's doing both. This also has the benefit that you'll be able to do lasso select without spatialpandas/geopandas as long as you specify `index_cols`.

@jonmmease please review

### DIFF
--- a/holoviews/element/selection.py
+++ b/holoviews/element/selection.py
@@ -17,26 +17,30 @@ class SelectionIndexExpr(object):
 
     _selection_streams = (Selection1D,)
 
+    def __init__(self, *args, **kwargs):
+        super(SelectionIndexExpr, self).__init__(*args, **kwargs)
+        self._index_skip = False
+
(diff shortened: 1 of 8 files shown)


jonmmease:
Thanks for working this out @philippjfr.  I haven't tested it, but the changes all look reasonable. Agreed that it would definitely be nice to have a couple tests to lock it down going forward.

PR #4644, opened 2020-10-13
Ensure PolyDrawCallback always has vdim data

philippjfr:
HoloViews tries to avoid sending data to the frontend if it knows that data isn't going to be used. This would usually go unnoticed but when attaching a CDS stream it can be the case that the data gets updated and now the `vdims` will be missing. A user that is relying on all the data from an element to be available will now get unexpected errors because that data is missing. This simply ensures that if the data is missing it is added on the Python side before the user sees it.


jlstevens:
I've just tested this PR and it fixes an issue I was facing that I was (wrongly) attributing to geoviews.

Issue #4052, opened 2019-10-10
datashaded image not zooming/panning properly


jbednar:
Using https://daac.ornl.gov/daacdata/cms/CMS_DARTE_V2/data/onroad_2017.tif and the following code, I can show a geotiff:

```
import xarray as xr, holoviews as hv, colorcet as cc
from holoviews.operation.datashader import datashade

hv.extension('bokeh', logo=False)
emissions = xr.open_rasterio('CMS_DARTE_V2_1735/data/onroad_2017.tif').load()[0]
datashade(hv.Image(emissions, ['x','y']), cmap=cc.fire).opts(width=500, invert_yaxis=True)
#hv.Image(emissions, ['x','y']).opts(cmap=cc.fire, width=500, invert_yaxis=True)
```

![image](https://user-images.githubusercontent.com/1695496/66581642-1390b700-eb46-11e9-9312-efd70dc23c37.png)

However, zooming and panning don't work properly; the image jumps around and gets cut off, and in some scenarios shows only garbage:

![image](https://user-images.githubusercontent.com/1695496/66582023-bba68000-eb46-11e9-967b-6d344bfb5766.png)
![image](https://user-images.githubusercontent.com/1695496/66582047-c4975180-eb46-11e9-977c-92e0a1202cd9.png)
![image](https://user-images.githubusercontent.com/1695496/66581990-aaf60a00-eb46-11e9-85c2-08acdab27773.png)

Panning in the x direction seems to work fine, so I'm suspecting that it's to do with the inverted y axis; panning in y leaves the data in the same spot in the plot axes but the bounding box of the data going up. If I uncomment the non-datashaded `hv.Image` line, zooming and panning work fine, but the image is only barely visible and of course it's quite slow.

I'm on git master for the various HoloViz projects:

holoviews=1.13.0a10.post6+gc2f0c7bf6
datashader=0.8.0a1.post4+ge9a2889
geoviews=1.6.4a3
hvplot=0.5.0a1.post6+g8784838
xarray=0.13.0

xavArtley:
Hello simple reproducer if it can help
```python
import numpy as np
import holoviews as hv
from holoviews.operation.datashader import rasterize
hv.extension('bokeh')

ls = np.linspace(0, 10, 2000)
xx, yy = np.meshgrid(ls, ls)

bounds=(-1,-1,1,1)   # Coordinate system: (left, bottom, right, top)
[...]

PR #4665, opened 2020-10-27
Inverted axes fixes for raster plots

philippjfr:
Fix https://github.com/holoviz/holoviews/issues/4052

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -897,26 +897,25 @@ class RangeXYCallback(Callback):
         data = {}
         if 'x0' in msg and 'x1' in msg:
             x0, x1 = msg['x0'], msg['x1']
+            if x0 > x1:
+                x0, x1 = x1, x0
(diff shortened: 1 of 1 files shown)


holoviews/plotting/bokeh/raster.py
```diff
@@ -96,9 +96,9 @@ def get_data(self, element, ranges, style):
 
         dh, dw = t-b, r-l
         if self.invert_xaxis:
-            l, r = r, l
+            dw = -dw
         if self.invert_yaxis:
-            b, t = t, b
+            dh = -dh
```
jbednar, line 101: D'oh! I guess this was always just a bug, since dh,dw have already been computed and thus l,r,b,t will just be ignored?  Damn.

jbednar (approved):

holoviews/plotting/bokeh/raster.py
philippjfr: I think it mostly works in a lot of scenarios, but not others. Honestly not quite sure.

PR #4672, opened 2020-10-31
Fix releases urls

Hoxbro:
Fixes releases.rst urls for version 1.13.5


philippjfr:
Thanks for catching this!

PR #4683, opened 2020-11-13
Support link_selections on dynamic overlays

jonmmease:
This PR adds support for performing `link_selections` on DynamicMaps that are the result of performing the `__mul__` operator on one or more DynamicMap layers.

This came up when trying to use `link_selections` and datashader together on top of a `Tiles` element.

e.g.
```python
link_selections(tiles * datashade(...))
```

The result of  `tiles * datashade(...)` is a `DynamicMap` that returns an `Overlay`. Previously we skipped these in `link_selections` because it's not possible to accurately recurse in the general case.

This PR handles only the very common case of DynamicMaps generated by the `__mul__` operator.

### DIFF
--- a/holoviews/selection.py
+++ b/holoviews/selection.py
@@ -164,7 +164,8 @@ class _base_link_selections(param.ParameterizedFunction):
                     self._selection_streams, hvobj, operations,
                     self._selection_expr_streams.get(hvobj, None), cache=self._cache
                 )
-            elif issubclass(hvobj.type, Overlay):
+            elif (issubclass(hvobj.type, Overlay) and
+                  getattr(hvobj.callback, "name", None) == "dynamic_mul"):
                 return Overlay([
                     self._selection_transform(el, operations=operations)
                     for el in callback.inputs
--- a/holoviews/selection.py
+++ b/holoviews/selection.py
@@ -164,6 +164,11 @@ class _base_link_selections(param.ParameterizedFunction):
                     self._selection_streams, hvobj, operations,
                     self._selection_expr_streams.get(hvobj, None), cache=self._cache
                 )
+            elif issubclass(hvobj.type, Overlay):
+                return Overlay([
+                    self._selection_transform(el, operations=operations)
+                    for el in callback.inputs
+                ]).collate()
             else:
                 # This is a DynamicMap that we don't know how to recurse into.
                 return hvobj
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -1020,7 +1020,7 @@ class SelectionExpr(Derived):
                 return
 
             selection_expr, bbox, region_element = selection
-                
+
             if selection_expr is not None:
                 break
 
@@ -1107,6 +1107,8 @@ class SelectionExprSequence(Derived):
         combined_region_element = None
 
         for selection_contents in stream_values[0]["values"]:
+            if selection_contents is None:
+                continue
             selection_expr = selection_contents['selection_expr']
             if not selection_expr:
                 continue
--- a/holoviews/tests/testselection.py
+++ b/holoviews/tests/testselection.py
@@ -278,6 +278,71 @@ class TestLinkSelections(ComparisonTestCase):
             )[()]
         )
 
+    @ds_skip
+    def test_datashade_in_overlay_selection(self):
+        points = Points(self.data)
+        layout = points * dynspread(datashade(points))
+
+        lnk_sel = link_selections.instance(unselected_color='#ff0000')
+        linked = lnk_sel(layout)
+        current_obj = linked[()]
+
+        # Check base points layer
+        self.check_base_points_like(current_obj[()].Points.I, lnk_sel)
+
+        # Check selection layer
+        self.check_overlay_points_like(current_obj[()].Points.II, lnk_sel, self.data)
+
+        # Check RGB base layer
+        self.assertEqual(
+            current_obj[()].RGB.I,
+            dynspread(
+                datashade(points, cmap=lnk_sel.unselected_cmap, alpha=255)
+            )[()]
+        )
+
+        # Check RGB selection layer
+        self.assertEqual(
+            current_obj[()].RGB.II,
+            dynspread(
+                datashade(points, cmap=lnk_sel.selected_cmap, alpha=255)
+            )[()]
+        )
+
+        # Perform selection of second and third point
+        selectionxy = TestLinkSelections.get_value_with_key_type(
+            lnk_sel._selection_expr_streams, hv.Points
+        ).input_streams[0].input_stream.input_streams[0]
+
+        self.assertIsInstance(selectionxy, SelectionXY)
+        selectionxy.event(bounds=(0, 1, 5, 5))
+        current_obj = linked[()]
+
+        # Check that base points layer is unchanged
+        self.check_base_points_like(current_obj[()].Points.I, lnk_sel)
+
+        # Check points selection layer
+        self.check_overlay_points_like(current_obj[()].Points.II, lnk_sel,
+                                       self.data.iloc[1:])
+
+        # Check that base RGB layer is unchanged
+        self.assertEqual(
+            current_obj[()].RGB.I,
+            dynspread(
+                datashade(points, cmap=lnk_sel.unselected_cmap, alpha=255)
+            )[()]
+        )
+
+        # Check selection RGB layer
+        self.assertEqual(
+            current_obj[()].RGB.II,
+            dynspread(
+                datashade(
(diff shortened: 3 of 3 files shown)


jbednar:
> This PR handles only the very common case of DynamicMaps generated by the __mult__ operator.

Thanks! What cases _wouldn't_ be handled?  I had thought that an hv.Overlay generated with the mult operator was the same as one generated any other way.

jonmmease:
> I had thought that an hv.Overlay generated with the mult operator was the same as one generated any other way.

That's what I had thought as well, but turns out that `hv.Overlay([dmap, el])` returns on `hv.Overlay` consisting of an `Element` and a `DynamicMap` while `dmap * el` returns a `DynamicMap` that produces an overlay.

> Thanks! What cases wouldn't be handled?

It doesn't handle the general case where someone writes a custom dynamic map callback function that returns an overlay, but I'm not sure there is a meaningful way to handle this case.

jbednar:
I wonder if the regular Overlay constructor should be extended to do the same as `__mul__` when it encounters a DynamicMap. @jlstevens ?

jlstevens:
> That's what I had thought as well, but turns out that hv.Overlay([dmap, el]) returns on hv.Overlay consisting of an Element and a DynamicMap while `dmap * el` returns a `DynamicMap` that produces an overlay.

That is correct. The issue here is that `__mul__` operator has the freedom to switch the resulting type while the `Overlay` constructor will always return an `Overlay` (without some heavy magic). @philippjfr and I discussed trying to pull out the machinery of `__mul__` into an operation but this is quite tricky unfortunately.

Nothing stops us having an `overlay` utility that simply applies `__mul__` to the supplied elements/dynamicmaps but I'm not sure whether that will help here (it would probably also be confusing). Of course `Overlay` could also have a `classmethod` that does this (probably a better idea) but again, I'm not sure how much this would help here.

philippjfr:
We could easily add a __new__ implementation that does the right thing for hv.Overlay([el, dmap]), which is probably preferable over asking the user to call collate in that case.

jlstevens:
I'm not sure how I feel about constructors that give you a different type as a result. Also, `dmap * el` is a perfectly valid structure that also has its uses. How would you toggle between these behaviors?

philippjfr:
Not sure what you mean by the second point. What I was suggesting wouldn't affect the * behavior so what would be toggled between?

jlstevens:
I think there are cases where `hv.Overlay([el, dmap])` defining an overlay of an element and a `DynamicMap` is a perfectly reasonable thing to want.

philippjfr:
I suppose link_selections can just do the introspection and apply the collate itself so a user doesn't have to.

jlstevens:
Right.

I don't think this really affects this PR either way as the question now is whether:

1. Users should call `.collate` before using `link_selections` 
2. Whether `link_selections` can just apply `.collate`
3. Whether we should change `Overlay`

I vote for option 2 as option 3 is a big drastic right now.

jbednar:
Thanks for explaining all that; I did suspect it was related to `collate()`. What I missed was that `*` wasn't returning an Overlay, even though Jon did say that. :-) I vote for option 2; if we detect in link_selections that it needs collating first, just collate it.

If link_selections does that, are there then any remaining overlay-like things (i.e. Overlays and things dynamically returning Overlays) that won't be supported for link_selections, assuming the things in the overlay are individually supported?

Also, what about elements in the Overlay that aren't supported for link_selections?  I would hope that those can just be ignored, so that their presence does not break the overall usability of the overlay.  Maybe it should warn for such Elements, so that the user knows some elements are being ignored, but allow the warning to be silenced with an explicit argument to link_selections. (And the warning could mention the name of that argument, so that it's easy for the user to do the right thing.) The reason I'm asking is that previously using link_selections with an overlay was very awkward and required learning new concepts (grabbing the ls object and applying it individually to the bits that work), and I'd really hope that only really weird circumstances would need special treatment like that.

jonmmease:
> 2. Whether link_selections can just apply .collate (affects this PR, doesn't affect the rest of HoloViews)

The trouble is that, currently, `collate` doesn't do anything with DynamicMaps that return overlays:

```python
import holoviews as hv
from holoviews.plotting.util import initialize_dynamic

dmap_overlay = hv.Points([(0, 0)]) * datashade(hv.Points([0, 0]))
dmap_overlay
```
```
:DynamicMap   []
   :Overlay
      .Points.I :Points   [x,y]
      .RGB.I    :RGB   [x,y]   (R,G,B,A)
```

```python
dmap_overlay.collate()
```
```
:DynamicMap   []
   :Overlay
      .Points.I :Points   [x,y]
      .RGB.I    :RGB   [x,y]   (R,G,B,A)
```

We could potentially move the approach that I used in this PR to update `collate` to return an `Overlay` of `Points` and `DynamicMap` in this case, but that seems like a larger change than just handling it in `link_selections` for now. Perhaps we could record this as something to revisit in version 2.

> If link_selections does that, are there then any remaining overlay-like things (i.e. Overlays and things dynamically returning Overlays) that won't be supported for link_selections, assuming the things in the overlay are individually supported?

I think this covers everything that we can reasonably cover. Elements in the overlay (dynamic or not) that don't support selections are retained unmodified.  There is not warning at the moment though. This worked previously with the result of `Overlay([...])`, but in the case of the `__mul__` operator with one or more `DynamicMap` objects, the whole overlay was ignored.  With this PR both cases should work the same way.

As I mentioned before, if someone builds a `DynamicMap` based on a custom callback function that happens to return an Overlay, there isn't really anything we can do to introspect into the function. So DynamicMaps built this way will still be returned unmodified.

jlstevens:
I must be confused or missing something...I thought the case that needs `collate` is `hv.Overlay([el, dmap])` and that everything worked when using `__mul__`? When using `__mul__` I wouldn't expect `collate` to change anything which is what you show in your example...

philippjfr:
Right, that was my understanding as well, specifically you said:

>This PR handles only the very common case of DynamicMaps generated by the `__mul__` operator.

So why would collate be needed for that case at all?

philippjfr:
I guess you assumed that we were suggesting that collate could somehow replace what you've done here, which is not the case. We were simply suggesting that we could cover the `hv.Overlay([..., dmap, ...])` as well by **combining** collate with what you've done here.

jonmmease:
Ohh, I see. Thanks for clarifying. I was confused because the `link_selections(hv.Overlay([dmap, el]))` case already worked (before this PR) without calling `.collate()`. It's handled here: https://github.com/holoviz/holoviews/blob/886b03d6af8f02855ac2f426dca1259d6f82cb03/holoviews/selection.py#L187-L189

This makes sense to me because it's the same logic path that regular (non-dynamic) overlays follow.

But yeah, if you all want I think we could probably call `.collate` on everything at the start of `link_selections` to steer that case through the code I added in this PR. But this might have other consequences for other input types that I don't foresee right now.  It would be a larger change with more risk of regression in my opinion, but does that sound preferable?

jbednar:
Right -- that's what I'm after. Keeping all the progress you've made in this PR, but using collate on Overlays so that it can go even further and cover more cases (hence my question about what's still _not_ covered if this extension is made to the PR).

jbednar:
>  It would be a larger change with more risk of regression in my opinion, but does that sound preferable?

I think it's worth a try, because it addresses a class of inputs that we know is going to fail, so it's a known fix against a possible but unknown possibility of breakage.  We can always revert it, but it does seem worth trying!

jonmmease:
> I think it's worth a try, because it addresses a class of inputs that we know is going to fail

It's not clear to me what these cases are.  The current state of this PR handles `__mul__` and `Overlay([...])` for dynamic and non-dynamic inputs.

jbednar:
>> What cases wouldn't be handled?

> It doesn't handle the general case where someone writes a custom dynamic map callback function that returns an overlay, but I'm not sure there is a meaningful way to handle this case.

I _think_ the proposal is to call .collate() for this case and thereby cover it. Right?

jonmmease:
I'll think on it some more, but right now I don't see how collate would help with that case. @philippjfr and @jlstevens?

philippjfr:
Yeah, I hadn't realized `hv.Overlay([..., dmap, ...])` was already covered but that does make sense. I'm -1 on adding collate too then since it could cause some regression. Sorry for even bringing it up since it derailed discussion a fair bit.

jonmmease:
> Sorry for even bringing it up since it derailed discussion a fair bit.

No, I appreciate you weighing in. There's a lot going on here so I really appreciate whatever comes to your mind 🙂

philippjfr:
The MacOS failures are unrelated. Merging.

jonmmease:
Thanks, @philippjfr!

PR #4686, opened 2020-11-15
Plotly map tiles support

jonmmease:
## Overviews
This PR enhances the plotly backend with full support for the `holoviews.Tiles` element for displaying maps.

cc @philippjfr @jlstevens @jbednar 

## Features
 - `holoviews.Tiles` element that can display raster maps from a tile server (like the Bokeh backend) or vector tiles maps from the Mapbox service.  Vector tiles from mapbox require a free account with mapbox.com and a user token, but no account or token is required otherwise.

 - Many elements can be overlayed on top of `Tiles` including: `Scatter`, `Points`, `Curve`, `Labels`, `RGB`, `Path`, `Rectangles`, `Bounds`, and `Segments`.

 - The `Range*`, `Bounds*`, and `Selection1d` streams are all supported for traces overlayed on top of `Tiles`.

 - As a consequence of this feature set, `datashader` and `link_selections` both work on elements overlayed on `Tiles`.

 - A new pair of static methods were added to `Tiles` class for converting between lat/lon coordinates and pseudo-Mercator coordinates: `Tiles.lon_lat_to_easting_northing` and `Tiles.easting_northing_to_lon_lat`

 - All of this works with the new Dash support.

 - I added a simple `Tiles` documentation page that matches the Bokeh version, but includes a small section on the vector tiles support.  I plan to add a few examples to the Dash HoloViews docs PR once this is released.

## Challenges
For posterity, here are a few of the challenges I ran into, all a consequence of how plotly.js relies on Mapbox for mapping support:

 - While Mapbox displays maps in the same pseudo-Mercator coordinate system as Bokeh, accepts and returns coordinate in the longitude, latitude coordinate system.  For consistency with the current paradigm, the external interface works with pseudo-Mercator and conversions are done internally in all interactions with plotly.js.

 - Plotly has a separate set of traces for display on top of mapbox layers. So the trace used to draw a line on a map is not the same trace as is used to draw a regular 2D line. Same for lines, text, images, etc.  To make this work with the HoloViews paradigm, many of the plotly plot classes were updated to be able to render themselves either to a 2D Cartesian coordinates, or on top of a mapbox map layer.  This geo mode is activated automatically whenever a geo-supported element is overlayed with a `Tiles` element.  An error is raised on render if an element without geo support is overlayed with a `Tiles` element.

 - Similarly to above, the zoom/pan/selection events emitted by Plotly.js are different for mapbox subplots than regular 2D Cartesian subplots, and they operate in lat/lon coordinates.  So the Plotly callback classes were all updated to handle the mapbox case as well as the Cartesian subplot case.

In the end, I think it all worked out pretty well and the user experience shouldn't be any more complicated than when using the Bokeh backend.

### All built-in raster tile maps

Screenshot_20201115_175756

### `datashader` and `link_selections` over vector tile layer

plotly_holoviews_map_datashader_link_selections

## Relationship to other PRs

For testing purposes, this PR contains the changes in https://github.com/holoviz/holoviews/pull/4683#issuecomment-727236514. I'll merge master to fix the conflicts once that one is merged.

I fixed a bug in plotly.js that caused RGB image layers to be pushed behind raster layers. https://github.com/plotly/plotly.js/pull/5269.  Until this fix is released, datashader can only really be used on top of the mapbox vector tile layers. After the fix is released in plotly.js, it will also work properly on top of raster tile layers.

## TODO
 - [x] Python 2.7 compatibility
 - [x] Visually inspect geo elements that cross the date boundary to confirm whether mapbox handles the wraparound internally.

### DIFF
--- a/holoviews/plotting/plotly/dash.py
+++ b/holoviews/plotting/plotly/dash.py
@@ -475,31 +475,39 @@ def to_dash(app, hvobjs, reset_button=False, graph_class=dcc.Graph):
             graph_id = graph_ids[fig_ind]
             # plotly_stream_types
             for plotly_stream_type, uid_to_streams_for_type in uid_to_stream_ids.items():
-                panel_prop = plotly_stream_type.callback_property
-                if panel_prop == "selected_data":
-                    if graph_id + ".selectedData" in triggered_prop_ids:
-                        # Only update selectedData values that just changed.
-                        # This way we don't the the may have been cleared in the
-                        # store above
-                        stream_event_data = plotly_stream_type.get_event_data_from_property_update(
-                            selected_dicts[fig_ind], initial_fig_dicts[fig_ind]
-                        )
-                        for uid, event_data in stream_event_data.items():
-                            if uid in uid_to_streams_for_type:
-                                for stream_id in uid_to_streams_for_type[uid]:
-                                    store_data["streams"][stream_id] = event_data
-                elif panel_prop == "viewport":
-                    if graph_id + ".relayoutData" in triggered_prop_ids:
-                        stream_event_data = plotly_stream_type.get_event_data_from_property_update(
-                            relayout_dicts[fig_ind], initial_fig_dicts[fig_ind]
-                        )
-
-                        for uid, event_data in stream_event_data.items():
-                            if event_data["x_range"] is not None or event_data["y_range"] is not None:
+                for panel_prop in plotly_stream_type.callback_properties:
+                    if panel_prop == "selected_data":
+                        if graph_id + ".selectedData" in triggered_prop_ids:
+                            # Only update selectedData values that just changed.
+                            # This way we don't the the may have been cleared in the
+                            # store above
+                            stream_event_data = plotly_stream_type.get_event_data_from_property_update(
+                                panel_prop, selected_dicts[fig_ind], initial_fig_dicts[fig_ind]
+                            )
+                            for uid, event_data in stream_event_data.items():
                                 if uid in uid_to_streams_for_type:
                                     for stream_id in uid_to_streams_for_type[uid]:
-                                        store_data["streams"][
-                                            stream_id] = event_data
+                                        store_data["streams"][stream_id] = event_data
+                    elif panel_prop == "viewport":
+                        if graph_id + ".relayoutData" in triggered_prop_ids:
+                            stream_event_data = plotly_stream_type.get_event_data_from_property_update(
+                                panel_prop, relayout_dicts[fig_ind], initial_fig_dicts[fig_ind]
+                            )
+
+                            for uid, event_data in stream_event_data.items():
+                                if event_data["x_range"] is not None or event_data["y_range"] is not None:
+                                    if uid in uid_to_streams_for_type:
+                                        for stream_id in uid_to_streams_for_type[uid]:
+                                            store_data["streams"][stream_id] = event_data
+                    elif panel_prop == "relayout_data":
+                        if graph_id + ".relayoutData" in triggered_prop_ids:
+                            stream_event_data = plotly_stream_type.get_event_data_from_property_update(
+                                panel_prop, relayout_dicts[fig_ind], initial_fig_dicts[fig_ind]
+                            )
+                            for uid, event_data in stream_event_data.items():
+                                if uid in uid_to_streams_for_type:
+                                    for stream_id in uid_to_streams_for_type[uid]:
+                                        store_data["streams"][stream_id] = event_data
 
         # Update store with derived/history stream values
         for output_id in reversed(stream_callbacks):
--- a/holoviews/plotting/plotly/dash.py
+++ b/holoviews/plotting/plotly/dash.py
@@ -479,8 +479,8 @@ def to_dash(app, hvobjs, reset_button=False, graph_class=dcc.Graph):
                     if panel_prop == "selected_data":
                         if graph_id + ".selectedData" in triggered_prop_ids:
                             # Only update selectedData values that just changed.
-                            # This way we don't the the may have been cleared in the
-                            # store above
+                            # This way we don't save values that may have been cleared
+                            # from the store above by the reset button.
                             stream_event_data = plotly_stream_type.get_event_data_from_property_update(
                                 panel_prop, selected_dicts[fig_ind], initial_fig_dicts[fig_ind]
                             )
--- a/holoviews/plotting/plotly/images.py
+++ b/holoviews/plotting/plotly/images.py
@@ -37,7 +37,7 @@ class RGBPlot(ElementPlot):
             # Create a dummy invisible scatter trace for this image.
             # This serves two purposes
             #  1. The two points placed on the corners of the image are used by the
-            #     autoscale logic to allow using the autoscale button to property center
+            #     autoscale logic to allow using the autoscale button to properly center
             #     the image.
             #  2. This trace will be given a UID, and this UID will make it possible to
             #     associate callbacks with the image element. This is needed, in particular
--- a/holoviews/plotting/plotly/images.py
+++ b/holoviews/plotting/plotly/images.py
@@ -5,6 +5,7 @@ import numpy as np
 from plotly.graph_objs.layout import Image as _Image
 
 from ...core.util import VersionError
+from ...element import Tiles
 from .element import ElementPlot
 from .selection import PlotlyOverlaySelectionDisplay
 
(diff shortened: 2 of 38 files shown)


holoviews/plotting/plotly/dash.py
```diff
@@ -475,31 +475,39 @@ def update_figure(*args):
...
-                            relayout_dicts[fig_ind], initial_fig_dicts[fig_ind]
-                        )
-
-                        for uid, event_data in stream_event_data.items():
-                            if event_data["x_range"] is not None or event_data["y_range"] is not None:
+                for panel_prop in plotly_stream_type.callback_properties:
+                    if panel_prop == "selected_data":
+                        if graph_id + ".selectedData" in triggered_prop_ids:
+                            # Only update selectedData values that just changed.
+                            # This way we don't the the may have been cleared in the
```
jbednar, line 482: ???

holoviews/plotting/plotly/images.py
```diff
@@ -17,28 +18,42 @@ class RGBPlot(ElementPlot):
...
+                'showlegend': False
+            }
+            return dict(mapbox=dict(layers=[layer]),
+                        traces=[dummy_trace])
+        else:
+            image = dict(datum, **options)
+            # Create a dummy invisible scatter trace for this image.
+            # This serves two purposes
+            #  1. The two points placed on the corners of the image are used by the
+            #     autoscale logic to allow using the autoscale button to property center
```
jbednar, line 40:
```suggestion
            #     autoscale logic to allow using the autoscale button to properly center
```

holoviews/plotting/plotly/tiles.py
```diff
@@ -0,0 +1,65 @@
...
+        return extents
+
+    def init_graph(self, datum, options, index=0, **kwargs):
+        return {'traces': [datum], "mapbox": options}
+
+    def generate_plot(self, key, ranges, element=None, is_geo=False):
+        """
+        Override to force is_geo to True
+
+        """
```
jbednar, line 64:
```suggestion
        """
```

holoviews/tests/plotting/plotly/testcallbacks.py
```diff
@@ -96,6 +96,46 @@ def setUp(self):
...
+                {'type': 'scattermapbox', 'uid': 'first', 'subplot': 'mapbox'},
+                {'type': 'scattermapbox', 'uid': 'second', 'subplot': 'mapbox2'},
+                {'type': 'scattermapbox', 'uid': 'third', 'subplot': 'mapbox3'}
+            ],
+            'layout': {
+                'title': {'text': 'Figure Title'},
+            }
+        }).to_dict()
+
+        # Precompue pair so lat/lon, easting/northing, mapbox coord values
```
jbednar, line 110:
```suggestion
        # Precompute pair so lat/lon, easting/northing, mapbox coord values
```

(Can't quite parse it, though!)

jbednar (commented):
Looks great! I can't see how it could be cleaner, given the impedance mismatch between how Plotly and Bokeh handle projections, and Plotly's restrictions on what can be overlaid on maps.

The web mercator conversion functions look good, but to me it seems like they ought to be bare functions, not static class methods, so that it's easier for people to call them individually. For instance, in the plotly/Tiles.ipynb file, seems like we should suggest that people use those functions to convert their data first.

holoviews/plotting/plotly/dash.py
jonmmease: Thanks! Yeah, that comment didn't make sense.  fixed in 2da9ef9

jonmmease:
Thanks for taking a look @jbednar!

> The web mercator conversion functions look good, but to me it seems like they ought to be bare functions, not static class methods, so that it's easier for people to call them individually.

What I liked about making them static methods is that it avoids the need to import them separately from the `Tiles` element.  In your view, why would bare functions be easier to call individually?

Regardless, it's not a major point and I'm happy to move them somewhere else if someone has a suggestion.

jlstevens:
> What I liked about making them static methods is that it avoids the need to import them separately from the Tiles element. In your view, why would bare functions be easier to call individually?

Why not both? :-) 

I would be happy for there to be staticmethods on `Tiles` (where they clearly make sense) that import and use functions supplied as utilities (bare functions). That way both use cases are supported and HoloViews would allow you to do these transforms without necessarily mentioning the `Tiles` element when it isn't relevant.

jonmmease:
> Why not both? :-)

Works for me! Does `holoviews.util.transform.(lon_lat_to_easting_northing|easting_northing_to_lon_lat)` make sense?

jlstevens:
Makes sense to me at least!

jonmmease:
Ok, in 5c0a4a9 I moved the implementations to standalone functions that are called from the `Tiles` static methods.

jbednar:
> In your view, why would bare functions be easier to call individually?

I guess easier isn't the right word; more like "conceptually cleaner". I.e., it feels dirty to call a static method on a class that I'm not using, making me feel like it's somehow tied to that class in particular, and shouldn't be used outside that context. Those functions are fully general, and now that they are available as bare functions, people can feel comfortable using them anywhere they are useful. Thanks!

jonmmease:
I merge in master, and now the only changes in the PR that are outside of the plotly plotting directory are the addition of the lat/lon coordinate conversion functions.

I don't see anything in the failing tests that look related to this PR, but let me know if there's anything I can help out with there.

philippjfr:
This all looks good to me but I have some questions about projections.

1. How are image projections handled? Does it just project the bounding box? If so won't this cause various distortions?
2. What happens for objects that cross the date boundary? Is that handled at all or will this cause weird issues wrapping around the entire globe?
3. How do you imagine extending this work in GeoViews by building on cartopy? Can/should we make the projection code here pluggable so we can simply override it in GeoViews?

holoviews/tests/element/testtiles.py
```diff
@@ -0,0 +1,114 @@
...
+        self.check_array_type_preserved(
+            np.array, np.ndarray,
+            lambda a, b: np.testing.assert_array_almost_equal(a, b, decimal=2)
+        )
+
+    def test_pandas_series(self):
+        self.check_array_type_preserved(
+            pd.Series, pd.Series,
+            lambda a, b: pd.testing.assert_series_equal(
+                a, b, atol=0.01, check_exact=False
```
philippjfr, line 110: atol does not exist for the py2 pandas version it's getting.

philippjfr: Sorry you still have to worry about py2 here!

jonmmease: Ahh, thanks for the pointer!

jonmmease:
Thanks for taking a look @philippjfr 

> This all looks good to me but I have some questions about projections.

> How are image projections handled? Does it just project the bounding box? If so won't this cause various distortions?

Mapbox is a little odd here in my opinion. The corners of the image are defined in lat/lon, but Mapbox doesn't apply any distortion to the image when laying it out on the map (i.e. the image pixels don't go through the web-mercator projection). So the image needs to be generated in web-mercator coordinates.  So the end result is that it works just like Bokeh, but I have to convert the corners to lat-lon.  Let me know if that didn't make sense.

> What happens for objects that cross the date boundary? Is that handled at all or will this cause weird issues wrapping around the entire globe?

I think Mapbox handles the wrap-around internally, but I'll visually confirm this for the various element types this evening.

> How do you imagine extending this work in GeoViews by building on cartopy? Can/should we make the projection code here pluggable so we can simply override it in GeoViews?

To be honest, I haven't looked much into how GeoViews handles this.  What I can say is that Mapbox can't display anything besides Web-mercator. I am interested in working through what from GeoViews would apply here. But is it all right if, for now, I open an issue in GeoViews and we push this off after this HoloViews release?

philippjfr:
> So the end result is that it works just like Bokeh, but I have to convert the corners to lat-lon. Let me know if that didn't make sense.

It does make sense but I guess it does introduce distortion if you provide a lat/lon image which is why something proper projection with cartopy is still preferable.

>To be honest, I haven't looked much into how GeoViews handles this. 

GeoViews defines one operation called `gv.project` which handles all projecting and dispatches to the correct cartopy routines depending on the type of data, e.g. points, vectors, images, quadmeshes, geometries are all handled separately.

>I am interested in working through what from GeoViews would apply here. But is it all right if, for now, I open an issue in GeoViews and we push this off after this HoloViews release?

That's fine and once the pandas test issues are fixed I'm happy to merge.

jonmmease:
> it does introduce distortion if you provide a lat/lon image which is why a proper projection with cartopy is still preferable.

Yeah, that's right.  So geoviews would run a lat/lon image through cartopy to distort it into Web-mercator?  Very cool!

> That's fine and once the pandas test issues are fixed I'm happy to merge.

Awesome, I'll fix the tests and make that GeoViews issue this evening.

philippjfr:
>Yeah, that's right. So geoviews would run a lat/lon image through cartopy to distort it into Web-mercator? Very cool!

But sadly fairly slow...

>Awesome, I'll fix the tests and make that GeoViews issue this evening.

Great, thanks!

jonmmease:
> What happens for objects that cross the date boundary? Is that handled at all or will this cause weird issues wrapping around the entire globe?

A quick check with a long line that goes all the way around the world.

newplot

The Mapbox and easting/northing to lon/lat conversions do the right thing. And mapbox automatically repeats the line as you pan around the world. So I think that's working fine.

jonmmease:
Tests passing. I think this is ready to go now @philippjfr

philippjfr:
Thanks, this is really exciting!

jonmmease:
Thanks!

PR #4690, opened 2020-11-18
Fix mysterious id_offset bug

philippjfr:
Can't explain the bug but this should fix it:

```
  File "C:\Miniconda3\envs\test-environment\lib\site-packages\holoviews\core\options.py", line 1751, in id_offset
    max_id = max(store_ids)+1 if len(store_ids) > 0 else 0
ValueError: max() arg is an empty sequence
```

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -1748,7 +1748,9 @@ class StoreOptions(object):
         max_ids = []
         for backend in Store.renderers.keys():
             store_ids = Store.custom_options(backend=backend).keys()
-            max_id = max(store_ids)+1 if len(store_ids) > 0 else 0
+            max_id = max(store_ids, default=0)
+            if store_ids:
+                max_id += 1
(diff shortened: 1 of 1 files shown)


holoviews/core/options.py
```diff
@@ -1748,7 +1748,7 @@ def id_offset(cls):
         max_ids = []
         for backend in Store.renderers.keys():
             store_ids = Store.custom_options(backend=backend).keys()
-            max_id = max(store_ids)+1 if len(store_ids) > 0 else 0
+            max_id = max(store_ids, default=0)+1
```
jbednar, line 1751: I can't say why it didn't work before, but won't this change it to return 1 in the empty list case, where it used to return 0?

philippjfr: Good point.

philippjfr: Thankfully tests are just as attentive as you!

PR #4691, opened 2020-11-19
Dask/cudf histogram fixes

jonmmease:
This PR has two small fixes for displaying histograms using dask and cudf.

 - First, the plotly histogram plot was passing Dask arrays into the Plotly Figure dictionary which plotly.py doesn't handle.
 - Second, the histogram operation code was still using the deprecated `normed` argument in one place. Numpy still supports this with a runtime deprecation warning, but cupy doesn't. The replacement is the `density` kwarg, which we were already using several places in the same function so I don't think backward compatibility should be affected at all. Also, the `normed` kwarg has been deprecated in numpy for a long time now (since 1.6.0)

Do these make sense @philippjfr ?

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -717,7 +717,7 @@ class histogram(Operation):
         if is_cupy:
             import cupy
             full_cupy_support = LooseVersion(cupy.__version__) > '8.0'
-            if not full_cupy_support and (normed or self.p.weight_dimension):
+            if not full_cupy_support and (normed or self.p.weight_dimension): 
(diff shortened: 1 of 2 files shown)


philippjfr:
I've got various cuDF fixes I still need to push up including a fix for `normed` and the Ibis PR completely refactors the histogram operation. So maybe drop the normed change for now.

jonmmease:
Sure thing! Thanks for working on all that

philippjfr:
Thanks @jonmmease!

PR #4693, opened 2020-11-20
Fix handling of custom matplotlib and bokeh colormaps

### LINKED ISSUE #4635: Strange rasterize colormapping in Large Data user guide
While discussing #4567 with @jbednar, we noticed the following strange output in the [Large Data User Guide](http://holoviews.org/user_guide/Large_Data.html)

image

This looks pretty strange as it looks like the *reverse* of the [fire colormap](https://colorcet.holoviz.org/) where black is low and white is high as the raster holds positive counts with *higher* counts in the middle:

</img>

Initially I thought this might be related to the recent work in #4567 but then realized that it can't be as that PR hasn't been merged yet.

Presumably it used to look like the GIF shown for the next code cell (i.e normal fire with a black background)?

</img>

philippjfr:
Two issues compounded causing our default 'fire' cmap to invert in recent releases. In particular recent versions of matplotlib changed how to look up colormaps and we had to adapt, however in doing so we were no longer picking up custom matplotlib colormaps (such as the fire cmap we were registering), which caused us to fall back to using the bokeh 'fire' palette. The issue there was that we had custom handling to invert non-matplotlib based bokeh palettes, because those are generally stored in reverse order, but since 'fire' in particular was stored in the correct order it was incorrectly inversed.

Fixes #4635

### DIFF
--- a/holoviews/plotting/util.py
+++ b/holoviews/plotting/util.py
@@ -684,7 +684,10 @@ def _list_cmaps(provider=None, records=False):
     if 'matplotlib' in provider:
         try:
             import matplotlib.cm as cm
-            mpl_cmaps = list(cm.cmaps_listed)+list(cm.datad)
+            if hasattr(cm, '_cmap_registry'):
+                mpl_cmaps = list(cm._cmap_registry)
+            else:
+                mpl_cmaps = list(cm.cmaps_listed)+list(cm.datad)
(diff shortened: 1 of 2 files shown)


jlstevens:
Thanks! Happy to test that large data is behaving properly again once this is merged.

jlstevens:
Glad to see this fixed!

Issue #4596, opened 2020-09-03
Spatial indexing for linked selections


jonmmease:
When performing linked selections on large tabular datasets, it would sometimes be useful to maintain spatial indices for subsets of columns in the dataset.

 - For pandas DataFrames, these could be RTree indices from spatialpandas (https://github.com/holoviz/spatialpandas).
 - For cudf, these could be quad-tree indices from cuspatial (https://docs.rapids.ai/api/cuspatial/stable/api.html#indexing)

These would make it possible to accelerate multi-dimensional range selections, and 2-dimensional geometric selections.

I'm picturing that we could create these indices on demand when selections are requested. But, what I'm trying to think through is, where could we store/cache them?  They would be associated with a specific DataFrame, so it would be nice to somehow store them alongside the DataFrame, or in the `Dataset`.  We might also want to extend the tabular data interface(s) to include spatial selection operations.

philippjfr:
>They would be associated with a specific DataFrame, so it would be nice to somehow store them alongside the DataFrame, or in the Dataset.

We have long kept the `Interface` types stateless but this kind of data would imo be a solid reason to allow making Interface instances which hold such data.

jonmmease:
> We have long kept the Interface types stateless but this kind of data would imo be a solid reason to allow making Interface instances which hold such data.


Yeah, that would be handy.  I also wonder if we'd want to consider moving the cache introduced in https://github.com/holoviz/holoviews/pull/4547 into these Interface instances.

jlstevens:
The `Interface` types are supposed to be stateless to make it easier to convert and cast between them. This isn't to say that stateful interfaces can't be supported, but I would prefer to not make them available by default i.e you could use them by specifying `datatype` explicitly. Alternatively, you could warn when converting in a lossy fashion but that might just be more confusing.

philippjfr:
If the state is explicitly restricted to extraneous metadata or optimization related, such as spatial indexing would, I wouldn't worry very much about that part being lossy.

PR #4695, opened 2020-11-20
Fix various issues related to data interfaces

philippjfr:
- `DaskInterface.persiste`was a typo
- The fix for masking on linked selections ensures that cuDFs don't error due to dropping to an array type rather than keeping the series type.

### DIFF
--- a/holoviews/core/data/dask.py
+++ b/holoviews/core/data/dask.py
@@ -75,7 +75,7 @@ class DaskInterface(PandasInterface):
         return dataset.clone(dataset.data.compute())
 
     @classmethod
-    def persiste(cls, dataset):
+    def persist(cls, dataset):
         return dataset.clone(dataset.data.persist())
 
     @classmethod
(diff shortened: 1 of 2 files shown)


philippjfr:
@jonmmease One thing I think would be very useful is testing the `SelectionDisplay._select` method against various data backends and various dtypes.

jonmmease:
> @jonmmease One thing I think would be very useful is testing the SelectionDisplay._select method against various data backends and various dtypes.

Not for this release, but I think I'd actually like to see if we can push most of this logic, including the caching, into the data interfaces.

It would take some rework, but there's already a sort of interface type case statement there:

```python
if dataset.interface.gridded:
    ...
    selection = ...
elif dataset.interface.multi:
    ...
    selection = ...
elif isinstance(element, (Curve, Spread)) and hasattr(dataset.interface, 'mask'):
    ...
    selection = ...
else:
    ...
    selection = ...
```
Basically, I'm wondering if we can push all of this into a new `select_expr` interface function.

And, if we change to using data interface instances (https://github.com/holoviz/holoviews/issues/4596), I'm thinking it would make sense to store this `select_expr` cache in the dataset interface object itself.

jonmmease:
Moving discussion to https://github.com/holoviz/holoviews/issues/4708

PR #4696, opened 2020-11-20
Add ibis count/count_nonzero aggregations, fix rowid handling

jonmmease:
This PR evolved a bit, here is what's currently included:
 - Add support for `len` and `np.count_nonzero` aggregation functions.
 - New test case covering aggregation functions.
 - Fixes to activate rowid-based operations with ibis 1.4.0.
 - Fixes to handle one vs zeros indexed rowids in OmniSci db vs SQLite.

----

**Original Description:**

This adds support for count aggregations using the `ibis` interface by mapping the python `len` function to `ibis.expr.operations.Count`.  Let me know if there's a better way to model this, but `len` is what I use with other interfaces to do a count aggregation.

I added a test to loop through the the aggregations, but there were issues with a few of them (which are commented out):

 - `cumsum` fails, and that actually makes sense to be since this doesn't produce a scalar value
 - all of the variance based operations fail in ibis. Here's a stack trace:

```
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1277, in _execute_context
    cursor, statement, parameters, context
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/sqlalchemy/engine/default.py", line 593, in do_execute
    cursor.execute(statement, parameters)
sqlite3.OperationalError: user-defined aggregate's 'finalize' method raised error
The above exception was the direct cause of the following exception:
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/IPython/core/interactiveshell.py", line 3418, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-1-433657610532>", line 1, in <module>
    result.dframe()
  File "/home/jmmease/PyDev/repos/holoviews/holoviews/core/data/__init__.py", line 184, in pipelined_fn
    result = method_fn(*args, **kwargs)
  File "/home/jmmease/PyDev/repos/holoviews/holoviews/core/data/__init__.py", line 1108, in dframe
    df = self.interface.dframe(self, dimensions)
  File "/home/jmmease/PyDev/repos/holoviews/holoviews/core/data/interface.py", line 42, in cached
    return method(*args, **kwargs)
  File "/home/jmmease/PyDev/repos/holoviews/holoviews/core/data/ibis.py", line 418, in dframe
    return dataset.data[dimensions].execute()
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/ibis/expr/types.py", line 202, in execute
    return execute(self, limit=limit, params=params, **kwargs)
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/ibis/client.py", line 352, in execute
    return backend.execute(expr, limit=limit, params=params, **kwargs)
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/ibis/client.py", line 215, in execute
    result = self._execute_query(query_ast, **kwargs)
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/ibis/client.py", line 220, in _execute_query
    return query.execute()
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/ibis/client.py", line 54, in execute
    self.compiled_sql, results=True, **kwargs
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/ibis/sql/alchemy.py", line 1212, in _execute
    return AlchemyProxy(self.con.execute(query))
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 2235, in execute
    return connection.execute(statement, *multiparams, **params)
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1011, in execute
    return meth(self, multiparams, params)
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/sqlalchemy/sql/elements.py", line 298, in _execute_on_connection
    return connection._execute_clauseelement(self, multiparams, params)
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1130, in _execute_clauseelement
    distilled_params,
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1317, in _execute_context
    e, statement, parameters, cursor, context
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1511, in _handle_dbapi_exception
    sqlalchemy_exception, with_traceback=exc_info[2], from_=e
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/sqlalchemy/util/compat.py", line 182, in raise_
    raise exception
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/sqlalchemy/engine/base.py", line 1277, in _execute_context
    cursor, statement, parameters, context
  File "/home/jmmease/anaconda3/envs/holoviews_dev/lib/python3.7/site-packages/sqlalchemy/engine/default.py", line 593, in do_execute
    cursor.execute(statement, parameters)
sqlalchemy.exc.OperationalError: (sqlite3.OperationalError) user-defined aggregate's 'finalize' method raised error
[SQL: SELECT t0."Gender", t0."Weight", t0."Height" 
FROM (SELECT t1."Gender" AS "Gender", _ibis_sqlite_sqrt(_ibis_sqlite_var_samp(t1."Height")) AS "Height", _ibis_sqlite_sqrt(_ibis_sqlite_var_samp(t1."Weight")) AS "Weight" 
FROM (SELECT t2."Gender" AS "Gender", t2."Weight" AS "Weight", t2."Height" AS "Height" 
FROM base.hetero AS t2) AS t1 GROUP BY t1."Gender" ORDER BY t1."Gender") AS t0
 LIMIT ? OFFSET ?]
[parameters: (10000, 0)]
(Background on this error at: http://sqlalche.me/e/13/e3q8)
```

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,6 +20,8 @@ jobs:
         exclude:
           - os: windows-latest
             python-version: 2.7
+          - os: macos-latest
+            python-version: 3.7
     timeout-minutes: 30
     defaults:
       run:
@@ -28,7 +30,7 @@ jobs:
       DESC: "Python ${{ matrix.python-version }} tests"
       HV_REQUIREMENTS: "unit_tests"
       PYTHON_VERSION: ${{ matrix.python-version }}
-      CHANS_DEV: "-c pyviz/label/dev -c bokeh"
+      CHANS_DEV: "-c pyviz/label/dev"
       CHANS: "-c pyviz"
       MPLBACKEND: "Agg"
       GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
@@ -61,6 +63,12 @@ jobs:
           git describe
           echo "======"
           conda list
+      - name: bokeh update
+        if: startsWith(matrix.python-version, 3.)
+        run: |
+          eval "$(conda shell.bash hook)"
+          conda activate test-environment
+          conda install "bokeh>=2.2"
       - name: matplotlib patch
         if: startsWith(matrix.python-version, 3.)
         run: |
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ holoviews.rc
 ghostdriver.log
 holoviews/.version
 .dir-locals.el
+.doit.db
+.vscode/settings.json
+holoviews/.vscode/settings.json
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ notifications:
 env:
   global:
     - PKG_TEST_PYTHON="--test-python=py37 --test-python=py27"
-    - CHANS_DEV="-c pyviz/label/dev -c bokeh"
+    - CHANS_DEV="-c pyviz/label/dev -c bokeh -c conda-forge"
     - CHANS="-c pyviz"
     - MPLBACKEND="Agg"
     - PYTHON_VERSION=3.7
@@ -69,7 +69,7 @@ jobs:
       install:
         - doit env_create $CHANS_DEV --python=$PYTHON_VERSION
         - source activate test-environment
-        - travis_wait 30 doit develop_install $CHANS_DEV -o $HV_REQUIREMENTS
+        - travis_wait 45 doit develop_install $CHANS_DEV -o $HV_REQUIREMENTS
         - if [ "$PYTHON_VERSION" == "3.6" ]; then conda uninstall matplotlib matplotlib-base --force; conda install $CHANS_DEV matplotlib=3.0.3 --no-deps; fi;
         - doit env_capture
         - hash -r
--- a/binder/environment.yml
+++ b/binder/environment.yml
@@ -32,3 +32,4 @@ dependencies:
   - bzip2
   - dask
   - scipy
+  - ibis-framework >= 1.3
--- a/examples/reference/elements/bokeh/Points.ipynb
+++ b/examples/reference/elements/bokeh/Points.ipynb
@@ -24,6 +24,7 @@
    "source": [
     "import numpy as np\n",
     "import holoviews as hv\n",
+    "import pandas as pd\n",
     "from holoviews import opts, dim\n",
     "hv.extension('bokeh')"
    ]
@@ -69,7 +70,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Although the simplest ``Points`` element simply mark positions in a two-dimensional space without any associated value this doesn't mean value dimensions (``vdims``) aren't supported. Here is an example with two additional quantities for each point, declared as the ``vdims`` ``'z'`` and ``'size'`` visualized as the color and size of the dots, respectively:"
+    "Although the simplest ``Points`` element simply marks positions in a two-dimensional space without any associated value, other value dimensions (``vdims``) are also supported. Here is an example with two additional quantities for each point, declared as the ``vdims`` ``'z'`` and ``'size'`` visualized as the color and size of the dots, respectively:"
    ]
   },
   {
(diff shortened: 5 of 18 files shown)


philippjfr:
Right, we should just remove cumsum and I'm happy to exclude the variance based aggregations from the tests for now as it's probably backend dependent and might work for some even if it doesn't work for sqlite which we're testing against. Is there a way to implement `np.count_nonzero` easily in addition to `len`?

jonmmease:
> ... np.count_nonzero easily in addition to len?

There actually was some special case handling for `count_nonzero`, but it was incorrect (identical to count), so I removed it in ffe3df9.

Looking at the Count operation in the ibis source code (I couldn't find it in the docs), it looks like there is some sort of `where` option available:

https://github.com/ibis-project/ibis/blob/a57347c4742d6b1353344141fc1752dca06f210e/ibis/expr/operations.py#L859-L864

But I'm not sure how to use it, or get a predicate in there.

jonmmease:
*Good news:* I think I actually got count_nonzero working in 7ab3ad3 

*Bad news:* To make sure count_nonzero was being tested I change a value in the test dataframe to 0, and whole bunch of other tests broke :frowning_face:  I'll take a look and see if the expected values just need to be updated.

jonmmease:
I think I fixed the expected test data values, but I'm getting unrelated failures locally having to do with the `has_rowid` check. 

I'm on the latest `ibis-framework` (1.4.0), and a bunch of tests are failing because the `has_rowid()` check is returning False. Do you know if that's expected?

jonmmease:
My test changes made a mess of things. I reverted them all. Currently `np.count_nonzero` is being tested, but not on a dataset that includes zeros.

jonmmease:
A few more observations:

1) rowid

We test for rowid support with:
```python
import ibis
return hasattr(ibis, "rowid")
```

I can't find any ibis version where this `ibis.rowid` exists.  Looking at https://github.com/ibis-project/ibis/pull/2345, it looks like we should check for rowid support with something like: `hasattr(ibis.expr.operations, "RowID")`.

2) After making the change above, a bunch of tests still fail because they compare unequal data because iloc is returning different results

For example, in `test_dataset_iloc_slice_rows_slice_cols`:

```python
self.table.dframe()
```
```
  Gender  Age  Weight  Height
0      M   10      15     0.8
1      M   16      18     0.6
2      F   12      10     0.8
```

iloc slice expression
```python
self.table.iloc[1:2, 1:3]
```
```
   Age  Weight
0   10      15
```

Based on https://github.com/ibis-project/ibis/blob/34849125ad29a54d62dffd50b2c6272b7db684ce/ibis/expr/api.py#L4369-L4370, it looks like sqlite supports rowid as a 1-indexed value.

And according to the ibis test code at https://github.com/ibis-project/ibis/blob/3779f7ba55e5d3384bafbad1af307fd0f271dfd8/ibis/tests/all/test_column.py#L5, it looks like rowid is 1-based for every ibis supported database except for omniscidb which is 0-based.

I think we'll need to adopt this logic of applying an offset of 1 unless using Omniscidb.

jonmmease:
> I think we'll need to adopt this logic of applying an offset of 1 unless using Omniscidb.

Hmm, but so far I can't find a way to lookup the current backend from an Ibis `TableExpr` object.

philippjfr:
@tonyfast @kcpevey any ideas about the rowid stuff?

jonmmease:
I think I got everything working in https://github.com/holoviz/holoviews/pull/4696/commits/5a78c3ba42f4f8e3afd1ea5990b7dfaf72d1ec64, but it would be good for someone to smoke test it with omniscidb if someone has a chance.

jonmmease:
Tests are still failing because CI is picking up ibis 1.3 instead of 1.4 where rowid is supported. Maybe because 1.4 isn't on the main anaconda channel yet?

jonmmease:
Ok, turns out I had broken the logic to skip tests when the rowid stuff isn't supported.  Before, this PR, that rowid test logic always returned false and so rowid operations were never enabled or tested.

Now I believe the rowid tests are working with ibis-framework 1.4 and are being skipped for ibis-framework < 1.4.

philippjfr:
Thanks, really appreciate you working on this!

PR #4697, opened 2020-11-20
Don't auto-import ibis

jonmmease:
Ibis was being auto-imported, causing an import error when using HoloViews without ibis installed.


philippjfr:
Thanks, I meant to remove that.

PR #4705, opened 2020-11-26
Fix datashader instability by disabling plot padding for Plotly RGB elements.

jonmmease:
The introduction of plot padding last release broke the default behavior of using `datashade` with the plotly backend in the notebook.  It resulted in the plot continually zooming out and crashing the kernel.

This  simple (and I think appropriate) fix is to disable padding, by default, for the RGB elements with the plotly backend. Even apart from datashader, I think zero padding is a better default for RGB images as it's a bit awkward for there to be an empty border around them.


jbednar:
Looks to me like this was just an oversight, because the Matplotlib and Bokeh backends already don't have padding for RGB or Image or Raster elements. I assume the same change should be made to Image and Raster?

jonmmease:
Thanks, that makes sense. Done in 67fcb26

PR #4715, opened 2020-11-30
[WIP] add macos 3.7

kcpevey:


Hoxbro:
For fun I was playing around with this and got it to work, see https://github.com/Hoxbro/holoviews/pull/1.

I don't know if every change is necessary, but I think it could be a good start fixing the problem. The first thing I will try is just to add conda-forge to channels. 

FYI I also observed that macos 3.6 uses a beta version of panel `0.10.0a35`.

kcpevey:
@Hoxbro Thanks for figuring that out! I was looking into memory issues, but it seems that was unrelated. Not sure why macos 3.6 is using a beta version of panel. That will be worth looking into - @philippjfr any ideas why that would be? Maybe I've missed passing an env var to the doit commands somewhere?

philippjfr:
>That will be worth looking into - @philippjfr any ideas why that would be?

All builds should use dev versions of Panel.

Hoxbro:
To clarify my earlier comment it uses an OLD dev version of Panel. I would assume it should be on `0.10.2` not `0.10.0`.

Maybe it is a dependencies issue, which maybe can be fixed by adding conda-forge to channel.

philippjfr:
Ah, thanks, I missed that. That is quite odd, I'll have to dig in a bit.

maximlt:
The CI is running on MacOs/Python 3.7 so this can be closed.

PR #4719, opened 2020-12-01
Allow supplying named cmap to (data)shade

philippjfr:


holoviews/operation/datashader.py
```diff
@@ -1292,6 +1293,12 @@ def _process(self, element, key=None):
         elif isinstance(self.p.cmap, Callable):
             colors = [self.p.cmap(s) for s in np.linspace(0, 1, 256)]
             shade_opts['cmap'] = map(self.rgb2hex, colors)
+        elif isinstance(self.p.cmap, basestring):
+            if self.p.cmap.startswith('#') or self.p.cmap in color_lookup:
+                shade_opts['cmap'] = self.p.cmap
+            else:
+                from ..plotting.util import process_cmap
+                shape_opts['cmap'] = process_cmap(self.p.cmap)
```
jlstevens, line 1301: `shade_opts`

jlstevens: Fixed in datashade_cmap

holoviews/operation/datashader.py
```diff
@@ -1292,6 +1293,12 @@ def _process(self, element, key=None):
         elif isinstance(self.p.cmap, Callable):
             colors = [self.p.cmap(s) for s in np.linspace(0, 1, 256)]
             shade_opts['cmap'] = map(self.rgb2hex, colors)
+        elif isinstance(self.p.cmap, basestring):
+            if self.p.cmap.startswith('#') or self.p.cmap in color_lookup:
+                shade_opts['cmap'] = self.p.cmap
+            else:
+                from ..plotting.util import process_cmap
+                shade_opts['cmap'] = process_cmap(self.p.cmap)
```
jbednar, line 1301: Seems like this will give confusing error messages if someone supplies a color not in the database. E.g. if they try "lightturquoise" instead of "paleturquoise", they'll now get a message "Supplied cmap lightturquoise not found among matplotlib, bokeh, or colorcet colormaps" instead of the previous message "Unknown color: 'lightturquoise'. But I guess it's more likely that people will pass a cmap by name than an individual color like this, so I suppose it's ok.

PR #4723, opened 2020-12-02
Updated Large Data

jbednar:
As mentioned in https://github.com/holoviz/holoviews/pull/4567, the Large Data notebook still needed a bit of work. I've done that in this PR, but there are a few problems:

- [x] 1. For `decimate(points) + rasterize(points) + rasterize(paths)`, even if I leave off `+ rasterize(paths)`, the colormapping behavior of `rasterize(points)` differs when `decimate(points)` is included. Without `decimate(points)`, the behavior looks reasonable, with colors rescaling to cover the full colormap at each zoom level. With `decimate(points)` in the layout, zooming makes the rasterized data get very light. Why should `decimate` affect a rasterized plot in any way?
- [x] 2. The example in Colormapping using `clim` does not currently show what it purports to show; both plots shown behave the same on zoom and pan; the one with clim defined doesn't do anything different. This is a new example, but I couldn't get the previous Image-based example to work with rasterize() either, even when I forced the z axis to have a different Dimension name, so I don't think it's specific to this example. I don't actually know how to change the x_y Count dimension names, but I assume that won't help since it didn't in the original example. It could still be me doing something wrong, though. In the old example, the colormap rescaled properly on each individual plot, just not when in a layout. On this version, the colormap seems fixed regardless of viewport, which is confusing me. Once it works, it would presumably need a GIF to illustrate it.
- [x] 3. The `xlim=(0.1,0.2), ylim=(0.1,0.2)))` example also presumably needs a gif, as the static website export will otherwise be broken.
- [x] 4. The nodata=None illustration is not working; both plots are identical. Again, am I doing something wrong?
- [x] 5. I haven't tested Large_Data with released versions of Bokeh; does it work reasonably when the new colorbar ticking is not available?
- [x] 6. Is it really `clim` on Bokeh and `clims` on the mpl backend and on shade()? Damn.
- [x] 7. The padding settings don't look right, in the comparison between datashaded and native (mpl) hv plots. Shouldn't there be padding for datashaded non-gridded types, just as for the corresponding non-datashaded non-gridded types (Path, Points, etc.)? 
- [x] 8. And shouldn't padding be 0 for non-datashaded Contour plots? The gap looks bad in regular hv plots; the datashaded ones look much better. In any case, can we make the padding behave the same whether or not we use rasterize?

image
image

### DIFF
--- a/examples/user_guide/15-Large_Data.ipynb
+++ b/examples/user_guide/15-Large_Data.ipynb
@@ -331,7 +331,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "hv.Layout([rasterize(hv.Points(df), vdim_prefix='').redim.nodata(Count=n).opts(**ropts).relabel(\"nodata=\"+str(n))\n",
+    "hv.Layout([rasterize(hv.Points(df), vdim_prefix='').redim.nodata(Count=n)\\\n",
+    "           .opts(**ropts, cnorm=\"eq_hist\").relabel(\"nodata=\"+str(n))\n",
     "           for n in [0, None]])"
    ]
   },
@@ -350,7 +351,7 @@
     "\n",
     "Dynamic spreading is typically more useful for interactive plotting, because it adjusts depending on how close the datapoints are to each other on screen. As of Datashader 0.12, both types of spreading are supported for both `rasterize()` and `shade()`, but previous Datashader versions only support spreading on the RGB output of `shade()`.\n",
     "\n",
-    "As long as you have Datashader 0.12 or later, you can compare the results in the two zoomed-in plots below, then zoom out to see that the plots are the same when points are clustered together to form a distribution:"
+    "As long as you have Datashader 0.12 or later, you can compare the results when you zoom the two plots below; when you zoom in far enough you should be able to see that the in the two zoomed-in plots below, then zoom out to see that the plots are the same when points are clustered together to form a distribution. (If running a live notebook; remove the semicolon so that you see the live output rather than the saved GIF.)"
    ]
   },
   {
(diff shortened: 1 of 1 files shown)


philippjfr:
> 4. The nodata=None illustration is not working; both plots are identical. Again, am I doing something wrong?

You cannot override the data level nodata at the plot option level.

philippjfr:
>7. The padding settings don't look right, in the comparison between datashaded and native (mpl) hv plots. Shouldn't there be padding for datashaded non-gridded types, just as for the corresponding non-datashaded non-gridded types (Path, Points, etc.)?

How would it know? Only the option transfer could achieve this.

philippjfr:
>8. And shouldn't padding be 0 for non-datashaded Contour plots? The gap looks bad in regular hv plots; the datashaded ones look much better. In any case, can we make the padding behave the same whether or not we use rasterize?

Really depends on the contents of Contours but you're right that for iso-contours extracted using the contours operation it makes more sense.

philippjfr:
>2. The example in Colormapping using clim does not currently show what it purports to show...On this version, the colormap seems fixed regardless of viewport, which is confusing me

Not surprising to me, the dynamic rescaling was always just an artifact of datashading because the RGB had no way of persisting ranges. We do not and have never enabled dynamic color rescaling by default anywhere.

philippjfr:
>Not surprising to me, the dynamic rescaling was always just an artifact of datashading because the RGB had no way of persisting ranges. We do not and have never enabled dynamic color rescaling by default anywhere.

That being said, if you do enable it with `axiswise=True` and `framewise=True` it still doesn't work because there's a bug :(

jlstevens:
To expand on point 4 where Philipp says:

> You cannot override the data level nodata at the plot option level.

What this means is you should be able to do it with `element.redim.nodata` but then the problem is using the vdim name...that can probably be solved if you set `vdim_prefix=''` in `rasterize` so you should be able to do something like

```python
hv.Layout([rasterize(hv.Points(df), vdim_prefix='').redim.nodata(Count=n).opts(**ropts).relabel("nodata="+str(n)
 for n in [0, None]])
```

jlstevens:
I've ticked point 2 though there is a new issue with the plots not being linked when they should be.

jbednar:
I've marked the padding issues as complete, but only because I forced padding=10 on the spreadable plots, and ignored the fact that `contours` should have no padding with or without Datashader, for now. Apart from the issue with plots not being linked, seems good to go.

jlstevens:
Looks good! Merging.

PR #4724, opened 2020-12-02
Fix range calculation in framewise/axiswise configuration

philippjfr:
This fixes a bug related to range calculations but only partially. The problematic situation is that when a stream triggers an update we only recompute the particular subplot(s) that are driven by that stream. This messes with range calculations because those work across plots and we have no way to update the overall range without triggering the range calculation at the top level plot. 

This PR fixes the situation where both axiswise and framewise are set because in that situation we do not have to recompute across all plots. However `framewise=True, axiswise=False` is still a problem and the only idea I have to address the problem is for top level plots to pass all the individual ranges from each element down to each subplot so that they can recompute the overall ranges by simply updating the ranges of the elements that are being updated. That is definitely more effort though.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1447,10 +1447,13 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         """
         Resets RangeXY streams if norm option is set to framewise
         """
-        if self.overlaid:
+        # Temporarily reverts this fix (see https://github.com/holoviz/holoviews/issues/4396)
+        # This fix caused PlotSize change events to rerender
+        # rasterized/datashaded with the full extents which was wrong
+        if self.overlaid or True:
             return
         for el, callbacks in self.traverse(lambda x: (x.current_frame, x.callbacks)):
-            if el is None:
+            if el is None:W
                 continue
             for callback in callbacks:
                 norm = self.lookup_options(el, 'norm').options
@@ -1466,6 +1469,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         """
         self._reset_ranges()
         reused = isinstance(self.hmap, DynamicMap) and (self.overlaid or self.batched)
+        self.prev_frame =  self.current_frame
         if not reused and element is None:
             element = self._get_frame(key)
         elif element is not None:
@@ -2367,6 +2371,7 @@ class OverlayPlot(GenericOverlayPlot, LegendPlot):
         """
         self._reset_ranges()
         reused = isinstance(self.hmap, DynamicMap) and self.overlaid
+        self.prev_frame =  self.current_frame
         if not reused and element is None:
             element = self._get_frame(key)
         elif element is not None:
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1453,7 +1453,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         if self.overlaid or True:
             return
         for el, callbacks in self.traverse(lambda x: (x.current_frame, x.callbacks)):
-            if el is None:W
+            if el is None:
                 continue
             for callback in callbacks:
                 norm = self.lookup_options(el, 'norm').options
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1447,7 +1447,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         """
         Resets RangeXY streams if norm option is set to framewise
         """
-        # Temporarily reverts this fix (see https://github.com/holoviz/holoviews/issues/4396)
+        # Skipping conditional to temporarily revert fix (see https://github.com/holoviz/holoviews/issues/4396)
         # This fix caused PlotSize change events to rerender
         # rasterized/datashaded with the full extents which was wrong
         if self.overlaid or True:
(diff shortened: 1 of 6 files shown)


jlstevens:
@philippjfr I realize this is a partial fix but the tests are green and I'm happy to merge this if you think this PR is ready.

philippjfr:
Let me try to expand the scope of this fix a bit.

holoviews/plotting/bokeh/element.py
```diff
@@ -1447,10 +1447,13 @@ def _reset_ranges(self):
...
         """
-        if self.overlaid:
+        # Temporarily reverts this fix (see https://github.com/holoviz/holoviews/issues/4396)
+        # This fix caused PlotSize change events to rerender
+        # rasterized/datashaded with the full extents which was wrong
+        if self.overlaid or True:
             return
         for el, callbacks in self.traverse(lambda x: (x.current_frame, x.callbacks)):
-            if el is None:
+            if el is None:W
```
jlstevens, line 1456: SyntaxError I think

philippjfr, line 1456:
```suggestion
            if el is None:
```

holoviews/plotting/bokeh/element.py
```diff
@@ -1447,7 +1447,10 @@ def _reset_ranges(self):
         """
         Resets RangeXY streams if norm option is set to framewise
         """
-        if self.overlaid:
+        # Temporarily reverts this fix (see https://github.com/holoviz/holoviews/issues/4396)
+        # This fix caused PlotSize change events to rerender
+        # rasterized/datashaded with the full extents which was wrong
+        if self.overlaid or True:
```
jlstevens, line 1453: hardcode `or True`? Doesn't that mean this branch always executes?

jlstevens: Maybe

# Skipping conditional to temporarily revert this fix ...

holoviews/plotting/plot.py
```diff
@@ -604,6 +604,7 @@ def compute_ranges(self, obj, key, ranges):
         over the whole animation) and finally compute the dimension
         ranges in each group. The new set of ranges is returned.
         """
+        prev_frame = getattr(self, 'prev_frame', None)
```
jlstevens, line 607: Not a big issue but `self.prev_frame` should probably be initialized to `None` in the appropriate constructor(s)...

holoviews/plotting/plot.py
```diff
@@ -687,9 +689,34 @@ def _get_norm_opts(self, obj):
                           if not any(spec[:i] in norm_opts.keys() for i in range(1, 4))})
         return norm_opts
 
+    @classmethod
+    def _merge_group_ranges(cls, ranges):
```
jlstevens, line 693: As a classmethod that doesn't even use `cls`, it should be easy to set up some unit tests for this function (could even be a utility tbh).

holoviews/plotting/plot.py
```diff
@@ -770,14 +797,17 @@ def _compute_group_range(cls, group, elements, ranges, framewise, robust, top_le
...
+                            'id': [], 'data': [], 'hard': [], 'soft': []
+                        }
 
                     if factors is not None:
                         if 'factors' not in group_ranges[dim_name]:
                             group_ranges[dim_name]['factors'] = []
                         group_ranges[dim_name]['factors'].append(factors)
                     else:
                         group_ranges[dim_name]['data'].append(drange)
+                    group_ranges[dim_name]['id'].append(id(el))
```
jlstevens, line 810: Guess this plotting class will accumulate ranges for all the elements it processes. Doubt it would cause much of a memory leak to worry about though...

holoviews/plotting/plot.py
```diff
@@ -825,44 +858,49 @@ def _compute_group_range(cls, group, elements, ranges, framewise, robust, top_le
...
-                        # Try to keep the same dtype
-                        expanded = np.array(expanded, dtype=dtype)
-                    except Exception:
-                        pass
-                dranges['factors'] = util.unique_array(expanded)
+            dranges = cls._merge_group_ranges(values)
             dim_ranges.append((gdim, dranges))
-        if prev_ranges and not (framewise and top_level):
+
+        # Merge local ranges into global range dictionary
```
jlstevens, line 867: This is the hardest bit to follow. As I find this hard to reason about the best thing to do is make sure there are lots of  decent unit tests!

jlstevens:
@philippjfr made a few comments but overall looks fine to me.

holoviews/plotting/plot.py
philippjfr: Don't really want it on all plots, only makes sense on element/overlay plots but this is used even on layout/grid plots.

philippjfr:
Let's see what the tests say.

jlstevens:
Failures on 2.7 but apparently not on 3.6/3.7:

```
FAIL: test_layout_framewise_matching_norm_update (holoviews.tests.plotting.bokeh.testlayoutplot.TestLayoutPlot)
  File "/home/runner/work/holoviews/holoviews/holoviews/tests/plotting/bokeh/testlayoutplot.py", line 67, in test_layout_framewise_matching_norm_update
    self.assertEqual(img1_cmapper.high, 4)
  File "/home/runner/work/holoviews/holoviews/holoviews/element/comparison.py", line 79, in assertEqual
    asserter(first, second)
  File "/home/runner/work/holoviews/holoviews/holoviews/element/comparison.py", line 56, in simple_equality
    raise cls.failureException(msg or standardMsg)
AssertionError: 40 != 4
```

jlstevens:
The tests are now passing, Python 2 included. Merging.

PR #4726, opened 2020-12-02
Fix for inplace dtype multiplication issue

jlstevens:
Simple fix to address inplace int/float multiplication issue in `SideHistogram`.


jlstevens:
Tests are now passing and this is a trivial fix. Merging.

PR #4727, opened 2020-12-02
Fix coloring by count on SideHistogramPlot

philippjfr:


jlstevens:
Thanks. Merging and tagging the next dev.

PR #4729, opened 2020-12-02
Fix RGB.load_image range

philippjfr:


jlstevens:
Looks good to me. Merging.

jlstevens:
Reverting as I think the real issue is that in the switch from matplotlib to PIL, the data changed from being floats in the range [0,1] to ints. The correct fix is to normalize the data from PIL.

Issue #4760, opened 2020-12-16
Conflicting info in docs for Points and Scatter


rchurt:
In the docs for [Scatter](http://holoviews.org/reference/elements/bokeh/Scatter.html), it says: Note: Although the Scatter element is superficially similar to the Points element (they can generate plots that look identical), the two element types are semantically quite different: Points are used to visualize data where the y variable is dependent. This semantic difference also explains why the histogram generated by hist call above visualizes the distribution of a different dimension than it does for Points.

In the docs for [Points](http://holoviews.org/reference/elements/bokeh/Points.html), it says: Note: Although the Scatter element is superficially similar to the Points element (they can generate plots that look identical), the two element types are semantically quite different. The fundamental difference is that Scatter is used to visualize data where the y variable is dependent. This semantic difference also explains why the histogram generated by hist call above visualizes the distribution of a different dimension than it does for Scatter.

So two points of confusion:
1. Each says the other is "used to visualize data where the y variable is dependent"
2. The statement "This semantic difference also explains why the histogram generated by hist call above visualizes the distribution of a different dimension than it does for [Points or Scatter]" is wrong because in both cases `hist` displays the `z` dimension.

jbednar:
Whoa, those are some old docs, and that glaring error must have been sitting there confusing people for many years now with apparently no one ever realizing it made no sense. Scatter is the one where y is dependent (i.e. one kdim (x)), and Points is one where y is independent (i.e. two kdims (x,y)). A PR on the docs would help clear that up! 

I have no explanation for point 2; any ideas @jlstevens and @philippjfr?

rchurt:
As a broader point, I think it would be useful to make significant changes to the `Scatter` docs, since right now they look almost identical to `Points`, but as the inconsistency mentioned above points out, these are two different functions. My vote would be to add some examples to the docs for `Scatter` modified from the docs for `Bars`, showing categorical data (where it's clear that y is dependent).

jbednar:
That would be great!

jlstevens:
Two mistaken characters make such a difference! As Jim says, it should be:

> Points are used to visualize data where the y variable is *in*dependent.

As for the second issue, I suspect it is also a case of old docs not being updated when they should have been.

Thanks for reporting this!

jlstevens:
Looks like the above referenced PR has addressed this issue. Closing but feel free to comment if you think it needs to be reopened.

PR #4764, opened 2020-12-17
Clarified Points and Scatter notebooks

jbednar:
Fixes issue 1 from https://github.com/holoviz/holoviews/issues/4760 . Still need to address issue 2.


examples/reference/elements/bokeh/Points.ipynb
```diff
@@ -147,10 +147,9 @@
...
+    "**Note**: Although the  ``Scatter`` element is superficially similar to the [``Points``](./Points.ipynb) element (they can generate plots that look identical), the two element types are semantically quite different. The fundamental difference is that [Scatter](./Scatter.ipynb) is used to visualize data where the *y* variable is *dependent*, unlike ``Points``. This semantic difference also explains why the histogram generated by the ``hist`` call above visualizes the distribution of a different dimension than it does for [``Scatter``](./Scatter.ipynb).\n",
     "\n",
+    "This difference means that ``Points`` elements can easily be overlaid with other elements that express independent variables in two-dimensional space, such as [``Raster``](./Raster.ipynb) types like [``Image``](./Image.ipynb). Conversely, ``Scatter`` expresses a dependent relationship between *x* and *y* and thus overlays easily with ``Chart`` types such as [``Curve``](./Curve.ipynb).\n",
```
philippjfr, line 152: While this is true conceptually people might be confused by this since scatter overlays on easter/image just fine.

philippjfr:
Thanks, one comment which you can ignore.

examples/reference/elements/bokeh/Points.ipynb
jbednar: Yeah; maybe you can think of a better way to say that. "most naturally"?

philippjfr: "Most naturally" works for me, "easily" was my main objection.

jbednar:
For point 2, my guess is that all the examples used to say `.hist()` and thus , but someone (potentially me!) noticed that the histograms didn't match between Points and Scatter, so they made Scatter use `.hist('z')` and thus made the histograms match.  I've reverted that, and now .hist() selects different dimensions on the two object types (always the first value dimension, i.e. 'y' on Scatter and 'z' on Points) .

I also removed any mention of .hist() on the Plotly versions, because those examples don't actually use histograms, presumably because AdjointLayouts are not supported for that backend.

examples/reference/elements/bokeh/Points.ipynb
jbednar: I'll update to "most naturally.

examples/reference/elements/bokeh/Points.ipynb
```diff
@@ -70,7 +70,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Although the simplest ``Points`` element simply marks positions in a two-dimensional space without any associated value, other value dimensions (``vdims``) are also supported. Here is an example with two additional quantities for each point, declared as the ``vdims`` ``'z'`` and ``'size'`` visualized as the color and size of the dots, respectively:"
+    "Although the simplest ``Points`` element simply marks positions in a two-dimensional space without any associated value, value dimensions (``vdims``)  are also supported. Here is an example with two additional quantities for each point, declared as the ``vdims``s ``z`` and ``size`` (visualized as the color and size of the dots, respectively):"
```
jbednar, line 73:
```suggestion
    "Although the simplest ``Points`` element simply marks positions in a two-dimensional space without any associated value, value dimensions (``vdims``) are also supported. Here is an example with two additional quantities for each point, declared as the ``vdims``s ``z`` and ``size`` (visualized as the color and size of the dots, respectively):"
```

examples/reference/elements/plotly/Points.ipynb
```diff
@@ -66,7 +66,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "Although the simplest ``Points`` element simply mark positions in a two-dimensional space without any associated value this doesn't mean value dimensions aren't supported. Here is an example with two additional quantities for each point, declared as the ``value_dimension``s *z* and &alpha; visualized as the color and size of the dots, respectively:"
+    "Although the simplest ``Points`` element simply marks positions in a two-dimensional space without any associated value, value dimensions (``vdims``)  are also supported. Here is an example with two additional quantities for each point, declared as the ``vdims``s ``z`` and ``size`` (visualized as the color and size of the dots, respectively):"
```
jbednar, line 69:
```suggestion
    "Although the simplest ``Points`` element simply marks positions in a two-dimensional space without any associated value, value dimensions (``vdims``) are also supported. Here is an example with two additional quantities for each point, declared as the ``vdims``s ``z`` and ``size`` (visualized as the color and size of the dots, respectively):"
```

jbednar:
Ok, this is ready for review and merge. It makes me go blind to try to read through these six nearly identical but subtly different notebooks, but I *think* each of them has the correct changes now.

jbednar:
I actually tried out Plotly with .hist(), and I think that was missing before because used to fail but it's working fine now, so I made the Plotly examples include .hist() to match the Bokeh and Matplotlib ones. I also removed other seemingly unnecessary differences like colormaps and slice ranges, so that all the notebooks now show similar stuff.

Oddly, the Bokeh Scatter notebook is failing with `KeyError: "One or more dimensions in the expression dim('z') could not resolve on ':Histogram   [y]   (y_count)'. Ensure all dimensions referenced by the expression are present on the supplied object."` , for the same code that works on the Matplotlib Scatter notebook. @philippjfr, can you take a look? Either I'm confused or there is a bug.

philippjfr:
@jbednar It was actually your fault 😄 The problem is that you dropped the `'z'` from `.hist` call, so the histogram was on the wrong dimension. The matplotlib version that "worked" was still coloring by the wrong dimension so wasn't really any good.

jbednar:
That was deliberate; it was _supposed_ to be a histogram on 'y'. Reverting that is not a fix; how can you get a histogram on 'y'?

PR #4776, opened 2021-01-03
Fixes for index based linked selections

philippjfr:
Fixes two issues with index based linked selections:

1. Ensures empty selection is not applied
2. Ensures that index cols are present on the selected element

Example:

```python
import hvplot.pandas
from bokeh.sampledata.autompg import autompg

scatter = autompg.hvplot.scatter('weight', 'accel')
table = autompg.hvplot.table()

link_selections(scatter + table, index_cols=['name', 'yr'])
```

Screen Shot 2021-01-03 at 1 36 57 PM

### DIFF
--- a/holoviews/element/selection.py
+++ b/holoviews/element/selection.py
@@ -26,15 +26,18 @@ class SelectionIndexExpr(object):
 
     def _get_index_selection(self, index, index_cols):
         self._index_skip = True
+        if not index:
+            return None, None, None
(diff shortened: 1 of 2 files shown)


jlstevens:
@philippjfr Would you like me to rebase this PR so we can get updated test results? The errors look like they may well be related to these changes but I think a rebase wouldn't hurt to try anyway...

philippjfr:
I rebased it a day ago, the error is real.

PR #4782, opened 2021-01-08
Update tap example

kcpevey:
Implementing Tap is something I struggle to get right every time. This is partially because the example is limited. In this PR I've added a little more detail and two simplified examples (one using Panel, one without). 

The additional examples came from [this](https://discourse.holoviz.org/t/example-of-using-holoviews-tapstream-with-panel/166) discourse discussion.


kcpevey:
Also, the current version of this notebook is throwing a warning `Log color mapper lower bound <= 0 and will not render corrrectly. Ensure you set a positive lower bound on the color dimension or using the `clim` option.` 

So I've updated the current example to fix that.

http://holoviews.org/reference/streams/bokeh/Tap.html

examples/reference/streams/bokeh/Tap.ipynb
```diff
@@ -39,23 +112,32 @@
...
+    "                     label='Average Weekly Measles Incidence').select(Year=(1928, 2002))\n",
     "\n",
     "# Declare Tap stream with heatmap as source and initial values\n",
     "posxy = hv.streams.Tap(source=heatmap, x=1951, y='New York')\n",
     "\n",
     "# Define function to compute histogram based on tap location\n",
     "def tap_histogram(x, y):\n",
     "    return hv.Curve(dataset.select(State=y, Year=int(x)), kdims='Week',\n",
-    "                   label='Year: %s, State: %s' % (x, y))\n",
+    "                   label=f'Year: {x}, State: {y}')\n",
```
jbednar, line 123: I forget; is HoloViews py2 only now? If not this change should presumably be reverted.

examples/reference/streams/bokeh/Tap.ipynb
```diff
@@ -20,13 +20,86 @@
...
+    "stream = hv.streams.Tap(source=points, x=np.nan, y=np.nan)\n",
+    "\n",
+    "# we use panel depends to callback the location function every time the x and y values on the stream change.\n",
+    "@pn.depends(stream.param.x, stream.param.y)\n",
+    "def location(x, y):\n",
+    "    \"\"\"Display pane showing the x and y values\"\"\"\n",
+    "    return pn.pane.Str(f'Click at {x:.2f}, {y:.2f}', width=200)\n",
+    "\n",
+    "# create a container for the plot and the text display\n",
+    "layout = pn.Row(points, location)\n",
```
jbednar, line 91: It's up to @philippjfr , but my personal vote is to encourage `pn.bind` rather than `@pn.depends`:

```suggestion
    "# make a function that displays the location when called.\n",
    "def location(x, y):\n",
    "    \"\"\"Display pane showing the x and y values\"\"\"\n",
    "    return pn.pane.Str(f'Click at {x:.2f}, {y:.2f}', width=200)\n",
    "\n",
    "# Display the points and the function output, updated\n",
    "# whenever the stream values change\n",
    "layout = pn.Row(points, pn.bind(location, x=stream.param.x, y=stream.param.y)\n",
```

jbednar (approved):
Thanks!

examples/reference/streams/bokeh/Tap.ipynb
kcpevey: ah. I just couldn't help myself... I'll revert.

examples/reference/streams/bokeh/Tap.ipynb
kcpevey: I have never heard of `pn.bind`. Interesting... Are there docs for that?

philippjfr: It's recent and sadly not documented. Had to edit the suggestion because it was missing a closing `)`.

jlstevens: I agree: it would be better to use `pn.bind` though really here `param.bind` seems more appropriate to me (same thing but doesn't support panel widgets, which aren't being used in this example).

philippjfr: param.bind does not yet exist.

jlstevens: Right. We should get it into a param dev release at least.

examples/reference/streams/bokeh/Tap.ipynb
kcpevey: this has been reverted

kcpevey:
@jbednar @philippjfr Comments have been addressed. I believe this is ready to be merged.

examples/reference/streams/bokeh/Tap.ipynb
jbednar: It doesn't _look_ reverted; was that in a separate PR?

Issue #4781, opened 2021-01-08
Supporting panel widgets directly as streams


jlstevens:
HoloViews supports turning parameters directly into streams such that `DynamicMap`s update when parameter values change (the parameters are promoted to a `Param` stream internally). Panel widgets are also based on param and I was hoping to use this support to drive a `DynamicMap` with panel widgets. Unfortunately, as soon as you try to use more than one panel widget this way, you encounter name clashes as all widget value parameters have the same name (specifically `widget.param.value`).

Given that parameters can be used as streams, I would like to use param widgets directly as well, using the widget name (or some sort of identifier label which I think would be useful in Panel more generally) as follows:

```python
import panel as pn
import holoviews as hv
hv.extension('bokeh')

checkbox = pn.widgets.Checkbox(name='My Checkbox')
selector = pn.widgets.Select(options=['a','b','c'], name='My Selector')

def callback(my_selector, my_checkbox):
    return hv.Curve([1,2,3])

hv.DynamicMap(callback, streams=[checkbox, selector])
```

This would be orthogonal to other approaches such as `pn.bind` (which should also be available from param, without special widget support). For me this approach would feel very natural and I don't think it precludes any other of the available approaches.

philippjfr:
I'm definitely supportive of finding a cleaner way to map parameterized objects to streams which makes it easy to rename parameter values. That said I'm not entirely excited about the escaping required to support arbitrary names and how a user would figure out the correct escaping for more complex names - it's very implicit, and I much prefer the explicit approaches we have developed and promoted for the past two years. Also note that this proposal is not backward compatible since right now using a parameterized as a stream automatically expands all of its parameters, which I agree is probably not all that useful or desirable, but is still worth noting.

For the sake of completeness I'll list current approaches to this problem, which are imo all much more explicit and at least from my perspective preferable:

```python
# pn.depends
@pn.depends(a=selector, b=checkbox)
def callback(a, b):
    return hv.Curve([1, 2, 3])

hv.DynamicMap(callback)

# pn.bind
def callback(a, b):
    return hv.Curve([1, 2, 3])

hv.DynamicMap(pn.bind(callback, a=selector, b=checkbox))

# Params stream
from holoviews.streams import Params

checkbox = pn.widgets.Checkbox(name='My Checkbox')
selector = pn.widgets.Select(options=['a','b','c'], name='My Selector')

def callback(my_selector, my_checkbox):
    return hv.Curve([1,2,3])

[...]

PR #4787, opened 2021-01-13
Supporting streams parameter as a dictionary

jlstevens:
This implements an idea raised in https://github.com/holoviz/holoviews/issues/4781 which is to support dictionary values in the `streams` parameter. This currently works:

</img>

### TODO

- [x] Detect parameters from linked streams to set up bokeh callbacks.
- [x]  Support class parameters.
- [x] Unit tests

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -19,7 +19,7 @@ from .layout import Layout, AdjointLayout, NdLayout, Empty
 from .ndmapping import UniformNdMapping, NdMapping, item_check
 from .overlay import Overlay, CompositeOverlay, NdOverlay, Overlayable
 from .options import Store, StoreOptions
-from ..streams import Stream, Params, LinkedStream, streams_list_from_dict
+from ..streams import Stream, Params, streams_list_from_dict
 
 
 
@@ -947,7 +947,7 @@ class DynamicMap(HoloMap):
                 stream.source = self
             if isinstance(stream, Params):
                 for p in stream.parameters:
-                    if isinstance(p.owner, LinkedStream) and p.owner.source is None:
+                    if isinstance(p.owner, Stream) and p.owner.source is None:
                         p.owner.source = self
 
         self.periodic = periodic(self)
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -19,7 +19,7 @@ from .layout import Layout, AdjointLayout, NdLayout, Empty
 from .ndmapping import UniformNdMapping, NdMapping, item_check
 from .overlay import Overlay, CompositeOverlay, NdOverlay, Overlayable
 from .options import Store, StoreOptions
-from ..streams import Stream, streams_list_from_dict
+from ..streams import Stream, Params, LinkedStream, streams_list_from_dict
 
 
 
@@ -945,6 +945,11 @@ class DynamicMap(HoloMap):
         for stream in self.streams:
             if stream.source is None:
                 stream.source = self
+            if isinstance(stream, Params):
+                for p in stream.parameters:
+                    if isinstance(p.owner, LinkedStream) and p.owner.source is None:
+                        p.owner.source = self
+
         self.periodic = periodic(self)
 
     @property
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -882,13 +882,13 @@ class DynamicMap(HoloMap):
         directly, otherwise it will be automatically wrapped in one.""")
 
     streams = param.List(default=[], constant=True, doc="""
-       List of Stream instances to associate with the DynamicMap or a
-       dictionary mapping parameters or panel widgets to callback
-       argument names. In the list case, the set of parameter values
-       across these streams will be supplied as keyword arguments to the
-       callback when the events are received, updating the streams. When
-       a dictionary is supplied, it will be automatically converted to
-       the equivalent list format.""" )
+       List of Stream instances to associate with the DynamicMap. The
+       set of parameter values across these streams will be supplied as
+       keyword arguments to the callback when the events are received,
+       updating the streams. Can also be supplied as a dictionary that
+       maps parameters or panel widgets to callback argument names that
+       will then be automatically converted to the equivalent list
+       format.""")
 
     cache_size = param.Integer(default=500, doc="""
        The number of entries to cache for fast access. This is an LRU
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -881,7 +881,7 @@ class DynamicMap(HoloMap):
         If the callable is an instance of Callable it will be used
         directly, otherwise it will be automatically wrapped in one.""")
 
-    streams = param.ClassSelector(default=[], class_=(dict, list), constant=True, doc="""
+    streams = param.List(default=[], constant=True, doc="""
        List of Stream instances to associate with the DynamicMap or a
        dictionary mapping parameters or panel widgets to callback
        argument names. In the list case, the set of parameter values
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -19,7 +19,7 @@ from .layout import Layout, AdjointLayout, NdLayout, Empty
 from .ndmapping import UniformNdMapping, NdMapping, item_check
 from .overlay import Overlay, CompositeOverlay, NdOverlay, Overlayable
 from .options import Store, StoreOptions
-from ..streams import Stream
+from ..streams import Stream, streams_list_from_dict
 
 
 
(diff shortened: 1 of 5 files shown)


jlstevens:
Since the last commit, linked streams also work:

image

jbednar:
Cool! And presumably `streams=dict(x=xy.param.x, y=xy.param.y)` also works?

jlstevens:
> And presumably streams=dict(x=xy.param.x, y=xy.param.y) also works?

Oops, yes! Fixed the screenshot above.

jlstevens:
Actually, there is a point to the original screenshot I posted:

</img>

This did not use to work with linked streams, even in the list format. You had to put `xy` into the streams list, not the parameters as it wouldn't set the callbacks to work with Bokeh (they were treated as normal parameters).

jlstevens:
And here is a simple screenshot showing how this will make working together with panel widgets easier:

</img>

jbednar:
Just for completeness of comparison, here's the way it would already be done now using `pn.bind`:

```python
def callback(x=1):
    return hv.Curve([1,2,3,x])

xy = hv.streams.PointerXY(x=1,y=2)
slider=pn.widgets.FloatSlider()
pn.Row(slider, pn.bind(callback,x=slider))
```

(Exactly the same apart from the last line.) Personally, I prefer the pn.bind version syntactically and semantically, but I'm very happy for the DynamicMap version to work as well, and there may be performance differences.

jlstevens:
> Personally, I prefer the pn.bind version... 

Well `pn.bind(callback,x=slider)` isn't a `DynamicMap` so it can't be put into an overlay, can't be fed into a HoloViews operation, won't have automatic normalization applied with other visual elements, is generally less compositional (can be put into a panel layout but not a HoloViews layout) etc. I'm not sure how you would hook up `PointerXY` to it either (you can pass the parameter but will the bokeh callbacks be registered?). For this simple example, `pn.bind` is pretty equivalent but I think it won't have the same number of affordances as a `DynamicMap` for more complex tasks.

Anyway, one thing that isn't working with this PR right now are class parameters (has to be an instance). I've added that to the TODO list to see if class parameters can be supported.

holoviews/streams.py
```diff
@@ -52,7 +52,9 @@ def streams_list_from_dict(streams):
         if 'panel' in sys.modules:
             from panel.depends import param_value_if_widget
             v = param_value_if_widget(v)
-        if isinstance(v, param.Parameter) and isinstance(v.owner, param.Parameterized):
+        if (isinstance(v, param.Parameter)
+            and (isinstance(v.owner, param.Parameterized)
+                 or issubclass(v.owner, param.Parameterized))):
```
philippjfr, line 57: Probably can simplify this too:

```suggestion
        if isinstance(v, param.Parameter) and v.owner is not None:
```

jlstevens: Sure!

jlstevens:
@philippjfr Ready for review/merge once the tests are green.

holoviews/tests/teststreams.py
```diff
@@ -1230,6 +1270,10 @@ def test_selection_expr_stream_hist_invert_xaxis_yaxis(self):
 
     def test_selection_expr_stream_polygon_index_cols(self):
         # Create SelectionExpr on element
+        try:
+            import shapely
+        except:
+            raise SkipTest('Shapely required for polygon selection')
```
philippjfr, line 1276: Thanks, though this requires **either** shapely or spatialpandas.

holoviews/tests/core/testdynamic.py
```diff
@@ -1,8 +1,10 @@
 import uuid
 import time
+import sys
 from collections import deque
 
 import param
+from unittest import SkipTest
```
philippjfr, line 7: Don't want to be too much of a stickler but imports are generally organized by stdlib then external libraries.

philippjfr:
Very minor comments, otherwise looks fine.

holoviews/tests/teststreams.py
jlstevens: Addressed in f9370a69c

holoviews/tests/core/testdynamic.py
jlstevens: Fixed in f9370a69c

jbednar:
> Well pn.bind(callback,x=slider) isn't a DynamicMap so it (1) can't be put into an overlay, (2) can't be fed into a HoloViews operation, (3) won't have automatic normalization applied with other visual elements, (4) is generally less compositional (can be put into a panel layout but not a HoloViews layout) etc. (5)I'm not sure how you would hook up PointerXY to it either (you can pass the parameter in but will the bokeh callbacks be registered to make it responsive to cursor position?).

I think _some_ of those things are true, and that's one reason I do support adding this functionality; often you do want a DynamicMap explicitly. But to take them point by point:

1. Sure, but you can have a callback that returns an overlay. 
2. Sure, but people writing a callback at the Panel level are probably more comfortable just writing more callbacks than they are with writing (and probably with using) a HoloViews operation explicitly.
3. I don't think using pn.bind affects normalization; see example below.
4. I personally favor blurring the distinction between Panel layouts and HoloViews layouts, making them have similar affordances where possible, at least partly for this reason.
5. I think that's true. I don't know how to make PointerXY be owned by a Curve that's inside a pn.bind callback.

```
import panel as pn, holoviews as hv, param
hv.extension("bokeh")
def callback(x=1, scale=1):
    return hv.Curve([1*scale,2*scale,3*scale,x])

slider1=pn.widgets.FloatSlider()
slider2=pn.widgets.FloatSlider()
pn.Column(pn.Row(slider1, slider2), 
          pn.Row(pn.bind(callback,x=slider1), 
                 pn.bind(callback,x=slider2,scale=2)))
```
image
(notice the shared scales; the axes are also linked if you drag the plots. Panel magic!)

jlstevens:
> I don't think using pn.bind affects normalization; see example below.

I didn't know that! Looks like panel is using HoloViews machinery behind the scenes to achieve this. Though how would you now specify axiswise versus framewise (or switch it off if desired)?

As for your other points, I agree that it is possible to work around (most of) the limitations using a certain coding style but I personally find it less compositional and less flexible. For instance, you can pull apart `dmap1 * dmap2 * dmap3` easily but if the overlay is built inside the callback, it is harder to pull the components apart without rewriting the callback (or making it a fair bit more complicated).

philippjfr:
>Though how would you now specify axiswise versus framewise (or switch it off if desired)?

The same way as in HoloViews, by setting axiswise=True/False, or you can explicitly opt out of axis sharing with `pn.pane.HoloViews(..., linked_axes=False)`.

>I think that's true. I don't know how to make PointerXY be owned by a Curve that's inside a pn.bind callback.

Don't do this, but you could simply assign `stream.source = curve`. That said if you did you'd be triggering a full rerender on each mouse move. Therefore you should just wrap your pn.bind callback in a `DynamicMap`.

holoviews/core/spaces.py
```diff
@@ -881,11 +881,12 @@ class DynamicMap(HoloMap):
         If the callable is an instance of Callable it will be used
         directly, otherwise it will be automatically wrapped in one.""")
 
-    streams = param.List(default=[], constant=True, doc="""
-       List of Stream instances to associate with the DynamicMap. The
-       set of parameter values across these streams will be supplied as
-       keyword arguments to the callback when the events are received,
-       updating the streams.""" )
+    streams = param.ClassSelector(default=[], class_=(dict, list), constant=True, doc="""
```
philippjfr, line 884:
```suggestion
    streams = param.List(default=[], constant=True, doc="""
```

holoviews/core/spaces.py
```diff
@@ -881,11 +881,12 @@ class DynamicMap(HoloMap):
...
+       List of Stream instances to associate with the DynamicMap or a
+       dictionary mapping parameters or panel widgets to callback
+       argument names. In the list case, the set of parameter values
+       across these streams will be supplied as keyword arguments to the
+       callback when the events are received, updating the streams.""" )
```
philippjfr, line 889: This should probably explain that dicts are accepted but are then converted into the list format.

holoviews/core/spaces.py
jlstevens: @jbednar I would prefer to give more explicit indication that `streams=` can be supplied as a dict or a list using `ClassSelector` but Philipp is also right that after the constructor it will always be a list. Happy to let you be the tie breaker!

holoviews/core/spaces.py
jlstevens: Addressed in e4dc46149

holoviews/core/spaces.py
jlstevens: I'll also note that `constant=True` has always been declared so you were never meant to set `streams` after the constructor has run.

jbednar: Good question. I guess `param.List` correctly reports what the `streams` parameter holds, so I agree with this suggestion. It's just important to document in the constructor's docstring that dict is also accepted.

PR #4792, opened 2021-01-16
Support spatialpandas DaskGeoDataFrame

philippjfr:
Adds support for DaskGeoDataFrame to the SpatialpandasInterface.

- [x] Implements https://github.com/holoviz/holoviews/issues/4234
- [x] Add tests
- [x] Ensure all methods are handled

PR #4794, opened 2021-01-19
Added inspect_points datashader operation

jlstevens:
WIP.

Here is the example code

```python
import numpy as np
from spatialpandas.geometry import PointArray
from spatialpandas import GeoDataFrame
from holoviews.operation.datashader import rasterize, dynspread, inspect_sample
import holoviews as hv
hv.extension('bokeh')

def gaussian_df(specs=(1.5,0,1.0),num=10000):
    np.random.seed(1)
    xs, ys = np.random.normal(specs[0],specs[2],num), np.random.normal(specs[1],specs[2],num)
    return xs,ys

xs, ys = gaussian_df()
sdf = GeoDataFrame({'geometry':PointArray((xs, ys)),'id':list(range(len(xs)))})
rasterized = dynspread(rasterize(hv.Points(sdf)))
rasterized * inspect_sample(rasterized).opts(color='red', tools=['hover'], size=6)
```
github1

Getting tap behavior is as easy as adding `streams=[hv.streams.Tap]` to the operation:

github1

## TODO

- [x] Make sure vdims of original data are preserved for hover information
- [x] Find out why the initial render looks blank until a small zoom in/out occurs
- [x] Handle normal dataframes (getting the x/y from the raster kdims), geopandas, and spatially indexed spatialpandas dataframes.
- [x] Tests
- [ ] Documentation

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1885,7 +1885,8 @@ class inspect_points(Operation):
 
     @classmethod
     def _mask_dataframe(cls, raster, x, y, xdelta, ydelta):
-        "Mask the dataframe with size around x and y"
+        "Mask the dataframe around the specified x and y position with
+        the given x and y deltas"
         ds = raster.dataset
         x0, x1, y0, y1 = x-xdelta, x+xdelta, y-ydelta, y+ydelta
         if 'spatialpandas' in ds.interface.datatype:
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -861,7 +861,7 @@ class regrid(AggregationOperation):
         for i, vd in enumerate(element.vdims):
             if element.interface is XArrayInterface:
                 if element.interface.packed(element):
-                    xarr = element.data[..., i]
+                    xarr = element.data[..., i] 
                 else:
                     xarr = element.data[vd.name]
                 if 'datetime' in (xtype, ytype):
@@ -1743,7 +1743,8 @@ class inspect_points(Operation):
     null_value = param.Number(default=0, doc="""
        Value of raster which indicates no hits. For instance zero for
        count aggregator (default) and commonly NaN for other (float)
-       aggregators""")
+       aggregators. For RGBA images, the alpha channel is used which means
+       zero alpha acts as the null value.""")
 
     value_bounds = param.NumericTuple(default=None, length=2, allow_None=True, doc="""
        If not None, a numeric bounds for the pixel under the cursor in
@@ -1769,14 +1770,26 @@ class inspect_points(Operation):
        columns and dtypes when no hits occur. The transformed dataframe
        is also what is passed to the Points object.""")
 
+    points_transformer = param.Callable(default=identity, doc="""
+      Function that transforms the hits dataframe""")
+
+    vdims = param.List(default=[], doc="""
+       The vdims of the returned Point element drawn from the source
+       dataframe. If a points_transformer is used, these columns need to
+       in the return value.""")
+
     # Stream values and overrides
     streams = param.List(default=[PointerXY])
     x = param.Number(default=0)
     y = param.Number(default=0)
 
     def _process(self, raster, key=None):
-        no_match = Points(self.p.hits_transformer(None))
-        try: val = raster[self.p.x,self.p.y]
+        no_match = Points(self.p.hits_transformer(None), vdims=self.p.vdims)
+        try:
+            if isinstance(raster, RGB):
+                val = raster[self.p.x,self.p.y, 'A']
+            else:
+                val = raster[self.p.x,self.p.y]
         except: val = self.p.null_value # Exception at the edges
         interface = raster.dataset.interface.datatype
         if  interface != 'spatialpandas':
@@ -1787,24 +1800,22 @@ class inspect_points(Operation):
             if (val < self.p.value_bounds[0]) or (val > self.p.value_bounds[1]):
                 self.p.hits = self.p.hits_transformer(None)
                 return no_match
+
         if val != self.p.null_value:
-            mask_size = self._distance_args(x_range, y_range, raster.data, self.pixels)
+            mask_size = self._distance_args(raster, x_range, y_range, self.pixels)
             masked = self._mask_dataframe(raster.dataset.data, self.p.x, self.p.y,
                                           mask_size, interface)
             dist_sorted = self._sort_by_distance(masked, self.p.x, self.p.y,
                                                  raster.kdims, interface)
             self.hits = self.p.hits_transformer(dist_sorted)
-            vdims=  ([] if (self.p.hits_transformer is identity)
-                     else [col for col in self.hits.columns
-                           if col not in ['x','y']])
-            point = Points(self.hits, vdims=vdims).iloc[:self.p.point_count]
+            point = Points(self.p.points_transformer(self.hits), vdims=self.p.vdims).iloc[:self.p.point_count]
             return point
         self.hits = self.p.hits_transformer(None)
         return no_match
 
     @classmethod
-    def _distance_args(cls, x_range, y_range, data, pixels):
-        xcount, ycount = data.dims['x'], data.dims['y']
+    def _distance_args(cls, element, x_range, y_range,  pixels):
+        ycount, xcount =  element.interface.shape(element, gridded=True)
         x_delta = abs(x_range[1] - x_range[0]) / xcount
         y_delta = abs(y_range[1] - y_range[0]) / ycount
         return (x_delta*pixels, y_delta*pixels)
@@ -1814,7 +1825,10 @@ class inspect_points(Operation):
         "Mask the dataframe with size around x and y"
         xdelta, ydelta = mask_size
         # If spatialpandas, pandas, geopandas etc. Currently assuming spatialpandas
-        return df.cx[x-xdelta:x+xdelta, y-ydelta:y+ydelta]
+        masked = df.cx[x-xdelta:x+xdelta, y-ydelta:y+ydelta]
+        if hasattr(masked, 'compute'):
+            masked = masked.compute()
+        return masked
 
     @classmethod
     def _get_xs_ys(cls, df, raster_kdims, interface):
@@ -1830,4 +1844,6 @@ class inspect_points(Operation):
         xs, ys = cls._get_xs_ys(df, raster_kdims, interface)
         dx, dy = xs - x, ys - y
         distances = pd.Series(dx*dx + dy*dy)
-        return df.iloc[distances.argsort()]
+        if hasattr(df.geometry, 'compute'):
+            df = df.compute()
+        return df.iloc[list(distances.argsort())]
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1794,8 +1794,8 @@ class inspect_base(Operation):
 
     hits = param.DataFrame(default=pd.DataFrame(), allow_None=True)
 
-    max_hits = param.Integer(default=1, doc="""
-       Maximum number of points to display within the mask of size
+    max_elements = param.Integer(default=1, doc="""
+       Maximum number of elements to display within the mask of size
        pixels. Points are prioritized by distance from the cursor
        point. This means that the default value of one shows the single
        closest sample to the cursor. Note that this limit is not applies
@@ -1836,7 +1836,7 @@ class inspect_base(Operation):
 
         self.hits = result
         df = self.p.transform(result)
-        return self._element(raster, df.iloc[:self.p.max_hits])
+        return self._element(raster, df.iloc[:self.p.max_elements])
 
     @classmethod
     def _distance_args(cls, element, x_range, y_range,  pixels):
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1790,8 +1790,8 @@ class inspect_points(Operation):
         if ((self.p.value_bounds and not (self.p.value_bounds[0] < val < self.p.value_bounds[1])) or
             val == self.p.null_value):
             empty_df = self._empty_df(raster.dataset)
+            df = self.p.points_transformer(empty_df)
             self.hits = self.p.hits_transformer(empty_df)
-            df = self.p.points_transformer(self.hits)
             vdims = self._vdims(raster, df)
             return Points(df, kdims=raster.kdims, vdims=vdims)
 
@@ -1800,8 +1800,8 @@ class inspect_points(Operation):
         mask_size = self._distance_args(raster, x_range, y_range, self.pixels)
         masked = self._mask_dataframe(raster, self.p.x, self.p.y, mask_size)
         dist_sorted = self._sort_by_distance(raster, masked, self.p.x, self.p.y)
+        df = self.p.points_transformer(dist_sorted)
         self.hits = self.p.hits_transformer(dist_sorted)
-        df = self.p.points_transformer(self.hits)
         vdims = self._vdims(raster, df)
         return Points(df, kdims=raster.kdims, vdims=vdims).iloc[:self.p.point_count]
 
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1773,45 +1773,41 @@ class inspect_points(Operation):
     points_transformer = param.Callable(default=identity, doc="""
       Function that transforms the hits dataframe""")
 
-    vdims = param.List(default=[], doc="""
-       The vdims of the returned Point element drawn from the source
-       dataframe. If a points_transformer is used, these columns need to
-       in the return value.""")
-
     # Stream values and overrides
     streams = param.List(default=[PointerXY])
     x = param.Number(default=0)
     y = param.Number(default=0)
 
     def _process(self, raster, key=None):
-        no_match = Points(self.p.hits_transformer(None), vdims=self.p.vdims)
         try:
             if isinstance(raster, RGB):
                 val = raster[self.p.x,self.p.y, 'A']
             else:
                 val = raster[self.p.x,self.p.y]
-        except: val = self.p.null_value # Exception at the edges
-        interface = raster.dataset.interface.datatype
-        if  interface != 'spatialpandas':
-            raise Exception('%r interface not yet supported' % interface)
+        except Exception:
+            val = self.p.null_value # Exception at the edges
+
+        if ((self.p.value_bounds and not (self.p.value_bounds[0] < val < self.p.value_bounds[1])) or
+            val == self.p.null_value):
+            df = self._empty_df(raster.dataset)
+            self.hits = self.p.hits_transformer(None)
+            return Points(self.points_transformer(self.hits), kdims=raster.kdims)
+
         x_range, y_range = raster.range(0), raster.range(1)
 
-        if self.p.value_bounds is not None:
-            if (val < self.p.value_bounds[0]) or (val > self.p.value_bounds[1]):
-                self.p.hits = self.p.hits_transformer(None)
-                return no_match
+        mask_size = self._distance_args(raster, x_range, y_range, self.pixels)
+        masked = self._mask_dataframe(raster, self.p.x, self.p.y, mask_size)
+        dist_sorted = self._sort_by_distance(raster, masked, self.p.x, self.p.y)
+        self.hits = self.p.hits_transformer(dist_sorted)
+        return Points(self.p.points_transformer(self.hits), vdims=self.p.vdims).iloc[:self.p.point_count]
 
-        if val != self.p.null_value:
-            mask_size = self._distance_args(raster, x_range, y_range, self.pixels)
-            masked = self._mask_dataframe(raster.dataset.data, self.p.x, self.p.y,
-                                          mask_size, interface)
-            dist_sorted = self._sort_by_distance(masked, self.p.x, self.p.y,
-                                                 raster.kdims, interface)
-            self.hits = self.p.hits_transformer(dist_sorted)
-            point = Points(self.p.points_transformer(self.hits), vdims=self.p.vdims).iloc[:self.p.point_count]
-            return point
-        self.hits = self.p.hits_transformer(None)
-        return no_match
+    @classmethod
+    def _empty_df(cls, dataset):
+        if 'dask' in dataset.interface.datatype:
+            return dataset.data.partitions[0].head(0)
+        elif dataset.interface.datatype in ['pandas', 'geopandas', 'spatialpandas']:
+            return dataset.data.head(0)
+        return dataset.iloc[:0].dframe()
 
     @classmethod
     def _distance_args(cls, element, x_range, y_range,  pixels):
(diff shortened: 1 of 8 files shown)


jlstevens:
I might prefer `inspect_point` in the end as `Points` is the return type and it is only suited to datashaded points right now. You can imagine fancier things in future e.g `inspect_segments`, `inspect_paths`, `inspect_spikes`, `inspect_polygons` etc etc.

holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,78 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+    Given datashaded aggregate (Image) output, return a single
+    (hoverable) point at the sample closest to the cursor.
+    """
+
+    pixels = param.Integer(default=3, doc="""
+       Number of pixels in data space around the cursor point to search
+       for hits in (with mask_shape). The hit in this mask that is
+       closest to the cursors position is displayed.""")
+
+    mask_shape = param.ObjectSelector(default='square', objects=['square'], doc="""
```
jbednar, line 1741: Maybe remove this, hard code it to a square, leave changing the shape for future work, then never do it? :-)

jbednar (commented):
Looks great so far!

holoviews/operation/datashader.py
jlstevens: Sure, in which case I would file an issue to consider other mask shapes once this is merged (as a feature request).

jbednar:
Can you maybe make the default shape an `o` (open circle) so that the marker doesn't cover up the datashaded data point?

jlstevens:
Operations shouldn't set those kinds of options. It is easy enough to add `fill_alpha=0`. e.g:

</img>

Would be happy to document this as an example of course.

holoviews/operation/datashader.py
jlstevens: Addressed in 022680df0. Will now be a feature request issue (if it seems worthwhile).

jlstevens:
Here is an example of how I expect the drilldown table to work:

drilldown

jlstevens:
Added `point_count` parameter so this operation can live up to the name `inspect_points` (i.e plural). Here is an example with (a maximum of) ten points shown within an expanded mask size of 50:

point_count

jlstevens:
I think this PR will probably need some of the fixes planned in https://github.com/holoviz/holoviews/pull/4792

jlstevens:
One comment Philipp made in #4796 (now closed) is that this operation should handle RGBs as well as RGBAs. I suggest raising an exception unless we want to declare a specific color as a null value.

holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,125 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+       dataframe. If a points_transformer is used, these columns need to
+       in the return value.""")
+
+    # Stream values and overrides
+    streams = param.List(default=[PointerXY])
+    x = param.Number(default=0)
+    y = param.Number(default=0)
+
+    def _process(self, raster, key=None):
+        no_match = Points(self.p.hits_transformer(None), vdims=self.p.vdims)
```
jlstevens, line 1787: Should indeed be `self.p.points_transformer`.

jbednar:
> I suggest raising an exception unless we want to declare a specific color as a null value.

Is it _essential_ to detect a null value, or just helpful? If it's not essential, I'd just print a helpful warning and then just skip checking for nulls, unless that makes it significantly more complex.

jlstevens:
> Is it essential to detect a null value, or just helpful?

It is not essential but if you can't detect null you can't optimize when inspecting sparse data. A warning seems fine.

jlstevens:
To make sure this isn't forgotten, I realized one bug we have: the null test should be over the mask area and not just a point test (i.e use the `pixels` parameter). 

We can also have an optimization where we pick the first sample we find falling within a single pixel (assuming anything after that doesn't matter).

jlstevens:
Other things I would like to attempt before merging:

- Support normal dataframes without a `.cx` method using normal slicing, issuing a warning recommending spatialpandas/geopandas. Maybe this warning could be disabled by a parameter if people *really* want to avoid those dependencies.
- Output a Polygon in the shape of the mask instead of the point. Seeing the actual search area seems really useful and would make things like drilldown histograms more useful.

Other than that (and any unaddressed items above), I am very happy with this new operation! At a later date we can consider `inspect_segments`, `inspect_paths`, `inspect_polygons` etc..

jlstevens:
Added an `inspect_mask` operation to show the mask that can be used as follows:

</img>

I think it could be improved further: ideally `inspector.mask(rasterized)` could be simply `inspector.mask` but then the processed element would need to be linked between the two operations.

jlstevens:
@jbednar @philippjfr Ready for review. Hoping for some feedback on the API (e.g parameter names, the mask operation, arrangement of the transformers. Once things settle, I'll add unit tests. Happy to add docs in this PR but can also open a new PR if we want the code merged sooner.

holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,193 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+        rect = np.array([(x-xdelta/2,y-ydelta/2), (x+xdelta/2, y-ydelta/2),
+                         (x+xdelta/2, y+ydelta/2), (x-xdelta/2, y+ydelta/2)])
+        data = {(str(kdims[0]),str(kdims[1])):rect}
+        return Polygons(data, kdims=kdims)
+
+
+class inspect_points(Operation):
+    """
+    Given datashaded aggregate (Image) output, return a single
+    (hoverable) point at the sample closest to the cursor.
```
jbednar, line 1774:
```suggestion
    Given datashaded aggregate (Image) output, return a set of
    (hoverable) points sampled from those near the cursor.
```

holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,193 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+    """
+
+    pixels = param.Integer(default=3, doc="""
+       Number of pixels in data space around the cursor point to search
+       for hits in. The hit within this box mask that is closest to the
+       cursors position is displayed.""")
```
jbednar, line 1780:
```suggestion
       Number of pixels in data space around the cursor point to search
       for hits in. The hit within this box mask that is closest to the
       cursor's position is displayed.""")
```

Is that true? I'd think we don't care about the actual distances, given the quantization that is forced by the pixel array, and instead preferentially return a datapoint from the actual pixel (any datapoint falling in that pixe, as all such points have the same logical distance given the limited resolution of a click), and barring that, any point falling within the square ring of pixels surrounding the central pixel, etc. up to the limit.

holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,193 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+
+    value_bounds = param.NumericTuple(default=None, length=2, allow_None=True, doc="""
+       If not None, a numeric bounds for the pixel under the cursor in
+       order for hits to be computed. Useful for count aggregators where
+       a value of (1,1000) would make sure no more than a thousand
+       samples will be searched.""")
+
+    hits = param.DataFrame(default=pd.DataFrame(), allow_None=True)
+
+    point_count = param.Integer(default=1, doc="""
```
jbednar, line 1796: max_points?

holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,193 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+       Maximum number of points to display within the mask of size
+       pixels. Points are prioritized by distance from the cursor
+       point. This means that the default value of one shows the single
+       closest sample to the cursor.""")
```
jbednar, line 1800: Again, I'm not sure that's well defined; seems like all datapoints falling into the cursor's pixel are equivalent, and they don't need to be the closest (e.g. to the cursor pixel's center) to be appropriate; I'd just take any of them as there's no particular order that can be established on them.

holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,193 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+
+    transformer = param.Callable(default=identity, doc="""
+       Function that can transform the hits dataframe before it is
+       passed to the hits_transformer and points_transformer for further
+       customization of the hits parameter and points output
+       respectively. Useful when no distinction is needed between the
+       hits_transformer and the points_transformer or when these later
+       transformers share the same initial code (e.g common merges and
```
jbednar, line 1808:
```suggestion
       transformers share the same initial code (e.g. common merges and
```

holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,193 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+       hits occur, the value passed to the callable is None. This can be
+       used to do multiple common tasks such as 1) subselecting
+       available columns 2) limiting the maximum number of hit rows
+       returned 3) performing joins with other data sources 4) returning
+       a dataframe with appropriate columns and dtypes when no hits
+       occur.""")
+
+    points_transformer = param.Callable(default=identity, doc="""
+      Function that transforms the hits dataframe after applying the
+      transformer and before it it pass to the Points. Can be used to
```
jbednar, line 1823:
```suggestion
      transformer and before it is passed to the Points. Can be used to
```

holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,193 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+      customize the value dimensions e.g to implement custom hover
```
jbednar, line 1824:
```suggestion
      customize the value dimensions e.g. to implement custom hover
```

holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,193 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+    @classmethod
+    def _distance_args(cls, element, x_range, y_range,  pixels):
+        ycount, xcount =  element.interface.shape(element, gridded=True)
+        x_delta = abs(x_range[1] - x_range[0]) / xcount
+        y_delta = abs(y_range[1] - y_range[0]) / ycount
+        return (x_delta*pixels, y_delta*pixels)
+
+    @classmethod
+    def _mask_dataframe(cls, raster, x, y, xdelta, ydelta):
+        "Mask the dataframe with size around x and y"
```
jbednar, line 1888: What is `size`?

holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,193 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+
+    @classmethod
+    def _get_xs_ys(cls, raster, df):
+        "Returns the x and y arrays or series for the given dataframe"
+        ds = raster.dataset
+        if 'spatialpandas' in ds.interface.datatype:
+            geom = ds.interface.geo_column(ds.data)
+            return df[geom].array.x, df[geom].array.y
+        xdim, ydim = raster.kdims
+        return df[xdim].values, df[ydim].values
```
jbednar, line 1906: It might be good to make this a module-level function; I've needed precisely this functionality to handle spatialpandas and regular pandas dataframes consistently.

holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,193 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+    @classmethod
+    def _sort_by_distance(cls, raster, df, x, y):
+        """
+        Returns a dataframe of hits within a given mask around a given
+        spatial location, sorted by distance from that location.
+        """
+        xs, ys = cls._get_xs_ys(raster, df)
+        dx, dy = xs - x, ys - y
+        distances = pd.Series(dx*dx + dy*dy)
+        return df.iloc[distances.argsort().values]
```
jbednar, line 1917: I don't know how expensive it is to calculate distances like this, but if it's quicker to run the query multiple times (1x1, 3x3, etc. up to max_points) and abort early, we should! In the worst case (clicking in a dense region), which is presumably where we have to worry the most about performance, I'd expect the 1x1 region nearly always to have at least one datapoint, so the other cases would only be reached in sparser regions where I'd think it's cheaper to do the query anyway.

jbednar (commented):
This will be very useful. I find the names `transformer`, `hits_transformer`, and `points_transformer` confusing (and verbose), though. E.g. doesn't `transformer` also transform hits, just as much as `hits_transformer` does? Maybe`match_transform`, `matches_to_points`, `matches_to_hits`?

It would be nice if this operation can be passed Points as an argument instead of being specialized for points, but I can't see how that could work.

Will it be able to e.g. select an entire connected trajectory, given one point on that trajectory?

holoviews/operation/datashader.py
jlstevens: Currently it is actually true. What you are suggesting is a way to terminate early (but you would also need to know when you've finally got a sample within the pixel, assuming that you want that granularity). That suggestion is a good one, might speed things up (though I'm not entirely sure as a vectorized operation may be faster than trying to early termination) but it is not implemented right now.

holoviews/operation/datashader.py
jlstevens: Good suggestion.

holoviews/operation/datashader.py
philippjfr: Shouldn't we just filter by max points before computing distances? We've already established that they are in the target region.

holoviews/operation/datashader.py
jlstevens: The size parameter (in pixels). I'll clarify accordingly.

holoviews/operation/datashader.py
philippjfr: Ah sorry then you can't establish which is closest. Ignore me.

jlstevens: > I don't know how expensive it is to calculate distances like this...

It will be O(N) in the number of hits found. If you have a million hits I imagine it will take a while to compute but that seems unlikely.

> ... but if it's quicker to run the query multiple times (1x1, 3x3, etc. up to max_points) and abort early, we should

Honestly, I think this is tricky to reason about because the performance is going to depend a lot on the distribution of the data being inspected. 

> 1x1, 3x3, etc. up to max_points

I'm not sure I follow. Those look like mask sizes to me so I don't see how that relates to `max_points`. Unless you meant up to the full mask size? If so, it sounds like you mean to query a single pixel mask, and if there are insufficient hits, try 2x2 etc. 

Such an algorithm would need multiple `.cx` queries which is likely to me *much* more expensive than computing the distances unless there are a huge number of matches in the mask. The possibly optimization I had in mind was different: do a chunk of distances and if you have `max_points` rows that are found within the center pixel, return them (basically this says any samples within the center pixel are considered equivalently good no matter where the cursor is in that pixel). Note that if you consider a raster to be an approximation to an underlying density, finding the minimum distance (as it does now) is the most correct thing to do...

I agree there may be cases where the distance computation becomes the limiting factor but my intuition is that the `.cx` filter will be the far bigger limiting factor in any realistic dataset (yes, you could try to inspect a dataframe with 100 million points with exactly the same position but this doesn't seem very realistic).

holoviews/operation/datashader.py
jlstevens: Addressed in ebb04cffa

holoviews/operation/datashader.py
jlstevens: >  I'd just take any of them as there's no particular order that can be established on them.

That is a possible optimization and not true in all cases. 

If the datashading is occurring at screen resolution and bokeh's hover/tap is at the same resolution then it is true (well, the above is perfectly accurate but limited by the precision supplied by Bokeh), but it is very easy to ask `rasterize` or `datashade` to output rasters where that is not true at all (i.e bins much bigger than screen pixels by specifying custom `x_sampling` and `y_sampling`).

As currently defined, this inspector will work correctly for *all* datashaded output and this docstring is correct. For a lot of common datashader output where things are configured to automatically make datashaded bins match screen pixels, then your suggested change would work because Bokeh isn't giving anything past screen pixel precision (and even this might not be the case, for all I know a mouse pointer can specify an x,y position that is finer than a pixel! What happens if you stick your screen into 640x480 resolution?). 

Because of all this, I think the docstring is well defined and correct even if the precision with which the cursor's position is specified is limited.

holoviews/operation/datashader.py
jlstevens: Addressed in efa5fff18

holoviews/operation/datashader.py
jlstevens: I'm happy to pull out utilities but I can't quite see it here. What would be the name of the utility and what would be its signature? A signature that involves an element and an arbitrary dataframe as input seems hard to explain in a docstring, what is their relation (note that the raster element here doesn't contain a DataFrame, it is its source dataset that does)?

jlstevens:
> Will it be able to e.g. select an entire connected trajectory, given one point on that trajectory?

No. This operation is explicitly `inspect_points` and I've mentioned earlier that there could be `inspect_paths`, `inspect_segments` etc. What you suggest would be possible when we flesh out this suite of operations.

I've addressed quite a few of your suggestions, thanks! The remaining items need a bit more discussion though.

> Maybe `match_transform`, `matches_to_points`, `matches_to_hits`?

Happy to use these though I don't find them particularly compelling either. Maybe `masked_transform`, `masked_to_points` and `masked_to_hits`? I don't see enough of a difference between the semantics of 'matches' and 'hits' otherwise. @philippjfr any other suggestions? What do you prefer?

philippjfr:
Could then provide a generic `inspect` operation which wraps around element specific ones much like rasterize does.

jlstevens:
> Could then provide a generic inspect operation which wraps around element specific ones much like rasterize does.

I like this but how would `inspect` know what kind of thing was datashaded/rasterized? Or in other words, what kind of element was turned into a raster? I think this is possible by inspecting `.pipeline` but not via `.dataset`.

holoviews/operation/datashader.py
philippjfr: Perhaps this should just be:

```python
ds = raster.dataset.clone(df)
return (ds.dimension_values(kd) for kd in raster.kdims)
```

holoviews/operation/datashader.py
philippjfr: Agree that this sounds significantly **more** expensive rather than less, particularly in the worst case. I don't see **any** scenario where repeated spatial queries will be faster than computing distances.

setup.py
```diff
@@ -42,7 +42,7 @@
     "ffmpeg",
     "cftime",
     "netcdf4",
-    "dask",
+    "dask<=2020.12.0",
```
philippjfr, line 45: Ideally we'd fix the regression in a separate PR, rather than pin this here.

holoviews/operation/datashader.py
jlstevens: That is a nice simplification and I like it. My question then is whether the kdims (default 'x' and 'y') can be set to something else when there is just a 'geometry' column. For instance, does redim work? In general having kdims that aren't actual columns in the dataframe is kinda weird (though I guess there is no alternatives for these geodataframes?).

setup.py
jlstevens: Sure, I'll make a quick PR, revert this and can rebase once merged.

holoviews/operation/datashader.py
jbednar: I must be missing something here. The worst case is when you click on a dense region of the plot, right? I.e. somewhere there are millions of points per pixel?  Is so how could looking in a 1x1 pixel area first be bad?  With a very high probability, there will be a point in the 1x1 pixel area and it will then get returned, leaving the total search space lower than if you had done a smaller mask.  It's true that in some other cases one will have to do multiple queries, but aren't those the easy cases, where there are few points?

setup.py
jlstevens: Addressed in a293d626. New PR is https://github.com/holoviz/holoviews/pull/4803

jlstevens:
The dask pinning is now in #4803 and we can decide what to do there. Tests on this PR will be failing unless we reintroduce the pin, fix the underlying issue (e.g update the tests?) or rebase after that PR is merged to master.

holoviews/operation/datashader.py
jlstevens: Applied this suggestion in 458508a and it seems to work fine. @philippjfr Just curious about my earlier questions though...I suspect we discussed this before but I can't quite remember (maybe it is a case where the situation is not ideal and not easy to address properly).

holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,188 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+        return ds.select(**query).dframe()
+
+    @classmethod
+    def _sort_by_distance(cls, raster, df, x, y):
+        """
+        Returns a dataframe of hits within a given mask around a given
+        spatial location, sorted by distance from that location.
+        """
+        ds = raster.dataset.clone(df)
+        print("BOO")
```
philippjfr, line 1908: Stray print.

holoviews/operation/datashader.py
philippjfr: >It's true that in some other cases one will have to do multiple queries, but aren't those the easy cases, where there are few points?

That's assuming a lot about the data distribution no? In any case though I think you aren't appreciating the difference in cost between the two operations, the 1x1 query will take pretty much the same amount of time as the 3x3 query. So yes, you'll save on the distance calculation but that step is (almost) never going to be expensive when compared to the spatial query, e.g. it takes 70 ms on 10 million points. So you're trading a small bit of speedup in the distance calculation in the best case against an `~N x` slowdown in the worst case, where N is the size of the `N pixel x N pixel` mask.

holoviews/operation/datashader.py
jlstevens: Addressed in 6745b6758

holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,187 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+            empty_df = self._empty_df(raster.dataset)
+            df = self.p.points_transformer(self.p.transformer(empty_df))
+            self.hits = self.p.hits_transformer(self.p.transformer(empty_df))
+            vdims = self._vdims(raster, df)
+            return Points(df, kdims=raster.kdims, vdims=vdims)
+
+        masked = self._mask_dataframe(raster, x, y, xdelta, ydelta)
+        dist_sorted = self._sort_by_distance(raster, masked, x, y)
+        df = self.p.points_transformer(self.p.transformer(dist_sorted))
+        self.hits = self.p.hits_transformer(self.p.transformer(dist_sorted))
```
philippjfr, line 1854:
```suggestion
        transformed = self.p.transformer(dist_sorted)
        df = self.p.points_transformer(transformed)
        self.hits = self.p.hits_transformer(transformed)
```

jlstevens:
Regarding naming, here are some generic (so that we have support `inspect_polygons`) semantic names we want to improve:

`common_dataframe_transform`: Suggestions: `transform_element`, `match_transform`
`output_element_data_transform`: Suggestions `transform`, `matches_to_points`,  `mask_to_points`
`output_dataframe_parameter_transform`: Suggestions `transform_hits`, `matches_to_hits`, `mask_to_hits`

holoviews/operation/datashader.py
jlstevens: More efficient, I agree.

holoviews/operation/datashader.py
jlstevens: > Is so how could looking in a 1x1 pixel area first be bad?

There is no ordering in which you look at pixels as it all works through a `cx` slice. Looking at one pixel takes just as long as a 100 (assuming you are in the same partition): you need to pass through all the rows and filter out the ones outside the chosen region.

holoviews/operation/datashader.py
```diff
@@ -1793,36 +1793,17 @@ class inspect_points(Operation):
...
-       a dataframe with appropriate columns and dtypes when no hits
-       occur.""")
-
-    points_transformer = param.Callable(default=identity, doc="""
-      Function that transforms the hits dataframe after applying the
-      transformer and before it is passed to the Points. Can be used to
-      customize the value dimensions e.g. to implement custom hover
-      behavior.""")
+       closest sample to the cursor. Note that this limit is not applies
+       to the hits parameter.""")
```
jbednar, line 1801: I suggested calling it `max_hits` when I thought it applied to the `hits` parameter; now that it's clarified I don't think `max_hits` is the right name.

jlstevens: Sure. How about `max_elements` then?

jbednar: Sure.

jlstevens: Addressed in 8811c71b9

philippjfr: Seems confusing and overloaded with HoloViews elements. It's more like max_geoms.

jbednar: I thought it _was_ a limit on hv elements?  Geoms seems confusing to me. Maybe max_selected? max_matches?

jlstevens: I vote for `max_indicators`

jlstevens: Addressed in 420e5da97. @philippjfr told me he had no objection to `max_indicators`.

jlstevens:
I've rebased off master which now has the dask testing fixed. Before I can write tests, it looks like `dynamic=False` needs fixed:

image

jlstevens:
Ok, `dynamic=False` works if you apply it to both operations:

</img>

jlstevens:
@jbednar @philippjfr All tests are passing and I am happy with the API. There are a few things left to do but I think we can merge now and cut a dev release:

- [ ] Unit tests for the polygon inspector.
- [ ] Documentation

jlstevens:
Going ahead and merging.

PR #4796, opened 2021-01-20
Added inspect_points operation

jlstevens:
This PR requires some of the fixes in https://github.com/holoviz/holoviews/pull/4792 and is essentially a way for me to continue work on https://github.com/holoviz/holoviews/pull/4794 until the aforementioned PR is merged.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1770,13 +1770,21 @@ class inspect_points(Operation):
        columns and dtypes when no hits occur. The transformed dataframe
        is also what is passed to the Points object.""")
 
+    points_transformer = param.Callable(default=identity, doc="""
+      Function that transforms the hits dataframe""")
+
+    vdims = param.List(default=[], doc="""
+       The vdims of the returned Point element drawn from the source
+       dataframe. If a points_transformer is used, these columns need to
+       in the return value.""")
+
     # Stream values and overrides
     streams = param.List(default=[PointerXY])
     x = param.Number(default=0)
     y = param.Number(default=0)
 
     def _process(self, raster, key=None):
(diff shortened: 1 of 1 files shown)


holoviews/operation/datashader.py
```diff
@@ -1725,3 +1725,125 @@ class directly_connect_edges(_connect_edges, connect_edges):
...
+
+    # Stream values and overrides
+    streams = param.List(default=[PointerXY])
+    x = param.Number(default=0)
+    y = param.Number(default=0)
+
+    def _process(self, raster, key=None):
+        no_match = Points(self.p.hits_transformer(None), vdims=self.p.vdims)
+        try:
+            if isinstance(raster, RGB):
```
philippjfr, line 1789: Shouldn't happen for datashaded images but not all RGBs have an alpha channel.

jlstevens: Good point. I'll make sure to check the channel number. If alpha is missing, that would be an error imho as I don't know what else you could do (maybe say white or black are null values but then you wouldn't have a single scalar, unless you use a hex color or something).

jbednar: I guess that could be a warning, and then just treat it as if the count is non-zero and non-NaN.

jlstevens:
Closing now that #4794 is rebased off master with the `dask_spatialpandas` PR merged.

holoviews/operation/datashader.py
jlstevens: In other words, disable the sparseness optimization and always run the spatial queries?

jbednar: Right.

PR #4798, opened 2021-01-23
Add single value numpy array to HLine and VLine input

Hoxbro:
This PR adds the ability to give a numpy array with a single value in it to HLine and VLine. 

Furthermore I renamed hline to vline in `test_vline_dimension_values`.


jlstevens:
Looks good to me! I'll merge soon unless @philippjfr has any suggestions...

PR #4806, opened 2021-01-28
Add automatic categorical legend for datashaded plots

philippjfr:
Inspects the pipeline of an RGB element and discovers the shade and rasterize operations used to generate an RGB to then automatically add a legend.

Screen Shot 2021-01-28 at 1 04 12 PM

- [x] Clean up operation and factor out operations to traverse pipeline
- [x] Factor out code to render legend plot and use across all interfaces
- [ ] Update legend when needed

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -253,21 +253,26 @@ class AggregationOperation(ResamplingOperation):
         'var':   rd.var,
         'std':   rd.std,
         'min':   rd.min,
-        'max':   rd.max
+        'max':   rd.max,
+        'count_cat': rd.count_cat
     }
 
     @classmethod
     def _get_aggregator(cls, element, agg, add_field=True):
         if isinstance(agg, str):
-            if agg not in self._agg_methods:
-                agg_methods = sorted(self._agg_methods)
+            if agg not in cls._agg_methods:
+                agg_methods = sorted(cls._agg_methods)
                 raise ValueError("Aggregation method '%r' is not known; "
                                  "aggregator must be one of: %r" %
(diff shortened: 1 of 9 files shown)


jbednar:
Does this still have the property that the previous hack did, of disrupting the axis range of a plot if that plot didn't happen to include 0,0? If so that needs documenting and explaining how to work around it.

philippjfr:
>Does this still have the property that the previous hack did, of disrupting the axis range of a plot if that plot didn't happen to include 0,0? If so that needs documenting and explaining how to work around it.

No, `apply_ranges=False` should avoid this.

philippjfr:
Actually no, not even that, it won't cause that issue because the plot won't even take part is axis range calculation in this approach.

jbednar:
Woohoo!!!

philippjfr:
Okay, happier with this now. I've put the operation behind a try/except as well so we don't cause any regressions in case I missed something.

jbednar:
How can I get text rather than numeric labels for a plot like this?

image

Also, I'm getting errors simply from importing HoloViews from this branch, with param 1.10.1 and pandas 1.1.5 installed:

image

philippjfr:
>Also, I'm getting warnings simply from importing HoloViews from this branch, with param 1.10.1 and pandas 1.1.5 installed:

Already fixed in https://github.com/holoviz/holoviews/pull/4821

>How can I get text rather than numeric labels for a plot like this?

Hmm, I suppose that's a problem. It uses what's actually in the data (which is how legends everywhere work). We could introduce a `legend_labels` option to allow remapping labels I suppose.

PR #4812, opened 2021-02-01
Disable zoom on axis for geographic plots

philippjfr:


jbednar:
Thanks! Are there any docs needed to tell people how to re-enable such zooming (e.g. if they are using non-geographic tiles)?

philippjfr:
Hmm, I suppose.

philippjfr:
Honestly the amount of low-level bokeh properties we set but don't expose directly is very large, so I'm not worried about a different default here and don't think this one deserves a special mention when a million others don't get one.

PR #4817, opened 2021-02-01
Handle isfinite for NumPy and Pandas masked arrays

philippjfr:
Ensures isfinite works on masked arrays.


holoviews/core/util.py
```diff
@@ -99,6 +100,9 @@ def __cmp__(self, other):
         if pandas_version > '0.23.0':
             from pandas.core.dtypes.generic import ABCExtensionArray
             arraylike_types = arraylike_types + (ABCExtensionArray,)
+        if pandas_version > '1.0':
```
philippjfr, line 103: Need to double check this version.

philippjfr: Turns out to be correct.

Issue #4807, opened 2021-01-29
Support streams dictionary format in operations


jlstevens:
Since https://github.com/holoviz/holoviews/pull/4787 was merged, `DynamicMap` supports a dictionary format for streams. For consistency, I think operations should support the same format (which can help resolve parameter name clashes):

```python
class example(Operation):
   streams = {'x':PointerXY.param.x, 'y':PointerXY.param.y, 'tap_x':Tap.param.x, 'tap_y':Tap.param.y}
   x = ...
   y = ...
   tap_x = ...
   tap_y = ...
```

The main wrinkle is that these are stream class parameters, not instance parameters (as in the current list format). I imagine the declared mapping would be transferred to stream instance when they are created.

Assigning to 1.14.2 as this isn't a change of API (but an extension of it) and it would be good to achieve consistency with what `DynamicMap` supports as soon as possible.

philippjfr:
This needs to go in today or tomorrow if we want it in the 1.14.2 release.

jlstevens:
Currently working on it. I'll know this evening if it is a quick addition or something that won't make it into the release.

jlstevens:
Addressed in the PR above (now merged).

PR #4818, opened 2021-02-01
Supporting streams dictionaries in operations

jlstevens:
WIP PR to address https://github.com/holoviz/holoviews/issues/4807.

For testing, I've defined the `streams` parameter of `inspect_points` as:

```python
    streams = param.ClassSelector(default=dict(x=PointerXY.param.x,
                                               y=PointerXY.param.y),
                                  class_=(dict, list))
```

And here is the corresponding, working test example:

PR

However, there are some serious limitations right now. The minor one is that I have to set `hv.streams.PointerXY.x = 0` and 
`hv.streams.PointerXY.y = 0` due to an error processing dictionaries when allowing the `None` default. While I'm sure this can be fixed, this seems to be a far more fundamental problem right now:

</img>

As operations use stream classes, not instances, how can I get back to the correct *class* with something like `Tap.param.x` when the owner of the parameter may be a superclass? The code needs to instantiate the correct stream instance. Any ideas @jbednar @philippjfr ?

If this can be solved quickly then I can work on this PR to get it into the next minor release. However, if this is a deeper issue (as I suspect it is) then this PR will need to wait for updates in param at the very least. Assuming this has a quick solution, the next steps would be:

- [x] Fix the dictionary keyword issue when the parameter values are initialized with `None`.
- [x] See if this dictionary support can be pushed further down into `Apply` and/or `Dynamic`.

### DIFF
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -216,9 +216,6 @@ class Operation(param.ParameterizedFunction):
         elif 'streams' not in kwargs:
             kwargs['streams'] = self.p.streams
 
-
-        if isinstance(kwargs['streams'], dict):
-            kwargs['streams'] = streams.streams_list_from_dict(kwargs['streams'])
         kwargs['per_element'] = self._per_element
         kwargs['link_dataset'] = self._propagate_dataset
         kwargs['link_inputs'] = self.p.link_inputs
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -11,7 +11,6 @@ from .options import Store
 from .overlay import NdOverlay, Overlay
 from .spaces import Callable, HoloMap
 from . import util, Dataset
-from .. import streams
 
 
 class Operation(param.ParameterizedFunction):
@@ -215,7 +214,6 @@ class Operation(param.ParameterizedFunction):
                 return self._apply(element)
         elif 'streams' not in kwargs:
             kwargs['streams'] = self.p.streams
-
         kwargs['per_element'] = self._per_element
         kwargs['link_dataset'] = self._propagate_dataset
         kwargs['link_inputs'] = self.p.link_inputs
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -11,6 +11,7 @@ from .options import Store
 from .overlay import NdOverlay, Overlay
 from .spaces import Callable, HoloMap
 from . import util, Dataset
+from .. import streams
 
 
 class Operation(param.ParameterizedFunction):
@@ -214,6 +215,10 @@ class Operation(param.ParameterizedFunction):
                 return self._apply(element)
         elif 'streams' not in kwargs:
             kwargs['streams'] = self.p.streams
+
+
+        if isinstance(kwargs['streams'], dict):
+            kwargs['streams'] = streams.streams_list_from_dict(kwargs['streams'])
         kwargs['per_element'] = self._per_element
         kwargs['link_dataset'] = self._propagate_dataset
         kwargs['link_inputs'] = self.p.link_inputs
--- a/holoviews/core/operation.py
+++ b/holoviews/core/operation.py
@@ -59,7 +59,7 @@ class Operation(param.ParameterizedFunction):
        visualization should update this stream with range changes
        originating from the newly generated axes.""")
 
-    streams = param.List(default=[], doc="""
+    streams = param.ClassSelector(default=[], class_=(dict, list), doc="""
         List of streams that are applied if dynamic=True, allowing
         for dynamic interaction with the plot.""")
 
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -96,9 +96,9 @@ class ResamplingOperation(LinkableOperation):
         width and height.
     """)
 
-    streams = param.List(default=[PlotSize, RangeXY], doc="""
-        List of streams that are applied if dynamic=True, allowing
-        for dynamic interaction with the plot.""")
+    streams = param.ClassSelector(default=[PlotSize, RangeXY], class_=(dict, list), doc="""
+       List or dictionary of streams that are applied if dynamic=True,
+       allowing for dynamic interaction with the plot.""")
 
     element_type = param.ClassSelector(class_=(Dataset,), instantiate=False,
                                         is_instance=False, default=Image,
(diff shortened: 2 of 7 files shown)


jlstevens:
One possible solution to the `owner` problem is e75743f05: simply make sure all parameters exist on each stream class without relying on inheritance.

The logic is that 1) this mechanism is primarily about linked streams, you may as well use parameters directly for the old non-linked case 2) linked streams have to be tied to the plotting backend which means they are unlikely to be user defined (and if they are, users would just need to know that they shouldn't rely on parameter inheritance) and 3) we can do something like the following to enforce that all the streams we provide are defined according to the rule:


```python
objs = [getattr(hv.streams, el) for el in dir(hv.streams)]
stream_classes = [obj for obj in objs if isinstance(obj, type) and  issubclass(obj, hv.streams.Stream) ]
for stream_class in stream_classes:
    for name, p in stream_class.param.params().items():
        if name != 'name':
            assert p.owner == stream_class, 'Owner %s and class %s' % (p.owner, stream_class)
```

jbednar:
Sounds good. Seems like only linked streams need testing like that at this time.

holoviews/core/operation.py
```diff
@@ -214,6 +215,10 @@ def __call__(self, element, **kwargs):
                 return self._apply(element)
         elif 'streams' not in kwargs:
             kwargs['streams'] = self.p.streams
+
+
+        if isinstance(kwargs['streams'], dict):
+            kwargs['streams'] = streams.streams_list_from_dict(kwargs['streams'])
```
philippjfr, line 221: I think we want something else here tbh, I'll take a look at this once you're done.

jlstevens:
@philippjfr I've pushed the changes we discussed in e71782f . Please do push better fixes to how the `Params` stream is handled!

holoviews/core/operation.py
jlstevens: Sure. I would also like to push this logic deeper into `Apply` and `Dynamic` so it can all be more generally applicable. This version was all about seeing is this whole idea is at all feasible...

holoviews/tests/teststreams.py
```diff
@@ -28,6 +28,18 @@ def test_all_stream_parameters_constant():
                                 % (name, stream_cls.__name__))
 
 
+def test_all_linked_stream_parameters_owners():
+    "Test to ensure operations can accept parameters in streams dictionary"
+    stream_classes = param.concrete_descendents(LinkedStream)
+    print(stream_classes)
```
philippjfr, line 34: Stray print

```suggestion
```

holoviews/util/__init__.py
```diff
@@ -929,9 +929,29 @@ def _get_streams(self, map_obj, watch=True):
...
+                    if p.name != name:
+                        streams[p.owner][p.name] = name
+                    else:
+                        streams[p.owner] = {}
+                else:
+                    params[name] = p
+            stream_specs = streams_list_from_dict(params)
+            stream_specs += [stream(rename=rename) for stream, rename in streams.items()]
+        else:
+            stream_specs = self.p.streams
```
philippjfr, line 950: This is my improved suggestion. If you want move this part out into a utility. I prefer this approach because it doesn't needlessly wrap stream parameters into `Params` streams but rather simply creates the stream instance the parameter is referring to and also avoids creating one stream per parameter by grouping by the parameter owner.

holoviews/core/operation.py
```diff
@@ -11,6 +11,7 @@
 from .overlay import NdOverlay, Overlay
 from .spaces import Callable, HoloMap
 from . import util, Dataset
+from .. import streams
```
philippjfr, line 14:
```suggestion
```

holoviews/core/operation.py
```diff
@@ -214,6 +215,7 @@ def __call__(self, element, **kwargs):
                 return self._apply(element)
         elif 'streams' not in kwargs:
             kwargs['streams'] = self.p.streams
+
```
philippjfr, line 218:
```suggestion
```

holoviews/streams.py
```diff
@@ -773,8 +776,21 @@ def reset(self):
...
+        if self._rename:
+            owner_updates = defaultdict(dict)
+            for (owner, pname), rname in self._rename.items():
+                if rname in kwargs:
+                    owner_updates[owner][pname] = kwargs[rname]
+            for owner, updates in owner_updates.items():
+                if isinstance(owner, Stream):
+                    owner.update(**updates)
+                else:
+                    owner.param.set_param(**kwargs)
```
philippjfr, line 788:
```suggestion
                    owner.param.set_param(**updates)
```

holoviews/util/__init__.py
jlstevens: I do prefer this. It relies on the `p.owner` being the correct stream class which you can then pass `rename` to (and I like the use of default dict here). This avoids the dangling stream instances that were created only to pass their parameters to `Params`. Thanks!

philippjfr:
Also happy with this now, just needs a few unit tests.

holoviews/util/__init__.py
jlstevens: No need to pull out a utility as I don't think it would make much sense out of context. Instead I've added a few clarifying comments in 324871a97

holoviews/operation/datashader.py
```diff
@@ -1742,7 +1742,8 @@ class inspect_mask(Operation):
        Size of the mask that should match the pixels parameter used in
        the associated inspection operation.""")
 
-    streams = param.List(default=[PointerXY])
+    streams = param.ClassSelector(default=dict(x=PointerXY.param.x,
+                                               y=PointerXY.param.y), class_=(dict, list))
```
philippjfr, line 1746: Why change these, the original seemed simpler in this case?

jlstevens: Addressed in dbb4ce390

holoviews/tests/operation/testdatashader.py
```diff
@@ -1126,3 +1130,22 @@ def test_points_inspection_5px_mask(self):
         points = inspect_points(self.img, max_indicators=3, dynamic=False, pixels=5, x=-0.1, y=-0.1)
         self.assertEqual(points.dimension_values('x'), np.array([0.2, 0.4, 0]))
         self.assertEqual(points.dimension_values('y'), np.array([0.3, 0.7, 0.99]))
+
+    def test_points_inspection_dict_streams(self):
+        Tap.x, Tap.y = 0.4, 0.7
+        points = inspect_points(self.img, max_indicators=3, dynamic=True,
+                                pixels=1, streams=dict(x=Tap.param.x, y=Tap.param.y))
+        self.assertEqual(len(points.streams), 1)
+        self.assertEqual(isinstance(points.streams[0], Tap), True)
```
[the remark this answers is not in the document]
jlstevens: Sure!

jlstevens:
@philippjfr Addressed your latest suggestions and the tests are green again. Merge?

PR #4826, opened 2021-02-08
Generalized inspect operation

jlstevens:
WIP.

This generalized operation inspects the pipeline to figure out which concrete `inspect_X` operation to dispatch to:

PR2

The idea is that for most uses, users only ever need to know about `inspect`.

### TODO

- [x] Make sure the `hits` dataframe is mirrored properly.
- [x] Support the `mask` operation.
- [x] Try to reduce the repetition of parameter definitions.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1935,3 +1935,67 @@ class inspect_poly(inspect_base):
         dx, dy = np.array(xs) - x, np.array(ys) - y
         distances = pd.Series(dx*dx + dy*dy)
         return df.iloc[distances.argsort().values]
+
+
+class inspect(Operation):
+    """
+    Generalized inspect operation that detects the appropriate indicator
+    type.
+    """
+
+    pixels = param.Integer(default=3, doc="""
+       Number of pixels in data space around the cursor point to search
+       for hits in. The hit within this box mask that is closest to the
+       cursor's position is displayed.""")
+
+    null_value = param.Number(default=0, doc="""
+       Value of raster which indicates no hits. For instance zero for
+       count aggregator (default) and commonly NaN for other (float)
+       aggregators. For RGBA images, the alpha channel is used which means
+       zero alpha acts as the null value.""")
+
+    value_bounds = param.NumericTuple(default=None, length=2, allow_None=True, doc="""
+       If not None, a numeric bounds for the pixel under the cursor in
+       order for hits to be computed. Useful for count aggregators where
+       a value of (1,1000) would make sure no more than a thousand
+       samples will be searched.""")
+
+    hits = param.DataFrame(default=pd.DataFrame(), allow_None=True)
+
+    max_indicators = param.Integer(default=1, doc="""
+       Maximum number of indicator elements to display within the mask
+       of size pixels. Points are prioritized by distance from the
+       cursor point. This means that the default value of one shows the
+       single closest sample to the cursor. Note that this limit is not
+       applies to the hits parameter.""")
+
+    transform = param.Callable(default=identity, doc="""
+      Function that transforms the hits dataframe before it is passed to
+      the Points element. Can be used to customize the value dimensions
+      e.g. to implement custom hover behavior.""")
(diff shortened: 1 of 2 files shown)


holoviews/operation/datashader.py
```diff
@@ -1935,3 +1935,71 @@ def _sort_by_distance(cls, raster, df, x, y):
...
+      Function that transforms the hits dataframe before it is passed to
+      the Points element. Can be used to customize the value dimensions
+      e.g. to implement custom hover behavior.""")
+
+    # Stream values and overrides
+    streams = param.ClassSelector(default=dict(x=PointerXY.param.x,
+                                               y=PointerXY.param.y),
+                                  class_=(dict, list))
+    x = param.Number(default=0)
+    y = param.Number(default=0)
```
philippjfr, line 1982: Can't you inherit all this from `inspect_base`?

jlstevens: There is a clash on the `_process` method but I agree that I would like to avoid this parameter duplication. Maybe `inspect` can own the parameters, `inspect_base` can inherit from it?

philippjfr: Doesn't the process just get overridden anyway?

jlstevens: That will work but I'll have to define this once the concrete operations exist:

```
    _dispatch = {Points : inspect_points,
                 Polygons : inspect_poly}
```

philippjfr: No objection to changing the hierarchy though.

jlstevens: Implemented in bbf79f094

jlstevens: I guess it doesn't have to be a linear chain but this approach means `inspect` doesn't have a load of methods inherited from `inspect_base` that it doesn't need.

jlstevens:
@philippjfr In 8741623e5 I added mirroring of the `hits` parameter using an internal instance. It works but I'm wondering if you have a better idea for how to do this (I think the current approach is ok, but not great).

philippjfr:
Pushed what I think is a cleaner approach. Didn't change the `watch` approach to mirroring the `hits` parameter though.

jlstevens:
Your suggestion looks good to me. Never copied the `ParamOverrides` across like that but it makes sense if it works. You also added the check to make sure the input type doesn't change that I had been thinking about.

Happy to merge?

holoviews/tests/operation/testdatashader.py
```diff
@@ -1144,8 +1154,16 @@ def test_points_inspection_dict_streams_instance(self):
...
\ No newline at end of file
+        self.assertEqual(points.streams[0].y, 0.3)
+
+    def test_polys_inspection_1px_mask(self):
+        if spatialpandas is None:
+            raise SkipTest('Polygon inspect tests require spatialpandas')
+        polys = inspect_polygons(self.polysrgb,
+                                 max_indicators=3, dynamic=False, pixels=1, x=-6, y=5)
+        self.assertEqual(polys.dimension_values('x'), np.array([6, 3, 7, 6]))
+        self.assertEqual(polys.dimension_values('y'), np.array([7, 2, 5, 7]))
```
philippjfr, line 1169: Would write this as:

```suggestion
        self.assertEqual(polys, hv.Polygons([{'x': [6, 3, 7, 6], 'y': [7, 2, 5, 7]}]))
```

philippjfr: Don't think that's quite right, still missing the vdim.

jlstevens: Sure. Happy to do the asserts that way...

jlstevens: Implemented in 7bba4800a

jlstevens:
Given that the tests are passing and that this API (within the datashader module at least) is nicer than having to import `inspect_points` and `inspect_polygons` individually, I'll go ahead and merge.

PR #4843, opened 2021-02-22
Fix for Dimension pickling

### LINKED ISSUE #4838: Incorrect pickle dump/load for Dimension
Holoviews 1.14.1, Python 3.7.6

`hv.Dimension` does not pickle correctly. As shown in example below, dimension loaded from pickle is different than the original.

```python
import holoviews as hv
import pickle

path = "tmp.pkl"
d = hv.Dimension('current', label='I', unit='mA')

with open(path, 'wb') as handle:
    pickle.dump(d, handle)

with open(path, 'rb') as handle:
    d_loaded = pickle.load(handle)
    
print(repr(d))
print(d.pprint_label)

print(repr(d_loaded))  # repr does not match
print(d_loaded.pprint_label)  # note that label is not used

```
result:
```
Dimension('current', label='I', unit='mA')
I (mA)
Dimension('current')
current (mA)
```
whereas I would expect to see:
```
Dimension('current', label='I', unit='mA')
I (mA)
Dimension('current', label='I', unit='mA')
I (mA)
```

jlstevens:
Fixes #4838 by removing the custom `__setstate__` on `Dimension`.

The key thing it was doing `self.label = self.name` which was needed a long time ago before `Dimension` has separate `name` and `label` concepts. The question is whether we still need this to try to unpickle *really* old pickles (>4 years old) in which case a fix for the custom `__setstate__` is needed instead of simply removing it.

### DIFF
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -348,6 +348,14 @@ class Dimension(param.Parameterized):
         """
         return hash(self.spec)
 
+    def __setstate__(self, d):
+        """
+        Compatibility for pickles before alias attribute was introduced.
+        """
+        super(Dimension, self).__setstate__(d)
+        if '_label_param_value' not in d:
+            self.label = self.name
+
     def __eq__(self, other):
         "Implements equals operator including sanitized comparison."
 
--- a/holoviews/core/dimension.py
+++ b/holoviews/core/dimension.py
@@ -348,13 +348,6 @@ class Dimension(param.Parameterized):
         """
         return hash(self.spec)
 
-    def __setstate__(self, d):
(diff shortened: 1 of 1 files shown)


philippjfr:
Hmm, that was always wrong seemingly. We should have checked if 'label' is in the state and only if it wasn't copied the name over.

jlstevens:
Agreed. The question now is what to do moving forwards.

philippjfr:
Seems to me that's still the right fix going forward.

jlstevens:
@philippjfr I've updated things so pickling now works without losing the fallback when the label parameter is missing. Happy to merge?

Issue #4842, opened 2021-02-22
MatplotlibDeprecationWarning - modifying the state of a globally registered colormap


randallpittman:
#### ALL software version info

OS: Linux Mint 20.1 Ulyssa
python 3.8.7
matplotlib 3.3.4
bokeh 2.2.3
holoviews 1.14.1
jupyter 1.0.0

#### Description of expected behavior and the observed behavior

In developing [colorcet PR #62](https://github.com/holoviz/colorcet/pull/62), I was testing my changes in the examples
with the `colorcet.plotting.swatches()`, which uses holoviews. For each plotted swatch, I got the following warning from matplotlib (python path omitted for brevity):

>  _\<python path\>_/site-packages/holoviews/plotting/mpl/element.py:982: MatplotlibDeprecationWarning: You are modifying the
>  state of a globally registered colormap. In future versions, you will not be able to modify a registered colormap in-place. To
>  remove this warning, you can make a copy of the colormap first.

#### Complete, minimal, self-contained example code that reproduces the issue

In a Jupyter Notebook:

```python
import holoviews as hv
from colorcet.plotting import swatch
hv.notebook_extension("matplotlib")
swatch("fire")
```

#### Screenshots or screencasts of the bug in action
![mpl_deprecation](https://user-images.githubusercontent.com/7116641/108668507-2fbd9b80-7490-11eb-86d1-66e9983181ae.png)

jlstevens:
Thanks for reporting this! I see Philipp has already added this to the next milestone which should be out soon...

jlstevens:
The merged PR referenced above fixes this issue. Closing.

PR #4852, opened 2021-03-02
Copying matplotlib cmap to avoid mutating global colormaps

jlstevens:
Fix for https://github.com/holoviz/holoviews/issues/4842.

Before:

image

After:

image


jlstevens:
@philippjfr Linux tests have now passed for all python versions. I don't think we need to wait for OSX and windows to pass before merging this simple fix...

PR #4859, opened 2021-03-06
Fix wrong subsection level for version number

Hoxbro:
Fixes a wrong subsection level for version number in https://holoviews.org/releases.html


philippjfr:
Thanks @Hoxbro!

PR #4898, opened 2021-04-19
Change how upper limit is taken for color_levels

### LINKED ISSUE #4897: Color levels not respected
holoviews                 1.13.5                   pypi_0    pypi
Python 3.8.5 (default, Sep  4 2020, 02:22:02) [Clang 10.0.0 ] :: Anaconda, Inc. on darwin

When passing colour_levels to an image opts `color_levels` I would expect them to be respected eg. the colormap is broken on the limits that are specified.

As you can see in the example below the color_levels specifies `0` as a limit however the purple step includes values greater than 0. (In the data provided this is off the eastern African coast)

data here: https://github.com/CEREGE-CL/netcdf_editor_app/blob/main/data/standard.nc
```
import xarray as xr
import holoviews as hv
import numpy
hv.extension('bokeh')

ds = xr.open_dataset('data/standard.nc')
var = 'Z'

abs_max = numpy.abs((ds[var].min(), ds[var].max())).max()

delta = 1000
clevels = [-abs_max] + [*numpy.arange(0, abs_max, delta)[::-1] * -1, * numpy.arange(0, abs_max, delta)[1:]] + [abs_max]
print('color_levels: ', clevels)

img = hv.Image(
    ds[var],
    group="Map",
).opts(
    invert_axes=True, tools=['hover'], width=600, aspect=2, colorbar=True, cmap='Category10', clim=(-abs_max, abs_max), color_levels = clevels
)

img

```

Screenshot 2021-04-19 at 10 42 18

WesleyTheGeolien:
Fixes #4897 

I believe that the offset is due to the N parameter in the `plotting.utils.color_intervals` code and how it is used if there is a clip.

Because of the int(round(...)) we can end up with more values in the camp than the N parameter. So when a slice is taken we should use the `Len(cmap)` rather than `N` which introduces an offset if `len(camp) != N`


philippjfr:
Thanks @WesleyTheGeolien! Will have to update the related tests:

>FAIL: test_color_intervals_clipped (holoviews.tests.plotting.testplotutils.TestBokehPaletteUtils)
>FAIL: test_path_continuously_varying_color_op (holoviews.tests.plotting.bokeh.testpathplot.TestPathPlot)

philippjfr:
Also not totally following the correctness of this fix, wouldn't it be more correct not to round and to use `floor` instead?

hoxbro:
With this change the reported problem now looks like this:

image

Respecting the limit, and also better colorbar axis.

PR #4901, opened 2021-04-19
Check before derefferencing mapper

rafiyr:


rafiyr:
Will fix s/mapper/mappers/ and resubmit

PR #4902, opened 2021-04-19
Check before derefferencing mappers

rafiyr:


rafiyr:
@philippjfr a small patch for the problem introduced in fb8c4263438b85602c70c6378f60fd15ed47c4e3

philippjfr:
Thanks @rafiyr!

PR #4904, opened 2021-04-21
Name dim transform params properly

### LINKED ISSUE #4903: `hv.dim` doesn't support multiple parametrized inputs per operation
If you try to have more than one parametrized input per deferred `hv.dim` operation, one of the widgets doesn't update.
This is because `hv.dim.params` overwrites keys in the returned dictionary for more than one input, causing the problem seen in holoviz/hvplot#593 (see there for a MRE).

Looking at the source for the `params` accessor:
```
@property
def params(self):
    params = {}
    for op in self.ops:
        op_args = list(op['args'])+list(op['kwargs'].values())
        for op_arg in op_args:
            ...
            if (isinstance(op_arg, param.Parameter) and
                isinstance(op_arg.owner, param.Parameterized)):
                  params[op_arg.name+str(id(op))] = op_arg      
    return params
```
All dictionary keys of `params` must be unique; to ensure this, `str(id(op))` is part of the key.
However, if one `op` maps to multiple `op_arg`s, then this no longer holds.
Changing it to `str(id(op_arg))` fixes holoviz/hvplot#593.

`holoviews 1.14.2`
`hvplot 0.7.1`

nitrocalcite:
Fixes #4903 and [holoviz/hvplot/#593](https://github.com/holoviz/hvplot/issues/593)

### DIFF
--- a/holoviews/util/transform.py
+++ b/holoviews/util/transform.py
@@ -322,7 +322,7 @@ class dim(object):
                     params.update(op_arg.params)
                 if (isinstance(op_arg, param.Parameter) and
                     isinstance(op_arg.owner, param.Parameterized)):
-                    params[op_arg.name+str(id(op))] = op_arg
+                    params[op_arg.name+str(id(op_arg))] = op_arg
                 
         return params
 


philippjfr:
Perfect, thank you! We'll try to get a new release out with the fix sooner rather than later.

PR #4907, opened 2021-04-23
Updated Dashboards user guide to show pn.bind first

jbednar:
The Dashboards user guide was written before Panel supported `pn.bind`, first showing a Parameterized class and then showing `pn.depends` while saying that the `pn.depends` approach tightly ties the GUI and analysis code together. `pn.bind` doesn't tie it together in that same way, and indeed no longer needs the dummy function that previously was needed so that `pn.depends` could  decorate the analysis function.  All told, I think it's cleaner and clearer to present `pn.bind` first, explicitly creating some widgets and then attaching them to DynamicMap arguments and operation arguments.  So I've done that, motivating it by needing to get a widget to control the operation argument.  Only after that do we get into Parameterized classes.

I think this approach works pretty well, but I'll admit I got lost a bit in the Parameterized sections. I don't think anything there is incorrect, but it seems a bit bewiildering.  What I'd want a user to get out of that section is a solid understanding of how to make dynamic apps with HoloViews + Panel. Panel's own docs can cover everything about how to make apps in general, but having responsive, efficient apps where widgets dynamically affect just a small aspect of a plot seems appropriate to explain here in the HoloViews docs.  I think that's implicit in the current material already, but it would be great if @philippjfr could read through that and more clearly express how one uses the features of HoloViews and Panel to make efficiently responsive apps.

### DIFF
--- a/examples/getting_started/1-Introduction.ipynb
+++ b/examples/getting_started/1-Introduction.ipynb
@@ -210,7 +210,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "points = hv.Points(station_info, ['lon','lat'])\n",
+    "points = hv.Points(station_info, ['lon','lat']).opts(color=\"red\")\n",
     "image + image * points"
    ]
   },
@@ -218,8 +218,9 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "\n",
     "On the left, we have the visual representation of the ``image`` object we declared. Using ``+`` we put it into a ``Layout`` together with a new compositional object created with the ``*`` operator called an ``Overlay``. This particular overlay displays the station positions on top of our image, which works correctly because the data in both elements exists in the same space, namely New York City.\n",
+    "\n",
+    "The `.opts()` method call for specifying the visual style is part of the HoloViews options system, which is described in the next ['Getting started' section](2-Customization.ipynb).\n",
     "\n",
     "This overlay on the right lets us see the location of all the subway stations in relation to our midnight taxi dropoffs. Of course,  HoloViews allows you to visually express more of the available information with our points.  For instance, you could represent the ridership of each subway by point color or point size. For more information see [Applying Customizations](../user_guide/03-Applying_Customizations.ipynb)."
    ]
@@ -306,15 +307,13 @@
     "\n",
     "composition.opts(\n",
     "    opts.Image(xrotation=90),\n",
-    "    opts.Points(color='deepskyblue', marker='v', size=6))"
+    "    opts.Points(color='red', marker='v', size=6))"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The `.opts()` method call for specifying the visual style is part of the HoloViews options system, which is described in the next ['Getting started' section](2-Customization.ipynb).\n",
-    "\n",
     "In the cell above we created and styled a composite object with a few short lines of code. Furthermore, this composite object relates tabular and array data and is immediately presented in a way that can be explored interactively. This way of working enables highly productive exploration, allowing new insights to be gained easily. For instance, after exploring with the slider we notice a hotspot of taxi dropoffs at 7am, which we can select as follows:"
    ]
   },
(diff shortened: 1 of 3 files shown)


MarcSkovMadsen:
I don't really understand that we have two names `pn.bind` and `pn.depends`. I can see they have a little bit different functionality and `pn.depends` is used as a decorator. But in principle they are doing the same job in my understanding. Binding a function to parameters/ widgets.

philippjfr:
All the changes look good, but I didn't see anything about the benefits of `pn.bind` (basically addressing the same question @MarcSkovMadsen is asking). I know what my answer is but it doesn't seem like it was addressed in the docs.

jbednar:
>  I didn't see anything about the benefits of pn.bind

Compared to what, pn.depends? I don't recommend pn.depends to anyone and my revised document doesn't mention or use pn.depends at all, and so it wouldn't make sense to include a comparison to pn.depends here in this document.

@MarcSkovMadsen, pn.depends was designed to be used at the point of function definition, while pn.bind is used later, typically at the point when a function is passed to Panel. The pros and cons are laid out at https://github.com/holoviz/panel/issues/1629, but my own opinion is that late binding is vastly preferable, as it keeps Panel-specific things together and separate from your analysis functions. pn.depends can be used for late binding but it has a confusing and unnatural syntax when used in that way, hence the need for pn.bind. Personally, I recommend scrubbing pn.depends from all of our docs; I cannot think of any case where it's the best approach, and introducing it just adds to confusion.

(param.depends, on the other hand, is perfectly reasonable; param.depends is meant for writing a class, where late binding is not relevant.)

philippjfr:
>Compared to what, pn.depends? I don't recommend pn.depends to anyone and my revised document doesn't mention or use pn.depends at all, and so it wouldn't make sense to include a comparison to pn.depends here in this document.

But you do use it in the class based context so you could say what you just told @MarcSkovMadsen. Which is that in a functional paradigm where you declare widgets explicitly (instead of parameters like in the class-based approach), `pn.bind` allows you to bind the widgets after the fact and keeps your function definition clean of UI specific consideration (i.e. the widgets). You don't have to point out that you can use `pn.depends` in a function based context too, to make that point.

jbednar:
I'm making a distinction here between param.depends, which I would use in the class-based context, and pn.depends, which I would use no where. Are you saying I should compare pn.bind with param.depends, then?

philippjfr:
Yes, at least something to say that both approaches allow you to separate UI concerns from your business logic and then maybe hint at that fact that combining pn.depends + widgets does not do that and that's the reason we're not recommending it. Just feel like people will have seen pn.depends around and might wonder why you wouldn't just use that everywhere.

jbednar:
I tried editing it just now to mention pn.depends, but I did not see any way to include it here that wouldn't be confusing and seem like a _non_ _sequitur_.  The problem is that this PR reorganizes the examples to show pn.bind first, because I think that's a better first thing to introduce, but pn.depends doesn't seem motivated until later, when param.depends is used. And later it it's very confusing to bring up pn.depends, which doesn't apply to the Parameterized case at all. 

Personally, I don't see any need to mention pn.depends here in the HoloViews docs. Sure, in the Panel docs, but why in HoloViews? This is a very basic Panel example, and the user needs to go to the Panel docs for just about everything at this point, so it seems unmotivated to start talking about pn.depends here just to say "don't use it". So I'd vote for merging this PR as-is unless there's a problem with the actual content included.

MarcSkovMadsen:
hmmm. My feel is that having both `param.depends`/ `pn.depends` and `pn.bind` will only complicate things instead of making things easier.

Is there not a way to implement `pn.depends` such that it can both work as the decorator and as `pn.bind` does? Enabling removal of `pn.bind` again.

jbednar:
The decorator works in a tricky way that makes it very awkward to make it behave normally like pn.bind does.  I don't think it needs to be confusing, as long as we eliminate pn.depends from all our examples. pn.bind is better in every example I have ever seen.  Then we have param.depends for the parameterized case, which is a class-based approach where late binding is not relevant, and then pn.bind for the widget+callback approach, where late binding is crucial. Two very different cases, two different solutions, both good.

Issue #3190, opened 2018-11-22
Documentation suggestion for arguments to hv.Path


ea42gh:
the hv.Path argument variations are complicated enough to warrant additional documentation.

For example:
Given 3 curves with coordinates x, y held in arrays of shape (N,3),
the matplotlib statement to plot these is  `plt.plot(x,y)`.

Using hv.Path, it is `hv.Path(zip(x.T, y.T, range(3)), vdims='color').options(color='color')`

ea42gh:
I see! OK.
I'll need some time to figure out all the various ways of setting the path.
Will run it by you when I have something

Issue #2601, opened 2018-04-21
Paths do not display legends


ea42gh:
```
hv.Path( [ [[1,4],[4,1]],   [[3,3],[2,2]]], label='foo' )*hv.Path( [ [[4,4],[1,1]],   [[3,2],[2,3]]], label='bah' )
```
does not produce legends

ea42gh:
Found the option:  show_legend=True
To be consistent with other elements, should this be the default?

jlstevens:
It seems there are some element types where you want legends on my default and others where you don't (e.g Curves versus Paths). If we decide to make such a distinction, what we need to do is document it and list the elements based on their defaults.

ea42gh:
Not sure why you say that? I have to explicitly add a `label=...`  to the call to get the legend.
So with paths, this becomes `hv.Path(...,label='tata').options( show_legend=True)`.

jlstevens:
It comes down to a judgment about how ``Paths`` are used. Whenever I have used this element myself, I know  I would always be switching the legend off. Of course, that doesn't mean it reflects everybody's experience so it is a value judgement about sensible default behavior.

ea42gh:
But that is the question I was trying to raise: if you know you do not want the paths to have a legend,
would you not define
`labeled_elements * hv.Path( data )`  as opposed to
`labeled_elements * hv.Path( data, label='path')` ?

jbednar:
Right; that's the question: @jlstevens, it's true that there are many uses of paths where a legend is not appropriate.  But are there a lot of uses for Paths *with labels* where a legend is not appropriate?  I agree with @ea42gh that a typical user would simply not declare the label in that case, and thus it seems to me that show_legend ought to be true here both to support that and to make things more consistent and predictable for users.

philippjfr:
I agree consistency is preferable here, especially since you still have to define a ``label`` to get a legend.

jlstevens:
I suppose when labels are defined you should have the same default for showing the legend.

I think I got confused thinking of ``Contours``, surely you can show the legend for the contour level values within a single ``Contours`` element without specifying a label? Or is this not possible?

philippjfr:
The value judgement is indeed more relevant to Contours/Polygons here, adding a ``label`` based legend isn't controversial I don't think. The problem is that Contours/Polygons support adding legends for each unique contour/polygon generated from the value dimension that is being colormapped in which case enabling the legend by default can be pretty horrible.

[...]

PR #4919, opened 2021-05-04
changed default show_legend = True for hv.Path elements

ea42gh:


ea42gh:
Addresses issue #2601

philippjfr:
Sorry I didn't get around to review this. The problem here is that in many cases you are plotting *a lot* of paths which might be colored by different values and you do not want to generate a legend for this case by default. This is the original reason this was disabled by default, so simply setting it to true is not something I'm okay with merging. I don't currently have an alternative suggestion though.

jbednar:
Should it respect the `legend_limit` based on the length of the path, as Overlays do?

ea42gh:
We had gone through a discussion of this in the past:
https://github.com/holoviz/holoviews/issues/2601
The conclusion there appears to have been that legends only appear if explicitley requested
with a label='...' argument.   Caveat Emptor

I also found
https://github.com/holoviz/holoviews/issues/3190
which I had totally forgotten about...

I seem to get bitten by this about once a year:
https://discourse.holoviz.org/t/paths-do-not-display-legends-by-default/2252

PR #4941, opened 2021-05-19
Tile source replacement warnings

jlstevens:
After quite some discussion with @jbednar  and @philippjfr, here is one proposal for how to handle the tile sources that no longer work.

To replace the `Wikipedia` tile source (showing how you set your own default to disable the warning):

image

And now that `CartoMidnight` and `CartoEco` no longer work:

image

I think this is as good as we can do once we explain the changes in the release notes. That said, I'm sure we can improve the warning messages further...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,9 @@
 Version 1.14.4
 ==============
 
-This release primarily focuses on a number of bug fixes. Many thanks
-to @Hoxbro, @nitrocalcite, @brl0, @hyamanieu, @rafiyr, @jbednar and
-@philippjfr for contributing.
+This release primarily focuses on a number of bug fixes. Many thanks to
+@Hoxbro, @nitrocalcite, @brl0, @hyamanieu, @rafiyr, @jbednar, @jlstevens
+and @philippjfr for contributing.
 
 Enhancements:
 
@@ -40,6 +40,15 @@ Documentation:
 - Updated docs to correctly declare Scatter kdims
   ([#4914](https://github.com/holoviz/holoviews/pull/4914))
 
+Compatibility:
+
+- The `Wikipedia` tile source is no longer available as it is no longer
+  being served outside the wikimedia domain. As one of the most
+  frequently used tile sources, HoloViews now issues a warning and
+  switches to the OpenStreetMap (OSM) tile source instead.
+- The `CartoMidnight` and `CartoEco` tile sources are no longer publicly
+  available. Attempting to use these tile sources will result in a
+  deprecation warning.
 
 Version 1.14.3
 ==============
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,11 @@ Documentation:
 
 Compatibility:
 
+Unfortunately a number of tile sources are no longer publicly
+available. Attempting to use these tile sources will now issue warnings
+unless `hv.config.raise_deprecated_tilesource_exception` is set to
+`True` in which case exceptions will be raised instead.
+
 - The `Wikipedia` tile source is no longer available as it is no longer
   being served outside the wikimedia domain. As one of the most
   frequently used tile sources, HoloViews now issues a warning and
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -163,16 +163,6 @@ class Config(param.ParameterizedFunction):
        Global default colormap for HeatMap elements. Prior to HoloViews
        1.14.0, the default value was the 'RdYlBu_r' colormap.""")
 
-    wikimedia_tile_source_replacement = param.String(default='OSM-with-warning', doc="""
-       If set to the special sentinal value of OSM-with-warning, replace
-       wikimedia tile source with OpenStreetMap (OSM) and warn. If set
-       explicitly with a suitable Tile source URL, apply replacement for
-       wikimedia tile source without warning.
-
-       This config parameter was introduced as wikimedia tile sources
-       can no longer be used outside the wikimedia domain, as of
-       April 2021.""")
-
     def __call__(self, **params):
         self.param.set_param(**params)
         return self
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -163,6 +163,10 @@ class Config(param.ParameterizedFunction):
        Global default colormap for HeatMap elements. Prior to HoloViews
        1.14.0, the default value was the 'RdYlBu_r' colormap.""")
 
+    raise_deprecated_tilesource_exception = param.Boolean(default=False,
+       doc=""" Whether deprecated tile sources should raise a
+       deprecation exception instead of issuing warnings.""")
+
     def __call__(self, **params):
         self.param.set_param(**params)
         return self
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -170,7 +170,8 @@ class Config(param.ParameterizedFunction):
        wikimedia tile source without warning.
 
        This config parameter was introduced as wikimedia tile sources
-       can no longer be used outside the wikimedia domain.""")
+       can no longer be used outside the wikimedia domain, as of
+       April 2021.""")
 
     def __call__(self, **params):
         self.param.set_param(**params)
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -163,6 +163,15 @@ class Config(param.ParameterizedFunction):
        Global default colormap for HeatMap elements. Prior to HoloViews
        1.14.0, the default value was the 'RdYlBu_r' colormap.""")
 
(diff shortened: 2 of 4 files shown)


jlstevens:
Warning instead of raising an exception:

image

holoviews/core/util.py
```diff
@@ -163,6 +163,15 @@ class Config(param.ParameterizedFunction):
...
        1.14.0, the default value was the 'RdYlBu_r' colormap.""")
 
+    wikimedia_tile_source_replacement = param.String(default='OSM-with-warning', doc="""
+       If set to the special sentinal value of OSM-with-warning, replace
+       wikimedia tile source with OpenStreetMap (OSM) and warn. If set
+       explicitly with a suitable Tile source URL, apply replacement for
+       wikimedia tile source without warning.
+
+       This config parameter was introduced as wikimedia tile sources
+       can no longer be used outside the wikimedia domain.""")
```
jbednar, line 173:
```suggestion
       can no longer be used outside the wikimedia domain, as of
       April 2021.""")
```

holoviews/element/tiles.py
```diff
@@ -157,8 +169,21 @@ def easting_northing_to_lon_lat(easting, northing):
 EsriReference = lambda: Tiles('https://server.arcgisonline.com/ArcGIS/rest/services/Reference/World_Reference_Overlay/MapServer/tile/{Z}/{Y}/{X}', name="EsriReference")
 ESRI = EsriImagery # For backwards compatibility with gv 1.5
 
+
+def wikimedia_replacement():
+    if util.config.wikimedia_tile_source_replacement == 'OSM-with-warning':
+        param.main.param.warning('Wikimedia tile source no longer available outside '
+                                 'wikimedia domain, switching to OpenStreetMap (OSM) tile '
```
jbednar, line 176:
```suggestion
                                 'wikimedia domain as of April 2021; switching to OpenStreetMap (OSM) tile '
```

holoviews/element/tiles.py
```diff
@@ -131,11 +131,23 @@ def easting_northing_to_lon_lat(easting, northing):
     )
 }
 
+def deprecation_warning(name, url, reason):
+    def deprecated_tilesource_warning():
+        param.main.param.warning('%s tile source is deprecated: %s' %  (name, reason))
```
jbednar, line 136:
```suggestion
        param.main.param.warning('%s tile source is deprecated and is likely to be unusable: %s' %  (name, reason))
```

jbednar (approved):

holoviews/element/tiles.py
```diff
@@ -131,11 +131,23 @@ def easting_northing_to_lon_lat(easting, northing):
...
-CartoEco = lambda: Tiles('https://3.api.cartocdn.com/base-eco/{Z}/{X}/{Y}.png', name="CartoEco")
 CartoLight = lambda: Tiles('https://cartodb-basemaps-4.global.ssl.fastly.net/light_all/{Z}/{X}/{Y}.png', name="CartoLight")
-CartoMidnight = lambda: Tiles('https://3.api.cartocdn.com/base-midnight/{Z}/{X}/{Y}.png', name="CartoMidnight")
+CartoMidnight = deprecation_warning('CartoMidnight',
+                                    'https://3.api.cartocdn.com/base-midnight/{Z}/{X}/{Y}.png',
+                                    'this tile source is no longer publicly available.')
+CartoEco = deprecation_warning('CartoEco',
+                               'https://3.api.cartocdn.com/base-eco/{Z}/{X}/{Y}.png',
+                               'this tile source is no longer publicly available.')
+
```
jbednar, line 150:
```suggestion
                                    'no longer publicly available.')
CartoEco = deprecation_warning('CartoEco',
                               'https://3.api.cartocdn.com/base-eco/{Z}/{X}/{Y}.png',
                               'no longer publicly available.')

```

holoviews/core/util.py
```diff
@@ -163,6 +163,16 @@ class Config(param.ParameterizedFunction):
...
 
+    wikimedia_tile_source_replacement = param.String(default='OSM-with-warning', doc="""
+       If set to the special sentinal value of OSM-with-warning, replace
+       wikimedia tile source with OpenStreetMap (OSM) and warn. If set
+       explicitly with a suitable Tile source URL, apply replacement for
+       wikimedia tile source without warning.
+
+       This config parameter was introduced as wikimedia tile sources
+       can no longer be used outside the wikimedia domain, as of
+       April 2021.""")
```
philippjfr, line 174: Damn GitHub reviews, thought I'd left this comment earlier but turns out I hadn't hit submit.

Personally I really don't think this config option is needed and don't think anyone should use it. Aliasing the Wikipedia tile source to refer to something other than the Wikipedia tile sources is weird and we're only doing it because it's used a bunch (and it's the default if you do `tiles=True` in hvplot). Any actual user should switch to the tile source they actually want, not do this aliasing.

jbednar: Seems like we need to change the hvplot default as well, then.

philippjfr: Definitely.

philippjfr: Fixed here: https://github.com/holoviz/hvplot/pull/610

jlstevens: I am not sure I agree...the idea is that some people might be fine with replacing wikipedia with OSM throughout but don't want to keep seeing a warning. I would be fine making it into a bool if setting a different replacement than OSM seems weird...

jlstevens: Or in other words, some people may be more annoyed by warnings appearing in old notebooks than the fact that the tile source that says it is wikipedia is now OSM...

philippjfr: >(and may not want to update all their existing code/notebooks).

But to set the config option they would be doing this anyway, no? The difference in effort between doing it once at the top and doing a search and replace doesn't seem particularly high, and actually replacing the tile source is preferable since we do want to eventually remove Wikipedia (since keeping the inaccurate alias around forever does not seem like a great idea). 

Anyway no very strong opinions here but I've always been pretty inclined against config options to control backward compatibility options since they aren't very discoverable and I prefer nudging people to do the right thing rather than letting them live in their outdated world any longer and thereby forcing them to make edits to their code twice.

jlstevens: > But to set the config option they would be doing this anyway, no?

Not necessarily: people can use the holoviews.rc to do it once only.

Happy to let @jbednar be the tie breaker!

philippjfr: >Not necessarily: people can use the holoviews.rc to do it once only (or in other ways that are environment/sitewide)

I actually also really dislike the idea of rc files setting options like this because it leads to reproducibility nightmares.

jbednar: I don't mind using rc files to set backwards compatibility options, because by definition the code involved is already broken. Doing that makes it _more_ reproducible than it was, in the sense that having set it things will now mostly work.  So I'd say such compatibility options are just about the only reason to use rc files, and only as a temporary measure while knowing you eventually need to update the scattered source code.

Still, in this case, I don't think adding the configurability is very useful, because it only addresses one of the tile sources, with the others still broken. So I'm inclined to simply swap in for Wikipedia with a warning, only because it's the default in current hvplot versions, and otherwise for there simply to be a warning and require that people update their code to use a working tile source.

jlstevens: > I actually also really dislike the idea of rc files setting options like this because it leads to reproducibility nightmares.

I feel the same way and only want rc files to be used when absolutely necessary (but it can be a useful escape hatch when needed as Jim says). 

> So I'm inclined to simply swap in for Wikipedia with a warning...

Ok, I'll get rid of the config option then.

philippjfr: If you could also add the changelog then we can get this merged and released.

jlstevens:
@jbednar @philippjfr How about a config option to allow the breakage of `CartoEco` and `CartoMidnight` be raised as exceptions instead of just warnings? 

Even if these tile sources aren't used very often, that would allow user to be *more* strict and detect if they have broken tile sources (e.g if using nbsmoke). To be honest, this makes me want to be the wikipedia config option back in, though this time to do the same thing and raise exceptions (which can make it easier to update the broken tile sources everywhere).

Maybe a single option would do: `hv.config.deprecated_tilesource_exception` (boolean and False by default)?

CHANGELOG.md
```diff
@@ -42,6 +42,11 @@ Documentation:
 
 Compatibility:
 
+Unfortunately a number of tile sources are no longer publicly
```
philippjfr, line 45: Great, but sadly need you to add this to doc/releases.rst as well.

jlstevens: Thanks for the reminder. I'll do that now.

jlstevens: Updated in 32d4af7bb

PR #4949, opened 2021-05-25
Switch to pytest

philippjfr:


stonebig:
Hi  Holoviews team,

This line of patch didn't make it to Holoviews-1.14.7 (nor 1.14.8 yet) , and seems to be an issue on Python-3.10

Could you include it ?

image

PR #4951, opened 2021-05-26
Fixed missing handling of nodata for count aggregator with column

jlstevens:
The `count_cat` aggregators were handled but `count` wasn't. Simple fix that shouldn't need much discussion once a corresponding unit test is added.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -315,7 +315,11 @@ class AggregationOperation(ResamplingOperation):
                                  "Ensure the aggregator references an existing "
                                  "dimension." % (column,element))
             if isinstance(agg_fn, (ds.count, ds.count_cat)):
-                vdims = dims[0].clone('%s %s Count' % (vdim_prefix, column), nodata=0)
+                if vdim_prefix:
+                    vdim_name = '%s%s Count' % (vdim_prefix, column)
+                else:
+                    vdim_name = '%s Count' % column
+                vdims = dims[0].clone(vdim_name, nodata=0)
             else:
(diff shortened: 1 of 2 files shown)


jbednar (approved):

jbednar (approved):
Thanks!

jlstevens:
The test failure for 3.6 ubuntu seems unrelated to this PR. I tried restarting the tests but the same failure reappeared:

```
self = <holoviews.plotting.bokeh.callbacks.PointDrawCallback object at 0x7f197416c908>
plot_id = None

    def initialize(self, plot_id=None):
        plot = self.plot
        stream = self.streams[0]
>       cds = plot.handles['source']
E       KeyError: 'source'
```

@philippjfr Any idea what this failure is about? Is it simply a transient that happened twice in a row?

At any rate, happy to see this PR merged now.

holoviews/operation/datashader.py
```diff
@@ -297,8 +297,11 @@ def _get_agg_params(self, element, x, y, agg_fn, bounds):
         params = dict(get_param_values(element), kdims=[x, y],
                       datatype=['xarray'], bounds=bounds)
 
-        kdim_list = '_'.join(str(kd) for kd in params['kdims'])
-        vdim_prefix = self.vdim_prefix.format(kdims=kdim_list)
+        if self.vdim_prefix is not None:
```
philippjfr, line 300: Not real different but no point doing the kdims join if there is no prefix string:

```suggestion
        if self.vdim_prefix:
```

jlstevens: Sure!

PR #4953, opened 2021-05-27
Removed unneeded list from dynamic map example (docs)

jenssss:
The `frequencies` list is defined, but not used or needed in the Dynamicmap reference notebook


jbednar (approved):
Thanks for the fix!

PR #4954, opened 2021-05-27
Optimize and clean up options system

philippjfr:
A number of optimizations improvements for the options system:

1. Options objects are no longer Parameterized's
2. Store maintains a `_lookup_cache` per backend which resets every time an option is set.

The lookup cache in particular is very important, in my particular example it reduced the number of lookups and therefore instantiation of Options objects from 2168 to 8, shaving off almost 40% of the rendering time.

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -598,12 +598,13 @@ class OptionTree(AttrTree):
     """
 
     def __init__(self, items=None, identifier=None, parent=None,
-                 groups=None, options=None, **kwargs):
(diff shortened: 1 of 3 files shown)


holoviews/core/options.py
```diff
@@ -32,19 +32,23 @@
    extension together.
 
 """
-import pickle
-import traceback
 import difflib
 import inspect
+import functools
```
philippjfr, line 37:
```suggestion
```

PR #4956, opened 2021-05-27
Ensure Params stream is treated as linked when it mirrors LinkedStream parameters

### LINKED ISSUE #4955: Streams not working with overlays when using bind or streams dict
This works when no overlay is involved:

```python
tap = hv.streams.Tap()
def test(x, y):
    return hv.VLine(x or 0)
dmap = hv.DynamicMap(pn.bind(test, x=tap.param.x, y=tap.param.y))
hv.Curve([1, 2, 3])  * dmap
```

image

But introducing an overlay breaks it:

```python
tap = hv.streams.Tap()
def test(x, y):
    return hv.Curve([1, 2, 3]) * hv.VLine(x or 0)
dmap = hv.DynamicMap(pn.bind(test, x=tap.param.x, y=tap.param.y))
dmap
```

Using the streams dictionary has the same issue:

```python
tap = hv.streams.Tap()
def test(x, y):
    return hv.Curve([1, 2, 3]) * hv.VLine(x or 0)
dmap = hv.DynamicMap(test, streams=dict(x=tap.param.x, y=tap.param.y))
dmap
```

The old streams list approach is also still working.

```python
>>> param.__version__, bokeh.__version__, hv.__version__
('1.10.1', '2.3.1', '1.14.4')
```

philippjfr:
Fixes #4955

### DIFF
--- a/holoviews/plotting/util.py
+++ b/holoviews/plotting/util.py
@@ -21,7 +21,7 @@ from ..core.util import (
     closest_match, is_number, isfinite, python2sort, disable_constant,
     arraylike_types
 )
-from ..streams import LinkedStream
+from ..streams import LinkedStream, Params
 from ..util.transform import dim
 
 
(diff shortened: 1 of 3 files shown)


philippjfr:
>Any idea when this regression happened? I am assuming that this was working at some point

Nope, never worked.

jlstevens:
> Nope, never worked.

I guess the streams dictionary format is fairly new so not *that* surprising. Guess no one has been using `bind` to return overlays either!

PR #4958, opened 2021-05-27
Hotfix version 1.14.5

jlstevens:
Currently just cherry picks #4956 but the list should be extended with all the other fixes we want to include.

Cherry picked:

- [x] #4956

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -91,13 +91,6 @@ jobs:
           doit develop_install $CHANS_DEV -o unit_tests
           pip uninstall -y holoviews
           doit pip_on_conda
-      - name: matplotlib patch
-        run: |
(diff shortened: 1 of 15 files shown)


philippjfr:
Tests failing due to https://github.com/holoviz/holoviews/issues/4959 which is currently pinned on master. Instead of porting the pin back to this branch we should just fix it and then port the compatibility fix back.

jlstevens:
Conda is having trouble running out of memory on 3.6 for linux and windows but all the other builds are passing.

I've dropped spatialpandas as an optional dependency for now until it is released and working with pandas >= 1.3.0. That might be one reason to release 1.14.6...

jlstevens:
Not closing this PR until I port my cftime change and plotly unit test changes back to master...

Issue #4968, opened 2021-06-10
AttributeError when saving a layout containing holomap with title set 


jenssss:
#### ALL software version info
holoviews                 1.14.4             pyhd3eb1b0_1
matplotlib                3.3.4            py39h06a4308_0
python                    3.9.5                h12debd9_4
Ubuntu 20.04.2 LTS

#### Description of expected behavior and the observed behavior

I get an `AttributeError` when trying to save a layout containing a holomap to a file when title has been set and using the matplotlib backend.

#### Complete, minimal, self-contained example code that reproduces the issue

Upon saving the following code to a file
```
import holoviews as hv
import numpy as np

hv.extension("matplotlib")

frequencies = [0.5, 0.75, 1.0, 1.25]

def sine_curve(phase, freq):
    xvals = [0.1* i for i in range(100)]
    return hv.Curve((xvals, [np.sin(phase+freq*x) for x in xvals]))

curve_dict = {f:sine_curve(0,f) for f in frequencies}
hmap = hv.HoloMap(curve_dict, kdims='frequency')
layout = hmap + hmap
layout.opts(title="")

hv.save(layout, 'sines.svg', fmt='svg')
```
and running `python holomap_export.py` I get the following `AttributeError`

<details>
 <summary>Error trace</summary>

```
Traceback (most recent call last):
  File "/home/jens/gits/mol_1d/mol_1d/AAf/test/holomap_export.py", line 18, in <module>
    hv.save(layout, 'sines.svg', fmt='svg')
  File "/home/jens/miniconda3/envs/test_hv/lib/python3.9/site-packages/holoviews/util/__init__.py", line 819, in save
    return renderer_obj.save(obj, filename, fmt=fmt, resources=resources,
  File "/home/jens/miniconda3/envs/test_hv/lib/python3.9/site-packages/holoviews/plotting/renderer.py", line 609, in save
    plot, fmt = self_or_cls._validate(obj, fmt)
  File "/home/jens/miniconda3/envs/test_hv/lib/python3.9/site-packages/holoviews/plotting/renderer.py", line 314, in _validate
    plot = self.get_plot(obj, renderer=self, **kwargs)
  File "/home/jens/miniconda3/envs/test_hv/lib/python3.9/site-packages/holoviews/plotting/renderer.py", line 243, in get_plot
    plot.update(init_key)
[...]

PR #4969, opened 2021-06-10
Workaround for AttributeError

jenssss:
See #4968


holoviews/plotting/mpl/plot.py
```diff
@@ -280,7 +280,7 @@ def _update_title(self, key):
         title = self._format_title(key) if self.show_title else ''
         if 'title' in self.handles:
             self.handles['title'].set_text(title)
-        else:
+        elif self.handles['axis'].figure is not None:
```
philippjfr, line 283: Let's be completely safe I guess:

```suggestion
        elif 'axis' in self.handles and self.handles['axis'].figure is not None:
```

Issue #4961, opened 2021-05-28
First call to different stream triggers unnecessary  calls in overlay


jlstevens:
Here is the example code:

```python
class Example(param.Parameterized):
    a= param.Event()
    b = param.Integer(default=0)

def cb_A(a):
    print('Callback A')
    return hv.Curve(np.random.rand(10))

def cb_B(b):
    print('Callback B')
    return hv.Curve(np.random.rand(10))

dmapa = hv.DynamicMap(cb_A, streams=dict(a=Example.param.a))
dmapb = hv.DynamicMap(cb_B, streams=dict(b=Example.param.b))
dmapa * dmapb
```

That on initialisation looks like this:

<img src="https://user-images.githubusercontent.com/890576/120017211-3980fe80-bfab-11eb-8b47-d6d167f9e0a7.png" width=60%></img>

As expected both callbacks are invoked to generate the initial display of the overlay.

Now triggering parameter `a`:

<img src="https://user-images.githubusercontent.com/890576/120017328-63d2bc00-bfab-11eb-8e9e-53735dbca37c.png" width=60%></img>

As expected only the blue curve has updated and callback `A` is called.

Now triggering parameter `b`:

<img src="https://user-images.githubusercontent.com/890576/120017472-88c72f00-bfab-11eb-9637-9e88546fbe63.png" width=60%></img>

Now only the red curve updates as expected but *both* callbacks are called which I would *not* expect. This means the random number generator is hit in callback `A` but the result is not displayed.

Subsequent updates to `b` only hit callback `B` (as expected) and only the red curve is updated.

<img src="https://user-images.githubusercontent.com/890576/120017675-c62bbc80-bfab-11eb-85e0-9f13889ad6eb.png" width=60%></img>

Now if `a` is triggered and then `b`, both callbacks are called when switching back to `b`:

<img src="https://user-images.githubusercontent.com/890576/120017880-03904a00-bfac-11eb-9f66-e14367fae4fa.png" width=60%></img>

This behavior makes it hard to work with callbacks that are stateful as a callback (here `B`) can be invoked when none of its parameters change.

PR #4976, opened 2021-06-16
Fixing CI tests

jlstevens:
I'm repurposing this PR to be a fix for tests and will open a new one tackling the original goals:

## Original PR goals

This PR aims to address https://github.com/holoviz/holoviews/issues/4961

Using the example described in that issue, here is the behavior before:

</img>

And the behavior now:

</img>

The ideas is as follows:

1. Event parameters should *not* be part of the memoization as event parameters are fundamentally transient (e.g. button clicks etc). For this reason they are now dropped from `Params.hashkey`.
2. Nonetheless, triggering event parameters should invoke the callback. This means `Params.transient` is True when any events are occurring and False otherwise.
3. I've disabled setting `transient` in the Params constructor as it seems to depend on the parameters used. Though @philippjfr maybe the behavior should be: `user_provided_transient_value` or `transient_if_event_parameters_are_triggered`? That would allow `transient` to be True for all none Event parameters (if that even makes sense and has a use case).

Will see how much the tests like this change shortly...


jlstevens:
@philippjfr One worry I have...

If you press a button to update the state of the visualization, *that* latest state should be associated with the (now shortened) memoization key. I suspect transient will just skip all that and drop the latest state returned by the callback. 

In other words, I think we want the result of the callback when `transient is True` to be memoized, even though the memoization key won't have changed. This is a little stateful (which is what memoization doesn't normally expect) but I think that would be the desired behavior here (e.g you click a button, the plot updates, now if the callback is called with the old signature, the memoized value shouldn't return the value before the update).

Does that make sense?

jlstevens:
I now think my concern above is moot: I can't create an example of the behavior I was worrying about. I think my concern might be valid for other streams but not `Param` streams where these `Event` parameters might appear. I *think* the reason it is ok is:

* If you set a value to its old value, nothing is triggered at the param level and nothing happens.
* If you call `foo.param.trigger` on a parameter, you are explicitly triggering it so you do expect the callback to be invoked (which it is).

Honestly, I'm still not certain though...I feel there must be some case where an older version of the return is memoized after an event has caused an update, causing a stale version to be displayed. At least it does seem to be a rare edge case if so!

jlstevens:
@philippjfr The one test failure seems to be a transient. Ready for an initial review...

jlstevens:
This PR has turned out to be more about getting the tests passing that about handling event parameters. I'll revert the changes to `streams.py` so we can merge and get tests passing on master (the one failure on windows for 3.6 is when conda runs out of memory). 

I'll then open a new PR pointing to this one where I reintroduce the transient behavior for `Event` parameters and add the unit tests.

jlstevens:
@jbednar Now ready to merge to get the tests passing on master (assuming the tests pass above, except for maybe windows + python 3.6 due to the conda memory error). The two things to note:

* I pinned `pandas <1.3.0` as 1.3.0 is causing lots of errors right now (111 unit test failures). Obviously this will need fixing but I think it is important to get the tests going green again.
* The `bokeh/label/dev` channel is added. I believe @philippjfr instructed me to do this: presumably the latest panel dev release now relies on the latest bokeh dev releases.

I'll then open a new PR for the changes I wanted to make to the `Param` stream. This time it should be easier to add unit tests now the rest of the tests are passing again.

jlstevens:
All tests passing except 3.6 on Windows (due to the out of memory issue with conda).

@jbednar Ready to merge.

PR #5005, opened 2021-07-06
Fix regressions in dim transforms

philippjfr:
Namespace object does not necessarily have the attribute you're looking up and the class may not be initialized yet when `__getattribute__` is called the first time.


jlstevens:
I've tested that this change fixes the unit tests in a separate PR. Merging.

PR #5014, opened 2021-07-14
fix: Rectangles and Segments color

### LINKED ISSUE #4981: Rectangles do not use default color cycle
holoviews: 1.14.4
jupyterlab: 2.3.1
bokeh: 2.3.2
matplotlib: 3.4.2

Overlaying rectangles should select different colors, just like Curve or Points.

```
import holoviews as hv
hv.extension('bokeh')

# All rectangles are blue
fig = hv.Rectangles([(0,0,1,1), (4,4,5,5)]) *  hv.Rectangles([(2,2,3,3)])
fig

# Rectangles now use the default color cycle
fig.options({'Rectangles': {'color': hv.Cycle('default_colors')}})
```

By default, no color cycle is used:
image

Once set to the default one, it works as expected
image

douglas-raillard-arm:
* Remove duplicated setting of Rectangles options
* Make Segments default to black, like all the other "line-only" shapes
* Make Rectangles use the default color cycle

Fixes #4981


holoviews/plotting/bokeh/__init__.py
```diff
@@ -191,11 +191,10 @@ def colormap_generator(palette):
 options.Ellipse = Options('style', color='black')
 options.Polygons = Options('style', color=Cycle(), line_color='black',
                            cmap=dflt_cmap)
-options.Rectangles = Options('style', cmap=dflt_cmap)
-options.Segments = Options('style', cmap=dflt_cmap)
 
 # Geometries
-options.Rectangles = Options('style', line_color='black')
+options.Rectangles = Options('style', color=Cycle(), cmap=dflt_cmap, line_color='black')
+options.Segments = Options('style', color='black', cmap=dflt_cmap)
```
jbednar, line 197: What's the motivation for having one of these be a cycle and the other be a fixed color?

douglas-raillard-arm: It looked like all the other "line-only" things like Segments are drawn in black, like error bars. I suppose it depends on the main use case.

For the Rectangles, I suppose it is an oversight when it was added, as the options were set twice with different values. Polygons sets a facecolor as well. OTH, things like Ellipse do not have any filling by default, so maybe Rectangles should also default to border-only in black.

douglas-raillard-arm: The current behavior is not aligned to anything else I tried though, since it is filled by default with blue. Elements like Ellipse are not filled by default at all.

douglas-raillard-arm:
As I went through my workarounds, I realized this was still open. @jbednar This rebases cleanly on `main`.

PR #5015, opened 2021-07-14
Allow plotting anonymous pandas.Series

### LINKED ISSUE #4970: Cannot plot unnamed pd.Series
holoviews: 1.14.4
jupyterlab: 2.3.1
bokeh: 2.3.2

Plotting pandas Series with `name` attribute set to None raises an exception.

```
import pandas as pd

series = pd.Series([1,2])
print(series.name)
# Raises exception
hv.Curve(series)

# Works because accessing df['a'] returns a series with name == 'a'
df = pd.DataFrame(dict(a=series))
hv.Curve(df['a'])

# Works as well
series2 = pd.Series([1,2])
series2.name = 'foo'
hv.Curve(series2)

# Fails with:
# ValueError: Dimension name cannot be the empty string
series3 = pd.Series([1,2])
series3.name = ''
hv.Curve(series3)
```

```
DataError                                 Traceback (most recent call last)
<ipython-input-29-41d62e9ba886> in <module>
      3 series = pd.Series([1,2])
      4 print(series.name)
----> 5 hv.Curve(series)

venv-3.9/lib/python3.9/site-packages/holoviews/element/selection.py in __init__(self, *args, **kwargs)
     19 
     20     def __init__(self, *args, **kwargs):
---> 21         super(SelectionIndexExpr, self).__init__(*args, **kwargs)
     22         self._index_skip = False
     23 

venv-3.9/lib/python3.9/site-packages/holoviews/element/chart.py in __init__(self, data, kdims, vdims, **params)
     50         if len(params.get('kdims', [])) == self._max_kdim_count + 1:
     51             self.param.warning('Chart elements should only be supplied a single kdim')
---> 52         super(Chart, self).__init__(data, **params)
     53 
     54     def __getitem__(self, index):

venv-3.9/lib/python3.9/site-packages/holoviews/core/data/__init__.py in __init__(self, data, kdims, vdims, **kwargs)
    339 
    340         validate_vdims = kwargs.pop('_validate_vdims', True)
--> 341         initialized = Interface.initialize(type(self), data, kdims, vdims,
    342                                            datatype=kwargs.get('datatype'))
    343         (data, self.interface, dims, extra_kws) = initialized

venv-3.9/lib/python3.9/site-packages/holoviews/core/data/interface.py in initialize(cls, eltype, data, kdims, vdims, datatype)
    254                 continue
    255             try:
--> 256                 (data, dims, extra_kws) = interface.init(eltype, data, kdims, vdims)
    257                 break
    258             except DataError:

venv-3.9/lib/python3.9/site-packages/holoviews/core/data/pandas.py in init(cls, eltype, data, kdims, vdims)
     72                     break
     73             if any(isinstance(d, (np.int64, int)) for d in kdims+vdims):
---> 74                 raise DataError("pandas DataFrame column names used as dimensions "
     75                                 "must be strings not integers.", cls)
     76 

DataError: pandas DataFrame column names used as dimensions must be strings not integers.

PandasInterface expects tabular data, for more information on supported datatypes see http://holoviews.org/user_guide/Tabular_Datasets.html
```

douglas-raillard-arm:
If a Series has no name (empty string or None), uses a default "y" name
to avoid exceptions.

Fixes #4970

### DIFF
--- a/holoviews/core/data/pandas.py
+++ b/holoviews/core/data/pandas.py
@@ -33,7 +33,8 @@ class PandasInterface(Interface):
         kdim_param = element_params['kdims']
         vdim_param = element_params['vdims']
         if util.is_series(data):
-            data = data.to_frame()
+            name = data.name or util.anonymous_dimension_label
+            data = data.to_frame(name=name)
         if util.is_dataframe(data):
             ncols = len(data.columns)
             index_names = data.index.names if isinstance(data, pd.DataFrame) else [data.index.name]
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -36,6 +36,8 @@ timedelta_types = (np.timedelta64, dt.timedelta,)
 arraylike_types = (np.ndarray,)
 masked_types = ()
 
+anonymous_dimension_label = '_'
+
 try:
     import pandas as pd
 except ImportError:


jbednar:
Thanks! Would "" work?

douglas-raillard-arm:
Agreed, "y" is a completely arbitrary. `" "` gives:
```

  File "/home/raillard/WFH/holoviews/holoviews/plotting/mpl/chart.py", line 76, in get_data
    style = self._apply_transforms(element, ranges, style)

  File "/home/raillard/WFH/holoviews/holoviews/plotting/mpl/element.py", line 557, in _apply_transforms
    v = dim(v)

  File "/home/raillard/WFH/holoviews/holoviews/util/transform.py", line 225, in __init__
    self.dimension = Dimension(obj)

  File "/home/raillard/WFH/holoviews/holoviews/core/dimension.py", line 277, in __init__
    raise ValueError('Dimension name cannot be the empty string')

ValueError: Dimension name cannot be the empty string
```
I suppose something strips it somewhere

douglas-raillard-arm:
Maybe `'-'` is better than `y` as it is some sort of neutral placeholder, and is minimally invasive visually

jbednar:
`_` seems like a good compromise, then. IPython already uses that for a placeholder name...

douglas-raillard-arm:
@jbednar I updated the PR with the underscore

jbednar:
The test failures (at least the first ones I noticed) seem to be due to spatialpandas, not this PR.

douglas-raillard-arm:
Yes, running pytest on master in a fresh venv currently brings a few failures, especially on Python 3.9. Some of them are coming from pandas (or maybe warnings), which is to be expected as pandas dropped 3.6 support, so testing on 3.6 can only use an old pandas frozen in time.

jbednar:
@philippjfr has fixed one of the Pandas issues, and @jlstevens is investigating the rest.  The fact that this PR works on 3.6 strongly implies that it will be fine once the Pandas issues are addressed separately, so I think it could be merged now. Ok, @philippjfr ?

philippjfr:
Happy merging this but if we're doing this I think we should define the name of the anonymous dimension globally somewhere and reuse it consistently, e.g. to allow plotting an unnamed carry.DataArray

douglas-raillard-arm:
Maybe we should allow `None` or `""` in the dataframe column name and then convert that to `"_"` inside `dimension_name()` ? This way all data sources should get the same treatment, while avoiding empty strings later in the pipeline where it can create problems.
https://github.com/holoviz/holoviews/blob/master/holoviews/core/dimension.py#L68

jbednar:
Ok, it sounds like the only thing left to do on this PR is to "define the name of the anonymous dimension globally somewhere and reuse it consistently, e.g. to allow plotting an unnamed xarray.DataArray". Presumably that just requires adding `anonymous_dimension_label='_'` somewhere. @philippjfr, any suggestion where to put that definition so that it would be accessible for any data type like that?

philippjfr:
`holoviews/core/util.py` is where most of those global definitions live. That file needs to be broken up but for now I'd probably still put it in there.

douglas-raillard-arm:
@philippjfr @jbednar Updated the PR with `holoviews.core.utils.anonymous_dimension_label = '_'`

jbednar (approved):
Looks good to me if the failing tests can be ignored. Thanks!

PR #5018, opened 2021-07-15
Sanitize uses of contextlib.contextmanager

douglas-raillard-arm:
* try/finally around the yield statement in order to execute the 2nd
  part if an exception is raised in the with statement block

* "except: raise" is unncessary, that is the default behaviour of any
  try/finally construct.


philippjfr:
Nice, thank you. Hope to have CI working properly again tomorrow but since 3.6 tests are passing and the changes look good I'll merge.

PR #5024, opened 2021-07-16
Align Bars with annotations on matplotlib

### LINKED ISSUE #4989: ErrorBars * Bars broken on matplotlib
holoviews: 1.14.4
jupyterlab: 2.3.1
bokeh: 2.3.2
matplotlib: 3.4.2

Error bars should be at the tip of the bars, not floating next to it.
Alignment is correct on bokeh but broken on matplotlib.

```
import holoviews as hv

hv.extension('matplotlib')
cats = ['a', 'b']
levels = [3, 4]

bars = hv.Bars(zip(cats, levels)).opts(alpha=1)
errors = hv.ErrorBars(zip(cats, levels, [0.5, 0.2], [1.5, 1]), vdims=['y', 'yerrneg', 'yerrpos'])
fig = bars * errors
fig
```

With matplotlib:
image

With bokeh:
image

Note that bokeh error bar is hidden by the bar itself, but that's for another bug report.

douglas-raillard-arm:
Align the bars with the annotations such as ErrorBars or Labels.

Fixes #4989


douglas-raillard-arm:
Not 100% sure if that fixes the issue in all paths

philippjfr:
I agree this seems like the right fix, the only worry I have here is that users in the past may have worked around the issue and now we're breaking all their workarounds.

douglas-raillard-arm:
Maybe an explicit changelog entry would solve that problem so that people are aware they can remove their workaround ?

holoviews/plotting/mpl/chart.py
```diff
@@ -983,12 +983,13 @@ def _create_bars(self, axis, element, ranges, style):
...
                     # mini offset needed or else combines with non-continuous
                     goffset += width / 1000
-
-                xpos = gidx+goffset if not continuous else xvals[gidx]
-                if not continuous:
-                    xticks.append(((xpos), grp_label, yalign))
+                if continuous:
+                    xpos = xvals[gidx]
+                else:
+                    xpos = gidx+goffset+0.5
```
philippjfr, line 989:
```suggestion
                    xpos = gidx+goffset
```

philippjfr:
Looks like this was already fixed in https://github.com/holoviz/holoviews/pull/6145

PR #5027, opened 2021-07-16
Added release dates to changelog

ablythed:


jlstevens:
Thanks!

Other than the fact that the date probably shouldn't be part of the heading, this looks great! Currently it looks like this:

</img>

jlstevens:
Looks much better now, thanks!

image

Merging.

Issue #5022, opened 2021-07-16
NdOverlay.hist raises AttributeError


peterroelants:
#### Description of expected behavior and the observed behavior 
Creating a histogram from an `NdOverlay` using `hist()` results in an `AttributeError`.

E.g. the line:
```python
nd_overlay.hist(groupby='cat', dimension=['x','y'])
```
raises:
```python-traceback
AttributeError: 'NoneType' object has no attribute 'hist'
```

I would have expected `NdOverlay.hist()` to create a histogram plot similar to the following figure (Not generated by this code):
<img src="https://user-images.githubusercontent.com/951093/125970995-e750d20e-b531-47cb-8a04-62956803c6ef.png"  width="200">
For example, the resulting histogram plot could be a new `NdOverlay` of histograms of the elements in the original `NdOverlay`.

#### Complete, minimal, self-contained example code that reproduces the issue

Reproduce by running:
```python
import numpy as np
import holoviews as hv

hv.extension('bokeh')


points_dct = {
    'p1': hv.Points(np.random.randn(1000, 2), kdims=['x', 'y'], label='p1'),
    'p2': hv.Points(np.random.randn(1000, 2) * 2 + 1, kdims=['x', 'y'], label='p2')
}
nd_overlay = hv.NdOverlay(points_dct, kdims='cat')
nd_overlay.hist(groupby='cat', dimension=['x','y'])
```

#### Stack traceback and/or browser JavaScript console output
The line `nd_overlay.hist(groupby='cat', dimension=['x','y'])` results in the following error output:
```python-traceback
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_58893/2897274094.py in <module>
      8 }
      9 nd_overlay = hv.NdOverlay(points_dct, kdims='cat')
---> 10 nd_overlay.hist(groupby='cat', dimension=['x','y'])

~/miniconda3/envs/lcms/lib/python3.9/site-packages/holoviews/core/overlay.py in hist(self, dimension, num_bins, bin_range, adjoin, index, **kwargs)
     75             raise TypeError("Please supply a suitable index or label for the histogram data")
     76 
---> 77         hists = self.get(index).hist(
     78             adjoin=False, dimension=dimension, bin_range=bin_range,
     79             num_bins=num_bins, **kwargs)

AttributeError: 'NoneType' object has no attribute 'hist'
```

It seems that the default index is `0`, which is not in the NdOverlay. From the debugger:
[...]

PR #5031, opened 2021-07-18
Show all histograms of an Overlay

peterroelants:
I created a PR to enable showing all histograms of an Overlay:

This PR:
- Fixes `NdOverlay.hist raises AttributeError`: https://github.com/holoviz/holoviews/issues/5022 . `NdOverlay.hist` works in this PR.
- New behaviour: `CompositeOverlay.hist()` shows histograms for all the elements in the overlay. This affects both `NdOverlay` and `Overlay` (e.g. see previous comment @ https://github.com/holoviz/holoviews/issues/5022#issuecomment-882006386)

Illustration of new `CompositeOverlay.hist()` behavior. Notice the histograms being overlaid for both elements in the overlay:

I would love to get some feedback on this PR

### DIFF
--- a/holoviews/core/overlay.py
+++ b/holoviews/core/overlay.py
@@ -7,6 +7,7 @@ Also supplies ViewMap which is the primary multi-dimensional Map type
 for indexing, slicing and animating collections of Views.
 """
 
+from collections import OrderedDict
 from functools import reduce
 import numpy as np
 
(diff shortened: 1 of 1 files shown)


holoviews/core/overlay.py
```diff
@@ -52,7 +53,7 @@ class CompositeOverlay(ViewableElement, Composable):
     _deep_indexable = True
 
     def hist(self, dimension=None, num_bins=20, bin_range=None,
-             adjoin=True, index=0, **kwargs):
+             adjoin=True, index=None, **kwargs):
```
peterroelants, line 56: I kept the `index` functionality, however I decided to change the default behavior to show all histograms (which I believe to be a better default if we have the option).

peterroelants:
I think I still have to write/update some tests related to this. However, I couldn't find a contributor's guide on the Holoviews website that explains the Holoviews testing frameworks and principles. Can anyone provide me a reference to how to write and run tests for Holoviews?

peterroelants:
I also have a notebook that illustrates some of the features of this PR: https://gist.github.com/peterroelants/32fadcdf4d27ff6228178fa9a8a2252b#file-overlay_hists-ipynb

You have to run the notebook yourself on this branch to see the effects (GitHub Gists crashed each time when I tried to upload the notebook with embedded Holoviews/Bokeh plots)

peterroelants:
Some things to be considered:
- Is it possible to inherit colormaps from the parent plot in some way? If so, how might I enable this?
- Currently, the legend of the overlay is replicated on each plot of the Adjoint (when `adjoin=True`), do we want this as a default? Or do we only want to show the legend on the main plot?

jbednar:
I'm traveling and only have my phone, but from the pictures it looks like a big improvement to me! Maybe @jlstevens  or @philippjfr will be able to take a look sooner.

jbednar:
One comment: it might be good to suppress the legend for the histogram, because it's redundant for an AdjointLayout and typically overlaps the bars a lot.

peterroelants:
@jbednar Don't worry, I'm not in rush to get this in.

About hiding the legends of the histograms, I tried looking into it, however it seems that legends are not controlled by Holoviews but by the renderers (Bokeh, Matplotlib). Do you know a way to hide the legends from Holoviews itself?

peterroelants:
@jbednar I updated this to hide the legend in the histograms by default, and added a parameter to still enable them if needed.

Please take your time while travelling. I'm not in a rush to get this in.

peterroelants:
@jbednar I updated this to hide the legend in the histograms by default, and added a parameter to still enable them if needed. I also made sure that axes are aligned in the case of `adjoin=True`

One weird thing that I cant explain is that the histogram plot sizes of Overlays are different compared to the histograms of plain elements. E.g.:


vs



Please take your time while travelling. I'm not in a rush to get this in.

philippjfr:
I think this is a sensible change but since it's a change in behavior would go into 2.0.

peterroelants:
> I think this is a sensible change but since it's a change in behavior would go into 2.0.

@philippjfr Should I file a PR against a different branch? Or leave this as-it-is for now?

philippjfr:
You're good, master is already in "2.0" mode, e.g. a bunch of deprecations for 2.0 have already happened.

jlstevens:
Thanks for the PR!

This change seems appropriate to me and as already mentioned this is a good time to allow backwards incompatible behavior.   

I've not yet checked what unit tests are running here but I would want some tests that make sure `index` still works to select a particular layer when necessary.

jbednar:
Looks great to me! @jlstevens , can you see how to avoid having the size change like that? The smaller size (equivalent to adding .opts(width=120) (or so) if I recall correctly) seems much more appropriate for an adjoint histogram, but I don't know why the size is changing when it's an overlay.

peterroelants:
@jlstevens I rebased this PR on latest master, and most tests seem to be passing. The ones that fail seem to fail due to unrelated reasons:
- [Out of memory in](https://github.com/holoviz/holoviews/pull/5031/checks?check_run_id=3259147156#step:7:230)
- [Plotly import error](https://github.com/holoviz/holoviews/pull/5031/checks?check_run_id=3259147098#step:11:1244)
- [Killed during conda install](https://github.com/holoviz/holoviews/pull/5031/checks?check_run_id=3259147049#step:7:189)
All of them are Python 3.6 tests

jlstevens:
Sorry for letting this languish!

I agree that those test failures are unrelated. I will now approve this PR (looks fine to me) but I will wait a bit to give @philippjfr  to also have a look. I'll merge at the end of the day regardless.

jlstevens (approved):
Looks good to me! 

I don't think we need to worry about backwards compatibility as it will end up on the 2.0 branch. The only thing that would be nice to add would be some additional unit tests...

jlstevens:
Actually, there is the size issue that @jbednar mentions. I'll have a look now...

jlstevens:
The sizing issue is unrelated to this PR: the problem is about how overlays become adjoint which is a separate problem.

@peterroelants  Can you have a look at the issues and report this if it hasn't been already? At which point I am happy to merge.

peterroelants:
@jlstevens I couldn't find an existing issue related to the adjoint overlays resulting in an unexpected plot size. I created a issue for it at https://github.com/holoviz/holoviews/issues/5072

jlstevens:
Thanks! 

I'm going to go ahead and merge. That way we can test and iron out any other issues before the 2.0 release.

PR #5035, opened 2021-07-19
Added release dates to releases.rst

ablythed:


jlstevens:
@ablythed looks good! 

My only worry is that the diff is quite large, probably because something is working differently than pandoc. In particular, the bulletpoints seem to have some extra whitespace indentation and at least some of the header definitions have changed.

Once you get the diff down to just the dates, I'm happy to merge!

ablythed:
@jlstevens Just updated. There were a couple of dates in there already, but some were a few days off from the ones I found in the log so I changed them - let me know if I should do something different there. Thanks for your help!

jlstevens:
Looks good, merging!

Issue #5036, opened 2021-07-20
DynamicMap.hist(dimension='dim') does not respect given dimension


peterroelants:
#### Description of expected behavior and the observed behavior

`DynamicMap.hist(dimension='y')` does not rerun histogram of the `'y'` dimension, but of the `'x'` dimension. It seems that it does not respect the given dimension.

#### Complete, minimal, self-contained example code that reproduces the issue

For example, the following code
```python
from collections import OrderedDict
import numpy as np
import holoviews as hv

hv.extension('bokeh')

def gen_points_overlay():
    points = [
        hv.Points(np.random.randn(1000, 2), kdims=['x', 'y'], label='p1'),
        hv.Points(np.random.randn(1000, 2) + np.array([5, -5]), kdims=['x', 'y'], label='p2')
    ]
    return hv.Overlay(points)

dmap_2d = hv.DynamicMap(gen_points_overlay)
dmap_2d.hist(dimension='y')
```

Results in the following figure:
<img src="https://user-images.githubusercontent.com/951093/126271151-793c84d6-b59c-4d81-a2d1-439e1c3ced70.png" alt="hist-wrong-dimension" width="400">

#### ALL software version info
```
Python implementation: CPython
Python version       : 3.9.6
IPython version      : 7.25.0

Compiler    : GCC 9.3.0
OS          : Linux
Release     : 5.10.0-1034-oem
Machine     : x86_64
Processor   : x86_64
CPU cores   : 16
Architecture: 64bit

holoviews: 1.14.5
numpy    : 1.21.1
bokeh    : 2.3.3

jupyter core     : 4.7.1
jupyter-notebook : 6.4.0
qtconsole        : not installed
ipython          : 7.25.0
ipykernel        : 6.0.3
jupyter client   : 6.1.12
jupyter lab      : 3.0.16
nbconvert        : 6.1.0
ipywidgets       : 7.6.3
nbformat         : 5.1.3
traitlets        : 5.0.5
```

peterroelants:
Please note that `hist()` does respect the dimension when not applied to a `DynamicMap`. So the issue seems to be specifically related to the `DynamicMap`:
```python
from collections import OrderedDict
import numpy as np
import holoviews as hv

hv.extension('bokeh')

points = [
    hv.Points(np.random.randn(1000, 2), kdims=['x', 'y'], label='p1'),
[...]

PR #5037, opened 2021-07-20
Forward DynamicMap.hist dimension parameter to histogram creation

peterroelants:
Fixes https://github.com/holoviz/holoviews/issues/5036

```python
DynamicMap.hist(dimension='y')
```
Results in:
* Before fix:
  hist-wrong-dimension
* After fix:
   hist-wrong-dimension


jbednar:
Looks good to me! The test failures look like they are to do with spatialpandas+pandas1.3, so I think this should be merged.

peterroelants:
Thanks for having a look.
I don't think I can merge myself, so someone with write access to the repo will have to merge this.

philippjfr:
Thanks @peterroelants!

PR #5044, opened 2021-07-23
Add warning to linked selection

peterroelants:
`hv.selection.link_selections` does not work on all objects, if it doesn't know how to recurse into an object it just returns the object (without a warning): https://github.com/holoviz/holoviews/blob/effaf45f9c2e7bf33c9f7ee830c2aa355badaae2/holoviews/selection.py#L180-L182

The absense of any indication that the `link_selections` might not work can be confusing, as I experience myself. I propose adding a warning to signal the user that the `link_selections` might not work as expected.

For example see:

Please let me know if you think this is a good idea

### DIFF
--- a/holoviews/selection.py
+++ b/holoviews/selection.py
@@ -179,6 +179,10 @@ class _base_link_selections(param.ParameterizedFunction):
                     callback.operation)
             else:
                 # This is a DynamicMap that we don't know how to recurse into.
+                self.param.warning(
+                    "linked selection: Encountered DynamicMap that we don't know "
+                    "how to recurse into:\n{!r}".format(hvobj)
+                )
                 return hvobj
         elif isinstance(hvobj, Element):
             # Register hvobj to receive selection expression callbacks


jbednar (commented):
Thanks!  On balance, I agree there should be a warning here, because there is always a way to call link_selections to avoid the warning if it's not relevant.  

I think the warnings should be handled differently for a couple of reasons, though. First, HoloViews does warnings via param using the logging module, so for consistency it would be a call to self.param.warning(). Second, using an f-string here generates the string representation even if the actual display is suppressed due to the logging level settings, which can slow things down. In this particular case, it wouldn't matter since it's behind an if statement and likely to be outweighed by the other slow parts of this operation, but still, I think it's better to keep our logging usage consistent by using old-style (alas, more verbose) string formatting as supported by the logging module (https://docs.python.org/3/howto/logging.html#logging-basic-tutorial).

peterroelants:
> I think the warnings should be handled differently for a couple of reasons, though. First, HoloViews does warnings via param using the logging module, so for consistency it would be a call to self.param.warning(). 

Done! I've updated the code to use `self.param.warning`

> Second, using an f-string here generates the string representation even if the actual display is suppressed due to the logging level settings, which can slow things down. In this particular case, it wouldn't matter since it's behind an if statement and likely to be outweighed by the other slow parts of this operation, but still, I think it's better to keep our logging usage consistent by using old-style (alas, more verbose) string formatting as supported by the logging module (https://docs.python.org/3/howto/logging.html#logging-basic-tutorial).

@jbednar I'm not sure what you mean by using "old-style" string formatting. Do you mean the [old style percentage based templating](https://docs.python.org/3/howto/logging.html#logging-variable-data)? Because even the Python logging documentation mentions that _"This is for backwards compatibility: the logging package pre-dates newer formatting options such as str.format()"_. I'm also unsure how this will prevent things slowing down because the `repr()` rendering should be independent of string formatting, and even the string representation of every `Dimensioned` seems to use the `repr()` representation. 

In the end, I decided to go use `"{!r}.format()` because I saw it used in other places in the codebase. Please let me know if this is ok, if you mean something else, and if I should change the `repr()` representation by another.

jbednar:
> @jbednar I'm not sure what you mean by using "old-style" string formatting. Do you mean the old style percentage based templating? Because even the Python logging documentation mentions that "This is for backwards compatibility: the logging package pre-dates newer formatting options such as str.format()". 

Yes, that's what I'm referring to -- the logging module only supports lazy evaluation for the old formatting options, so you have to use the old formatting when you use the logging module (directly or via self.param.warning) if you want the result to be lazy (and thus efficient). E.g.:

```
>>> import logging as lg
>>> class A():
...     def __repr__(self):
...         print("Running")
...         return "A()"
... 
>>> lg.warning("Msg %s",a)
Running
WARNING:root:Msg A()
>>> lg.info("Msg %s",a)
>>> lg.info("Msg %s" % (a))
Running
>>> lg.info(f"Msg {a}")
Running
>>> 
```

Notice that in all but one of the calls above, the repr is being evaluated, even though in most cases the result is simply discarded anyway. Only one of those options, the one using old-style formatting like `lg.info("Msg %s",a)`, skips evaluating the repr of A (no "Running" message) if the logging message is not actually being shown. Time taken to construct the repr can be quite significant for a complex object `a`, and should be avoided for any call that might be in an inner loop of some code.

Again, here it doesn't really matter since the call is behind an if statement, but @philippjfr and @jlstevens , this is your repo, so can you please state what you expect people to do for logging messages? Looking around the codebase I see only diversity, not a clear indication that logging messages are constructed safely.  That may well be ok, if applied solely to warning and error calls -- they only occur for error conditions anyway, so who cares? But for .info() and .debug() calls we have found cases where the code was many times slower, for no functional reason.  So my proposal would be that people can use any formatting they want for .warning and .error, but need to use old-style for .message, .verbose, and .debug. Sound ok? 

@peterroelants , sorry for dragging you into this discussion; really it's an issue for the development team! :-)

philippjfr:
Just merged for now. As for the logging, I think it's something we've done a terrible job at forever. So could you open an issue to discuss it?

PR #5050, opened 2021-07-29
fix(core.data.interface): check if data.interface is a class

geronimos:
check if data.interface is a class before calling issubclass()

Fixes https://github.com/holoviz/hvplot/issues/645


jbednar (approved):
Looks perfect!

philippjfr:
Thanks @geronimos!

Issue #5057, opened 2021-08-07
Support xyzservices TileProvider objects in hv.Tiles


martinfleis:
#### Is your feature request related to a problem? Please describe.

With the @geopandas team, we have built a new package called [`xyzservices`](https://github.com/geopandas/xyzservices) to serve as a unified place for XYZ tile management. The idea behind this is that other packages will depend on that instead of having their own system and list of supported tiles hard-coded. It would be great to have holoviews on board with the idea, either by supporting `xyzservices.TileProvider` as an input of `hv.Tiles` or even refactoring `tiles.py` module to depend directly on `xyzservices`.

`xyzservices` itself does not have any dependencies which makes it a non-problematic dependency.

It is worth mentioning that `xyzservices` stores providers' metadata in JSON under `share`, accessible for anyone if you have a use case for that.

I am happy to do any changes in `xyzservices` needed to make it as useful for you as possible if that allows centralising maintenance of providers in one place.

#### Describe the solution you'd like

1) support `TileProvider` as an input of `hv.Tiles`. 

```py
import xyzservices.providers as xyz

hv.Tiles(xyz.CartoDB.Positron)
```

All the info `Tiles` require can be fetched from the `TileProvider` object under the hood.

2) refactor tile management

You can remove a big chunk of the maintenance burden by refactoring tile management in `tiles.py` using `xyzservices` directly. In the first step, you can remove all metadata and simply map your existing classes to TileProvider objects. In the second, you can even dynamically support the whole `xyzservices` library of providers instead of the pre-defined set you have now.

#### Describe alternatives you've considered

I know that 2) may sound like a bit ambitious proposal but 1) is pretty straightforward to implement already. So I'd say that having at least the first part would be nice.

#### Additional context
Support of `xyzservices` is being implemented across the ecosystem, from geopandas' own packages (new built-in foilum-based plotting and `contextily`) to `leafmap` (https://github.com/giswqs/leafmap/issues/91) and `ipyleaflet` (https://github.com/jupyter-widgets/ipyleaflet/issues/854) (+ hopefully more).

If there's anything we would need to implement in `xyzservices` to make all this work, I am all ears.

It seems that @philippjfr [would welcome that](https://twitter.com/PhilippJFR/status/1422517252222504961) :).

jlstevens:
Thanks @martinfleis for the suggestion!

Interestingly enough, @jbednar @philippjfr and myself recently discovered `xyzservices` and briefly discussed this suggestion already! 

[...]

PR #5062, opened 2021-08-09
Support xyzservices.TileProvider as hv.Tiles input

martinfleis:
Implements suggestion 1) from #5057.

Bokeh and Plotly backends now support `xyzservices.TileProvider` as input for `hv.Plot`:

```py
import xyzservices.providers as xyz

hv.Tiles(xyz.CartoDB.Positron)
```

I have not added tests for the Bokeh backend as I have no idea where it should be. Can you point me to the right direction?

I'll add the same to GeoViews once this gets reviewed and we agree on specifics of the support.


philippjfr:
This is great, thanks so much! There should probably be a `holoviews/tests/plotting/bokeh/test_tiles.py`. Happy to add the scaffolding for that or you can just copy the rough structure in one of the `test_{element}.py` files in the same directory.

jlstevens:
Thanks for the contribution!

Hopefully we can get the tests on master working again at which point this PR can be rebased and merged.

jlstevens:
Sorry for not getting back to this sooner!

Seems like your last commit didn't trigger the tests. Could you push another commit to see if the tests run this time? Alternatively, you could do another rebase against latest master...

martinfleis:
Seems to work now. Awaiting approval.

jlstevens:
Approved. I'm ready to merge assuming the tests go green. Do you agree @philippjfr ?

martinfleis:
The new tests are all skipped in CI since xyzservices are not in any of envs. As I said above

> We should also add xyzservices to CI envs but I am not certain how that works in here. xyzservices requres Python 3.7 so adding it to setup.py as you do with other deps would probably break your 3.6-based tests.

jlstevens:
The  `xyzservices` package on`conda-forge` is a noarch package and I successfully installed it into a conda 3.6 environment. It looks like the issue with 3.6 is on import (`from __future__ import annotations`) so what might work is to install `xyzservices` into the CI env across python versions and these specific tests can check for `python > 3.6` before importing `xyzservices` inline to run. Does that make sense?

jlstevens:
Note that if you rebase again, the windows tests should pass. It would be nice to get these tests running on CI for Python > 3.6 so I won't merge just yet. @philippjfr Any opinions on whether it is ok to install `xyzservices` but make the tests conditional on the Python version?

jlstevens:
@philippjfr At this point I would be happy to just merge. Do you agree?

philippjfr:
The 3.6 test failures are unrelated and otherwise this is passing. So I'll close.

PR #5082, opened 2021-09-14
Attempt to fix environment solving

philippjfr:
Unpin a bunch of stuff to see if that a) fixes the solve and b) still lets the tests pass.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,12 +46,29 @@ jobs:
           doit ecosystem_setup
           doit env_create ${{ env.CHANS_DEV}} --python=${{ matrix.python-version }}
       - name: doit develop_install
+        if: matrix.os != 'macos-latest'
         run: |
           eval "$(conda shell.bash hook)"
           conda activate test-environment
           conda install ${{ env.CHANS_DEV }} "pip<21.2.1"
           conda list
           doit develop_install ${{ env.CHANS_DEV}} -o ${{ env.HV_REQUIREMENTS }}
+          conda install "panel=0.12.1" "nbconvert>5"
+          python -c "from param import version; print(version.Version.setup_version('.', 'holoviews', archive_commit='$Format:%h$'))"
+          echo "-----"
+          git describe
+          echo "======"
+          conda list
+        # Temporary step for Mac only due doit to develop_install failing
+      - name: doit develop_install on Mac
+        if: matrix.os == 'macos-latest'
+        run: |
+          eval "$(conda shell.bash hook)"
+          conda activate test-environment
+          conda install ${{ env.CHANS_DEV }} "pip<21.2.1"
+          conda list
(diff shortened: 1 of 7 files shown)


maximlt:
Ok this is still failing unfortunately. As for the environment, we add conda-forge in the mix which made things trickier probably. Removing it leads to a state where the env creation is failing on Mac only. It fails with this error:
```
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
TaskFailed - taskid:develop_install
Command failed: 'pip install --no-deps --no-build-isolation -e .' returned -9
```
I've been able to reproduce it locally but I can't get my head around why this fails. I then tried to run the apparently failing command (`pip install ...`) directly and it succeeded. I added this piece in `dodo.py`:
```
import pyctdev._conda
# Mocking python_develop originally set to pip install ...
pyctdev._conda.python_develop = echo "TEST"'
```

And well:
```
Preparing transaction: ...working... done
Verifying transaction: ...working... done
Executing transaction: ...working... done
TaskFailed - taskid:develop_install
Command failed: 'echo "TEST"' returned -9
```

I don't understand what is going on there!

As for the tests, the CI is failing on Python 3.6 because it's pulling a dev release of panel that has an [issue](https://github.com/holoviz/panel/issues/2662) with 3.6. I've seen at least another test failing, due to holoviews itself this time.

jbednar:
Bokeh 2.4 [requires Python 3.7+](https://docs.bokeh.org/en/dev-2.4/docs/releases.html), and with Panel tied tightly to Bokeh, does that mean we have to disable Python 3.6 tests?

philippjfr:
Shouldn't have to since Holoviews does not have to pin tightly to any specific Panel version.

Hoxbro:
A small comment about removing conda-forge is that I think it will only postpone the problems and not fix them, as the packages in default channels are not as up-to-date with the latest releases. This may give new users (and developers) problems when installing the holoviz stack (from pip or with conda-forge), with a newer version of numpy, pandas, or others.

Another thing to be noted is that many scientific packages follow [NEP29](https://numpy.org/neps/nep-0029-deprecation_policy.html), which will make it harder to solve the environment for python 3.6 in the future.

maximlt:
This test `MPLRendererTest.test_render_holomap_not_embedded` is failing now on Ubuntu py38 and Mac py37. It seemed like a transient to @jlstevens but it would be worth having a look into it.

I'll submit a small commit soon to change a logging step, let's see whether this test still fails and if so how.

jlstevens:
To get the hotfix release out for bokeh 2.4 support I will go ahead and merge. The one failure looks like an unusual transient and I now think there are fewer transients to worry about.

There are a number of TODO items to follow up on in subsequent PRs:

- [ ] Fix the issue with Sankey (test skipped right now)
- [ ] Improve/fix the `render_holomap_not_embedded` tests
- [ ] Improve the testing infrastructure and remove the hacks if possible.

Thanks @maximlt for figuring out the workarounds!

PR #5084, opened 2021-09-16
Updating testing environment to use Bokeh 2.4

jlstevens:
The last batch of CI testing fixes ended up pinning panel to 0.12.1 which meant bokeh 2.4 was not being used for testing.


jlstevens:
Now an older panel/bokeh version is being used on 3.6 as bokeh seems to have dropped support for it. For 3.7 and 3.8, bokeh 2.4.0 is now used.

The `hover_tool_overlay_renderers` unit test is now also commented out with the flakey `render_holomap_not_embedded` tests. I will now merge and we can further improve the testing infrastructure in later PRs.

PR #5087, opened 2021-09-20
Attempt to deploy the docs

maximlt:
Currently `ibis-framework` cannot be installed on Ubuntu from conda-forge (https://github.com/conda-forge/ibis-framework-feedstock/issues/54), yet it can be installed from defaults.

The only dependency that is not available on defaults from the `-o doc` option is awscli. It's thus removed from `setup.py` and installed after `develop install` from conda-forge. Note this is already done for selenium and geckodriver. After this change, `develop_install` won't rely on conda-forge anymore.

This PR also adds the possibility to trigger the docs workflow manually from the UI (https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/).

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -9,7 +9,7 @@ on:
   workflow_dispatch:
     
 jobs:
-  test_suite:
+  build_docs:
     name: Documentation
     runs-on: 'ubuntu-latest'
     timeout-minutes: 120
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -57,6 +57,7 @@ jobs:
           conda activate test-environment
           conda list
           doit develop_install ${{ env.CHANS_DEV}} -o doc
+          conda install -c pyviz "jupyter_client<7"
           conda install -c conda-forge geckodriver selenium awscli
       - name: doit env_capture
         run: |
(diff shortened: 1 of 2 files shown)


jlstevens:
Happy to merge whenever you say it is ready (now I've seen it build and the output looks good).

maximlt:
@jlstevens I squashed some commits and forced push. And created a new 1.14.7 branch against which this PR is now on.  I've also removed the 1.14.6a4 dummy tag I used to trigger the docs build.

The manual trigger button doesn't appear as of yet. I have in mind that once this is merged it should appear.

jlstevens:
> And created a new 1.14.7 branch against which this PR is now on. 

I believe what should happen is this:

1. This PR should be against master.
2. Then this can be squash merged to master (I would double check recent fixes to CI are on master..I think so!)
3. The `branch_1.14.7` branch should be a PR opened against `branch_1.14`
4. The squash-merged commit from this PR should be cherry picked to it.

That way the doc fix here is on *both* master and the 1.14.7 hotfix. @philippjfr Do you agree?

philippjfr:
>That way the doc fix here is on both master and the 1.14.7 hotfix. @philippjfr Do you agree?

Yes, that's the correct approach, generally things should land on master first and backported if needed. Indeed I suggest we create a `backport` label for PRs. All PRs which are to land in a hotfix release should then be labeled with `backport` and have the release assigned.

Issue #5100, opened 2021-10-19
Get the current state of a DynamicMap and set bounds to cache_size


maximlt:
I would like to be able to get the current state of a DynamicMap, i.e. the values of the widgets that I see. `.last_key` cannot be used in that case, since it returns the last key of the `.keys()` dictionary, which is not the current state if that particular state has already been visited before (it was already added in the dictionary then).

One workaround here is to set `cache_size` to 1, then the dictionary only ever has one entry and `last_key` returns the current state, the only one being actually stored. I tried to set `cache_size` to 0 and got a weird error, this Integer parameter should get bounds, e.g. `(1, None)`.

PR #5105, opened 2021-10-31
Add bounds to the cache_size Parameter

maximlt:
Partially addresses https://github.com/holoviz/holoviews/issues/5100

The minimum size for the cache is 1.


jlstevens:
Thanks for the PR!

The fix is simple enough that I'll just go ahead and merge.

PR #5106, opened 2021-10-31
Add current_key property to DynamicMap

maximlt:
Partially addresses https://github.com/holoviz/holoviews/issues/5100

@jlstevens I've done some manual tests and saw that this works as I expect it. However `DynamicMap.__getitem__`  handles quite a number of different cases and I'm not sure I understand them all. It'd be great if you could let me know if it's the right implementation. If so, I'll document it and add some tests.

### DIFF
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -826,6 +826,8 @@ class DynamicMap(HoloMap):
 
         self.periodic = periodic(self)
 
+        self._current_key = None
+
     @property
     def opts(self):
         return Opts(self, mode='dynamicmap')
@@ -856,7 +858,7 @@ class DynamicMap(HoloMap):
     @property
     def current_key(self):
         """Returns the current key value."""
-        return getattr(self, '_current_key', None)
+        return self._current_key
 
     def _stream_parameters(self):
         return util.stream_parameters(
--- a/holoviews/core/spaces.py
+++ b/holoviews/core/spaces.py
@@ -853,6 +853,11 @@ class DynamicMap(HoloMap):
                 unbounded_dims.append(str(kdim))
         return unbounded_dims
 
+    @property
+    def current_key(self):
+        """Returns the current key value."""
+        return getattr(self, '_current_key', None)
+
     def _stream_parameters(self):
         return util.stream_parameters(
             self.streams, no_duplicates=not self.positional_stream_args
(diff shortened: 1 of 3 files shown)


jlstevens:
Other than deciding on a name, this looks like the correct approach to me!

holoviews/core/spaces.py
```diff
@@ -853,6 +853,11 @@ def unbounded(self):
                 unbounded_dims.append(str(kdim))
         return unbounded_dims
 
+    @property
+    def current_key(self):
+        """Returns the current key value."""
+        return getattr(self, '_current_key', None)
```
jlstevens, line 859: I think I would prefer you declare `self._current_key=None` in the constructor and just return `self._current_key`. That way you can prevent anyone from overwriting `current_key` without needing to use `getattr` here.

maximlt: Done!

maximlt:
@jlstevens as for the naming I chose `current_key` because of the existing property `last_key`. But I'd be fine with other options, maybe `state` ?

jlstevens:
How about `last_call`? 

That makes sense to me as the crucial aspect of a `DynamicMap` is that it uses a callable. What you think @philippjfr ?

philippjfr:
As a user I'd be confused by `last_call`, personally I also prefer `current_key` especially since it's symmetric with `last_key`.

jlstevens:
Ok, I still prefer `last_call` as the use of a callable is what is unique about DynamicMaps. Happy to let @jbednar be the tie-breaker...

philippjfr:
`last_call` just makes no sense to me, as a user I'd have to guess what it actually returns and if you made me guess I'd definitely think it would give me the value of the last call not the arguments used to generate it. I don't think we'll need @jbednar as a tie breaker though, since we've got @maximlt :)

maximlt:
I think I actually still prefer `current_key`, as Philipp said I would have a hard time guessing the output of `last_call`, while  it's more obvious with `current_key`.

jlstevens:
Ok, happy to let you override me but I'm not going to pretend that the difference between `last_key` and `current_key` is going to be clear to anybody. ;-p

jlstevens:
Thanks! 

Happy to merge when the tests finish.

jlstevens:
Tests are green. Merging.

PR #5109, opened 2021-11-05
Pin freetype on windows

maximlt:
Issue already observed on the CI of colorcet and datashader. The new release of freetype (2.11.0) on the `main` channel breaks matplotlib plotting somehow.


jlstevens:
Thanks @maximlt!

I'll merge now though we need to either open a PR to revert this change when it is fixed or set a constraint to explicitly exclude this version of freetyle (which is probably a good idea!)

PR #5131, opened 2021-11-13
Handle the empty string as a group name

### LINKED ISSUE #5130: hv.Image(group="") no longer works
Until HoloViews 1.13.5 it was legal to use a group name of "" for an Image, but starting with v1.14.0 that gives an error:

```python
import numpy as np, holoviews as hv
hv.extension()

ls = np.linspace(0, 10, 200)
xx, yy = np.meshgrid(ls, ls)
hv.Image(np.sin(xx)*np.cos(yy), group="").opts(cmap="greys")
```
image

In a brief investigation I couldn't see what code had changed, but apparently something has. In any case, searching for "capitalize" in https://github.com/holoviz/holoviews/blob/master/holoviews/core/util.py reveals various code paths that implicitly depend on the length of a string being nonzero, which doesn't seem like a safe assumption in general and is problematic in this case. 

Here, obviously group could simply be omitted, but this came up for existing code in colorcet that accepted a user argument for the group, and when the user wants a blank group they now have to set the group to the very non-intuitive value "Image", whereas the much more reasonable group="" was sufficient to suppress printing the group name before.

jbednar:
Fixes #5130 by making `hv.utils.capitalize` work properly for empty strings and by considering the empty string a tree attribute.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -346,7 +346,7 @@ def tree_attribute(identifier):
     These custom attributes start with a capitalized character when
     applicable (not applicable to underscore or certain unicode characters)
     """
-    if not identifier:
+    if identifier == '':
         return True
     if identifier[0].upper().isupper() is False and identifier[0] != '_':
         return True
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -346,6 +346,8 @@ def tree_attribute(identifier):
     These custom attributes start with a capitalized character when
     applicable (not applicable to underscore or certain unicode characters)
     """
+    if not identifier:
+        return True
     if identifier[0].upper().isupper() is False and identifier[0] != '_':
         return True
     else:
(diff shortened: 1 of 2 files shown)


holoviews/core/util.py
```diff
@@ -346,6 +346,8 @@ def tree_attribute(identifier):
     These custom attributes start with a capitalized character when
     applicable (not applicable to underscore or certain unicode characters)
     """
+    if not identifier:
```
jlstevens, line 349: This would match anything that is falsey even though it should only ever match the empty string. I would prefer to be explicit, i.e.:

`if identifier=='':`

jlstevens: I can update this when I add a unit test or two

jbednar: That's fine with me, though the rest of the function will then error if someone were to pass 0 or None; not sure if that's any better behavior, but maybe?

jlstevens: A 0 or None should be error conditions imho: those shouldn't be passed as identifiers in any situation.

jbednar: Right; I just mean that this isn't like a Parameterized object that's actually checking for such things; 0 and None will just error later at some point. Anyway, I'm fine with however you want to handle that!

jlstevens:
Looks good and after a minor change and a new unit test, all the tests are green.

Thanks @jbednar for the PR! Merging

PR #5133, opened 2021-11-15
Do not merge partially overlapping Stream callbacks

philippjfr:
In HoloViews any layout or grid of plots tries to merge stream callbacks that have the same Bokeh models as their source. The problem was that even stream callbacks that only partially overlapped in their handles were being merged. For example if you had two RangeXY callbacks attached to separate plots, if both were attached to the same Range1D object they would get merged. This is obviously incorrect since a) the fact that the x_range is the same does not mean the y_range will also match and b) even if x_range and y_range match they might be transposed.

This PR ensures that only totally compatible stream callbacks are merged.

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -454,12 +454,14 @@ class Callback(object):
         # callbacks on one handle to be merged
         handle_ids = [id(h) for h in cb_handles]
         cb_hash = tuple(handle_ids)+(id(type(self)),)
+        print(cb_hash in self._callbacks)
         if cb_hash in self._callbacks:
             # Merge callbacks if another callback has already been attached
             cb = self._callbacks[cb_hash]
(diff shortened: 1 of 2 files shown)


jlstevens:
Thanks for tracking this down!

Can you think of any cases where the previous, unsafe merge operation would appear to work correctly but would now no longer appear to work after this PR? I think this is the correct thing to do but I want to avoid issues where the old behavior appeared to be doing the right thing...

philippjfr:
>Can you think of any cases where the previous, unsafe merge operation would appear to work correctly but would now no longer appear to work after this PR?

Not really, at worst this new version will be slightly less efficient (not in any way that matters honestly) but at least it's always correct. We could probably have a version that allows merging stream callbacks where the sources are transposed, e.g. x_range/y_range are swapped but that would complicate the Stream triggering logic a whole bunch for not much gain.

philippjfr:
@jlstevens Added some tests, please merge once passing.

PR #5138, opened 2021-11-16
Fix datetime clipping on RangeXY stream

philippjfr:
Fixes https://github.com/holoviz/holoviews/issues/5078

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -668,23 +668,23 @@ class RangeXYCallback(Callback):
         data = {}
         if 'x0' in msg and 'x1' in msg:
             x0, x1 = msg['x0'], msg['x1']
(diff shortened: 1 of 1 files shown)


jlstevens:
So the code to convert from timestamps was already there but being executed after the comparisons were already attempted? :-S

Happy to merge but I'll let you consider whether you think it is worth adding a unit test first.

philippjfr:
>So the code to convert from timestamps was already there but being executed after the comparisons were already attempted? :-S

It was needed for other stuff, I didn't anticipate that x0 could be float but x1 be a datetime or vice versa.

philippjfr:
>Happy to merge but I'll let you consider whether you think it is worth adding a unit test first.

Honestly don't have the bandwidth to keep plugging away at HV.

jlstevens:
Ok then! Merging.

Issue #4943, opened 2021-05-22
hv.BoxWhisker/Violin doesn't work with xarray datasets


ahuang11:
```
import xarray as xr
import holoviews as hv

hv.extension("bokeh")

ds = xr.tutorial.open_dataset("air_temperature").isel(time=slice(0, 5))

hv.BoxWhisker(ds.isel(lon=0), ["lat"], ["air"])
```

This works:
```
import xarray as xr
import holoviews as hv

hv.extension("bokeh")

ds = xr.tutorial.open_dataset("air_temperature").isel(time=slice(0, 5))

hv.BoxWhisker(ds.isel(lon=0).to_dataframe(), ["lat"], ["air"])
```

philippjfr:
Yeah, this is one reason why hvPlot is usually a better entrypoint.

jlstevens:
@philippjfr Do you think this is a case of `wontfix` or maybe a doc fix to point to `hvplot` or maybe even something that should be addressed at the holoviews level?

PR #5140, opened 2021-11-16
Validate dimensionality of xarray interface data

philippjfr:
Add validation with helpful error message to catch issues like in https://github.com/holoviz/holoviews/issues/4943

### DIFF
--- a/holoviews/core/data/xarray.py
+++ b/holoviews/core/data/xarray.py
@@ -207,6 +207,20 @@ class XArrayInterface(GridInterface):
                             "for all defined kdims, %s coordinates not found."
                             % not_found, cls)
 
+        for vdim in vdims:
+            da = data[vdim.name]
+            undeclared = [
+                c for c in da.coords if c not in kdims and da[c].shape and
+                np.multiply.reduce(da[c].shape)>1]
+            if undeclared:
+                raise DataError(
+                    'The coordinates on the %r DataArray do not match the '
+                    'provided key dimensions (kdims). The following coords '
+                    'were left unspecified: %r. If you are requesting a '
+                    'lower dimensional view such as a histogram cast '
+                    'the xarray to a columnar format using the .to_dataframe '
+                    'or .to_dask_dataframe methods before providing it to '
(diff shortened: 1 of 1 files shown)


holoviews/core/data/xarray.py
```diff
@@ -207,6 +207,26 @@ def retrieve_unit_and_label(dim):
...
+            if any(len(da.coords[c].shape) > 1 for c in da.coords):
+                continue
+            undeclared = [
+                c for c in da.coords if c not in kdims and len(da[c].shape) == 1 and
+                da[c].shape[0] > 1]
+            if undeclared:
+                raise DataError(
+                    'The coordinates on the %r DataArray do not match the '
+                    'provided key dimensions (kdims). The following coords '
+                    'were left unspecified: %r. If you are requesting a '
```
jlstevens, line 225: I think this could be written more clearly. Here is how I think about things:

1. If a kdim is missing from the data, then that is an error.
2. If the data has a higher dimensionality (i.e. a superset of the kdims) I don't see why there would be an error: HoloViews should just work with the key dimensions requested (and I guess the remaining dimensions get considered as vdims).

Is case 1 happening here? In which case I am happy to suggest a new error message. If it is case 2. then i don't know why extra coordinate dimensions would cause any issue.

philippjfr: The problem is that gridded interface (like the `XArrayInterface`) are not column stores so if you are under-specifying the dimensionality of your data it cannot function correctly. It cannot orient the arrays consistently, it cannot add dimensions, it cannot report the correct shape etc. We could add support to GridInterfaces to somehow have consistent flattened representations and disallowing multi-dimensional return values but that's a **major** undertaking. So as long as that is true we have two options:

1. Automatically cast the data to a columnar representation, i.e. calling `.to_dataframe` or `.to_dask_dataframe` on xarray objects
2. Raise the error above

jlstevens: Thanks for explaining. In that case, I'll merge.

jlstevens:
Looks good! Merging.

PR #5149, opened 2021-11-27
Handle unsigned integer dtype in datashader aggregate operation

Hoxbro:
Fixes https://github.com/holoviz/hvplot/issues/670

<s>Still need to make tests.</s>

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -427,8 +427,10 @@ class aggregate(AggregationOperation):
             vals = df[d.name]
             if not is_custom and len(vals) and isinstance(vals.values[0], cftime_types):
                 vals = cftime_to_timestamp(vals, 'ns')
-            elif df[d.name].dtype.kind == 'M':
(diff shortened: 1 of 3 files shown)


jbednar:
Thanks, @Hoxbro ! Looks like a very reasonable fix to me, but I can't tell why the tests are failiing.

Hoxbro:
I was wondering if I should add a check that raises an error if an uint64 array has values higher than 2 ** 64 // 2 - 1. 
Alternatively could be not accepting uint64 as a dtype. What are your thoughts on this?

jbednar:
I'm happy to simply fail for uint64; without proper plumbing all the way through there's no way that could work reliably without checking all the data, and checking all the data seems slow, so better just to leave it to the user to convert to a signed array if they want to plot that (or if they really want it, to put uint support in Datashader!).

maximlt:
Looks good to me. @jbednar if you get the chance to review this I could include it in the next release.

jbednar (approved):

PR #5151, opened 2021-11-29
CI: Fix before release

maximlt:
Attempts to fix the CI on Python 3.6, and take advantage of that to do some clean up in the test workflow.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -79,18 +79,18 @@ jobs:
           echo "======"
           conda list
       # Temporarily pin panel and nbconvert
-      - name: Pin panel and nbconvert on py36
+      - name: Pin panel on py36
         if: matrix.python-version == '3.6'
         run: |
           eval "$(conda shell.bash hook)"
           conda activate test-environment
-          conda install -c pyviz "panel=0.12.1" "nbconvert>5"
+          conda install -c pyviz "panel=0.12.1"
       - name: Pin panel and nbconvert on py>36
         if: matrix.python-version != '3.6'
         run: |
           eval "$(conda shell.bash hook)"
           conda activate test-environment
-          conda install -c pyviz "panel=0.12.2" "nbconvert>5"
+          conda install -c pyviz "panel=0.12.2"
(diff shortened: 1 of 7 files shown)


maximlt:
This also fixes #5156. The broken tests could be fixed by wrapping the `.get_array()` calls into `numpy.asarray()` calls. I found two occurrences of `.get_array()` in holoviews' source code but don't expect them to break because of the changes  in Matplotlib, so I left them unchanged. The *examples* test suite needed no change at all. I also ran the test suite locally with both matplotlib 3.4.3 and 3.5.0.

.github/workflows/test.yml
```diff
@@ -79,24 +79,12 @@ jobs:
...
-          eval "$(conda shell.bash hook)"
-          conda activate test-environment
-          conda install -c pyviz "panel=0.12.2" "nbconvert>5"
-      - name: Patch with an older version of freetype since 2.11.0 is broken with Matplotlib on Windows
-        if: (matrix.os == 'windows-latest')
-        run: |
-          eval "$(conda shell.bash hook)"
-          conda activate test-environment
-          conda install ${{ env.CHANS_DEV}} "freetype=2.10.4"
+          conda install -c pyviz "panel=0.12"
```
jlstevens, line 87: Pinning panel *should* be fine for holoviews testing (and I am happy to merge this) but long-term it would be better if we could let `panel` float to the latest version.

setup.py
```diff
@@ -71,7 +71,7 @@
     'path.py',
     'matplotlib >=3',
     'nbsmoke >=0.2.0',
-    'nbconvert <6',
+    'nbconvert',
```
jlstevens, line 74: Great!

.github/workflows/test.yml
maximlt: Yes in this particular case I've had to pin panel there are one or more dev releases in the 0.13 series that can be installed on Python 3.6 but that are actually not working on Python 3.6.

jlstevens:
Given that it is unlikely that your comment broke the tests I will go ahead and merge. :-)

Thanks for the fix!

PR #5152, opened 2021-12-01
Delay projection comparison to optimize geoviews

maximlt:
`geoviews` depends on `cartopy` for all sorts of projections related operations. `cartopy` now (0.20) relies on `pyproj` for coordinate transformations, which is more robust and versatile. There's one downside though, at least for `geoviews`: now an equality comparison that is False, like `projection == 'not a projection'` is significantly slower (about 1,000,000x) than before. See this [comment](https://github.com/SciTools/cartopy/issues/1895#issuecomment-982982168) for a simple benchmark. This makes the creation of some geoviews plots [very slow](https://github.com/holoviz/geoviews/issues/529). The equality comparison is slower because a CRS/projection can be created in many different ways, so `PROJ`, the underlying library `pyproj` is using, will exhaust all possibilities until it finds out that the projections are not equal, and this apparently takes a lot of time.

`holoviews` had many occurrences of the comparison `self.projection == '3d'`, which in this PR I replace by `isinstance(self.projection, str) and self.projection == '3d'`. This solves the issue of slowing down `geoviews`. While I believe this shouldn't impact `holoviews` performance-wise, it may be worthwhile checking this is not the case. @philippjfr @jlstevens if that is relevant, could you point me to `holoviews` plots that make a large number of calls to `get_extents`?

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -693,7 +693,8 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         elif axis == 'y':
             axis_obj = plot.yaxis[0]
 
-        if self.geographic and self.projection == 'mercator':
+        if (self.geographic and isinstance(self.projection, str)
+            and self.projection == 'mercator'):
             dimension = 'lon' if axis == 'x' else 'lat'
             axis_props['ticker'] = MercatorTicker(dimension=dimension)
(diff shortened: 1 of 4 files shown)


jlstevens:
I very much doubt the extra `isinstance` check would have a big performance impact and this proposed changed definitely makes sense for geoviews. Seems like a very reasonably change to me!

maximlt:
Merging then given your positive comment @jlstevens !

Issue #5146, opened 2021-11-20
FutureWarning: casting datetime64[ns] values to int64 with .astype(...) is deprecated 


Hoxbro:
#### ALL software version info
```
numpy.__version__ = '1.20.3'
holoviews.__version__ = '1.14.6'
```

#### Description of expected behavior and the observed behavior
Casting datetime64[ns] values to int64 with .astype(...) is deprecated 

#### Complete, minimal, self-contained example code that reproduces the issue

``` python
import hvplot.pandas  # noqa
import pandas as pd

df = pd.DataFrame({
    "time": pd.date_range("2020", "2021" ,100),
    "data": 100 * [1],
})

df.hvplot.line("time", "data", datashade=True)

```

#### Stack traceback and/or browser JavaScript console output
```
C:\Users\shh\miniconda3\envs\clean_holoviz\lib\site-packages\holoviews\operation\datashader.py:428: FutureWarning: casting datetime64[ns] values to int64 with .astype(...) is deprecated and will raise in a future version. Use .view(...) instead.
  df[d.name] = vals.astype('int64')
```

jbednar:
Sounds like a straightforward PR to make from the error message, but then if so I'm not sure why it's deprecated and not simply done automatically any more, so maybe there is more to it than that.

philippjfr:
Yeah, I don't really understand the reasoning behind this change tbh. We rely on this a lot all over the place iirc.

Hoxbro:
The warning was added here: https://github.com/pandas-dev/pandas/pull/38544

jbednar:
Oh my; that link leads to a *lot* of discussion: https://github.com/pandas-dev/pandas/issues/22384#issuecomment-909660696

I'm not at all sure that the issue has been resolved one way or another, i.e. whether the deprecation will be implemented or reverted, so I'd be inclined to wait on this until such a change is released definitively in pandas.

Hoxbro:
The change will also not be as straightforward as the error message for Holoviews, which also can handle Dask DataFrames and Series, which doesn't have a view attribute. So a function to convert would look something like this:

``` python
def convert_to_int64(x):
    if pd.isna(x).any():
        raise ValueError("Cannot convert NaN/NaT values to integer")

    if dd and isinstance(x, (dd.DataFrame, dd.Series)):  # Dask does not have view
        return x.astype("int64")
    else:
        return x.view("int64")
```

jbednar:
[...]

PR #5157, opened 2021-12-06
Test suite maintenance

maximlt:
Fixes https://github.com/holoviz/holoviews/issues/5146

And cleans up the test suite a little, mostly by suppressing warnings generated by dependencies.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -473,7 +473,13 @@ class aggregate(AggregationOperation):
                         x_range=x_range, y_range=y_range)
 
         dfdata = PandasInterface.as_dframe(data)
-        agg = getattr(cvs, glyph)(dfdata, x.name, y.name, agg_fn)
+        # Suppress numpy warning emitted by datashader
+        with warnings.catch_warnings():
+            warnings.filterwarnings(
+                action='ignore', message='casting datetime64',
+                category=FutureWarning
+            )
+            agg = getattr(cvs, glyph)(dfdata, x.name, y.name, agg_fn)
         if 'x_axis' in agg.coords and 'y_axis' in agg.coords:
             agg = agg.rename({'x_axis': x, 'y_axis': y})
         if xtype == 'datetime':
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -473,7 +473,8 @@ class aggregate(AggregationOperation):
                         x_range=x_range, y_range=y_range)
 
         dfdata = PandasInterface.as_dframe(data)
-        # Suppress numpy warning emitted by datashader
+        # Suppress numpy warning emitted by dask:
+        # https://github.com/dask/dask/issues/8439
         with warnings.catch_warnings():
             warnings.filterwarnings(
                 action='ignore', message='casting datetime64',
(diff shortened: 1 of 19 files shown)


maximlt:
The test suite (`test_unit`) is down to a few warnings now 👍 

@jlstevens it's ready to be reviewed. Particularly this commit that adds a utility to suppress Numpy warnings when casting a datetim64 to int64: https://github.com/holoviz/holoviews/pull/5157/commits/fca56eac39d8763ffdf80186fb3e2cc91aacf930

holoviews/operation/datashader.py
```diff
@@ -473,7 +473,13 @@ def _process(self, element, key=None):
                         x_range=x_range, y_range=y_range)
 
         dfdata = PandasInterface.as_dframe(data)
-        agg = getattr(cvs, glyph)(dfdata, x.name, y.name, agg_fn)
+        # Suppress numpy warning emitted by datashader
```
jlstevens, line 476: Is there a corresponding issue/PR to fix this warning at the datashader level? Would be good to have a link to point to so this can be removed once fixed in datashader.

holoviews/operation/element.py
```diff
@@ -760,8 +761,17 @@ def _process(self, element, key=None):
                 edges = edges.astype('datetime64[ns]').astype('int64')
         else:
             hist_range = self.p.bin_range or element.range(selected_dim)
+            # Suppress a warning emitted by Numby when datetime or timedelta scalars
```
jlstevens, line 764: Numby sounds like a fun project! :-)

holoviews/tests/element/test_tiles.py
```diff
@@ -107,7 +107,7 @@ def test_pandas_series(self):
         self.check_array_type_preserved(
             pd.Series, pd.Series,
             lambda a, b: pd.testing.assert_series_equal(
-                a, b, check_exact=False, check_less_precise=True,
```
jlstevens, line 110: Any idea what `check_less_precise` was about?

jlstevens:
Other than a few minor comments above, this looks good!

Happy to merge when those points are addressed.

holoviews/tests/element/test_tiles.py
maximlt: Not precisely (haha sorry). It'd defined [here](https://pandas.pydata.org/docs/reference/api/pandas.testing.assert_frame_equal.html) and given its name I assumed that if the test passes without it, then it should be safe.

holoviews/operation/datashader.py
maximlt: I will open an issue in datashader and link it in this comment.

holoviews/operation/element.py
maximlt: Will fix in a commit! :)

holoviews/operation/datashader.py
maximlt: There was actually an issue in datashader which I've linked in the comment.

maximlt:
@jlstevens I addressed your comments.

jlstevens:
Great, thanks!

The comment fixes should have no bearing on the tests so I'll go ahead and merge.

PR #5161, opened 2021-12-07
CI: improvements to the docs build

maximlt:
Before releasing 1.14.7, let's see how that goes.


maximlt:
The docs build has been successful and the failing test is a transient. Merging!

PR #5179, opened 2022-01-10
Swap LooseVersion for packaging.Version subclass

philippjfr:
Not sure what to do about this. Also if we backport to the 1.14 branch we will probably have to conditionally import if `sys.version.major < 3`.

PR #5180, opened 2022-01-10
Using packaging.version.Version as LooseVersion

jlstevens:
An alternate approach to https://github.com/holoviz/holoviews/pull/5179: although the subclass approach in that PR would require fewer lines to be changed overall, this approach is more idiomatic in that no new classes are introduced.

Note that `packaging` is not expended to be a new dependency of HoloViews as HoloViews depends on Panel which in turn depends on Bokeh which pulls in `packaging`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,7 +16,7 @@ install_requires = [
     "pyviz_comms >=0.7.4",
     "panel >=0.9.5",
     "colorcet",
-    'packaging',
+    "packaging",
     "pandas >=0.20.0",
 ]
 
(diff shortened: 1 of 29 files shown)


setup.py
```diff
@@ -16,6 +16,7 @@
     "pyviz_comms >=0.7.4",
     "panel >=0.9.5",
     "colorcet",
+    'packaging',
```
philippjfr, line 19: Consistency :)

```suggestion
    "packaging",
```

philippjfr:
Thanks!

jlstevens:
@philippjfr Now this PR has been rebased, the tests are going green. Ready for final review/merge.

jlstevens:
The Python 3.8, macos-latest build failed but only because it hit the 60 minutes limit afaik.

PR #5184, opened 2022-01-19
Enhance resolve_dependent_value if value is list or tuples

Hoxbro:
Recursive calling `resolve_dependent_value` if value is a list or tuple. If this implementation crash and burn in the CI an alternative is the changes described here https://github.com/holoviz/hvplot/issues/697#issuecomment-1013895746.

This PR should make it able to run:
``` python
import holoviews as hv
import panel as pn
from bokeh.sampledata.autompg import autompg_clean as df

hv.extension("bokeh")

xaxis = pn.widgets.RadioButtonGroup(name="X axis", options=["mpg", "yr"])
ds = hv.Dataset(df)
transform = hv.util.transform.df_dim("*")
transform.groupby(["origin", xaxis]).apply(ds, keep_index=True, compute=False).mean()
```

Will add a unit test later this week.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -2228,3 +2228,30 @@ def cast_array_to_int64(array):
             category=FutureWarning,
         )
         return array.astype('int64')
+
+
+def flatten(line):
+    """
+    Flatten an arbitrarily nested sequence.
+
+    Inspired by: pd.core.common.flatten
+
+    Parameters
+    ----------
+    line : sequence
+        The sequence to flatten
+
+    Notes
+    -----
+    This only flattens list, tuple, and dict sequences.
+
+    Returns
+    -------
+    flattened : generator
+    """
+
+    for element in line:
+        if any(isinstance(element, tp) for tp in (list, tuple, dict)):
+            yield from flatten(element)
+        else:
+            yield element
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -1517,8 +1517,15 @@ def resolve_dependent_value(value):
        resolved.
     """
     range_widget = False
-    if isinstance(value, (tuple, list)):
+    if isinstance(value, list):
         value = [resolve_dependent_value(v) for v in value]
+    elif isinstance(value, tuple):
+        value = tuple(resolve_dependent_value(v) for v in value)
+    elif isinstance(value, dict):
+        value = {
+            resolve_dependent_value(k): resolve_dependent_value(v) for k, v in value.items()
+        }
+
     if 'panel' in sys.modules:
         from panel.widgets import RangeSlider, Widget
         range_widget = isinstance(value, RangeSlider)
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -1505,14 +1505,15 @@ def is_param_method(obj, has_deps=False):
 def resolve_dependent_value(value):
     """Resolves parameter dependencies on the supplied value
 
-    Resolves parameter values, Parameterized instance methods and
-    parameterized functions with dependencies on the supplied value.
+    Resolves parameter values, Parameterized instance methods,
+    parameterized functions with dependencies on the supplied value,
+    including such parameters embedded in a list or tuple.
 
     Args:
        value: A value which will be resolved
 
     Returns:
-       A new dictionary where any parameter dependencies have been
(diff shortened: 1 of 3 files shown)


philippjfr:
Nice, presumably we also need to ensure that it actually watches the widget/parameter though. Don't recall how offhand so if you don't see a way let me know and I'll dig into it.

Hoxbro:
I don't know if anything is needed in holoviews, but in hvplot the function `hvplot.interactive._find_widgets` should be updated with something similar to this.

holoviews/core/util.py
```diff
@@ -1516,6 +1516,8 @@ def resolve_dependent_value(value):
        resolved.
     """
     range_widget = False
+    if isinstance(value, (tuple, list)):
+        value = [resolve_dependent_value(v) for v in value]
```
jbednar, line 1520: Seems like this would silently convert tuples to lists, which I assume would cause problems. So maybe:

```suggestion
    if isinstance(value, list):
        value = [resolve_dependent_value(v) for v in value]
    elif isinstance(value, tuple):
        value = tuple([resolve_dependent_value(v) for v in value])
```

jbednar: Also, should there be a case to handle `dict` arguments, presumably recursively calling `resolve_dependent_value` on all the values of the dict? There are pandas methods that accept dictionaries...

Hoxbro:
This should now watch the widget/parameter. See:

https://user-images.githubusercontent.com/19758978/150646119-d321fd21-6d63-4a89-9ac9-bbd1c538a4e5.mp4



The code:
``` python
import holoviews as hv
import hvplot.pandas  # noqa
import panel as pn
from bokeh.sampledata.autompg import autompg_clean as df

hv.extension("bokeh")

idf = df.interactive()

cylinders = pn.widgets.IntSlider(start=4, end=8, step=2, name="Cylinders")
mfr = pn.widgets.ToggleGroup(
    options=["ford", "chevrolet", "plymouth", "honda", "sbaru", "volkswagen"],
    value=["ford", "chevrolet", "plymouth", "honda", "sbaru", "volkswagen"],
)
xaxis = pn.widgets.RadioButtonGroup(name="X axis", options=["mpg", "yr"])
yaxis = pn.widgets.RadioButtonGroup(name="Y axis", options=["hp", "weight"])

ipipeline = (
    idf[(idf.mfr.isin(mfr)) & (idf.cyl == cylinders)]
    .groupby(["origin", xaxis])[yaxis]
    .mean()
)

pn.Column(ipipeline.hvplot(by="origin"), sizing_mode="stretch_both").servable()
```

I have noted the linecolor change when changing the yaxis, I don't know if this is a feature or a bug.

Hoxbro:
I have added an unittest. When trying to run the test suite on my computer it stops at `holoviews/tests/plotting/bokeh/test_server.py` and when I ctrl+c it returns `TypeError: cannot create weak reference to 'NoneType' object`. This seems to be related to changes made here https://github.com/holoviz/panel/pull/2597.

holoviews/core/util.py
Hoxbro: This is now done. One of the failed test was because of this conversion.

jbednar:
This is really cool, @Hoxbro ! Does anyone know of other Pandas or Xarray API argument types that remain unsupported? So far it seems like scalars, lists, tuples, and dictionaries are handled. I don't expect there are any arguments that take sets? Any other containers worth adding?

holoviews/core/util.py
```diff
@@ -1505,17 +1505,27 @@ def is_param_method(obj, has_deps=False):
 def resolve_dependent_value(value):
     """Resolves parameter dependencies on the supplied value
 
-    Resolves parameter values, Parameterized instance methods and
-    parameterized functions with dependencies on the supplied value.
+    Resolves parameter values, Parameterized instance methods,
+    parameterized functions with dependencies on the supplied value,
+    including such parameters embedded in a list or tuple.
```
Hoxbro, line 1510: This needs to be updated again with dictionary but will wait with an update until a final decision have been made.

jbednar:
@philippjfr , I'm eager to get this in place; is there anything else needed before merging, other than the docs fix above?

philippjfr:
Nice, all looks good. Will merge.

PR #5185, opened 2022-01-19
Look at id if owner has no name for streams

Hoxbro:
This caused a problem when multiple widgets did not have names and would therefore only update the last widget with no name. The problem is seen in https://github.com/holoviz/hvplot/issues/697#issuecomment-1013895746.

### DIFF
--- a/holoviews/streams.py
+++ b/holoviews/streams.py
@@ -768,7 +768,7 @@ class Params(Stream):
         for p in self.parameters:
             pkey = (p.owner, p.name)
             pname = self._rename.get(pkey, p.name)
-            key = ' '.join([str(id(p)), pname])
+            key = ' '.join([str(id(p.owner)), pname])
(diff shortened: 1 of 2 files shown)


holoviews/streams.py
```diff
@@ -768,7 +768,7 @@ def hashkey(self):
         for p in self.parameters:
             pkey = (p.owner, p.name)
             pname = self._rename.get(pkey, p.name)
-            key = ' '.join([p.owner.name, pname])
+            key = ' '.join([p.owner.name or str(id(p)), pname])
```
Hoxbro, line 771: This does not fix when the names are the same. Should the key just be `str(id(p))` or something similar?

philippjfr: Honestly, yes, that's probably preferable.

philippjfr:
Thanks @Hoxbro!

PR #5188, opened 2022-01-20
refactoring code with truth value test 

zjzh:
refactoring code with truth value test which is more pythonic, concise, readable and efficient; how do think this change which has practical value？


jbednar:
Thanks for the suggestion! However, layer_number is explicitly a numeric value (obtained from `range()`), and so I think it's clearer to test only numeric values, to convey that more clearly.  Testing as a truth value I'd think is less clear in this case.

PR #5197, opened 2022-01-28
use Renderer's center property for HoloViews pane

stas-sl:
Hi, I noticed that there are 2 places where HoloViewsPane is instantiated: in one case its center property is set using Renderer's center property, in another it is set to always = True, not sure if this is intended behavior, but I assume this should be fixed, as currently you need to wrap you plot into panel manually if you don't want to center it, which is a bit inconvenient.


philippjfr:
Thanks @stas-sl!

stas-sl:
Oops 😳, now it should be better

stas-sl:
Hmm, I'm not sure now how the error could be caused by my changes, but if so, probably it is more complicated than I thought...

PR #5200, opened 2022-02-03
Fix xaxis/yaxis options with Matplotlib 

maximlt:
Fixes a bug introduced in https://github.com/holoviz/holoviews/pull/5152 and adds a test that would have caught it.


jlstevens:
Easily done and hard to track down after. Thanks for the fix!

jlstevens:
All tests are green. Merging!

PR #5201, opened 2022-02-03
Clean-up the CI and add Python 3.9 and maybe 3.10

maximlt:
- [x] Check the warnings


maximlt:
`ibis-framework` cannot be installed with `conda-forge`. They've recently published conda packages dedicated to each backend such as `ibis-sqlite` which solves this issue. However they're only available for ibis >= 2 while Holoviews code only supports ibis < 2. This hasn't be realized before since the channel used in CI was defaults where newer versions of ibis aren't yet available.

jlstevens:
Ready to merge?

PR #5203, opened 2022-02-07
Fix overlay of two-level categorical plots and HLine

philippjfr:
Previously `np.concatenate` would error in various scenarios when overlaying a two-level categorical plot with a plot that had no categories.


jlstevens:
All tests are green except for 3.6 and 3.8 on macos. As these are slow and 3.7 is green, I'll go ahead and merge.

PR #5208, opened 2022-02-15
Allow lower dimensional views on arbitrary dimensioned elements

philippjfr:
Recent enforcement of the dimensionality of gridded data when passed to an element has caused some issues. In particular it is technically valid to create elements which support arbitrary dimensionality to be given a higher dimensional array but only express indexes only for a subset of those dimensions. Reasons you might want to do this:

1) You want to aggregate the array along the non-declared dimensions
2) The element itself aggregates the data during the plotting step (e.g. stats plots like box-whisker and violin)

Fixes current errors in the doc build.

Fixes https://github.com/holoviz/hvplot/issues/698

### DIFF
--- a/holoviews/core/data/grid.py
+++ b/holoviews/core/data/grid.py
@@ -329,7 +329,10 @@ class GridInterface(DictInterface):
         dropped = [dims.index(d) for d in dims
                    if d not in dataset.kdims+virtual_coords]
         if dropped:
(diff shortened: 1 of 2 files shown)


jlstevens:
Not all the tests have passed yet but at least one test has passed per platform/python version. The test suite will be run again fully for the 1.14.8 hotfix release anyway and on master. Merging!

PR #5209, opened 2022-02-15
Fix escaping of HTML on Div element

philippjfr:
The `HTML` model in Panel has required the HTML text to be escaped for **a long time**.


jlstevens:
All platforms and python versions have at least one test passing. Merging.

PR #5215, opened 2022-02-21
Added missing packaging.version

Hoxbro:
I was wondering if the rename of Version to LooseVersion should be removed, as I think it confuses more than it helps.

### DIFF
--- a/holoviews/plotting/bokeh/stats.py
+++ b/holoviews/plotting/bokeh/stats.py
@@ -309,7 +309,7 @@ class BoxWhiskerPlot(CompositeElementPlot, ColorbarPlot, LegendPlot):
             factors = list(unique_iterator(factors))
 
         if self.show_legend:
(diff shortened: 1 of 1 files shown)


holoviews/plotting/bokeh/stats.py
```diff
@@ -309,7 +309,7 @@ def get_data(self, element, ranges, style):
             factors = list(unique_iterator(factors))
 
         if self.show_legend:
-            legend_prop = 'legend_field' if bokeh_version >= '1.3.5' else 'legend'
+            legend_prop = 'legend_field' if bokeh_version >= LooseVersion('1.3.5') else 'legend'
```
jlstevens, line 312: Well caught! I thought I had replaced all of these but seems I missed this one :-(

jlstevens:
Some tests timed out but the rest passed. Merging and thanks for the PR!

PR #5216, opened 2022-02-21
Don't load any html in the notebook when no logo is required

maximlt:
This avoids calling `publish_display_data` in a notebook when calling `hv.extension()` with `logo=False` because unfortunately in JupyterLab each call to `publish_display_data` ends up adding an empty line.

The call to `publish_display_data` loads HTML from the `load_notebook.html` template which only contains the logos base64 encoded. So it seems safe to avoid loading this HTML when no logo is required.

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -195,7 +195,7 @@ class notebook_extension(extension):
 
         # Create a message for the logo (if shown)
         if not same_cell_execution and p.logo:
-            self.load_hvjs(logo=p.logo,
+            self.load_logo(logo=p.logo,
                            bokeh_logo=  p.logo and ('bokeh' in resources),
                            mpl_logo=    p.logo and (('matplotlib' in resources)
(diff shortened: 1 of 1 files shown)


maximlt:
Two checks were cancelled because they timed out after 1hr. Looking forward to the new libmamba solver to speed things up, and maybe the caching Mridul has started to implement in hvPlot.

holoviews/ipython/__init__.py
```diff
@@ -194,7 +194,7 @@ def __call__(self, *args, **params):
                                                   "hv-extension-comm")
 
         # Create a message for the logo (if shown)
-        if not same_cell_execution:
+        if not same_cell_execution and p.logo:
             self.load_hvjs(logo=p.logo,
```
jlstevens, line 198: Given that panel handles the JS now, I guess this is now misnamed (an artifact of history). Could you double check it is only used to load HTML and if so, rename it to `self.load_extension_html` (or similar)?

Issue #4837, opened 2021-02-12
Add streamlines plot


neighthan:
[matplotlib] and [plotly] both support streamline plots (a different way of visualizing vector fields than the quiver-style plots holoviews already supports). As far as I could tell, holoviews doesn't support such a plot (though googling mostly gave me results about using data streams in holoviews, so it's possible I missed something). I think it'd be nice to add this plot type. Alternatively, if holoviews doesn't want to have too many built-in plot types, is there an easy way from the user side to add a new plot type that already exists in one of the backends? I'm just making streamplots in matplotlib right now, but then I can't, e.g., combine them with my other holoviews plots in overlays.

(bokeh doesn't seem to have built-in support for streamlines, but there's an [official example] in their gallery showing how this can be done).

[matplotlib]: https://matplotlib.org/stable/gallery/images_contours_and_fields/plot_streamplot.html
[plotly]: https://plotly.com/python/streamline-plots/
[official example]: https://docs.bokeh.org/en/latest/docs/gallery/streamline.html

jlstevens:
I think this would be a pretty cool type of element to support with pretty general applicability. Tagging this as a feature request. What do you think @philippjfr?

philippjfr:
+1, should be easy enough to implement.

astroboylrx:
Any updates on this? Thx in advance.

jbednar:
I tagged it Wishlist to indicate that it's a reasonable but unfunded request that no one has volunteered to implement (though Philipp did sound like he was considering it. :-)

PR #5242, opened 2022-03-23
Streamlines element

poplarShift:
I'm starting to need this so had a go at #4837.

It might be good to eventually include points at either the origin or destination of a given streamline. It should be easy enough but I've become a bit rusty in holoviews...

PS: I'm not at the university anymore so got to be counting hours now, but still interested in getting this in.

### DIFF
--- a/holoviews/element/path.py
+++ b/holoviews/element/path.py
@@ -16,6 +16,13 @@ from .geom import Geometry
 from .selection import SelectionPolyExpr
 
 
+class Streamlines(SelectionPolyExpr, Geometry):
+
+    vdims = param.List(default=[Dimension('u'), Dimension('v')])
+
+    group = param.String(default='Streamlines', constant=True)
+
+
 class Path(SelectionPolyExpr, Geometry):
     """
     The Path element represents one or more of path geometries with
--- a/holoviews/plotting/bokeh/__init__.py
+++ b/holoviews/plotting/bokeh/__init__.py
@@ -7,7 +7,7 @@ from param import concrete_descendents
 
 from ...core import (Store, Overlay, NdOverlay, Layout, AdjointLayout,
                      GridSpace, GridMatrix, NdLayout, config)
-from ...element import (Curve, Points, Scatter, Image, Raster, Path,
+from ...element import (Curve, Points, Scatter, Image, Raster, Path, Streamlines,
                         RGB, Histogram, Spread, HeatMap, Contours, Bars,
                         Box, Bounds, Ellipse, Polygons, BoxWhisker, Arrow,
                         ErrorBars, Text, HLine, VLine, HSpan, VSpan, Spline, Spikes,
@@ -38,6 +38,7 @@ from .heatmap import HeatMapPlot, RadialHeatMapPlot
 from .hex_tiles import HexTilesPlot
 from .links import LinkCallback # noqa (API import)
 from .path import PathPlot, PolygonPlot, ContourPlot
+from .streamlines import StreamlinePlot
 from .plot import GridPlot, LayoutPlot, AdjointLayoutPlot
 from .raster import RasterPlot, RGBPlot, HSVPlot, QuadMeshPlot
 from .renderer import BokehRenderer
@@ -92,6 +93,7 @@ associations = {Overlay: OverlayPlot,
                 Bounds:   PathPlot,
                 Ellipse:  PathPlot,
                 Polygons: PolygonPlot,
+                Streamlines: StreamlinePlot,
 
                 # Geometry
                 Rectangles:    RectanglesPlot,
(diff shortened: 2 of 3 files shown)


jbednar:
Cool! For most elements we have a default border, but I believe that is set to zero for RGB, HSV, Raster, VectorField, and Image elements. I'd propose that the border be zero for Streamlines and Contour elements as well.

philippjfr:
That's easily configured by overriding the `padding` parameter on the plotting class with a value of 0.

poplarShift:
Thanks for the heads up, will look at that next time I'm in there!

maximlt:
@poplarShift this is a very nice feature that we would like to put in the next major release, do you need any help to complete your PR?

poplarShift:
Hi @maximlt 
Some open questions would be:
1. Is it OK to just use the bokeh streamline integration code? Its core is RK4 which is pretty standard AFAIK.
2. Is there any sensible way of putting arrows on the lines? I don't think so but I may be missing some (new?) Bokeh features
3. I've set the padding argument to default to 0, but padding is the same as before. If somebody could make a guess why not, that would be great!
4. I would ideally like to be able to supply unstructured data (not regularly gridded velocities) - though I can't promise when I get around to that. So maybe that's best handled in a future PR?

maximlt:
Hey thanks for the update! I'll ping @jlstevens who is more likely to be able to answer your questions.

jlstevens:
Thanks for the PR, this looks great!

> 1. Is it OK to just use the bokeh streamline integration code? Its core is RK4 which is pretty standard AFAIK.

If you are worried about licensing issues, I'm sure it is ok. *Ideally* this code would be public API importable from Bokeh, but in this case, inlining the code is also fine.

> 2. Is there any sensible way of putting arrows on the lines? I don't think so but I may be missing some (new?) Bokeh features

Not that I am aware of, but I'll go ahead and ping @mattpap for feedback on these two bokeh related questions.

Also, as long as arrows support can be added as a plot option later, I don't think those are necessary for the first release of this element.

> 3. I've set the padding argument to default to 0, but padding is the same as before. If somebody could make a guess why not, that would be great!

I've got no immediate insights into this padding issue, I'm afraid! Happy to help track this down once we are happy with everything else.

> I would ideally like to be able to supply unstructured data (not regularly gridded velocities) - though I can't promise when I get around to that. So maybe that's best handled in a future PR?

As long as the constructor of the element doesn't change (i.e. the element semantics are clear), I agree this can be in a future PR. Perhaps you could just give an example of what you think this data would look like and then we can postpone supporting that format till later?

philippjfr:
> I've set the padding argument to default to 0, but padding is the same as before. If somebody could make a guess why not, that would be great!

Would have to know where and how you set it. I would recommend setting it as a parameter on the plotting class, i.e.:

```
    padding = param.ClassSelector(default=0, class_=(int, float, tuple))
```

poplarShift:
> > I've set the padding argument to default to 0, but padding is the same as before. If somebody could make a guess why not, that would be great!
> 
> Would have to know where and how you set it. I would recommend setting it as a parameter on the plotting class, i.e.:
> 
> ```
>     padding = param.ClassSelector(default=0, class_=(int, float, tuple))
> ```

Thanks for weighing in, that's exactly what I did. Will have to look at that again then...

poplarShift:
> Thanks for the PR, this looks great!
> 
> > 1. Is it OK to just use the bokeh streamline integration code? Its core is RK4 which is pretty standard AFAIK.
> 
> If you are worried about licensing issues, I'm sure it is ok. _Ideally_ this code would be public API importable from Bokeh, but in this case, inlining the code is also fine.

It's from the bokeh documentation https://docs.bokeh.org/en/latest/docs/gallery/streamline.html, so not directly from the codebase.

> > I would ideally like to be able to supply unstructured data (not regularly gridded velocities) - though I can't promise when I get around to that. So maybe that's best handled in a future PR?
> 
> As long as the constructor of the element doesn't change (i.e. the element semantics are clear), I agree this can be in a future PR. Perhaps you could just give an example of what you think this data would look like and then we can postpone supporting that format till later?

The current gridded input format is
```
( (N, ), (M, ), (N, M), (N, M) )
```
so yes, that would have to change if we were instead to pass unstructured (columnar) data. We could reconstruct the regular grid by doing some binning/averaging preprocessing internally to make the current streamline integration algorithm work. Integrating directly on unstructured data is a bit more involved but I haven't actually done a lot of research on other implementations.

mattpap:
> Is there any sensible way of putting arrows on the lines? I don't think so but I may be missing some (new?) Bokeh features

Arrow heads can be added to a subset of glyphs via glyph decorations. Currently only `Arc` and `Segment` glyphs support this. See an example here https://github.com/bokeh/bokeh/blob/branch-3.0/examples/plotting/file/trefoil.py#L48-L52. I expect to implement broader support in bokeh 3.1. Note also that the existing API and features are experimental and subject to change.

PR #5247, opened 2022-03-28
Support buffer of unspecified length

pepijndevos:
Currently a buffer requires a specified length, but this is not ideal in situations where you want to stream a large but finite dataset of unknown size. You could set a large size, but how large is large enough, or too large? This PR allows setting `size=None` for a buffer that'll just grow until you clear it or run out of memory.


jbednar (approved):
Looks useful to me! Presumably should be documented somewhere.

PR #5249, opened 2022-03-28
Loading JS once in the given cell

jlstevens:
PR using the information now exposed in https://github.com/holoviz/pyviz_comms/pull/87 to avoid repeatedly loading JS in the same cell (e.g. if multiple extensions are in the first cell of a notebook).

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -191,12 +191,14 @@ class notebook_extension(extension):
             Renderer.comm_manager.get_client_comm(notebook_extension._process_comm_msg,
                                                   "hv-extension-comm")
 
+        same_cell_execution = getattr(self, '_repeat_execution_in_cell', False)
         # Create a message for the logo (if shown)
-        self.load_hvjs(logo=p.logo,
-                       bokeh_logo=  p.logo and ('bokeh' in resources),
-                       mpl_logo=    p.logo and (('matplotlib' in resources)
-                                                or resources==['holoviews']),
-                       plotly_logo= p.logo and ('plotly' in resources))
+        if not same_cell_execution:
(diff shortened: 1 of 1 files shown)


philippjfr:
Does not seem sufficient, you'll also have to handle these lines:

        for r in [r for r in resources if r != 'holoviews']:
            Store.renderers[r].load_nb(inline=p.inline)
        Renderer.load_nb(inline=p.inline)

and we probably also need to think about adding warning checking that if `same_cell_execution` is true and `'plotly' not in panel.extension._loaded_extensions` that the user has to make sure they run `hv.extension('plotly')` or `pn.extension('plotly')` first.

holoviews/ipython/__init__.py
```diff
@@ -183,20 +183,23 @@ def __call__(self, *args, **params):
         if hasattr(config, 'comms') and comms:
             config.comms = comms
 
-        for r in [r for r in resources if r != 'holoviews']:
-            Store.renderers[r].load_nb(inline=p.inline)
-        Renderer.load_nb(inline=p.inline)
+        if not same_cell_execution:
```
philippjfr, line 186: Variable is not defined yet, you define it below.

jlstevens: Oops! Well spotted.

jlstevens: Fixed in [b946bbf](https://github.com/holoviz/holoviews/pull/5249/commits/b946bbfa895fc49678725bad5f95e0d20a918c6d)

maximlt:
I've merged master and the tests all passed 🎉 Ready to be merged then I think @jlstevens

jlstevens:
Thanks for fixing the tests! @philippjfr please merge if you think this PR is ready.

Issue #5261, opened 2022-04-06
Unable to control minor ticks on HeatMap


jbednar:
According to https://holoviews.org/user_guide/Customizing_Plots.html#font-sizes, there are plot options to control `minor_xticks` and `minor_ticks`, but I can't find anything to control the minor ticking behavior on a HeatMap:

```python
import numpy as np, holoviews as hv
hv.extension('bokeh', width=100)

data = [(i, chr(97+j),  i*j) for i in range(5) for j in range(5) if i!=j]
hm = hv.HeatMap(data).sort().opts(width=200, height=200)
(hm.options(xticks=0) + hm.options(xticks=1) + hm.options(xticks=2) + hm.options(xticks=4) +  hm.options(xticks=5)).cols(5)
```

<img width="1148" alt="image" src="https://user-images.githubusercontent.com/1695496/162039849-f0d50372-8f15-47e5-879e-a9ee3a199e80.png">

I.e., the only option available appears to be `xticks`, but apart from the default (with lots of minor ticks) and 0 (with no ticks at all), there don't seem to be any useful values for a heatmap. A value of 1 seems broken, in that it puts ticks but only labels one of them, and none of the other values achieve having one tick per cell in this plot, which is what I was trying to achieve, i.e. the default for the Matplotlib backend:

![image](https://user-images.githubusercontent.com/1695496/162040346-60ece64a-7cb6-49b1-ab5f-02e4e70ffc01.png)

marcbernot:
In the exemple, the x-axis is a `LinearAxis`. This seems problematic ; shouldn't both axes be `CategoricalAxis` (as stated in the documentation you link to)? And then, this would be logical that there is no notion of minor_ticks.
```python
import holoviews as hv
import pandas as pd

hv.extension('bokeh', width=100)
data = [(i, chr(97+j),  i*j) for i in range(5) for j in range(5) if i!=j]
hm = hv.HeatMap(data)

df = pd.DataFrame(data,columns = ['x','y','z'])
hm = hv.HeatMap(df)

brenderer = hv.renderer('bokeh')
fig = brenderer.get_plot_state(hm)
fig.axis
```
result is `[LinearAxis(id='1673', ...), CategoricalAxis(id='1677', ...)]`.

The figure you look for can be achieved by casting `df['x'] = df['x'].astype(str)`. However I think, this should work with no explicit casting (or such a casting should be taken care of by holoviews).

[...]

Issue #5108, opened 2021-11-05
Suggesting that by default a Heatmap has no annotations with the Matplotlib backend


maximlt:
Annotations don't look good in a Heatmap with a large amount of data and those that are rendered in the high values range are barely visible with the default color map. Heatmaps generated with the Bokeh backend don't have annotations by default.

See this example taken from the reference gallery:
![image](https://user-images.githubusercontent.com/35924738/140487561-3cdb38ea-8694-4328-ab9f-bfddaf1dd010.png)

jbednar:
Fine by me!

jlstevens:
I agree this should be trivial for someone to change but note that this affects backwards compatibility. That means this is suitable for 2.0 but should *not* go into a hotfix release.

prj99:
@maximlt may I work with this issue

maximlt:
Yes please @prj99! :)

PR #5262, opened 2022-04-06
Heatmap Fixes

jbednar:
The HeatMap and Continuous_Coordinates notebooks were written long ago, and various changes to the docs over the years have caused them to have issues that we apparently never noticed:

- Previously, HeatMap had a default colormap of `RdYlBu_r`, while the Raster types had a default of `fire`. Both defaults were updated to `kbc_r`, but Continuous_Coordinates was still overriding `RdYlBu_r` to match the Raster `fire`, and now the result is _not_ matching the Raster `kbc_r`. Removed this override now that the defaults match (as they should).
- Matplotlib's HeatMap shows labels overlaid by default (which itself is an issue; see https://github.com/holoviz/holoviews/issues/5108), while Bokeh does not. The Bokeh approach is a better choice anyway, for composability, but switching to the Bokeh version meant that those labels disappeared. Added an example of adding labels explicitly to HeatMap.ipynb, and added labels to Continuous_Coordinates since the text was assuming the heat map had explicit labels. Note that overlaying labels looks pretty bad unless `padding=0`, so I added that to the example.
- Matplotlib doesn't show minor ticks by default on a HeatMap, which is appropriate for coarse grids. When adding the Bokeh version of this Element, it looks like there was an attempt to turn off minor ticks with `.opts(xticks=None)`, but that doesn't appear to have any effect. Moreover, there doesn't seem to be a reliable way to turn off those ticks at all, for which I opened a separate ticket https://github.com/holoviz/holoviews/issues/5261 . For now, just accepted having minor ticks shown.

Before this PR:
image

After this PR:
image

### DIFF
--- a/examples/reference/elements/bokeh/HeatMap.ipynb
+++ b/examples/reference/elements/bokeh/HeatMap.ipynb
@@ -47,14 +47,30 @@
    "source": [
     "data = [(i, chr(97+j),  i*j) for i in range(5) for j in range(5) if i!=j]\n",
     "hm = hv.HeatMap(data).sort()\n",
-    "hm.opts(xticks=None)"
+    "hm"
    ]
   },
   {
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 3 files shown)


jlstevens:
Looks good, thanks!

PR #5264, opened 2022-04-07
Add support for setting line_width on datashader line aggregation

philippjfr:
Adds support for setting `line_width` on datashader aggregation operations if datashader version >= 0.14.

Also adds support for `rescale_discrete_levels` option on `shade` operation.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1246,6 +1246,13 @@ class shade(LinkableOperation):
         undersaturation, i.e. poorly visible low-value datapoints, at
         the expense of the overall dynamic range..""")
 
+    rescale_discrete_levels = param.Boolean(default=False, doc="""
+        If ``cnorm='eq_hist`` and there are only a few discrete values,
+        then ``rescale_discrete_levels=True`` decreases the lower
+        limit of the autoranged span so that the values are rendering
+        towards the (more visible) top of the ``cmap`` range, thus
+        avoiding washout of the lower values.  Has no effect if
+        ``cnorm!=`eq_hist``.""")
 
     @classmethod
     def concatenate(cls, overlay):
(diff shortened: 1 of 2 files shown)


holoviews/operation/datashader.py
```diff
@@ -336,7 +336,24 @@ def _get_agg_params(self, element, x, y, agg_fn, bounds):
...
+        lines are drawn using a simple algorithm with a blocky
+        single-pixel width based on whether the line passes through
+        each pixel or does not. If greater than one, lines are drawn
+        with the specified width using a slower and more complex
+        antialiasing algorithm with fractional values along each edge,
+        so that lines have a more uniform visual appearance across all
+        angles. Line widths between 0 and 1 effectively use a
+        line_width of 1 pixel but with a proportionate reduction in
+        the strength of each pixel, approximating the visual
+        appearance of a subpixel line width.""")
```
jlstevens, line 352: Looks good though might be worth mentioning a `line_width` of 1 as a special value: the antialiased version of the old default behavior.

jlstevens: I also see arguments for the default being `None` (old behavior) or setting the default to `1` (nicer looking lines, need to set it back to None to restore the old behavior exactly).

jlstevens: Actually, given that this feature isn't going to be in the datashader version people have right now, it is probably best to stick with `None`. This docstring should mention the datashader version that antialiasing/line_width gets used.

Essentially, people may expect to see antialiasing because they set `line_width` but they have an old datashader version. Now I am leaning towards setting it to 1 after the new datashader release has been out for a while.

jlstevens:
Tests are passing so feel free to merge if you think this is ready.

holoviews/operation/datashader.py
jbednar: I'd guess that most people will get an updated Datashader at the same time they will get an updated HoloViews, so though it's good to have error checking I think we should assume the support is available in Datashader when setting defaults.  I don't think `line_width=1` has any special significance we need to mention.

jlstevens: That isn't necessarily true unless we want to pin to greater than the new datashader version in the new release (which I am not convinced we should do, especially for a hotfix release).

jbednar: It's not *necessarily* true, but isn't it roughly true? I.e., when someone creates a new environment with HoloViews or hvPlot, seems like they'll get a new Datashader too, since that will also be available and compatible. It's true they could be updating only and precisely HoloViews, but that takes extra work, and seems less likely.

PR #5268, opened 2022-04-08
Set rescale_discrete_levels=True by default

jbednar:
The `eq_hist` colormapper used with Datashader is designed for large datasets with a wide distribution of values, but as documented in https://github.com/holoviz/datashader/issues/357 the behavior when there are only a few distinct value levels has always been confusing and misleading. Datashader 0.14 adds an option `rescale_discrete_levels=True`, now supported in HoloViews, allowing users to avoid this issue. However, this option is False by default, and most users are never going to be aware that the option exists. This PR enables the option by default, which has good and bad consequences:

- Good: Users will now see better colormapping behavior, with points visible at all zoom levels and no longer jumping between the lowest and highest colormap values at a tiny change in zoom level, without having to learn about this option or modify any of their code. This is arguably a long-overdue bug fix.
- Bad: Results will change for plots rendered with `eq_hist` that have only a small number of discrete levels. That's by design, but it is strictly a compatibility change, which is odd to introduce in a patch release.
- Bad: Results will differ between `rasterize` and `datashade`, because Bokeh does not yet support `rescale_discrete_levels`. So there is some argument for waiting until Bokeh does support it to make this the default.

Basically, if we consider this a bugfix, we should introduce it now; patch releases are for bugfixes. If we consider it a new feature, we should wait until the next major or at least minor release to enable it by default.

I'm ok with either option, as long as the options have been duly considered and decided between.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1265,8 +1265,8 @@ class shade(LinkableOperation):
         the lower limit of the autoranged span so that the values are
         rendering towards the (more visible) top of the ``cmap`` range, 
         thus avoiding washout of the lower values.  Has no effect if
-        ``cnorm!=`eq_hist``. Set to False to match historical unscaled
-        behavior, prior to HoloViews 1.14.4.""")
+        ``cnorm!=`eq_hist``. Set this value to False if you need to
+        match historical unscaled behavior, prior to HoloViews 1.14.4.""")
 
     @classmethod
     def concatenate(cls, overlay):
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1246,13 +1246,14 @@ class shade(LinkableOperation):
         undersaturation, i.e. poorly visible low-value datapoints, at
         the expense of the overall dynamic range..""")
 
-    rescale_discrete_levels = param.Boolean(default=False, doc="""
+    rescale_discrete_levels = param.Boolean(default=True, doc="""
         If ``cnorm='eq_hist`` and there are only a few discrete values,
-        then ``rescale_discrete_levels=True`` decreases the lower
-        limit of the autoranged span so that the values are rendering
-        towards the (more visible) top of the ``cmap`` range, thus
-        avoiding washout of the lower values.  Has no effect if
-        ``cnorm!=`eq_hist``.""")
+        then ``rescale_discrete_levels=True`` (the default) decreases
+        the lower limit of the autoranged span so that the values are
+        rendering towards the (more visible) top of the ``cmap`` range, 
+        thus avoiding washout of the lower values.  Has no effect if
+        ``cnorm!=`eq_hist``. Set to False to match historical unscaled
+        behavior, prior to HoloViews 1.14.4.""")
 
     @classmethod
     def concatenate(cls, overlay):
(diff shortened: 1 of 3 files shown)


maximlt:
@jbednar should this also be merged before 1.15.0?

jbednar:
Yes

jbednar:
Yes

holoviews/operation/datashader.py
```diff
@@ -1259,13 +1259,14 @@ class shade(LinkableOperation):
...
-        limit of the autoranged span so that the values are rendering
-        towards the (more visible) top of the ``cmap`` range, thus
-        avoiding washout of the lower values.  Has no effect if
-        ``cnorm!=`eq_hist``.""")
+        then ``rescale_discrete_levels=True`` (the default) decreases
+        the lower limit of the autoranged span so that the values are
+        rendering towards the (more visible) top of the ``cmap`` range, 
+        thus avoiding washout of the lower values.  Has no effect if
+        ``cnorm!=`eq_hist``. Set to False to match historical unscaled
+        behavior, prior to HoloViews 1.14.4.""")
```
maximlt, line 1269: @jbednar Then this last sentence should be removed?

jbednar: I don't think so, but maybe it needs clarifying:
```suggestion
        ``cnorm!=`eq_hist``. Set this value to False if you need to
        match historical unscaled behavior, prior to HoloViews 1.14.4.""")
```

jbednar:
Bokeh 2.4.3 EqHistColorMapper now supports `rescale_discrete_levels`, so I think this PR should now be merged. Do we need to do anything so that the same default applies in the Bokeh case?

philippjfr:
Already seems to be true in the bokeh case, but really we need to implement it for matplotlib.

holoviews/plotting/mpl/element.py
```diff
@@ -719,6 +719,14 @@ class ColorbarPlot(ElementPlot):
         If not 'neither', make pointed end(s) for out-of- range values."""
     )
 
+    rescale_discrete_levels = param.Boolean(default=True, doc="""
+        If ``cnorm='eq_hist`` and there are only a few discrete values,
+        then ``rescale_discrete_levels=True`` decreases the lower
+        limit of the autoranged span so that the values are rendering
+        towards the (more visible) top of the palette, thus
+        avoiding washout of the lower values.  Has no effect if
+        ``cnorm!=`eq_hist``.""")
```
jbednar, line 728:
```suggestion
        ``cnorm!=`eq_hist``. Set this value to False if you need to
        match historical unscaled behavior, prior to HoloViews 1.14.4.""")
```

holoviews/plotting/mpl/util.py
```diff
@@ -404,52 +404,55 @@ def convert(cls, value, unit, axis):
...
+            c = 1.5 - 2*m
+            multiple = m*discrete_levels + c
+            if (multiple > 1):
+                lower_span = 1 - multiple
+
+        cdf_bins = np.linspace(lower_span, 1, n+1)
+        binning = np.interp(cdf_bins, cdf, eq_bin_centers)
+        if not self._rescale:
+            binning[0] = low
+        binning[-1] = high
```
jbednar, line 455: This looks good to me, as it brings in the simplifications and improved colorbar calculations now in Bokeh, but I can't follow every step in the logic to detect if there are any holes. Maybe some tests needed?

jbednar (commented):
The MPL changes look good, thanks!

holoviews/plotting/mpl/util.py
philippjfr: Will follow up when I get a chance but honestly can't justify spending much more time on this.

jbednar: Would definitely be good to have some serious Matplotlib users and potential contributors involved in HoloViews and hvPlot, if anyone is reading this! :-)

PR #5278, opened 2022-04-21
Added FAQ about sharing only a single axis

jbednar:
Users are very often confused about axiswise normalization and shared_axes. I added one FAQ, explaining how to get only a single axis to be shared, but I think there are a few other FAQs that need addressing:

- Q: When do I use `shared_axes` and when do I use `axiswise=True`?
- Q: How do I make HoloViews plots responsive in a Panel app? Should I use `sizing_mode='both'` or `responsive=True` or both? Do I put them on the individual plots, the layout, the Panel container, or...? Help!

If someone knows the answers, please add them here!

### DIFF
--- a/doc/FAQ.rst
+++ b/doc/FAQ.rst
@@ -94,8 +94,20 @@ normalized independently by changing ``framewise`` to True:
 
     your_holomap.opts(framewise=True)
 
+**Q: How do I make only a single axis be shared across a layout?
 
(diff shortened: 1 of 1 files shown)


jlstevens:
Thanks! 

This is indeed a genuinely frequently asked question and confusing enough that some explanation here is appropriate. merging.

Issue #5283, opened 2022-04-25
Easily control PlotSize scale per plot or per operation


jbednar:
Right now, if you want high-resolution Datashader plots on a Retina (scaled) display, you can set that at the stream class level:

```python
from holoviews.streams import PlotSize
PlotSize.scale=2.0
```

However, if you just one a single plot to have a different size scale, it's very awkward, because `datashade` and `rasterize` don't normally expose the PlotSize stream that they use. Can we have a `size_scale` argument for those two operations that will get mapped into the `scale` argument of the `PlotSize` stream so that users can easily control it?

Hoxbro:
I'm 95% sure this was fixed with the work done in #5288, so I will close this issue. Feel free to open it again if I'm wrong.

PR #5288, opened 2022-04-29
Added pixel_density parameter to ResamplingOperation base class

jlstevens:
PR to address https://github.com/holoviz/holoviews/issues/5283. 

WIP (not yet tested!) but I am opening the PR so we can find a decent name for the parameter. I quite like `pixel_density` but am open to hearing better suggestions or objections...

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -80,7 +80,7 @@ class ResamplingOperation(LinkableOperation):
        Pixel ratio applied to the height and width. Useful for higher
        resolution screens where the PlotSize stream reports 'nominal'
        dimensions in pixels that do not match the physical pixels. For
-       instance, setting pixel_ratio=2 can give better results on retina
+       instance, setting pixel_ratio=2 can give better results on Retina
        displays.""")
 
     x_range  = param.Tuple(default=None, length=2, doc="""
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -76,7 +76,7 @@ class ResamplingOperation(LinkableOperation):
     width = param.Integer(default=400, doc="""
        The width of the output image in pixels.""")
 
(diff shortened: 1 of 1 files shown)


holoviews/operation/datashader.py
```diff
@@ -76,6 +76,13 @@ class ResamplingOperation(LinkableOperation):
     width = param.Integer(default=400, doc="""
        The width of the output image in pixels.""")
 
+    pixel_ratio = param.Number(default=1, bounds=(1,None), doc="""
+       Pixel ratio applied to the height and width. Useful for higher
+       resolution screens where the PlotSize stream reports 'nominal'
+       dimensions in pixels that do not match the physical pixels. For
+       instance, setting pixel_ratio=2 can give better results on retina
```
jbednar, line 83:
```suggestion
       instance, setting pixel_ratio=2 can give better results on Retina
```

holoviews/operation/datashader.py
```diff
@@ -202,7 +209,9 @@ def _get_sampling(self, element, x, y, ndim=2, default=None):
         xs, ys = (np.linspace(xstart+xunit/2., xend-xunit/2., width),
                   np.linspace(ystart+yunit/2., yend-yunit/2., height))
 
-        return ((xstart, xend), (ystart, yend)), (xs, ys), (width, height), (xtype, ytype)
+        width *= self.p.pixel_ratio
+        height *= self.p.pixel_ratio
+        return ((xstart, xend), (ystart, yend)), (xs, ys), (int(width), int(height)), (xtype, ytype)
```
jbednar, line 214: Up to you on style, but my own vote would be to apply int() in the same operation as applying the ratio, so that the height and width datatype doesn't briefly differ from the expected integer type.

jlstevens: Addressed in ca7935045

jlstevens:
I'll merge so the docs can be updated (and the feature tested). Happy to add some unit tests separately.

PR #5290, opened 2022-04-30
Updated Large Data guide to match changes in Datashader

jbednar:
- Increased resolution of line plots
- Set default Datashader size to match final rendered size (ResamplingOperation.width=500) so that exported plots match live notebook results 
- Added antialiased/thick lines where appropriate, instead of spread
- Made Datashaded plots match Bokeh versions better


jbednar:
Affected by https://github.com/holoviz/holoviews/pull/5288, so if that merges before this one, should update this one to use the mechanism added there.

jbednar:
The test failures on 3.6 appear to be about environments, not this PR's changes.

PR #5294, opened 2022-05-01
Fix the CI on Python 3.6

maximlt:
with a quick attempt to run it again on Python 2.7.


maximlt:
Updates on this PR which is far from being completed:
* The tests on Python 3.6 are hanging. It doesn't seem to be one particular test, it's just the step that doesn't end. Will have to debug this locally maybe to understand what's going on.
* I can't run the tests on Python 2 on master because holoviews has updated its minimum Panel version. I would have to do a merge request to the 1.14.9 branch directly (from a branch created from 1.14).

jlstevens:
All python versions are passing with builds passing on linux and windows. I'll go ahead and merge even though now macos builds have passed yet (I'm not expecting any issues).

PR #5297, opened 2022-05-04
Add compatibility to Python 2.7 again

maximlt:
Until HoloViews 2.0 compatibility to Python 2.7 is meant to be ensured. This PR reactivates the CI to run on Python 2.7 and will certainly have to make a few changes to the code base to get the tests to pass (removing fstrings, etc.).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -109,3 +109,63 @@ jobs:
           eval "$(conda shell.bash hook)"
           conda activate test-environment
           codecov
+  test_suite_26:
+    name: Pytest on ${{ matrix.python-version }}, ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
+        python-version: ['2.7']
+    timeout-minutes: 60
+    defaults:
+      run:
+        shell: bash -l {0}
+    env:
+      MPLBACKEND: "Agg"
+      CHANS_DEV: "-c pyviz -c bokeh -c conda-forge -c nodefaults"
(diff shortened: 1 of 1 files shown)


Hoxbro:
Has there ever been any discussion in releasing a version 1.15, which is (almost) identical to version 1.14.9 but will remove dependencies of Python versions below 3.7?

maximlt:
I don't think so but I may be wrong. I'll let @jlstevens chime in since he has opinion about this whole topic.

jlstevens:
> Has there ever been any discussion about releasing version 1.15? Which is (almost) identical to version 1.14.9 but will remove dependencies of Python versions below 3.7.

This isn't a half bad idea!

In the release notes/CHANGELOG we wrote:

> It is important to note that version 1.14 will be the last HoloViews release supporting Python 2.

This leaves the possibility of a a 1.15 hotfix series (where dropping py2 is the breaking change) which lets us do this without also having to get version 2 out! Thoughts @philippjfr ?

philippjfr:
Agree, seems like a good idea. 2.0 isnt any closer at this point so that would make our lives easier.

maximlt:
Ok so deal, 1.15 will drop support of Python 2.7 (🎉). Should it also drop support of Python 3.6?

philippjfr:
IMO yes, since latest bokeh and panel do not support 3.6 and the only way this actually makes our lives easier if people on different python versions don't end up on completely different versions of our stack.

jlstevens:
As we need a release ASAP and python 2 is actually broken right now (while 3.6 isn't), I suggest we release 1.15 to drop python 2 and revisit 3.6 support as a separate discussion later.

jlstevens:
Python 2 support has been fixed in https://github.com/holoviz/holoviews/pull/5218 for 1.14.9. We will drop support in 1.15.0.

PR #5302, opened 2022-05-06
Don't skip each renderer's load_nb call when multiple extension calls are made in a single cell

maximlt:
Attempt to fix an issue observed when building hvPlot's docs, where no Plotly plot was displayed.

Each holoviews renderer (bokeh, mpl, plotly) has a `load_nb` method that actually doesn’t load any code in the notebook, it’s just doing some extra setting. Plotly’s one does the equivalent of `pn.extension('plotly')` minus the insertion of JS/CSS into the notebook, already done elsewhere. The new pyviz_comms/holoviews/panel have added a way to only load the code once in a notebook cell, even when `.extension()` is called multiple times. At the holoviews level though this also avoids calling each renderer’s `load_nb` method, leading in Plotly’s case to Panel not being set up properly.

The suggested solution is not to skip calling each renderer's `load_nb` method.

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -184,9 +184,9 @@ class notebook_extension(extension):
             config.comms = comms
 
         same_cell_execution = getattr(self, '_repeat_execution_in_cell', False)
+        for r in [r for r in resources if r != 'holoviews']:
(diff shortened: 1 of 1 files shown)


maximlt:
@jlstevens it'd be nice if you could have a look at this PR (to get it in the next release if it's OK).

Issue #5303, opened 2022-05-10
Blue plot


Hoxbro:
Noticed a blue plot on the holoviews documentation page [here](https://holoviews.org/getting_started/Introduction.html#compositional-overlays).  It could be related to https://github.com/holoviz/datashader/issues/1051.

![image](https://user-images.githubusercontent.com/19758978/167686525-003a62d3-a684-4b64-b175-5cf1ad210500.png)

jbednar:
Yes, the blue is fairly distracting here, and I think we should do the same thing we do in Datashader, which is to make a distinction between no data (zero counts show up transparent) and data (non-zero counts are mapped to the colorbar). Bokeh has options for this, but I can't seem to make them work:

```python
import pandas as pd, numpy as np, holoviews as hv
hv.extension('bokeh')

taxi_dropoffs = {hour:arr for hour, arr in np.load('../assets/hourly_taxi_data.npz').items()}
bounds = (-74.05, 40.70, -73.90, 40.80)

image = hv.Image(taxi_dropoffs['0'], ['lon','lat'], bounds=bounds)
image.opts(width=450, tools=['hover'], clipping_colors={'NaN': 'transparent'}, colorbar=True)
image
```
<img width="465" alt="image" src="https://user-images.githubusercontent.com/1695496/167697512-7d1bc52d-00b4-400e-8ac7-c0cc4a0e364b.png">

```python
image.opts(nodata=0)
```
<img width="458" alt="image" src="https://user-images.githubusercontent.com/1695496/167697616-8b617a3c-e785-4ae6-82de-04832718cc25.png">

```python
image.data = np.where(image.data == 0, np.nan, image.data)
image
```
<img width="459" alt="image" src="https://user-images.githubusercontent.com/1695496/167697692-36a02ec8-3f4b-44f6-8dab-d299aeea0ba7.png">

We'd want it to look like the last image here, but to get that to work I had to manipulate the NumPy array directly; shouldn't it work to just specify `nodata`?

Hoxbro:
This works without manipulating the array directly `image.redim.nodata(z=0)`

jbednar:
Thanks! @jlstevens , any idea how `nodata=`*x* is supposed to be used if not like this?

jlstevens:
`image.opts(width=450, tools=['hover'], clipping_colors={'NaN': 'transparent'}, colorbar=True, nodata=0)` should work as `nodata` is both an option on dimensions and a plot option for images.

[...]

PR #5309, opened 2022-05-16
Replaced zeros with nans in hourly_taxi_data.npz

jlstevens:
Fixes the getting started example by replacing zeros in the raster data with nans:

image

Addresses https://github.com/holoviz/holoviews/issues/5303


jbednar:
Thanks. I think that's a better approach than https://github.com/holoviz/holoviews/pull/5304 in that it keeps the user-visible story simpler; I didn't like having to explain `nodata`!

PR #5312, opened 2022-05-17
Expose `rescale_discrete_levels` in the Bokeh backend

maximlt:
This colormap parameter has just been added to Bokeh (https://github.com/bokeh/bokeh/pull/12121) and will be released in part of Bokeh `2.4.3`.

@jbednar I've set it to `True` by default, can you confirm this is correct?

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1759,6 +1759,14 @@ class ColorbarPlot(ElementPlot):
     logz = param.Boolean(default=False, doc="""
          Whether to apply log scaling to the z-axis.""")
 
+    rescale_discrete_levels = param.Boolean(default=True, doc="""
+        If ``cnorm='eq_hist`` and there are only a few discrete values,
+        then ``rescale_discrete_levels=True`` decreases the lower
+        limit of the autoranged span so that the values are rendering
+        towards the (more visible) top of the palette, thus
+        avoiding washout of the lower values.  Has no effect if
+        ``cnorm!=`eq_hist``.""")
+
     symmetric = param.Boolean(default=False, doc="""
         Whether to make the colormap symmetric around zero.""")
 
@@ -1961,6 +1969,7 @@ class ColorbarPlot(ElementPlot):
 
     def _get_cmapper_opts(self, low, high, factors, colors):
         if factors is None:
+            opts = {}
             if self.cnorm == 'linear':
                 colormapper = LinearColorMapper
             if self.cnorm == 'log' or self.logz:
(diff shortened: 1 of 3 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -1983,14 +1992,15 @@ def _get_cmapper_opts(self, low, high, factors, colors):
                                       "Note that the option cnorm='eq_hist' requires "
                                       "bokeh 2.2.3 or higher.")
                 colormapper = EqHistColorMapper
+                if bokeh_version >= LooseVersion('2.4.3'):
+                    opts['rescale_discrete_levels'] = True
```
philippjfr, line 1996:
```suggestion
                    opts['rescale_discrete_levels'] = self.rescale_discrete_levels
```

philippjfr:
One comment :) Otherwise looks good.

holoviews/plotting/bokeh/element.py
maximlt: Brh, time to close the day! Thanks!

maximlt:
Thanks for catching that Philipp!

@jbednar that last commit doesn't invalid the docs build that just completed and was built from this PR: http://dev.holoviews.org/index.html

holoviews/plotting/bokeh/element.py
```diff
@@ -1983,14 +1992,15 @@ def _get_cmapper_opts(self, low, high, factors, colors):
                                       "Note that the option cnorm='eq_hist' requires "
                                       "bokeh 2.2.3 or higher.")
                 colormapper = EqHistColorMapper
+                if bokeh_version >= LooseVersion('2.4.3'):
```
jbednar, line 1995:
```suggestion
                if bokeh_version > LooseVersion('2.4.2'):
```

At least for testing, this guard allow Bokeh dev releases like 2.4.3rc1, so it can't demand 2.4.3 or later.

jbednar:
Looks good, thanks! Yes, I think the default should be `rescale_discrete_levels=True`. On my testing, in every case where I could spot a difference from when it is `False`, the `True` behavior was more useful, less surprising, and more appropriate.  So I think this will finally address the 2017 issue https://github.com/holoviz/datashader/issues/357 where I first complained about this behavior! 

Unfortunately, by default this behavior doesn't actually get invoked, because HoloViews doesn't use `eq_hist` by default with Datashader output. So the benefits of `rescale_discrete_levels` don't apply to these plots that still look fairly broken in their website version but would look better with `eq_hist`:

image
image


image
image


image
image

PR #5313, opened 2022-05-18
Drop Python 2.7

maximlt:
Drop Python 2.7 in the 1.15.0 release.

### DIFF
--- a/examples/getting_started/1-Introduction.ipynb
+++ b/examples/getting_started/1-Introduction.ipynb
@@ -14,7 +14,7 @@
     "\n",
     "## Why HoloViews?\n",
     "\n",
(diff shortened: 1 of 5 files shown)


maximlt:
I've kept Python 2 aliases like `basestring = str` in the `util.py` module as:

* the test suite installs Panel 0.12.6 on Python 3.6, which needs the `unicode` one
* other packages like hvplot and geoviews rely on `basestring` being available in `util.py`.  I've opened PRs to remove that dependency (https://github.com/holoviz/hvplot/pull/757 and https://github.com/holoviz/geoviews/pull/572). But there may be more usage of these aliases around, so let's just keep them.

PR #5316, opened 2022-05-20
Use any as default aggregator for Curve

philippjfr:
Better behavior for simple time-series or other Curve plots where self-intersections are spurious and entirely dependent on zoom-level.

bokeh_plot - 2022-05-20T174814 455
bokeh_plot - 2022-05-20T174801 032


philippjfr:
`aggregate` is a bit of an ugly beast that precedes the idea of having separate operations to dispatch to. The real solution is to actually split that operation up into aggregate_curve, aggregate_path and aggregate_points and aggregate_ndoverlay. I could create those now I suppose and then split up the actual implementation later.

jbednar:
On balance, though the `any` aggregator behaves well for simple cases, those seem like they are also the ones that work well with Bokeh already. With Datashader, a curve can intersect itself many times in a single pixel if it has many more samples than there are pixels on the screen, and with `any` there's no way to determine if the curve is going back and forth equally over a large range, or going up and down in a small range many times but then having one outlying value.

PR #5324, opened 2022-05-30
CI: remove the runs on Python 3.6

maximlt:
HoloViews 1.15 needs "panel >=0.13.1" which supports Python >= 3.7. We no longer need the test suite to run on Python 3.6 then.


maximlt:
While the proportion of downloads was > 5% 6 months ago it went down to < 2%, which sort of justifies dropping support for Python 3.6 (it's also EOL). Sorry Jean-Luc!

image

jlstevens:
Given the situation with 3.6 usage has changed a fair bit in the past 6 months, this seems fine.

PR #5325, opened 2022-05-30
Adding geodataframe to DataConversion

Hoxbro:
First part (1/2) of fix for https://discourse.holoviz.org/t/small-multiples-aka-facetted-plots/3760/2

This will make it possible to run:
``` python
import geopandas as gpd
import hvplot.pandas

countries = gpd.read_file(gpd.datasets.get_path('naturalearth_lowres'))
countries.hvplot(geo=True, by="continent")
```

I'm actually not sure if `cuDF` should be added to `df_interfaces`

Before fix:
image

After fix:
image

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -129,7 +129,8 @@ class DataConversion(object):
             else:
                 selected = self._element
         else:
-            if pd and issubclass(self._element.interface, PandasInterface):
+            df_interfaces = tuple(v for k, v in Interface.interfaces.items() if "dataframe" in k.lower())
+            if issubclass(self._element.interface, df_interfaces):
                 ds_dims = self._element.dimensions()
                 ds_kdims = [self._element.get_dimension(d) if d in ds_dims else d
                             for d in groupby+kdims]


holoviews/core/data/__init__.py
```diff
@@ -129,7 +129,8 @@ def __call__(self, new_type, kdims=None, vdims=None, groupby=None,
             else:
                 selected = self._element
         else:
-            if pd and issubclass(self._element.interface, PandasInterface):
+            df_interfaces = tuple(v for k, v in Interface.interfaces.items() if "dataframe" in k.lower())
```
maximlt, line 132: I would suggest this change to be a little more explicit.

```suggestion
            df_interfaces = tuple(
                v
                for k, v in Interface.interfaces.items()
                if k in ('dataframe', 'geodataframe')
            )
```

Hoxbro: The only thing I'm a little bit afraid of with this change is if a new `dataframe` interface is implemented, it will not automatically be added to `df_interfaces` and therefore potentially go down the wrong code path.  On the other hand, it could properly also be seen as a good thing that a new interface is not automatically added to `df_interfaces`.  :sweat_smile: 

I have no strong opinion about this, so I will be cowardly and let you decide if the suggested changes should be implemented.

maximlt: Actually looking at the code base I see that the cuDF and Dask interfaces inherit from PandasInterface:

```python
In [3]: param.concrete_descendents(hv.core.PandasInterface)
Out[3]: 
{'PandasInterface': holoviews.core.data.pandas.PandasInterface,
 'cuDFInterface': holoviews.core.data.cudf.cuDFInterface,
 'DaskInterface': holoviews.core.data.dask.DaskInterface}
```

So they will pass the `issubclass` check and the code in the if blocks will be executed for these two interfaces.

There are two interfaces that inherit from `MultiInterface` but declare their `base_interface` as being PandasInterface:
```python
In [15]: [kls for kls in param.concrete_descendents(hv.core.MultiInterface).values() if hasattr(kls, '
    ...: base_interface') and issubclass(kls.base_interface, hv.core.PandasInterface)]
Out[15]: 
[holoviews.core.data.spatialpandas.SpatialPandasInterface,
 holoviews.core.data.spatialpandas_dask.DaskSpatialPandasInterface]
```

Maybe the code should be changed to pass the `if` check too. But I'm really unfamiliar with this part of the code base to really tell.

The `GeoPandasInterface` declared in GeoViews inherits from `MultiInterface` but doesn't declare any `base_interface`:
```python
class GeoPandasInterface(MultiInterface):

    types = ()

    datatype = 'geodataframe'

    multi = True
```

Maybe it should?

Thoughts on this @jlstevens ?

maximlt:
@philippjfr we had a discussion about this a couple of weeks ago, can you remember what  you decided about this? I remember there was a suggestion to move the interface from GeoViews to HoloViews. It would be nice if we could summarize the plan so that someone could pick up that work.

holoviews/core/data/pandas.py
```diff
@@ -14,7 +14,20 @@
 from .util import finite_range
 
 
-class PandasInterface(Interface):
+class PandasApi:
+    """
```
jlstevens, line 18: As API is an acronym I would call it `PandasAPI`.

PR #5326, opened 2022-06-02
Test suite clean up and fix for the pip build

maximlt:
This turned into a bigger PR than I thought. My original intention was to fix the pip build that appeared to be broken when it last run on the scheduled jobs. I saw that the grouped dependencies in `setup.py` where not so clear, so started to re-organize them and checked in particular the ones required to run the tests (not the notebooks). That led me to simplify a lot of imports in the tests. For instance in many places `pandas` was assumed to be optional while it's now a required dependency of HoloViews, same for Panel. The tests also now need Bokeh, Matplotlib and Plotly to be installed to run. This made sense to me, they're so critical to HoloViews.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -29,9 +29,6 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: "100"
-      - uses: actions/setup-python@v2
-        with:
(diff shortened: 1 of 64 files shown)


maximlt:
I've manually launched the docs and the packages workflows in dry-run mode and they succeeded, merging then.

PR #5330, opened 2022-06-07
Fix hv.help when pattern is set

maximlt:
I think this was just totally untested so the code was not being updated. I've added a minimal test.

cc @jlstevens


jlstevens:
Looks good to me! Ready to merge?

maximlt:
Yep!

Issue #5255, opened 2022-04-04
Datashader throws exception when Buffer is cleared


pepijndevos:
#### ALL software version info
<details><summary>version info</summary>
<pre>
$ pip freeze
aiocouch==2.2.1
aiohttp==3.8.1
aiosignal==1.2.0
anyio==3.5.0
argon2-cffi==21.3.0
argon2-cffi-bindings==21.2.0
async-timeout==4.0.2
attrs==21.4.0
Babel==2.9.1
backcall==0.2.0
bleach==4.1.0
bokeh==2.4.2
certifi==2021.10.8
cffi==1.15.0
chardet==4.0.0
charset-normalizer==2.0.10
chevron==0.14.0
click==8.0.4
cloudpickle==2.0.0
colorcet==3.0.0
dask==2022.3.0
datashader==0.13.0
datashape==0.5.2
debugpy==1.5.1
decorator==5.1.1
defusedxml==0.7.1
Deprecated==1.2.13
distributed==2022.3.0
docker==5.0.3
entrypoints==0.3
escapism==1.0.1
frozenlist==1.3.0
fsspec==2022.2.0
HeapDict==1.0.1
holoviews==1.14.8
hvplot==0.7.3
ibm-cloud-sdk-core==3.13.2
ibmcloudant==0.0.41
idna==3.3
ipykernel==6.6.1
ipython==7.31.0
ipython-genutils==0.2.0
ipywidgets==7.6.5
iso8601==1.0.2
jedi==0.18.1
Jinja2==3.0.3
json5==0.9.6
jsonschema==4.4.0
jupyter-bokeh==3.0.4
jupyter-client==7.1.0
jupyter-core==4.9.1
jupyter-repo2docker==2021.8.0
jupyter-server==1.13.2
-e git+https://github.com/jupyterhub/jupyter-server-proxy.git@26d7569a7d2e7c806fcafe3b538ecbe061e7c0d8#egg=jupyter_server_proxy
jupyterlab==3.2.7
jupyterlab-pygments==0.1.2
jupyterlab-server==2.10.3
jupyterlab-widgets==1.0.2
llvmlite==0.38.0
locket==0.2.1
Markdown==3.3.6
MarkupSafe==2.0.1
matplotlib-inline==0.1.3
mistune==0.8.4
more-itertools==8.12.0
msgpack==1.0.3
multidict==6.0.2
multipledispatch==0.6.0
nbclassic==0.3.5
nbclient==0.5.9
[...]

PR #5331, opened 2022-06-08
Checking for empty data ranges in raster elements

jlstevens:
PR to address https://github.com/holoviz/holoviews/issues/5255

The datashader operation was using `_empty_agg` to return an empty `Image` element which then failed to display with a `ZeroDivisionError` after attempting a bounds calculation. I'm not sure this is quite the right fix but now this element displays without error:

image

Though I now notice that this isn't enough as both the matplotlib and plotly backends still have trouble: plotly complains with `Out of range float values are not JSON compliant` and matplotlib complains with `ValueError: zero-size array to reduction operation minimum which has no identity`.  

Looks like plotting libraries don't believe that empty images are a thing either! ;-p

### DIFF
--- a/holoviews/element/raster.py
+++ b/holoviews/element/raster.py
@@ -299,7 +299,7 @@ class Image(Selection2DExpr, Dataset, Raster, SheetCoordinateSystem):
             data_bounds = self.bounds.lbrt()
 
         non_finite_bounds = all(not util.isfinite(v) for v in bounds.lbrt())
-        data_ranges = self.interface.range(self, kdims[0]) + self.interface.range(self, kdims[1])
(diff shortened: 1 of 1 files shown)


jlstevens:
Actually, the 'plotly' JSON error is being raised in Bokeh via Panel. Not sure what to think here...

jbednar:
This change looks very reasonable, but it sounds like the additional fixes for Plotly and Matplotlib should be folded in as well.

jlstevens:
Looks like this approach breaks a load of tests. I'll close this PR as superseded by https://github.com/holoviz/holoviews/pull/5334

PR #5338, opened 2022-06-17
Add support for rendering in pyodide/pyscript

philippjfr:
Implements a `pyodide_extension` that registers MIME rendering hooks that work in pyodide/pyscript.

### DIFF
--- a/holoviews/__init__.py
+++ b/holoviews/__init__.py
@@ -1,4 +1,4 @@
-import os, io
+import io, os, sys
 
 import numpy as np # noqa (API import)
 import param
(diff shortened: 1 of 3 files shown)


MarcSkovMadsen:
Will this enable streaming to plots or from plots? I.e. dynamically updating the plot data or getting hover/ selection etc. events.

philippjfr:
> Will this enable streaming to plots or from plots? I.e. dynamically updating the plot data or getting hover/ selection etc. events.

This is already possible with master if you render plots manually using the `pn.io.pyodide.write` or `.servable(target=...)` API.

PR #5343, opened 2022-06-23
Minor optimizations to renderers

Hoxbro:
Fixes #5340 

The primary reason for the memory leak was because `plot.document` was set to `curdoc` here:
https://github.com/holoviz/holoviews/blob/22db2d1dd3b5d0b42a08f00645e04f693ad63dbb/holoviews/plotting/bokeh/renderer.py#L72

Where `plot` is an element that does not have `document` set up as a `param.Parameter` when setting `document` as a parameter, I get the following memory consumptions (cont. from [here](https://github.com/holoviz/holoviews/issues/5340#issuecomment-1161491369)):

Sankey:
sankey

Curve:
curve

With the current changes, it still seems to use a little more memory than without `hv.render`.

PR #5344, opened 2022-06-24
Test updates following release of datashader 0.14.1

ianthomas23:
This PR fixes the HoloViews test failures which were identified in PR #5343. There were two sets of failures, both relating to HoloViews tests needing to be updated following changes released in Datashader 0.14.1. They were:

1. Improved `trimesh` rendering.
2. The default setting of `rescale_discrete_levels=True` in `shade`.

The former required changing the comparison arrays; I have visually checked all of these. The latter required setting `rescale_discrete_levels=False` and then the tests pass.

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -1227,6 +1227,13 @@ def is_number(obj):
     else: return False
 
 
+def is_float(obj):
+    """
+    Checks if the argument is a floating-point scalar.
+    """
+    return isinstance(obj, (float, np.floating))
+
+
 def is_int(obj, int_like=False):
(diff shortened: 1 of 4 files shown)


ianthomas23:
There are some extra test failures in CI where we are using equality for floating-point comparisons rather than allowing some tolerance. The actual problem is that we compare the types of two objects before deciding which comparison operator to choose and the code concludes that `float` and `np.float64` are not the same types and drops to a default exact comparison. The solution is therefore to allow any two floating-point types to use the floating-point comparison rather than direct equality.

Issue #5341, opened 2022-06-21
rasterize with opts raises warning in latest dev release


Hoxbro:
#### ALL software version info
``` yaml
datashader : 0.14.0
holoviews  : 1.15.0a2
```

#### Description of expected behavior and the observed behavior
The combination of `rasterize` and using `opts` will give a warning with the latest dev release of holoviews. The warning is not present with version 1.14. The warning is only present with `rasterize` and not without it.

It seems the warning is only coming from this function:
https://github.com/holoviz/holoviews/blob/83ee1813d28fd870f61e7607fd191346d509cccd/holoviews/core/spaces.py#L460-L468

#### Complete, minimal, self-contained example code that reproduces the issue

``` python
import holoviews as hv
from holoviews.operation.datashader import rasterize
hv.extension("bokeh")

plot = rasterize(hv.Curve(range(10)))
(plot + plot).opts(shared_axes=False)

```

#### Screenshots or screencasts of the bug in action

With `holoviews` 1.15:
![image](https://user-images.githubusercontent.com/19758978/174781890-60e1027e-e4f9-446f-9a08-9e2ec18c1c80.png)

With `holoviews` 1.14:
![image](https://user-images.githubusercontent.com/19758978/174782022-f8f14c4c-1d53-4812-aa56-f347f57c5706.png)

PR #5345, opened 2022-06-27
Ensure setting opts on DynamicMap does not generate deprecation warnings

philippjfr:
The `DynamicMap._style` method is responsible for applying options set on a `DynamicMap`. It does so by looking up a custom options tree and then setting that on the returned object. The format of that custom options tree is broken down by the options group (i.e. 'style', 'norm' and 'plot') and then gets applied using the public `.opts()` call. This is problematic because `.opts` explicitly checks that options are not set using the deprecated option groups format. Therefore we bypass that call and directly apply the option groups. 

Fixes https://github.com/holoviz/holoviews/issues/5341


jlstevens:
Thanks for the clear explanation and the change looks fine. Happy to see this merged when the tests pass.

PR #5349, opened 2022-07-05
Version 1.15.0

jlstevens:
This PR is still WIP and will update the CHANGELOG and releases.rst for the upcoming 1.15.0 release.

- [x] There are still a few outstanding PRs I need to review before the list of complete
- [x] There are some new PRs merged since this set of PRs was computed that also need review.
- [x] Update `releases.rst`

PR #5350, opened 2022-07-07
Added final CHANGELOG entry for 1.15.0 release

jlstevens:
Addresses final comment in https://github.com/holoviz/holoviews/pull/5349 regarding the need for an entry in the compatibility section to inform users of changes to the Sankey algorithm.


jlstevens:
I think that is it! Going to merge and tag the release.

PR #5390, opened 2022-08-13
Add support for stepwise Area plots

j-svensmark:
Add support for the `step` option of the matplotlib `fill_between` method.
https://matplotlib.org/stable/api/_as_gen/matplotlib.axes.Axes.fill_between.html


philippjfr:
Thanks @j-svensmark!

PR #5397, opened 2022-08-18
Fix polar limits

### LINKED ISSUE #5396: Scatter polar projection main plot disappears when theta above 5.3 radians
hv version '1.14.9'

```
        
IPython          : 8.2.0
        ipykernel        : 6.13.0
        ipywidgets       : 7.7.0
        jupyter_client   : 7.2.2
        jupyter_core     : 4.9.2
        jupyter_server   : not installed
        jupyterlab       : not installed
        nbclient         : 0.6.0
        nbconvert        : 6.5.0
        nbformat         : 5.3.0
        notebook         : 6.4.11
        qtconsole        : 5.3.0
        traitlets        : 5.1.1
```

Chrome Version 104.0.5112.81 (Official Build) (64-bit)

 I am having trouble with the display of polar Scatter when the values go above 5.3 radians. Most of the plot disappears.

```
hv.extension('matplotlib')

theta = np.arange(0, 5.4, 0.1)

r = np.ones(len(theta))

scatter = hv.Scatter((theta, r), 'theta', 'r')
scatter.options(projection='polar', show_grid=True)

```

image

If you reduce the radian value then it is okay

```
hv.extension('matplotlib')

theta = np.arange(0, 5.3, 0.1)

r = np.ones(len(theta))

scatter = hv.Scatter((theta, r), 'theta', 'r')
scatter.options(projection='polar', show_grid=True)
```
image

Hoxbro:
Fixes #5396 

image

### DIFF
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -336,6 +336,10 @@ class ElementPlot(GenericElementPlot, MPLPlot):
                     axis.set_zlim(bottom=zmin)
                 if valid_lim(zmax):
                     axis.set_zlim(top=zmax)
+        elif isinstance(self.projection, str) and self.projection == "polar":
+            _, b, _, t = coords
+            l = 0
+            r = 2 * np.pi
         else:
             l, b, r, t = coords
 
--- a/holoviews/tests/plotting/matplotlib/test_elementplot.py
+++ b/holoviews/tests/plotting/matplotlib/test_elementplot.py
@@ -7,6 +7,8 @@ from holoviews.streams import Stream
 from .test_plot import TestMPLPlot, mpl_renderer
 
 from matplotlib.ticker import FormatStrFormatter, FuncFormatter, PercentFormatter
+from matplotlib.projections import PolarAxes
+
 
 class TestElementPlot(TestMPLPlot):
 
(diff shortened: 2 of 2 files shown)


jlstevens:
Thanks for the quick fix!

The diff looks fine to me and the additional unit test gives me confidence. Could you explain why this bug manifested itself for the scatter element but not for a curve? And why did it seem to work up to a certain angle? I'm guessing the polar axis wasn't being defined for a full circle.

Why would the polar axes be ok for one element and not another?

Once my curiosity is satisfied, I am happy to merge. :-)

Hoxbro:
I don't think any of the plots worked there was always a slice missing. 

https://github.com/holoviz/holoviews/blob/4c118b3c44ccdc33ac6130122c0290957f25d30b/holoviews/plotting/mpl/element.py#L321
Returns the following for the three examples in #5396: 
```
a = [-0.53, -0.2, 5.83, 2.2]
b = [-0.52, -0.2, 5.72 2.2]
c = [0.0, -0.2, 5.3, 2.2]
```
So the reason why the Curve plot is fine is that the `extents` is not padded. 

Then somewhere, a round-trip of going from coordinates back to radians is done, which can be faked with something like this `np.arcsin(np.sin(x)))` which for `a` gives `l=-0.53` and `r=-0.45`, which gives the small slice.  For `b` we get  `l=-0.52` and `r=-0.56`, which gives the almost full circle. 

I hope this satisfies your curiosity.

jlstevens:
> I hope this satisfies your curiosity.

Yes thanks. merging!

PR #5413, opened 2022-08-30
Fixed declaration of Scatter as Selection1DExpr

jlstevens:
The semantics of chart elements means they should all be `Selection1DExpr`: `Points` is the appropriate, corresponding 2D element.

### DIFF
--- a/holoviews/element/chart.py
+++ b/holoviews/element/chart.py
@@ -5,7 +5,7 @@ from ..core import util
 from ..core import Dimension, Dataset, Element2D, NdOverlay, Overlay
 from ..core.dimension import process_dimensions
 from .geom import Rectangles, Points, VectorField # noqa: backward compatible import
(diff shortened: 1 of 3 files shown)


jlstevens:
I do want this in the upcoming minor release - I'm sure the change here is the semantically correct thing to do so I'll need to update the tests accordingly.

jlstevens:
I've now updated the unit tests though I am a little suspicious of some of them...

For instance, it looks like I can comment out `invert_xaxis=True` and `invert_yaxis=True` in the `test_selection_expr_stream_invert_xaxis_yaxis` test [here](https://github.com/holoviz/holoviews/blob/master/holoviews/tests/test_streams.py#L1185) and the test still passes! 

Do you agree that something is off @philippjfr ?

jlstevens:
Tests have now passed. @philippjfr please merge if it looks good to you.

philippjfr:
Looks good.

PR #5414, opened 2022-09-01
WIP: Ensure rectilinear rasters support geometry selections

philippjfr:
Somewhat hacky fix to ensure that:

- `Image` correctly implements dim expression selections
- Geometry selections work on rectilinear coordinates

### DIFF
--- a/holoviews/core/data/xarray.py
+++ b/holoviews/core/data/xarray.py
@@ -422,6 +422,10 @@ class XArrayInterface(GridInterface):
             return data.T.flatten() if flat and not keep_index else data
         elif expanded:
             data = cls.coords(dataset, dim.name, expanded=True)
+            if keep_index:
+                da = dataset.data[dataset.vdims[0].name].copy().rename(dim.name)
+                da.data[:] = data
+                return da
             return data.T.flatten() if flat else data
         else:
             if keep_index:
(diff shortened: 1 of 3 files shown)


holoviews/core/data/xarray.py
```diff
@@ -422,6 +422,10 @@ def values(cls, dataset, dim, expanded=True, flat=True, compute=True, keep_index
             return data.T.flatten() if flat and not keep_index else data
         elif expanded:
             data = cls.coords(dataset, dim.name, expanded=True)
+            if keep_index:
+                da = dataset.data[dataset.vdims[0].name].copy().rename(dim.name)
+                da.data[:] = data
+                return da
```
philippjfr, line 428: Need to find a better solution here. This basically ensures that `keep_index` + `expanded` returns a 2D coordinate array even for a rectilinear (i.e. 1D coordinate) array. This implementation does not handle:

- packed vdims
- dask backed value arrays

holoviews/element/selection.py
```diff
@@ -73,7 +73,7 @@ def spatial_select_gridded(xvals, yvals, geometry):
         target = Image((xs, ys, np.empty(ys.shape+xs.shape)))
         poly = Polygons([geometry])
         mask = rasterize(poly, target=target, dynamic=False, aggregator='any')
-        return mask.dimension_values(2, flat=False)
+        return mask.interface.values(mask, mask.vdims[0], keep_index=True)
     else:
         mask = spatial_select_columnar(xvals.flatten(), yvals.flatten(), geometry)
         return mask.reshape(xvals.shape)
```
philippjfr, line 79: This branch should also ensure that the mask that is returned is still an xarray object if the input is an xarray object.

PR #5452, opened 2022-09-19
Return `decimate` samples in order

### LINKED ISSUE #5451: Interpolated charts (Curve, Area, etc) plot data in the provided order, but decimate unsorts data
| Library | Version |
| -- | -- |
| python | 3.9.13 |
| holoviews | 1.15.0 |
| bokeh | 2.4.3 |
| pandas | 1.4.4 |
| numpy | 1.23.3 |

I should be able to decimate Curve/Area charts the same way I can decimate scatter charts.  Decimating interpolated charts currently results in garbled output.

```python
import numpy as np
import pandas as pd
import holoviews as hv

hv.extension('bokeh')
x = np.linspace(0, 10, 100)
y1 = np.sin(x)
y2 = np.cos(x)

table = hv.Table((x, y1, y2), 'x', ['y1', 'y2'])
hv.Area(table)  # See Figure 1

---------------------------

from holoviews.operation import decimate
decimate(hv.Area(table), max_samples = 50)  # See Figure 2
```

Figure 1
image

Figure 2
image

ddrinka:
Closes #5451
The Area chart type does not function properly with out-of-order data. When sampling for `decimate` results, return data in the original order.

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -909,7 +909,8 @@ class decimate(Operation):
 
         if len(sliced) > self.p.max_samples:
             prng = np.random.RandomState(self.p.random_seed)
-            return sliced.iloc[prng.choice(len(sliced), self.p.max_samples, False)]
+            choice = prng.choice(len(sliced), self.p.max_samples, False)
+            return sliced.iloc[np.sort(choice)]
         return sliced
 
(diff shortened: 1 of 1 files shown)


Hoxbro:
Thank you for the issue and the fix.

The abstract decimate plot is now a thing of the past!

PR #5455, opened 2022-09-24
Faster hashing for arrays and pandas objects

philippjfr:
Our hashing implementations for arrays and pandas objects were both horrifically slow. Here we use the approach also used by Panel and Streamlit which is to sample 1m values and then hashing that. The chance of collisions is still vanishingly small and for a 2500 x 2500 array this approach is about 35x faster (7 seconds -> 200 milliseconds).

Fixes https://github.com/holoviz/panel/issues/3874

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -1,9 +1,11 @@
 import sys, warnings, operator
 import builtins as builtins   # noqa (compatibility)
+import hashlib
 import json
 import time
 import types
 import numbers
+import pickle
 import inspect
 import itertools
 import string
@@ -39,6 +41,11 @@ arraylike_types = (np.ndarray,)
 masked_types = ()
 
 anonymous_dimension_label = '_'
+
+_NP_SIZE_LARGE = 1_000_000
+_NP_SAMPLE_SIZE = 1_000_000
+_PANDAS_ROWS_LARGE = 1_000_000
+_PANDAS_SAMPLE_SIZE = 1_000_000
 
 pandas_version = LooseVersion(pd.__version__)
 try:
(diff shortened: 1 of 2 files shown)


philippjfr:
Okay, I'll add some tests when I get a chance.

jlstevens:
As there are some pathological cases where this may fail (e.g. a giant zero array with a single value in one place) is there any way we could at least document what failure looks like (even if it is only developer docs)?

jlstevens:
As I'm planning to merge this for the imminent minor release, I will mention this in the release notes. Hopefully, you can add a few tests before the release, if not nevermind and we will merge anyway!

philippjfr:
Thanks @Jhsmit. The problem is that we also need to handle 10 or 100s of millions of rows performantly and sampling is the only way to do so performantly. Completely avoiding cash collisions when there's sampling involved is impossible so I was just hoping to reduce it a little more.

Jhsmit:
I got the need for the subsampling, not sure if we understand eachtother but the point is that the use of `.sum()` that can give problems.

The issue is that `hash_pandas_object` calculates integer hashes per row in the dataframe, thus by summing up this result you get identical hashes if you change the order of the rows.


```python
np.random.seed(43)
df = pd.DataFrame(np.random.rand(10, 2))
b"%s" % pd.util.hash_pandas_object(df).sum()

>>> b'\x1f\x95$`\xe2\xa2y\xb7'
```

```python

df_sorted = df.sort_values([0])

b"%s" % pd.util.hash_pandas_object(df_sorted ).sum()

 >>> b'\x1f\x95$`\xe2\xa2y\xb7'
```

Although they are very different dataframes. But instead of sum just update the hash with the next value, the hashes are different:

```python
h = hashlib.new("md5")
for val in pd.util.hash_pandas_object(df):
    h.update(str(val).encode("UTF-8"))
h.hexdigest()

>>> '7fc844625ab097d980f13778233ccc56'
```

```python
h = hashlib.new("md5")
for val in pd.util.hash_pandas_object(df_sorted ):
    h.update(str(val).encode("UTF-8"))
h.hexdigest()
>>> '5684f7e2e039c879f7234774bc0e9585"
```

While the performance is the same.

philippjfr:
Ah sorry, my fault. I had originally been thinking about a separate suggestion of adding column sums of the unsampled dataframe to reduce the pathological cases but I now notice I only ever mentioned that on the internal slack. Only just noticed that the implementation here also uses `.sum()`. I'll use your version now.

PR #5457, opened 2022-09-28
Revert "Raising DataError when passing DataFrame with int columns to Dataset"

Hoxbro:
Reverts holoviz/holoviews#5354

As it is a breaking change. It will be reverted and added in `1.16` or `2.0` (whatever comes first). 

For now, a warning will be given when using a DataFrame with integers as column names.


jlstevens:
Looks good to me! Ready to merge @Hoxbro ?

PR #5460, opened 2022-09-28
Use colorcet register_cmap

Hoxbro:
When https://github.com/holoviz/colorcet/pull/90 is merged and released, it will remove the DeprecationWarning.

### DIFF
--- a/holoviews/plotting/mpl/__init__.py
+++ b/holoviews/plotting/mpl/__init__.py
@@ -2,9 +2,8 @@ import os
 
 from matplotlib import rc_params_from_file
 from matplotlib.colors import ListedColormap, LinearSegmentedColormap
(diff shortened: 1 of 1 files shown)


philippjfr:
Either need to have a fallback or need to pin the new version of colorcet.

Hoxbro:
The `register_cmap` function has been in colorcet as a "normal" matplotlib import for a long time, and the updated `register_cmap` function added in https://github.com/holoviz/colorcet/pull/90 should be backward compatible.  So I don't think a pin is necessary, but correct me if I'm wrong.

philippjfr:
Ah did not know it was already there before the version. Ignore me then.

jlstevens:
Colorcet has now been released with that PR merged there. Merging here too, thanks!

PR #5467, opened 2022-10-03
CI: pin pydata-sphinx-theme to 0.9.0

maximlt:
pydata-sphinx-theme 0.10.1 doesn't support usage of `self` in a toc. Pinning it for now.


maximlt:
Merging as the docs workflow passed.

PR #5472, opened 2022-10-06
Upgrade warning and set stacklevel to first file outside holoviews

Hoxbro:
The warning will now emit a FutureWarning in a notebook. 

I found a way to determine the first stack level outside holoviews, hopefully making it easier to understand.

image

### DIFF
--- a/holoviews/core/data/pandas.py
+++ b/holoviews/core/data/pandas.py
@@ -1,11 +1,7 @@
-import os
-import inspect
-from functools import lru_cache
-from warnings import warn
-
 import numpy as np
 import pandas as pd
 
+from ...util._exception import deprecation_warning
 from .interface import Interface, DataError
 from ..dimension import dimension_name
 from ..element import Element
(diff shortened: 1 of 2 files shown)


jlstevens:
Happy to see this merged!

holoviews/core/data/pandas.py
```diff
@@ -16,7 +18,27 @@
 @lru_cache(maxsize=None)
 def deprecation_warning(msg):
```
maximlt, line 19: If this function can be reused elsewhere, then let's make it a general shared utility?

holoviews/core/data/pandas.py
```diff
@@ -16,7 +18,27 @@
 @lru_cache(maxsize=None)
```
maximlt, line 18: Wondering if this is actually needed. Aren't warnings generally just shown once? [Python docs](https://docs.python.org/3.10/library/warnings.html) says that "Repetitions of a particular warning for the same source location are typically suppressed.".

holoviews/core/data/pandas.py
```diff
@@ -16,7 +18,27 @@
...
+    stacklevel = 1
+    while frame:
+        fname = inspect.getfile(frame)
+        if (fname.startswith(pkg_dir) or fname.startswith(param_dir)) and not fname.startswith(test_dir):
+            frame = frame.f_back
+            stacklevel += 1
+        else:
+            break
+
+    warn(msg, FutureWarning, stacklevel=stacklevel)
```
maximlt, line 41: 👍

holoviews/core/data/pandas.py
Hoxbro: `lru_cache` can be removed. 

The main reason for it being there (in my mind at least) is not to "spam" `hvplot` users (example without no `lru_cache`):
image

All these spams arise because it is a new location for each call. If the location is the same, the warning will only be raised once instead of ten times.
 
image

maximlt: Let's keep it then, that looks horrible otherwise!

holoviews/core/data/pandas.py
Hoxbro: I actually tried adding it to `holoviews.utils.__init__.py` but got `ImportError: cannot import name 'deprecation_warning' from partially initialized module 'holoviews.util' (most likely due to a circular import)`, which is why it stayed in the file (until now).

PR #5470, opened 2022-10-05
Fix warnings

Hoxbro:
Working through the backlogs of warnings from the test suite.

### Warnings after

``` python-traceback
../../../../miniconda3/envs/holoviz/lib/python3.9/site-packages/plotly/graph_objs/__init__.py:288
../../../../miniconda3/envs/holoviz/lib/python3.9/site-packages/plotly/graph_objs/__init__.py:288
../../../../miniconda3/envs/holoviz/lib/python3.9/site-packages/plotly/graph_objs/__init__.py:288
../../../../miniconda3/envs/holoviz/lib/python3.9/site-packages/plotly/graph_objs/__init__.py:288
  /home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/plotly/graph_objs/__init__.py:288: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    if LooseVersion(ipywidgets.__version__) >= LooseVersion("7.0.0"):
```

It needs to be fixed by plotly upstream: https://github.com/plotly/plotly.py/pull/3897 

``` python-traceback
holoviews/tests/core/data/test_spatialpandas.py: 21 warnings
  /home/shh/Development/holoviz/repos/holoviews/holoviews/core/data/spatialpandas.py:806: VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray.
    data = {k: v if isscalar(v) else np.asarray(v) for k, v in data.items()}
```
It depends on the future of spatialpandas.

``` python-traceback
holoviews/tests/operation/test_datashader.py::DatashaderRasterizeTests::test_rasterize_dask_trimesh
holoviews/tests/operation/test_datashader.py::DatashaderRasterizeTests::test_rasterize_dask_trimesh_implicit_nodes
holoviews/tests/operation/test_datashader.py::DatashaderRasterizeTests::test_rasterize_dask_trimesh_with_node_vdims
  /home/shh/miniconda3/envs/holoviz/lib/python3.9/pickle.py:329: DeprecationWarning: Please use `append` from the `scipy.optimize` namespace, the `scipy.optimize.slsqp` namespace is deprecated.
    obj = getattr(obj, subpath)

holoviews/tests/operation/test_datashader.py::DatashaderRasterizeTests::test_rasterize_dask_trimesh
[...]

PR #5473, opened 2022-10-07
Remove iterrows for loop from connect_edges_pd

Hoxbro:
In PR #5470 I noticed the use of `df.iterrows` and a for loop, which could be removed. 

`df.iterrows` is known to be notorious slow. 

image


jbednar (approved):

maximlt:
LGTM

PR #5481, opened 2022-10-14
histogram op: raise error on non numerical data

### LINKED ISSUE #906: Make it very clear that `hist` only support numerical or datetime fields
I'm trying to show that it is much easier to use the HoloViz ecosystem to develop something like https://www.youtube.com/watch?v=htXgwEXwmNs with crossfiltering.

One of the plots is a plotly `histogram`.

It works like

```python
import plotly.express as px
df = px.data.tips()
px.histogram(df, "day", height=400,)
```

But if I run

```python
import plotly.express as px
import hvplot.pandas
df = px.data.tips()
df.hvplot.hist("day", height=400)
```

I get

```bash
numpy.core._exceptions._UFuncNoLoopError: ufunc 'multiply' did not contain a loop with signature matching types (dtype('<U3'), dtype('float64')) -> None
```

This is not a useful error message and took me quite some time to figure out what was wrong. Pandas `.plot` handles this much nicer.

```python
import plotly.express as px
df = px.data.tips()
df.hist("day", height=400)
```

```bash
ValueError: hist method requires numerical or datetime columns, nothing to plot.
```

**Please provide a similar useful error message and document it in the reference guide and docstring. Thanks.**

## Additional Context

The next issue is that I should then probably do my own aggregations and use HoloViews `bars` which does not support linked selections. Hmmmm. Its not easy.

maximlt:
Fixes https://github.com/holoviz/hvplot/issues/906 in hvPlot

The main goal of this PR was to validate the data to be processed by the `histogram` operation, that data must be numerical/datetime-like to make sense. Guarding is tricky in HoloViews with all the data types supported, I hope this is alright (the tests passed locally). While I was at it I've also done the same with for the `bins` parameter.


holoviews/operation/element.py
```diff
@@ -700,6 +700,14 @@ class histogram(Operation):
       Used for setting a common style for histograms in a HoloMap or AdjointLayout.""")
 
     def _process(self, element, key=None):
+
+        bad_bins_msg = 'bins only accept numerical data.'
```
jlstevens, line 704: I think this message could be improved a bit. How about:

'Histogram binning can only be performed on numerical data'

holoviews/operation/element.py
```diff
@@ -700,6 +700,14 @@ class histogram(Operation):
...
+        if (isinstance(self.p.bins, np.ndarray) and
+           (self.p.bins.dtype.kind in 'SU' or self.bins.dtype.kind == 'O' and not isdatetime(self.bins))):
+            raise TypeError(bad_bins_msg)
+        elif isinstance(self.p.bins, (list, tuple)) and any(not is_number(bin) for bin in self.p.bins):
+            raise TypeError(bad_bins_msg)
```
jlstevens, line 709: Given all these boolean expressions, it would be good to pull them out onto separate lines for clarity.

holoviews/operation/element.py
```diff
@@ -716,8 +724,16 @@ def _process(self, element, key=None):
 
         if hasattr(element, 'interface'):
             data = element.interface.values(element, selected_dim, compute=False)
+            dtype = element.interface.dtype(element, selected_dim)
```
jlstevens, line 727: Is `dtype` not easily available with `compute=False`?

jlstevens:
Made a few comments and overall looks good. Could you add a test to see if the type error is raised when expected?

Hoxbro:
I can see that the improved error message requested for categorial data in https://github.com/holoviz/hvplot/issues/906 have been solved in #5540. E.g, this code now gives a more descriptive error message:

``` python
import plotly.express as px
import hvplot.pandas
df = px.data.tips()
df.hvplot.hist("day", height=400)
```

I can't at a quick glance see if this PR also fixes other issues with categorical data in histograms.

hoxbro:
No answer in a year, so I will close this PR.

PR #5494, opened 2022-10-24
Improve error message for hv.opts without a plotting backend

### LINKED ISSUE #5463: Confusing error from hv.opts prior to hv.extension()
Current published and master versions of HoloViews give a bewildering error when trying to create element-specific options objects:

image

hv.opts apparently only populates its various accessors when hv.extension loads, but the error message doesn't convey anything to me about "run hv.extension()". I've encountered this issue several times over the years, usually when I'm trying to isolate some code that (in my mind) has nothing to do with a particular backend. Each time I get completely sidetracked -- I'm trying to debug one thing, and now suddenly I can't even set an option on hv.opts.Image? It's particularly a problem in a Jupyter notebook, when I restart the kernel, then try to look at a tiny bit of code that has nothing to do with actual plotting, and suddenly things like this don't work. Eventually I do figure out that it's because hv.extension needs to be run, but each time I've now wasted a good bit of time.

Is there a way that instead of failing with this obscure message, instead it could print something helpful, like "No subobject of type Image has been registered; maybe you need to run hv.extension to declare a plotting backend?"?

Hoxbro:
Fixes #5463

With this PR a better error message is given when a plotting backend is not initialized with `hv.extension`

image

### DIFF
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -48,8 +48,24 @@ def examples(path='holoviews-examples', verbose=False, force=False, root=__file_
         print('Cannot find %s' % tree_root)
 
 
+class OptsMeta(param.parameterized.ParameterizedMetaclass):
+    """
+    Improve error message when running something
+    like: 'hv.opts.Curve()' without a plotting backend.
+    """
 
-class opts(param.ParameterizedFunction):
+    def __getattr__(self, attr):
+        try:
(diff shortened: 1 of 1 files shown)


jbednar (approved):
Thanks!

jlstevens:
Looks good to me and the CI errors are unrelated. Merging.

PR #4712, opened 2020-11-30
Implement support for robust color limits

philippjfr:
Implement https://github.com/holoviz/holoviews/issues/4680

PR #5495, opened 2022-10-28
Fix `clim_percentile`

maximlt:
The `clim_percentile` option was added in https://github.com/holoviz/holoviews/pull/4712 but not tested and contained two small bugs, the first one making the option never available basically. I've tested `clim_percentile` as I was trying to run the plotting guide of xarray and was wondering if we had something similar to their `robust=True` option.

I'd like to write a test for this now. I'm not sure how to obtain the ranges so here's what I came up with:

```python
import holoviews as hv
import numpy as np

hv.extension('bokeh')

arr = np.random.rand(10,10)
arr[0, 0] = -100
arr[-1, -1] = 100
im = hv.Image(arr).opts(clim_percentile=True)

bk_renderer = hv.Store.renderers['bokeh']
plot = bk_renderer.get_plot(im)

low, high = plot.ranges[('Image',)]['z']['robust']

assert low > 0
assert high < 1
```

The second issue I have is, if this test is alright, where to save it?


jlstevens:
Looks good to me! Does look like the original implementation was rather buggy...

@maximlt Ready to merge? All tests are passing except on macos...

jlstevens:
Ah, I guess you want to add some docs first!

maximlt:
This was previously untested and I'd like to add at least a test before merging. The thing is that I'm not quite sure where to write the test, and if the test I suggest (in the OP) is right.

Sure why not documenting that properly. Where would you recommend adding that?

jlstevens:
I'm ok with testings this as part of the bokeh plotting tests only. Technically the change applies to all backends, so arguably there should a similar test for matplotlib and plotly but I would say that is entirely optional. 

The approach you've suggested above looks ok to me. Happy to see this fix merged as soon as the tests goes in!

Hoxbro:
I will merge this as I want this fix into the next release.

Issue #5407, opened 2022-08-25
.opts / .options warning raised when no options are set


maximlt:
When calling `.opts` on an element without passing any options a warning is emitted.
```python
import holoviews as hv
hv.extension('bokeh')

hv.Scatter([0]).opts()
```
```
WARNING:param.main: Calling the .opts method with options broken down by options group (i.e. separate plot, style and norm groups) is deprecated. Use the .options method converting to the simplified format instead or use hv.opts.apply_groups for backward compatibility.
```

Users are unlikely to write that directly, but libraries like GeoViews use that in various places, for example:
https://github.com/holoviz/geoviews/blob/master/geoviews/element/geo.py#L149-L150

maximlt:
Maybe it's just a matter of removing this condition:
https://github.com/holoviz/holoviews/blob/master/holoviews/core/util.py#L248-L249

jbednar:
I can't tell what that used to point to, but I do think there should be an easy check that would fix this, and it's important to avoid those false alarms.

Hoxbro:
I noticed that an empty `opts` will reset previous options, this should not be the case in my opinion.

![image](https://user-images.githubusercontent.com/19758978/198291991-06ee7566-ab0b-4197-959b-ebcc9333a0e7.png)

jlstevens:
As `.opts.clear()` exists already, the easiest thing is to do nothing when you have an empty`.opts` (so it is a no-op, without the warning).

PR #5496, opened 2022-10-28
Fix for empty opts warning and incorrect clearing semantics

jlstevens:
Simple fix for #5407.

Discussing this with @maximlt @Hoxbro, we will follow up with a PR to finally remove the public version of this format which we will merge before the next minor release.


jlstevens:
Here is the follow up PR: https://github.com/holoviz/holoviews/pull/5497

jlstevens:
Tests seems to have passed (except the macos tests that have been failing recently).

I'll merge later today unless I hear any objections!

PR #5497, opened 2022-10-28
Removed support for apply_groups for old option groups signature

jlstevens:
This PR should be merged before the next minor release.


holoviews/core/accessors.py
```diff
@@ -566,7 +566,7 @@ def __call__(self, *args, **kwargs):
                        "group (i.e. separate plot, style and norm groups) is deprecated. "
                        "Use the .options method converting to the simplified format "
```
maximlt, line 567: I'm never sure, what do we recommend usually between calling `.opts` and `.options`?

holoviews/core/accessors.py
```diff
@@ -566,7 +566,7 @@ def __call__(self, *args, **kwargs):
...
                        "instead or use hv.opts.apply_groups for backward compatibility.")
-                param.main.param.warning(msg)
+                raise Exception(msg)
```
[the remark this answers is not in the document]
jlstevens: Good suggestion. Addressed in 89a59b85f

holoviews/core/accessors.py
jlstevens: I think the reason `.options` was recommended was because it isn't in-place and is therefore the right analogue to the old syntax.

However, we generally do recommend `.opts` so I think I agree - we should suggest `.opts` here instead of `.options`. Not entirely clear cut though!

maximlt:
The test suite is badly failing. We totally missed this call to `_dispatch_opts` in `Dimensioned.options`.
https://github.com/holoviz/holoviews/blob/89a59b85fabbd6f9c192ac3f415449611cc7ccae/holoviews/core/dimension.py#L1283

As I think it was suggested by Jean-Luc it seems that the options being broken down in options groups is used internally.

After reverting partially what we removed in https://github.com/holoviz/holoviews/pull/5497/commits/54b444cc1e2fece110d9c6909963c404b088348b I ran the test suite and saw 73 tests failing. I was surprised that the warnings were never surfaced in the test suite output (or were they???). My assumption is that since the warning was emitted with `param.warning` this means that it's actually using Python's `logging` module and not `warnings` module. I guess this type of "warning" doesn't have a special status to Pytest, it's just a log.

If this is really the case, we should make sure there's no other similar case.

maximlt:
I'm not done yet there are still failing tests. I think some of them will have to be removed as they were testing the old behavior. Some fail in ways that I didn't expect so maybe there's something more to it, let's see.

Hoxbro:
@jlstevens, as we already have a dev release of 1.16, can you take a last look at this PR and merge it when you have time?

PR #5502, opened 2022-10-31
Support jupyterlite

### LINKED ISSUE #5475: holoviews 1.15.1 doesn't work in Jupyterlite

holoviews: 1.15.1
jupyterlite: Version 0.1.0-beta.13

```python
import piplite
await piplite.install('holoviews==1.15.1')

import holoviews as hv
hv.extension('bokeh')

hv.Curve([1, 2, 3])
```

raise following exception:

```
ImportError: cannot import name 'document' from 'js' (unknown location)
```

Here is the reason:

https://github.com/jupyterlite/jupyterlite/issues/618

Hoxbro:
Fixes #5475

The solution to the problem is to use the existing `notebook_extension` and not the `pyodide_extension` inside the jupyterlite: 

image

Therefore a small check is added to see if holoviews is inside jupyterlite, and depending on that, either use `notebook_extension` or `pyodide_extension`.

### DIFF
--- a/holoviews/__init__.py
+++ b/holoviews/__init__.py
@@ -121,6 +121,8 @@ except ImportError:
 
 if '_pyodide' in sys.modules:
     from .pyodide import pyodide_extension, in_jupyterlite
+    # The notebook_extension is needed inside jupyterlite,
+    # so the override is only done if we are not inside jupyterlite.
     if not in_jupyterlite():
         extension = pyodide_extension
     del pyodide_extension, in_jupyterlite
--- a/holoviews/__init__.py
+++ b/holoviews/__init__.py
@@ -120,7 +120,10 @@ except ImportError:
             raise Exception("IPython notebook not available: use hv.extension instead.")
 
 if '_pyodide' in sys.modules:
-    from .pyodide import pyodide_extension as extension # noqa (API import)
+    from .pyodide import pyodide_extension, in_jupyterlite
+    if not in_jupyterlite():
+        extension = pyodide_extension
+    del pyodide_extension, in_jupyterlite
 
 # A single holoviews.rc file may be executed if found.
 for rcfile in [os.environ.get("HOLOVIEWSRC", ''),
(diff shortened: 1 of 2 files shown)


holoviews/pyodide.py
```diff
@@ -74,6 +74,10 @@ def render_png(element):
 def render_svg(element):
     return render_image(element, 'svg')
 
+def in_jupyterlite():
+    import js
+    return hasattr(js, "_JUPYTERLAB")
```
jlstevens, line 79: Looks a little hackish...is there no public/official way to tell what your environment is?

jlstevens:
Other than one minor comment, looks good to merge! 

(Ignoring the issue with CI for macos of course!)

holoviews/__init__.py
```diff
@@ -120,7 +120,10 @@ def __call__(self, *args, **opts): # noqa (dummy signature)
             raise Exception("IPython notebook not available: use hv.extension instead.")
 
 if '_pyodide' in sys.modules:
-    from .pyodide import pyodide_extension as extension # noqa (API import)
+    from .pyodide import pyodide_extension, in_jupyterlite
+    if not in_jupyterlite():
```
jlstevens, line 124: Maybe add a comment to say in jupyterlite the normal extension works just fine?

holoviews/pyodide.py
Hoxbro: I agree, but I can't seem to find a better way to do it. Maybe it just needs a fresh set of eyes; you can give it a try [here](https://jupyter.org/try-jupyter/lab).

jlstevens: Jupyter has always made this sort of thing needlessly difficult - I wish they would allow a public API and honestly, I'm impressed you found anything that works!

jlstevens:
I'm going to rebase now CI tests are passing on macos so that this can be the last code PR in the release...

jlstevens:
Will merge once the tests are green!

PR #5507, opened 2022-11-03
Require bokeh<3

basnijholt:
It seems like HoloViews is not compatible with Bokeh 3.0.0. Here I pin bokeh.


basnijholt:
I noticed that bokeh is pinned in panel, so this is probably not needed.

PR #5509, opened 2022-11-07
Maintenance of the code

Hoxbro:
This PR has done a couple of things to update the code:
1) Convert (most) of the format strings to f-strings. 
    - This is done automatically with [flynt](https://github.com/ikamensh/flynt) for 34169217cdf837289d0c8e8e98c8f069e9a70822 and small manually changes in  47112338b2575ba1b12da0ee7bbeef14e91a0018 and f2ce53e70bc32d954c6c11936541f191a99a8109.
2) Remove bare `except:` in bcd6cda2ad277504d805796185ffc9c40d8c9d74.
   - This is either done with `except Exception:` or `except ImportError`.  
3) Remove checks if pandas are installed in c4a94a2d9b0afb14c7d1c8fb4f36d2ba2835f2c2, as it is a required dependency.
4) Remove Python 2 code in 6dc73155cc8ca7072bafc41d941da34e5bb370c5 and 74201a9d4318d86377f1b4c389b89af02ee23a3a
5) Cleanup bokeh imports in 223eea1f56c8a72b13b06de2b12b46725cc1319d.
    - `MercatorTileSource`, `cartesian_to_axial`, `built_in_themes` was in Bokeh 1.4.0

I noticed that file `...interface` does not exist, so I think the following lines should be updated or removed if no longer needed. 
https://github.com/holoviz/holoviews/blob/e5224c6a278829ca6165ac04b36da6d5f55825cf/holoviews/plotting/bokeh/__init__.py#L19-L22

### DIFF
--- a/holoviews/__init__.py
+++ b/holoviews/__init__.py
@@ -140,7 +140,7 @@ for rcfile in [os.environ.get("HOLOVIEWSRC", ''),
             try:
                 exec(code)
             except Exception as e:
-                print("Warning: Could not load %r [%r]" % (filename, str(e)))
+                print(f"Warning: Could not load {filename!r} [{str(e)!r}]")
         del f, code
         break
     del filename
--- a/holoviews/annotators.py
+++ b/holoviews/annotators.py
@@ -224,7 +224,7 @@ class Annotator(PaneBase):
         super().__init__(None, **params)
         self.object = self._process_element(object)
         self._table_row = Row()
-        self.editor = Tabs(('%s' % param_name(self.name), self._table_row))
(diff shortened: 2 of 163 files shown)


philippjfr:
>I noticed that file ...interface does not exist, so I think the following lines should be updated or removed if no longer needed.

Yes please remove, it no longer exists. Since we now require Python >= 3.7 we could also consider swapping OrderedDict for regular dicts but that can be a separate PR.

philippjfr (approved):
I see that `DFrame` import was already removed so this looks good to me. Really appreciate this, thanks!

PR #5513, opened 2022-11-18
Add support to Python 3.11

maximlt:
Python 3.11 removed `inspect.getargspec` that was deprecated since Python 3.0. HoloViews no longer used it, in favor of `inspect.getfullargspec`. However, HoloViews still used directly the named tuple `ArgSpec` returned by that function, that was also removed in Python 3.11 and caused the breakage. There was no warning for that removal, probably as `ArgSpect` (and the newer `FullArgSpec`) aren't part of the Python public API.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -133,7 +133,7 @@ jobs:
             pip install pyctdev
         - name: doit develop_install
           run: |
-            doit ecosystem=pip develop_install -o flakes -o tests -o examples_tests
+            doit ecosystem=pip develop_install -o flakes -o tests_core
         - name: doit env_capture
           run: |
             doit ecosystem=pip env_capture
@@ -143,6 +143,3 @@ jobs:
         - name: doit test_unit
           run: |
             doit ecosystem=pip test_unit
-        - name: test examples
-          run: |
-            doit ecosystem=pip test_examples
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -34,7 +34,8 @@ jobs:
       fail-fast: false
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        # python-version: ['3.7', '3.8', '3.9', '3.10']
+        python-version: ['3.11']
     timeout-minutes: 90
     defaults:
       run:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -106,3 +106,42 @@ jobs:
         run: |
           conda activate test-environment
           codecov
+
+  test_pip:
+      name: Pip tests on ${{ matrix.os }} with Python ${{ matrix.python-version }}
+      runs-on: ${{ matrix.os }}
+      strategy:
+        fail-fast: false
+        matrix:
+          os: ['ubuntu-latest']
+          python-version: ["3.11"]
+      steps:
+        - name: Checkout source
+          uses: actions/checkout@v3
+          with:
+            fetch-depth: 0
+        - name: Install Python ${{ matrix.python-version }}
+          uses: actions/setup-python@v4
+          with:
+            python-version: ${{ matrix.python-version }}
+        - name: Update setuptools
+          run: |
+            pip install --upgrade setuptools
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yaml
```diff
@@ -34,7 +34,8 @@ jobs:
       fail-fast: false
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
-        python-version: ['3.7', '3.8', '3.9', '3.10']
+        # python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
```
Hoxbro, line 37: Should I uncomment this line, or do we want to be more selective? Like `['3.7', '3.9', '3.11']`

maximlt: Yep 5 of them is a lot :( !

We thought about this for Panel and I suggested to always include the bounds (of course) and not run the CI on one of the latest versions as they are the ones that are less used. In hindsight maybe what we want is actually test more and versions that are less used, as for the other ones it's likely we'll get user reporting problems.

So, well, I don't know :) But this isn't very important in the end to be honest. And whatever subset we pick, we should have the weekly cron job run on the full matrix.

Pick whatever subset you prefer!

holoviews/tests/plotting/plotly/test_distributionplot.py
```diff
@@ -1,10 +1,19 @@
+from unittest import SkipTest
+
 from holoviews.element import Distribution
 
 from .test_plot import TestPlotlyPlot
 
 
 class TestDistributionPlot(TestPlotlyPlot):
 
+    def setUp(self):
```
maximlt, line 10: Instead shouldn't we add `scipy` to the `tests_core` dependencies? Seems like it's pretty important to test the Distribution elements as part of the unit test suite.

setup.py
```diff
@@ -46,18 +49,22 @@
...
     'ffmpeg',
     'cftime',
     'scipy',
     'selenium',
-    'ipython >=5.4.0',
 ]
 
+# Packages not working on python 3.11 becauase of numba
+if sys.version_info < (3, 11):
+    extras_require['tests'] += [
```
maximlt, line 63: Alright yes I guess it's the best option. Let's not forget to remove that as soon as we can!

maximlt:
> I started from the changes suggested in #5511, but they did not pass the test suite. So I iterated on it.
> 
> Mainly there were two issues.
> 
>     1. `FullArgSpec`´s `varkw` is `ArgSpec`' s `keywords`.
> 
>     2. `FullArgSpec` have more information in its `namedtuple` than `ArgSpec`. The extra information is not needed.
> 
> 
> The solution to the two issues could also have been to change to FullArgSpec, but I found it less intrusive to do it this way.

Ok thanks for the details! I guess for now this is enough to get things to work. But when I see that `FullArgSpec` contains `kwonlyargs` (keyword-only parameters) I suspect that HoloViews `util.argspec` doesn't not cover all the possible signature forms offered by Python 3. If you think so too, let's open an issue about that.

holoviews/tests/plotting/plotly/test_distributionplot.py
Hoxbro: We use `test` and not `test-core` for our test suite, so we test the `scipy` in our CI. 

In the debug mode, `test_core` was used and gave an error. I think it is acceptable not to have `scipy` in `test-core`.

setup.py
Hoxbro: I will open a new PR with the revert as soon as this PR is merged!

.github/workflows/test.yaml
Hoxbro: I have now added `['3.7', '3.9', '3.11']`. Should I take a look at the cron job in a new PR?

holoviews/tests/plotting/plotly/test_distributionplot.py
maximlt: Ok thanks I agree.

setup.py
maximlt: Great thanks!

.github/workflows/test.yaml
maximlt: Yep let's leave that to a new PR. Btw I think we should run the full matrix when a release is made too.

PR #5530, opened 2022-11-30
Adding numba-dependent packages back for python 3.11

hoxbro:
This should be merged when numba which supports python 3.11 is released.


maximlt:
Just noting that in this PR we should try to run the examples tests on Python 3.11, as numba might be the only thing that wasn't compatible (or not!).

Issue #5533, opened 2022-12-02
Intermitent error when running a module


maximlt:
I have observed a very strange bug. It occurs intermittently when the test suite of Panel runs, when the `doc/user_guide/APIs.md` file is being executed. I've extracted its code in a module, and I run this module with Python directly, so not in a Notebook context and not in a Bokeh server context.

```python
import hvplot.pandas
from bokeh.sampledata.autompg import autompg

def autompg_plot(x='mpg', y='hp', color='#058805'):
    return autompg.hvplot.scatter(x, y, c=color, padding=0.1)

columns = list(autompg.columns[:-2])
autompg_plot()
import panel as pn
pn.extension()
def fn(a,b): return f'Arguments: {a,b}'
slider = pn.widgets.FloatSlider(value=0.5)

bound_fn = pn.bind(fn, a=slider, b=2)
bound_fn()
pn.Row(slider, bound_fn)
x = pn.widgets.Select(value='mpg', options=columns, name='x')
y = pn.widgets.Select(value='hp', options=columns, name='y')
color = pn.widgets.ColorPicker(name='Color', value='#AA0505')

pn.Row(pn.Column('## MPG Explorer', x, y, color),
       pn.bind(autompg.hvplot.scatter, x, y, c=color))
x = pn.widgets.Select(value='mpg', options=columns, name='x')
y = pn.widgets.Select(value='hp', options=columns, name='y')
color = pn.widgets.ColorPicker(name='Color', value='#AA0505')

@pn.depends(x, y, color)
def plot(xval, yval, colorval):
    return autompg.hvplot.scatter(xval, yval, c=colorval)

pn.Row(
    pn.Column('## MPG Explorer', x, y, color),
    plot
)
color = pn.widgets.ColorPicker(name='Color', value='#4f4fdf')
layout = pn.interact(autompg_plot, x=columns, y=columns, color=color)

pn.Row(pn.Column('## MPG Explorer', layout[0]), layout[1])
import param

class MPGExplorer(param.Parameterized):

    x = param.Selector(objects=columns)
    y = param.Selector(default='hp', objects=columns)
    color = param.Color(default='#0f0f0f')

    @param.depends('x', 'y', 'color') # optional in this case
    def plot(self):
        return autompg_plot(self.x, self.y, self.color)

explorer = MPGExplorer()

pn.Row(explorer.param, explorer.plot)
x = pn.widgets.Select(value='mpg', options=columns, name='x')
[...]

Issue #5430, opened 2022-09-07
Repeated calls to Overlay, Curve, and .opts() causes StoreOptions to fail


dycw:
#### ALL software version info

Python 3.9, with a simple `Makefile`:

```
.ONESHELL:
SHELL=/bin/bash
.PHONY: env
env:
        [ -d .venv ] && rm -rf .venv/
        PYENV_VERSION=3.9.13 python -m venv --clear .venv
        ! [ -L activate ] && ln -s .venv/bin/activate
        source activate
        pip install --upgrade numpy holoviews hypothesis pip pre-commit pytest
```

Run with `make env`.

#### Description of expected behavior and the observed behavior

Not to crash. Instead, `hypothesis` marks this as a flaky test: one that fails initially, but then succeeded upon shrinking.

#### Complete, minimal, self-contained example code that reproduces the issue

```
# test_plots.py

from holoviews import Curve, Overlay
from holoviews.plotting import bokeh  # type: ignore # noqa
from hypothesis import given, settings
from hypothesis.extra.numpy import arrays
from hypothesis.strategies import data, integers, lists


@given(data=data())
@settings(deadline=None, max_examples=10000)
def test_main(data):
    n = data.draw(integers(1, 5))
    arrs = data.draw(lists(arrays(float, n), min_size=1, max_size=5))

    _ = Overlay(items=[Curve(arr) for arr in arrs])  # this will succeed

    if True:  # this will fail
        _ = Overlay(items=[Curve(arr).opts(show_grid=True) for arr in arrs])

    if False:  # this will fail
        _ = Overlay(items=[Curve(arr).opts(tools=["hover"]) for arr in arrs])
```

#### Stack traceback and/or browser JavaScript console output

Run with `pytest test_plots.py`:

```
E           hypothesis.errors.Flaky: Hypothesis test_main(data=data(...)) produces unreliable results: Falsified on the first call but did not on a subsequent one
E           Falsifying example: test_main(
E               data=data(...),
E           )
E           Draw 1: 3
E           Draw 2: [array([6.10351562e-05, 6.10351562e-05, 6.10351562e-05])]
E           Failed to reproduce exception. Expected:
E           data = data(...)
E
E               @given(data=data())
E               @settings(max_examples=1000)
E               def test_main(data):
E                   n = data.draw(integers(1, 5))
E                   arrs = data.draw(
E                       lists(
E                           arrays(float, n, elements=floats(-1.0, 1.0)),
E                           min_size=1,
[...]

PR #5535, opened 2022-12-02
Fix potential race condition in the Options system

maximlt:
Fixes https://github.com/holoviz/holoviews/issues/5430
Fixes https://github.com/holoviz/holoviews/issues/5533

We think that this small change avoids a race condition in the Options system where the garbage collector cleans up a `weakref` in this line:

```python
max_id = max(store_ids)+1 if len(store_ids) > 0 else 0
```

It should not cause any other issues, because this would have been used when Python 2 was supported, at the time `.keys()` was returning a list. We may encounter this sort of issue in other places, if we don't handle weakrefs properly?

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -1787,7 +1787,7 @@ class StoreOptions(object):
         """
         max_ids = []
         for backend in Store.renderers.keys():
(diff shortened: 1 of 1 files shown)


maximlt:
Failing on 3.11 for unrelated reasons, merging.

PR #5539, opened 2022-12-02
When sorting colormap records, replace None with empty string

### LINKED ISSUE #5538: Issue: Listing colormap records sometimes raises TypeError
Calling `holoviews.plotting.list_cmaps` with `records=True` sometimes raises `TypeError`:

``` python
In [1]: from holoviews.plotting import list_cmaps

In [2]: cmaps = list_cmaps(records=True)
  Cell In [2], line 1
    cmaps = list_cmaps(records=True)
  File ...\holoviews\plotting\util.py:794 in list_cmaps
    return list(unique_iterator(python2sort(matches,
  File ...\holoviews\core\util.py:1128 in unique_iterator
    for item in seq:
  File ...\holoviews\core\util.py:1211 in <genexpr>
    return itertools.chain.from_iterable(sorted(group, key=key) for group in groups)
TypeError: '<' not supported between instances of 'str' and 'NoneType'
```

(I've abbreviated the traceback's paths.)

The error arises when `sorted` tries to order `None` and a `str`, which occurs because the sort key contains `None` from the `bg` or `source` elements of some `CMapInfo` objects.

For some (but few, in my tests) runs of *Python*, the call succeeds, presumably because the lower-level `_list_cmaps` function returns the colormap records in such an order that `sorted` needs to compare only `str` elements of the sort key.

The version of *holoviews* is v1.15.3a2.

stanwest:
Closes #5538.

### DIFF
--- a/holoviews/plotting/util.py
+++ b/holoviews/plotting/util.py
@@ -791,8 +791,10 @@ def list_cmaps(provider=None, records=False, name=None, category=None, source=No
 
     # Return results sorted by category if category information is provided
     if records:
-        return list(unique_iterator(python2sort(matches,
-                    key=lambda r: (r.category.split(" ")[-1],r.bg,r.name.lower(),r.provider,r.source))))
+        return list(unique_iterator(python2sort(
+            matches,
+            key=lambda r: (r.category.split(" ")[-1], r.bg or "", r.name.lower(), r.provider, r.source or "")
+        )))
     else:
         return list(unique_iterator(sorted([rec.name for rec in matches], key=lambda n:n.lower())))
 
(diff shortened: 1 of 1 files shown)


stanwest:
If the sort key must be a tuple of strings to avoid causing the `python2sort` function to raise `TypeError`, then one may use built-in sorting instead.  Furthermore, the sorted result need not go through `unique_iterator`, because the sorted items came from a set and are already unique.  Finally, the outer `list` may be removed, because the built-in `sorted` returns a list.  I'm pushing an additional commit with this suggestion.

jlstevens:
Thanks for the PR! 

At a quick glance this looks it should be fine but it needs a closer look before merging.

Hoxbro (approved):
LGTM

Thank you for your contribution :1st_place_medal:

stanwest:
> Thanks for the PR!

> Thank you for your contribution 🥇

You're welcome, and thanks for merging.

PR #5540, opened 2022-12-03
Check for categorical data to histogram

MarcSkovMadsen:
Addresses [hvplot/#998](https://github.com/holoviz/hvplot/issues/998) by providing an error message if the user tries to create a `histogram` from categorical data.

Before the user would get the error message `numpy.core._exceptions._UFuncNoLoopError: ufunc 'multiply' did not contain a loop with signature matching types (dtype('<U1'), dtype('float64')) -> None` 

This is not an error message that a new user to python and plotting would be able to make sense of. It takes a long time to realize that theoretically a histogram is for numerical data only. For example users can find plotly express examples of creating "histograms" from categorical data. In their documentation its the second example you see. I think they realized this is a very common use case.


MarcSkovMadsen:
The tests fails on Python 3.11 only and I believe its not related to this PR. This should be ready.

jlstevens:
@MarcSkovMadsen Change makes good sense to me. Hopefully the tests will now pass (including 3.11) after the merge commit at which point I am happy to see this merge.

jlstevens:
Looks like all tests are now green!

Issue #5168, opened 2021-12-16
Incorrect behavior of DynamicMap returning Overlay of certain Element types


TheoMathurin:
Versions:
holoviews 1.14.6
panel 0.12.4
bokeh 2.4.1

I encountered this bug both in notebooks and in a bokeh server.

When a DynamicMap driven by a MultiChoice or MultiSelect returns an overlay of elements such as Ellipse, Box or Bounds, there is a strange behavior where the glyphs that get rendered do not correspond to the inputs. Sometimes even the number of elements is incorrect. No error are raised in Python or in JavaScript.

Note that using a NdOverlay instead solves the issue for some reason.

See code and corresponding screen cast below

```
import holoviews as hv
import panel as pn
hv.extension('bokeh')

def make_ellipses_overlay(radii):
    if radii:
        return hv.Overlay([hv.Ellipse(0, 0, 2*radius) for radius in radii])
    else:
        return hv.Overlay([hv.Ellipse(0, 0, 0)])
    
def make_ellipses_ndoverlay(radii):
    if radii:
        return hv.NdOverlay({radius: hv.Ellipse(0, 0, 2*radius) for radius in radii})
    else:
        return hv.NdOverlay({0: hv.Ellipse(0, 0, 0)})

radii_select = pn.widgets.MultiChoice(options=[2, 4, 8], width=180)
ellipses_overlay = hv.DynamicMap(pn.bind(make_ellipses_overlay, radii=radii_select)).opts(data_aspect=1, xlim=(-5, 5), ylim=(-5, 5)).opts(title='Overlay: incorrect behavior')
ellipses_ndoverlay = hv.DynamicMap(pn.bind(make_ellipses_ndoverlay, radii=radii_select)).opts(data_aspect=1, xlim=(-5, 5), ylim=(-5, 5)).opts(title='NdOverlay: correct behavior')

pn.Row(ellipses_overlay, ellipses_ndoverlay, radii_select)

```

https://user-images.githubusercontent.com/49269671/146362499-33d7c03a-97a0-4f45-a0ff-4837d8f513c9.mp4

jbednar:
That's a scary bug! I  can reproduce it without clicking on anything in a Jupyter notebook if I first plot it:

```python
import holoviews as hv, panel as pn
hv.extension('bokeh')

def make_ellipses_overlay(radii):
    if radii:
        return hv.Overlay([hv.Ellipse(0, 0, 2*radius) for radius in radii])
    else:
[...]

PR #5555, opened 2022-12-09
Ensure dynamically created subplots can be updated

philippjfr:
This was a very bad bug that definitely should have been caught.

Fixes https://github.com/holoviz/holoviews/issues/5168


jlstevens:
Thanks for the fix!

As you say this was resulting in very confusing behavior which my explain why I've always reached for `NdOverlays` when returning from DynamicMaps.

Will merge when the tests pass for inclusion in the immediate follow up release.

jlstevens:
Tests passed. Merging now as this is an important fix.

PR #5563, opened 2022-12-15
Enhance Tiles example notebook

sandhujasmine:
- added a note to the lon_lat_to_easting_northing() function used for coordinate transformation from (long, lat) to web mercator
- remove deprecated tile sources when plotting so we don't have empty plots and JS errors

### DIFF
--- a/examples/reference/elements/bokeh/Tiles.ipynb
+++ b/examples/reference/elements/bokeh/Tiles.ipynb
@@ -54,6 +54,13 @@
    "source": [
     "One thing to note about tile sources is that they are always defined in the [pseudo-Mercator projection](https://epsg.io/3857), which means that if you want to overlay any data on top of a tile source the values have to be expressed as eastings and northings. If you have data in another projection, e.g. latitudes and longitudes, it may make sense to use [GeoViews](http://geoviews.org/) for it to handle the projections for you.\n",
     "\n",
+    ">\n",
+    "> **Note**: \n",
+    "> \n",
+    ">   Holoviews provides functions to project longitude, latitude into Web Mercator coordinates. \n",
+    ">   See [`hv.util.transform.lon_lat_to_easting_northing(longitude, latitude)`](https://holoviews.org/reference_manual/holoviews.element.html?highlight=lon_lat_to_easting_northing#holoviews.element.Tiles.lon_lat_to_easting_northing)\n",
+    ">\n",
+    "\n",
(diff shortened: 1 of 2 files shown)


examples/reference/elements/bokeh/Tiles.ipynb
```diff
@@ -79,7 +86,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "hv.Layout([ts().relabel(name) for name, ts in hv.element.tiles.tile_sources.items()]).opts(\n",
+    "valid_src = {k: v for k, v in hv.element.tiles.tile_sources.items() if v.__name__ != 'deprecated_tilesource_warning'}\n",
```
jlstevens, line 89: Maybe we should have a variable in `hv.element.tiles` of the tile sources that aren't being deprecated?

jlstevens:
Thanks for the PR!

I made one suggestion... I think we should have a full tile lists (as we currently do) but also one of tiles that we think are supported without needing deprecation warnings...

holoviews/element/tiles.py
```diff
@@ -188,6 +191,8 @@ def wikimedia_replacement():
 OpenTopoMap = lambda: Tiles('https://a.tile.opentopomap.org/{Z}/{X}/{Y}.png', name="OpenTopoMap")
 Wikipedia = wikimedia_replacement
 
-tile_sources = {k: v for k, v in locals().items() if isinstance(v, FunctionType) and k not in
+_all_tile_sources = {k: v for k, v in locals().items() if isinstance(v, FunctionType) and k not in
                 ['ESRI', 'lon_lat_to_easting_northing', 'easting_northing_to_lon_lat',
                  'deprecation_warning', 'wikimedia_replacement']}
+
+tile_sources = {k: v for k, v in _all_tile_sources.items() if v.__name__ != 'deprecated_tilesource_warning'}
```
jlstevens, line 198: Should be ok to do this, especially given the comment you added. Definitely makes the notebook cleaner!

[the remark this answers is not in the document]
jlstevens: Looks good to me!

As a sanity check I would grep the notebooks and source to see if `tile_sources` was used anywhere else where the deprecated tiles might have been expected (I don't think so). Otherwise, I'm happy with this approach.

jlstevens:
Looks ready and so I'll merge. If you have any follow on suggestions or changes, please open another PR.

PR #5580, opened 2023-01-02
Update copyright to only contain start year

Hoxbro:
It will be easier if we only include the start year in the license file. An explanation of why can be seen [here](https://hynek.me/til/copyright-years/)


jlstevens:
I'm happy with this approach. As no one had any strong objection when we discussed this, I'll go ahead and merge.

PR #5584, opened 2023-01-06
Fix hv.Empty not working in AdjointLayout plot

### LINKED ISSUE #5559: Trying to add marginal histogram to just one axis
I am in need to add a histogram to one axis similar to the question posted in https://discourse.holoviz.org/t/how-to-add-a-histogram-to-just-the-x-axis/133/1

I was trying to reproduce the code recommendation but failing the same as others in the thread. The suggestion was to report it in GitHub but not able to find previous issues.

Hoxbro:
Fixes #5559

Small regression, which made the following code error. 

``` python
import holoviews as hv
import numpy as np

hv.extension("bokeh")

scatter = hv.Scatter(np.random.rand(100, 2))
scatter << hv.Empty() << scatter.hist(adjoin=False)

```
image

### DIFF
--- a/holoviews/core/layout.py
+++ b/holoviews/core/layout.py
@@ -62,8 +62,8 @@ class Empty(Dimensioned, Composable):
 
     group = param.String(default='Empty')
 
-    def __init__(self):
-        super().__init__(None)
+    def __init__(self, **params) -> None:
+        super().__init__(None, **params)
 
 
 
--- a/holoviews/core/layout.py
+++ b/holoviews/core/layout.py
@@ -62,7 +62,7 @@ class Empty(Dimensioned, Composable):
 
     group = param.String(default='Empty')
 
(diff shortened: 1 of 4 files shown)


holoviews/core/layout.py
```diff
@@ -62,8 +62,8 @@ class Empty(Dimensioned, Composable):
 
     group = param.String(default='Empty')
 
-    def __init__(self):
-        super().__init__(None)
+    def __init__(self, **params) -> None:
```
philippjfr, line 65: We have not started adding type hints to HoloViews. I'm in favor but that should be a project level decision.

```suggestion
    def __init__(self, **params):
```

jlstevens:
Looks good to me! 

Happy to merge once there are some unit tests. Docs updates can be in a separate PR.

PR #5587, opened 2023-01-11
Add deserialization of non-deserialized base64 data in `CDSCallback._process_msg`

Hoxbro:
Fixes https://github.com/holoviz/geoviews/issues/584

The data in the `msg` `CDSCallback._process_msg` received had not been deserialized in some cases. This PR adds a deserialization step.

Values for byte-ordering were found [here](https://github.com/bokeh/bokeh/blob/7f0ebed4702e719485c6eafe857307f195bffa87/src/bokeh/core/serialization.py#L145).

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -1,4 +1,5 @@
 import asyncio
+import base64
 import time
 
 from collections import defaultdict, OrderedDict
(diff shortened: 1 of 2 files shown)


philippjfr:
>Values for byte-ordering were found here.

Just FYI byte ordering can only ever be big or little, see [Endianness](https://en.wikipedia.org/wiki/Endianness).

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -924,6 +925,14 @@ def _process_msg(self, msg):
                 values = new_values
             elif any(isinstance(v, (int, float)) for v in values):
                 values = [np.nan if v is None else v for v in values]
+            elif isinstance(values, list) and len(values) == 4 and isinstance(values[3], list):
```
philippjfr, line 928: Don't think this condition is quite sufficient as this could in theory occur when receiving polygon or path data. Probably need one more condition like `values[2] in ('big', 'little')`

Hoxbro:
Yes. 

This was more of a comment toward Numpy's other byte-order characters [xref](https://numpy.org/doc/stable/reference/generated/numpy.dtype.byteorder.html).

PR #5588, opened 2023-01-11
Ibis v4 support

hoxbro:
Ibis v4.0 was released a couple of days ago, which makes our CI fail if we don't pin it. See [here](https://github.com/holoviz/holoviews/actions/runs/3891636637/jobs/6642118177)

This PR fixes two exceptions and accommodates FutureWarning emitted.


jlstevens:
Looks good to me and good to see that Ibis is being actively developed again!

PR #5598, opened 2023-01-23
use matplotlib's public api to list the colormaps

### LINKED ISSUE #5476: external matplotlib colormaps cannot be found
```
holoviews: 1.15.1
matplotlib: 3.6.0
cmocean: 2.0
```

I've been trying to use the colormaps provided by `cmocean` in a new environment, and got an error stating that the colormap I requested couldn't be found.

It seems that the newest version of `matplotlib` removed `matplotlib.cm._cmap_registry`, which was used here:
https://github.com/holoviz/holoviews/blob/f051576deb8d21ec845a71cfd6f334af5362d64b/holoviews/plotting/util.py#L693-L696

Making `_cmap_registry` an alias of `_colormaps` allows me to use the external colormaps:
```python
import matplotlib.cm as cm

cm._cmap_registry = cm._colormaps
```

However, I don't think that's a long-term solution since both are internal API, so maybe use `list(matplotlib.colormaps)` instead?

cc @tacaswell, @QuLogic

```python
import cmocean
from holoviews.plotting import list_cmaps

list_cmaps(provider="matplotlib")
```

keewis:
- [x] closes #5476

This uses `matplotlib.colormaps` to get the colormaps (if it is available) instead of `matplotlib.cm._cmap_registry`, which was removed in `matplotlib=3.6` without warning because it was internal API.

### DIFF
--- a/holoviews/plotting/util.py
+++ b/holoviews/plotting/util.py
@@ -699,8 +699,12 @@ def _list_cmaps(provider=None, records=False):
 
     if 'matplotlib' in provider:
         try:
+            import matplotlib as mpl
             from matplotlib import cm
-            if hasattr(cm, '_cmap_registry'):
+
+            if hasattr(mpl, "colormaps"):
+                mpl_cmaps = list(mpl.colormaps)
+            elif hasattr(cm, '_cmap_registry'):
                 mpl_cmaps = list(cm._cmap_registry)
             else:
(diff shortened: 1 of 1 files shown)


Hoxbro:
This looks good to me. 

A small thing - can you add your comment in the code about the matplotlib version in the code (for the future)?

Hoxbro:
Thank you for the PR :+1:

PR #5606, opened 2023-01-30
Add convenience function to Opts

Hoxbro:
Inspired by [this](https://discourse.holoviz.org/t/can-i-read-what-opts-have-been-applied-to-my-plot/4872) discussion on discourse and my own experience when trying to find the values set in `Opts`. 

image

I'm not sure if `__setitem__` should be added, as it could add more confusion.

### DIFF
--- a/holoviews/core/accessors.py
+++ b/holoviews/core/accessors.py
@@ -649,3 +649,20 @@ class Opts(metaclass=AccessorPipelineMeta):
 
         kwargs['clone'] = False if clone is None else clone
         return self._obj.options(*new_args, **kwargs)
+
+    def __getitem__(self, item):
+        options = self.get().kwargs
+        if item in options:
+            return options[item]
+        else:
+            raise KeyError(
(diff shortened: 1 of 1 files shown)


philippjfr:
Definitely happy to provide `__getitem__`, however we have to consider `__setitem__` carefully. Generally HoloViews objects should not be modified inplace and the same really extends to their options. The concern around this is many fold:

- Multiple copies of an Element may point to the same underlying `Options` object. If we modify the Options inplace all copies will reflect the changed values.
- Allowing inplace modification can give users the expectations that the options respond reactively, i.e. you might expect that `.opts['color'] = 'red'` updates all current plots to be red when it really just affects future plots AND if something happened to trigger a re-render of the plot.

Hoxbro:
I already was on the fence about `__setitem__` and will remove it.

jlstevens:
Just to echo what is said, getting options this way is fine, setting them this way is certainly more questionable...

PR #5609, opened 2023-02-03
Apply autorange on data change events

philippjfr:
- Had to schedule the `CustomJS` callback using `setTimeout` because the `ColumnDataSource.data` callback fires before the glyphs (and more importantly their indexes) are updated.
- We remove all usage of `cb_obj` in the callback which allows us to reuse the callback between both data and range updaates.
- We add the autorange callback to a new `_js_on_data_callbacks` instance variable which then gets applied to all sources attached to a `GlyphRenderer`

## ToDo

- [x] Check inverted axes work with this
- [x] Guard against failed lookups in `Bokeh.index`
- [x] Guard the `view.child_views` lookup to avoid errors in layouts
- [x] Somehow filter out GlyphRenderers which were generated by plots with apply_ranges=False (suggest using `.tags` property to mark them)

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -11,7 +11,7 @@ from bokeh.core.properties import value
 from bokeh.document.events import ModelChangedEvent
 from bokeh.models import (
     BinnedTicker, ColorBar, ColorMapper, CustomJS, EqHistColorMapper,
-    Legend, Renderer, Title, tools,
+    GlyphRenderer, Legend, Renderer, Title, tools,
 )
 from bokeh.models.axes import CategoricalAxis, DatetimeAxis
 from bokeh.models.formatters import (
@@ -221,6 +221,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
 
         # Whether axes are shared between plots
         self._shared = {'x': False, 'y': False}
+        self._js_on_data_callbacks = []
 
         # Flag to check whether plot has been updated
         self._updated = False
@@ -985,7 +986,6 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         else:
             p0, p1 = self.padding, self.padding
 
-
         if dim == 'x':
             lower, upper = self.xlim
             lower = None if (lower is None) or np.isnan(lower) else lower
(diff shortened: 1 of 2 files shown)


jlstevens:
I think holoviews is doing what I asked for here even though the behavior is a little crazy!

flipping_inverted

jbednar:
> I think holoviews is doing what I asked for here even though the behavior is a little crazy!

That doesn't seem right at all. inverting should be about how that particular plot is displayed, not any other plot...

philippjfr:
>That doesn't seem right at all. inverting should be about how that particular plot is displayed, not any other plot...

Linked axes are shared axes, so there's no way to avoid this.

jbednar:
> Linked axes are shared axes, so there's no way to avoid this.

I'd say that there's no _current_ way. :-) I'd consider this a bug, since I think the intent of the user is clear here: share the range extents, don't share the orientation. Addressing it seems very difficult, so maybe there can be a warning saying if `invert_{xy}axis` must be set to the same value on all linked axes (so that users realize they need to set it on both or neither, with the current implementation).

jlstevens:
@philippjfr I've finished off the tasks mentioned in your TODO list - the tags suggestion works great! Ready for review/merge once the test go green.

holoviews/plotting/bokeh/element.py
```diff
@@ -997,10 +1000,13 @@ def _setup_autorange(self):
...
         # Clean this up in bokeh 3.0 using View.find_one API
-        self.state.js_on_event('rangesupdate', CustomJS(code=f"""
-          const ref = cb_obj.origin.id
+        callback = CustomJS(code=f"""
+        const cb = function() {{
+          const ref = plot.id
+
           const find = (view) => {{
-            for (const sv of view.child_views) {{
+            let iterable = view.child_views === undefined ? []  : view.child_views
```
philippjfr, line 1008: One thing I missed is that it would be nice if this could traverse through Tabs which I think don't have child_views directly (although they might).

philippjfr: Seems like that should work.

jlstevens: Tried the following test with tabs and autoranging is still working!

image

philippjfr (commented):
Happy to see this merged. Since this is my PR I can't approve, so feel free to approve yourself and merge.

jlstevens:
I'll add a line to the user guide about auto ranging applying after data updates and then I'll merge.

jlstevens:
I'll go ahead and merge.

PR #5610, opened 2023-02-03
Add sorting of imports

hoxbro:
This PR adds sorting of the import.  

- cf2d73009bc8f0144c3b0dd1503bf1502e2f075a removed conditional imports below the "normal" import block and also removed old IPython version checks.
- 5c282c2b01e93ca4730e284e53906f04460b3c0a removed the sensitivity of the `..util` import by giving it a namespace alias.
- 6a76250824b91c3aed856dd7241f4817586ef1f9 is running the sorting.

### DIFF
--- a/holoviews/__init__.py
+++ b/holoviews/__init__.py
@@ -120,7 +120,7 @@ except ImportError:
             raise Exception("IPython notebook not available: use hv.extension instead.")
 
 if '_pyodide' in sys.modules:
-    from .pyodide import pyodide_extension, in_jupyterlite
+    from .pyodide import in_jupyterlite, pyodide_extension
     # The notebook_extension is needed inside jupyterlite,
     # so the override is only done if we are not inside jupyterlite.
     if not in_jupyterlite():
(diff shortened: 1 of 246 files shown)


jlstevens:
Looks good but I don't like it when imports get split onto multiple lines: in those cases, that suggests there are enough things being imported that a namespace should be used instead.

There are a few places where this happens for inline imports, but I'm happy to let those slide (as long as they are rare!)

philippjfr:
Thanks for doing this. I don't much like the one import per link behavior  and prefer the behavior in Panel, i.e.:

```python
from .image import (  # noqa
    GIF, ICO, JPG, PDF, PNG, SVG,
)
```

vs.

```python
from .image import (  # noqa
    GIF,
    ICO,
    JPG,
    PDF,
    PNG,
    SVG,
)
```

MarcSkovMadsen:
Just go with isort and black. Do what they do.

Metallica: Nothing else matters :-)

philippjfr:
Going to close since this PR can easily be re-applied and the conflicts make this unusable in the current state.

PR #5612, opened 2023-02-06
Add simple option to tox.ini to fix PyPi build

Hoxbro:
The Pypi build is failing because `flakes` has been removed. This adds a simple option that only installs holoviews with no optional dependencies.


Hoxbro:
This fixes the PyPi build, see: https://github.com/holoviz/holoviews/actions/runs/4111435632

PR #5621, opened 2023-02-10
Implement multi-yaxis support

philippjfr:
This adds multi-yaxis support to the Bokeh backend.

Screen Shot 2023-02-10 at 16 21 11

# ToDo

- [x] Clean up utilities
- [x] Add autorange support for multi-axes 
- [x] Fix linking of shared axes
- [x] Implement axis labels correctly
- [x] Handle correct logx/logy mappings
- [x] Correctly determine categorical axis creation per axis
- [x] Clean up `_get_factors`
- [ ] Handle dynamic axis creation when DynamicMap OverlayPlot adds new elements

### DIFF
--- a/examples/user_guide/Customizing_Plots.ipynb
+++ b/examples/user_guide/Customizing_Plots.ipynb
@@ -698,9 +698,9 @@
     "### Twin axes\n",
     "*(Available in HoloViews > 1.17)*\n",
     "\n",
-    "HoloViews now supports displaying overlays containing two different value dimensions as twin axes for chart elements. To maintain backwards compatibility, this feature is only enabled by setting the `multi_y=True` option on the overlay:\n",
+    "HoloViews now supports displaying overlays containing two different value dimensions as twin axes for chart elements. To maintain backwards compatibility, this feature is only enabled by setting the `multi_y=True` option on the overlay.\n",
     "\n",
-    "To illustrate, here is an overlay containing three curves with two value dimensions ('A' and 'B'). The following shows the default, backwards compatible behavior:"
+    "To illustrate, here is an overlay containing three curves with two value dimensions ('A' and 'B'). Setting `multi_y=True` then maps these two value dimensions to twin-axes:"
    ]
   },
   {
@@ -710,7 +710,7 @@
    "outputs": [],
    "source": [
     "overlay = hv.Curve([1, 2, 3], vdims=['A']) * hv.Curve([2, 3, 4], vdims=['A']) * hv.Curve([3, 2, 1], vdims=['B'])\n",
-    "overlay"
+    "overlay.opts(multi_y=True)"
    ]
   },
   {
@@ -719,24 +719,6 @@
    "source": [
     "Additional value dimensions do map to additional axes but be aware that support of multi axes beyond twin axes is currently considered experimental.\n",
     "\n",
-    "#### Setting `multi_y=True`\n",
-    "\n",
-    "We can now enable twin axes by setting `multi_y=True` on the overlay:"
-   ]
-  },
-  {
-   "cell_type": "code",
-   "execution_count": null,
-   "metadata": {},
-   "outputs": [],
-   "source": [
-    "overlay.opts(multi_y=True)"
-   ]
-  },
-  {
-   "cell_type": "markdown",
-   "metadata": {},
-   "source": [
     "The first value dimension is mapped to the left-hand axis and the second value dimension maps to the right axis. Note that the two axes are individually zoomable by hovering over them and using the Bokeh wheelzoom tool.\n",
     "\n",
     "\n",
--- a/examples/user_guide/Customizing_Plots.ipynb
+++ b/examples/user_guide/Customizing_Plots.ipynb
@@ -44,7 +44,8 @@
     "    - [Inverting axes](#inverting-axes): Flipping the x-/y-axes and inverting an axis\n",
     "    - [Axis labels](#axis-labels): Setting axis labels using dimensions and options\n",
     "    - [Axis ranges](#axis-ranges): Controlling axes ranges using dimensions, padding and options\n",
-    "    - [Axis ticks](#axis-ticks): Controlling axis tick locations, labels and formatting"
+    "    - [Axis ticks](#axis-ticks): Controlling axis tick locations, labels and formatting\n",
+    "    - [Twin axes](#twin-axes): Enabling twin axes"
    ]
   },
   {
@@ -688,6 +689,88 @@
    "outputs": [],
    "source": [
     "bars.opts(xrotation=45)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Twin axes\n",
+    "*(Available in HoloViews > 1.17)*\n",
+    "\n",
+    "HoloViews now supports displaying overlays containing two different value dimensions as twin axes for chart elements. To maintain backwards compatibility, this feature is only enabled by setting the `multi_y=True` option on the overlay:\n",
+    "\n",
+    "To illustrate, here is an overlay containing three curves with two value dimensions ('A' and 'B'). The following shows the default, backwards compatible behavior:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "overlay = hv.Curve([1, 2, 3], vdims=['A']) * hv.Curve([2, 3, 4], vdims=['A']) * hv.Curve([3, 2, 1], vdims=['B'])\n",
+    "overlay"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "tags": []
+   },
+   "source": [
+    "Additional value dimensions do map to additional axes but be aware that support of multi axes beyond twin axes is currently considered experimental.\n",
+    "\n",
+    "#### Setting `multi_y=True`\n",
+    "\n",
+    "We can now enable twin axes by setting `multi_y=True` on the overlay:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "overlay.opts(multi_y=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "The first value dimension is mapped to the left-hand axis and the second value dimension maps to the right axis. Note that the two axes are individually zoomable by hovering over them and using the Bokeh wheelzoom tool.\n",
+    "\n",
+    "\n",
+    "#### Supported `multi_y` options\n",
+    "\n",
+    "When `multi_y` is enabled, you can set individual axis options on the elements of the overlay.\n",
+    "\n",
+    "In this example, the left axis uses the default options while the right axis is an inverted, autoranged, log axis with a set `ylim`:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
+    "(hv.Curve([1, 2, 3], vdims=['A']) * hv.Curve([2, 3, 4], vdims=['B']).opts(autorange='y', invert_yaxis=True, logy=True, ylim=(1,10))).opts(multi_y=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Supported options for customizing individual axes are `apply_ranges`, `autorange='y'`, `invert_yaxis`, `logy` and `ylim`.\n",
+    "\n",
+    "Note that as of HoloViews 1.17.0, `multi_y` does not have streaming plot support"
    ]
   }
  ],
--- a/examples/user_guide/Customizing_Plots.ipynb
+++ b/examples/user_guide/Customizing_Plots.ipynb
@@ -706,9 +706,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "tags": []
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "overlay = hv.Curve([1, 2, 3], vdims=['A']) * hv.Curve([2, 3, 4], vdims=['A']) * hv.Curve([3, 2, 1], vdims=['B'])\n",
@@ -717,9 +715,7 @@
   },
   {
    "cell_type": "markdown",
-   "metadata": {
-    "tags": []
-   },
+   "metadata": {},
    "source": [
     "Additional value dimensions do map to additional axes but be aware that support of multi axes beyond twin axes is currently considered experimental.\n",
     "\n",
@@ -731,9 +727,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "tags": []
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "overlay.opts(multi_y=True)"
@@ -756,9 +750,7 @@
   {
    "cell_type": "code",
    "execution_count": null,
-   "metadata": {
-    "tags": []
-   },
+   "metadata": {},
    "outputs": [],
    "source": [
     "(hv.Curve([1, 2, 3], vdims=['A']) * hv.Curve([2, 3, 4], vdims=['B']).opts(autorange='y', invert_yaxis=True, logy=True, ylim=(1,10))).opts(multi_y=True)"
@@ -781,5 +773,5 @@
   }
  },
  "nbformat": 4,
- "nbformat_minor": 2
+ "nbformat_minor": 4
 }
--- a/examples/user_guide/Customizing_Plots.ipynb
+++ b/examples/user_guide/Customizing_Plots.ipynb
@@ -762,7 +762,7 @@
    "source": [
     "Supported options for customizing individual axes are `apply_ranges`, `autorange='y'`, `invert_yaxis`, `logy` and `ylim`.\n",
     "\n",
-    "Note that as of HoloViews 1.17.0, `multi_y` does not have streaming plot support"
+    "Note that as of HoloViews 1.17.0, `multi_y` does not have streaming plot support and that linked streams are not yet aware of additional y-axes."
    ]
   }
  ],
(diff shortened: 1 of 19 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -927,46 +980,50 @@ def _update_ranges(self, element, ranges):
...
+        if shared:
+            shared = (axis_range.start, axis_range.end)
+            low, high = util.max_range([(low, high), shared])
+        if invert: low, high = high, low
+        if not isinstance(low, util.datetime_types) and log and (low is None or low <= 0):
+            low = 0.01 if high > 0.01 else 10**(np.log10(high)-2)
+            self.param.warning(
+                "Logarithmic axis range encountered value less "
+                "than or equal to zero, please supply explicit "
+                "lower-bound to override default of %.3f." % low)
```
jbednar, line 1013:
```suggestion
                "lower bound to override default of %.3f." % low)
```

jbednar (commented):
Very cool! How does a user configure this behavior? I think we should introduce it as an option at first since it's a change in behavior that some people may previously have been relying on (e.g. by defining dimensions only on the first plot in an overlay, rather than (correctly) defining them on all elements of the overlay).

philippjfr:
@jlstevens and I discussed this and I am still strongly of the opinion that we should never enable this by default. Right now you enable this with the `multi_y` option and I think we should keep that. What we discussed as an alternative is that if you have multiple clashing `vdims` we will simply put all of the vdim labels on the axis (probably separated by a pipe).

My concern with enabling it by default continues to be that it is likely to be much more annoying than useful. Particularly when creating plots directly from DataFrames or xarray you will often overlay things with different column/variable names and a slight mismatch shouldn't result in multiple axes.

jbednar:
Thanks; I didn't spot `multi_y` in the PNG. Sounds good.

jlstevens:
I think I am happy with `multi_y` as an additional parameter: it gates the functionality preserving backwards compatibility. Then, to make things more convenient again (but backwards incompatible) for HoloViews 2.0 we can simply set the default to `True`.

philippjfr:
As I said above and suggested to you, I vote strongly never to enable it by default.

jlstevens:
> As I said above and suggested to you, I vote strongly never to enable it by default.

As I have no strong opinion this is also fine by me! That said,  I also think other people may disagree and that this is something to discuss again when the time comes. Specifying an additional flag is more cumbersome and inconvenient though in this case, I could easily be convinced that it is being explicit is the overriding factor.

jbednar:
We'll see! I'll try setting it on for my workflows, and see what problems it causes in practice. Meanwhile, no need to decide about future possible changes to defaults.

orozcojd:
Is this expected to work with Datashaded plots?

philippjfr:
Yes, absolutely.

jbednar:
> Was wondering if there are plans to include some option to configure an holoviews.Area element with multiple axes such that colors are swapping when y1 and y2 referenced to different axes are crossing.

@GeoVizNow, that sounds unrelated to having multiple y axes, and seems like it should be doable already by overlaying multiple hv.Area plots cropped in comparison to a reference curve. Seems like a discourse.holoviz.org topic to propose and see if someone can fill in an example.

holoviews/plotting/mixins.py
```diff
@@ -223,3 +235,26 @@ def _get_coords(self, element, ranges, as_string=True):
...
+class MultiDistributionMixin:
+
+    def _get_axis_dims(self, element):
+        return element.kdims, element.vdims[0]
+
+    def get_extents(self, element, ranges, range_type='combined', **kwargs):
+        l,b,r,t =  super().get_extents(
+            element, ranges, range_type, 'categorical', element.vdims[0]
+        )
+        return b,l,t,r  # WHY IS THIS NEEDED FOR BOXWHISKER AND VIOLIN?
```
philippjfr, line 249: Indeed, definitely something odd here.

holoviews/plotting/bokeh/element.py
```diff
@@ -1251,10 +1252,19 @@ def _init_glyph(self, plot, mapping, properties):
             del properties['legend_label']
 
         # ALERT: This only handles XYGlyph types right now
-        if 'x' in mapping and mapping['x'] in plot.extra_x_ranges:
-            properties['x_range_name'] = mapping['x']
-        if 'y' in mapping and mapping['y'] in plot.extra_y_ranges:
-            properties['y_range_name'] = mapping['y']
+        if 'x' in mapping:
+            x = mapping['x']
+            if isinstance(x, Field):
```
Hoxbro, line 1257: This was added in Bokeh 3. There is in the utility folder the following function `property_to_dict`

holoviews/plotting/bokeh/element.py
```diff
@@ -1241,6 +1241,10 @@ def _init_glyph(self, plot, mapping, properties):
         """
         Returns a Bokeh glyph object.
         """
+        try:
+            from bokeh.transform import Field
+        except ImportError:
+            from bokeh.core.property.vectorization import Field
```
philippjfr, line 1248: Why is this inlined?

jlstevens: Was a quick test to see if it is the right thing. Bokeh 3 tests are now passing but bokeh 2 is failing pretty badly and it shouldn't be because the imports were inlined here.

Dunno if the Bokeh 2 test failures are related but I agree that these imports shouldn't be inlined once we got Bokeh 2 passing again.

jlstevens: It *might* well be related, as in 2025b5d, bokeh 2 tests were *mostly* passing except for the extra fixes that worked in Bokeh 3 but made Bokeh 2 fail a lot more. Maybe these two `Field` classes aren't really interchangeable?

jlstevens:
Bokeh 3 tests are now green! :-)

However, something since 2025b5d has broken Bokeh 2 badly (which now hangs. My suspicion is that  it could be due to the `Field` import but I could be wrong...

holoviews/plotting/bokeh/element.py
```diff
@@ -1256,16 +1252,13 @@ def _init_glyph(self, plot, mapping, properties):
             del properties['legend_label']
 
         # ALERT: This only handles XYGlyph types right now
+        mapping = property_to_dict(mapping)
         if 'x' in mapping:
```
jlstevens, line 1256: Just realized that this was introduced as a fix after unit tests were already passing so the fact that unit tests are passing locally with this change means nothing. I suppose if the unit tests for Bokeh 2 start passing that will at least be somewhat informative...

jlstevens: Bokeh 2 unit tests are now passing. The question now is whether the test that was failing in examples remains fixed with this change. At least we can see that the issue with Bokeh 2 is related to this code.

holoviews/plotting/bokeh/element.py
jlstevens: Issue is moot now I am using `property_to_dict` and the tests are passing.

jlstevens:
@philippjfr I consider df81780e9 a first cut at supporting log for multi axes (not happy with it yet). Other than cleanup/refactoring where exactly this logic lives, is this right general approach? 

My main worry though is having to remove `logy` from `_propagate_options` as that will affect backwards compatibility...

jlstevens:
The code now supports all the key features I expect from this PR. While the axis support is in bokeh's `ElementPlot` and while this supports more than two (i.e. twin) axes, at this stage I want to document and add unit tests for the following twin-axis use cases for `Chart` elements (this is the immediate and common need). 

Here are examples of the cases this PR officially supports starting with the following element definitions. I plan to use the notebook containing these examples for the docs and to add unit tests:

### Element definitions (note the different vdims)

```python
orange_scatter = hv.Scatter([9, 8, 7, 6, 5], vdims='A').opts(color='orange')
blue_scatter = hv.Scatter([1, 2, 3], vdims='B').opts(color='blue')
orange_scatter + blue_scatter
```

image

### Backwards compatible behavior by default

Without `multi_y`, overlays behave the same as before:

image

### Twin axes using `multi_y=True`

First axis goes left, second goes right

image

### Linked twin axes in layouts

linked_twin

### Independent autoranging per y-axis

Note the autoranging on data is also per-axis.

#### Autoranging left

autorange_left

#### Autoranging right

autorange_right

#### Autoranging both

autorange_both

### Respecting `apply_ranges=False` (per axis)

no_apply_range

### Respecting `ylim` (per axis)

image


### Respecting `ylim` (per axis) with `autorange='y'`

both_ylim_autorange

## What isn't working for twin axes that should

- [ ] Setting `yaxis='right'` and `yaxis='left'` should let you flip the axes round: right now the axes do flip but the labels do not.


@jbednar @philippjfr Ignoring the full generality of the multi-axis case, I think this is a decent enough set of features to make interactive twin axes useful. Do you agree? Is there any important feature missing?

jbednar:
Looks great to me! 

I'm not sure I 100% agree with the "Autoranging left" and "Autoranging right" behaviour; seems like auto-range on y should kick in when zooming on _x_ or on the whole plot, _not_ when explicitly zooming on _y_. I.e., auto-ranging is meant to be helpful, not prohibitive -- it should automatically choose a good range for y when the datapoints being shown have changed, _not_ prevent you from zooming on the _y_ axis at all! But I'm pretty sure that is a complaint unrelated to this PR. 

If you agree with that claim, please open a separate issue about the y autoranging, as I think this PR is good already.

jlstevens:
I have now modified option propagation in a way that should only change behavior when an overlay has `multi_y=True`, leaving everything else the same. This seems to be working and is easy enough to reason about, at least when it comes to seeing that backwards compatibility is preserved in general.

In addition, I've now implemented support for per-axis `invert_y`:

### `invert_yaxis` on the left
image

### `invert_yaxis` on the right

image

It also works with autoranging:

### `invert_yaxis` on the left with autoranging

invert_autorange_left

### `invert_yaxis` on the right with autoranging

invert_autorange_right

@jbednar 

> But I'm pretty sure that is a complaint unrelated to this PR.

That is correct: I see your point about the autoranging behavior but what this PR does is apply the existing autoranging behavior to twin axes. Per axis, the autoranging behavior you see is the same as what it was already. I'll file an issue once this PR is wrapped up.

Feature-wise, the main issue I see right now is the `yaxis` position is not respected to flip the axes correctly. While I think I have a fix for the axes in my stash, getting the axis labels working looks like it is going to be a fair bit more involved to get right. While this is an issue, I'm not sure this should hold up the PR as you can trivially achieve what you want by swapping the components of the overlay instead.

Now twin axes support all the basic y-axis options, I'm now going to add unit tests for what is supported so far.

jlstevens:
While testing with DynamicMaps, I see an issue with linked streams. Here is an example:

dmap_twin

Note that despite the different in the axis scales, `tap1.y` and `tap2.y` are the same even though the position of the cursor corresponds to two different places in data space, according to the axis the stream should be attached to. Talking to @philippjfr, it seems all streams conveying information about y axes will need updating to handle `multi_y`. This isn't going to be trivial as `cb_obj` is a `plot` model which doesn't make it easy to access extra y axes and figure out the correct data coordinates associated with the respective element/axis.

As this would a fair bit of work, I suggest this is tackled in a separate PR. If we need to release before this can be addressed, we can state that this is experimental in the release notes (saying streams involving extra y axis information are not yet supported and that streaming data is not supported). Even with these limitations, streams involving the x-axis are useful (e.g. LTTB only needs information from `RangeX` and we LTTB is useful for curves in a twin axis configuration).

jlstevens:
After fixing an issue with extent initialization of `OverlayPlot`, the following example that doesn't use y-axis based streams now works:

lttb_dmap_twin

In addition, I now warn when `multi_y=True` and streams are being used with unreliable y-axis data values.

At this point, I think fixing the position option (and corresponding labels) is something that I will warn about as being unsupported for now and can be tackled in a separate PR.

jlstevens:
The current issues that I see right now are:

1. Many streams do not work properly but at least they will warn that they do not yet work with twin axes.
2. (Low priority) No support for streaming, can be documented as such
3. The biggest problem happens when trying to switch the position of the axes around:

Since c4db54bdc, this is the behavior (to document it here):

image

The axes themselves are right but the wrong label is on the left (instead of 'A' it should be 'B') and the label on the right is missing entirely. I tried issuing a warning based on the `yaxis` option value but that ended up needing a partial solution that didn't address the bad label behavior at all. At this point, I would consider just documenting this as one of the limitations of twin axes that will be addressed later.

jlstevens:
@philippjfr Here is what I think the plan should be:

- I'll add some xfail tests as you suggest
- This PR gets reviewed
- I push a commit which I believe improves the position handling. I'm not entirely happy with the code and so I would appreciate review of this particular commit. It doesn't fix the labels but I think it is functionally an improvement (even if I might need some help refactoring the code it a bit)
- We can then decide whether to keep or revert that commit.
- Merge!

jlstevens:
@philippjfr @Hoxbro I've added docs and some xfail tests.

Ready for review/merge. 

I've got what I think is a partial fix for the yaxis position handling but I think it will just confuse the diff - I'll either make a new branch off this one if it hasn't been merged yet or a new one off `main` if it has to propose this fix (and hopefully figure out a fix for labels as well).

jlstevens:
Adding a note as a reminder that we will need a proper fix for the new dask release and to unpin in setup.py!

holoviews/plotting/bokeh/element.py
```diff
@@ -841,15 +955,41 @@ def _update_grid(self, plot):
...
+        streaming = False
+        multi_dim = 'x' if self.invert_axes else 'y'
+        for axis_dim, extra_y_range in self.handles[f'extra_{multi_dim}_ranges'].items():
+            _, b, _, t = self.get_extents(element, ranges, dimension=axis_dim)
+            factors = self._get_dimension_factors(element, ranges, axis_dim)
+            extra_scale = self.handles[f'extra_{multi_dim}_scales'][axis_dim] # Assumes scales and ranges zip
+            log = isinstance(extra_scale, LogScale)
+            self._update_range(
+                extra_y_range, b, t, factors,
+                extra_y_range.tags[1]['invert_yaxis'] if extra_y_range.tags else False,
```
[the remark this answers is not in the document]
jlstevens: No as when there are tags at all, the specs are in the first position (position zero) and the remaining axis-specific information (i.e. the dict) is in the second position.

holoviews/plotting/bokeh/element.py
```diff
@@ -2330,6 +2502,11 @@ class OverlayPlot(GenericOverlayPlot, LegendPlot):
                           'margin', 'aspect', 'data_aspect', 'frame_width',
                           'frame_height', 'responsive', 'fontscale']
 
+    def __init__(self, overlay, **kwargs):
+        self._multi_y_propagation = self.lookup_options(overlay, 'plot').options.get('multi_y', False)
+        super().__init__(overlay, **kwargs)
+        self._multi_y_propagation = False
```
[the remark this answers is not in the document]
jlstevens: To be safe: the change from the default behavior is when `self._multi_y_propagation = True` so I immediately set it to False afterwards. I can try removing it and it *should* be ok but I cannot be 100% sure.

holoviews/plotting/bokeh/util.py
```diff
@@ -1000,14 +1001,22 @@ def match_ax_type(ax, range_type):
...
     else:
         return range_type in ('auto', 'log')
 
 
+def match_yaxis_type_to_range(yax, range_type, range_name):
+    "Apply match_ax_type to the y-axis found by the given range name "
+    for axis in yax:
+        if axis.y_range_name == range_name:
+            return match_ax_type(axis, range_type)
+    raise Exception('No axis with given range found')
```
philippjfr, line 1017:
```suggestion
    raise ValueError('No axis with given range found')
```

holoviews/plotting/bokeh/element.py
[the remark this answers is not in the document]
jlstevens: I suppose so, for these particular models. But isn't that assumption reasonable for these models created by HoloViews?

holoviews/plotting/bokeh/element.py
```diff
@@ -141,6 +185,8 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         be specified as a two-tuple of the form (vertical, horizontal)
         or a four-tuple (top, right, bottom, left).""")
 
+    multi_y = param.Boolean(default=False)
```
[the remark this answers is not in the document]
jlstevens:
```suggestion
    multi_y = param.Boolean(default=False, doc="""
       Enables multiple axes (one per value dimension) in
       overlays and useful for creating twin-axis plots.

       When enabled, axis options are no longer propagated between the
       elements and the overlay container, allowing customization on a
       per-axis basis.""")
```

holoviews/plotting/bokeh/element.py
```diff
@@ -339,149 +391,125 @@ def _get_hover_data(self, data, element, dimensions=None):
...
+                                dim_range = extra_range
+
+        if dim_range and not (range_type is FactorRange and not isinstance(dim_range, FactorRange)):
+            shared_name = extra_range_name or ('x-main-range' if pos == 0 else 'y-main-range')
+            self._shared[shared_name] = True
+
+        return dim_range
+
+    def _axis_props(self, plots, subplots, element, ranges, pos, *, dim=None,
+                    range_tags_extras=[], extra_range_name=None):
```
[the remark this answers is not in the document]
philippjfr: Let's clean these up once we add a linter check.

holoviews/plotting/bokeh/element.py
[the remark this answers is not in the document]
philippjfr: I think for now it's reasonable if not 100% safe.

holoviews/plotting/plot.py
```diff
@@ -1379,7 +1407,7 @@ def _get_range_extents(self, element, ranges, range_type, xdim, ydim, zdim):
 
         return (x0, y0, x1, y1)
 
-    def get_extents(self, element, ranges, range_type='combined', xdim=None, ydim=None, zdim=None):
+    def get_extents(self, element, ranges, range_type='combined', dimension=None, xdim=None, ydim=None, zdim=None):
```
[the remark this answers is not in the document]
philippjfr: Let's make a concerted effort to move to keyword only signatures at some point.

examples/user_guide/Customizing_Plots.ipynb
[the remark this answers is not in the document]
jlstevens: Addressed in 55d2be324

Hoxbro:
As discussed with @philippjfr, this PR is in a good state and can be merged, and then I will tag a new dev version. 

Thank you for all your work here, @jlstevens! I know how you have loved every single second of bringing this fantastic functionality to life!

jlstevens:
> I know how you have loved every single second of bringing this fantastic functionality to life!

Of course! :-)

GeoVizNow:
Thank you for working up this great new feature!

PR #5623, opened 2023-02-16
Replace "three" with "four" for consistency

michaelaye:
There are four different types of axes discussed now, so this should be "four" instead of "three".


Hoxbro:
Thank you

Issue #5613, opened 2023-02-06
holoviews.Dataset.transform(flag=expr) and link_selections.selection_expr = expr not working with 'spatial_select' dimension expression when dataframe index differs from 0-based step one index


GeoVizNow:
Thank you for building the great stack of software in the holoviz ecosystem! I recently came across something that appears as a bug or at least a change in functionality compared to the holoviews 1.14.5 version that I was previously using. I am now at 1.15.2. 

#### ALL software version info
os: Windows 10 Enterprise
pandas: 1.5.2
holoviews: 1.15.2
numpy: 1.23.5
bokeh: 2.4.3
datashader: 0.14.3

#### Description of expected behavior and the observed behavior
When applying a spatial_select in a dimension expression to a holoviews.Dataset using dataset.transform(**{'name: expr}).dframe(..) it is expected that it would work correctly independent of what are the input pandas dataframe index values. Now, in holoviews 1.15.2, my experience is that the .transform is working erroneously/unexpected when given a pandas dataframe where the index deviates from a zero-based 1-incrementing monotonous index. 
#### Imports
```
import pandas as pd
import numpy as np
import holoviews as hv
from holoviews import opts
from holoviews.selection import link_selections
from holoviews.operation.datashader import datashade
from io import StringIO
hv.extension('bokeh')
```
#### Working example:
```
# Prepare test data
data_text = (
    """
         idx_fail| idx_ok| name|   a|    b
              101|      0|   P1|  7.|   3.
              102|      1|   P2|  3.|   4.
              103|      2|   P3|  .5|   3.
              104|      3|   P4|  2.|   2.
              105|      4|   P5|  1.|   2.
              106|      5|   P6|  1.|   1.
    """)

# read in data using zero based index
df = pd.read_csv(StringIO(data_text), sep='|', skipinitialspace=True, index_col=1)
df

ls = link_selections.instance(unselected_alpha=.2)
points = hv.Points(df, kdims=['a', 'b']).opts(size=26, fill_color='yellow', line_color='black')
labels = hv.Labels(df, kdims=['a', 'b'], vdims=['name'])  # .opts(fontsize={'labels':'7pt'})
plot = datashade(points, x_sampling=1, y_sampling=.5).opts(padding=.3)

arr = np.array(
    [
        [3.0, 1.7],
        [0.3, 1.7],
        [0.3, 2.7],
        [3.0, 2.7]
    ]
)

[...]

PR #5625, opened 2023-02-18
For closed issue #5613. Add test to ensure spatial_select to dataset.transform from non-zero indexed dataframe is applied to right indexes 

GeoVizNow:
I added a test for this as it recently failed in 1.15.2 as described in issue 5613. Then, when running the updated 1.15.4, I realized there was not a problem in that version. So probably accounted for in 1.15.3 or 1.15.4.

Even though I am uncertain if there is any test to fetch this problem if reappearing and hence the following test. I realize I do not follow the setup with tests in classes, but hope you will consider the content of the test anyhow. And potentially you can recommended how you would like it modified.

### DIFF
--- a/holoviews/tests/util/test_transform.py
+++ b/holoviews/tests/util/test_transform.py
@@ -526,7 +526,6 @@ class TestDimTransforms(ComparisonTestCase):
 
 def test_dataset_transform_by_spatial_select_expr_index_not_0_based():
     """Ensure 'spatial_select' expression works when index not zero-based.
-    
     Use 'spatial_select' defined by four nodes to select index 104, 105.
     Apply expression to dataset.transform to generate new 'flag' column where True
     for the two indexes."""
--- a/holoviews/tests/util/test_transform.py
+++ b/holoviews/tests/util/test_transform.py
@@ -4,8 +4,13 @@ Unit tests for dim transforms
 import pickle
 import warnings
 
+import holoviews as hv
+from io import StringIO
+hv.extension('bokeh')
+
 from collections import OrderedDict
 from unittest import skipIf
+import pytest
 
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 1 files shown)


GeoVizNow:
By the way I should have mentioned one experience: I had troubles getting the command 
`doit develop_install -c pyviz/label/dev -c conda-forge -o tests -o examples`
to run when trying to make an editable install. Eventually I found (by googling) that if pinning setuptools to version '63.4.3' in the 'pyproject.toml', then the editable install went fine.

holoviews/tests/util/test_transform.py
```diff
@@ -4,8 +4,13 @@
 import pickle
 import warnings
 
+import holoviews as hv
+from io import StringIO
+hv.extension('bokeh')
```
Hoxbro, line 9: `hv.extension` should not be run in the test suite, and it isn't needed for your test to pass.

holoviews/tests/util/test_transform.py
```diff
@@ -520,3 +525,56 @@ def test_pickle(self):
...
+        101|   P1|  7.|   3.
+        102|   P2|  3.|   4.
+        103|   P3|  .5|   3.
+        104|   P4|  2.|   2.
+        105|   P5|  1.|   2.
+        106|   P6|  1.|   1.
+        """)
+
+    # read in data using non-zero based index
+    return pd.read_csv(StringIO(data_text), sep='|', skipinitialspace=True, index_col=0)
```
Hoxbro, line 550: I think you have overengineered this. You could get a similar result by having `pd.DataFrame({"a": [7, 3, 0.5, 2, 1, 1], "b": [3, 4, 3, 2, 2, 1]}, index=range(101, 107))`.

holoviews/tests/util/test_transform.py
```diff
@@ -520,3 +525,56 @@ def test_pickle(self):
...
+ 
+    
+
+def test_dataset_transform_by_spatial_select_expr_index_not_0_based(df_fixture):
```
Hoxbro, line 554: The df_fixture should be removed, and create the DataFrame inside this test.

holoviews/tests/util/test_transform.py
```diff
@@ -520,3 +525,56 @@ def test_pickle(self):
...
+    """Ensure 'spatial_select' expression works when index not zero-based.
+    
+    Use 'spatial_select' defined by four nodes to select index 104, 105.
+    Apply expression to dataset.transform to generate new 'flag' column where True
+    for the two indexes."""
+    arr = np.array(
```
Hoxbro, line 560: This variable could be called geometry.

Hoxbro (changes requested):
Thank you for the test. I have left some comments, but in general, it looks good. :slightly_smiling_face:

holoviews/tests/util/test_transform.py
GeoVizNow: Sounds good, I remove it.

holoviews/tests/util/test_transform.py
GeoVizNow: I could modify this if you prefer. I just started to appreciate to see tabular input in test where the arrangment matters.

holoviews/tests/util/test_transform.py
GeoVizNow: Ok, then I will remove the fixture.

holoviews/tests/util/test_transform.py
GeoVizNow: Thanks, I will rename it.

holoviews/tests/util/test_transform.py
Hoxbro: I would very much prefer it. It takes focus away from the actual test code.

GeoVizNow: Then I sure will remove it.

GeoVizNow:
> Thank you for the test. I have left some comments, but in general, it looks good. 🙂

Thank you for reviewing, appreciate your guidance here :-)

GeoVizNow:
I think the test is adjusted to the better now.

jbednar:
Thanks for adapting it!

Hoxbro (approved):

Hoxbro:
> Thank you for reviewing, appreciate your guidance here :-)

No, thank you for taking the time to make this PR :+1:

PR #5630, opened 2023-03-01
Fix formatting on FAQ

wendrul:


Hoxbro:
Thank you for the fix :+1:

PR #5631, opened 2023-03-01
Fix legend display when using categorical datashade on GPU

### LINKED ISSUE #5619: Categorical datashade does not display legend if using GPU
This was originally reported on datashader as https://github.com/holoviz/datashader/issues/1178. There were actually 2 bugs, one in datashader and one in holoviews. The datashader bug is fixed in https://github.com/holoviz/datashader/pull/1179 so that it no longer raises an error. To reproduce the behaviour here it will be necessary to use a datashader branch containing that fix.

Minimal reproducer here:
```python
import datashader as ds
import holoviews as hv
import pandas as pd
hv.extension('bokeh')

df = pd.DataFrame(dict(
    x    = [0.0, 1.0, 0.0, 1.0, 0.0],
    y    = [0.0, 1.0, 1.0, 0.0, 0.5],
    cat  = ['a', 'b', 'a', 'b', 'a'],
))
df.cat = df.cat.astype("category")

if 1:
    import cudf
    df = cudf.DataFrame.from_pandas(df)

curve = hv.Curve(df)
datashade(curve, aggregator=ds.by("cat"))
```
If you run this code without the `cupy` block and then with the `cupy` block, these are the 2 outputs you see:
cpu gpu

So a legend is displayed if you are using `numpy` but not if using `cupy`. I presume the problem is similar to the datashader problem in the area of `cupy` array conversions which are mostly very similar to `numpy` but not always.

ianthomas23:
Fixes #5619.

When attempting a categorical datashade using a `cudf.DataFrame` the legend was not displayed. This was caused by the following line
https://github.com/holoviz/holoviews/blob/029ba951eaefff03d59f44195b6b466f70a584d0/holoviews/plotting/util.py#L1333
which gave the error
```
TypeError: StringIndex object is not iterable. Consider using `.to_arrow()`, `.to_pandas()` or `.values_host` if you wish to iterate over the values.
```
which explains the problem and offers a solution. I have chosen to wrap the offending line in a `try` block and use the `values_host` attribute if it fails. The use of `values_host` is consistent with that in `core/data/cudf.py`.

I have tested locally using `pandas`, `dask`, `cudf` and `dask-cudf` dataframes and they all give the correct result now. I have not added a test to this PR as we don't use `cudf` in github actions CI, but if one is required and someone can direct me to where it should be located I will happily do so.

### DIFF
--- a/holoviews/plotting/util.py
+++ b/holoviews/plotting/util.py
@@ -1334,7 +1334,7 @@ class categorical_legend(Operation):
                 cats = list(hvds.data.dtypes[column].categories)
             except TypeError:
                 # Issue #5619, cudf.core.index.StringIndex is not iterable.
-                cats = list(hvds.data.dtypes[column].categories.values_host)
+                cats = list(hvds.data.dtypes[column].categories.to_pandas())
             if cats == ['__UNKNOWN_CATEGORIES__']:
                 cats = list(hvds.data[column].cat.as_known().categories)
         else:
--- a/holoviews/plotting/util.py
+++ b/holoviews/plotting/util.py
@@ -1330,7 +1330,11 @@ class categorical_legend(Operation):
             return
         column = agg.column
         if hasattr(hvds.data, 'dtypes'):
-            cats = list(hvds.data.dtypes[column].categories)
+            try:
+                cats = list(hvds.data.dtypes[column].categories)
+            except TypeError:
+                # Issue #5619, cudf.core.index.StringIndex is not iterable.
+                cats = list(hvds.data.dtypes[column].categories.values_host)
             if cats == ['__UNKNOWN_CATEGORIES__']:
                 cats = list(hvds.data[column].cat.as_known().categories)
         else:


holoviews/plotting/util.py
```diff
@@ -1330,7 +1330,11 @@ def _process(self, element, key=None):
             return
         column = agg.column
         if hasattr(hvds.data, 'dtypes'):
-            cats = list(hvds.data.dtypes[column].categories)
+            try:
+                cats = list(hvds.data.dtypes[column].categories)
+            except TypeError:
+                # Issue #5619, cudf.core.index.StringIndex is not iterable.
+                cats = list(hvds.data.dtypes[column].categories.values_host)
```
Hoxbro, line 1337:
```suggestion
                cats = list(hvds.data.dtypes[column].categories.to_pandas())
```

I like this more because it could be used for other DataFrame libraries if needed. An example is `polars` which have the same method.
https://pola-rs.github.io/polars/py-polars/html/reference/dataframe/api/polars.DataFrame.to_pandas.html

ianthomas23: Here we are dealing with some `Index` class rather than a `DataFrame` or `Series`, and I don't know what `polars` uses for indexes. But yes, it is a good idea to standardise.

PR #5633, opened 2023-03-01
Update test CI to use  Panel dev release instead of git commit

Hoxbro:


Hoxbro:
The reason why Python 3.11 fails is that the latest Dask does not support Bokeh 3, and an older version of Dask without the pin has not been released for Python 3.11. 

Example `mamba create -n tmptmptmp python=3.11 dask bokeh=3 --dry-run`

image

PR #5634, opened 2023-03-01
Improve build setup

abravalheri:
Hi @maximlt and maintainers, this PR was created in response to https://github.com/pypa/setuptools/discussions/3841.

Please feel free to ignore/close or use parts of it in other PRs.

The main idea is to simplify `setup.py` to keep only configuration that needs to be dynamically computed (and `package_dir` since it is directly related them).

Everything else that is static can be moved to `pyproject.toml`.

Notes:

- You can use `package_dir` config to logic remap package directories to other locations (as long as they are inside the same project root).
  - This should obviate the need for copying the `examples` folder.
- Newer versions of `pip` / `setuptools` deprecate direct usage of `python setup.py ...`, these might even no longer be called.
  - Therefore it makes less sense to keep checks for `develop` and `install` in `sys.argv`.
- The installation environment should already come with both `param` and `pyct` pre-installed in the correct versions (specified in `pyproject.toml [build-system]`). Therefore version checks for those dependencies can be removed.
- As long as `pip` is relatively modern (released a couple of years ago), it should be possible for a package to depend on itself+extras.
  - This is a way of avoid repeating dependencies in extras, while still specifying them statically.

(If you have an intention of using this PR, please double check if I did not miss anything in `pyproject.toml` when doing the conversion...)

(I also believe that it should be OK to remove `graft holoviews/examples` from `MANIFEST.in`, but I haven't checked that).

---

Please note that when installing in editable mode some static analysis tools like vscode code intel and mypy may have trouble to find the package. This happens because static analysis tools may - understandably - have trouble to deal with dynamic parts of the Python language, such as `MetaPathFinder` and `PathEntryFinder`.

To workaround that you can use `pip install -e . --config-settings editable_mode=compat` for the installation. Note however that the package `holoviews.examples` will not be accessible for imports if the project is installed in that way.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,8 +2,178 @@
 requires = [
     "param >=1.7.0",
     "pyct >=0.4.4",
-    "setuptools >=30.3.0",
+    "setuptools>=61.2",
 ]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "holoviews"
+readme = "README.md"
+authors = [{name = "Jean-Luc Stevens and Philipp Rudiger", email = "holoviews@gmail.com"}]
+maintainers = [{name = "HoloViz Developers", email = "developers@pyviz.org"}]
+license = {text = "BSD"}
+description = "Stop plotting your data - annotate your data and let it visualize itself."
+classifiers = [
+    "License :: OSI Approved :: BSD License",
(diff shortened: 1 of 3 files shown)


hoxbro:
Thank you for the PR. I did use part of it when upgrading our build setup.

PR #5388, opened 2022-08-12
Compatibility changes with bokeh 3.0

Hoxbro:
For now, this is only done to be able to run panel tests in https://github.com/holoviz/panel/pull/3752. 

With these changes, it is also possible to run the following with Bokeh version 3.0:
``` python
import holoviews as hv
hv.extension("bokeh")
```

I have been pretty brutal and removed what is not needed, it could be added back in the future.

Widgetbox does not exist anymore, see https://github.com/bokeh/bokeh/issues/9334.

PR #5640, opened 2023-03-06
Merge tools in grid and layout plots

philippjfr:
In bokeh3 the tool merging is a lot more flexible but by default it is also a lot more conservative. The basic idea behind the new approach is:

> If the properties of the tools match then merge them otherwise don't.

There is a list of properties that are excluded from this logic which covers the 'renderer' and 'overlay' properties by default. Looking at a specific example we see this:

Screen_Shot_2023-03-06_at_11 34 41

- `BoxZoomTool` does not merge because the geo plot has set `match_aspect`
- `WheelZoomTool` does not merge because the geo plot has disabled  `zoom_on_axis`  (because the axes aren't rendered)
- `HoverTool` does not merge because the `tooltips` don't match
- The box select and lasso select tools don't merge because they have different `js_property_callbacks`

With this PR applied we avoid also skip `match_aspect`, `js_property_callbacks` and `zoom_axis` and I'm also considering adding `tooltips` to the list:

Screen Shot 2023-03-06 at 11 59 54

### DIFF
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -587,7 +587,7 @@ class GridPlot(CompositePlot, GenericCompositePlot):
                         sizing_mode=self.sizing_mode,
                         toolbar_location=self.toolbar)
         plot = self._make_axes(plot)
-        if bokeh3:
+        if bokeh3 and hasattr(plot, "toolbar"):
             plot.toolbar = merge_tools(plots)
 
         title = self._get_title_div(self.keys[-1])
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -639,10 +639,7 @@ class GridPlot(CompositePlot, GenericCompositePlot):
             if self.shared_yaxis:
                 x_axis.margin = (0, 0, 0, 50)
                 r1, r2 = r1[::-1], r2[::-1]
-            if bokeh3:
-                plot = gridplot([r1, r2], toolbar_location=None)
-            else:
-                plot = gridplot([r1, r2])
+            plot = gridplot([r1, r2])
         elif y_axis:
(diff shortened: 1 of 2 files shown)


philippjfr:
After some internal discussion I've decided the logic that makes most sense for us is to always merge tools of the same type  **unless** they define what I've called `disambiguation_properties`, which by default include `name`, `icon`, `tags`, and `description`.

holoviews/plotting/bokeh/plot.py
```diff
@@ -582,11 +582,13 @@ def initialize_plot(self, ranges=None, plots=[]):
             else:
                 passed_plots.append(None)
 
-        toolbar_location = None if bokeh3 else self.toolbar
```
philippjfr, line 585: @Hoxbro Do you remember why this was needed?

Hoxbro: See [this](https://github.com/holoviz/holoviews/pull/5388#issuecomment-1305969977) comment and the next.

philippjfr: Fixed the logic in https://github.com/holoviz/holoviews/pull/5640/commits/4a0b16ee21c615d831c04c6044af70c0137a2496

Hoxbro:
61f7c6364ad6ed5b40cccd959b26f6e2afad55bd is to remove toolbars for nested gridplots:

``` python
import holoviews as hv
hv.extension("bokeh")

data = {(i, j): hv.Curve(range(i + j)) for i in range(1, 3) for j in range(2, 4)}
hv.GridSpace(hv.GridSpace(data) + hv.GridSpace(data))
```

| Before | After |
| ---------- | --------|
| image | Screenshot 2023-03-08 12 35 46 |

philippjfr:
Might not be completely done but this is a huge improvement.

PR #5641, opened 2023-03-06
Update Param methods and bump pins to >=1.12,<3.0

maximlt:
No longer use `param.params()` and `param.defaults()` that have been deprecated and may be removed in Param 2.0.

Waiting for an answer to https://github.com/holoviz/param/issues/692#issuecomment-1456271626 to make sure I used the right thing to replace `param.params()`.


Hoxbro:
Because it is somewhat related, can you clean up this function in this PR?

https://github.com/holoviz/holoviews/blob/b1414e93375b989bfe880527be43f7b673e441b2/setup.py#L150-L169

maximlt:
Well hopefully it's going to go away soon when we move to pyproject.toml and hopefully find a suitable replacement for `param.version`. Does it currently raise a warning or something?

Hoxbro:
> Does it currently raise a warning or something?

No, I thought this was a good PR to clean up some old code.

maximlt:
@Hoxbro this is ready to be merged. I don't feel there's a need to spend time cleaning up the version function, hopefully it will all go away when we move to hatch.

holoviews/plotting/bokeh/element.py
```diff
@@ -561,7 +561,8 @@ def _plot_properties(self, key, element):
         if self.border is not None:
             for p in ['left', 'right', 'top', 'bottom']:
                 plot_props['min_border_'+p] = self.border
-        lod = dict(self.param.defaults().get('lod', {}), **self.lod)
+        defaults = {k: v.default for k, v in self.param.objects().items()}
+        lod = dict(defaults.get('lod', {}), **self.lod)
```
Hoxbro, line 565:
```suggestion
        lod = dict(self.params["lod"].default, **self.lod) if "lod" in self.param else self.lod
```

I don't think a complete dict lookup of `param.objects()` is needed here.

holoviews/plotting/plot.py
```diff
@@ -300,7 +300,14 @@ def selection_display(self, obj):
         return getattr(plt_class, 'selection_display', None)
 
     def _define_interface(self, plots, allow_mismatch):
-        parameters = [{k:v.precedence for k,v in plot.param.params().items()
+        def param_objects(obj):
+            if hasattr(obj.param, 'objects'):
+                return obj.param.objects()
+            else:
+                # Deprecated in Param 1.12.0
+                return obj.param.params()
```
Hoxbro, line 308: You haven't done this other than here.  

We should just pin to param 1.12.0 and use `xxx.param.objects()` everywhere. I can see Panel has it pinned to that version too.

maximlt: I was wondering why I did that here only and not everywhere, and then noticed that the other places were in some test files. If I remember correctly, my reasoning was that the tests are more likely to be run with the latest Param, while the code itself is going to be run with whatever version of Param users happen to have.

Weird reasoning maybe! And I should have indicated that in the comment to make the review easier for sure.

I'll bump the minimum required version of Param to 1.12.0.

setup.py
```diff
@@ -11,7 +11,7 @@
 
 setup_args = {}
 install_requires = [
-    "param >=1.9.3,<2.0",
+    "param >=1.12.0,<2.0",
```
Hoxbro, line 14: Shouldn't the upper pin be removed?

maximlt: Yep I guess we'll need to update the upper pin soon, to get HoloViews (and others) to be tested with alpha versions of Param. I'll update it to `<3.0`.

Hoxbro (approved):

Issue #5615, opened 2023-02-08
Error when using groupby with hvplot


lforesta:
I'm using hvplot to plot gridded data grouped by time, as shown in this example: https://hvplot.holoviz.org/user_guide/Gridded_Data.html#2d

I can run that example without issues, but when I use my data I get an error from holoviews (see below). My cmd is:
`my_data.hvplot(groupby='time') ` where `my_data` is a `xr.DataArray`.

#### ALL software version info
```
Ubuntu 22.04
Python 3.9.16
hvplot                   0.8.2
geoviews                 1.9.5
holoviews                1.15.2
bokeh                    2.4.3
```

#### Description of expected behavior and the observed behavior
Should plot the gridded data with a slider (over the time dimension).

#### Complete, minimal, self-contained example code that reproduces the issue

The issue is probably in something missing in the input file I use, which is a custom tif file (see DataArray structure below).

```
my_data
<xarray.DataArray 'bareness' (time: 2, lat: 5051, lon: 6035)>
dask.array<getitem, shape=(2, 5051, 6035), dtype=float32, chunksize=(2, 1000, 1000), chunktype=numpy.ndarray>
Coordinates:
  * time         (time) datetime64[ns] 2018-01-06 2018-01-11
  * lon          (lon) float64 -62.42 -62.42 -62.42 ... -62.05 -62.05 -62.05
  * lat          (lat) float64 -27.7 -27.7 -27.7 -27.71 ... -28.01 -28.01 -28.01
    spatial_ref  int64 0
    mask         (lat, lon) bool False False False False ... False False False
Attributes:
    STATISTICS_VALID_PERCENT:  0
    scale_factor:              1.0
    add_offset:                0.0
    long_name:                 ('2018-01-01T00:00:00+00:00', '2018-01-06T00:0...

my_data.hvplot(groupby='time') -> returns error posted below
```

#### Stack traceback and/or browser JavaScript console output
```
File ~/repos/hvplot-env/.venv/lib/python3.9/site-packages/holoviews/core/options.py:1426, in Store.render(cls, obj)
   1424 data, metadata = {}, {}
   1425 for hook in hooks:
-> 1426     ret = hook(obj)
   1427     if ret is None:
   1428         continue
...
    298 if not isinstance(library, list):
    299     library = [library]
--> 300 return any([obj.__module__.split('.')[0].startswith(l) for l in library])

AttributeError: 'collections.OrderedDict' object has no attribute '__module__'
[...]

PR #5646, opened 2023-03-08
Only use long_name if it is a string

Hoxbro:
Fixes https://github.com/holoviz/holoviews/issues/5615

@jlstevens I know we discussed having a warning saying that the long name should be a string, but coming back to the code after a week, I think I would rather ignore the long name if it is not a string.


jlstevens:
I think that is another very reasonable policy.

Hoxbro:
Great to here will merge this then.

PR #5650, opened 2023-03-08
Update setup.py

nareshdevasani:
Removing invalid dependency -     'mpl_sample_data >=3.1.3',
mpl-sample-data is not available in PyPI.


hoxbro:
Thank you for the PR. We have removed the package in another PR.

PR #5654, opened 2023-03-10
Remove deprecations

Hoxbro:
This PR removes old deprecations:

- Non-working tile sources
- Having a non-string as a column name in a DataFrame

I also added warnings.py (and removed _exceptions.py) to align how Panel does it.


holoviews/element/tiles.py
```diff
@@ -127,27 +126,10 @@ def easting_northing_to_lon_lat(easting, northing):
     )
 }
 
-def deprecation_warning(name, url, reason):
-    def deprecated_tilesource_warning():
```
jlstevens, line 131: This deprecation approach for tile sources was added not so long ago...removing it now might be a bit too soon!

jlstevens:
Made one quick comment, otherwise looks good!

holoviews/element/tiles.py
Hoxbro: This was added two years ago is that too soon?

https://github.com/holoviz/holoviews/pull/4941

Hoxbro: The notebook is working. Though, I update the text to reflect it.

examples/reference/elements/bokeh/Tiles.ipynb
```diff
@@ -77,7 +77,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "The full set of predefined tile sources can be accessed on the ``holoviews.element.tiles.tile_sources`` dictionary. Some tile sources from CartoDB are no longer publicly available and marked with a deprecation warning. These are filtered below:"
+    "The full set of predefined tile sources can be accessed on the `holoviews.element.tiles.tile_sources` dictionary. "
```
Hoxbro, line 80:
```suggestion
    "The full set of predefined tile sources can be accessed on the ``holoviews.element.tiles.tile_sources`` dictionary."
```

holoviews/element/tiles.py
jlstevens: Thanks!

Issue #5239, opened 2022-03-18
Rasterize not working in combination with multiple rectangles with DateTime coordinates


StevenWarmelink:
#### ALL software version info
pandas==1.3.5
holoviews==1.14.8
bokeh==2.3.3
python==3.8.5

#### Description of expected behavior and the observed behavior
When rasterizing multiple hv.Rectangles, I expect to be able to supply Timestamp coordinates; this works fine without the rasterization or for single Rectangles. However, when multiple Rectangles with Timestamp coordinates are rasterized, the supplied timestamps are out of bounds. 

EDIT: This also occurs when a single rectangle is supplied in a list when calling hv.Rectangles.

#### Complete, minimal, self-contained example code that reproduces the issue

```
from holoviews.operation.datashader import rasterize
import pandas as pd
import holoviews as hv
import bokeh 
hv.extension("bokeh",logo=False)

dts         = pd.to_datetime(['1.15.2019', '2.15.2019'])
rectangle_1 = (dts[0], 0, dts[1], 0.3)
rectangle_2 = (dts[0], 0.5, dts[1], 0.9)
rects       = hv.Rectangles([rectangle_1, rectangle_2])
rast_rects  = rasterize(rects)
display(rast_rects)

```

#### Stack traceback and/or browser JavaScript console output
```
WARNING:param.dynamic_operation: Callable raised "OutOfBoundsDatetime('Out of bounds nanosecond timestamp: 51008-08-27 00:00:00')".
Invoked as dynamic_operation(height=400, scale=1.0, width=400, x_range=(numpy.datetime64('2019-01-15T00:00:00.000000000'), numpy.datetime64('2019-02-15T00:00:00.000000000')), y_range=(0.1, 0.9))
WARNING:param.dynamic_operation: Callable raised "OutOfBoundsDatetime('Out of bounds nanosecond timestamp: 51008-08-27 00:00:00')".
Invoked as dynamic_operation(height=400, scale=1.0, width=400, x_range=(numpy.datetime64('2019-01-15T00:00:00.000000000'), numpy.datetime64('2019-02-15T00:00:00.000000000')), y_range=(0.1, 0.9))
Traceback (most recent call last):
  File "/home/steven/anaconda3/lib/python3.8/site-packages/holoviews/plotting/util.py", line 277, in get_plot_frame
    return map_obj[key]
  File "/home/steven/anaconda3/lib/python3.8/site-packages/holoviews/core/spaces.py", line 1344, in __getitem__
[...]

PR #5662, opened 2023-03-14
Initial Pandas 2 support

Hoxbro:
Ref: https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -781,7 +781,7 @@ class spikes_aggregate(LineAggregationOperation):
             df['y0'] = np.array(0, df.dtypes[y.name])
             yagg = ['y0', y.name]
         if xtype == 'datetime':
-            df[x.name] = cast_array_to_int64(df[x.name])
+            df[x.name] = cast_array_to_int64(df[x.name].astype('datetime64[ns]'))
 
         params = self._get_agg_params(element, x, y, agg_fn, (x0, y0, x1, y1))
 
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -781,7 +781,7 @@ class spikes_aggregate(LineAggregationOperation):
             df['y0'] = np.array(0, df.dtypes[y.name])
             yagg = ['y0', y.name]
         if xtype == 'datetime':
-            df[x.name] = cast_array_to_int64(df[x.name].astype('datetime64[us]'))
+            df[x.name] = cast_array_to_int64(df[x.name])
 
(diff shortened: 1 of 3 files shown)


Hoxbro:
The datetime change is because Pandas is now beginning to support non-nanosecond precision values:

``` python
import pandas as pd
from datetime import datetime

s0 = pd.Series([datetime(2020, 1, 1)])
s1 = s0.astype("datetime64[us]")
s2 = s1.astype("int64")

print(pd.__version__)
print(s0[0], s0.dtype)
print(s1[0], s1.dtype)
print(s2[0], s2.dtype)
```

Outputs:
``` bash
1.5.3
2020-01-01 00:00:00 datetime64[ns]
2020-01-01 00:00:00 datetime64[ns]
1577836800000000000 int64
```

``` python-traceback
2.0.0rc0
2020-01-01 00:00:00 datetime64[ns]
2020-01-01 00:00:00 datetime64[us]
1577836800000000 int64
```

Also beginning to get this warning from xarray because of this:
``` python-traceback
/home/shh/Repos/holoviz/holoviews/holoviews/operation/datashader.py:848: UserWarning: Converting non-nanosecond precision datetime values to nanosecond precision. This behavior can eventually be relaxed in xarray, as it is an artifact from pandas which is now beginning to support non-nanosecond precision values. This warning is caused by passing non-nanosecond np.datetime64 or np.timedelta64 values to the DataArray or Variable constructor; it can be silenced by converting the values to nanosecond precision ahead of time.
  agg[ydim] = (agg[ydim]/1e3).astype('datetime64[us]')
```

holoviews/operation/datashader.py
```diff
@@ -820,13 +820,13 @@ def _process(self, element, key=None):
         (x_range, y_range), (xs, ys), (width, height), (xtype, ytype) = info
         ((x0, x1), (y0, y1)), (xs, ys) = self._dt_transform(x_range, y_range, xs, ys, xtype, ytype)
 
-        df = element.interface.as_dframe(element)
+        df = element.interface.as_dframe(element).copy()
```
maximlt, line 823: Is the copy required?

maximlt: Are you trying to sneak in a fix for https://github.com/holoviz/holoviews/issues/5239? :D

(If so, it should probably be done in another PR?)

Hoxbro: Seems like I did... 

Though, I don't think keeping the fix in this PR is unreasonable. As the changes are related to pandas and datetime.

Hoxbro:
xarray does not yet support pandas 2.0: https://github.com/pydata/xarray/issues/7716

PR #5664, opened 2023-03-15
Rectangles use Bokeh quad not rect glyph

ianthomas23:
This fixes a bug using `hv.Rectangles` on logarithmic axes when using the Bokeh WebGL backend. Related Bokeh issue is https://github.com/bokeh/bokeh/issues/12734. Here is an example of the problem:

Screenshot 2023-03-15 at 17 07 51

The cause of the problem is that `hv.Rectangles`, which are defined by min and max `x` and `y` values, are implemented using bokeh `rect` glyphs which use centre coordinates, `width` and `height`. The center coordinates are calculated as the means of the `x` and `y` coordinates, which makes sense for linear axes but not necessarily for logarithic. The solution is to use the Bokeh `quad` glyph instead which is characterised by min and max `x` and `y` so no transform of these parameters is required. Here is the output using this PR:

Screenshot 2023-03-15 at 17 07 16

`quad` glyph has been in Bokeh since at least version 1.4.0 (https://docs.bokeh.org/en/1.4.0/docs/reference/models/glyphs/quad.html?highlight=quad#bokeh.models.glyphs.Quad) in November 2019.

I haven't added a new test as I don't think there are any specific webgl tests implemented.

### DIFF
--- a/holoviews/plotting/bokeh/geometry.py
+++ b/holoviews/plotting/bokeh/geometry.py
@@ -50,7 +50,7 @@ class RectanglesPlot(GeomMixin, LegendPlot, ColorbarPlot):
 
     _allow_implicit_categories = False
     _nonvectorized_styles = base_properties + ['cmap']
-    _plot_methods = dict(single='rect')
+    _plot_methods = dict(single='quad')
     _batched_style_opts = line_properties + fill_properties
     _color_style = 'fill_color'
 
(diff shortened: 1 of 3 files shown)


philippjfr (approved):
Looks good to me!

Hoxbro:
It also looks good to me. 

Is this ready to be merged?

ianthomas23:
> Is this ready to be merged?

Yes.

PR #5667, opened 2023-03-17
Set the bounds of the `kdims` List parameter

maximlt:
I just ran the unit test suite of HoloViews with Param installed from https://github.com/holoviz/param/pull/605 and a few tests failed. They failed when `kdims` `List` is overridden in a sub Parameterized class, and the overriding `List` is not defining explicitly the new `bounds`. In the current state of Param (1.x) , `bounds` is not inherited from any super class but is always the default of `(0, None)` (meaning a list of any length). With the tested PR `bounds` is inherited.

The change made in this PR was to set explicitly the `bounds` in the subclasses. Alternatively, one could have loosen `bounds` in the super classes. The commits list what change was made and why. For instance, one could have set  `bounds` of `Chart.kdims` (super class) to `(0, None)` instead of updating `bounds` on `Spikes.kdims` (sub class) to `(0, None)`. I didn't spend much time thinking about doing the alternative 

I noticed there are a few typos in the commits, here are the changes made:
- set bounds to `(0,None)` to `Spikes.vdims` to override `Chart.vdims` bounds of `(1, None)`
- set bounds to `(3,3)` to `Path3D.vdims` to override `Geometry.vdims` bounds of `(2, 2)`
- set bounds to `(3,3)` to `Scatter3D.vdims` to override `Geometry.vdims` bounds of `(2, 2)`
- set bounds to `(3,3)` to `TriSurface.vdims` to override `Geometry.vdims` bounds of `(2, 2)`

### DIFF
--- a/holoviews/element/chart.py
+++ b/holoviews/element/chart.py
@@ -215,7 +215,7 @@ class Spikes(Selection1DExpr, Chart):
 
     kdims = param.List(default=[Dimension('x')], bounds=(1, 1))
 
-    vdims = param.List(default=[])
+    vdims = param.List(default=[], bounds=(0, None))
 
     _auto_indexable_1d = False
 
--- a/holoviews/element/chart3d.py
+++ b/holoviews/element/chart3d.py
@@ -140,7 +140,7 @@ class Path3D(Element3D, Path):
 
     kdims = param.List(default=[Dimension('x'),
                                 Dimension('y'),
-                                Dimension('z')])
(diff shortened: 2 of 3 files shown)


maximlt:
@Hoxbro this could be merged safely, and need to be merged not to break HoloViews with Param 2.0.

Hoxbro (approved):

Hoxbro:
Thank you for this :+1:

PR #5669, opened 2023-03-23
Add real legend_cols support for Bokeh

Hoxbro:
In Bokeh 3.1 support for `ncols` and `nrows` was added. This PR implements that. Before that `legend_cols` was only used to determine if the legend should be horizontal or vertical.

###  Current behavior
current_bokeh

### Behavior after this PR
pr_bokeh

### Matplotlib behavior
matplotlib

``` python
import holoviews as hv
import panel as pn

hv.extension("bokeh")
# hv.extension("matplotlib")


curves = [hv.Curve([i] * 10, label=str(i)) for i in range(10)]
data = [(0, 0, "A"), (1, 1, "B"), (2, 2, "C")]
optss = [
    dict(show_legend=True, legend_cols=2),
    dict(show_legend=True, legend_cols=2, legend_position="top"),
    dict(show_legend=True),
    dict(show_legend=True, legend_position="top"),
]


plots = []
for opts in optss:
    ol = hv.Overlay(curves).opts(**opts)
    bar = hv.Points(data, vdims=["y", "color"]).opts(color="color", **opts)
    plots.append(pn.Column(ol, bar))

pn.Row(*plots).servable()
```

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -2174,8 +2174,8 @@ class ColorbarPlot(ElementPlot):
 
 class LegendPlot(ElementPlot):
 
-    legend_cols = param.Integer(default=False, doc="""
-        Whether to lay out the legend as columns.""")
+    legend_cols = param.Integer(default=0, bounds=(0, None), doc="""
+        Number of columns for legend.""")
 
     legend_labels = param.Dict(default=None, doc="""
         Label overrides.""")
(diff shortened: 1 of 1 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -2218,12 +2218,16 @@ def _process_legend(self, plot=None):
             or not self.show_legend):
             legend.items[:] = []
         else:
-            plot.legend.orientation = 'horizontal' if self.legend_cols else 'vertical'
+            if bokeh3 and self.legend_cols:
+                plot.legend.nrows = self.legend_cols
```
Hoxbro, line 2222: `nrows` are used here to get the desired effect:

|   |   | 
| -- | -- |
| 0 | 2 |
|1 | 3 | 

If `ncols` are used:
|   |   | 
| -- | -- |
| 0 | 1 |
|2 | 3 |

philippjfr: Seems odd, would expect exactly the opposite 🤷

philippjfr (approved):
Very, very excited to finally have support for this one. The changes look good.

PR #5677, opened 2023-04-11
fix anchor links

Yura52:
Currently, the links are broken because of the upper case


MarcSkovMadsen:
Hi @Hoxbro and @Yura52 

The problem is that now the links are broken if used as notebooks in Jupyterlab. I also worked on fixing this. There is an issue somewhere where I believe I noted its not possible to make these links work both in Jupyterlab and in Sphinx docs.

😄

PR #5693, opened 2023-04-24
Ensure .stack works on Areas with different vdims

### LINKED ISSUE #5689: Area stack requires individual elements to share the same vdim
HoloViews 1.15.4
Bokeh 2.4.3
Python 3.10
In notebooks and server

####

I need to stack several columns of a dataframe. However `hv.Area.stack` yields an empty plot.

```python
import pandas as pd
import holoviews as hv
hv.extension('bokeh')

df = pd.DataFrame({'x': [1, 2, 3], f'y_1': [1, 2, 3], 'y_2': [6, 4, 2], 'y_3': [8, 1, 2]})
overlay = []
for col in ['y_1', 'y_2', 'y_3']:
    overlay.append(hv.Area(df, kdims='x', vdims=col, label=col))
overlay = hv.Overlay(overlay)

hv.Area.stack(overlay)
```

This is because it is expected that the different elements will share the same vdim. [In the docs](https://holoviews.org/reference/elements/bokeh/Area.html) a numpy array is used and the elements share the same `value` vdim. But I think the need to display stacked columns of a dataframe is quite common. You could circumvent the issue by feeding `Areas` columns instead of the dataframe, but in my situation I need to use the API like `Element(df, kdims=kdims, vdims=vdims)`.

In the [source code](https://github.com/holoviz/holoviews/blob/38a4350606e7154f9858d432ca23c84d0321e4b8/holoviews/element/chart.py#L238), the Overlay is cast to a NdOverlay as follows:

```
:Overlay
   .Area.Y_1 :Area   [x]   (y_1)
   .Area.Y_2 :Area   [x]   (y_2)
   .Area.Y_3 :Area   [x]   (y_3)
```
becomes
```
:NdOverlay   [Element]
   :Area   [x]   (y_3)
```

 and the resulting dframe that is used to build the stacked areas is:
```
           y_1  y_2  y_3
Element x               
0       1  1.0  NaN  NaN
        2  2.0  NaN  NaN
        3  3.0  NaN  NaN
1       1  NaN  6.0  NaN
        2  NaN  4.0  NaN
        3  NaN  2.0  NaN
2       1  NaN  NaN  8.0
        2  NaN  NaN  1.0
        3  NaN  NaN  2.0
```

Then the vdim is determined to be `y_3` throughout the for loop that follows and the corresponding dataframe column is used for each element of the stack, meaning you end up with NaNs everywhere.

Here is a modified version to make it work without a complete overhaul:
```python
    @classmethod
    def stack(cls, areas, baseline_name='Baseline'):
        """
        Stacks an (Nd)Overlay of Area or Curve Elements by offsetting
        their baselines. To stack a HoloMap or DynamicMap use the map
        method.
        """
        if not len(areas):
            return areas
        is_overlay = isinstance(areas, Overlay)
        if is_overlay:
            element_vdims = [[element.vdims[0], baseline_name] for element in areas]
            areas = NdOverlay({i: el for i, el in enumerate(areas)})
        df = areas.dframe(multi_index=True)
        levels = list(range(areas.ndims))
        # vdim = areas.last.vdims[0]
        # vdims = [vdim, baseline_name]
        baseline = None
        stacked = areas.clone(shared_data=False)
        for (key, sdf), vdims in zip(df.groupby(level=levels, sort=False), element_vdims):
            vdim = vdims[0]
            sdf = sdf.droplevel(levels).reindex(index=df.index.unique(-1), fill_value=0)
            if baseline is None:
                sdf[baseline_name] = 0
            else:
                sdf[vdim.name] = sdf[vdim.name] + baseline
                sdf[baseline_name] = baseline
            baseline = sdf[vdim.name]
            stacked[key] = areas[key].clone(sdf, vdims=vdims)
        return Overlay(stacked.values()) if is_overlay else stacked
```

Result:
image

TheoMathurin:
Fixes #5689 

I have yet to try this on main, but works fine on 1.15.4.

### DIFF
--- a/holoviews/element/chart.py
+++ b/holoviews/element/chart.py
@@ -246,9 +246,9 @@ class Area(Curve):
         is_overlay = isinstance(areas, Overlay)
         if is_overlay:
             areas = NdOverlay({i: el for i, el in enumerate(areas)})
-        vdims = [[element.vdims[0], baseline_name] for element in areas]
         df = areas.dframe(multi_index=True)
         levels = list(range(areas.ndims))
+        vdims = [[el.vdims[0], baseline_name] for el in areas]
         baseline = None
         stacked = areas.clone(shared_data=False)
         for (key, sdf), element_vdims in zip(df.groupby(level=levels, sort=False), vdims):
--- a/holoviews/element/chart.py
+++ b/holoviews/element/chart.py
@@ -245,16 +245,16 @@ class Area(Curve):
             return areas
         is_overlay = isinstance(areas, Overlay)
         if is_overlay:
-            element_vdims = [[element.vdims[0], baseline_name] for element in areas]
+            vdims = [[element.vdims[0], baseline_name] for element in areas]
             areas = NdOverlay({i: el for i, el in enumerate(areas)})
         else:
-            element_vdims = [areas.last.vdims[0], baseline_name]
+            vdims = [areas.last.vdims[0], baseline_name]
         df = areas.dframe(multi_index=True)
         levels = list(range(areas.ndims))
         baseline = None
         stacked = areas.clone(shared_data=False)
-        for (key, sdf), vdims in zip(df.groupby(level=levels, sort=False), element_vdims):
-            vdim = vdims[0]
+        for (key, sdf), element_vdims in zip(df.groupby(level=levels, sort=False), vdims):
+            vdim = element_vdims[0]
             sdf = sdf.droplevel(levels).reindex(index=df.index.unique(-1), fill_value=0)
             if baseline is None:
                 sdf[baseline_name] = 0
@@ -262,5 +262,5 @@ class Area(Curve):
                 sdf[vdim.name] = sdf[vdim.name] + baseline
                 sdf[baseline_name] = baseline
             baseline = sdf[vdim.name]
-            stacked[key] = areas[key].clone(sdf, vdims=vdims)
+            stacked[key] = areas[key].clone(sdf, vdims=element_vdims)
         return Overlay(stacked.values()) if is_overlay else stacked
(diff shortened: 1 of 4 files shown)


Hoxbro:
Thank you. The changes look good to me. Will it be possible for you to add a unit test? You can use the original example and verify the Baseline is correct. This can be found here: 
image
 
The test can be added here:
https://github.com/holoviz/holoviews/blob/main/holoviews/tests/plotting/bokeh/test_areaplot.py

TheoMathurin:
Ok, maybe something along these lines?

```python
df = pd.DataFrame({'x': [1, 2, 3], 'y_1': [1, 2, 3], 'y_2': [6, 4, 2], 'y_3': [8, 1, 2]})
overlay = hv.Overlay([hv.Area(df, kdims='x', vdims=col, label=col) for col in ['y_1', 'y_2', 'y_3']])

plot = hv.Area.stack(overlay)
baselines = [np.array([0, 0, 0]), np.array([1., 2., 3.]), np.array([7., 6., 5.])]

for n, baseline in zip(plot.data, baselines):
    self.assertEqual(plot.data[n].data.Baseline, baseline)
```

Hoxbro:
Yes. That looks good to me.

TheoMathurin:
Ok, I guess the test be included for all three backends? I see that there are `test_areaplot` modules for plotly and matplotlib as well.

There's also the `test_operation` module, I don't know if it could be an alternative since there are already tests on the stack operation there.

Hoxbro:
> Alright I guess the test should be included for all three backends? 

It would be nice, but it is up to you if you want to do it. 

> There's also the test_operation

Just ignore this. Operations are pretty cumbersome, and we want to add a test for your simple example.

Hoxbro:
Thank you for PR :+1:

TheoMathurin:
You're welcome!

PR #5695, opened 2023-04-25
HEP1 - Drop support for Python 3.7

Hoxbro:
With Python 3.7's end of life at the end of June and the upcoming release of Holoviews 1.17, it makes sense to have Holoviews support Python 3.8 and up.

Following: HEP1 at https://github.com/holoviz/holoviz/pull/362


philippjfr:
I'd be in favor of doing this after the 1.16 release. There's no pressing need right now since we still support older versions of Bokeh/Panel.

Hoxbro:
The bump is not related to Bokeh/Panel going to 3.8, but that it will become much more time-consuming supporting old versions of Python, _especially_ in the CI. Note that we will still support older versions of Bokeh/Panel with this change. 

Python 3.7 is going out of life in two months and in my opinion, this is close enough to stop supporting it. If we had a more consistent release of HoloViews, a reason could be made to wait, but looking at the history of the last two minor releases, a year can easily go by. 1.15 was released on July 7, 2022, and 1.14 was released on December 3, 2020.

PR #5703, opened 2023-05-01
Support partial notebook extension loading

philippjfr:
Requires https://github.com/holoviz/panel/pull/4752

### DIFF
--- a/holoviews/plotting/renderer.py
+++ b/holoviews/plotting/renderer.py
@@ -642,7 +642,7 @@ class Renderer(Exporter):
         return options
 
     @classmethod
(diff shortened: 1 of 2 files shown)


holoviews/plotting/renderer.py
```diff
@@ -642,12 +642,17 @@ def validate(cls, options):
         return options
 
     @classmethod
-    def load_nb(cls, inline=True):
+    def load_nb(cls, inline=True, reloading=same_cell_execution):
```
philippjfr, line 645:
```suggestion
    def load_nb(cls, inline=True, reloading=False):
```

PR #5719, opened 2023-05-17
Speed up import time

Hoxbro:
The main focus is to get the import time down. Done with `python -X importtime -c "import holoviews" 2> log.txt && tuna log.txt`

For now, I have moved IPython, Ibis, and Panel to be lazy imports.  

Current main (taking 1.256 seconds to load):
image

With this PR (takes 0.338 seconds to load):
image


Hoxbro:
Good catch. I think I made those changes before I added the sys.module change. I have reverted it now.

PR #5729, opened 2023-05-26
Add enable_mathjax and set it and inline to False

Hoxbro:
Follow up to https://github.com/holoviz/panel/pull/4895 and implements https://github.com/holoviz/holoviews/issues/5730


philippjfr (approved):
LGTM!

PR #5744, opened 2023-06-06
Work with non-gridded xarray dataset

hoxbro:
Attempt at supporting request in https://github.com/holoviz/hvplot/issues/1080.

This is very much still work-in-progress.

### DIFF
--- a/holoviews/core/data/xarray.py
+++ b/holoviews/core/data/xarray.py
@@ -186,7 +186,7 @@ class XArrayInterface(GridInterface):
             data = xr.DataArray(value_array, coords=coords, **xr_kwargs)
         else:
             arrays = {}
-            for dim in kdims + vdims:
+            for dim in vdims:
                 arr = data[dim.name]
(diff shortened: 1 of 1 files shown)


Hoxbro:
My basic thought process is to treat data with empty `vdims`  as tabular data, which makes it possible to do the following: 

``` python
import holoviews as hv
import numpy as np
import xarray as xr

hv.extension("bokeh")

ds = xr.Dataset(
    data_vars={
        "lat": ("trajectory", np.linspace(30, 35, 15)),
        "lon": ("trajectory", np.linspace(16, 20, 15)),
        "example": ("trajectory", np.linspace(16, 20, 15)),
    },
    coords={"trajectory": np.arange(15)},
)

hv.Dataset(ds, kdims=["lat", "lon"], vdims=[]).to(hv.Points)  
```

image

And when not setting `vdims` to an empty list we get this (`hv.Dataset(ds, kdims=["lat", "lon"]).to(hv.Points)`):
image

PR #5751, opened 2023-06-09
Implementation of ImageStack 

Hoxbro:
WIP

``` python
import holoviews as hv
import numpy as np
import panel as pn

hv.extension("bokeh")


def categorical_data():
    n = 20
    samples = 1000
    rng = np.random.default_rng(92478)

    centers = [(0.3, 0.3), (0.7, 0.5), (0.3, 0.7)]
    radii = [0.2, 0.3, 0.15]
    ncats = len(radii)

    dx = 1.0 / n
    data = np.zeros((n, n, ncats), dtype=np.float32)
    data[:] = np.nan

    for k in range(ncats):
        x = rng.normal(centers[k][0], radii[k], samples)
        y = rng.normal(centers[k][1], radii[k], samples)
        i = (x / dx).astype(int)
        j = (y / dx).astype(int)
        for ii, jj in zip(i, j):
            if 0 <= ii < n and 0 <= jj < n:
                if np.isnan(data[jj, ii, k]):
                    data[jj, ii, k] = 1.0
                else:
                    data[jj, ii, k] += 1.0
    return data


data = categorical_data()

colors = ["red", "blue", "green"]
a = hv.ImageStack(data).opts(
    # colorbar=True,
    width=800,
    height=800,
    cmap=colors,
    num_colors=100,
)

pn.panel(a).servable()
```

Reference for me:
https://github.com/holoviz/holoviews/pull/4806
https://github.com/bokeh/bokeh/pull/12356


philippjfr:
- `shade` now has to support ingesting an ImageStack as well as an NdOverlay of Image
- Add Compositor definitions for matplotlib and plotly to convert an ImageStack -> RGB
- Finish plot implementation by finishing support for `invert_axes`, `invert_xaxis` and `inter_yaxis`.
- Add reference notebook for `ImageStack`
- Add tests

Nice to have:

- Automatically add a legend for the categories in the ImageStack

ahuang11:
```
import holoviews as hv
import xarray as xr
import numpy as np
import panel as pn

hv.extension("bokeh")


da = xr.DataArray(coords={"x": [1, 2, 3], "y": [6, 7, 8]}, dims=["x", "y"])

da_1 = da.copy()
da_1.name = "a"
da_1.values = [[np.nan, np.nan, 1], [np.nan] * 3, [np.nan] * 3]

da_2 = da.copy()
da_2.name = "b"
da_2.values = [[np.nan] * 3, [1, 1, np.nan], [np.nan] * 3]

da_3 = da.copy()
da_3.name = "c"
da_3.values = [[np.nan] * 3, [np.nan] * 3, [1, 1, 1]]

ds = xr.merge([da_1, da_2, da_3])

colors = ["red", "blue", "green"]
a = hv.ImageStack(ds).opts(
    # colorbar=True,
    cmap=colors,
    num_colors=100,
    title="original",
)
b = a.clone().opts(invert_axes=True, title="invert axes")

c = a.clone().opts(invert_yaxis=True, title="invert yaxis")
d = a.clone().opts(invert_xaxis=True, title="invert xaxis")
pn.GridBox(
    pn.panel(a, linked_axes=False),
    pn.panel(b, linked_axes=False),
    pn.panel(c, linked_axes=False),
    pn.panel(d, linked_axes=False),
    ncols=2,
)
```

image

ahuang11:
Does hover need support?

Hoxbro:
> Does hover need support?

Hover support is out of scope for this PR as it also needs work in Bokeh (or a new release). See https://github.com/bokeh/bokeh/pull/13193, https://github.com/bokeh/bokeh/issues/13200, and https://github.com/bokeh/bokeh/issues/13354.

ahuang11:
```
import holoviews as hv
import xarray as xr
import numpy as np
import panel as pn
from holoviews.operation.datashader import shade, datashade

hv.extension("bokeh")


da = xr.DataArray(coords={"x": [1, 2, 3], "y": [6, 7, 8]}, dims=["x", "y"])

da_1 = da.copy()
da_1.name = "a"
da_1.values = [[np.nan, np.nan, 1], [np.nan] * 3, [np.nan] * 3]

da_2 = da.copy()
da_2.name = "b"
da_2.values = [[np.nan] * 3, [1, 1, np.nan], [np.nan] * 3]

da_3 = da.copy()
da_3.name = "c"
da_3.values = [[np.nan] * 3, [np.nan] * 3, [1, 1, 1]]

ds = xr.merge([da_1, da_2, da_3])

colors = ["red", "blue", "green"]
a = hv.ImageStack(ds).opts(
    # colorbar=True,
    cmap=colors,
    num_colors=100,
    title="original",
)
# pn.GridBox(
#     pn.panel(a, linked_axes=False),
#     pn.panel(b, linked_axes=False),
#     pn.panel(c, linked_axes=False),
#     pn.panel(d, linked_axes=False),
#     ncols=2,
# )

shade(a)
```

image

ahuang11:
mpl & plotly

image

image

ahuang11:
Weirdly using numpy array transposes the plot
```python
import holoviews as hv
import xarray as xr
import numpy as np
import panel as pn

hv.extension("bokeh")

x = np.arange(0, 3)
y = np.arange(5, 8)
a = np.array([[np.nan, np.nan, 1], [np.nan] * 3, [np.nan] * 3])
b = np.array([[np.nan] * 3, [1, 1, np.nan], [np.nan] * 3])
c = np.array([[np.nan] * 3, [np.nan] * 3, [1, 1, 1]])

ds = xr.Dataset(
    {"a": (["x", "y"], a), "b": (["x", "y"], b), "c": (["x", "y"], c)},
    coords={"x": x, "y": y},
)

colors = ["green", "red", "blue"]
(
    hv.ImageStack(ds, kdims=["x", "y"], vdims=["a", "b", "c"]).opts(cmap=colors) + 
    hv.ImageStack((x, y, a, b, c), kdims=["x", "y"], vdims=["a", "b", "c"]).opts(cmap=colors)
)
```

image

ahuang11:
Okay, I guess Image does the same, so maybe it's not an issue / separate issue
image

philippjfr:
Yeah definitely an issue but a separate issue.

holoviews/element/raster.py
```diff
@@ -517,13 +517,18 @@ class ImageStack(Image):
...
+
     group = param.String(default='ImageStack', constant=True)
 
     _ndim = 3
 
     _vdim_reductions = {1: Image}
 
     def __init__(self, data, kdims=None, vdims=None, **params):
+        if vdims is None:
+            raise ValueError("ImageStack requires vdims to be defined.")
```
ahuang11, line 531: I felt like this was necessary because if it defaulted to only z, the other layers do not show up. On the other hand, I wasn't sure how to elegantly handle all the possible input types (dict, ds, da, tuple), so I just force the user to specify vdims.

holoviews/element/raster.py
```diff
@@ -517,13 +517,18 @@ class ImageStack(Image):
     - overlay/ndoverlay of Images
     """
 
+    vdims = param.List(doc="""
```
ahuang11, line 520: I do not include `bounds=(0, None)` here because it errors without a default, but I also don't want the default to be `z` because it only plots the first layer.

ahuang11:
I believe this is ready for initial review, but I don't understand the invert_x/yaxis tests because although the tests pass, the inverted data matches the original data and it renders as expected.

holoviews/tests/plotting/bokeh/test_rasterplot.py
```diff
@@ -150,3 +151,78 @@ def test_rgb_invert_yaxis(self):
...
+        x = np.arange(0, 3)
+        y = np.arange(5, 8)
+        a = np.array([[np.nan, np.nan, 1], [np.nan] * 3, [np.nan] * 3])
+        b = np.array([[np.nan] * 3, [1, 1, np.nan], [np.nan] * 3])
+        c = np.array([[np.nan] * 3, [np.nan] * 3, [1, 1, 1]])
+
+        img_stack = ImageStack((x, y, a, b, c), kdims=["x", "y"], vdims=["a", "b", "c"])
+        plot = bokeh_renderer.get_plot(img_stack.opts(invert_xaxis=True))
+        source = plot.handles['source']
+        self.assertEqual(source.data['image'][0][0], a.T)
```
ahuang11, line 184: Why isn't this flipped?
```suggestion
        self.assertEqual(source.data['image'][0][0], a.T[::-1])
```

holoviews/element/raster.py
philippjfr: I think it might be nice to at least handle 3D arrays and xarray DataArray/Datasets automatically and error out for anything else.

holoviews/tests/plotting/bokeh/test_rasterplot.py
philippjfr: Isn't axis inversion handled by flipping the x and y values, rather than flipping the actual data?

holoviews/operation/flatten.py
```diff
@@ -0,0 +1,19 @@
+try:
+    from .datashader import shade
+except ImportError:
+    shade = None
+
+
+class flatten_stack(shade):
```
philippjfr, line 7: Since this is really just a plotting helper I'd probably move this into `plotting/util.py` Also please inline the datashader import, we should not import datashader unless the user uses this feature.

philippjfr (commented):
A few comments, otherwise this looks good.

ahuang11:
Okay, I think the tests are good to go now.

I believe the spatialpandas test is unrelated because https://github.com/holoviz/holoviews/actions/runs/6037187496/job/16381021321?pr=5874 from https://github.com/holoviz/holoviews/pull/5874 is also failing.

In regards to Bokeh 2 tests, Simon says "ImageStack was added in Bokeh 3.1, so it will fail on Bokeh 2 ci test. You can ignore those or comment them out as I will remove Bokeh 2 support before next release."

setup.py
```diff
@@ -41,6 +41,7 @@
     'dash >=1.16',
     'codecov',
     'ipython >=5.4.0',
+    'pandas <2.1',  # for spatialpandas recursion https://github.com/holoviz/spatialpandas/issues/124
```
Hoxbro, line 44:
```suggestion

```

Is no longer needed.

ahuang11:
I'm confused why xr.DataArray is considered GriddedInterface while xr.Dataset is considered XArrayInterface

image
image

I've don't understand how this line decides the interface:
```python

                                           datatype=kwargs.get('datatype'))
        (data, self.interface, dims, extra_kws) = initialized
```

holoviews/core/data/__init__.py
```diff
@@ -324,6 +324,7 @@ def __init__(self, data, kdims=None, vdims=None, **kwargs):
         kdims, vdims = kwargs.get('kdims'), kwargs.get('vdims')
 
         validate_vdims = kwargs.pop('_validate_vdims', True)
+        print(type(self))
```
ahuang11, line 327:
```suggestion
```

ahuang11:
I fixed allowing 3D xr.DataArrays:
image

However, I don't really know how to address the HoverTool. Is that the responsibility of Bokeh?
image

Hoxbro (commented):
I just had a quick look at the code and left some comments. 

For the hover problem, try to filter the `_hover_opts` method. For now, I think we should remove these extra dimensions, as we can only get one layer of information for now. When Bokeh 3.3 is released, we can extract more information because of https://github.com/bokeh/bokeh/pull/13193 (maybe we can already get this with a dev release). Further down the line, we should be able to get even better support with https://github.com/bokeh/bokeh/issues/13200.

(Merged in main to be able only to be able to see the changes made in this PR)

holoviews/core/data/xarray.py
```diff
@@ -121,8 +121,12 @@ def retrieve_unit_and_label(dim):
...
                                 cls)
             if not packed:
-                vdims = [vdim]
-                data = data.to_dataset(name=vdim.name)
+                if vdim in data.dims:
+                    data = data.to_dataset(vdim.name)
+                    vdims = [asdim(vd) for vd in data.data_vars]
+                else:
+                    vdims = [vdim]
+                    data = data.to_dataset(name=vdim.name)
```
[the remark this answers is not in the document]
ahuang11:
```
(method) def to_dataset(
    dim: Hashable = None,
    *,
    name: Hashable = None,
    promote_attrs: bool = False
) -> Dataset
Convert a DataArray to a Dataset.

Parameters
dim : Hashable, optional
    Name of the dimension on this array along which to split this array into separate variables. If not provided, this array is converted into a Dataset of one variable.
name : Hashable, optional
    Name to substitute for this array's name. Only valid if dim is not provided.
```

Hoxbro:
It is a limitation in Bokeh. Try to read the issues I mentioned in my last post.

ahuang11:
> For the hover problem, try to filter the _hover_opts method. For now, I think we should remove these extra dimensions, as we can only get one layer of information for now.

Okay. To build upon that though, I don't think we can even get one layer though; so do we just keep x and y?

Hoxbro:
> So do we just keep x and y?

yes

holoviews/operation/datashader.py
```diff
@@ -1252,9 +1249,16 @@ def _process(self, element, key=None):
...
             bounds = element.bounds
 
-        vdim = element.vdims[0].name
-        array = element.data[vdim]
         kdims = element.kdims
+        if isinstance(element, ImageStack):
+            vdim = element.vdims
+            array = element.data
+            if hasattr(array, "to_array"):
+                array = array.to_array("z")
```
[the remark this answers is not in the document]
ahuang11: It creates a new variable "z" so yes I suppose so.

image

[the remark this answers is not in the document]
ahuang11: I add handling (converting to array) here because without that, the downstream code would crash:
```python
if np.isnan(array.data).all():
```

ahuang11:
Should I be worried about the core tests?

Hoxbro:
> Eh I just wrapped try/except.

I would rather see something like this:
https://github.com/holoviz/holoviews/blob/d3f345871e51d407b4ff8857d0b9010be468ce02/holoviews/tests/core/data/test_binneddatasets.py#L225-L228

holoviews/plotting/bokeh/raster.py
```diff
@@ -303,8 +303,7 @@ def get_data(self, element, ranges, style):
 
     def _hover_opts(self, element):
         xdim, ydim = element.kdims
-        print(xdim)
-        super()._hover_opts(element)
+        return [(xdim.pprint_label, '$x'), (ydim.pprint_label, '$y')], {}
```
Hoxbro, line 306: I would like to see a comment about why it does not work right now and what it needed to get it to work in the future.

holoviews/operation/datashader.py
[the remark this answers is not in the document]
Hoxbro: I assume one of your new tests is for this case?

holoviews/tests/plotting/bokeh/test_rasterplot.py
```diff
@@ -150,3 +151,138 @@ def test_rgb_invert_yaxis(self):
...
+        x = np.arange(0, 3)
+        y = np.arange(5, 8)
+        a = np.array([[np.nan, np.nan, 1], [np.nan] * 3, [np.nan] * 3])
+        b = np.array([[np.nan] * 3, [1, 1, np.nan], [np.nan] * 3])
+        c = np.array([[np.nan] * 3, [np.nan] * 3, [1, 1, 1]])
+
+        img_stack = ImageStack((x, y, a, b, c), kdims=["x", "y"], vdims=["a", "b", "c"])
+        plot = bokeh_renderer.get_plot(img_stack)
+        source = plot.handles['source']
+        self.assertEqual(source.data['image'][0][0], a.T)
```
Hoxbro, line 165: At some point, I want to transition to using pytest. 

`self.assertEqual` has special power in the holoviews testing framework, e.g., deep comparison of elements. 

For these though it seems like it is just a "simple" comparison that can be done with `np.testing.assert_array_almost_equal`.

philippjfr (approved):
Apart from the minor comments left by @Hoxbro and I this looks great.

holoviews/tests/plotting/plotly/test_imagestack.py
```diff
@@ -0,0 +1,46 @@
...
+        a = np.array([[np.nan, np.nan, 1], [np.nan] * 3, [np.nan] * 3])
+        b = np.array([[np.nan] * 3, [1, 1, np.nan], [np.nan] * 3])
+        c = np.array([[np.nan] * 3, [np.nan] * 3, [1, 1, 1]])
+        image_stack = ImageStack(
+            (x, y, a, b, c), kdims=["x", "y"], vdims=["a", "b", "c"]
+        )
+        try:
+            assert isinstance(plotly_renderer.get_plot(image_stack), RGBPlot)
+        except ImportError:
+            return SkipTest("Skip test if datashader is unavailable for core tests")
```
Hoxbro, line 23: I would like this check to be very explicit. e.g., at the top of the method:
``` python
try:
    import datashader
except ImportError:
   raise SkipTest("Test requires datashader")
```

PR #5763, opened 2023-06-15
Sync legends of LayoutPlot

### LINKED ISSUE #5674: Support syncing between visibility legend labels
As shown in https://github.com/holoviz/hvplot/issues/937#issuecomment-1498749914 it is possible to sync the visibility between lines in different figures. 

This seems like a great addition to have in Holoviews. 

https://user-images.githubusercontent.com/19758978/230332683-

Hoxbro:
Closes #5674 

TODO: 
- [x] Need to add some tests
- [x] Work if click_policy is `hide` (replace `muted` with `visibility`). 
- [x] Check whether `sync_location` is needed other places
- [x] Work with column, row

``` python
import holoviews as hv
import hvplot.pandas
import numpy as np
import pandas as pd
from holoviews.plotting.bokeh.util import select_legends

df = pd.DataFrame(np.arange(10)[:, None] * np.array([1, 2, 3]), columns=list("ABC"))
plot = (df.hvplot() + df.hvplot() + df.hvplot() + df.hvplot()).cols(2)
select_legends(plot, 0, "top_left")
```

https://github.com/holoviz/holoviews/assets/19758978/91817cc3-27ed-45dd-9e3b-27df7ed09134

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1171,7 +1171,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         if 'legend_field' in properties and 'legend_label' in properties:
             del properties['legend_label']
         if "name" not in properties:
-            properties["name"] = properties.get("legend_label") or properties.get("legend_field", None)
+            properties["name"] = properties.get("legend_label") or properties.get("legend_field")
         renderer = getattr(plot, plot_method)(**dict(properties, **mapping))
         return renderer, renderer.glyph
 
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1170,6 +1170,8 @@ class ElementPlot(BokehPlot, GenericElementPlot):
             plot_method = plot_method[int(self.invert_axes)]
         if 'legend_field' in properties and 'legend_label' in properties:
             del properties['legend_label']
+        if "name" not in properties:
+            properties["name"] = properties.get("legend_label") or properties.get("legend_field", None)
         renderer = getattr(plot, plot_method)(**dict(properties, **mapping))
         return renderer, renderer.glyph
 
--- a/holoviews/plotting/bokeh/util.py
+++ b/holoviews/plotting/bokeh/util.py
@@ -437,6 +437,7 @@ def sync_legends(plot_layout):
         if fig.legend:
             click_policies.add(fig.legend.click_policy)
 
+    click_policies.remove("none")  # If legend is not visible, click_policy is "none"
     if len(click_policies) > 1:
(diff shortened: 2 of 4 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -1170,6 +1170,8 @@ def _init_glyph(self, plot, mapping, properties):
             plot_method = plot_method[int(self.invert_axes)]
         if 'legend_field' in properties and 'legend_label' in properties:
             del properties['legend_label']
+        if "name" not in properties:
+            properties["name"] = properties.get("legend_label") or properties.get("legend_field", None)
```
ahuang11, line 1174: I think it defaults to None anyways.
```suggestion
            properties["name"] = properties.get("legend_label") or properties.get("legend_field")
```

holoviews/plotting/bokeh/util.py
```diff
@@ -407,6 +410,79 @@ def merge(tool, group):
...
+    policy = mapping.get(next(iter(click_policies)))
+    code = f"dst.{policy} = src.{policy}"
+    for item in items.values():
+        for src, dst in permutations(item, 2):
+            src.js_on_change(
+                policy,
+                CustomJS(code=code, args=dict(src=src, dst=dst)),
+            )
+
+def one_legend(plot_layout, legend_no=0, legend_position="top_right"):
```
ahuang11, line 459: What if we have legend_no as a list so users can have it displayed on multiple figures and maybe rename function as `limit_legends`? Also thoughts on renaming to `figure_index`?

ahuang11 (commented):
This will be useful! Just left some comments

holoviews/plotting/bokeh/element.py
Hoxbro: Correct. It is weird as I'm doing the right thing on the same line... It must have been some refactoring I did :man_shrugging:

holoviews/plotting/bokeh/util.py
Hoxbro: I'm not opposed to that idea. The function name would definitely need a rename. Not sure if `limit_legends` is the right one, though.

philippjfr (approved):
Looks great!

philippjfr:
One test at least would be nice though.

Hoxbro:
I could "easily" add a test to `select_legends`, but it should mostly be seen as a small utility. 

I have a hard time figuring out to make a unit test for `sync_legends`, as the changes are made when clicking on the canvas with changes happening in the browser.

philippjfr:
The unit test would just check that for different types of plots the CustomJS callback is set up correctly.

Hoxbro:
That should also be easy :)

dcherian:
Amazing thank you!

PR #5764, opened 2023-06-16
Add UI tests

Hoxbro:
Mostly copied from Panel setup

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -102,7 +102,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
+        os: ['ubuntu-latest']
     timeout-minutes: 60
     defaults:
       run:
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -95,3 +95,50 @@ jobs:
         run: |
           conda activate test-environment
           codecov
+  ui_test_suite:
+    name: UI tests on ${{ matrix.os }} with Python 3.9
+    needs: [pre_commit]
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
+    timeout-minutes: 60
+    defaults:
+      run:
+        shell: bash -el {0}
+    env:
+      DESC: "Python ${{ matrix.python-version }} tests"
+      PYTHONIOENCODING: "utf-8"
+      PANEL_LOG_LEVEL: info
+      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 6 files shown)


.github/workflows/test.yaml
```diff
@@ -95,3 +95,50 @@ jobs:
...
           conda activate test-environment
           codecov
+  ui_test_suite:
+    name: UI tests on ${{ matrix.os }} with Python 3.9
+    needs: [pre_commit]
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
```
maximlt, line 105: Our experience on Panel shows that it's easy to write flaky tests. The more UI tests you run, the more likely it is one of these tests will fail. Because of that, and because I don't think the OS is so relevant to UI tests (to me these are integration tests, that shouldn't cover so much code that isn't covered by unit tests), I'd really consider just running theses tests on one OS, and I'd pick ubuntu as it's the fastest!

(The browser tested is maybe more relevant than the OS for UI tests, let's not get into that now though :) )

holoviews/tests/ui/bokeh/test_layout.py
```diff
@@ -0,0 +1,29 @@
+import time
+
+import pytest
+
+pytestmark = pytest.mark.ui
```
maximlt, line 5: There's a warning about this line in the logs. There's more code that should be added to `conftest.py` maybe? https://github.com/holoviz/panel/blob/60ba9c91990044e8f06c3cc9ca2bc66c4dab73b1/panel/tests/conftest.py#L81

Links to and copy of the warning:
https://github.com/holoviz/holoviews/actions/runs/5290689182/jobs/9575200902?pr=5764#step:4:288

```python
holoviews/tests/ui/bokeh/test_layout.py:5
  /home/runner/work/holoviews/holoviews/holoviews/tests/ui/bokeh/test_layout.py:5: PytestUnknownMarkWarning: Unknown pytest.mark.ui - is this a typo?  You can register custom marks to avoid this warning - for details, see https://docs.pytest.org/en/stable/how-to/mark.html
    pytestmark = pytest.mark.ui
```

maximlt (commented):
That's a good start, it's just missing some HoloViews tests :D Do you intend to add some in this PR directly?

I've left some comments. I'm also going to add something that will fall right into the nice to have category and isn't a blocker for this PR. pytest's output shows all the skipped tests which isn't super useful:
image
Not showing that doesn't seem to be super straightforward (https://github.com/pytest-dev/pytest/issues/3730).

.github/workflows/test.yaml
Hoxbro: One OS is fine by me.

Hoxbro:
> Not showing that doesn't seem to be super straightforward

Looks good but we still want to show skipped UI tests, correct?

PR #5767, opened 2023-06-18
Only merge changed instance params with rasterize 

### LINKED ISSUE #5358: In this example, OSM Tiles look great, then disappear!

```
conda list | grep -E "bokeh|geoviews|datashader|hvplot|holovies|panel"

bokeh                     2.4.3            py39hf3d152e_0    conda-forge
bokeh-root-cmd            0.1.2              pyhd8ed1ab_0    conda-forge
datashader                0.14.0             pyh6c4a22f_0    conda-forge
geoviews                  1.9.5              pyhd8ed1ab_0    conda-forge
geoviews-core             1.9.5              pyha770c72_0    conda-forge
hvplot                    0.8.0              pyh6c4a22f_0    conda-forge
jupyter-panel-proxy       0.1.0                      py_0    conda-forge
panel                     0.13.1             pyhd8ed1ab_0    conda-forge
```

OSM basemap tiles appear nicely, then disappear.  See 15s movie here: 
https://www.screencast.com/t/JhREGjulrH

Full reproducible notebook here:
https://nbviewer.org/gist/5818739f4aa65e6ed81b1f1501c1be82

Hoxbro:
Supersedes #5351
Fixes #5358

`rasterize` has `expand=True` as default, whereas  `regrid` has it as `expand=False`. When applying all the parameters from the rasterize class, `regrid` was set to have `expand=True`. 

This change makes it so only non-default values are used.

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -29,10 +29,9 @@ from ..core import (
 from ..core.data import (
     Dataset, PandasInterface, XArrayInterface, DaskInterface, cuDFInterface
 )
-from ..core.data.util import compare
 from ..core.util import (
     cast_array_to_int64, cftime_types, cftime_to_timestamp,
-    datetime_types, dt_to_int, get_param_values
+    datetime_types, dt_to_int, get_param_values, isequal
 )
 from ..element import (Image, Path, Curve, RGB, Graph, TriMesh,
                        QuadMesh, Contours, Spikes, Area, Rectangles,
@@ -1381,7 +1380,7 @@ class rasterize(AggregationOperation):
         all_supplied_kws = set()
         non_default_params = {
             k: v for k, v in self.param.values().items()
-            if not compare(v, self.param[k].default)
+            if not isequal(v, self.param[k].default)
         }
         for predicate, transform in self._transforms:
             merged_param_values = dict(non_default_params, **self.p)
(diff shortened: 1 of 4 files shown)


holoviews/operation/datashader.py
```diff
@@ -1378,8 +1378,12 @@ def _process(self, element, key=None):
         # Potentially needs traverse to find element types first?
         all_allowed_kws = set()
         all_supplied_kws = set()
+        non_default_params = {
+            k: v for k, v in self.param.values().items()
+            if not isequal(v, self.param[k].default)
+        }
```
philippjfr, line 1384: Hmm, having trouble figuring out if this is really correct, it seems like since this is a `ParameterizedFunction` and instance values are never overridden then `non_default_params` will always be empty.

Hoxbro: One test fails when removing rasterize defaults: `test_segments_aggregate_sum_instance`. Which contains `rasterize.instance`. So by removing the default, this does not work:

``` python
import holoviews as hv
import numpy as np
import xarray as xr

hv.extension("bokeh")

da = xr.DataArray(
    np.arange(100.0).reshape(10, 10),
    coords=dict(x=np.arange(10.0), y=np.arange(10.0)),
)
ri = rasterize.instance(x_range=(-1, 10), y_range=(-1, 10))
ri(hv.Image(da))
```

Hoxbro: I have updated `non_default_params` to `instance_params` to take account for that in https://github.com/holoviz/holoviews/pull/5767/commits/046c2e0e63445f922f68b36ac025707be2318c4b. 

I have left `isequal` in as I feel it is a good util to have, even though it is not used in this PR anymore.

philippjfr (approved):
Looks good to me!

Issue #5517, opened 2022-11-22
test_plotutils.py::TestBokehUtils::test_py2js_funcformatter* tests fail


bnavigator:
#### ALL software version info

- holoviews 1.15.2 sdist from PyPI
- Python 3.8, 3.9, 3.10 on openSUSE Tumbleweed (openbuildservice, rpmbuild)
- bokeh 2.4.3
- pscript 0.7.7

#### Description of expected behavior and the observed behavior
Trying to update to rpm package of holoviews for openSUSE Tumleweed, I get the test suite failures below. I do not know enough JS in order to determine if the differently generated code is okay or why the long boilerplate is there. Am I missing a dependency, perhaps?

#### Complete, minimal, self-contained example code that reproduces the issue

```shell
docker run -it opensuse/tumbleweed
```

```shell
zypper ar https://download.opensuse.org/repositories/devel:/languages:/python:/numeric/openSUSE_Tumbleweed/devel:languages:python:numeric.repo
zypper in wget tar gzip python310-{base,pytest,colorcet,numpy,packaging,pandas,panel,param,pyviz-comms,bokeh2,pscript}
# (trust devel repo key)
wget https://files.pythonhosted.org/packages/source/h/holoviews/holoviews-1.15.2.tar.gz
tar xf holoviews-1.15.2.tar.gz
cd holoviews-1.15.2
pytest -vv holoviews/tests/plotting/test_plotutils.py -k "test_py2js_funcformatter"
```

#### Stack traceback and/or browser JavaScript console output
```
=================================================================== test session starts ====================================================================
platform linux -- Python 3.10.8, pytest-7.1.2, pluggy-1.0.0 -- /usr/bin/python3.10
cachedir: .pytest_cache
rootdir: /holoviews-1.15.2
collected 73 items / 70 deselected / 3 selected                                                                                                            

holoviews/tests/plotting/test_plotutils.py::TestBokehUtils::test_py2js_funcformatter_arg_and_kwarg FAILED                                            [ 33%]
holoviews/tests/plotting/test_plotutils.py::TestBokehUtils::test_py2js_funcformatter_single_arg FAILED                                               [ 66%]
holoviews/tests/plotting/test_plotutils.py::TestBokehUtils::test_py2js_funcformatter_two_args FAILED                                                 [100%]

========================================================================= FAILURES =========================================================================
__________________________________________________ TestBokehUtils.test_py2js_funcformatter_arg_and_kwarg ___________________________________________________

[...]

Issue #5327, opened 2022-06-03
Drop support of formatting Bokeh ticks with `pscript`


maximlt:
Ticks can be formatted by providing a function which will be compiled to JS using `pscript` (if installed) when using Bokeh. Bokeh has dropped support of this kind of approach based on `pscript`, and it has apparently never worked really well in HoloViews.

maximlt:
@jlstevens if you agree with that, which version would be a good target? 1.15 or later?

jlstevens:
Do you know when Bokeh dropped support? If it has been over a year or so, I think we can indeed drop this in 1.15.

maximlt:
It seems it was removed some time in 2019 for the 2.0 release of Bokeh which happened in March 2020: https://github.com/bokeh/bokeh/pull/8965

MarcSkovMadsen:
I'm actually using the support in some of my apps.

PR #5776, opened 2023-06-22
Deprecate functionality

Hoxbro:
This PR deprecated two items: 
- `BoundingRegionParameter` as discussed [here](https://github.com/holoviz/holoviews/pull/5768#discussion_r1233685836)
- py2js support as`py2js_tickformatter` (which we are not even testing right now)
   - #5327 
   - #5517

I also bumped ResamplingOperation to 1.18.

### DIFF
--- a/holoviews/core/boundingregion.py
+++ b/holoviews/core/boundingregion.py
@@ -10,6 +10,7 @@ File originally part of the Topographica project.
 import param
 from param.parameterized import get_occupied_slots
 from .util import datetime_types
+from ..util.warnings import deprecated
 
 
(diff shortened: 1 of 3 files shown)


philippjfr (approved):

philippjfr (approved):
This looks good. 

- `py2js` was removed in Bokeh looong ago and is not maintained so removing it is the only sensible move.
- `ResamplingOperation` was already deprecated
- `BoundingRegionParameter` was almost certainly never used by anything, it got inherited from topographica and is some of the oldest code in HoloViews

Issue #5789, opened 2023-07-01
Select_mask & decimate fail when dtype changes from numeric to datetime


chrisbotica:
#### ALL software version info
holoviews (working off of main)
panel 1.1.1
bokeh 3.1.1
tornado 6.3.2
python 3.10.8
same error on Windows 10/Replit unix/Mac 13.4

#### Description of expected behavior and the observed behavior

When changing the axis dtypes from datetime to float (and vice-versa) in a ```decimate(DynamicMap([]))``` two main bugs occur:
1. (in ```core/data/interface/select_mask```) The mask initialization fails to catch changing dtype (```not None```) 
2. (in ```operation/element/decimate```) The plot range in the decimate element (```self.p.range```) fails to update fast enough, as opposed to the element range (```element.range(0), element.range(1)```)

Will submit a pull request to address these two errors. 

Note that these errors are obscured by the larger DynamicMap-axis-change bug (#5405), but since I'm circumventing that by resetting the dmap (see background & code below), these two errors are uncovered. This is useful to fix now / keep in mind for later

Background: 
- Changing axis dtypes in a DynamicMap is super helpful, because it is a hv element and allows us to attach streams, make overlays, or apply any other hv element related function
- The current workaround in https://github.com/holoviz/holoviews/issues/5405#issuecomment-1225844874 from @Hoxbro is to use pn.bind to essentially reset the frame; however the output is a pn.bind function that cannot be further combined with holoviews streams (see https://github.com/holoviz/holoviews/issues/5788#issue-1784123763  )
- My workaround is to essentially reset the dynamic map using an external ```pn.bind``` function, all while not changing the frame. For more details, please see my comments here: https://github.com/holoviz/holoviews/issues/5591#issuecomment-1482250369

#### Complete, minimal, self-contained example code that reproduces the issue

Workaround that uncovers ```decimate``` issues. Note that when decimate is not used, there is no error.

```
import holoviews as hv
import panel as pn
import pandas as pd
import datetime
import numpy as np
from holoviews.operation import decimate
from datetime import datetime, timedelta
from holoviews.selection import link_selections

hv.extension('bokeh')

df = pd.DataFrame({'x1': np.random.rand(1000), 
                   'x2': np.random.rand(1000), 
                   't': np.array([datetime(2023,1,1) + timedelta(seconds=1) * i for i in range(1000)]),
                   'y' : np.random.rand(1000)})

x_sel = pn.widgets.Select(options=['x1', 'x2', 't'],
[...]

Issue #5405, opened 2022-08-24
DynamicMap fails when dtype changes on axis


GitRay:
I posted this as a question on Discourse a week ago, but did not get a response. I've played around some more in the last week and am pretty sure it is a bug and not usage error, so I'm posting it here.

Windows 10 Enterprise setup:
Chrome 104.0.5112.102
Python 3.9.13
Holoviews 1.15.0
Bokeh 2.4.3
jupyter_bokeh 3.0.4
jupyterlab 3.4.5

[Discourse writeup](https://discourse.holoviz.org/t/dynamicmap-fails-when-dtype-changes-on-axis/4126?u=gitray)

The above Discourse writeup has lots of detail, but I'll put a more concise version here.

If I create a simple DynamicMap where the only kdim is the column name of a DataFrame, I expect to be able to switch between column names and have the graph update the Y axis with the data in this column. This works fine as long as the data is of the same dtype that was originally used to construct the graph. If columns have mixed dtypes, the DynamicMap does not correctly update the axis type and so the render fails. 

Here is a sample of code which exhibits the behavior. This is the worst case, where changing to the column "class" results in an Exception. Most cases just result in undesired behavior of the chart.

```
import holoviews as hv
#hv.extension('matplotlib')
hv.extension('bokeh')
import pandas as pd

# creating a dataframe with both an int and an obj dtype
df = pd.DataFrame({
    'count':    [ 0,    1,    2 ], 
    'distance': [ 3,    4,    7 ],
    'class':    ['A',  'B',  'C']
})

# set up a DynamicMap where the user can select the y_axis column
y_dim = hv.Dimension('y_column', values=df.columns.tolist(), default='distance')
def return_scatter(col_name):
    return hv.Scatter(df, kdims='count', vdims=[col_name])
hv.DynamicMap(return_scatter, kdims=y_dim).opts(framewise=True, axiswise=True)
```

The above code results in an exception when "class" is selected:
Traceback (most recent call last):
  File "C:\Users\rcathcart\.conda\envs\py39\lib\site-packages\pyviz_comms\__init__.py", line 338, in _handle_msg
    self._on_msg(msg)
  File "C:\Users\rcathcart\.conda\envs\py39\lib\site-packages\panel\viewable.py", line 292, in _on_msg
    doc.unhold()
  File "C:\Users\rcathcart\.conda\envs\py39\lib\site-packages\bokeh\document\document.py", line 799, in unhold
    self.callbacks.unhold()
[...]

PR #5790, opened 2023-07-01
Enable changing dtypes for decimate

chrisbotica:
This fixes the issues uncovered in #5789 : Select_mask & decimate fail when dtype changes from numeric to date time. Note that without the workaround in #5789, the two bugs are first hidden by the issue in #5405..

This PR:
- updates the element interface mask to allow for ```np.nan```
  - The mask initialization fails to catch changing dtype (e.g., np.nans from datetime objects). "None" types will not account for these
- changes the decimate xy lims to current element ranges
  - For decimate, the plot range updates slower than the element range (e.g., whenever the x, y dtypes change from a dynamic map). The element range reflect the current ranges in the new dtype, whereas the self.p.range reflects the previous dtype

### DIFF
--- a/holoviews/core/data/interface.py
+++ b/holoviews/core/data/interface.py
@@ -354,9 +354,9 @@ class Interface(param.Parameterized):
             if isinstance(sel, slice):
                 with warnings.catch_warnings():
                     warnings.filterwarnings('ignore', r'invalid value encountered')
-                    if sel.start is not None:
+                    if sel.start is not None and sel.start is not np.nan:
                         mask &= sel.start <= arr
-                    if sel.stop is not None:
+                    if sel.stop is not None and sel.stop is not np.nan:
                         mask &= arr < sel.stop
             elif isinstance(sel, (set, list)):
                 iter_slcs = []
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -899,8 +899,8 @@ class decimate(Operation):
         if element.interface not in column_interfaces:
             element = element.clone(tuple(element.columns().values()))
 
-        xstart, xend = self.p.x_range if self.p.x_range else element.range(0)
-        ystart, yend = self.p.y_range if self.p.y_range else element.range(1)
+        xstart, xend = element.range(0)
+        ystart, yend = element.range(1)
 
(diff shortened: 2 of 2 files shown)


holoviews/operation/element.py
```diff
@@ -899,8 +899,8 @@ def _process_layer(self, element, key=None):
         if element.interface not in column_interfaces:
             element = element.clone(tuple(element.columns().values()))
 
-        xstart, xend = self.p.x_range if self.p.x_range else element.range(0)
-        ystart, yend = self.p.y_range if self.p.y_range else element.range(1)
+        xstart, xend = element.range(0)
```
philippjfr, line 902: This doesn't seem correct, it is important that you can pass in the required `x_range` and `y_range`

philippjfr (changes requested):
Thank you @chrisbotica! The changes for the nan handling seem good but passing in the x_range and y_range to select on is essential to the operation.

holoviews/operation/element.py
```diff
@@ -899,8 +899,8 @@ def _process_layer(self, element, key=None):
...
+        ystart, yend = element.range(1)
```
philippjfr:
```suggestion
```

holoviews/core/data/interface.py
```diff
@@ -358,9 +358,9 @@ def select_mask(cls, dataset, selection):
             if isinstance(sel, slice):
                 with warnings.catch_warnings():
                     warnings.filterwarnings('ignore', r'invalid value encountered')
-                    if sel.start is not None:
+                    if sel.start is not None and sel.start is not np.nan:
```
hoxbro:
```suggestion
                    if sel.start is not None and not np.isnan(sel.start):
```

holoviews/core/data/interface.py
```diff
@@ -358,9 +358,9 @@ def select_mask(cls, dataset, selection):
...
                         mask &= sel.start <= arr
-                    if sel.stop is not None:
+                    if sel.stop is not None and sel.stop is not np.nan:
```
hoxbro:
```suggestion
                    if sel.stop is not None and not np.isnan(sel.stop):
```

PR #5791, opened 2023-07-02
Update FileArchive repr for Param 2.0

maximlt:
Seen with @jlstevens when reviewing https://github.com/holoviz/param/pull/767, there was a remaining call to `Parameterized.pprint` that is going to be removed in Param 2.0. It's replaced by `Parameterized.param.pprint` that has been available since Param 1.12.0, 1.12.0 being the minimum version required by HoloViews currently so there's no need for any compatibility code.


Hoxbro:
LGTM

Issue #3486, opened 2019-02-11
Allow passing x, y velocities to VectorField (alternative to angle, magnitude)


poplarShift:
As discussed in https://github.com/pyviz/geoviews/issues/295, it would be nice to be able to pass x/y velocities to VectorField instead of angle and magnitude. On rectilinear grids it doesn't matter much, but in geographic contexts `angle` quickly becomes confusing.

However, in view of ultimate support for Datetime axes and stickplots and the like (see here: https://github.com/pyviz/holoviews/issues/2467 and here: http://www.whoi.edu/science/PO/coastal/ANDRILL_2010_Mooring/), keeping the angle/magnitude option makes sense too.

kcpevey:
+1   I'd also like the option of passing in either format.

rsignell-usgs:
The reason other vector plotting routines (like [matplotlib quiver](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.quiver.html)) use U and V instead of speed and direction are:
* those are the variables actually used in numerical weather, ocean and hydro models
* they don't suffer from the same ambiguity that direction does.   Is it direction using Math convention? Met convention?  Ocean convention?

I actually think we should deprecate spd and dir in favor of U and V.

poplarShift:
> I actually think we should deprecate spd and dir in favor of U and V.

Last time this was up there seemed to be a consensus that stickplots (#3504) and other vectorfields where the x/y coordinates do not necessarily occupy the same space as the arrows (when there is a datetime axis, for instance) should be handled as part of the same element. But otherwise I agree.

PR #5797, opened 2023-07-06
Allow vectorfield from U/V

ahuang11:
Closes https://github.com/holoviz/holoviews/issues/3486

image

```python
import pandas as pd
import numpy as np
import holoviews as hv
hv.extension("bokeh")

x = np.linspace(-1, 1, 4)
X, Y = np.meshgrid(x, x)
U, V = 5 * X, 5 * Y

df = pd.DataFrame({
    "x": X.flatten(),
    "y": Y.flatten(),
    "u": U.flatten(),
    "v": V.flatten(),
})
angle = np.pi / 2 - np.arctan2(U, V)
mag = np.hypot(U, V)

hv.VectorField.from_uv(df, ["x", "y"], ["u", "v"])
```

### DIFF
--- a/holoviews/element/geom.py
+++ b/holoviews/element/geom.py
@@ -52,6 +52,35 @@ class VectorField(Selection2DExpr, Geometry):
     vdims = param.List(default=[Dimension('Angle', cyclic=True, range=(0,2*np.pi)),
                                 Dimension('Magnitude')], bounds=(1, None))
 
+    @classmethod
+    def from_uv(cls, data, kdims=None, vdims=None, **params):
+        if isinstance(data, tuple):
+            xs, ys, us, vs = data
+        else:
+            us = data[vdims[0]]
+            vs = data[vdims[1]]
+
+        uv_magnitudes = np.hypot(us, vs)  # unscaled
+        radians = np.pi / 2 - np.arctan2(us, vs)
+
+        if isinstance(data, tuple):
+            reorganized_data = (xs, ys, radians, uv_magnitudes)
+        else:
+            # calculations on this data could mutate the original data
+            # here we do not do any calculations; we only store the data
+            reorganized_data = {}
+            for kdim in kdims:
+                reorganized_data[kdim] = data[kdim]
+            reorganized_data["Angle"] = radians
+            reorganized_data["Magnitude"] = uv_magnitudes
+            for vdim in vdims[2:]:
+                reorganized_data[vdim] = data[vdim]
+            vdims = [
+                Dimension('Angle', cyclic=True, range=(0, 2 * np.pi)),
+                Dimension('Magnitude')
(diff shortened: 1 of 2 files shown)


holoviews/element/geom.py
```diff
@@ -52,6 +52,35 @@ class VectorField(Selection2DExpr, Geometry):
...
+    @classmethod
+    def from_uv(cls, data, kdims=None, vdims=None, **params):
+        if isinstance(data, tuple):
+            xs, ys, us, vs = data
+        else:
+            us = data[vdims[0]]
+            vs = data[vdims[1]]
+
+        uv_magnitudes = np.hypot(us, vs)  # unscaled
+        radians = np.pi / 2 - np.arctan2(us, vs)
```
Hoxbro, line 64: Why not `radians = np.pi / 2 - np.arctan2(-us, -vs)`?

ahuang11: Vectorfield conventions are "to"

Hoxbro: Should we add a comment about this?

ahuang11:
```python
import numpy as np

x = 0
y = 5
print(
    np.rad2deg(np.arctan2(-y, -x)),  # meteorological (from)
    np.rad2deg(np.arctan2(-y, -x) + np.pi),  # oceanographic (to)
    np.rad2deg(np.arctan2(x, y)), # mathematical
)
```
-90.0 90.0 0.0

holoviews/element/geom.py
```diff
@@ -52,6 +52,37 @@ class VectorField(Selection2DExpr, Geometry):
     vdims = param.List(default=[Dimension('Angle', cyclic=True, range=(0,2*np.pi)),
                                 Dimension('Magnitude')], bounds=(1, None))
 
+    @classmethod
+    def from_uv(cls, data, kdims=None, vdims=None, **params):
+        if isinstance(data, tuple):
+            xs, ys, us, vs = data
+        else:
+            us = data[vdims[0]]
```
philippjfr, line 60: This doesn't seem safe, `vdims` could contain Dimensions.

holoviews/element/geom.py
```diff
@@ -52,6 +52,37 @@ class VectorField(Selection2DExpr, Geometry):
     vdims = param.List(default=[Dimension('Angle', cyclic=True, range=(0,2*np.pi)),
                                 Dimension('Magnitude')], bounds=(1, None))
 
+    @classmethod
+    def from_uv(cls, data, kdims=None, vdims=None, **params):
```
philippjfr, line 56: `data` here could be one of N many formats and handling them all is basically impossible. What I would suggest is that you create a Dataset from the kdims and vdims and then use `.dimension_values()` on it to look up the arrays.

PR #5800, opened 2023-07-10
Add args for initial ranges to RangeToolLink

### LINKED ISSUE #5437: Set initial selected range for RangeToolLink
#### Is your feature request related to a problem? Please describe.

As described on [stackoverflow](https://stackoverflow.com/questions/73647097/predefine-selection-of-holoviews-rangetoollink), users want to adjust the initial selection range in the RangeToolLink, rather than having it select the entire range from the start. There is a solution using hooks (see the [accepted answer](https://stackoverflow.com/a/73649363/3149975)), however, there should be a simpler way to do it.

#### Describe the solution you'd like
For the axes being linked, allow a start and end value in some way. A complication is that more than one axis is allowed to be linked.

One possible solution that avoids axes' names from having to be args:
`RangeToolLink(source, target, axes=['x', 'y'], start = {'x':1, 'y':2}, end={'x':3, 'y':4})`

droumis:
fixes #5437

Allows users to specify a start and end to the axes ranges in the call to RangeToolLink. Previously, this required a hook (#5437).

**update:** changed args and logic to boundsx/y rather than axes_start/end

```python
import pandas as pd
import holoviews as hv
from holoviews import opts

hv.extension('bokeh')
from bokeh.sampledata.stocks import AAPL

aapl_df = pd.DataFrame(AAPL['close'], columns=['close'], index=pd.to_datetime(AAPL['date']))
aapl_df.index.name = 'Date'

aapl_curve = hv.Curve(aapl_df, 'Date', ('close', 'Price ($)'))

tgt = aapl_curve.relabel('AAPL close price').opts(width=800, labelled=['y'], toolbar='disable')

src = aapl_curve.opts(width=800, height=100, yaxis=None, default_tools=[])

RangeToolLink(src, tgt, axes=['x', 'y'], boundsx= (pd.to_datetime('2006'), pd.to_datetime('2010')), boundsy=(None, 400))

layout = (tgt + src).cols(1)
layout.opts(opts.Layout(shared_axes=False, merge_tools=False))
```

image

### DIFF
--- a/examples/gallery/demos/bokeh/timeseries_range_tool.ipynb
+++ b/examples/gallery/demos/bokeh/timeseries_range_tool.ipynb
@@ -63,7 +63,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "tgt = aapl_curve.relabel('AAPL close price').opts(width=800, labelled=['y'], toolbar='disable')\n",
+    "tgt = aapl_curve.relabel('AAPL close price').opts(width=800, height=300, labelled=['y'], toolbar='disable')\n",
     "src = aapl_curve.opts(width=800, height=100, yaxis=None, default_tools=[])\n",
     "\n",
     "RangeToolLink(src, tgt)\n",
@@ -71,14 +71,63 @@
     "layout = (tgt + src).cols(1)\n",
     "layout.opts(opts.Layout(shared_axes=False, merge_tools=False))"
    ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Specify range"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "Use the `axes_start` and `axes_end` parameters to specify the initial range of the ``RangeTool``:"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "tgt = aapl_curve.relabel('AAPL close price').opts(width=800, height=300, labelled=['y'], toolbar='disable')\n",
+    "src = aapl_curve.opts(width=800, height=100, yaxis=None, default_tools=[])\n",
+    "\n",
+    "RangeToolLink(src, tgt, axes_start={'x':pd.to_datetime('2010')}, axes_end={'x': pd.to_datetime('2012')})\n",
+    "\n",
+    "layout = (tgt + src).cols(1)\n",
+    "layout.opts(opts.Layout(shared_axes=False, merge_tools=False))"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
   }
  ],
  "metadata": {
+  "kernelspec": {
+   "display_name": "Python 3 (ipykernel)",
+   "language": "python",
+   "name": "python3"
+  },
   "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
    "name": "python",
-   "pygments_lexer": "ipython3"
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.11.4"
   }
  },
  "nbformat": 4,
- "nbformat_minor": 2
+ "nbformat_minor": 4
 }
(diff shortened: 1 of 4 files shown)


holoviews/plotting/bokeh/links.py
```diff
@@ -138,10 +138,16 @@ class RangeToolLinkCallback(LinkCallback):
...
+        for axis in ['x', 'y']:
+            if axis in link.axes:
+                axes[f'{axis}_range'] = target_plot.handles[f'{axis}_range']
+                bounds = getattr(link, f'bounds{axis}', None)
+                if bounds is not None:
+                    start, end = bounds
+                    if start is not None:
+                        axes[f'{axis}_range'].start = start
+                    if end is not None:
+                        axes[f'{axis}_range'].end = end
```
ahuang11, line 150: I wonder if we can reduce the indentation like this?
```python
for axis in ['x', 'y']:
    if axis not in link.axes:
        continue

    axes[f'{axis}_range'] = target_plot.handles[f'{axis}_range']
    bounds = getattr(link, f'bounds{axis}', None)
    if bounds is None:
        continue

    start, end = bounds
    if start is not None:
        axes[f'{axis}_range'].start = start
    if end is not None:
        axes[f'{axis}_range'].end = end
```


(Relevant video explanation: https://www.youtube.com/watch?v=CFRhGnuXG-4)

holoviews/plotting/links.py
```diff
@@ -104,6 +104,12 @@ class RangeToolLink(Link):
     axes = param.ListSelector(default=['x'], objects=['x', 'y'], doc="""
         Which axes to link the tool to.""")
 
+    boundsx = param.Tuple(default=None, length=2, doc="""
+        (start, end) bounds for the x axis""")
+
+    boundsy = param.Tuple(default=None, length=2, doc= """
+        (start, end) bounds for the y axis""")
```
ahuang11, line 111:
```suggestion
        (start, end) bounds for the y-axis""")
```

holoviews/plotting/bokeh/links.py
droumis: thanks! this is a really helpful concept and video :)

holoviews/plotting/links.py
```diff
@@ -104,6 +104,12 @@ class RangeToolLink(Link):
     axes = param.ListSelector(default=['x'], objects=['x', 'y'], doc="""
         Which axes to link the tool to.""")
 
+    boundsx = param.Tuple(default=None, length=2, doc="""
+        (start, end) bounds for the x axis""")
```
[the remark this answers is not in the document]
droumis: added, thanks!

droumis:
one quirk - when you press the 'reset' button in the Bokeh toolbar, users would probably expect the plot to go back to the initial state that was displayed with constrained rangetool bounds... However, upon 'reset', the rangetool currently fills the whole data space. 

I don't think this should block this PR.. just to note for a future enhancement

examples/gallery/demos/bokeh/timeseries_range_tool.ipynb
Hoxbro, line 1: I would consider using pure `datetime` objects.

holoviews/plotting/links.py
```diff
@@ -104,6 +104,12 @@ class RangeToolLink(Link):
     axes = param.ListSelector(default=['x'], objects=['x', 'y'], doc="""
         Which axes to link the tool to.""")
 
+    boundsx = param.Tuple(default=None, length=2, doc="""
+        (start, end) bounds for the x-axis""")
+
+    boundsy = param.Tuple(default=None, length=2, doc= """
```
Hoxbro, line 110:
```suggestion
    boundsy = param.Tuple(default=None, length=2, doc="""
```

Very _very_ nit. Only do this if you make other changes...

Hoxbro (approved):
Some small comments. That definitely are not blockers.

examples/gallery/demos/bokeh/timeseries_range_tool.ipynb
droumis: thanks for the feedback. I'm assuming you mean something like this:

`RangeToolLink(src, tgt, axes=['x', 'y'], boundsx= (datetime.datetime.strptime('2006', '%Y'), datetime.datetime.strptime('2010', '%Y')), boundsy=(None, 400))`

For my own understanding, why do you recommend using datetime instead of pd here?

Hoxbro: I would recommend: 
``` 
from datetime import datetime
... # Other stuff
boundsx=(datetime(2006, 1, 1) , datetime(2010, 1, 1))
```

The reason is only that it is more explicit.

Hoxbro:
Thank you for the PR. Great job!

PR #5802, opened 2023-07-12
Update Build and Docs CI for dev release

Hoxbro:
Fixes failing CI build and docs

- Also update docs to use holoviz_tasks/install
- Uploaded thumbnails to S3 so it doesn't need to create them for each documentation build. 
- Also fixing a problem with https://holoviews.org/gallery/demos/matplotlib/network_graph.html erroring out

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -27,7 +27,7 @@ jobs:
     timeout-minutes: 120
     defaults:
       run:
-        shell: bash -l {0}
+        shell: bash -el {0}
     env:
       DESC: "Documentation build"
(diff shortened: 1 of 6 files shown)


Hoxbro:
I think we have a perfect storm of things not working right now:

On Mac Mamba fails with `Library not loaded: '@rpath/libarchive.13.dylib'`
https://github.com/holoviz/holoviews/actions/runs/5529593220/jobs/10087878559#step:2:2137

and on Windows Conda fails with ` Warning: ERROR during elevated execution.`
https://github.com/holoviz/holoviews/actions/runs/5529593220/jobs/10087878828#step:2:1363

PR #5805, opened 2023-07-13
Support `ds.where` and `ds.summary` and add selector 

Hoxbro:
Example code:

``` python
import datashader as ds
import holoviews as hv
import numpy as np
import pandas as pd

hv.extension("bokeh")

num = 10000
np.random.seed(1)

dists = {
    cat: pd.DataFrame(
        {
            "x": np.random.normal(x, s, num),
            "y": np.random.normal(y, s, num),
            "s": s,
            "val": val,
            "cat": cat,
        }
    )
    for x, y, s, val, cat in [
        (2, 2, 0.03, 0, "d1"),
        (2, -2, 0.10, 1, "d2"),
        (-2, -2, 0.50, 2, "d3"),
        (-2, 2, 1.00, 3, "d4"),
        (0, 0, 3.00, 4, "d5"),
    ]
}

agg = ds.where(ds.min("s"))

plot = rasterize(hv.Points(df), aggregator=agg).opts(
    tools=["hover"], colorbar=True, width=500
)
```
With `agg = ds.where(ds.min("val"))`:

https://github.com/holoviz/holoviews/assets/19758978/930da0d1-674c-4aac-9c50-07065197cc1b

With `agg = ds.where(ds.min("s"))`:

https://github.com/holoviz/holoviews/assets/19758978/8559aae1-f0da-43ed-a0af-c57f23157c35

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -390,11 +390,6 @@ class aggregate(LineAggregationOperation):
             return NdOverlay(layers, kdims=[data.get_dimension(agg_fn.column)])
 
     def _apply_datashader(self, dfdata, cvs_fn, agg_fn, agg_kwargs, x, y):
-        if isinstance(agg_fn, ds.where):
-            # Only calculating the index no matter the column
-            # We will calculate the value later
-            agg_fn = ds.where(agg_fn.selector)
-
         # Suppress numpy warning emitted by dask:
         # https://github.com/dask/dask/issues/8439
         with warnings.catch_warnings():
@@ -405,7 +400,7 @@ class aggregate(LineAggregationOperation):
             agg = cvs_fn(dfdata, x.name, y.name, agg_fn, **agg_kwargs)
 
         is_where_index = ds15 and isinstance(agg_fn, ds.where) and isinstance(agg_fn.column, rd.SpecialColumn)
-        is_summary_index = isinstance(agg_fn, ds.summary) and "index" in agg.data
+        is_summary_index = isinstance(agg_fn, ds.summary) and "index" in agg
         if is_where_index or is_summary_index:
             if is_where_index:
                 data = agg.data
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -164,6 +164,8 @@ class AggregationOperation(ResampleOperation2D):
             col = agg_fn.column if not isinstance(agg_fn.column, rd.SpecialColumn) else agg_fn.selector.column
             vdims = sorted(params["vdims"], key=lambda x: x != col)
             # TODO: Should we add prefix to all of the where columns.
+        elif agg_name == "Summary":
+            vdims = list(agg_fn.keys)
         elif column:
             dims = [d for d in element.dimensions('ranges') if d == column]
             if not dims:
@@ -185,8 +187,6 @@ class AggregationOperation(ResampleOperation2D):
                 vdims.nodata = 0
         else:
             vdims = Dimension(f'{vdim_prefix}{agg_name}', label=agg_name, nodata=0)
-        if agg_name == "Summary":
-            vdims = list(agg_fn.keys)
 
         params['vdims'] = vdims
         return params
@@ -363,9 +363,9 @@ class aggregate(LineAggregationOperation):
                 params["vdims"] = [params["vdims"]]
             sum_agg = ds.summary(**{str(params["vdims"][0]): agg_fn, "index": ds.where(sel_fn)})
             agg = self._apply_datashader(dfdata, cvs_fn, sum_agg, agg_kwargs, x, y)
-            # First sel is index, which we don't care
             sel_prefix = f"{type(sel_fn).__name__}_{sel_fn.column}"
-            sel_vdims = [Dimension(s, label=f"{sel_prefix} {s}") for s in list(agg)[1:] if s not in params["vdims"]]
+            _ignore = [*params["vdims"], "index"]
+            sel_vdims = [Dimension(s, label=f"{sel_prefix} {s}") for s in agg if s not in _ignore]
             params["vdims"] = [*params["vdims"], *sel_vdims]
         else:
             agg = self._apply_datashader(dfdata, cvs_fn, agg_fn, agg_kwargs, x, y)
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -363,9 +363,8 @@ class aggregate(LineAggregationOperation):
                 params["vdims"] = [params["vdims"]]
             sum_agg = ds.summary(**{str(params["vdims"][0]): agg_fn, "index": ds.where(sel_fn)})
             agg = self._apply_datashader(dfdata, cvs_fn, sum_agg, agg_kwargs, x, y)
-            sel_prefix = f"{type(sel_fn).__name__}_{sel_fn.column}"
             _ignore = [*params["vdims"], "index"]
-            sel_vdims = [Dimension(s, label=f"{sel_prefix} {s}") for s in agg if s not in _ignore]
+            sel_vdims = [s for s in agg if s not in _ignore]
             params["vdims"] = [*params["vdims"], *sel_vdims]
         else:
             agg = self._apply_datashader(dfdata, cvs_fn, agg_fn, agg_kwargs, x, y)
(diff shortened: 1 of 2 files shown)


holoviews/operation/datashader.py
```diff
@@ -148,7 +153,13 @@ def _get_agg_params(self, element, x, y, agg_fn, bounds):
             category = agg_fn.cat_column
             agg_fn = agg_fn.reduction
         column = agg_fn.column if agg_fn else None
-        if column:
+        agg_name = type(agg_fn).__name__.title()
+        if agg_name == "Where":
+            # Set the first item to be the selector column.
+            col = agg_fn.column if not isinstance(agg_fn.column, rd.SpecialColumn) else agg_fn.selector.column
+            vdims = sorted(params["vdims"], key=lambda x: x == col, reverse=True)
```
Hoxbro, line 160:
```suggestion
            vdims = sorted(params["vdims"], key=lambda x: x != col)
```

jlstevens: Makes sense to me!

jlstevens:
Looking good!

holoviews/operation/datashader.py
```diff
@@ -163,13 +186,11 @@ def _get_agg_params(self, element, x, y, agg_fn, bounds):
             else:
                 vdims = dims[0].clone(vdim_prefix + column)
         elif category:
-            agg_name = type(agg_fn).__name__.title()
             agg_label = f'{category} {agg_name}'
```
philippjfr, line 189: Ah, I see now. Ignore the other suggestions I guess.

holoviews/operation/datashader.py
```diff
@@ -58,13 +59,19 @@ class AggregationOperation(ResampleOperation2D):
...
-    aggregator = param.ClassSelector(class_=(ds.reductions.Reduction, str),
-                                     default=ds.count(), doc="""
+    aggregator = param.ClassSelector(class_=(rd.Reduction, rd.summary, str),
+                                     default=rd.count(), doc="""
         Datashader reduction function used for aggregating the data.
         The aggregator may also define a column to aggregate; if
         no column is defined the first value dimension of the element
         will be used. May also be defined as a string.""")
 
+    selector = param.ClassSelector(class_=(rd.min, rd.max, rd.first, rd.last),
```
philippjfr, line 69: Assume I'm stupid, can you explain to me how aggregator, selector and where as an aggregator interact?

Hoxbro: A selector is a subset of the aggregator function, which selects an item for the data / dataframe. With this PR, we support `rd.first`, `rd.last`, `rd.min`, and `rd.max` in HoloViews. For the next examples, we have a dataframe with two columns `s` and `val`, which are both numeric.

`rd.where` can use a selector to look up an index. `rd.where(rd.min("val"))` will find the index of the minimum value. By passing in a column like `rd.where(rd.min("val"), "s")` will find the s-value where the `val` has its minimum value.

The selectors can be passed into the rasterize-aggregator parameter with or without the where. Suppose we want to use an aggregator which is not a selector like `rd.Count` but still wants some information about the underlying data. We will use the rasterize-selector parameter and pass in `ds.first("val")`. Behind the scene, this will then use `rd.summary` to combine the aggregator `rd.Count` with `ds.where(ds.first("val"))` and only do one pass one data shader pass on data and return the output of both.

A special thing is done in HoloViews when passing in `rd.where(rd.first("val")` into rasterize, where it will not return the index itself but all of the row data for that index. This means in the previous example, we will get a hover tool containing `Count` (from the aggregator), `val`, and `s` (from the selector).

philippjfr: Thank you, **I think** that all makes sense to me but we will have to have detailed examples to explain this.

philippjfr (approved):

jlstevens:
I'll be taking a closer look soon but my initial question is whether you have any idea how `first_n` and `last_n` might work as selectors? I suppose the `DataSet` could have `m columns x n selector` layers? While this would quickly blow up  for wide data with lots of columns (or large `n`) I assume this is the natural extension?

Alternatively, `first` and `last` should probably behave like `first_n` and `last_n` where `n=1`...

holoviews/operation/datashader.py
```diff
@@ -338,22 +360,27 @@ def _process(self, element, key=None):
...
-            )
-            agg = getattr(cvs, glyph)(dfdata, x.name, y.name, agg_fn, **agg_kwargs)
+        cvs_fn = getattr(cvs, glyph)
+
+        if sel_fn:
+            if isinstance(params["vdims"], (Dimension, str)):
+                params["vdims"] = [params["vdims"]]
+            sum_agg = ds.summary(**{str(params["vdims"][0]): agg_fn, "index": ds.where(sel_fn)})
+            agg = self._apply_datashader(dfdata, cvs_fn, sum_agg, agg_kwargs, x, y)
+            _ignore = [*params["vdims"], "index"]
```
jlstevens, line 370: While it is a good default, the name "index" is a magic value. I could be wrong but couldn't this clash with a column name?

holoviews/operation/datashader.py
```diff
@@ -365,6 +392,39 @@ def _process(self, element, key=None):
...
+                data = agg.index.data
+            neg1 = data == -1
+            for col in dfdata.columns:
+                if col in agg.coords:
+                    continue
+                val = dfdata[col].values[data]
+                if val.dtype.kind == 'f':
+                    val[neg1] = np.nan
+                elif val.dtype.kind == "O":
+                    val[neg1] = "-"
```
jlstevens, line 421: This line was causing me problems for a categorical 'language' column in an example I demoed - I had to work around it by appending `"-"` as a row so it would appear as a possible category value.

Is this fixed? I guess the point of this is that when there is no data to aggregate you need some sort of special 'no category' value - the equivalent of `NaN` for categories...

holoviews/operation/datashader.py
jlstevens: @Hoxbro can correct me if I am wrong, but here is my simplified understanding of how to think about this:

The selector is used to specify which index value is used to build all the `Dataset` layers (e.g `first` or `max` on some dimension) and this needs to be combined with the `aggregator` (which may or may not be a selector) in `summary` to do a single pass. That way the aggregator (e.g. `mean`) which is not a selector can be put on the bottom layer (to make it visible and plotted) in the `Dataset` and the rest of the layers (computed from the selector) comes from the supplied selector.

What `where` does is make something like `max('foo')` go from giving you the max value of `'foo'` to the max *index* of foo. This is done automatically and internally when supplying `max('foo')` as the selector.

Hope that makes sense!

What I don't yet get is what happens if you *explicitly* supply a `where` aggregator as a selector. I think this is what @Hoxbro is talking about with '... where it will not return the index itself but all of the row data for that index.' but I have to admit I don't really understand what that means yet...

jbednar: > A special thing is done in HoloViews when passing in rd.where(rd.first("val") into rasterize, where it will not return the index itself but all of the row data for that index. 

I think this needs to be configurable; returning all the row data provides good pixel-to-pixel interactivity but if there are many rows or some rows are structured data like polygons or text the result may slow down the aggregation step significantly. So I think we need a good default here that someone can override when they disagree with the default. 

My gut tells me that returning the index and then dynamically requesting the row will give the best user experience, in that it makes all the information available without any special setup and without slowing down aggregations. But I don't think that's what it does now (and would require a custom hover tool?), and also won't always be optimal, since when there are only a couple of hover fields and they are simple scalars (e.g. categories) then it makes sense to send that data to the client as the complete screen-sized array. Plus then if it's not a dynamic plot hover will still be fully supported. 

So, I think this behavior should be configurable, and after it's configurable we can decide on the default, which my _guess_ (without actually trying it) would be the index array rather than all the row's columns.

jbednar:
> While this would quickly blow up for wide data with lots of columns (or large n) I assume this is the natural extension?

With the index array approach, first_3 should always only be 3x larger than the aggregate array, while with the approach of returning all columns the size and time taken would scale with the number and size of columns.  Seems unsafe as a default!  

An intermediate approach could be to return a fixed number of scalar columns only (up to e.g. 3) by default, but that seems quite arbitrary.

holoviews/operation/datashader.py
philippjfr: I'm not totally following the issue here. You're saying it would error because `'-'` didn't exist as a category? What precisely errored?

philippjfr:
I'd say we handle the `first_n` and `last_n` cases later.

holoviews/operation/datashader.py
philippjfr: Looks like @Hoxbro resolved this.

philippjfr:
I'm going to merge. I'll look into the first_n and last_n thing separately.

Issue #5807, opened 2023-07-14
`responsive=True` alongside `rasterize=True` breaks dynamic


ahuang11:
The plot no longer updates after zooming in; can be reproduced by serving on browser (not just in VSCode)

```python
import numpy as np
import pandas as pd
import hvplot.pandas

df = pd.DataFrame(np.random.multivariate_normal((0,0), [[0.1, 0.1], [0.1, 1.0]], (500000,)))
points = df.hvplot.points('0', '1', rasterize=True, tiles=True, global_extent=True).opts(responsive=True)
hvplot.show(points)
```

```
Traceback (most recent call last):
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/plotting/bokeh/callbacks.py", line 317, in process_on_event
    self.on_msg(msg)
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/plotting/bokeh/callbacks.py", line 187, in on_msg
    raise e
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/plotting/bokeh/callbacks.py", line 179, in on_msg
    Stream.trigger(streams)
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/streams.py", line 186, in trigger
    subscriber(**dict(union))
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/plotting/plot.py", line 231, in refresh
    raise e
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/plotting/plot.py", line 227, in refresh
    self._trigger_refresh(stream_key)
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/plotting/plot.py", line 244, in _trigger_refresh
    self.update(key)
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/plotting/plot.py", line 937, in update
    item = self.__getitem__(key)
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/plotting/plot.py", line 423, in __getitem__
    self.update_frame(frame)
[...]

Issue #5425, opened 2022-09-06
Support responsive=True, data_aspect=True, and datashader operations


Hoxbro:
#### Is your feature request related to a problem? Please describe.
I will like to be able to use `responsive=True`, `data_aspect=True`, and datashader operations. Right now, this raises a `TypeError`. See the example below.

This _could_ be seen as a bug, but I will mark it as a feature request.

#### Additional context

<details>

<summary> Code </summary>

``` python
import xarray as xr
import numpy as np
import holoviews as hv
from holoviews.operation.datashader import rasterize

hv.extension("bokeh")

da = xr.DataArray(
    np.arange(10000).reshape(100, 100),
    coords=dict(
        x=np.linspace(0, 3000, 100), y=np.linspace(0, 3000, 100)
    ),
)

image = hv.Image(da).opts(responsive=True, data_aspect=True)
image

rasterize(image)
```
</details>

<details>

<summary> Traceback </summary>

``` python-traceback
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
File ~/miniconda3/envs/holoviz/lib/python3.9/site-packages/IPython/core/formatters.py:973, in MimeBundleFormatter.__call__(self, obj, include, exclude)
    970     method = get_real_method(obj, self.print_method)
    972     if method is not None:
--> 973         return method(include=include, exclude=exclude)
    974     return None
    975 else:

File ~/Development/holoviz/repos/holoviews/holoviews/core/dimension.py:1294, in Dimensioned._repr_mimebundle_(self, include, exclude)
   1287 def _repr_mimebundle_(self, include=None, exclude=None):
   1288     """
   1289     Resolves the class hierarchy for the class rendering the
   1290     object using any display hooks registered on Store.display
   1291     hooks.  The output of all registered display_hooks is then
   1292     combined and returned.
   1293     """
-> 1294     return Store.render(self)

File ~/Development/holoviz/repos/holoviews/holoviews/core/options.py:1418, in Store.render(cls, obj)
   1416 data, metadata = {}, {}
   1417 for hook in hooks:
-> 1418     ret = hook(obj)
   1419     if ret is None:
   1420         continue

File ~/Development/holoviz/repos/holoviews/holoviews/ipython/display_hooks.py:277, in pprint_display(obj)
    275 if not ip.display_formatter.formatters['text/plain'].pprint:
    276     return None
--> 277 return display(obj, raw_output=True)

[...]

PR #5808, opened 2023-07-14
Fix aspect handling when plot size is still unknown

philippjfr:
Fixes https://github.com/holoviz/holoviews/issues/5807
Fixes https://github.com/holoviz/holoviews/issues/5425


Hoxbro:
This also looks like it could fix https://github.com/holoviz/holoviews/issues/5425

Issue #4910, opened 2021-04-26
datashading error


ahuang11:
Sometimes happens:
```
  File "~/anaconda3/envs/wxqsee/lib/python3.7/site-packages/holoviews/core/sheetcoords.py", line 166, in __init__
    r1,r2,c1,c2 = Slice._boundsspec2slicespec(self.lbrt,self)
  File "~/anaconda3/envs/wxqsee/lib/python3.7/site-packages/holoviews/core/sheetcoords.py", line 519, in _boundsspec2slicespec
    t_idx = int(np.ceil(t_m-0.5))
ValueError: cannot convert float NaN to integer
```

jbednar:
Any way to reproduce?  Presumably t_m here is NaN, implying that the bounds are NaN, so I'd suspect an empty plot is being encountered somehow?

ahuang11:
Zoom too close on some points that are spread apart I think?

jbednar:
If you can find an example plot we can run and give a step-by-step guide, we can check it out!

ahuang11:
Maybe not related to datashader, maybe related to https://discourse.holoviz.org/t/show-individual-points-when-zoomed-else-datashade/2204

ahuang11:
Okay I got a reproducible example; has to do with responsive=True.

Basically go wild zooming (zoom in really fast and out) then the error should pop up eventually.

```
import datashader as ds
import holoviews as hv
import geoviews as gv
import numpy as np
import cartopy.crs as ccrs
import pandas as pd
import panel as pn
from holoviews.operation.datashader import rasterize
hv.extension("bokeh")


df = pd.DataFrame(np.random.rand(1000, 3), columns=["x", "y", "z"]) * 10

points = gv.Points(df, kdims=["x", "y"], vdims=["z"], crs=ccrs.PlateCarree())

def filter_points(points, x_range, y_range):
    if x_range is None or y_range is None:
        return points
    return points[x_range, y_range]

def hover_points(points, threshold=50):
    if len(points) > threshold:
        return points.iloc[:0]
    return points

range_stream = hv.streams.RangeXY(source=points)
streams=[range_stream]

filtered = points.apply(filter_points, streams=streams)
shaded = rasterize(filtered, streams=streams, aggregator="max", x_sampling=0.25, y_sampling=0.25)
hover = filtered.apply(hover_points)

dynamic_hover = (shaded * hover)

[...]

PR #5809, opened 2023-07-14
Prevent edge cases in XArrayInterface from crashing

ahuang11:
Closes https://github.com/holoviz/geoviews/issues/663 and https://github.com/holoviz/holoviews/issues/4910

The examples in the issues above no longer crashes. I suspect datashade was peddling numpy arrays into XArrayInterface, but there was no handling for that so this PR implements that.

Also, there was NaNs in the bounds and casting to int raised an exception; this simply sets NaNs to 0s to prevent it from crashing.

### DIFF
--- a/holoviews/core/sheetcoords.py
+++ b/holoviews/core/sheetcoords.py
@@ -515,13 +515,18 @@ class Slice(np.ndarray):
         t_m,l_m = scs.sheet2matrix(l,t)
         b_m,r_m = scs.sheet2matrix(r,b)
 
-        l_idx = int(np.ceil(l_m-0.5))
-        t_idx = int(np.ceil(t_m-0.5))
+        l_idx = np.ceil(l_m-0.5)
+        t_idx = np.ceil(t_m-0.5)
         # CBENHANCEMENT: Python 2.6's math.trunc()?
-        r_idx = int(np.floor(r_m+0.5))
-        b_idx = int(np.floor(b_m+0.5))
+        r_idx = np.floor(r_m+0.5)
+        b_idx = np.floor(b_m+0.5)
 
-        return t_idx,b_idx,l_idx,r_idx
+        # sometimes contains NaNs so just set them to 0
+        # rather than raising an exception
+        indices = np.array([t_idx, b_idx, l_idx, r_idx])
+        indices[np.isnan(indices)] = 0
+
+        return tuple(indices.astype(int))
 
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


holoviews/core/sheetcoords.py
```diff
@@ -515,13 +515,18 @@ def _boundsspec2slicespec(boundsspec,scs):
         t_m,l_m = scs.sheet2matrix(l,t)
         b_m,r_m = scs.sheet2matrix(r,b)
 
-        l_idx = int(np.ceil(l_m-0.5))
-        t_idx = int(np.ceil(t_m-0.5))
+        l_idx = np.ceil(l_m-0.5)
+        t_idx = np.ceil(t_m-0.5)
         # CBENHANCEMENT: Python 2.6's math.trunc()?
```
Hoxbro, line 520: I think this can safely be removed :slightly_smiling_face:

holoviews/core/sheetcoords.py
```diff
@@ -515,13 +515,18 @@ def _boundsspec2slicespec(boundsspec,scs):
...
-        r_idx = int(np.floor(r_m+0.5))
-        b_idx = int(np.floor(b_m+0.5))
+        r_idx = np.floor(r_m+0.5)
+        b_idx = np.floor(b_m+0.5)
 
-        return t_idx,b_idx,l_idx,r_idx
+        # sometimes contains NaNs so just set them to 0
+        # rather than raising an exception
```
Hoxbro, line 525: I'm not completely sure if this is the right thing to do. If our lbrt coordinates are nans, isn't that usually a sign something has gone wrong beforehand?

Do you still see this with the changes in `holoviews/core/data/xarray.py`?

ahuang11: You're right; it seems I don't have this issue anymore. I'll remove this then.

holoviews/core/sheetcoords.py
```diff
@@ -517,7 +517,6 @@ def _boundsspec2slicespec(boundsspec,scs):
 
         l_idx = int(np.ceil(l_m-0.5))
         t_idx = int(np.ceil(t_m-0.5))
-        # CBENHANCEMENT: Python 2.6's math.trunc()?
         r_idx = int(np.floor(r_m+0.5))
```
Hoxbro, line 520: Out of scope for this PR. But because these are scalars, it is significantly faster* to use the built-in math module:

image

*) Only relative - absolute it does not make a difference :upside_down_face:

Hoxbro (approved):

Hoxbro:
LGTM. Thank you for the fix :+1:

ahuang11:
Cool 😎 no problem! This bug had me do work arounds 2 years ago; I'm glad it's finally fixed haha

ahuang11:
Cool 😎 no problem! This bug had me do work arounds 2 years ago; I'm glad it's finally fixed haha

PR #5811, opened 2023-07-14
Allow and handle non-parameters in rasterize.instance()

ahuang11:
I don't fully understand this...

If I use HoloViews alone / try to reproduce with HoloViews:
```python
import numpy as np
import pandas as pd
import holoviews as hv
hv.extension("bokeh")

df = pd.DataFrame(
    np.random.multivariate_normal((0, 0), [[0.1, 0.1], [0.1, 1.0]], (500000,))
)
df.columns = ["a", "b"]

curve = hv.Curve(df, kdims=["a"], vdims=["b"]).opts("Curve", line_width=1)
hv.operation.apply_when(curve, operation=rasterize, predicate=lambda x: len(x) > 1000)
```
It runs fine...

However, if I use hvplot with this [PR](https://github.com/holoviz/hvplot/pull/1103/files):
```python
import numpy as np
import pandas as pd
import hvplot.pandas

df = pd.DataFrame(
    np.random.multivariate_normal((0, 0), [[0.1, 0.1], [0.1, 1.0]], (500000,))
)
df.columns = ["a", "b"]
df.hvplot("a", "b", rasterize=True, aggregation_threshold=1000)
```

I get the following traceback:
```
File [~/repos/holoviews/holoviews/operation/datashader.py:1389](https://file+.vscode-resource.vscode-cdn.net/Users/ahuang/repos/sketches/~/repos/holoviews/holoviews/operation/datashader.py:1389), in rasterize._process(self, element, key)
   1387 all_allowed_kws = set()
   1388 all_supplied_kws = set()
-> 1389 instance_params = {
   1390     k: getattr(self, k) for k in self.__instance_params
   1391 }
   1392 for predicate, transform in self._transforms:
   1393     merged_param_values = dict(instance_params, **self.p)

File [~/repos/holoviews/holoviews/operation/datashader.py:1390](https://file+.vscode-resource.vscode-cdn.net/Users/ahuang/repos/sketches/~/repos/holoviews/holoviews/operation/datashader.py:1390), in (.0)
   1387 all_allowed_kws = set()
   1388 all_supplied_kws = set()
   1389 instance_params = {
-> 1390     k: getattr(self, k) for k in self.__instance_params
   1391 }
   1392 for predicate, transform in self._transforms:
   1393     merged_param_values = dict(instance_params, **self.p)

AttributeError: 'rasterize' object has no attribute 'line_width'
```

### DIFF
--- a/holoviews/tests/operation/test_datashader.py
+++ b/holoviews/tests/operation/test_datashader.py
@@ -9,8 +9,9 @@ import pytest
 from holoviews import (
     Dimension, Curve, Points, Image, Dataset, RGB, Path, Graph, TriMesh,
     QuadMesh, NdOverlay, Contours, Spikes, Spread, Area, Rectangles,
-    Segments, Polygons, Nodes
+    Segments, Polygons, Nodes, DynamicMap, Overlay
 )
+from holoviews.util import render
 from holoviews.streams import Tap
 from holoviews.element.comparison import ComparisonTestCase
 from numpy import nan
@@ -1182,7 +1183,11 @@ class DatashaderRasterizeTests(ComparisonTestCase):
         output = apply_when(
             curve, operation=custom_rasterize, predicate=lambda x: len(x) > 1000
         )
-        assert isinstance(output, Image)
+        render(output, "bokeh")
+        assert isinstance(output, DynamicMap)
+        overlay = output.items()[0][1]
+        assert isinstance(overlay, Overlay)
+        assert len(overlay) == 2
 
 class DatashaderSpreadTests(ComparisonTestCase):
 
--- a/holoviews/tests/operation/test_datashader.py
+++ b/holoviews/tests/operation/test_datashader.py
@@ -1174,14 +1174,15 @@ class DatashaderRasterizeTests(ComparisonTestCase):
     def test_rasterize_apply_when_instance_with_line_width(self):
         df = pd.DataFrame(
             np.random.multivariate_normal(
-            (0, 0), [[0.1, 0.1], [0.1, 1.0]], (500000,))
+            (0, 0), [[0.1, 0.1], [0.1, 1.0]], (100,))
         )
         df.columns = ["a", "b"]
 
         curve = Curve(df, kdims=["a"], vdims=["b"])
+        # line_width is not a parameter
         custom_rasterize = rasterize.instance(line_width=2)
         output = apply_when(
-            curve, operation=custom_rasterize, predicate=lambda x: len(x) > 1000
+            curve, operation=custom_rasterize, predicate=lambda x: len(x) > 10
         )
         render(output, "bokeh")
         assert isinstance(output, DynamicMap)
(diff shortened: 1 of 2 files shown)


Hoxbro:
This is a MRE:

``` python
import holoviews as hv
import numpy as np
import pandas as pd

hv.extension("bokeh")

df = pd.DataFrame(
    np.random.multivariate_normal((0, 0), [[0.1, 0.1], [0.1, 1.0]], (500000,))
)
df.columns = ["a", "b"]

curve = hv.Curve(df, kdims=["a"], vdims=["b"])
inst = rasterize.instance(line_width=2)
hv.operation.apply_when(curve, operation=inst, predicate=lambda x: len(x) > 1000)
```

`__instance_params` was added in https://github.com/holoviz/holoviews/pull/5767, and by the look of how it worked before this fix should be okay.

Is it possible to add a small unit test?

holoviews/tests/operation/test_datashader.py
```diff
@@ -1169,6 +1171,24 @@ def test_rasterize_image_expand_default(self):
         output = img.data["z"].to_numpy()
         assert np.isnan(output).any()
 
+    def test_rasterize_apply_when_instance_with_line_width(self):
+        df = pd.DataFrame(
```
Hoxbro, line 1175: I think we should reduce the size of this DataFrame. No need for it to be this big for a unit test.

holoviews/tests/operation/test_datashader.py
```diff
@@ -1169,6 +1171,24 @@ def test_rasterize_image_expand_default(self):
...
+            np.random.multivariate_normal(
+            (0, 0), [[0.1, 0.1], [0.1, 1.0]], (500000,))
+        )
+        df.columns = ["a", "b"]
+
+        curve = Curve(df, kdims=["a"], vdims=["b"])
+        custom_rasterize = rasterize.instance(line_width=2)
```
Hoxbro, line 1182: Maybe add a comment about what is so special about `line_width` e.g. it is not a parameter.

Hoxbro (commented):
I left some comments, but maybe it is correct behavior to error out here. 

When using `.instance()` and one of the inputs is not a parameter (here `line_width`), it is not using it. Beforehand this will just silently fail, but with the recent changes, it gives an error, which I'm leaning toward being a good thing.   

Take a look at these examples (with your changes / the behavior before https://github.com/holoviz/holoviews/pull/5767):
image

``` python
import holoviews as hv
import numpy as np
import pandas as pd

hv.extension("bokeh")
np.random.seed(10)

df = pd.DataFrame(
    np.random.multivariate_normal((0, 0), [[0.1, 0.1], [0.1, 1.0]], (500000,)),
    columns=["a", "b"],
)
curve = hv.Curve(df, kdims=["a"], vdims=["b"])

rasterize.instance(line_width=2)(curve).opts(xlim=(1.1, 1.2), ylim=(1.1, 1.2))
rasterize(curve, line_width=2).opts(xlim=(1.1, 1.2), ylim=(1.1, 1.2))
```
So maybe in the `hvplot` code, we should use a partial function for the operation:
`hv.operation.apply_when(curve, operation=partial(rasterize, line_width=2), predicate=lambda x: len(x) > 1000)`

Hoxbro:
I have updated the code to save non-parameter in the instance creation, so it works as one would expect:

image

philippjfr:
Looks great, thanks!

PR #5814, opened 2023-07-19
Avoid setting `batched` before the class is initialized

maximlt:
@Hoxbro feel free to reject this PR since it's not a fix per say.

What used to happen is showed in the example below, where the `batched` Parameter of `GenericOverlayPlot` was set (`self.batched = batched`) before `Parameterized.__init__` was called. Setting a Parameter value before a Parameterized class is initialized (at the Param sense) has been supported somehow in Param for a long time.  https://github.com/holoviz/param/pull/766 broke that support temporarily, while https://github.com/holoviz/param/pull/790 is an attempt at restoring it before Param 2.0 is released. Breaking that support temporarily allowed me to spot that this `batched` case was the only occurrence of a Parameter value being set early across a few code bases (Panel, Lumen, HoloViews). Because of that, and because it feels to me like an anti-pattern, I was tempted to open this PR to avoid this, 

```
class GenericElementPlot(param.Parameterized):

    ....

    def __init__(self, batched=False, **params):
        self.batched = batched
        super().__init__(**params)

class GenericOverlayPlot(GenericElementPlot):

    ...

    batched = param.Boolean()

    def __init__(self, batched=True, **params):
        super().__init__(batched=batched, **params)

# When GenericOverlayPlot is instantiated `batched` is set before `Parameterized.__init__` is called.
GenericOverlayPlot()
```

### DIFF
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -1164,7 +1164,6 @@ class GenericElementPlot(DimensionedPlot):
         self.zorder = zorder
         self.cyclic_index = cyclic_index
         self.overlaid = overlaid
-        self.batched = batched
         self.overlay_dims = overlay_dims
 
         if not isinstance(element, (HoloMap, DynamicMap)):
@@ -1179,7 +1178,7 @@ class GenericElementPlot(DimensionedPlot):
             self.stream_sources = compute_overlayable_zorders(self.hmap)
 
         plot_element = self.hmap.last
-        if self.batched and not isinstance(self, GenericOverlayPlot):
+        if batched and not isinstance(self, GenericOverlayPlot):
             plot_element = plot_element.last
 
         dynamic = isinstance(element, DynamicMap) and not element.unbounded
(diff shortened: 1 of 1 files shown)


maximlt:
I also thought about adding `batched` as a Parameter to `GenericElementPlot`. Then I think I got confused by `self.batched = batched` being in between of other non Parameter attributes being set like `zorder`, `cyclic_index` or `overlay_dims` and wondered if it made sense for `batched` to be made more "officially" an input of `GenericElementPlot` while it seems to be dedicated to handle specifically NdOverlays?

jbednar:
> I also thought about adding batched as a Parameter to GenericElementPlot. 

I agree that it's very strange to have GenericElementPlot accept and respect `batched` without `batched` being a Parameter on GenericElementPlot, only its subclass.  Seems very fishy and it's hard to see any justification for doing that. I agree it's truly bizarre that generally GEP has only a `batched` attribute except that sometimes it has a `batched` Parameter (if it's also a GenericOverlayPlot).

PR #5819, opened 2023-07-21
Disable Dask's conversion of strings in test 

Hoxbro:
Ref: https://docs.dask.org/en/stable/changelog.html#v2023-7-1

In the future, we should look into how our library handles the pyarrow backend in pandas.  This should be done with `pd.options.mode.string_storage = "pyarrow"` [ref](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.StringDtype.html).

But this seems like a bigger undertaking and for now this make our test go green again.

### DIFF
--- a/holoviews/tests/core/data/__init__.py
+++ b/holoviews/tests/core/data/__init__.py
@@ -1,7 +1,0 @@
-try:
-    # From Dask 2023.7,1 they now automatic convert strings
-    # https://docs.dask.org/en/stable/changelog.html#v2023-7-1
-    import dask
-    dask.config.set({"dataframe.convert-string": False})
-except Exception:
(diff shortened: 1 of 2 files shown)


holoviews/tests/core/data/__init__.py
```diff
@@ -0,0 +1,7 @@
+try:
+    # From Dask 2023.7,1 they now automatic convert strings
+    # https://docs.dask.org/en/stable/changelog.html#v2023-7-1
+    import dask
+    dask.config.set({"dataframe.convert-string": False})
+except Exception:
+    pass
```
philippjfr, line 7: Probably better to have this in a `conftest.py`?

PR #5826, opened 2023-07-24
Support RangeXY streams on multi-axes

philippjfr:
This PR does four things:

- Generalize the logic for populating `Plot.handles` with the correct x/y-range and x/y-axis 
- Generalize the logic for mapping the correct axes/ranges to the GlyphRenderer
- Fix Linked Stream Callback hashing to ensure that plots with different axes don't reuse the same callback
- Look up RangeXY data on ranges rather than use event data   
 

- [x] Add tests

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -400,7 +400,6 @@ class Callback:
 
         # Hash the plot handle with Callback type allowing multiple
         # callbacks on one handle to be merged
-        handle_ids = [id(h) for h in cb_handles]
         hash_ids = [id(h) for h in hash_handles]
         cb_hash = tuple(hash_ids)+(id(type(self)),)
         if cb_hash in self._callbacks:
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -87,6 +87,9 @@ class Callback:
     # The plotting handle(s) to attach the JS callback on
     models = []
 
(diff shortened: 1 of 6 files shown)


holoviews/plotting/bokeh/callbacks.py
```diff
@@ -87,6 +87,9 @@ class Callback:
     # The plotting handle(s) to attach the JS callback on
     models = []
 
+    # Additional handles to hash on for uniqueness
+    extra_handles = []
```
philippjfr, line 91: We need to tell the callback to hash itself based on all the handles so we don't reuse a callback attached to different axes. It's a bit ugly as you can see below.

holoviews/plotting/bokeh/element.py
```diff
@@ -1714,10 +1737,9 @@ def initialize_plot(self, ranges=None, plot=None, plots=None, source=None):
             plot = self._init_plot(key, style_element, ranges=ranges, plots=plots)
             self._init_axes(plot)
         else:
-            self.handles['xaxis'] = plot.xaxis[0]
-            self.handles['x_range'] = plot.x_range
-            self.handles['yaxis'] = plot.yaxis[0]
-            self.handles['y_range'] = plot.y_range
+            axes, plot_ranges = self._find_axes(plot, element)
+            self.handles['xaxis'], self.handles['yaxis'] = axes
+            self.handles['x_range'], self.handles['y_range'] = plot_ranges
```
philippjfr, line 1742: We correctly populate the handles on each subplot.

Issue #5823, opened 2023-07-21
Warn when using `multi_y` with datashader


jlstevens:
Given that the new `multi_y` option implies multiple data spaces (in y at least) which isn't supported, it might be worth having a warning saying that the datashader support is not ready (instead of warning about the streams which the user may not realize is a general issue relating to the datashader support)

jlstevens:
Assigning this to myself for 1.17.0 as this should be simple to implement. Also, I am giving myself a reminder to also update the docs to mention that `multi_y` zooming only properly works for Bokeh >=3.2 :-)

jlstevens:
In PR https://github.com/holoviz/holoviews/pull/5826 `RangeXY` is no longer warning: it might be worth warning to suggest the need to overlay rasterized layers to make twin axis rasterization work properly.

jbednar:
Using the released holoviews 1.17.0, rasterizing an overlay definitely seems problematic:

```python
import holoviews as hv
from holoviews.operation.datashader import datashade,rasterize
hv.extension("bokeh")

overlay = (hv.Curve([1, 4, 2], vdims=['A']) * 
           hv.Curve([2, 3, 8], vdims=['A']) * 
           hv.Curve([3, 2, 0], vdims=['B']))
```
<img width="876" alt="image" src="https://github.com/holoviz/holoviews/assets/1695496/97c4a40e-c65b-43e0-bd68-e53ec9542b86">

(no warning, and not a usable plot, missing big parts of the rendering).

Overlaying rasterized lines works better:

```python
overlay = (rasterize(hv.Curve([1, 4, 2], vdims=['A']), line_width=3) * 
           rasterize(hv.Curve([2, 3, 8], vdims=['A']), line_width=3) * 
           rasterize(hv.Curve([3, 2, 0], vdims=['B']), line_width=3)).opts(width=700, multi_y=True)
overlay
```

<img width="709" alt="image" src="https://github.com/holoviz/holoviews/assets/1695496/1e747509-d574-4ea6-a6e8-7d7db8d92d35">

However, I get a bunch of `assert 0 < size <= self._size` AssertionErrors when I zoom on any axis, and the plot fails to re-render at the new resolution appropriate for this zoom (zooming on B here):

<img width="706" alt="image" src="https://github.com/holoviz/holoviews/assets/1695496/b1c0a32a-77af-4701-8226-98abbc6bdb90">
[...]

PR #5829, opened 2023-07-25
Added warning when datashading with multi_y=True

jlstevens:
Simple PR to address https://github.com/holoviz/holoviews/issues/5823

Now that `RangeXY` is extra axis aware, you still need to overlay rasterize calls to datashade with `multi_y=True`. This PR adds such a warning/suggestion for the user until we can support a `RangeXMultiY` stream (or similar).

### DIFF
--- a/holoviews/operation/resample.py
+++ b/holoviews/operation/resample.py
@@ -114,6 +114,14 @@ class ResampleOperation2D(ResampleOperation1D):
         inst._precomputed = {}
         return inst
 
+
+    def __call__(self, element, **kwargs):
+        opts = element.opts.get('plot', 'bokeh')
+        opts = {} if opts is None else opts.kwargs
+        if 'multi_y' in opts:
+            self.param.warning('Overlay separate datashader operation calls to rasterize with multi_y=True')
+        return super().__call__(element, **kwargs)
+
     def _get_sampling(self, element, x, y, ndim=2, default=None):
         target = self.p.target
         if not isinstance(x, list) and x is not None:


jlstevens:
Note that I am hardcoding 'bokeh' as the backend: `multi_y` is currently only supported by bokeh anyway so I am making the assumption that any presence of `multi_y` means Bokeh is being used.

holoviews/operation/resample.py
```diff
@@ -114,6 +114,14 @@ def instance(self_or_cls,**params):
         inst._precomputed = {}
         return inst
 
+
+    def __call__(self, element, **kwargs):
+        opts = element.opts.get('plot', 'bokeh')
+        opts = {} if opts is None else opts.kwargs
+        if 'multi_y' in opts:
```
philippjfr, line 121:
```suggestion
        if 'multi_y' in opts and isinstance(element, CompositeOverlay):
```

philippjfr: Not sure that's quite right, but I don't see a good reason to warn if you just provide an Element. Also the `multi_y` may be set after the fact and then this warning won't do anything. So overall I don't think this approach is very good.

jlstevens: Fair enough. Where would you warn from when using datashader with `multi_y=True` then?

jlstevens: One other alternative of course is to not warn at all and to document this instead...

philippjfr:
Sadly I don't think this approach is workable, it checks only bokeh, it can't detect with multi-y is applied after the fact. The only approach I can see working is to inspect the pipeline on the plot class to detect this scenario.

PR #5830, opened 2023-07-25
Handle issue with multiple legends in Bokeh sync_legends

philippjfr:
Fix for this error in the hvplot test suite: https://github.com/holoviz/hvplot/actions/runs/5633558389/job/15262593959?pr=1114#step:4:2352


maximlt:
I was able to reproduce the error locally running the failing hvPlot notebook and I confirm this change allows to get rid of the error.

maximlt (approved):

PR #5848, opened 2023-08-08
Add reset_start and reset_end to RangeToolLink

Hoxbro:
This change will return to the bounds set when clicking the reset button.

### DIFF
--- a/holoviews/plotting/bokeh/links.py
+++ b/holoviews/plotting/bokeh/links.py
@@ -151,8 +151,10 @@ class RangeToolLinkCallback(LinkCallback):
             start, end = bounds
             if start is not None:
                 axes[f'{axis}_range'].start = start
+                axes[f'{axis}_range'].reset_start = start
             if end is not None:
                 axes[f'{axis}_range'].end = end
+                axes[f'{axis}_range'].reset_end = end
 
         tool = RangeTool(**axes)
         source_plot.state.add_tools(tool)
--- a/holoviews/tests/plotting/bokeh/test_links.py
+++ b/holoviews/tests/plotting/bokeh/test_links.py
@@ -52,6 +52,8 @@ class TestLinkCallbacks(TestBokehPlot):
         tgt_plot = plot.subplots[(0, 0)].subplots['main']
         self.assertEqual(tgt_plot.handles['x_range'].start, x_start)
         self.assertEqual(tgt_plot.handles['x_range'].end, x_end)
+        self.assertEqual(tgt_plot.handles['x_range'].reset_start, x_start)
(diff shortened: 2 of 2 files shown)


holoviews/plotting/bokeh/links.py
```diff
@@ -151,8 +151,10 @@ def __init__(self, root_model, link, source_plot, target_plot):
             start, end = bounds
             if start is not None:
                 axes[f'{axis}_range'].start = start
+                axes[f'{axis}_range'].reset_start = start
             if end is not None:
                 axes[f'{axis}_range'].end = end
+                axes[f'{axis}_range'].reset_end = end
```
droumis, line 157: I think we should add a option to revert to previous behavior, like:

```suggestion
            if start is not None:
                axes[f'{axis}_range'].start = start
            if end is not None:
                axes[f'{axis}_range'].end = end
            if link.reset_to_bounds:
                axes[f'{axis}_range'].reset_start = start
                axes[f'{axis}_range'].reset_end = end
```

and then add the following in holoviews/plotting/links.py - RangeToolLink class

```
    reset_to_bounds = param.Boolean(default=True, doc="""
        If True (default), the 'reset' action returns the plot to its initial bounds.
        If False, it uses Bokeh's default reset behavior.""")`
```

droumis:
The code looks good to me and behaves in the way I would expect now.. see the video below.

https://github.com/holoviz/holoviews/assets/6613202/27f8d5dd-7003-4031-8c7a-25dbf9269ddc


Now when a user clicks the "reset" button in a Bokeh-backed plot, the axis range will revert to these specific values that were the range of the initial display, rather than the prior behavior which I think was ranging to the full data extents of the current data.

Although I think this is the correct default behavior, a potential side effect is that if an application relies on the existing default behavior of the "reset" action, this might introduce unexpected behavior. I can't really think of an instance where this would be an issue, except for the case where data in the plot is updated or changes dynamically. Then, the reset_start and reset_end values will not be updated to the new data extents and the reset action may not show the entirety of the new data. I doubt this will be an issue and I think this PR corrects what should be the default behavior, but I added a suggestion to include an option to revert to the previous behavior. 

Otherwise, ready to merge

Hoxbro:
The previous behavior was "only" introduced in 1.17 (implemented in #5800), so I feel no one has had time to rely on that behavior. I can, like you, not think of a reason why it would be relied upon. 

If you have set `boundsx` or `boundsy`, the reset button should return to the original state, which is what these changes do.   

I would rather not add your suggested setting for this PR if you don't mind. If there is a demand for it, we can easily add it, but we cannot easily remove it if we add it to this PR.

droumis:
sounds good!

PR #5854, opened 2023-08-11
Remove warnings

Hoxbro:
Removing warnings from the CI. The goal is to have warnings raise an error in the CI. 

Gridplot warning will be handled in: https://github.com/holoviz/holoviews/pull/5873

xarray warnings will be handled in another PR, as I'm not sure about what to do with them right now: 
``` python-traceback
UserWarning: rename 'x' to 'x' does not create an index anymore. Try using swap_dims instead or use set_index after rename to create an indexed coordinate.
UserWarning: Converting non-nanosecond precision datetime values to nanosecond precision. This behavior can eventually be relaxed in xarray, as it is an artifact from pandas which is now beginning to support non-nanosecond precision values. This warning is caused by passing non-nanosecond np.datetime64 or np.timedelta64 values to the DataArray or Variable constructor; it can be silenced by converting the values to nanosecond precision ahead of time.

### DIFF
--- a/holoviews/core/data/ibis.py
+++ b/holoviews/core/data/ibis.py
@@ -194,7 +194,7 @@ class IbisInterface(Interface):
             import ibis
             order = ibis.desc if reverse else ibis.asc
             return dataset.data.order_by([order(dataset.get_dimension(x).name) for x in by])
-        if ibis4():
+        elif ibis4():
             # Tuple syntax will be removed in Ibis 7.0:
             # https://github.com/ibis-project/ibis/pull/6082
             return dataset.data.order_by([(dataset.get_dimension(x).name, not reverse) for x in by])
--- a/holoviews/core/data/ibis.py
+++ b/holoviews/core/data/ibis.py
@@ -23,11 +23,6 @@ def ibis4():
     return ibis_version() >= Version("4.0")
 
 
-@lru_cache
-def ibis6():
-    return ibis_version() >= Version("6.0")
-
-
 class IbisInterface(Interface):
 
     types = ()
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: holoviews/core/data/ibis.py [hunk 1]
```diff
@@ -23,6 +23,11 @@ def ibis4():
     return ibis_version() >= Version("4.0")
 
 
+@lru_cache
+def ibis6():
+    return ibis_version() >= Version("7.0")
```
Hoxbro, line 28:
```suggestion
    return ibis_version() >= Version("6.0")
```

holoviews/core/data/ibis.py
```diff
@@ -190,7 +190,13 @@ def dtype(cls, dataset, dimension):
 
     @classmethod
     def sort(cls, dataset, by=[], reverse=False):
+        if ibis_version() >= Version("6.0"):
+            import ibis
+            order = ibis.desc if reverse else ibis.asc
+            return dataset.data.order_by([order(dataset.get_dimension(x).name) for x in by])
         if ibis4():
```
Hoxbro, line 197:
```suggestion
        elif ibis4():
```

holoviews/core/data/xarray.py
```diff
@@ -172,7 +173,17 @@ def retrieve_unit_and_label(dim):
...
+                            # xarray emit a warning when datetime/timedelta is not nanosecond,
+                            # and then convert it to nanosecond precision. The full warning is:
+                            #   Converting non-nanosecond precision {case} values to nanosecond precision.
+                            #   This behavior can eventually be relaxed in xarray, as it is an artifact from
+                            #   pandas which is now beginning to support non-nanosecond precision values.
+                            #   This warning is caused by passing non-nanosecond np.datetime64 or
+                            #   np.timedelta64 values to the DataArray or Variable constructor; it can be
+                            #   silenced by converting the values to nanosecond precision ahead of time.
+                            warnings.filterwarnings("ignore", "Converting non-nanosecond precision")
+                            arr = xr.DataArray(arr, coords=coords, **xr_kwargs)
```
Hoxbro, line 186: I don't know if this is the correct way to do it. Alternative ways:

1) Ignore the warning in pyproject.toml
2) Convert the data in the unit test to nanoseconds beforehand.

PR #5856, opened 2023-08-15
Improve support for NdOverlay and Overlay in downsample1d

Hoxbro:
Before the following code will always return the non-downsampled ndoverlay/overlay. 

``` python
        if len(element) <= self.p.width:
            return element
```


philippjfr (approved):
Looks good!

PR #5865, opened 2023-08-25
Update tests to new Param 2 error messages

maximlt:


Hoxbro:
Is the test failing because a new release is needed?

I have added a `param ==1.13` pin for the Bokeh 2 tests in https://github.com/holoviz/holoviews/pull/5844. So we need to add `if bokeh3` to get all tests to pass.

maximlt:
Yep they're failing because they're relying on code that hasn't been released yet! But I don't want to force Panel to rely on the latest Param all the time so I'll adapt them to be compatible.

PR #5867, opened 2023-08-28
Drop requirement for NdMapping to be defined with OrderedDict

philippjfr:
Since Python dict is now ordered the limitation around requiring an NdMapping input dict to be an `OrderedDict` is no longer necessary.


Hoxbro:
Can you also remove?
https://github.com/holoviz/holoviews/blob/04d7fa6691dbf8bb8605fe35ef73fb6fb05b0db2/holoviews/element/tabular.py#L44-L46

Issue #5869, opened 2023-08-29
Handle Bokeh toolbar UserWarning


Hoxbro:
With the release of Bokeh 3.2.2 we are starting to get the following warnings both in CI and in a simple addition of plots. 

``` python-traceback
/home/shh/Repos/holoviz/holoviews/holoviews/plotting/bokeh/plot.py:987: UserWarning: found multiple competing values for 'toolbar.active_drag' property; using the latest value
  layout_plot = gridplot(
/home/shh/Repos/holoviz/holoviews/holoviews/plotting/bokeh/plot.py:987: UserWarning: found multiple competing values for 'toolbar.active_scroll' property; using the latest value
  layout_plot = gridplot(
````

``` python
import holoviews as hv
import pandas as pd

hv.extension("bokeh")
hv.Curve(range(10)) + hv.Curve(range(10))
```

![image](https://github.com/holoviz/holoviews/assets/19758978/3227cba8-a9ed-478f-82bd-0a71a20c73a1)

droumis:
can confirm I receive these too

PR #5873, opened 2023-08-30
Ignore Bokeh UserWarning for multiple active properties in toolbar

Hoxbro:
Resolves https://github.com/holoviz/holoviews/issues/5869

The warnings we are seeing are the tools we are actually setting in: 

https://github.com/holoviz/holoviews/blob/6dcdb9aa4fe48be9df1110033e12ca83874fd3a1/holoviews/plotting/bokeh/element.py#L698-L705

We are using our merge_tools for merging tools instead of Bokeh. 

No special attention has been given to Bokeh 2, as it will be removed in https://github.com/holoviz/holoviews/pull/5891

### DIFF
--- a/holoviews/plotting/bokeh/plot.py
+++ b/holoviews/plotting/bokeh/plot.py
@@ -1,3 +1,4 @@
+import warnings
 from itertools import groupby
 from collections import defaultdict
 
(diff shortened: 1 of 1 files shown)


holoviews/plotting/bokeh/plot.py
```diff
@@ -586,13 +586,13 @@ def initialize_plot(self, ranges=None, plots=[]):
...
         plot = gridplot(plots[::-1],
-                        merge_tools=self.merge_tools,
+                        merge_tools=False,
                         sizing_mode=self.sizing_mode,
                         toolbar_location=self.toolbar)
         if self.sync_legends:
             sync_legends(plot)
         plot = self._make_axes(plot)
-        if bokeh3 and hasattr(plot, "toolbar"):
+        if bokeh3 and hasattr(plot, "toolbar") and self.merge_tools:
```
philippjfr, line 595: Does that mean this was previously incorrect? Also what happens in the bokeh2 case now?

Hoxbro: I think so. 

I have ignored Bokeh 2 as I plan to remove it soon with https://github.com/holoviz/holoviews/pull/5891. We don't even test for it anymore after the ImageStack PR was merged.

philippjfr: Okay, sounds good.

philippjfr (approved):
Looks good, only just realized that the new `merge_tools` code is something I only implemented recently as part of the Bokeh 3.0 migration. The logic in there makes sense and matches our required semantics much better than what Bokeh does.

PR #5874, opened 2023-08-31
support python3-only systems

MeggyCal:
Python 2 has been dead for some time and some rolling distros (like openSUSE Tumbleweed) don't have the `python` binary anymore. I think that checking for the `python3` binary is more inclusive and doesn't change the nature of the test.

### DIFF
--- a/holoviews/tests/util/test_init.py
+++ b/holoviews/tests/util/test_init.py
@@ -14,6 +14,6 @@ def test_no_blocklist_imports():
         print(", ".join(mods), end="")
         """
 
-    output = check_output(['python', '-c', dedent(check)])
+    output = check_output(['python3', '-c', dedent(check)])
 
     assert output == b""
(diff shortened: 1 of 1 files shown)


Hoxbro:
I think this could cause problems with Windows. Can you make it an if statement? like
``` python
if shutil.which("python"):
    check_output(['python', '-c', dedent(check)])
else:
    check_output(['python3', '-c', dedent(check)])
```

MeggyCal:
sorry for the force-pushes :( like this?

Hoxbro:
> sorry for the force-pushes :( like this?

No worries. Force push is part of life. And yes.

Hoxbro:
Thank you for the PR.

Failing PRs are unrelated.

MeggyCal:
I thank you for the review :) it's been an honor.

PR #5878, opened 2023-09-07
Ibis: update deprecated call to `order_by`

maximlt:
This PR takes care of this warning:

```
holoviews/tests/core/data/test_ibisinterface.py: 16 warnings
  /Users/mliquet/miniconda3/envs/hv39/lib/python3.9/site-packages/ibis/expr/types/relations.py:1268: FutureWarning: `table.order_by((key, True)) and table.order_by((key, False)) syntax` is deprecated as of v6.0, removed in v7.0; Use ibis.desc(key) or ibis.asc(key) instead
```

### DIFF
--- a/holoviews/core/data/ibis.py
+++ b/holoviews/core/data/ibis.py
@@ -191,7 +191,9 @@ class IbisInterface(Interface):
     @classmethod
     def sort(cls, dataset, by=[], reverse=False):
         if ibis4():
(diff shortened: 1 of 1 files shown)


Hoxbro:
I have also removed it in https://github.com/holoviz/holoviews/pull/5854. But I haven't gotten around to finishing up that PR :slightly_frowning_face:

maximlt:
Oh yes I missed that!

PR #5881, opened 2023-09-11
Allow to link to an `Overlay`

maximlt:
Partly extracted from https://github.com/holoviz/holoviews/pull/5840

The goal of this PR is to allow linking (with a `RangeToolLink`) to an `Overlay`, the example in mind being a minimap targeting a plot with subcoordinates (subcoordinates being implemented as an `Overlay`).

Overriding `.clone` at the `Overlay` level is done to be more specific instead of what was done in https://github.com/holoviz/holoviews/pull/5840 at the `LabelledData` level.

Overriding `clone` is only required to get `test_range_tool_link_callback_single_axis_overlay_target_image_source` to pass, i.e. `test_range_tool_link_callback_single_axis_overlay_target` runs fine without it. While the two tests look very much alike, the former uses an `Image` as a source. I saw that there's a [`Compositor` registered for the `Image` element](https://github.com/holoviz/holoviews/blob/14869550861111538584840503f237113a403726/holoviews/plotting/__init__.py#L16), which makes the code run through a pretty hairy path when creating a simple `Image` (`.map`, `.collapse_element`) and leads to the creation of internal/intermediate overlays.

I would be happy to:

- have someone more knowledgeable in the `Compositor` logic explain what is going on there :)
- add more tests to:
  - cover the breadth of the envisioned feature; so far I've only tested the `RangeToolLink` to a simple overlay of curves, with either an `Image` or a `Curve` as source.
  - maybe cover what those changes (that still feel a little hacky to, because I feel I'm just playing with a house of cards!) could break

As for documenting these changes, I haven't found any place to update where links were described as limited to elements. There are two gallery examples that demonstrate how to use the `RangeToolLink`, it seems to me it wouldn't improve them so much to show how to link to an overlay. My suggestion instead would be to show that this can be done adding an EEG example to the gallery (in or after https://github.com/holoviz/holoviews/pull/5840).

### DIFF
--- a/holoviews/core/overlay.py
+++ b/holoviews/core/overlay.py
@@ -289,7 +289,7 @@ class Overlay(ViewableTree, CompositeOverlay, Layoutable, Overlayable):
 
     def clone(self, data=None, shared_data=True, new_type=None, link=True,
               *args, **overrides):
-        if link:
+        if data is None and link:
             overrides['plot_id'] = self._plot_id
         return super().clone(data, shared_data=shared_data, new_type=new_type, link=link, *args, **overrides)
 
--- a/holoviews/core/overlay.py
+++ b/holoviews/core/overlay.py
@@ -287,6 +287,12 @@ class Overlay(ViewableTree, CompositeOverlay, Layoutable, Overlayable):
     def shape(self):
         raise NotImplementedError
 
+    def clone(self, data=None, shared_data=True, new_type=None, link=True,
+              *args, **overrides):
+        if link:
+            overrides['plot_id'] = self._plot_id
+        return super().clone(data, shared_data=shared_data, new_type=new_type, link=link, *args, **overrides)
+
 
 class NdOverlay(Overlayable, UniformNdMapping, CompositeOverlay):
     """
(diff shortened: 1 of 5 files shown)


philippjfr:
>have someone more knowledgeable in the Compositor logic explain what is going on there :)

Compositors are a really, really old mechanism in HoloViews and were originally designed to take in an Overlay and run some operation to collapse multiple layers into a single layer (this was primarily used for combining two Image layers into a single RGB layer). Later we repurposed this as a general mechanism to allow operations to be applied as part of the plotting code. You can register a Compositor to match on one or more layers of a plot (this is done using `.map` I believe) and once a match is found `.collapse_element` is called to apply the operation (it's a misnomer for Compositors that work on a single Element but wasn't worth renaming). In the case of `Image` we have a Compositor that replaces the `nodata` values, which then results in the creation of the intermediate Overlay.

>maybe cover what those changes (that still feel a little hacky to, because I feel I'm just playing with a house of cards!) could break

Agreed, unfortunately I have no good suggestions on what needs to be tested. When I made the changes in #5840 I did notice a few tests starting to fail with my changes, and it's a good sign that isn't happening in this PR.

>My suggestion instead would be to show that this can be done adding an EEG example to the gallery (in or after #5840).

Agreed that sounds nice.

philippjfr (approved):
I'm happy to see this merged. If we could think of more tests that would be nice but as mentioned above I have no good way of figuring out what tests would make sense to add.

PR #5890, opened 2023-09-16
Remove all use of OrderedDict

Hoxbro:
`OrderedDict` is no longer required as `dict` has been in insertion order since Python 3.7.


holoviews/plotting/bokeh/element.py
```diff
@@ -762,7 +762,7 @@ def _axis_properties(self, axis, key, plot, dimension=None,
         Returns a dictionary of axis properties depending
         on the specified axis.
         """
-        # need to copy dictionary by calling dict() on it
+        # need to copy dictionary by calling {} on it
```
Hoxbro, line 765:
```suggestion
        # need to copy dictionary by calling dict() on it
```

PR #5891, opened 2023-09-16
Dropping support for Bokeh 2

Hoxbro:
#5751 removed Bokeh 2 from the test suite.

This PR continues it and removes support from the code base.

The sorting of import will be done in another PR, where I will enable automatic linting for it.


philippjfr (approved):
Great, thank you! All the changes seem straightforward and uncontroversial. LGTM

PR #5897, opened 2023-09-19
Fix `Path.datatype` parameter type

maximlt:
From `ObjectSelector` to `List`, `List` being the type of `datatype` in the base class `Dataset`.


Hoxbro:
Looks good

PR #5900, opened 2023-09-21
DOC: Add *Linking Bokeh plots* guide to the toc

maximlt:
I knew this guide existed but often struggled to find it, now I know why :)


Hoxbro:
Thank you for finding the lost documentation :+1:

Issue #5740, opened 2023-06-04
LaTeX Support for Labels (Title, Axes, Ticks, Colorbar, etc.)


cdeciampa:
Package versions:
- holoviews = 1.16.1
- datashader = 0.15.0
- bokeh = 3.1.1

As of Bokeh 3.0, [LaTeX is now supported](https://docs.bokeh.org/en/3.1.1/docs/user_guide/styling/mathtext.html), but I don't think it's supported by holoviews yet. Using `r"Upwelling Longwave Flux W$$m^{-2}$$"` doesn't work for the colorbar label:

![test_olr](https://github.com/holoviz/holoviews/assets/66640080/f7ca8079-2af0-41c3-b4aa-de3bb6ecc3a0)

And using that same label but for the title throws a KeyError:
```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Cell In[28], line 1
----> 1 hv.save(mpas_raster_plot, '../figs/test_olr.png', dpi=300, center=False)

File /storage/work/c/cnd5285/mambaforge/envs/uviz/lib/python3.9/site-packages/holoviews/util/__init__.py:807, in save(obj, filename, fmt, backend, resources, toolbar, title, **kwargs)
    805     if formats[-1] in supported:
    806         filename = '.'.join(formats[:-1])
--> 807 return renderer_obj.save(obj, filename, fmt=fmt, resources=resources,
    808                          title=title)

File /storage/work/c/cnd5285/mambaforge/envs/uviz/lib/python3.9/site-packages/holoviews/plotting/renderer.py:594, in Renderer.save(self_or_cls, obj, basename, fmt, key, info, options, resources, title, **kwargs)
    592         if fmt in MIME_TYPES:
    593             basename = '.'.join([basename, fmt])
--> 594     plot.layout.save(basename, embed=True, resources=resources, title=title)
    595     return
    597 rendered = self_or_cls(plot, fmt)

File /storage/work/c/cnd5285/mambaforge/envs/uviz/lib/python3.9/site-packages/panel/viewable.py:985, in Viewable.save(self, filename, title, resources, template, template_variables, embed, max_states, max_opts, embed_json, json_prefix, save_path, load_path, progress, embed_states, as_png, **kwargs)
    939 def save(
    940     self, filename: str | os.PathLike | IO, title: Optional[str] = None,
    941     resources: Resources | None = None, template: str | Template | None = None,
   (...)
[...]

PR #5904, opened 2023-09-23
Actually enable Mathjax with enable_mathjax

Hoxbro:
This should now enable Mathjax without both in a notebook and when running `panel serve`.

``` python
import holoviews as hv
import panel as pn

hv.extension("bokeh", enable_mathjax=True)

a = hv.Rectangles([[0, 0, 1, 1]]).opts(title=r"$$\alpha \beta$$", xlabel="$$y$$")

pn.panel(a).servable()
```

Should fix the problem described in https://github.com/holoviz/holoviews/issues/5740#issuecomment-1705372739

### DIFF
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -715,6 +715,9 @@ class extension(_pyviz_extension):
 
         import panel as pn
 
+        if params.get("enable_mathjax", False) and selected_backend == "bokeh":
+            pn.extension("mathjax")
+
         if pn.config.comms == "default":
             if "google.colab" in sys.modules:
                 pn.config.comms = "colab"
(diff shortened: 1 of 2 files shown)


holoviews/util/__init__.py
```diff
@@ -725,6 +725,10 @@ def __call__(self, *args, **params):
                 self._ignore_bokeh_warnings()
                 return
 
+        if params.get("enable_mathjax", True) and selected_backend == "bokeh":
+            pn.extension("mathjax")
```
Hoxbro, line 729: I'm not sure if this is the best way to do it, open for suggestion @philippjfr

holoviews/util/__init__.py
```diff
@@ -725,6 +725,10 @@ def __call__(self, *args, **params):
                 self._ignore_bokeh_warnings()
                 return
 
+        if params.get("enable_mathjax", True) and selected_backend == "bokeh":
```
Hoxbro, line 728:
```suggestion
        if params.get("enable_mathjax", False) and selected_backend == "bokeh":
```

PR #5909, opened 2023-09-27
Use Python 3.12 for core test

Hoxbro:
Test to see if everything runs on the latest version of Python 3.12


setup.py
```diff
@@ -38,7 +38,6 @@
     'bokeh',
     'pillow',
     'plotly >=4.0',
-    'dash >=1.16',
```
Hoxbro, line 41: Not a core package. Moved it down to tests.

Hoxbro:
Need the libmamba solver to work on Python 3.12.

holoviews/core/util.py
```diff
@@ -929,7 +929,7 @@ def isfinite(val):
     if pandas_version >= Version('1.0.0'):
         if finite is pd.NA:
             return False
-        return finite & (~pd.isna(val))
+        return finite & ~pd.isna(np.asarray(val))
```
Hoxbro, line 932: From this Python 3.12 warning: ` Bitwise inversion '~' on bool is deprecated. This returns the bitwise inversion of the underlying int object and is usually not what you expect from negating a bool. Use the 'not' operator for boolean negation or ~int(x) if you really want the bitwise inversion of the underlying int.`

And val sometimes could be None or a tuple of None. 
Screenshot 2023-10-10 13 46 46

PR #5910, opened 2023-09-28
Use contourpy for contour calculations

### LINKED ISSUE #5899: Contours not displaying properly with matplotlib 3.8
In an environment running an up-to-date Holoviz set of packages:
```
geoviews = 1.10.1
holoviews = 1.17.1
bokeh = 3.2.2
panel = 1.2.3
```
I’ve discovered that both line and filled contours do not properly display if Matplotlib 3.8.0 is installed. Things work properly with MPL 3.7.3.

Here is a fully-reproducible code block:
```
import geoviews as gv
import xarray as xr

gv.extension("bokeh")

Z = xr.open_dataset(
    "http://thredds.atmos.albany.edu:8080/thredds/dodsC/atm533/2012103000_z500_era5.nc"
).z

gv.LineContours(Z.isel(time=0), kdims=["longitude", "latitude"], vdims=["z"])
```

Here are a couple of screenshots (with a background map layer) showing the expected (pre-MPL 3.8) and problematic (MPL 3.8.0) visualizations:

Good (MPL 3.7.3):
Geoviews_MPL373

Problematic (MPL 3.8.0):
Geoviews_MPL380

ianthomas23:
This is a WIP to use ContourPy for contour calculations. It was originally intended just to fix #5899, caused by a change within Matplotlib of how contour data is internally stored, but it is easier and better just to use ContourPy directly. This is faster and avoids unnecessary generation and manipulation of extra contour data such as Matplotlib `path` codes.

Notes:

1. ContourPy must be available of course, but it is a compulsory dependency of both Bokeh and Matplotlib.
2. Processing of the returned contour line and filled polygon data is performed separately as the data formats differ.
3. Matplotlib is still used for conversion of dates to and from numbers. It may be possible to replace these with NumPy or Pandas equivalents.
4. Matplotlib is also used for auto generation of required contour levels when the `levels` kwarg is an integer. This currently uses the `MaxNLocator` class. The required functionality could be directly vendored within HoloViews to completely remove the need for Matplotlib, e.g. if using the Bokeh backend.
5. There are issues using dates/times for `x`, `y`, and/or `z` arrays. There were no tests of these so have added some. Unfortunately some of these new tests such as `test_image_contours_x_datetime` pass regardless of the comparison `contour` in the `assertEqual`. I suspect that `assertEqual` silently ignores datetimes.
6. Datetime `z` data implies datetime `levels` too. To fully support this we may not be able to rely on `MaxNLocator` but might need to also use or implement some of Matplotlib's `DateLocator` functionality.

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -645,25 +645,26 @@ class contours(Operation):
             for lower_level, upper_level in zip(levels[:-1], levels[1:]):
                 filled = cont_gen.filled(lower_level, upper_level)
                 # Only have to consider last index 0 as we are using contourpy without chunking
-                points = filled[0][0]
+                if (points := filled[0][0]) is None:
+                    continue
+
                 exteriors = []
                 interiors = []
-                if points is not None:
-                    if any(data_is_datetime[0:2]):
-                        points = points_to_datetime(points)
+                if any(data_is_datetime[0:2]):
+                    points = points_to_datetime(points)
 
-                    offsets = filled[1][0]
-                    outer_offsets = filled[2][0]
+                offsets = filled[1][0]
+                outer_offsets = filled[2][0]
 
-                    # Loop through exterior polygon boundaries.
-                    for jstart, jend in zip(outer_offsets[:-1], outer_offsets[1:]):
-                        if exteriors:
-                            exteriors.append(empty)
-                        exteriors.append(points[offsets[jstart]:offsets[jstart + 1]])
+                # Loop through exterior polygon boundaries.
+                for jstart, jend in zip(outer_offsets[:-1], outer_offsets[1:]):
+                    if exteriors:
+                        exteriors.append(empty)
+                    exteriors.append(points[offsets[jstart]:offsets[jstart + 1]])
 
-                        # Loop over the (jend-jstart-1) interior boundaries.
-                        interior = [points[offsets[j]:offsets[j + 1]] for j in range(jstart+1, jend)]
-                        interiors.append(interior)
+                    # Loop over the (jend-jstart-1) interior boundaries.
+                    interior = [points[offsets[j]:offsets[j + 1]] for j in range(jstart+1, jend)]
+                    interiors.append(interior)
                 level = (lower_level + upper_level) / 2
                 geom = {
                     element.vdims[0].name:
@@ -677,16 +678,17 @@ class contours(Operation):
             for level in levels:
                 lines = cont_gen.lines(level)
                 # Only have to consider last index 0 as we are using contourpy without chunking
-                points = lines[0][0]
-                if points is not None:
-                    if any(data_is_datetime[0:2]):
-                        points = points_to_datetime(points)
+                if (points := lines[0][0]) is None:
+                    continue
 
-                    offsets = lines[1][0]
-                    if offsets is not None and len(offsets) > 2:
-                        # Casting offsets to int64 to avoid possible numpy UFuncOutputCastingError
-                        offsets = offsets[1:-1].astype(np.int64)
-                        points = np.insert(points, offsets, np.nan, axis=0)
+                if any(data_is_datetime[0:2]):
+                    points = points_to_datetime(points)
+
+                offsets = lines[1][0]
+                if offsets is not None and len(offsets) > 2:
+                    # Casting offsets to int64 to avoid possible numpy UFuncOutputCastingError
+                    offsets = offsets[1:-1].astype(np.int64)
+                    points = np.insert(points, offsets, np.nan, axis=0)
                 geom = {
                     element.vdims[0].name:
                     num2date(level) if data_is_datetime[2] else level,
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -561,7 +561,7 @@ class contours(Operation):
 
     def _process(self, element, key=None):
         try:
-            from contourpy import contour_generator, FillType, LineType
+            from contourpy import FillType, LineType, contour_generator
         except ImportError:
             raise ImportError("contours operation requires contourpy.")
 
@@ -588,7 +588,7 @@ class contours(Operation):
                                    "for filled contour calculations.")
 
             try:
-                from matplotlib.dates import num2date, date2num
+                from matplotlib.dates import date2num, num2date
             except ImportError:
                 raise ImportError("contours operation using datetimes requires matplotlib.") from None
 
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -596,6 +596,7 @@ class contours(Operation):
                 contours = contour_type([], [xdim, ydim], vdims)
                 return (element * contours) if self.p.overlaid else contours
             else:
+                # The +1 is consistent with Matplotlib's use of MaxNLocator for contours.
                 locator = MaxNLocator(levels + 1)
                 levels = locator.tick_values(zmin, zmax)
         else:
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -573,6 +573,9 @@ class contours(Operation):
         # if any data is a datetime, transform to matplotlib's numerical format
         data_is_datetime = tuple(isdatetime(arr) for k, arr in enumerate(data))
         if any(data_is_datetime):
+            if any(data_is_datetime[:2]) and self.p.filled:
+                raise RuntimeError("Datetime spatial coordinates are not supported "
+                                   "for filled contour calculations.")
             data = tuple(
                 date2num(d) if is_datetime else d
                 for d, is_datetime in zip(data, data_is_datetime)
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -546,11 +546,6 @@ class contours(Operation):
 
     def _process(self, element, key=None):
         try:
-            from matplotlib.dates import num2date, date2num
-        except ImportError:
-            raise ImportError("contours operation requires matplotlib.")
-
-        try:
             from contourpy import contour_generator, FillType, LineType
         except ImportError:
             raise ImportError("contours operation requires contourpy.")
@@ -576,6 +571,11 @@ class contours(Operation):
             if any(data_is_datetime[:2]) and self.p.filled:
                 raise RuntimeError("Datetime spatial coordinates are not supported "
                                    "for filled contour calculations.")
+            try:
+                from matplotlib.dates import num2date, date2num
+            except ImportError:
+                raise ImportError("contours operation using datetimes requires matplotlib.")
+
             data = tuple(
                 date2num(d) if is_datetime else d
                 for d, is_datetime in zip(data, data_is_datetime)
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -545,13 +545,15 @@ class contours(Operation):
 
     def _process(self, element, key=None):
         try:
-            from matplotlib.contour import QuadContourSet
-            from matplotlib.axes import Axes
-            from matplotlib.figure import Figure
             from matplotlib.dates import num2date, date2num
+            from matplotlib.ticker import MaxNLocator
         except ImportError:
             raise ImportError("contours operation requires matplotlib.")
-        extent = element.range(0) + element.range(1)[::-1]
+
+        try:
+            from contourpy import contour_generator, FillType, LineType
+        except ImportError:
+            raise ImportError("contours operation requires contourpy.")
 
         xs = element.dimension_values(0, True, flat=False)
         ys = element.dimension_values(1, True, flat=False)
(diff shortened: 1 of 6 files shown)


holoviews/operation/element.py
```diff
@@ -545,13 +545,15 @@ class contours(Operation):
...
     def _process(self, element, key=None):
         try:
-            from matplotlib.contour import QuadContourSet
-            from matplotlib.axes import Axes
-            from matplotlib.figure import Figure
             from matplotlib.dates import num2date, date2num
+            from matplotlib.ticker import MaxNLocator
         except ImportError:
             raise ImportError("contours operation requires matplotlib.")
-        extent = element.range(0) + element.range(1)[::-1]
```
ianthomas23, line 554: This is no longer used as I don't think it is required. There is explicit code to set the `x` and `y` arrays and pass them through, so the `extent` kwarg would have been ignored.

holoviews/operation/element.py
```diff
@@ -583,61 +585,99 @@ def _process(self, element, key=None):
...
+                # Only have to consider last index 0 as we are using contourpy without chunking
+                if (points := lines[0][0]) is None:
+                    continue
+
+                if any(data_is_datetime[0:2]):
+                    points = points_to_datetime(points)
+
+                offsets = lines[1][0]
+                if len(offsets) > 2:
+                    # Casting offsets to int64 to avoid possible numpy UFuncOutputCastingError
```
ianthomas23, line 670: This is a concern. It isn't needed for `test_operation.py` but is for one of the KDE tests in `test_statsoperations.py`.

holoviews/tests/operation/test_operation.py
```diff
@@ -122,11 +165,21 @@ def test_qmesh_curvilinear_edges_contours(self):
     def test_image_contours_filled(self):
         img = Image(np.array([[0, 1, 0], [3, 4, 5.], [6, 7, 8]]))
         op_contours = contours(img, filled=True, levels=[2, 2.5])
-        data = [[(0., 0.166667, 2.25), (0.333333, 0.166667, 2.25), (0.333333, 0.2, 2.25), (0., 0.222222, 2.25),
-                 (-0.333333, 0.111111, 2.25), (-0.333333, 0.055556, 2.25), (0., 0.166667, 2.25)]]
+        data = [[(-0.333333, 0.111111, 2.25), (-0.333333, 0.055556, 2.25), (0., 0.166667, 2.25),
```
ianthomas23, line 168: This is the same polygon as before but with a different start point.

holoviews/operation/element.py
Hoxbro: What was the type before hand?

Hoxbro:
> 1. ContourPy must be available of course, but it is a compulsory dependency of both Bokeh and Matplotlib.

Let us add it to `extras_require['tests_core']` in setup.py for good measure. 

> 2. Processing of the returned contour line and filled polygon data is performed separately as the data formats differ.

Ok. 

> 3. Matplotlib is still used for conversion of dates to and from numbers. It may be possible to replace these with NumPy or Pandas equivalents.

I think it should be safe to convert it to `int64`. At some point, I think matplotlib used another datetime format than epoch, which is likely why custom handling was needed.  

> 4. Matplotlib is also used for auto generation of required contour levels when the `levels` kwarg is an integer. This currently uses the `MaxNLocator` class. The required functionality could be directly vendored within HoloViews to completely remove the need for Matplotlib, e.g. if using the Bokeh backend.

Looking at the code for `MaxNLocator` it should be somewhat easy to get the needed functionality directly without having to import matplotlib.  

> 5. There are issues using dates/times for `x`, `y`, and/or `z` arrays. There were no tests of these so have added some. Unfortunately some of these new tests such as `test_image_contours_x_datetime` pass regardless of the comparison `contour` in the `assertEqual`. I suspect that `assertEqual` silently ignores datetimes.

I think assertEqual converts it before comparing it. I would suggest using pandas or numpy testing modules directly.  

> 6. Datetime `z` data implies datetime `levels` too. To fully support this we may not be able to rely on `MaxNLocator` but might need to also use or implement some of Matplotlib's `DateLocator` functionality.

Wouldn't a conversion of datetime data to int64 solve this? Maybe not perfect, but good enough. We can always add a more complex implementation if there is a demand for it.

holoviews/operation/element.py
ianthomas23: With a clean environment this is now needed to avoid the "UFunc add" error in many tests (`test_operation.py` as well as `test_statsoperations.py`). The complaining line is in numpy `function_base.py` when adding extra values to these `uint32` offsets. The values added could in theory be negative (although they are always positive in practice) and hence it doesn't like the unsigned-ness of `uint32`. `int64` is the next largest signed integer, so it seems a sensible choice.

ianthomas23:
For the conversion of datetimes to numbers, shouldn't it be floats rather than ints to include fractional seconds? For example something like
```python
as_floats = some_datetimes.astype('datetime64[s]').astype(np.float64)
back_to_datetimes = as_floats.astype('datetime64[s]')
```

philippjfr:
Really appreciate this @ianthomas23, I've been meaning to do this ever since you first mentioned contourpy.

ianthomas23:
This is functionally complete and ready for review. I have kept the use of Matplotlib for conversions between datetimes and floats but the attempted `import matplotlib` only occurs if such a conversion is required. Datetimes can be used for contour line plots, but using datetimes for x or y coordinates of filled contour plots raises a `RuntimeError` as it is not supported. 

I have added some extra tests.

Longer term we can consider replacing the use of Matplotlib's datetime to float conversions with our own. I started looking at this but it turned out to be more work than this conversion to ContourPy is and hence should not be attempted as part of this PR.

Example:
```python
import numpy as np
import holoviews as hv
hv.extension('bokeh')

x = np.linspace(0.0, 10.0, 40)
y = np.linspace(0.0, 2*np.pi, 20)
x2, y2 = np.meshgrid(x, y)
z = np.cos(x2 - 0.2*y2)*np.sin(y2)
img = hv.Image((x, y, z))
hv.operation.contours(img, filled=True).opts(cmap="rainbow4", colorbar=True, width=800)
```

Screenshot 2023-10-13 at 11 23 48

holoviews/operation/element.py
```diff
@@ -571,6 +568,15 @@ def _process(self, element, key=None):
...
         data_is_datetime = tuple(isdatetime(arr) for k, arr in enumerate(data))
         if any(data_is_datetime):
+            if any(data_is_datetime[:2]) and self.p.filled:
+                raise RuntimeError("Datetime spatial coordinates are not supported "
+                                   "for filled contour calculations.")
+
+            try:
+                from matplotlib.dates import num2date, date2num
+            except ImportError:
+                raise ImportError("contours operation using datetimes requires matplotlib.")
```
Hoxbro, line 578:
```suggestion
                raise ImportError("contours operation using datetimes requires matplotlib.") from None
```

No need to see two import errors.

ianthomas23: Nice. I didn't know this syntax.

holoviews/tests/operation/test_statsoperations.py
```diff
@@ -61,7 +61,7 @@ def test_bivariate_kde_contours(self):
         kde = bivariate_kde(bivariate, n_samples=100, x_range=(0, 1),
                             y_range=(0, 1), contours=True, levels=10)
         self.assertIsInstance(kde, Contours)
-        self.assertEqual(len(kde.data), 9)
+        self.assertEqual(len(kde.data), 11)
```
Hoxbro, line 64: Curious of why this value is changed?

holoviews/tests/operation/test_operation.py
```diff
@@ -126,13 +232,71 @@ def test_qmesh_curvilinear_edges_contours(self):
...
+            delta = 0.5*(levels[1] - levels[0])
+            assert np.min(levels) <= z.min() + delta
+            assert np.max(levels) >= z.max() - delta
+
+    def test_image_contours_filled_x_datetime(self):
+        x = np.array(['2023-09-01', '2023-09-05', '2023-09-09'], dtype='datetime64')
+        y = np.array([6, 7])
+        z = np.array([[0, 2, 0], [0, 2, 0]])
+        img = Image((x, y, z))
+        msg = 'Datetime spatial coordinates are not supported for filled contour calculations.'
```
Hoxbro, line 296: Be careful with `.` as it is treated as any character by `pytest.raises` match is treated as regex. It does not matter here, though.

holoviews/tests/operation/test_operation.py
```diff
@@ -126,13 +232,71 @@ def test_qmesh_curvilinear_edges_contours(self):
...
         self.assertEqual(op_contours, polys)
 
+    def test_image_contours_filled_auto_levels(self):
+        z = np.array([[0, 1, 0], [3, 4, 5.], [6, 7, 8]])
+        img = Image(z)
+        for nlevels in range(3, 20):
+            op_contours = contours(img, filled=True, levels=nlevels)
+            levels = [item['z'] for item in op_contours.data]
+            print(nlevels, len(levels), levels)
+            assert len(levels) <= nlevels + 1
```
Hoxbro, line 286: Could this be `==`?

holoviews/tests/operation/test_operation.py
```diff
@@ -126,13 +232,71 @@ def test_qmesh_curvilinear_edges_contours(self):
...
-        data = [[(0., 0.166667, 2.25), (0.333333, 0.166667, 2.25), (0.333333, 0.2, 2.25), (0., 0.222222, 2.25),
-                 (-0.333333, 0.111111, 2.25), (-0.333333, 0.055556, 2.25), (0., 0.166667, 2.25)]]
-        polys = Polygons(data, vdims=img.vdims[0].clone(range=(2, 2.5)))
+        # Contour level outside of data limits
+        op_contours = contours(img, filled=True, levels=[20.0, 23.0])
+        polys = Polygons([[]], vdims=img.vdims[0].clone(range=(20.0, 23.0)))
         self.assertEqual(op_contours, polys)
 
+    def test_image_contours_filled_auto_levels(self):
+        z = np.array([[0, 1, 0], [3, 4, 5.], [6, 7, 8]])
```
Hoxbro, line 280: Is the `5.` to check if I review thoroughly enough? :upside_down_face:

holoviews/util/locator.py
```diff
@@ -0,0 +1,171 @@
+"""
+Minimal set of functionality of Matplotlib's MaxNLocator to choose contour
+levels without having to have Matplotlib installed.
```
Hoxbro, line 3: Maybe add the Matplotlib version that was used for inspiration?

holoviews/util/locator.py
```diff
@@ -0,0 +1,171 @@
...
+            Offset subtracted from the data limits prior to calculating tick
+            locations.
+        """
+        if step <= 0:
+            raise ValueError("'step' must be positive")
+        self.step = step
+        self._offset = abs(offset)
+
+    def closeto(self, ms, edge):
+        # Allow more slop when the offset is large compared to the step.
```
Hoxbro, line 32: Is slop correct?

holoviews/tests/operation/test_operation.py
```diff
@@ -126,13 +232,71 @@ def test_qmesh_curvilinear_edges_contours(self):
...
+        img = Image(z)
+        for nlevels in range(3, 20):
+            op_contours = contours(img, filled=True, levels=nlevels)
+            levels = [item['z'] for item in op_contours.data]
+            print(nlevels, len(levels), levels)
```
Hoxbro, line 285:
```suggestion
```

holoviews/operation/element.py
```diff
@@ -583,61 +589,94 @@ def _process(self, element, key=None):
...
         levels = self.p.levels
         zmin, zmax = element.range(2)
-        if isinstance(self.p.levels, int):
+        if isinstance(levels, int):
             if zmin == zmax:
                 contours = contour_type([], [xdim, ydim], vdims)
                 return (element * contours) if self.p.overlaid else contours
-            data += (levels,)
+            else:
+                locator = MaxNLocator(levels + 1)
```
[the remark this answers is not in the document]
ianthomas23: I'll add a comment.

holoviews/tests/operation/test_statsoperations.py
ianthomas23: Me too. I don't know why, I haven't looked into the KDE code. On `main` the contour levels used are
```python
[0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0]
```
and on this branch they are
```python
[0.1, 0.2, 0.3, 0.4, 0.5, 0.6, 0.7, 0.8, 0.9, 1.0, 1.1]
```
Given that the data is in the range 0 to 1, the new values are better than the old I suppose.

holoviews/tests/operation/test_operation.py
ianthomas23: I'll make it a raw string.

holoviews/tests/operation/test_operation.py
ianthomas23: That is copied and pasted from previous image tests. I have confirmed that it gives the same results with or without the decimal point.

holoviews/util/locator.py
ianthomas23: Yes.

holoviews/util/locator.py
ianthomas23: It is copy-pasted from mpl, I don't wish to understand or change it!

holoviews/tests/operation/test_operation.py
ianthomas23: Oops. Well spotted, at least I didn't leave any swear words in it.

Hoxbro (approved):

Hoxbro:
Thank you for modernizing the contouring code to use state-of-the-art algorithms :+1:

PR #5918, opened 2023-10-03
Update to compatibility with latest param.rx API

philippjfr:
param2.0.0rc6 changed the API for getting the current value of an rx object to `.rx.value`


philippjfr:
Apologies for merging without tests passing but there's a bunch of blockers that currently prevent building working Panel packages that would allow the test suite to pass and this change is trivial.

Issue #5919, opened 2023-10-03
Title font size not working in plotly backend


jj-github-jj:
#### ALL software version info
holoviews v 1.17.1
#### Description of expected behavior and the observed behavior
can not change font size of title
#### Complete, minimal, self-contained example code that reproduces the issue

```
# code goes here between backticks
import holoviews as hv
import pandas as pd
import hvplot.pandas

hvplot.extension('plotly')
from plotly import graph_objs as go

# show all jupyter outputs
from IPython.core.interactiveshell import InteractiveShell
InteractiveShell.ast_node_interactivity = "all"

# Create a dataframe
df = pd.DataFrame({'x': [1, 2, 3], 'y': [4, 5, 6]})

# Plot the dataframe using hvplot.plotly
plot = df.hvplot()

# Change the font size of the title
_=plot.opts(title='test2',fontsize={'title': 26})
#_=plot.opts(title={'text':'test2','font':{'size':42}})

# Display the plot
plot

```

#### Screenshots of fix that I am having trouble creating a pull request for since I get a github error 
stating I am not a validated contributor. But anyway the fix is shown in the screenshot
<img width="1196" alt="Screenshot 2023-10-03 150412" src="https://github.com/holoviz/holoviews/assets/79281080/ccc7983f-b738-4f87-a930-16cb9d5f92df">

- [ ] I may be interested in making a pull request to address this

PR #5920, opened 2023-10-03
plotly title font size fix

jj-github-jj:
Fixes https://github.com/holoviz/holoviews/issues/5919

fixes plotly title font size


Hoxbro (commented):
Thank you for your contribution +1: 

I think it would be cleaner to do something like this. What do you think?
``` python
  def _format_title(self, key, separator=' '):
      """
      Formats the title of the plot.
      """
      title = super()._format_title(key, separator)

      if self.fontsize is not None and "title" in self.fontsize:
          title = {"text": title, "font": {"size": self.fontsize["title"]}}

      return title
```

A test that confirms that the text size is correct would also be good to have. Which should be possible with `hv.render(plot)`. 

And the failing test is likely not related to your PR.

jj-github-jj:
> Thank you for your contribution +1:
> 
> I think it would be cleaner to do something like this. What do you think?
> 
> ```python
>   def _format_title(self, key, separator=' '):
>       """
>       Formats the title of the plot.
>       """
>       title = super()._format_title(key, separator)
> 
>       if self.fontsize is not None and "title" in self.fontsize:
>           title = {"text": title, "font": {"size": self.fontsize["title"]}}
> 
>       return title
> ```
> 
> A test that confirms that the text size is correct would also be good to have. Which should be possible with `hv.render(plot)`.
> 
> And the failing test is likely not related to your PR.

Agree with your changes. Thanks

Hoxbro:
Are you up for implementing them?

PR #5921, opened 2023-10-04
Build conda without pyctdev

Hoxbro:


Hoxbro:
I will merge this and cut a dev release.

Future improvements to the infrastructure should be done when we upgrade to Hatch.

PR #5925, opened 2023-10-05
Fix OrderedDict regression

Hoxbro:
PR https://github.com/holoviz/holoviews/pull/5890 which changed `OrderedDict` to `dict`. had a regression which was discovered in hvplot https://github.com/holoviz/hvplot/pull/1164


Hoxbro:
This fix is not well received by our test suite.

Have to find the underlying issue.

Hoxbro:
An example with `hvplot`

``` python
import geoviews as gv
import xarray as xr
from holoviews.operation import contours

gv.extension("bokeh")

ds = xr.tutorial.open_dataset("air_temperature").isel(time=0)

p1 = gv.QuadMesh(ds, kdims=["lon", "lat"])
p2 = contours(p1, filled=True)
p2
```

Hoxbro:
It's not an ideal fix, but it no longer errors in our CI or the example above.

PR #5932, opened 2023-10-09
Raise exception in compare_dataset and fixes to comply

### LINKED ISSUE #5931: Testing framework fails to raise exceptions
In this code on `main`:

https://github.com/holoviz/holoviews/blob/adbfe2c6e82b08df4bbbf14e3a2421e0f2bfb2c3/holoviews/element/comparison.py#L511-L532

there are two lines in which exceptions are created but they are not `raise`d so no errors are reported to the user. This is a bug, all other uses of `cls.failureException` are `raise`d.

Hoxbro:
<s>Let's see how bad it is.</s>  It is not too bad, some small bugs. 

Resolves #5931

### DIFF
--- a/holoviews/element/comparison.py
+++ b/holoviews/element/comparison.py
@@ -17,6 +17,7 @@ methods on all objects as comparison operators only return Booleans and
 thus would not supply any information regarding *why* two elements are
 considered different.
 """
+import contextlib
 from functools import partial
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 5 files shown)


holoviews/tests/core/test_callable.py
```diff
@@ -218,29 +218,29 @@ class TestDynamicMapInvocation(ComparisonTestCase):
 
     def test_dynamic_kdims_only(self):
         def fn(A,B):
-            return Scatter([(A,2)], label=A)
+            return Scatter([(B,2)], label=A)
```
Hoxbro, line 221: This was a bug in our test.

holoviews/tests/core/test_dynamic.py
```diff
@@ -740,7 +740,7 @@ def fn2(x, y):
 
         overlaid = dmap * dmap2
         overlay = overlaid[()]
-        self.assertEqual(overlay.Scatter.I, fn(0, 0))
+        self.assertEqual(overlay.Scatter.I, fn(None, None))
```
Hoxbro, line 743: At some point, it could be that the dataframe in scatter returned (0, 0). But now it returns (None, None)

holoviews/tests/element/test_graphelement.py
```diff
@@ -227,9 +227,9 @@ def test_chord_constructor_with_vdims(self):
 
     def test_chord_constructor_self_reference(self):
         chord = Chord([('A', 'B', 2), ('B', 'A', 3), ('A', 'A', 2)])
-        nodes = np.array(
-            [[-0.5, 0.866025, 0],
-             [0.5, -0.866025, 1]]
+        nodes = pd.DataFrame(
```
Hoxbro, line 230: It could be at some point this worked with an array. This seem like a safe change to made.

holoviews/element/comparison.py
```diff
@@ -522,14 +522,17 @@ def compare_dataset(cls, el1, el2, msg='Dataset'):
...
-            if d1.dtype != d2.dtype:
-                cls.failureException(f"{msg} {dim.pprint_label} columns have different type."
-                                     + f" First has type {d1}, and second has type {d2}.")
-            if d1.dtype.kind in 'SUOV':
-                if list(d1) == list(d2):
-                    cls.failureException(f"{msg} along dimension {dim.pprint_label} not equal.")
-            else:
-                cls.compare_arrays(d1, d2, msg)
+            try:
+                np.testing.assert_equal(d1, d2)
```
Hoxbro, line 526: This is mainly to avoid exceptions if one type is `int` and the other is `float`.

Hoxbro:
Geoviews and hvplot do not seems to be affected by these changes running their test locally.

PR #5934, opened 2023-10-11
Maintenance update

Hoxbro:
Updating pre-commit and complying with new spelling mistakes. 

Also removing verbose flag for pytest as I hate how much I have to scroll to see output of tests.


holoviews/core/data/xarray.py
```diff
@@ -604,7 +604,7 @@ def select(cls, dataset, selection_mask=None, **selection):
         if dropped and not indexed:
             data = data.expand_dims(dropped)
             # see https://github.com/pydata/xarray/issues/2891
-            # since we only exapanded on dimnesions of size 1
+            # since we only expanded on dimnesions of size 1
```
Hoxbro, line 607:
```suggestion
            # since we only expanded on dimensions of size 1
```

PR #5936, opened 2023-10-12
Drop Python 3.8

Hoxbro:


philippjfr (approved):
Looks good and conforms with HEP1.

Issue #5565, opened 2022-12-16
Test suite are sensitive to run order


hoxbro:
When trying to apply parallelization with `pytest-xdist` to the test suite (in `holoviews/pytest`), I noticed that it is very dependent on the run order.  

If the order is randomized with something like `pytest-random-order` and the following command: `pytest holoviews/tests --random-order-bucket=class`, the test suite will fail. 

Looking at the code a culprit for this could be having `Store.renderers['bokeh']` out in the open and not inside the `setUp` of the class.

Hoxbro:
A reason could also be this line:
https://github.com/holoviz/holoviews/blob/b40916ed35166a40f7de9f620372b8cfe93f7713/holoviews/plotting/plotly/dash.py#L35

Hoxbro:
Another case related is `pytest holoviews/tests -k InspectorTests`

![image](https://user-images.githubusercontent.com/19758978/225260634-32525c73-9775-4c10-8bbf-5c1f935c589e.png)

Hoxbro:
Another one was discovered at: https://github.com/holoviz/holoviews/pull/5870

Where RGBElement_ImageInterfaceTests.test_reduce_to_single_values can't run with this command: `pytest holoviews/tests -k test_reduce_to_single_values but runs fine with `pytest holoviews/tests/core/`. In this case, I added a try/except as it errors out on core test.

Hoxbro:
Works: pytest holoviews/tests/core 
Fails: pytest holoviews/tests/core/data/test_imageinterface.py::RGBElement_ImageInterfaceTests::test_reduce_to_single_values


Tried running [detect-test-pollution](https://github.com/asottile/detect-test-pollution) with the following change to the codebase:

https://github.com/holoviz/holoviews/blob/c7db69525c560723d8c0f995bee1152b031e98ed/holoviews/tests/core/data/test_imageinterface.py#L461-L476

with:

``` python
class RGBElement_ImageInterfaceTests(BaseRGBElementInterfaceTests):

    datatype = 'image'

    __test__ = True

    def test_reduce_to_single_values(self):
        try:
            super().test_reduce_to_single_values()
        except Exception:
            return
        else:
            assert False
```

And used this command a couple of time `detect-test-pollution --failing-test holoviews/tests/core/data/test_imageinterface.py::RGBElement_ImageInterfaceTests::test_reduce_to_single_values --tests holoviews/tests/core`

I got the following polluted problems:
[...]

PR #5954, opened 2023-10-25
Part 1, modernize test suite

Hoxbro:
This is part 1 of many parts, which will begin two tasks for the HoloViews test suite: 

1. Drop the standard library `unittest` framework with `self.assertEqual` and replace it with pytest. 
    1.a. I plan to have an `assert_element_equal` for comparison of holoviews elements and then use `assert`, `{np, pd, xr}.testing` for types not directly HoloViews type. This will make it easier to see what is being tested at a glance. 
   2.a. Be able to use fixtures. I have in this PR added `bokeh_backend`, `mpl_backend`, and `plotly_backend`. 
2. Make the test run in random order. Which then makes it possible to run with `pytest-xdist`.  ref #5565

The focus of this PR has been updating the test, which saves files to disk; this has been done by using pytest `tmp_path` and minor updates to support `pathlib` for these methods. 

The second part is `test_statselements.py`, where `TestStatisticalCompositor` was sensitive to the run order because it needed `hv.renderer('matplotlib')` to be called.  I have updated the whole file to pytest.


holoviews/ipython/__init__.py
```diff
@@ -44,6 +44,7 @@ def setUp(self):
         except Exception:
                 raise SkipTest("IPython could not be started")
 
+        self.ip.displayhook.flush = lambda: None  # To avoid gc.collect called in it
```
Hoxbro, line 47: This is not directly related to any failing test but gives a significant overhead when running the test suite.

philippjfr (approved):

Issue #5951, opened 2023-10-23
Support GeoPandas passthrough to Datashader


ianthomas23:
The upcoming 0.16 release of Datashader will include support for datashading GeoPandas geometric primitives (holoviz/datashader#1285 and holoviz/datashader#1293). HoloViews should be modified to also support this rather than the current behaviour of converting to SpatialPandas equivalents for datashading.

PR #5958, opened 2023-10-25
Add support for rasterizing geopandas dataframes directly

philippjfr:
Should consider whether to move the geopandas interface from geoviews to holoviews, but here's it working:

```python
import geopandas as gpd
import geoviews
import holoviews as hv

from geodatasets import get_path

hv.extension('bokeh')

nybb = gpd.read_file(get_path("nybb"))

polys = hv.Polygons(nybb, datatype=['geodataframe'])

rasterize(polys)
```

bokeh_plot - 2023-10-25T175156 792

Fixes https://github.com/holoviz/holoviews/issues/5951


jbednar:
Nice! My understanding from @ianthomas23 was that the geopandas rendering was slower than the spatialpandas rendering. Are there any implications that we need to deal with? E.g. do we need to let users decide between directly rendering as geoparquet and converting to spatialpandas first?  If the rendering is a lot faster, converting to sp might take more time initially and use more memory but then give faster zooms. I haven't tried the two alternatives, though, so this is just speculation.

ianthomas23:
We are in a difficult situation here with regard to automatic conversion from geopandas to spatialpandas or not. If we had never previously converted then we would be fine to always leave the data in the format passed to us, and provide documentation on how to convert between the two. But if we turn off automatic conversion from geopandas to spatialpandas then users with a certain class of problem (data above a certain size) will experience a performance loss.

jbednar:
Yes, but we can also go by whether SpatialPandas is installed, using it if it's installed and passing through to GeoPandas support if it isn't. I think that should cover most cases right now, and if eventually we stopped using SpatialPandas, that branch would never get executed, and we can silently move on.

philippjfr:
The explicit thing to do is to respect what the user provided, if they passed in geopandas we should keep in geopandas.

PR #5983, opened 2023-11-14
Use sys.executable for check_output

Hoxbro:
This seems more correct


Hoxbro:
@MeggyCal, will this still work for you?

MeggyCal:
Yes, that would still work :) thank you.

PR #5985, opened 2023-11-16
Update contour line calculations to use ContourPy's LineType.ChunkCombinedNan

### LINKED ISSUE #5957: Update contour line calculations to use ContourPy's LineType.ChunkCombinedNan
The upcoming 1.2 release of ContourPy will include a data format for contour lines tailored to how Bokeh and HoloViews accept lines as NaN-separate numpy arrays (https://github.com/contourpy/contourpy/pull/296). When this release is out on PyPI and conda-forge, we need to update HoloViews to use this new `LineType.ChunkCombinedNan` if `contourpy >= 1.2` is installed. This will improve the performance of calculating contour lines (strictly speaking the calculation takes about the same amount of time but the approach avoids allocating new numpy arrays to insert the NaN into).

Code changes required (wrapped in a version check) in `holoviews/operation/element.py`:

- The `line_type` in the `contour_generator` call.
- Handling the returned data from the `ContourGenerator.lines()` call.

ianthomas23:
Fixes #5957.

ContourPy 1.2.0 includes a new `LineType` called `ChunkCombinedNan` that returns the coordinates of multiple contour lines as a single 2D array with `nan` used as a separator between different lines. This is the format used internally by HoloViews for contour lines, and it is faster and uses less RAM if ContourPy inserts the `nan` when the contours are calculated rather than HoloViews doing it afterwards.

There is a version check for ContourPy and if `>= 1.2` then `LineType.ChunkCombinedNan` is used, otherwise the original `LineType.ChunkCombinedOffset` is used and HoloViews inserts the `nan`.  The new complexity is that if we are using datetime coordinates then we cannot pass `nan` to Matplotlib's `num2date` function, so here such `nan` are identified and set to zero using a mask, then `num2date` is called and the mask used again to set those values back to `nan`. This is unfortunate but I cannot see a better way that doesn't involve writing a new version of `num2date`.

Locally tested using both ContourPy 1.1.1 and 1.2.0.

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -561,7 +561,7 @@ class contours(Operation):
 
     def _process(self, element, key=None):
         try:
-            from contourpy import FillType, LineType, contour_generator
+            from contourpy import FillType, LineType, contour_generator, __version__ as contourpy_version
         except ImportError:
             raise ImportError("contours operation requires contourpy.")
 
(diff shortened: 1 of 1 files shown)


ianthomas23:
The reduction in coverage is the `contourpy < 1.2` code path that we aren't explicitly testing in CI when we use `contourpy >= 1.2`.

Hoxbro:
LGTM, thank you for the PR :+1:

PR #5994, opened 2023-11-21
Handle rasterized output which is ImageStack

Hoxbro:
Because rasterize now can output an `ImageStack`, we need to also allow it to accept `color_key`

### DIFF
--- a/holoviews/tests/operation/test_datashader.py
+++ b/holoviews/tests/operation/test_datashader.py
@@ -1,6 +1,7 @@
 import datetime as dt
 from unittest import SkipTest, skipIf
 
+import colorcet as cc
 import numpy as np
 import pandas as pd
 import pytest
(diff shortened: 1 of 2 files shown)


holoviews/tests/operation/test_datashader.py
```diff
@@ -1530,3 +1531,17 @@ def test_uint64_dtype():
...
         rasterize(curve, dynamic=False, height=10, width=10)
+
+
+def test_imagestack_datashader_color_key():
+    d = np.arange(23)
+    df = pd.DataFrame({"x": d, "y": d, "language": list(map(str, d))})
+    points = Points(df, ["x", "y"], ["language"])
+
+    # This will run rasterize which outputs an ImageStack
+    op = datashade(
```
jlstevens, line 1542: I don't get why `datashade` is the operation used here instead of `rasterize` directly - `shade` means server-side colormixing which returns an RGB. Why not test that `rasterize` accepts `color_key` more directly?

jlstevens:
Looks good to me!

jlstevens (approved):

PR #5997, opened 2023-11-23
Enforce labels

Hoxbro:
It is always taking longer than it should for me to sort the changelog. This is an attempt to reduce the time needed by tagging a PR with one of the following labels: `type: bug`, `type: enhancement`, `type: feature`, `type: docs`, `type: maintenance`, `type: compatibility`, and then have the automatic changelog generator sort them for me.


Hoxbro:
Will merge this and see how it goes. We can always revert if it to big of a hassle.

Issue #5908, opened 2023-09-26
Using `rasterize` on the source plot breaks the RangeToolLink


droumis:
Starting with the [subcoordinate_y PR](https://github.com/holoviz/holoviews/pull/5840), running `rasterize` on a target plot image of a RangeToolLink (aka minimap) breaks the link. There is no error, but the RangeToolLink Box no longer appears on the target.

<details> 
<summary> Code </summary>

```
import numpy as np
import holoviews as hv
from bokeh.models import HoverTool
from holoviews.plotting.links import RangeToolLink
from scipy.stats import zscore
from holoviews.operation.datashader import rasterize

hv.extension('bokeh')

N_CHANNELS = 10
N_SECONDS = 5
SAMPLING_RATE = 200
INIT_FREQ = 2  # Initial frequency in Hz
FREQ_INC = 5  # Frequency increment
AMPLITUDE = 1

# Generate time and channel labels
total_samples = N_SECONDS * SAMPLING_RATE
time = np.linspace(0, N_SECONDS, total_samples)
channels = [f'EEG {i}' for i in range(N_CHANNELS)]

# Generate sine wave data
data = np.array([AMPLITUDE * np.sin(2 * np.pi * (INIT_FREQ + i * FREQ_INC) * time)
                     for i in range(N_CHANNELS)])

hover = HoverTool(tooltips=[
    ("Channel", "@channel"),
    ("Time", "$x s"),
    ("Amplitude", "$y µV")
])

channel_curves = []
for channel, channel_data in zip(channels, data):
    ds = hv.Dataset((time, channel_data, channel), ["Time", "Amplitude", "channel"])
    curve = hv.Curve(ds, "Time", ["Amplitude", "channel"], label=channel)
    curve.opts(
        subcoordinate_y=True, color="black", line_width=1, tools=[hover],
    )
    channel_curves.append(curve)

eeg = hv.Overlay(channel_curves, kdims="Channel").opts(
    xlabel="Time (s)", ylabel="Channel", show_legend=False, aspect=3, responsive=True,
)

y_positions = range(N_CHANNELS)
yticks = [(i , ich) for i, ich in enumerate(channels)]

z_data = zscore(data, axis=1)

minimap = rasterize(hv.Image((time, y_positions , z_data), ["Time (s)", "Channel"], "Amplitude (uV)"))
minimap = minimap.opts(
[...]

PR #6003, opened 2023-12-01
Fix linking elements that are transformed by a Compositor

maximlt:
Fixes https://github.com/holoviz/holoviews/issues/5908

The issue to fix could basically be simplified as linking a rasterized element to any other element:
image

`rasterize` generates `Image` elements which have at least one `Compositor` registered, e.g. `apply_nodata`. The `Compositor` transformation is applied in the following lines. I noticed that, in the example above, the `_plot_id` value of the `DynamicMap` changed after `obj.map(...)` is executed, but not of the left (target) `Curve`. And indeed, re-setting "manually" the original `_plot_id` value of the `DynamicMap` allowed the link to be correctly setup.

https://github.com/holoviz/holoviews/blob/7a4fc21149840656c9dd81464e985893499cf4c3/holoviews/core/options.py#L1002-L1004

The `obj.map(...)` call is pretty complex, however, the issue can be reproduced with a much simpler version that doesn't involve any `Compositor`. In this example the mapped operation is a bare identity function and `datashade` is used instead of `rasterize` to avoid involving any `Compositor`. This is enough for the code to fail at linking the elements:

```python
import numpy as np
import holoviews as hv
hv.extension('bokeh')

curve = hv.Curve(np.arange(10))
dcurve = datashade(curve)
RangeToolLink(dcurve, curve, axes=['x', 'y'])
layout = (curve + dcurve).opts(shared_axes=False)
layout = layout.map(lambda obj: obj)  # Breaks the link
layout
```

image

I tried to adjust `LabelledData.map()` so that it doesn't break the links but I wasn't very successful. So instead I decided to just try to reset the `_plot_id`  values after `obj.map(...)` is called, and it worked. Interestingly that also allowed to remove the `Overlay.clone` override added in https://github.com/holoviz/holoviews/pull/5881, as in that PR this override was only necessary to link from an `Image` to an `Overlay`, which is now covered by the fix from this PR.

Trying this fix against the code provided in the issue:
Kapture 2023-12-01 at 21 08 13

@philippjfr :

- How robust do you think that change is?
- It might well be this is more of a workaround for a proper fix in `LabelledData.map()` (that calls `obj.clone(shared=False)` which prevent the links from being preserved IIUC) or in some other place.

### DIFF
--- a/holoviews/plotting/util.py
+++ b/holoviews/plotting/util.py
@@ -1129,6 +1129,8 @@ def hex2rgb(hex):
 
 class apply_nodata(Operation):
 
+    link_inputs = param.Boolean(default=True)
+
     nodata = param.Integer(default=None, doc="""
         Optional missing-data value for integer data.
         If non-None, data with this value will be replaced with NaN so
--- a/holoviews/tests/plotting/bokeh/test_links.py
+++ b/holoviews/tests/plotting/bokeh/test_links.py
@@ -1,6 +1,6 @@
 import numpy as np
 import pytest
-from bokeh.models import ColumnDataSource
+from bokeh.models import ColumnDataSource, RangeTool
 
 from holoviews.core.spaces import DynamicMap
 from holoviews.element import Curve, Image, Path, Points, Polygons, Scatter, Table
@@ -12,7 +12,6 @@ from .test_plot import TestBokehPlot, bokeh_renderer
 class TestLinkCallbacks(TestBokehPlot):
 
     def test_range_tool_link_callback_single_axis(self):
-        from bokeh.models import RangeTool
         array = np.random.rand(100, 2)
         src = Curve(array)
         target = Scatter(array)
@@ -26,7 +25,6 @@ class TestLinkCallbacks(TestBokehPlot):
         self.assertIs(range_tool.y_range, None)
 
     def test_range_tool_link_callback_single_axis_overlay_target(self):
-        from bokeh.models import RangeTool
         array = np.random.rand(100, 2)
         src = Curve(array)
         target = Scatter(array, label='a') * Scatter(array, label='b')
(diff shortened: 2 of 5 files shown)


droumis:
great work! works well for me

holoviews/plotting/util.py
```diff
@@ -1129,6 +1129,8 @@ def hex2rgb(hex):
 
 class apply_nodata(Operation):
 
+    link_inputs = param.Boolean(default=True)
```
Hoxbro, line 1132: It could be nice with a doc here.

holoviews/tests/plotting/bokeh/test_links.py
```diff
@@ -53,8 +50,41 @@ def test_range_tool_link_callback_single_axis_overlay_target_image_source(self):
...
+            kdims=['a']
+        ).redim.range(a=(0.1,1))
+        target = Curve(np.arange(10))
+        RangeToolLink(src, target)
+        layout = target + src
+        plot = bokeh_renderer.get_plot(layout)
+        tgt_plot = plot.subplots[(0, 0)].subplots['main']
+        src_plot = plot.subplots[(0, 1)].subplots['main']
+        range_tool = src_plot.state.select_one({'type': RangeTool})
+        self.assertEqual(range_tool.x_range, tgt_plot.handles['x_range'])
```
Hoxbro, line 66: I'm slowly trying to move away from unittest to pytest (will take time). So I would prefer to use `assert` here, even though you are mimicking the existing code.

holoviews/util/__init__.py
```diff
@@ -900,8 +900,10 @@ def __call__(self, map_obj, **params):
         callback = self._dynamic_operation(map_obj)
         streams = self._get_streams(map_obj, watch)
         if isinstance(map_obj, DynamicMap):
-            dmap = map_obj.clone(callback=callback, shared_data=self.p.shared_data,
-                                 streams=streams)
+            params = {'shared_data': self.p.shared_data}
```
Hoxbro, line 903: Is there any reason not to put all kwargs in the params dictionary?

philippjfr: Nope, I was planning to do that when we were working together on it but ran out of time.

holoviews/plotting/util.py
maximlt: It's inherited from the `Operation` class.

holoviews/tests/plotting/bokeh/test_links.py
maximlt: Done in https://github.com/holoviz/holoviews/pull/6003/commits/c6460fc03cbacec592d20c0f234fbaa63421e5ec

PR #6005, opened 2023-12-04
Fixing URLs to bokeh project

magic-lantern:
Noticed that the link to Bokeh examples don't work. I think this is the new and correct URL.


Hoxbro:
Thank you for the updated links :+1:

PR #6007, opened 2023-12-05
Add extra check to detect if we are in jupyterlite

Hoxbro:
`_JUPYTERLAB` is no longer there when running jupyterlite, so found another way to detect it. 

https://jupyter.org/try-jupyter/lab/

Also adding `pyodide_http.patch_all()` like what is done in Panel.

### DIFF
--- a/holoviews/pyodide.py
+++ b/holoviews/pyodide.py
@@ -77,7 +77,7 @@ def render_svg(element):
 
 def in_jupyterlite():
     import js
-    return hasattr(js, "_JUPYTERLAB")
+    return hasattr(js, "_JUPYTERLAB") or hasattr(js, "webpackChunk_jupyterlite_pyodide_kernel_extension")
 
 #-----------------------------------------------------------------------------
 # Public API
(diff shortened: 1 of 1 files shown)


philippjfr:
>Also adding pyodide_http.patch_all() like what is done in Panel.

Doesn't that automatically get inherited from Panel?

holoviews/pyodide.py
```diff
@@ -17,6 +17,12 @@
 #-----------------------------------------------------------------------------
 _background_task = set()
 
+try:
+    import pyodide_http
+    pyodide_http.patch_all()
+except Exception:
+    pass
```
philippjfr, line 24: Does not seem needed, import `panel.io.pyodide` will already do that.

Hoxbro: You are right.

Hoxbro: I got a bit confused as to why I couldn't download files in Jupyterlite, but it was a cors issue.

Issue #6013, opened 2023-12-07
DynamicMap callable called unnecessarily in Overlay containing sources of RangeXY stream


TheoMathurin:
Holoviews 1.18.1
Panel 1.3.4
Bokeh 3.3.1

Maybe related to #4961 but I'm filing a separate issue as the behaviour I'm reporting does not happen on earlier holoviews versions (_e.g._ 1.17.0).

Consider the code below.

```python
import numpy as np
import holoviews as hv
import panel as pn

pn.extension()


def make_img(value):
    print('make_img called')
    return hv.Image(np.sin(xx)*np.cos(yy), bounds=(-1, -1, 1, 1))


def make_points(x_range, y_range):
    return hv.Points([(0, 0)]).opts(size=10, color='red')


ls = np.linspace(0, 10, 200)
xx, yy = np.meshgrid(ls, ls)
dummy_int = pn.widgets.IntInput(name='Dummy int', value=5)
xy_range = hv.streams.RangeXY()
points = hv.DynamicMap(make_points, streams=[xy_range])
img = hv.DynamicMap(pn.bind(make_img, value=dummy_int))

col = pn.Column(dummy_int, img * points)
col.servable()
```

Previously, and as you would expect, the `make_img` callback would be called once on render and then obviously if you changed the value of the driving parameter.

Now, it's called two times on initial render and then each time you zoom or pan and thus trigger range updates, while the associated `DynamicMap` has nothing to do with XY ranges.

The number of unnecessary executions on render increases as you add more elements that depend on `RangeXY` (including _e.g._ datashaded elements). As a result in my real-world app, I have several callables that are each called 4 or 5 times on initial render. I do not think this is intended behaviour.

Issue #6008, opened 2023-12-05
Can't mix pn.bind with streams in DynamicMaps anymore


TheoMathurin:
HoloViews 1.18.1
Param 2.0.1
Panel 1.3.4

#### Description of expected behavior and the observed behavior
I used to be able to wrap a `pn.bind` within a DynamicMap and add a `streams` argument like the following:
`dmap = hv.DynamicMap(pn.bind(function, arg=some_widget), streams=[some_stream])`

This is useful for plots driven both by widgets and streams.

But now this causes an error where expected inputs are missing, maybe related to the new Param version.

#### MRE

In a panel server

```python
import holoviews as hv
import panel as pn

pn.extension()


def make_plot(z, x_range, y_range):
    print(z)
    print(x_range, y_range)
    return hv.Curve([1, 2, 3, 4])


slider = pn.widgets.IntSlider(name='Slider', start=0, end=10)
range_xy = hv.streams.RangeXY()
dmap = hv.DynamicMap(pn.bind(make_plot, z=slider), streams=[range_xy])

pn.Column(slider, dmap).servable()
```

#### Stack traceback and/or browser JavaScript console output
```python
return eval_fn()(*combined_args, **combined_kwargs) Traceback (most recent call last):
  File "/home/mathurin/miniconda3/envs/hv/lib/python3.12/site-packages/bokeh/application/handlers/code_runner.py", line 229, in run
    exec(self._code, module.__dict__)
  File "/home/mathurin/PycharmProjects/test/toast.py", line 18, in <module>
    pn.Column(slider, dmap).servable()
  File "/home/mathurin/miniconda3/envs/hv/lib/python3.12/site-packages/panel/viewable.py", line 394, in servable
    self.server_doc(title=title, location=location) # type: ignore
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mathurin/miniconda3/envs/hv/lib/python3.12/site-packages/panel/viewable.py", line 1023, in server_doc
    model = self.get_root(doc)
            ^^^^^^^^^^^^^^^^^^
  File "/home/mathurin/miniconda3/envs/hv/lib/python3.12/site-packages/panel/layout/base.py", line 309, in get_root
    root = super().get_root(doc, comm, preprocess)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #6014, opened 2023-12-07
Only evaluate `rx` if it is an Reactive Expression 

Hoxbro:
should improve what is seen in #6013

Want to add a UI test that checks that the bound function is only run once.


TheoMathurin:
With this fix, the function is called only once.

Also, it does not reintroduce https://github.com/holoviz/holoviews/issues/6008

PR #6017, opened 2023-12-08
Add viewport downsample algorithm

Hoxbro:
Simple proof-of-concept of how we can use downsample to only send the data in the viewport to the frontend. 

``` python
import holoviews as hv
import numpy as np

hv.extension("bokeh")
downsample1d(hv.Curve(np.random.rand(1000)), algorithm="viewport").opts(xlim=(0, 10))
```

https://github.com/holoviz/holoviews/assets/19758978/0826bef1-7a00-4d02-8ddb-4735eecf10a4


droumis:
@Hoxbro . This would be a really fantastic addition. I want to make sure that it works for all the relevant use cases that we are encountering right now. 

First, it looks like any type of downsampling does not work with `subcoordinate_y=True`

<details> <summary> Code </summary>

```python
import holoviews as hv
import numpy as np
hv.extension("bokeh")

c1 = hv.Curve(np.random.rand(1000), label='c1').opts(subcoordinate_y=True)
c2 = hv.Curve(np.random.rand(1000), label='c2').opts(subcoordinate_y=True)

curves = hv.Overlay(c1*c2).opts(xlim=(5, 10))
downsample1d(curves, algorithm="viewport")
```

</details>

https://github.com/holoviz/holoviews/assets/6613202/6c67ceb4-f1b0-4738-8602-47f9e8aa42ff

Hoxbro:
`subcoordinate_y=True` should now work:


https://github.com/holoviz/holoviews/assets/19758978/e4b71c6a-5a49-4afc-9929-23579c29c187

holoviews/operation/downsample.py
```diff
@@ -176,7 +181,9 @@ def _process(self, element, key=None):
             return element.clone(elements)
 
         if self.p.x_range:
-            element = element[slice(*self.p.x_range)]
+            xs = element.dimension_values(0)
+            x0, x1 = (np.argmin(np.abs(xs-coord)) for coord in self.p.x_range)
+            element = element.iloc[max(x0 - 1, 0):(x1 + 2)]
```
philippjfr, line 186: What was wrong with the previous approach?

philippjfr: This seems like bad practice because it means you are forcing out-of-core data to be loaded into memory just to apply this slicing. If the previous approach didn't work properly then the data interfaces need to be fixed instead of implementing a custom solution here.

Hoxbro: > What was wrong with the previous approach?

This could leave white space at the start/end of the plot (using the original example). Indicates that there is no more data, even when there is.  

image

> This seems like bad practice because it means you are forcing out-of-core data to be loaded into memory just to apply this slicing.

Noted, but this is still very much WIP, so it is not ready for a critical review. Even more, this is pretty harshly written, which could have been better framed as a question.  

>  If the previous approach didn't work properly then the data interfaces need to be fixed instead of implementing a custom solution here.

I don't think there is anything wrong with the data interface, here I just want the next datapoints outside of the viewport. This could be upstreamed to the data interface, but I'm not sure about the implementation yet.

philippjfr: Sounds good, I think we have long needed but never implemented snapping semantics for the `DataInterface.select` operation. Implementing it for all data interfaces is definitely a good bit of work though.

>Noted, but this is still very much WIP, so it is not ready for a critical review. Even more, this is pretty harshly written, which could have been better framed as a question.

That certainly wasn't the intent, I think without having that consideration being explicitly pointed out there's much chance that anyone would have considered this without having dask and ibis backends top-of-mind. So it was meant as a heads up, not a criticism.

Hoxbro:
I have added a workaround when having an x limit set on the element (`downsample1d(hv.Curve(np.ones(1_000)), algorithm="viewport").opts(xlim=(5, 10))`).  This is not very elegant, but I can't spend more time on it. I have talked with Philipp, and did not know a better implementation at the top of his head, but maybe he has a better workaround, which at least does not need to have a separate algorithm to handle it. 

The problem is the downsampling works by looking at the stream `x_range`. The first run will be `None`, meaning the viewport algorithm will send all data to the browser. When the plot shows up in the browser a stream event will trigger and change the  `x_range`  to the  `xlim` range. This is not as big of a problem for the other algorithms as they do an actual downsample to the width pixel size (400 by default).

Hoxbro:
We can't send an empty plot either, as this will make the y-axis between 0 and 1, didn't notice this before as I was using random data. 

So I'm now sending 400 points (200 from the start and 200 from the end) to get an approximation of the range.

holoviews/operation/downsample.py
```diff
@@ -176,7 +183,33 @@ def _process(self, element, key=None):
...
+            # backend with the viewport algorithm does not make any sense.
+            # This is not very elegant.
+            size = element.dataset.shape[0]
+            mask1 = element.dataset.interface.select_mask(
+                element.dataset, {element.kdims[0]: slice(self.p.width // 2)},
+            )
+            mask2 = element.dataset.interface.select_mask(
+                element.dataset, {element.kdims[0]: slice(size - self.p.width // 2, size)},
+            )
+            element = element[mask1 | mask2]
```
Hoxbro, line 211: An alternative is to switch to `nth`  downsample when `self.p.algorithm == "viewport-xlim" and self.p.x_range is None`.

droumis:
#6061 is relevant, given that this PR is primarily slicing

philippjfr:
Set the stream value during the stream Callback initialization:

```
    def initialize(self, plot_id=None):
        super().initialize(plot_id)
        for stream in self.streams:
            msg = self._process_msg({})
            stream.update(**msg)
```

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -404,6 +404,7 @@ def set_callback(self, handle):
         if self.on_changes:
             change_handler = lambda attr, old, new: (
                 asyncio.create_task(self.on_change(attr, old, new))
+                 if self.plot.document else None
```
Hoxbro, line 407:
```suggestion
                if self.plot.document else None
```

philippjfr (approved):
Looks good, thanks!

PR #6024, opened 2023-12-11
Support ImageStack in dynspread

### LINKED ISSUE #5993: Support new `ImageStack` element in `dynspread` operation
Presumably this needs to be fixed in `spread`.

```python
import pandas as pd
from holoviews.operation.datashader import dynspread
df = pd.DataFrame({'x':[-16.8, 7.3], 'y': [-0.42, 13.6], 'language':['Marathi', 'Luganda']})
points = hv.Points(df, ['x','y'], ['language'])
op = rasterize(points, aggregator=ds.by('language', ds.count()))
dynspread(op)  # Works when dynspread not applied
```

hoxbro:
Resolves #5993

https://github.com/holoviz/holoviews/assets/19758978/c11c3e78-eefd-4aff-98a2-2ff1da983104


jlstevens:
Glad to see that the current codepath in datashader seems to work!

Two things to note:

1. As an extra validation, I would like to see dynspread using datashade next to dynspread using ImageStack - if everything is working correctly they should basically look the same.
2. I think more work is needed to support imagestack fully in the way we are imagining. For instance, supporting the spreading of layers that are not colormixed and included to added hover information. Such layers may not be numeric (e.g. text!) but should also be spread. The way I would imagine it is as if you are spreading a i64 image (an index per unique value) and then using those values as indices back to the data.

While the first thing would be good to check before merging this PR, the second thing is a fair bit more work and can be made into a feature request and addressed in a future PR.

droumis:
@philippjfr: "To efficiently spread multiple layers, we may have to do this in Datashader"

hoxbro:
I will add a small test and then merge this. We can always improve functionality later.

PR #6025, opened 2023-12-11
Support dictionary cmaps for ImageStack

### LINKED ISSUE #6012: ImageStack should support dictionaries for cmap
This currently works with a list:

```python
import numpy as np
x = np.arange(0, 3)
y = np.arange(5, 8)
a = np.array([[np.nan, np.nan, 1], [np.nan] * 3, [np.nan] * 3])
b = np.array([[np.nan] * 3, [1, 1, np.nan], [np.nan] * 3])
c = np.array([[np.nan] * 3, [np.nan] * 3, [1, 1, 1]])

img_stack = hv.ImageStack((x, y, a, b, c), kdims=["x", "y"], vdims=["a", "b", "c"])
img_stack.opts(cmap=["red", "green", "blue"])
```

But using a dictionary (which is often more convenient when you don't know the layer order, or need to map a category to a colour) does not work:

```python
import numpy as np
x = np.arange(0, 3)
y = np.arange(5, 8)
a = np.array([[np.nan, np.nan, 1], [np.nan] * 3, [np.nan] * 3])
b = np.array([[np.nan] * 3, [1, 1, np.nan], [np.nan] * 3])
c = np.array([[np.nan] * 3, [np.nan] * 3, [1, 1, 1]])

img_stack = hv.ImageStack((x, y, a, b, c), kdims=["x", "y"], vdims=["a", "b", "c"])
img_stack.opts(cmap={'a':"red", 'b':"green", 'c':"blue"})
```

Hoxbro:
Fixes #6012 

I'm not sure if I'm a fan of the solution.

### DIFF
--- a/holoviews/plotting/bokeh/raster.py
+++ b/holoviews/plotting/bokeh/raster.py
@@ -263,18 +263,11 @@ class ImageStackPlot(RasterPlot):
         if isinstance(style.get("cmap"), dict):
             dict_cmap = style["cmap"]
             missing = [vd.name for vd in vdims if vd.name not in dict_cmap]
-            extra = [k for k in dict_cmap if k not in vdims]
             if missing:
                 missing_str = "', '".join(sorted(missing))
                 raise ValueError(
                     "The supplied cmap dictionary must have the same "
                     f"value dimensions as the element. Missing: '{missing_str}'"
-                )
-            elif extra:
-                extra_str = "', '".join(sorted(extra))
-                self.param.warning(
-                    f"The supplied cmap dictionary has extra value dimensions: "
-                    f"'{extra_str}'. Ignoring these value dimensions."
                 )
             keys, values = zip(*dict_cmap.items())
             style["cmap"] = list(values)
--- a/holoviews/plotting/bokeh/raster.py
+++ b/holoviews/plotting/bokeh/raster.py
@@ -275,7 +275,8 @@ class ImageStackPlot(RasterPlot):
             colors=colors, group=group, name=name
         )
         if dict_cmap:
-            indices = [list(dict_cmap).index(vd.name) for vd in element.vdims]
+            keys_cmap = list(dict_cmap)
+            indices = [keys_cmap.index(vd.name) for vd in element.vdims]
         else:
             num_elements = len(element.vdims)
             step_size = len(cmapper.palette) // num_elements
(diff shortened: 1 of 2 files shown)


holoviews/plotting/bokeh/raster.py
```diff
@@ -245,13 +245,28 @@ def _get_cmapper_opts(self, low, high, factors, colors):
 
     def _get_colormapper(self, eldim, element, ranges, style, factors=None,
                          colors=None, group=None, name='color_mapper'):
+        dict_cmap = False
+        if isinstance(style.get("cmap"), dict):
+            dict_cmap = style["cmap"]
+            style["cmap"] = list(dict_cmap.values())
```
jlstevens, line 251: Logically it makes more sense to validate for missing value dimensions before setting this (though for readability I can see why you might prefer this order)

Hoxbro: If we don't convert the `style["cmap"]` to a list the colormapper will be of type `CategoricalColorMapper`, and `WeightedStackColorMapper` only accepts a `ContinuousColorMapper`. So the conversion to a list is to get the allowed colormapper.

jlstevens: Thanks for the explanation. I thought it would be something like that but it is good to confirm.

ahuang11:
I think this is ready for review; tests failing because: `Package: bokeh      Current: 3.3.4   (2024-01-25)	Latest: 3.4.0   (2024-03-14)`

holoviews/plotting/bokeh/raster.py
```diff
@@ -258,13 +258,38 @@ def _get_cmapper_opts(self, low, high, factors, colors):
...
+            dict_cmap = style["cmap"]
+            missing = [vd.name for vd in vdims if vd.name not in dict_cmap]
+            extra = [k for k in dict_cmap if k not in vdims]
+            if missing:
+                missing_str = "', '".join(sorted(missing))
+                raise ValueError(
+                    "The supplied cmap dictionary must have the same "
+                    f"value dimensions as the element. Missing: '{missing_str}'"
+                )
+            elif extra:
```
jlstevens, line 273: Thinking about it, the cmap might well be a fixed mapping for categories that simply gets supplied: while missing entries is definitely a problem, I'm not sure warning when you have extra entries is going to be more helpful than annoying. 

For instance, if your element is viewing a portion of the data space where certain categories do not appear, I'm not sure you need to warn about that the key is over-specified for that particular case.

ahuang11: Okay I removed it

PR #6056, opened 2024-01-01
Expose setting hard navigable bounds

### LINKED ISSUE #1019: Defining hard bounds when panning and zooming in bokeh backend
Currently when plotting using bokeh the axes are unbounded letting you zoom and pan to potentially empty parts of the data space. Since bokeh 0.11 it has been possible to bound the axis ranges to stop you from zooming or panning outside the desired region. We should decide how to expose this? Should axes ranges always be hard bounded or should setting an explicit dimension range bound them? There's a variety of behaviors we could consider here and just wanted to get discussion around this started.

droumis:
This PR closes #1019, which pertains to the lack of hard bounds when panning and zooming using the Bokeh backend. 

Why? Previously, the axes were unbounded, allowing users to zoom and pan to potentially empty parts of the data space. This is particularly annoying when a plot captures the scroll of a webpage/notebook/app that it is a part of. This PR ~~changes the default behavior~~ adds a flag, preventing zooming or panning beyond the data range. update: uses the more extreme of data ranges or xlim/ylim. Update, if dim ranges are set, they are used as hard bounds.

To enable this new behavior, use `apply_hard_bounds=True`.

Limitation: Currently, setting hard navigable bounds with Bokeh means that if you hit a single extent (e.g. the lower x dim), the zoom won't continue to work for the remaining directions (upper x dim, or any y dim).. Bokeh has an option called `maintain_focus=False` for zoom tools to enable continued zoom. Future work could potentially enable this option in HoloViews when `apply_hard_bounds=True`.

Todo:
- [x] add tests
- [x] add tests for datetime axes
- [x] add test for dmap updates
- [x] add docs

Demo:

<details> <summary> code </summary>

```python

hb_default = hv.Curve(([1,2,3], [1,2,1]), label='apply hard bounds Default')
hb_false = hv.Curve(([1,2,3], [4,2,3]), label='apply hard bounds False').opts(apply_hard_bounds=False, color='black')
hb_true = hv.Curve(([-1,0,1,2,3,4], [3,2,1,0,-1,-2]), label='apply hard bounds True').opts(apply_hard_bounds=True, color='red')
hb_overlay = (hb_true * hb_false).opts(title='True/False overlay', show_legend=False)

(hb_default + hb_false + hb_true + hb_overlay).opts(shared_axes=False)
```

</details>

https://github.com/holoviz/holoviews/assets/6613202/c3e50116-18ff-44f0-b800-3211e790b802

Only limits the x dim for subcoord_y plots:

<details> <summary> code </summary>

```python
import numpy as np
import holoviews as hv
from bokeh.models import HoverTool
from scipy.stats import zscore

hv.extension('bokeh')

N_CHANNELS = 10
N_SECONDS = 5
SAMPLING_RATE = 200
INIT_FREQ = 2  # Initial frequency in Hz
FREQ_INC = 5  # Frequency increment
AMPLITUDE = 1

# Generate time and channel labels
total_samples = N_SECONDS * SAMPLING_RATE
time = np.linspace(0, N_SECONDS, total_samples)
channels = [f'EEG {i}' for i in range(N_CHANNELS)]

# Generate sine wave data
data = np.array([AMPLITUDE * np.sin(2 * np.pi * (INIT_FREQ + i * FREQ_INC) * time)
                     for i in range(N_CHANNELS)])

hover = HoverTool(tooltips=[
    ("Channel", "@channel"),
    ("Time", "$x s"),
    ("Amplitude", "$y µV")
])

channel_curves = []
for channel, channel_data in zip(channels, data):
    ds = hv.Dataset((time, channel_data, channel), ["Time", "Amplitude", "channel"])
    curve = hv.Curve(ds, "Time", ["Amplitude", "channel"], label=channel)
    curve.opts(
        subcoordinate_y=True, color="black", line_width=1, tools=[hover],
        apply_hard_bounds=True,
    )
    channel_curves.append(curve)

eeg = hv.Overlay(channel_curves, kdims="Channel").opts(
    xlabel="Time (s)", ylabel="Channel", show_legend=False, aspect=3, responsive=True,
)


y_positions = range(N_CHANNELS)
yticks = [(i , ich) for i, ich in enumerate(channels)]

z_data = zscore(data, axis=1)

minimap = hv.Image((time, y_positions , z_data), ["Time (s)", "Channel"], "Amplitude (uV)")
minimap = minimap.opts(
    cmap="RdBu_r", xlabel='Time (s)', alpha=.5, yticks=[yticks[0], yticks[-1]],
    height=150, responsive=True, default_tools=[], clim=(-z_data.std(), z_data.std())
)


RangeToolLink(
    minimap, eeg, axes=["x", "y"],
    boundsx=(None, 2), boundsy=(None, 6.5)
)

dashboard = (eeg + minimap).opts(merge_tools=False).cols(1)
dashboard
```

</details>

https://github.com/holoviz/holoviews/assets/6613202/2f9fb3d7-d9db-4950-a5fb-15860662b1ba

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -105,9 +105,9 @@ class ElementPlot(BokehPlot, GenericElementPlot):
     align = param.ObjectSelector(default='start', objects=['start', 'center', 'end'], doc="""
         Alignment (vertical or horizontal) of the plot in a layout.""")
 
-    apply_hard_bounds = param.Boolean(default=None, allow_None=True, doc="""
+    apply_hard_bounds = param.Boolean(default=False, doc="""
         If True, the navigable bounds of the plot will be set based
-        on the extents of the data. If False, the bounds will not be set.""")
+        on the larger of extents of the data+padding, xlim/ylim, dim ranges.""")
 
     autorange = param.ObjectSelector(default=None, objects=['x', 'y', None], doc="""
         Whether to auto-range along either the x- or y-axis, i.e.
@@ -1939,15 +1939,11 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         is unbounded.
         """
 
-        # Skip if the element doesn't have an 'extents' attribute
-        if not hasattr(element, 'extents'):
-            return
-
         def validate_bound(bound):
             """Validate a single bound, returning None if it is not a valid number"""
-            return bound if isinstance(bound, (int, float)) and not np.isnan(bound) else None
+            return bound if util.isfinite(bound) else None
 
-        min_extent_x, min_extent_y, max_extent_x, max_extent_y = map(validate_bound, self.get_extents(element, ranges))
+        min_extent_x, min_extent_y, max_extent_x, max_extent_y = map(validate_bound, self.get_extents(element, ranges, lims_as_soft_ranges=True))
 
         def set_bounds(axis, min_extent, max_extent):
             """Set the bounds for a given axis, using None if both extents are None or identical"""
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -107,7 +107,8 @@ class ElementPlot(BokehPlot, GenericElementPlot):
 
     apply_hard_bounds = param.Boolean(default=False, doc="""
         If True, the navigable bounds of the plot will be set based
-        on the larger of extents of the data+padding, xlim/ylim, dim ranges.""")
+        on the more extreme of extents between the data or xlim/ylim ranges.
+        If dim ranges are set, the hard bounds will be set to the dim ranges.""")
 
     autorange = param.ObjectSelector(default=None, objects=['x', 'y', None], doc="""
         Whether to auto-range along either the x- or y-axis, i.e.
@@ -1931,8 +1932,8 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         """
         Apply hard bounds to the x and y ranges of the plot. If xlim/ylim is set, limit the
         initial viewable range to xlim/ylim, but allow navigation up to the abs max between
-        the data + pad range and xlim/ylim. If dim range is set (e.g. via redim.range), use
-        it as the hard bounds.
+        the data range and xlim/ylim. If dim range is set (e.g. via redim.range), enforce
+        as hard bounds.
 
         """
 
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1949,7 +1949,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
             if min_extent == max_extent:
                 self.handles[axis].bounds = None
             else:
-                self.handles[axis].bounds = (min_extent, max_extent) if min_extent is not None or max_extent is not None else None
+                self.handles[axis].bounds = (min_extent, max_extent)
 
         set_bounds('x_range', min_extent_x, max_extent_x)
         set_bounds('y_range', min_extent_y, max_extent_y)
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1899,9 +1899,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
                 if style_element.label in plot.extra_y_ranges:
                     self.handles['y_range'] = plot.extra_y_ranges.pop(style_element.label)
 
-        # If apply_hard_bound is not set but projection has been set, avoid applying hard bounds by default
-        # since a projection typically means that the plot is either geographic, polar, or 3d
-        if self.apply_hard_bounds or (self.projection is None and self.apply_hard_bounds is None):
+        if self.apply_hard_bounds:
             self._apply_hard_bounds(element, ranges)
 
         self.handles['plot'] = plot
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1946,10 +1946,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
 
         def set_bounds(axis, min_extent, max_extent):
             """Set the bounds for a given axis, using None if both extents are None or identical"""
-            if min_extent == max_extent:
-                self.handles[axis].bounds = None
-            else:
-                self.handles[axis].bounds = (min_extent, max_extent)
+            self.handles[axis].bounds = None if min_extent == max_extent else (min_extent, max_extent)
 
         set_bounds('x_range', min_extent_x, max_extent_x)
         set_bounds('y_range', min_extent_y, max_extent_y)
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1928,11 +1928,34 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         return plot
 
     def _apply_hard_bound(self, element, ranges):
-        # Set the navigable bounds
-        xmin, ymin, xmax, ymax = self.get_extents(element, ranges)
-        if not all(np.isnan([xmin, ymin, xmax, ymax])):
-            self.handles['x_range'].bounds = (xmin, xmax)
-            self.handles['y_range'].bounds = (ymin, ymax)
+        """
+        Apply hard bounds to the x and y ranges of the plot.
+
+        Sets the navigable bounds of the plot based on the extents
+        of the given element and ranges. If an extend is numeric and not NaN, it is
+        used as is. Otherwise, it is set to None, which means that end of the axis
+        is unbounded.
+        """
+
+        # Skip if the element doesn't have an 'extents' attribute
+        if not hasattr(element, 'extents'):
+            return
+
+        def validate_bound(bound):
+            """Validate a single bound, returning None if it is not a valid number"""
+            return bound if isinstance(bound, (int, float)) and not np.isnan(bound) else None
+
+        min_extent_x, min_extent_y, max_extent_x, max_extent_y = map(validate_bound, self.get_extents(element, ranges))
+
+        def set_bounds(axis, min_extent, max_extent):
+            """Set the bounds for a given axis, using None if both extents are None or identical"""
+            if min_extent == max_extent:
+                self.handles[axis].bounds = None
+            else:
+                self.handles[axis].bounds = (min_extent, max_extent) if min_extent is not None or max_extent is not None else None
+
+        set_bounds('x_range', min_extent_x, max_extent_x)
+        set_bounds('y_range', min_extent_y, max_extent_y)
 
     def _setup_data_callbacks(self, plot):
         if not self._js_on_data_callbacks:
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -105,7 +105,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
     align = param.ObjectSelector(default='start', objects=['start', 'center', 'end'], doc="""
         Alignment (vertical or horizontal) of the plot in a layout.""")
 
-    apply_hard_bounds = param.Boolean(default=True, allow_None=True, doc="""
+    apply_hard_bounds = param.Boolean(default=None, allow_None=True, doc="""
         If True, the navigable bounds of the plot will be set based
         on the extents of the data. If False, the bounds will not be set.""")
 
@@ -1899,8 +1899,10 @@ class ElementPlot(BokehPlot, GenericElementPlot):
                 if style_element.label in plot.extra_y_ranges:
                     self.handles['y_range'] = plot.extra_y_ranges.pop(style_element.label)
 
-        if self.apply_hard_bounds:
-            self._apply_hard_bound(element, ranges)
+        # If apply_hard_bound is not set but projection has been set, avoid applying hard bounds by default
+        # since a projection typically means that the plot is either geographic, polar, or 3d
+        if self.apply_hard_bounds or (self.projection is None and self.apply_hard_bounds is None):
+            self._apply_hard_bounds(element, ranges)
 
         self.handles['plot'] = plot
 
@@ -1927,7 +1929,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
 
         return plot
 
-    def _apply_hard_bound(self, element, ranges):
+    def _apply_hard_bounds(self, element, ranges):
         """
         Apply hard bounds to the x and y ranges of the plot.
 
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1940,7 +1940,9 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         def validate_bound(bound):
             return bound if util.isfinite(bound) else None
 
-        min_extent_x, min_extent_y, max_extent_x, max_extent_y = map(validate_bound, self.get_extents(element, ranges, range_type='combined', lims_as_soft_ranges=True))
+        min_extent_x, min_extent_y, max_extent_x, max_extent_y = map(
+            validate_bound, self.get_extents(element, ranges, range_type='combined', lims_as_soft_ranges=True)
+        )
 
         def set_bounds(axis, min_extent, max_extent):
             """Set the bounds for a given axis, using None if both extents are None or identical"""
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1946,7 +1946,10 @@ class ElementPlot(BokehPlot, GenericElementPlot):
 
         def set_bounds(axis, min_extent, max_extent):
             """Set the bounds for a given axis, using None if both extents are None or identical"""
-            self.handles[axis].bounds = None if min_extent == max_extent else (min_extent, max_extent)
+            try:
+                self.handles[axis].bounds = None if min_extent == max_extent else (min_extent, max_extent)
+            except ValueError:
+                self.handles[axis].bounds = None
 
         set_bounds('x_range', min_extent_x, max_extent_x)
         set_bounds('y_range', min_extent_y, max_extent_y)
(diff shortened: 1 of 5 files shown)


droumis:
I was getting some test errors when an element didn't have an extents attribute, so I'm skipping when this is the case. I'm assuming hard bounds are irrelevant for such objects.

droumis:
I'm also setting bounds to None for cases where the element.extents are identical (e.g. (10.0, 10.0)) to avoid some test errors like:

<details><summary> example test error: </summary>

```

____________________ TestPointPlot.test_op_ndoverlay_value _____________________

self = <holoviews.tests.plotting.bokeh.test_pointplot.TestPointPlot testMethod=test_op_ndoverlay_value>

    def test_op_ndoverlay_value(self):
        markers = ['circle', 'triangle']
        overlay = NdOverlay({marker: Points(np.arange(i)) for i, marker in enumerate(markers)}, 'Marker').opts('Points', marker='Marker')
>       plot = bokeh_renderer.get_plot(overlay)

holoviews/tests/plotting/bokeh/test_pointplot.py:503:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
holoviews/plotting/bokeh/renderer.py:68: in get_plot
    plot = super().get_plot(obj, doc, renderer, **kwargs)
holoviews/plotting/renderer.py:240: in get_plot
    plot.update(init_key)
holoviews/plotting/plot.py:955: in update
    return self.initialize_plot()
holoviews/plotting/bokeh/element.py:2930: in initialize_plot
    child = subplot.initialize_plot(ranges, plot, plots)
holoviews/plotting/bokeh/element.py:1890: in initialize_plot
    self._apply_hard_bound(element, ranges)
holoviews/plotting/bokeh/element.py:1937: in _apply_hard_bound
    set_bounds('x_range', xmin, xmax)
holoviews/plotting/bokeh/element.py:1935: in set_bounds
    self.handles[axis].bounds = (min_bound, max_bound) if min_bound is not None or max_bound is not None else None
../../opt/miniconda3/envs/hv-dev-tests/lib/python3.9/site-packages/bokeh/core/has_props.py:334: in __setattr__
    return super().__setattr__(name, value)
../../opt/miniconda3/envs/hv-dev-tests/lib/python3.9/site-packages/bokeh/core/property/descriptors.py:331: in __set__
    value = self.property.prepare_value(obj, self.name, value)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = Nullable(MinMaxBounds(Auto, Tuple(Float, Float), Tuple(Nullable(Float), Float), Tuple(Float, Nullable(Float)), Tuple(T...able(TimeDelta)), Tuple(Datetime, Datetime), Tuple(Nullable(Datetime), Datetime), Tuple(Datetime, Nullable(Datetime))))
owner = Range1d(id='p42539', ...), name = 'bounds', value = (0.0, 0.0)

    def prepare_value(self, owner: HasProps | type[HasProps], name: str, value: Any, *, hint: DocumentPatchedEvent | None = None) -> T:
        if value is Intrinsic:
            value = self._raw_default()
        if value is Undefined:
            return value

        error = None
        try:
            if validation_on():
                hinted_value = self._hinted_value(value, hint)
                self.validate(hinted_value)
        except ValueError as e:
            for tp, converter in self.alternatives:
                if tp.is_valid(value):
                    value = converter(value)
                    break
            else:
                error = e

        if error is None:
            value = self.transform(value)
        else:
            obj_repr = owner if isinstance(owner, HasProps) else owner.__name__
>           raise ValueError(f"failed to validate {obj_repr}.{name}: {error}")
E           ValueError: failed to validate Range1d(id='p42539', ...).bounds: expected either None or a value of type MinMaxBounds(Auto, Tuple(Float, Float), Tuple(Nullable(Float), Float), Tuple(Float, Nullable(Float)), Tuple(TimeDelta, TimeDelta), Tuple(Nullable(TimeDelta), TimeDelta), Tuple(TimeDelta, Nullable(TimeDelta)), Tuple(Datetime, Datetime), Tuple(Nullable(Datetime), Datetime), Tuple(Datetime, Nullable(Datetime))), got (0.0, 0.0)

../../opt/miniconda3/envs/hv-dev-tests/lib/python3.9/site-packages/bokeh/core/property/bases.py:365: ValueError

```

</details>

droumis:
I disabled hard bounds for one of the UI tests ([test_multi_axis_rangexy](https://github.com/holoviz/holoviews/pull/6056/commits/cff8c93862b212bfee0428e9d7979331c063337e)) in order to have it pass.

holoviews/plotting/bokeh/element.py
```diff
@@ -1906,6 +1914,36 @@ def initialize_plot(self, ranges=None, plot=None, plots=None, source=None):
 
         return plot
 
+    def _apply_hard_bound(self, element, ranges):
+        """
+        Apply hard bounds to the x and y ranges of the plot.
+
+        Sets the navigable bounds of the plot based on the extents
+        of the given element and ranges. If an extend is numeric and not NaN, it is
```
philippjfr, line 1922:
```suggestion
        of the given element and ranges. If an extent is numeric and not NaN, it is
```

holoviews/plotting/bokeh/element.py
```diff
@@ -1906,6 +1914,36 @@ def initialize_plot(self, ranges=None, plot=None, plots=None, source=None):
...
+        used as is. Otherwise, it is set to None, which means that end of the axis
+        is unbounded.
+        """
+
+        # Skip if the element doesn't have an 'extents' attribute
+        if not hasattr(element, 'extents'):
```
philippjfr, line 1928: `extents` is soft-deprecated so making anything conditional on it is weird. That said I think most elements do have it.

holoviews/plotting/bokeh/element.py
```diff
@@ -1906,6 +1914,36 @@ def initialize_plot(self, ranges=None, plot=None, plots=None, source=None):
...
+            return
+
+        def validate_bound(bound):
+            """Validate a single bound, returning None if it is not a valid number"""
+            return bound if isinstance(bound, (int, float)) and not np.isnan(bound) else None
```
philippjfr, line 1933: Would use `hv.core.util.isfinite instead here.

holoviews/plotting/bokeh/element.py
```diff
@@ -1881,6 +1885,10 @@ def initialize_plot(self, ranges=None, plot=None, plots=None, source=None):
             if self._subcoord_overlaid:
                 if style_element.label in plot.extra_y_ranges:
                     self.handles['y_range'] = plot.extra_y_ranges.pop(style_element.label)
+
+        if self.apply_hard_bounds:
+            self._apply_hard_bound(element, ranges)
```
philippjfr, line 1890: Will also have to call this on every plot update.

philippjfr (commented):
Needs a few unit tests to make sure that:

1. The bounds are correctly set for both numeric and datetime axes
2. The bounds update when the element updates

philippjfr (commented):
One more comment, it's not 100% clear to me if the hard bounds should always be drawn from the combined bounds, i.e. data range, `Dimension.range`, `Dimension.soft_range` and `{x|ylim` or whether that behavior should be configurable.

holoviews/plotting/bokeh/element.py
```diff
@@ -105,6 +105,10 @@ class ElementPlot(BokehPlot, GenericElementPlot):
     align = param.ObjectSelector(default='start', objects=['start', 'center', 'end'], doc="""
         Alignment (vertical or horizontal) of the plot in a layout.""")
 
+    apply_hard_bounds = param.Boolean(default=True, doc="""
```
philippjfr, line 108: Not 100% sure about changing the default, I think it may be sensible but I'm not 100% sure. Particularly if we default to applying the combined ranges, e.g. if the x\ylim or the Dimension.range is smaller than the data range what would even be the point of sending that additional data that you can't even see?

holoviews/plotting/bokeh/element.py
droumis: replaced with 
```python
return bound if util.isfinite(bound) else None
```
since `util.isfinite` seems to handle datetimes just fine

holoviews/plotting/bokeh/element.py
droumis: We spoke last week and said that perhaps defaulting to True might be ok, as long as it defaults to False if projection is set. So after that, I used:

```python
# If apply_hard_bound is not set but projection has been set, avoid applying hard bounds by default
# since a projection typically means that the plot is either geographic, polar, or 3d
if self.apply_hard_bounds or (self.projection is None and self.apply_hard_bounds is None):
    self._apply_hard_bounds(element, ranges)
```

But I've since determined that it's best to keep hard bounds default as false so I've simplified

```python
if self.apply_hard_bounds:
    self._apply_hard_bounds(element, ranges)
```

holoviews/plotting/bokeh/element.py
droumis: defaulting to False now

droumis:
Several issues remain.

We want it to use the largest of data + padding, xlim/ylim, dimension ranges, but it is not yet achieving that.

This also does not seem to work for y-axis with `subcoordinate_y` plots

droumis:
> We want it to use the largest of data + padding, xlim/ylim, dimension ranges, but it is not yet achieving that.

We determined that dimension ranges are a special, strict case and that a more semantically accurate handling of them includes using them as absolute hard bounds whenever specified, and when `apply_hard_bounds=True`.

So the current plan when `apply_hard_bounds=True`:
- If dimension ranges are set (e.g. via`.redim.range`), allow navigation only up to these bounds, regardless of whether the data extends beyond.
- Otherwise, use the more extreme of either the data+padding ranges or, if specified, xlim/ylim ranges.

droumis:
> This also does not seem to work for y-axis with `subcoordinate_y` plots

We determined that the it's best if the y-axis for `subcoordinate_y` plots remain unconstrained in all cases, so this bug is now a feature.

droumis:
Also need to test overlaid plots, especially when one or both have dim ranges set..

droumis:
still need to add docs but code is ready for review

holoviews/plotting/bokeh/element.py
```diff
@@ -1919,6 +1928,30 @@ def initialize_plot(self, ranges=None, plot=None, plots=None, source=None):
...
+        initial viewable range to xlim/ylim, but allow navigation up to the abs max between
+        the data range and xlim/ylim. If dim range is set (e.g. via redim.range), enforce
+        as hard bounds.
+
+        """
+
+        def validate_bound(bound):
+            return bound if util.isfinite(bound) else None
+
+        min_extent_x, min_extent_y, max_extent_x, max_extent_y = map(validate_bound, self.get_extents(element, ranges, range_type='combined', lims_as_soft_ranges=True))
```
Hoxbro, line 1943:
```suggestion
        min_extent_x, min_extent_y, max_extent_x, max_extent_y = map(
            validate_bound, self.get_extents(element, ranges, range_type='combined', lims_as_soft_ranges=True)
        )
```
Nit. The line is pretty long.

holoviews/plotting/bokeh/element.py
```diff
@@ -1919,6 +1928,30 @@ def initialize_plot(self, ranges=None, plot=None, plots=None, source=None):
...
+
+        def set_bounds(axis, min_extent, max_extent):
+            """Set the bounds for a given axis, using None if both extents are None or identical"""
+            if min_extent == max_extent:
+                self.handles[axis].bounds = None
+            else:
+                self.handles[axis].bounds = (min_extent, max_extent) if min_extent is not None or max_extent is not None else None
```
Hoxbro, line 1950:
```suggestion
                self.handles[axis].bounds = (min_extent, max_extent) if min_extent is not None else None
```

If `min_extent` is not None here `max_extent` isn't either because of `min_extent == max_extent`.

holoviews/tests/plotting/bokeh/test_elementplot.py
```diff
@@ -993,3 +994,96 @@ def test_clim_percentile(self):
...
         assert high < 1
+
+class TestApplyHardBounds(TestBokehPlot):
+    def test_apply_hard_bounds(self):
+        """Test `apply_hard_bounds` with a single element."""
+        x_values = np.linspace(10, 50, 5)
+        y_values = np.array([10, 20, 30, 40, 50])
+        curve = Curve((x_values, y_values)).opts(apply_hard_bounds=True)
+        plot = bokeh_renderer.get_plot(curve)
+        self.assertEqual(plot.handles['x_range'].bounds, (10, 50))
```
Hoxbro, line 1005:
```suggestion
        assert plot.handles['x_range'].bounds == (10, 50)
```

Because I, at some point, want to transition from unit test to pytest. Same with other `self.assertEqual`.

Hoxbro (commented):
I have left some small comments, but other than that, this looks good to me.

holoviews/plotting/bokeh/element.py
droumis: done

holoviews/plotting/bokeh/element.py
droumis: Actually, I think we can take this a step further.. if at least one of min_extent, max_extent is not None then we want to return the tuple.. so I think it can just be:

```python
if min_extent == max_extent:
    self.handles[axis].bounds = None
else:
    self.handles[axis].bounds = (min_extent, max_extent)
```

Hoxbro: If that is the case, you can make it into a one-liner

holoviews/plotting/plot.py
```diff
@@ -1423,7 +1423,7 @@ def _get_range_extents(self, element, ranges, range_type, xdim, ydim, zdim):
 
         return (x0, y0, x1, y1)
 
-    def get_extents(self, element, ranges, range_type='combined', dimension=None, xdim=None, ydim=None, zdim=None, **kwargs):
+    def get_extents(self, element, ranges, range_type='combined', dimension=None, xdim=None, ydim=None, zdim=None, lims_as_soft_ranges=False, **kwargs):
```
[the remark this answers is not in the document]
droumis: hmm are you hinting at the `get_extents` method of the `class GenericOverlayPlot`?

It doesn't appear to be necessary because even for overlay plots, GenericElementPlot's version of `get_extents` with `lims_as_soft_ranges` set to True seems to be called anyway. But honestly it's all pretty confusing and get_extents gets called numerous times so I could be missing something.

droumis:
observation, bar charts are only constrained on the continuous axis. I think that's ok for now.. it's not enough to block this PR I'm hoping

```python
data = [('one',8),('two', 10), ('three', 16), ('four', 8), ('five', 4), ('six', 1)]
bars = hv.Bars(data, hv.Dimension('Car occupants'), 'Count').opts(apply_hard_bounds=True)
bars
```

https://github.com/holoviz/holoviews/assets/6613202/b3bc8a60-7159-4053-97ec-392603066512

droumis:
observation: adding vspans to a plot that already has `apply_hard_bounds=True` doesn't screw up the hard range. ✅ 

Also, setting `apply_hard_bounds=True` directly on the vspan/hspan element is pointless as they are funtionaly unbounded.

image

holoviews/plotting/bokeh/element.py
droumis: pushed this change, let me know if you disagree

holoviews/tests/plotting/bokeh/test_elementplot.py
```diff
@@ -16,11 +16,12 @@
...
 )
 
-from holoviews.core import DynamicMap, HoloMap, NdOverlay
+from holoviews import opts
+from holoviews.core import DynamicMap, HoloMap, NdOverlay, Overlay
 from holoviews.core.util import dt_to_int
 from holoviews.element import Curve, HeatMap, Image, Labels, Scatter
 from holoviews.plotting.util import process_cmap
-from holoviews.streams import PointDraw, Stream
+from holoviews.streams import PointDraw, Stream, param
```
Hoxbro, line 24: is this just param? If so, do a direct import.

holoviews/plotting/plot.py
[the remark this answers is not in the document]
Hoxbro: I'm not hinting at anything specific; just that other classes could also need to implement `lims_as_soft_ranges` if they don't have the `super`. But I think we should wait and see if there is a demand for it, before doing anything.

Hoxbro (approved):

holoviews/plotting/bokeh/element.py
```diff
@@ -1919,6 +1928,29 @@ def initialize_plot(self, ranges=None, plot=None, plots=None, source=None):
...
+        min_extent_x, min_extent_y, max_extent_x, max_extent_y = map(
+            validate_bound, self.get_extents(element, ranges, range_type='combined', lims_as_soft_ranges=True)
+        )
+
+        def set_bounds(axis, min_extent, max_extent):
+            """Set the bounds for a given axis, using None if both extents are None or identical"""
+            self.handles[axis].bounds = None if min_extent == max_extent else (min_extent, max_extent)
+
+        set_bounds('x_range', min_extent_x, max_extent_x)
+        set_bounds('y_range', min_extent_y, max_extent_y)
```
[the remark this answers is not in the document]
droumis: I just tested a heatmap and hit an error because we can't set the bounds for a FactorRange.. so I just updated this function call to 

```python
        def set_bounds(axis, min_extent, max_extent):
            """Set the bounds for a given axis, using None if both extents are None or identical"""
            try:
                self.handles[axis].bounds = None if min_extent == max_extent else (min_extent, max_extent)
            except ValueError:
                self.handles[axis].bounds = None
```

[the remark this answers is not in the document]
droumis: at least for now, I think this is the best we can do with categorical axes.. see https://github.com/bokeh/bokeh/issues/13830

jbednar:
> bar charts are only constrained on the continuous axis

Is it possible to disable zooming and panning on the non-continous axis of a bar chart? All it ever seems to do is screw up the plot; I don't recall ever wanting to do it deliberately and it very often messes up the plot as I'm working with it.

droumis:
not declaratively, but you can do something like this if you know which axis is going to be non-continuous:

`default_tools=['ywheel_zoom', 'ypan', 'reset']`


https://github.com/holoviz/holoviews/assets/6613202/eb5d9c46-6eb0-4bf9-9071-edb19bcfa1e8

Issue #6058, opened 2024-01-05
[GOAL] Support viewing of medium, multi-channel timeseries data


droumis:
#### Is your feature request related to a problem? Please describe.
Target use cases of stacked timeseries commonly utilize a significant number of lines and samples which requires aggregation or downsampling in order to send to the browser. Currently, due to performance limitations, the standard HoloViews+Bokeh approach to this visualization with `subcoordinate_y` is only useable with a small part of the typical data size range.

#### Describe the solution you'd like
Let's shoot to make it not only possible, but smooth/performant to visualize and interact with:
**Medium Size Dataset**: too big for browser but fits in memory. For instance, a data size target of 100 stacked traces, each with 1,000 (16 bit) samples per second for 10,000 seconds. That's one billion samples and about 2 GB.

### Medium Size
- [x] Step 1: Downsample1d operation should slice once if all elements in an NdOverlay share the same underlying DataFrame (implemented in #6059) [@philippjfr @Hoxbro]
- [x] Step 2: Allow HoloViews to operate directly on a DataFrame with an index #6061 [@philippjfr @Hoxbro]
- [x] Step 3: Use tsdownsample library for downsampling if available #6059 [@philippjfr @Hoxbro]
- [ ] Step 4: [Add HoloViews' index support to hvPlot](https://github.com/holoviz/hvplot/issues/1292) [@maximlt]

## Notes for Medium Size:
- @philippjfr suggested that our implementation of LTTB downsampling would not scale sufficiently well for large datasets. This suggestion is backed up by comments in the [plotly resampler](https://github.com/predict-idlab/plotly-resampler/blob/2b79498eaa79cba891310dfcec3860936af3207b/plotly_resampler/aggregation/aggregators.py#L46): "for large datasets, [LTTB] can be much slower than other algorithms (e.g. MinMax) due to the higher cost of calculating the areas of triangles...`LTTB` doesn't scale super-well when moving to really large datasets, so when dealing with more than 1 million samples, you might consider using [`MinMaxLTTB`][aggregation.aggregators.MinMaxLTTB]"
- So either we could try an implementation of MinMaxLTTB (or even rely on the rust implementation of [tsdownsample](https://github.com/predict-idlab/tsdownsample) as plotly does - note, the author of tsdownsample [had offered to help](https://github.com/holoviz/holoviews/issues/2934#issuecomment-1367950689)!), ...
[...]

PR #6059, opened 2024-01-05
Use tsdownsample library for downsampling if available

philippjfr:
Partially resolves -> https://github.com/holoviz/holoviews/issues/6058

### DIFF
--- a/holoviews/operation/downsample.py
+++ b/holoviews/operation/downsample.py
@@ -194,7 +194,6 @@ _ALGORITHMS = {
     'minmax': _min_max,
     'minmax-lttb': _min_max_lttb,
     'm4': _m4,
-
 }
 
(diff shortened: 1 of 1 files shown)


holoviews/operation/downsample.py
```diff
@@ -183,9 +240,12 @@ def _process(self, element, key=None):
         if ys.dtype == np.bool_:
             ys = ys.astype(np.int8)
         downsample = _ALGORITHMS[self.p.algorithm]
+        kwargs = {}
         if self.p.algorithm == "lttb" and isinstance(element, Area):
```
Hoxbro, line 244:
```suggestion
        if "lttb" in self.p.algorithm and isinstance(element, Area):
```

Issue #2537, opened 2018-04-07
Multiindex support?


spott:
Are there any plans on supporting multiindexes in holoviews in the future?

jlstevens:
This issue is related to #2310. I don't believe this is currently supported but we will investigate it adding support in future.

jlstevens:
I've had a discussion with @philippjfr about handling wide data and that involved using multiindex iirc although that was mostly as an internal optimization. How exactly would you like to see multiindex supported?

KevinYanesG:
Hi! Any news on this? :)

jlstevens:
Unfortunately not though I think we should seriously consider this feature for holoviews 2.0!

hCraker:
Bump! Any news on if holoviews 2.0 is in the works and if it supports multiindex?

droumis:
I'm going to close this in light of #6061 being merged, which will likely be included in the upcoming minor HoloViews release. There will always be more work to be done, but I think general 'multiindex support' is too broad and more appropriate for a discussion thread on the [HoloViz Discourse or Discord](https://holoviz.org/community.html). If there are specific/actionable features requests or bugs, please file a new issue on GitHub.

PR #6061, opened 2024-01-06
Implement support for retaining Pandas index

philippjfr:
The Pandas support in HoloViews has always been severely hampered by the fact that we do not properly support indexes, i.e. if a user references an index column we are forced to call `.reset_index()` which is inefficient and breaks one core principle of HoloViews, which is that we simply provide thin wrappers around data. It also has significant performance implications both from a memory perspective and from a speed perspective since indexes can provide significant speedups when indexing or performing aggregations. One other major issue solved by supporting indexes is the problem of having multiple elements providing a view onto the same DataFrame, e.g. when you have an NdOverlay of curves each visualizing a different column.

Implements https://github.com/holoviz/holoviews/issues/2537

- [x] Implement all operations correctly
  - [x] `__init__`
  - [x] `validate`
  - [x] `dtype`
  - [x] `dimension_type`
  - [x] `range`
  - [x] `iloc` (check multi-index)
  - [x] `values` (check multi-index)
  - [x] `sort`
  - [x] `reindex`
  - [x] `sample`
- [x] Optimized operations
  - [x] `select`
  - [x] `groupby`
- [x] Check operations that use `dataset.data` directly do not make (now invalid) assumptions about indexes
- [x] Add tests

### DIFF
--- a/holoviews/core/data/pandas.py
+++ b/holoviews/core/data/pandas.py
@@ -325,7 +325,7 @@ class PandasInterface(Interface, PandasAPI):
     @classmethod
     def reindex(cls, dataset, kdims=None, vdims=None):
         data = dataset.data
-        if hasattr(data, 'index') and isinstance(data.index, pd.MultiIndex):
+        if isinstance(data.index, pd.MultiIndex):
             kdims = [kdims] if isinstance(kdims, (str, Dimension)) else kdims
             data = data.reset_index().set_index(list(map(str, kdims)), drop=True)
         return data
@@ -363,7 +363,7 @@ class PandasInterface(Interface, PandasAPI):
                 idx or "index": v
                 for idx in df.index.names
                 if isinstance((v := selection.get(idx)), slice)
-            }
+            } if not isinstance(df.index, pd.MultiIndex) else {}
             if selection.keys() - indexes.keys():
                 selection_mask = cls.select_mask(dataset, selection)
             else:
@@ -382,13 +382,13 @@ class PandasInterface(Interface, PandasAPI):
 
     @classmethod
     def values(
-            cls,
-            dataset,
-            dim,
-            expanded=True,
-            flat=True,
-            compute=True,
-            keep_index=False,
+        cls,
+        dataset,
+        dim,
+        expanded=True,
+        flat=True,
+        compute=True,
+        keep_index=False,
     ):
         dim = dataset.get_dimension(dim, strict=True)
         isindex = cls.isindex(dataset, dim)
--- a/holoviews/core/data/pandas.py
+++ b/holoviews/core/data/pandas.py
@@ -33,6 +33,8 @@ class PandasInterface(Interface, PandasAPI):
     @classmethod
     def dimension_type(cls, dataset, dim):
         name = dataset.get_dimension(dim, strict=True).name
+        if name in cls.indexes(dataset.data):
+            return cls.index_values(dataset, dim).dtype.type
         idx = list(dataset.data.columns).index(name)
         return dataset.data.dtypes.iloc[idx].type
 
@@ -46,9 +48,7 @@ class PandasInterface(Interface, PandasAPI):
             data = data.to_frame(name=name)
         if util.is_dataframe(data):
             ncols = len(data.columns)
-            index_names = data.index.names if isinstance(data, pd.DataFrame) else [data.index.name]
-            if index_names == [None]:
-                index_names = ['index']
+            index_names = cls.indexes(data)
             if eltype._auto_indexable_1d and ncols == 1 and kdims is None:
                 kdims = list(index_names)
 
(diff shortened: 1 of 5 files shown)


droumis:
may partially resolve #6058

holoviews/core/data/pandas.py
```diff
@@ -75,16 +71,9 @@ def init(cls, eltype, data, kdims, vdims):
                 )
 
             # Handle reset of index if kdims reference index by name
```
philippjfr, line 73: Should probably remove this comment.

holoviews/core/data/pandas.py
```diff
@@ -75,16 +71,9 @@ def init(cls, eltype, data, kdims, vdims):
...
-                    continue
-                if any(kd == ('index' if name is None else name)
-                       for name in index_names):
-                    data = data.reset_index()
-                    break
-
-            if kdims:
+            if len(kdims) == len(index_names) and {dimension_name(kd) for kd in kdims} == set(index_names):
+                pass
+            elif kdims:
```
philippjfr, line 76: Maybe cleaner, maybe not

```suggestion
            if kdims and not (len(kdims) == len(index_names) and {dimension_name(kd) for kd in kdims} == set(index_names)):
```

holoviews/core/data/pandas.py
```diff
@@ -332,12 +359,21 @@ def sort(cls, dataset, by=None, reverse=False):
...
+            indexes = {
+                idx or "index": v
+                for idx in df.index.names
+                if isinstance((v := selection.get(idx)), slice)
+            } if not isinstance(df.index, pd.MultiIndex) else {}
+            if selection.keys() - indexes.keys():
+                selection_mask = cls.select_mask(dataset, selection)
+            else:
+                for v in indexes.values():
+                    df = df[v]
```
philippjfr, line 371: Wouldn't it be better to build up a full `indexes` tuple and then use `df.loc[indexes, :]`? Also I think you may have to transform two-tuples into slices here, since `.select(index=(0, 3))` should be treated like `.select(index=slice(0, 3))` (maybe that's handled elsewhere though).

holoviews/core/data/pandas.py
```diff
@@ -346,20 +382,24 @@ def select(cls, dataset, selection_mask=None, **selection):
...
+        expanded=True,
+        flat=True,
+        compute=True,
+        keep_index=False,
     ):
         dim = dataset.get_dimension(dim, strict=True)
-        data = dataset.data[dim.name]
+        isindex = cls.isindex(dataset, dim)
+        if isindex:
+            data = cls.index_values(dataset, dim)
```
philippjfr, line 396: I genuinely don't know but does `index.get_level_values(dimension.name)` always return an array of the same length as the DataFrame itself?

philippjfr (commented):
Looks good, left a few suggestions and questions for you.

holoviews/core/data/pandas.py
Hoxbro: > Wouldn't it be better to build up a full indexes tuple and then use df.loc[indexes, :]

This fast-path only works for a single Index. Without transforming it into an array, I couldn't figure out an easy way to get it to work with `pd.MultiIndex`. So, even if it is a for-loop, it is only a single iteration. If I recall correctly, I did some tests and saw better speeds with `df[...]` over `df.loc[...]`. 

> Also I think you may have to transform two-tuples into slices here

It is not. It could also be for a tuple size of 3, right?

holoviews/core/data/pandas.py
Hoxbro: I think so, the way thinks are structured behind the scenes looks like a list of tuples for MultiIndex.

image

``` python
import pandas as pd

frame = pd.DataFrame({"number": [1, 1, 2, 2], "color": ["red", "blue", "red", "blue"]})
index = pd.MultiIndex.from_frame(frame, names=frame.columns)
df = pd.DataFrame(range(4), index=index, columns=["values"])
df.index.get_level_values("number")

File: holoviews/core/data/pandas.py (hunk 3 above)
Comment (Hoxbro) [reply holoviews/core/data/pandas.py:369]: I have pushed for the tuples.

Comment (philippjfr) [reply holoviews/core/data/pandas.py:369]: >Without transforming it into an array, I couldn't figure out an easy way to get it to work with pd.MultiIndex

Not totally following this bit. Want me to take a quick look at this?

File: holoviews/core/data/pandas.py [hunk 5]
```diff
@@ -327,17 +352,73 @@ def sort(cls, dataset, by=None, reverse=False):
...
+            sel = selection[idx]
+            if isinstance(sel, tuple) and len(sel) < 4:
+                sel = slice(*sel)
+            elif not isinstance(sel, (list, slice)):
+                sel = [sel]
+            if isinstance(sel, slice) and nindex > 1 and not sorted_index and level>depth:
+                # If the index is not monotonic we cannot slice
+                return {}
+            index_sel[idx] = sel
+        print(index_sel)
```
philippjfr, line 398:
```suggestion
```

PR #6063, opened 2024-01-08
Ensure partial methods can be used as dmap callables

### LINKED ISSUE #6062: DynamicMap cannot handle partial bound method callable
HoloViews 1.18.1
Bokeh 3.3.1

If you provide a bound method callable with `functools.partial` to a DynamicMap it will crash. It works fine with bound functions and bound static methods.

```python
import holoviews as hv
import numpy as np
import functools


class MyClass:

    def __init__(self):
        pass

    def make_dmap(self):
        dmap = hv.DynamicMap(functools.partial(self.make_plot, freq=1), kdims=['phase'])
        return dmap.redim.range(phase=(0.5, 1))

    def make_dmap_static(self):
        dmap = hv.DynamicMap(functools.partial(self.make_plot_static, freq=1), kdims=['phase'])
        return dmap.redim.range(phase=(0.5, 1))

    def make_plot(self, phase, freq):

    @staticmethod
    def make_plot_static(phase, freq):


MyClass().make_dmap_static()  # will work
MyClass().make_dmap()  # will not work
```

```
 File "/home/mathurin/PycharmProjects/holoviews/toast.py", line 15, in make_dmap
    dmap = hv.DynamicMap(functools.partial(self.make_plot, freq=1), kdims=['phase'])
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mathurin/miniconda3/envs/holoviews/lib/python3.11/site-packages/holoviews/core/spaces.py", line 807, in __init__
    self._posarg_keys = util.validate_dynamic_argspec(
                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/mathurin/miniconda3/envs/holoviews/lib/python3.11/site-packages/holoviews/core/util.py", line 499, in validate_dynamic_argspec
    raise KeyError(f'Callable {name!r} accepts more positional arguments than '
KeyError: "Callable 'functools.partial(<bound method MyClass.make_plot of <bokeh_app_45540ef9abcd44a88a8ac4cb605b7b56.MyClass object at 0x7f8a5e413310>>, freq=1)' accepts more positional arguments than there are kdims and stream parameters"
```

In the `argspec` function in holoviews/core/util.py [here](https://github.com/holoviz/holoviews/blob/main/holoviews/core/util.py#L427C8-L427C8), we could simply check whether `calllable_obj.func` for partials is a method and if so filter out the first argument:

```python
...
    elif isinstance(callable_obj, partial): # partials
        arglen = len(callable_obj.args)
        spec = inspect.getfullargspec(callable_obj.func)
        args = [arg for arg in spec.args[arglen:] if arg not in callable_obj.keywords]
        if inspect.ismethod(callable_obj.func):
            args = args[1:]
...
```

TheoMathurin:
Fixes #6062 

I guess tests should be added in `tests/core/test_callable.py`.


TheoMathurin:
I've added some tests in the `test_callable` module both for full and partial methods. For this purpose an instance method is added to `CallableClass` with arguments so that its `argspec` can be tested.

More could be done as only some test classes have been covered. Note that all callable types are not systematically tested and I don't think this PR is the place to do it.

holoviews/core/util.py
```diff
@@ -445,6 +445,10 @@ def argspec(callable_obj):
         arglen = len(callable_obj.args)
         spec = inspect.getfullargspec(callable_obj.func)
```
Hoxbro, line 446: I think we should unpack the partial before this line and maybe at the start of the elif statement.

TheoMathurin: Maybe but I'm not really sure.

In fact the following also works and yields the same `args` (both for simple partial method and nested partial methods):
```python
elif isinstance(callable_obj, partial): # partials
    arglen = len(callable_obj.args)
    while isinstance(callable_obj.func, partial): # Working with the inner-most partial from here
        callable_obj = callable_obj.func
    spec = inspect.getfullargspec(callable_obj.func)
    args = [arg for arg in spec.args[arglen:] if arg not in callable_obj.keywords]
    if inspect.ismethod(callable_obj.func):
        args = args[1:]
```

TheoMathurin: Well in fact we don't even have to loop because the outer partial's `func` object is already the original method. It looks like nesting partials is akin to merging the keywords supplied at the different levels, so the result is not really nested.

Therefore I would argue the change required is really minimal and only involves checking whether `callable_obj.func` is a method or not.

Hoxbro: Great to hear.

Hoxbro: And don't worry about unrelated failing tests, I'm working on them in https://github.com/holoviz/holoviews/pull/6078

holoviews/tests/core/test_callable.py
```diff
@@ -66,6 +69,9 @@ def test_callable_class_name(self):
     def test_callable_class_call_method(self):
         self.assertEqual(Callable(CallableClass().__call__).name, 'CallableClass')
 
+    def test_callable_instance_method(self):
+        self.assertEqual(Callable(CallableClass().someinstancemethod).name, 'CallableClass.someinstancemethod')
```
Hoxbro, line 73: I know the other tests use `self.assertEqual`, but I want to transition away from using the unittest framework to pytest. So can you change all of your added tests to `assert a == b`?

Hoxbro:
Thank you for this PR, great work.

PR #6074, opened 2024-01-17
Update to work with Pandas 2.2 release candidate

Hoxbro:
Also removes a warning emitted from xarray 2024.1

Still, one warning causing problem: `Passing a SingleBlockManager to GeoSeries is deprecated and will raise in a future version. Use public APIs instead.`. This is likely a problem that needs to be fixed in spatial pandas. Issue here: https://github.com/holoviz/spatialpandas/issues/137

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -142,6 +142,7 @@ jobs:
         run: |
           conda activate test-environment
           bokeh sampledata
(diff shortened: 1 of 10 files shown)


Hoxbro:
I will merge this if tests pass, as this should work with the currently released pandas.

PR #6075, opened 2024-01-18
Optimize downsample1d when data is shared between layers

philippjfr:
Another optimization for `downsample1d` when multiple layers share the same data. Since the majority of time is spent slicing the data we can apply an optimization when all elements in an (Nd)Overlay that is being downsampled share the same underlying data and are indexed along the same key dimension(s). 

In this case we simply get the first element slice it with the current x_range and then pass it down to each separate `self._process` call.

Here's the test case I was using:

```python
import holoviews as hv
import numpy as np
import pandas as pd

hv.extension('bokeh')

N = 1_000_000
index = pd._testing.makeDateIndex(k=N, freq='h', name='date')
df = pd.DataFrame({chr(65+i): np.random.randn(N).cumsum() for i in range(25)}, index=index).reset_index()

overlay = hv.NdOverlay({col: hv.Curve(df, 'date', col) for col in df.columns[1:]})

d = downsample1d(overlay).opts(responsive=True, height=300)
d
```

bokeh_plot - 2024-01-18T155254 315

```python
i = 2000
size_stream = d.streams[0]

%%timeit -r 10 -n 1
global i
i += 1
size_stream.event(width=i)
```

Results before:

> 1.41 s ± 60.8 ms per loop (mean ± std. dev. of 10 runs, 1 loop each)

Results after:

> 319 ms ± 40.8 ms per loop (mean ± std. dev. of 10 runs, 1 loop each)

Addresses part of https://github.com/holoviz/holoviews/issues/6058

Ping @droumis and @Hoxbro

### DIFF
--- a/holoviews/tests/operation/test_downsample.py
+++ b/holoviews/tests/operation/test_downsample.py
@@ -1,4 +1,5 @@
 import numpy as np
+import pandas as pd
 import pytest
 
 import holoviews as hv
@@ -10,7 +11,8 @@ except ImportError:
     tsdownsample = None
 
 algorithms = _ALGORITHMS.copy()
-algorithms.pop('viewport', None)  # viewport return slice(len(data)) no matter the width
+algorithms.pop("viewport", None)  # viewport return slice(len(data)) no matter the width
+
 
 @pytest.mark.parametrize("plottype", ["overlay", "ndoverlay"])
 def test_downsample1d_multi(plottype):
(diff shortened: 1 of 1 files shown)


Hoxbro:
@philippjfr, this should be ready for review; I can't request you review it as you are the original author of the issue.

holoviews/tests/operation/test_downsample.py
```diff
@@ -28,6 +30,46 @@ def test_downsample1d_multi(plottype):
             assert value.size == downsample1d.width
 
 
+def test_downsample1d_shared_data():
+    runs = [0]
+
+    class mocksample(downsample1d):
```
philippjfr, line 36: Probably would have used `unittest.mock.patch` on the Interface but this is fine.

PR #6083, opened 2024-01-25
Add test for downsample algorithms

Hoxbro:
Add simple tests to check if we can run the downsample algorithms.

Unimport is inspired by this: https://mastodon.social/@sscherfke/111116160705738755

It seems like tsdownsample is not available on Windows on conda-forge.

### DIFF
--- a/holoviews/tests/conftest.py
+++ b/holoviews/tests/conftest.py
@@ -1,4 +1,6 @@
 import contextlib
+import sys
+from collections.abc import Callable
 
 import pytest
 from panel.tests.conftest import (  # noqa
@@ -80,3 +82,17 @@ def plotly_backend():
     hv.Store.current_backend = 'plotly'
     yield
     hv.Store.current_backend = prev_backend
+
+@pytest.fixture
+def unimport(monkeypatch: pytest.MonkeyPatch) -> Callable[[str], None]:
+    """
+    Return a function for unimporting modules and preventing reimport.
+    """
+
+    def unimport_module(modname: str) -> None:
+        # Remove if already imported
+        monkeypatch.delitem(sys.modules, modname, raising=False)
+        # Prevent import:
+        monkeypatch.setattr(sys, "path", [])
+
+    return unimport_module
--- a/holoviews/tests/conftest.py
+++ b/holoviews/tests/conftest.py
@@ -87,6 +87,8 @@ def plotly_backend():
 def unimport(monkeypatch: pytest.MonkeyPatch) -> Callable[[str], None]:
     """
     Return a function for unimporting modules and preventing reimport.
+
+    This will block any new modules from being imported.
     """
 
     def unimport_module(modname: str) -> None:
--- a/holoviews/tests/operation/test_downsample.py
+++ b/holoviews/tests/operation/test_downsample.py
@@ -1,7 +1,13 @@
+import numpy as np
 import pytest
 
 import holoviews as hv
-from holoviews.operation.downsample import downsample1d
+from holoviews.operation.downsample import _ALGORITHMS, downsample1d
+
+try:
+    import tsdownsample
+except ImportError:
+    tsdownsample = None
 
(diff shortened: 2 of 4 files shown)


holoviews/tests/operation/test_downsample.py
```diff
@@ -18,3 +24,27 @@ def test_downsample1d_multi(plottype):
     for n in figure_values:
         for value in n.data.values():
             assert value.size == downsample1d.width
+
+
+@pytest.mark.parametrize("algorithm", _ALGORITHMS.values(), ids=_ALGORITHMS)
```
maximlt, line 29: I think the way these two tests are written leads to repeating the same tests (non-`tsdownsample` tests run twice) and skipping tests when `tsdownsample` is unimported while it should be a success (in `test_downsample_algorithm`)?

This is maybe more explicit?

```py
TSD_ALGOS = {k: v for k, v in _ALGORITHMS.items() if k in ['minmax', 'minmax-lttb', 'm4']}

@pytest.mark.parametrize("algorithm", TSD_ALGOS.values(), ids=TSD_ALGOS)
def test_downsample_algorithm(algorithm, unimport):
    unimport("tsdownsample")
    x = np.arange(1000)
    y = np.random.rand(1000)
    width = 20
    with pytest.raises(NotImplementedError):
        algorithm(x, y, width)
```

holoviews/tests/conftest.py
```diff
@@ -80,3 +82,17 @@ def plotly_backend():
...
+def unimport(monkeypatch: pytest.MonkeyPatch) -> Callable[[str], None]:
+    """
+    Return a function for unimporting modules and preventing reimport.
+    """
+
+    def unimport_module(modname: str) -> None:
+        # Remove if already imported
+        monkeypatch.delitem(sys.modules, modname, raising=False)
+        # Prevent import:
+        monkeypatch.setattr(sys, "path", [])
```
maximlt, line 96: That's kind of crazy, but if it works it works :) Could you add a comment in the docstring saying that it prevents importing anything really, not just `modname`.

maximlt (commented):
Thanks for spending time writing these tests :) I left a few comments.

setup.py
```diff
@@ -59,6 +59,9 @@
     'dash >=1.16',
 ]
 
+if os.name != "nt":
```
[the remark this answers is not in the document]
Hoxbro: It is not available on Conda-forge leading to failing tests on Windows. 

tsdownsample seems to be supported on Windows on PyPi.

holoviews/tests/conftest.py
Hoxbro: Will do.

setup.py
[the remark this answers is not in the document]
maximlt: Ok cool thanks, maybe a comment can precise that.

holoviews/tests/operation/test_downsample.py
Hoxbro: There is an overlap with lttb which works with and without tsdownsample. 

I originally had it passing, but liked I got feedback that the test was skipped for tsdownsample only functions. 

I think there is a lot of ways to do this. I preferred running some test twice than using logic as it could get out of sync. If you much more prefer it your way I will update the tests.

maximlt: Ok I see. The only problem I have then with the first test is that I can comment out `unimport("tsdownsample")` and will get no error with and without `tsdownsample` installed, it'll either skip or succeed.

Hoxbro: I agree with you that it is not ideal. I think it is okay, as your scenario changes the behavior of the unit test itself. 

I will gladly listen if you have a proposal on how to test it.

maximlt: Ok yes that's fine by me.

PR #6103, opened 2024-02-06
enh: Improve error message for non-matching types with aspect set to equal

hoxbro:
Fixes https://github.com/holoviz/holoviews/issues/6089

This can create the Matplotlib problem without `rasterize`.  `rasterize` will have an `Image` aspect set to 'equal', whereas `Quadmesh` have an aspect set to `square`. 

I have chosen to raise a TypeError instead of adding custom logic as it does not really make sense to take the `equal` of an axis with numbers and an axis with timedelta.   An alternative could be to set the aspect to 1. 

``` python
import holoviews as hv
import numpy as np
import pandas as pd

hv.extension("matplotlib")

X = pd.date_range(start="1/1/2018", end="1/08/2018", periods=100)
Y = np.linspace(1, 100, 100)
Z = np.random.randn(100, 100)

hv.QuadMesh((X, Y, Z)).opts(aspect='equal')
```

### TODO
- [x] Improve plotly
- [x] Add tests


holoviews/plotting/plotly/element.py
```diff
@@ -318,6 +318,19 @@ def get_aspect(self, xspan, yspan):
         """Computes the aspect ratio of the plot
 
         """
+        if self.aspect == 'equal':
```
hoxbro: We can combine the two if statements.

thuydotm: Done!

hoxbro:
Thank you

Unfortunate that the one starting the PR couldn't finish it...

thuydotm (approved):

Issue #6119, opened 2024-02-19
Hover tooltip minimal target size


adriengcql:
When the markers in a plot are small, it is not always easy to aim at a point to get the tooltip.

![tooltip1](https://github.com/holoviz/holoviews/assets/26626141/4d8ce8ea-a43d-4f33-af70-16ffae8a839a)
![tooltip2](https://github.com/holoviz/holoviews/assets/26626141/4cc21a18-9138-44b9-b173-cfeee3cb69af)

It would be great if the area around the marker also triggered the tooltip display.

A workaround is to increase the marker size, but this tends to make the plot less readable.

Hoxbro:
I don't think an API in Bokeh makes this possible. @mattpap can you confirm this? And if is not possible, is there an open feature request around this? The only one I could find was https://github.com/bokeh/bokeh/issues/7338.

mattpap:
This https://github.com/bokeh/bokeh/issues/6806 is the relevant issue. Currently every glyph/annotation/etc. manages hit threshold on its own, if at all. Luckily marker/scatter glyph supports `hit_dilation` property (default 1px), so this should help with this particular issue.

PR #6121, opened 2024-02-20
Expose hit_dilation in PointPlot

hoxbro:
Fixes https://github.com/holoviz/holoviews/issues/6119

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -6,107 +6,140 @@ on:
       - "v[0-9]+.[0-9]+.[0-9]+a[0-9]+"
       - "v[0-9]+.[0-9]+.[0-9]+b[0-9]+"
       - "v[0-9]+.[0-9]+.[0-9]+rc[0-9]+"
-  # Dry-run only
   workflow_dispatch:
   schedule:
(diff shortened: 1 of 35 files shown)


holoviews/plotting/bokeh/chart.py
```diff
@@ -35,6 +35,9 @@ class PointPlot(LegendPlot, ColorbarPlot):
         The current selection as a list of integers corresponding
         to the selected items.""")
 
+    hit_dilation = param.Integer(default=1, bounds=(1, None), doc="""
```
philippjfr, line 38: Should probably just be added to `style_opts` rather than making it a plot parameter.

Hoxbro: I did not know that it would not be applied in style_opts if it was also a parameter. I have added it as style_opts now.

philippjfr: Sorry if that wasn't clear.

Hoxbro: It was clear. Before your comment, I had tried it as a parameter and also added it as `style_opts`, which did not work. After your comment things clicked.

PR #6122, opened 2024-02-20
Add a zoom tool per subcoordinate_y group

maximlt:
closes https://github.com/holoviz/holoviews/issues/5901

When users assign a `group` to an Element with `subcoordinate_y` enabled, a dedicated `wheel_zoom` is created per group. The order of the zoom tools matches how the groups are displayed, i.e. the top tool controls the group displayed at the top of the chart. Each tool has a custom tooltip that contains the group name.

```python
import numpy as np
import panel as pn
import holoviews as hv
hv.extension('bokeh')

x = np.linspace(0, 10*np.pi)
curves = []
j = 0.5
for group in ['A', 'B', 'C']:
    for i in range(2):
        yvals = j * np.sin(x)
        curve = hv.Curve((x + i*np.pi/2, yvals), label=f'{group}{i}', group=group).opts(
            subcoordinate_y=True
        )
        curves.append(curve)
        j += 1

hv.Overlay(curves).opts(show_legend=False, width=400, height=400)
```
Kapture 2024-02-20 at 15 18 03

Users can add extra `zoom_in` and `zoom_out` tools. The UX isn't great but we agreed not to spend too much time trying to improve it (I've seen worse!).

```python
for group in ['A', 'B', 'C']:
    for i in range(2):
        yvals = j * np.sin(x)
        curve = hv.Curve((x + i*np.pi/2, yvals), label=f'{group}{i}', group=group).opts(
            subcoordinate_y=True, tools=['zoom_in', 'zoom_out']
        )
        curves.append(curve)
        j += 1

hv.Overlay(curves).opts(show_legend=False, width=400, height=800)
```
Kapture 2024-02-20 at 15 21 29

When overlaid with other non subcoordinate_y elements, the default zoom tool is displayed.
```python
for group in ['A', 'B', 'C']:
    for i in range(2):
        yvals = j * np.sin(x)
        curve = hv.Curve((x + i*np.pi/2, yvals), label=f'{group}{i}', group=group).opts(
            subcoordinate_y=True,
        )
        curves.append(curve)
        j += 1

(hv.Overlay(curves) * hv.Overlay([hv.VSpan(i, i + 2) for i in range(2, 20, 4)])).opts(
    show_legend=False, width=400, height=500
)
```
Kapture 2024-02-20 at 15 25 40

Implementation-wise, I opted for adding a post-processing step, instead of having to changing the structure of the zoom handle directory (accounting for groups) as it required changes in various places that are hard to follow. In the end, the implementation isn't so complex.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -2881,6 +2881,18 @@ class OverlayPlot(GenericOverlayPlot, LegendPlot):
         """
         # First, just process and validate the groups and their content.
         groups = defaultdict(list)
+
+        # If there are groups AND there are subcoordinate_y elements without a group.
+        if any(el.group != type(el).__name__ for el in overlay) and any(
+            el.opts.get('plot').kwargs.get('subcoordinate_y', False)
+            and el.group == type(el).__name__
+            for el in overlay
+        ):
+            raise ValueError(
+                'The subcoordinate_y overlay contains elements with a defined group, each '
+                'subcoordinate_y element in the overlay must have a defined group.'
+            )
+
         for el in overlay:
             # group is the Element type per default (e.g. Curve, Spike).
             if el.group == type(el).__name__:
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -2846,7 +2846,6 @@ class OverlayPlot(GenericOverlayPlot, LegendPlot):
         self.handles['hover_tools'] = hover_tools
         return init_tools
 
-
     def _merge_tools(self, subplot):
         """
         Merges tools on the overlay with those on the subplots.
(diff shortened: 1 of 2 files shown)


droumis:
I made the normalization per group a [separate issue](https://github.com/holoviz/holoviews/issues/6123). I think it could probably go into your postprocessing step.

droumis:
If not all the curves are assigned a group, then none of the group-wise zooming takes effect. We should either issue a warning/error to assign all curves to a group, or just group any that aren't already assigned into a new group. Probably the former is good enough and the simpler approach.

For example, in the code below, not including a group arg for the B group breaks all the intended grouping behavior.

```python
import numpy as np
import panel as pn
import holoviews as hv
hv.extension('bokeh')

x = np.linspace(0, 10*np.pi)
curves = {}
j = 0.5
for i, group in enumerate(['A', 'B', 'C']):
    for i in range(2):     
        yvals = j * np.sin(x)
        label = f'{group}{i}'
        if group == "B":
            curve = hv.Curve((x + i*np.pi/2, yvals), label=label, group=group).opts(
                subcoordinate_y=True, tools=['zoom_in', 'zoom_out', 'hover']
            )
        else:
            curve = hv.Curve((x + i*np.pi/2, yvals), label=label).opts(
                subcoordinate_y=True, tools=['zoom_in', 'zoom_out', 'hover']
            )
        curves[(group, label)] = curve
        j += 1

pn.Row(hv.Overlay(curves, kdims=['group', 'label']).opts(show_legend=False, width=600, height=500)).servable()

```

image

philippjfr (approved):
Looks good to me!

Issue #6123, opened 2024-02-20
Subcoordinate-y normalization per group


droumis:
If groups are specified, normalize the y-range of curve-like elements per group, by default.

PR #6124, opened 2024-02-20
Add operation for group-wise normalisation

maximlt:
resolves https://github.com/holoviz/holoviews/issues/6123

Very much WIP and I've never written any operation! But it works:

```python
import numpy as np
import panel as pn
import holoviews as hv

from holoviews.operation.normalization import normalize_group

hv.extension('bokeh')

x = np.linspace(0, 10*np.pi)
curves = []
j = 0.5
for group in ['A', 'B', 'C']:
    for i in range(2):
        yvals = j * np.sin(x)
        curve = hv.Curve((x + i*np.pi/2, yvals), label=f'{group}{i}', group=group).opts(
            subcoordinate_y=True,
        )
        curves.append(curve)
        j += 1

os = hv.Overlay(curves).opts(show_legend=False, width=400, height=400)
normalize_group(os)
```

image

@droumis let's chat more about the CZI use case we're trying to solve, with a much much larger dataset that comes with more constraints.

### DIFF
--- a/holoviews/operation/normalization.py
+++ b/holoviews/operation/normalization.py
@@ -190,6 +190,8 @@ class subcoordinate_group_ranges(Operation):
     """
 
     def _process(self, overlay, key=None):
+        if not getattr(overlay, 'subcoordinate_y', False):
+            return overlay
         vmins = defaultdict(list)
         vmaxs = defaultdict(list)
         include_chart = False
--- a/holoviews/operation/normalization.py
+++ b/holoviews/operation/normalization.py
@@ -206,6 +206,7 @@ class subcoordinate_group_ranges(Operation):
         new = []
         for el in overlay:
             if not isinstance(el, Chart):
+                new.append(el)
                 continue
             y_dimension = el.vdims[0]
             y_dimension = y_dimension.clone(range=minmax[el.group])
--- a/holoviews/operation/normalization.py
+++ b/holoviews/operation/normalization.py
@@ -20,7 +20,7 @@ import param
 from ..core import Overlay
 from ..core.operation import Operation
 from ..core.util import match_spec
-from ..element import Raster
+from ..element import Chart, Raster
 
 
 class Normalization(Operation):
@@ -188,6 +188,8 @@ class normalize_group(Operation):
         vmins = defaultdict(list)
         vmaxs = defaultdict(list)
         for el in overlay:
+            if not isinstance(el, Chart):
+                continue
             vmin, vmax = el.range(1)
             vmins[el.group].append(vmin)
             vmaxs[el.group].append(vmax)
@@ -198,6 +200,8 @@ class normalize_group(Operation):
         }
         new = []
         for el in overlay:
+            if not isinstance(el, Chart):
+                continue
             y_dimension = el.vdims[0]
             y_dimension = y_dimension.clone(range=minmax[el.group])
             new.append(el.redim(**{y_dimension.name: y_dimension}))
--- a/holoviews/operation/normalization.py
+++ b/holoviews/operation/normalization.py
@@ -192,12 +192,17 @@ class subcoordinate_group_ranges(Operation):
     def _process(self, overlay, key=None):
         vmins = defaultdict(list)
         vmaxs = defaultdict(list)
+        include_chart = False
         for el in overlay:
             if not isinstance(el, Chart):
                 continue
             vmin, vmax = el.range(1)
             vmins[el.group].append(vmin)
             vmaxs[el.group].append(vmax)
+            include_chart = True
+
+        if not include_chart:
+            return overlay
 
         minmax = {
             group: (np.min(vmins[group]), np.max(vmaxs[group]))
--- a/holoviews/operation/normalization.py
+++ b/holoviews/operation/normalization.py
@@ -200,6 +200,5 @@ class normalize_group(Operation):
         for el in overlay:
             y_dimension = el.vdims[0]
             y_dimension = y_dimension.clone(range=minmax[el.group])
-            el = el.redim(**{y_dimension.name: y_dimension})
-            new.append(el)
+            new.append(el.redim(**{y_dimension.name: y_dimension}))
         return overlay.clone(data=new)
--- a/holoviews/operation/normalization.py
+++ b/holoviews/operation/normalization.py
@@ -21,7 +21,6 @@ from ..core import Overlay
 from ..core.operation import Operation
 from ..core.util import match_spec
 from ..element import Raster
-from ..util.transform import dim
 
 
 class Normalization(Operation):
@@ -199,7 +198,8 @@ class normalize_group(Operation):
         }
         new = []
         for el in overlay:
-            dim_name = el.dimensions()[1].name
-            expr = (dim(dim_name) - minmax[el.group][0]) / (minmax[el.group][1] - minmax[el.group][0])
-            new.append(el.transform(**{dim_name: expr}))
+            y_dimension = el.vdims[0]
+            y_dimension = y_dimension.clone(range=minmax[el.group])
+            el = el.redim(**{y_dimension.name: y_dimension})
+            new.append(el)
         return overlay.clone(data=new)
--- a/holoviews/operation/normalization.py
+++ b/holoviews/operation/normalization.py
@@ -12,13 +12,16 @@ operations per element type. Unlike display normalization, data
 normalizations result in transformations to the stored data within
 each element.
 """
+from collections import defaultdict
 
+import numpy as np
 import param
 
 from ..core import Overlay
 from ..core.operation import Operation
 from ..core.util import match_spec
 from ..element import Raster
+from ..util.transform import dim
 
 
 class Normalization(Operation):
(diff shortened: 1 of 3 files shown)


maximlt:
Take 2 on this. This time the operation doesn't update the data but just records the min-max per group in the `range` tuple parameter of the y dimension. The aim is to rely on a concept HoloViews already has that defining this range is respected by when an element is plotted.

image

Setting the `range` tuple on the elements part of a subcoordinate_y overlay doesn't work out of the box. I made a simple (hacky for the moment) change to `element.py` so that in this context the `range` value takes precedence over the other approaches.

The code is the same as in the original post, with the same output, except this time the data is unchanged.

image

holoviews/operation/normalization.py
```diff
@@ -175,3 +177,29 @@ def _normalize_raster(self, raster, key):
...
+
+        minmax = {
+            group: (np.min(vmins[group]), np.max(vmaxs[group]))
+            for group in vmins
+        }
+        new = []
+        for el in overlay:
+            y_dimension = el.vdims[0]
+            y_dimension = y_dimension.clone(range=minmax[el.group])
+            el = el.redim(**{y_dimension.name: y_dimension})
```
jlstevens, line 203: No need to bind `el` here, an be put straight into the append.

holoviews/operation/normalization.py
```diff
@@ -175,3 +177,29 @@ def _normalize_raster(self, raster, key):
...
+            vmins[el.group].append(vmin)
+            vmaxs[el.group].append(vmax)
+
+        minmax = {
+            group: (np.min(vmins[group]), np.max(vmaxs[group]))
+            for group in vmins
+        }
+        new = []
+        for el in overlay:
+            y_dimension = el.vdims[0]
```
jlstevens, line 201: Should probably check if the element is an instance of `Chart` first...

holoviews/plotting/bokeh/element.py
```diff
@@ -486,10 +486,15 @@ def _axis_props(self, plots, subplots, element, ranges, pos, *, dim=None,
 
         range_el = el if self.batched and not isinstance(self, OverlayPlot) else element
 
+        if pos == 1 and 'subcoordinate_y' in range_tags_extras and dim and dim.range != (None, None):
+            dims = [dim]
+            v0, v1 = dim.range
+            axis_label = str(dim)
+            specs = ((dim.name, dim.label, dim.unit),)
```
jlstevens, line 493: Technically `ylim` should be supported as well...I'll open an issue about this.

holoviews/operation/normalization.py
```diff
@@ -175,3 +177,29 @@ def _normalize_raster(self, raster, key):
                 if range:
                     norm_raster.data[:,:,depth] /= range
         return norm_raster
+
+
+class normalize_group(Operation):
+    """
+    Group-wise min-max normalisation.
```
jlstevens, line 184: Docstring should say more including that the range on the dimensions are being computed...

holoviews/operation/normalization.py
```diff
@@ -175,3 +177,29 @@ def _normalize_raster(self, raster, key):
                 if range:
                     norm_raster.data[:,:,depth] /= range
         return norm_raster
+
+
+class normalize_group(Operation):
```
jlstevens, line 182: I think this should say something about subcoordinates as the output will only work in that case...horrible name but semantically it is something like `subcoordinate_group_ranges`

holoviews/plotting/bokeh/element.py
jlstevens: An issue about this bit of code once this PR is merged ;-p

philippjfr:
Just to clarify, this operation is meant to be invoked explicitly by a user or could we use compositors to automatically apply this if `subcoordinate_y` is set?

holoviews/operation/normalization.py
```diff
@@ -175,3 +177,38 @@ def _normalize_raster(self, raster, key):
...
+        }
+        new = []
+        for el in overlay:
+            if not isinstance(el, Chart):
+                new.append(el)
+                continue
+            y_dimension = el.vdims[0]
+            y_dimension = y_dimension.clone(range=minmax[el.group])
+            new.append(el.redim(**{y_dimension.name: y_dimension}))
+        return overlay.clone(data=new)
```
philippjfr, line 214: Might suggest that you check if any of the layers were actually Charts and then make this operation a complete no-op if none of them were, i.e. simply return `overlay` as is without cloning it.

maximlt:
> Just to clarify, this operation is meant to be invoked explicitly by a user or could we use compositors to automatically apply this if `subcoordinate_y` is set?

I didn't think about compositors so yes it is meant to be invoked explicitly. If it was applied through a compositor, would it be possible for users to disable it if they don't want this behavior?

(I'll fix the conflicts)

philippjfr:
>If it was applied through a compositor, would it be possible for users to disable it if they don't want this behavior?

Sure, if we exposed a `subcoordinate_group_normalize` option (or ideally something nicer) it could be made contingent on that being enabled.

holoviews/operation/normalization.py
maximlt: Implemented in https://github.com/holoviz/holoviews/pull/6124/commits/b3965b4dbcdaaa0410812b8a9918c9436f90935b

maximlt:
> > If it was applied through a compositor, would it be possible for users to disable it if they don't want this behavior?
> 
> Sure, if we exposed a `subcoordinate_group_normalize` option (or ideally something nicer) it could be made contingent on that being enabled.

So you would prefer this was implemented and applied by default via a registered compositor? If so, I may need some pointers as I just tried that but didn't manage to get the compositor operation registered on an "Overlay" to run.

philippjfr:
I guess I'd like to hear what others think. Should this behavior be automatic or toggleable with a plot option?

jlstevens:
My own feeling is that this operation addresses a very niche use-case. Having the operation be explicit helps make it a bit clearer how things work, especially as it is an unusual way to use the group/label system (which we have generally been moving away from over the years).

My own feeling is that we should document this operation and simple tell users how to register a suitable compositor if they really want it (i.e. if they will be using this feature all the time) and *not* make it a default compositor that is enabled by default.

So in short, my own feelings are:

* Should this operation be usable in a compositor? Yes
* Should we have that compositor enabled by default? No (we should just document it for people who need it)

philippjfr:
>I'm not sure how to convert the operation to be usable in a compositor.

It already is, my request to make it a no-op was the main thing that's needed.

>As far as I can see, compositors apply to an operation to each element of an Overlay. 

Compositors operate at the level of an Overlay and can combine one or more layers.

philippjfr:
I'm fine with merging this without the compositor for now.

philippjfr:
I am a little unclear about one thing though, @jlstevens makes it sound this operation is extremely niche but I thought it is the desired behavior for grouped layers when `subcoordinate_y` is enabled. Is that not the case? Certainly [the issue @droumis created](https://github.com/holoviz/holoviews/issues/6123) explicitly says: 

> If groups are specified, normalize the y-range of curve-like elements per group, **by default**.

philippjfr:
>Nice! Are you done reviewing it?

I'd like an answer to my question above first, as it currently stands this does not address the ask outlined in the issue.

jlstevens:
I suppose I consider all use of `subcoordinate_y` fairly niche.

Essentially, I am wary of the additional complexity and that this normalization behavior based on group isn't entirely intuitive (good docs will help!). If we do as Philipp suggests and only do anything when `subcoordinate_y` is enabled, I am happy to see this compositor on by default, if we think this is the right behavior for `subcoordinate_y`.

philippjfr:
Got you, yes the proposal was never to enable this for all overlays, only for subcoordinate plots. I had vaguely recalled you could make a compositor dependent on a plot option already but I guess that was me misremembering as I can't find any mention of that now.

maximlt:
> I'd like an answer to my question above first, as it currently stands this does not address the ask outlined in the issue.

Good catch, I'll see with Demetris.

Meanwhile:

> we'd have to add support for patterns that match on any number of elements

Yes true! I've also tried `"Chart * Chart"` but that doesn't work.

---

I'll make the operation a no-op is the overlay is not a subcoord-y overlay and will chat with Demetris.

maximlt:
> I'll make the operation a no-op is the overlay is not a subcoord-y overlay

I'm actually not sure how to find whether an overlay is a subcoord-y overlay in the operation. `subcoordinate_y` is a Parameter of (Bokeh) `ElementPlot` and propagated to `OverlayPlot`, except that it seems to propagation happens after the operation runs, `overlay.opts.get('plot').kwargs.get('subcoordinate_y', False)` is always False as far as I can see.

I've added some warnings inspired by the checks added in https://github.com/holoviz/holoviews/pull/6122.

@droumis can you remind me of the outcome of our discussion on Thursday about applying the normalization per group **by default or not**? I think we said we're fine with this not being the default behavior but I'd like you to confirm this (my brain 🧀 ). If not, `Compositor` will need to be extended to:

- Add support for patterns that match any number of elements. Currently, you can only specify e.g. `Curve * Curve * Curve` which means it'll apply to the first 3 Curves of an overlay.
- The operation is meant to work for all `Chart` element types but as far as I can see patterns are only applied to specific types, i.e. `Chart * Chart * Chart` doesn't work.

maximlt:
> Yea, we agreed that because of the complexity involved, it would be sufficient (for now) to require group-wise normalization to be explicitly enabled, and so not applied by default.

Thanks! Then I'll just ask @philippjfr to have a look again at this PR 🙏

philippjfr (approved):

jlstevens (approved):
Looks good to me!

Operations typically don't use the plot options but here they are simply checked to generate a warning. That seems entirely reasonable to me.

PR #6130, opened 2024-02-26
Update test for Bokeh 3.4

Hoxbro:
Noticed here:

https://github.com/holoviz/holoviews/actions/runs/8020821711/job/21924068272

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -132,7 +132,7 @@ jobs:
           name: unit_test_suite
           python-version: ${{ matrix.python-version }}
           channel-priority: strict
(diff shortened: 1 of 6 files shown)


Hoxbro:
Now, we also test the unit test with the latest panel and Bokeh dev release.

Will add the check latest release in a follow up PR, when the next version of Panels dev release have been made.

Hoxbro:
@mattpap, do you know what causes the change in lasso_select between 3.3 and 3.4? And if the change is expected and therefore okay?

mattpap:
> do you know what causes the change in lasso_select between 3.3 and 3.4? And if the change is expected and therefore okay?

Gesture and UI event handling was redesigned, among other things. It seems OK and aligned with the difference between old and new approaches.

PR #6144, opened 2024-03-04
General maintenance

Hoxbro:


setup.py
```diff
@@ -28,7 +28,7 @@
 
 # Test requirements
 extras_require['tests_core'] = [
-    'pytest',
+    'pytest <8.1',  # https://github.com/computationalmodelling/nbval/issues/202
```
Hoxbro, line 31: FYI @maxime, pytest 8.1 breaks nbval (and also nbsmoke), which likely needs a pin in hvplot.  

@philippjfr, I don't think this will affect Panel (right now) as `pytest-playwright` pins to 7.

Issue #6154, opened 2024-03-18
Datashader update causing an error in nyc_taxi plot


Azaya89:
### ALL software version info
active environment : examples-gallery-manage
active env location : /Users/mac/anaconda3/envs/examples-gallery-manage
                    shell level : 2
           user config file : /Users/mac/.condarc
populated config files : /Users/mac/.condarc
             conda version : 24.1.0
   conda-build version : 3.27.0
            python version : 3.11.7.final.0
                          solver : libmamba (default)
         virtual packages : __archspec=1=m2
                                        __conda=24.1.0=0
                                        __osx=14.0=0
                                        __unix=0=0
      base environment : /Users/mac/anaconda3  (writable)
      conda av data dir : /Users/mac/anaconda3/etc/conda
  conda av metadata url : None
           channel URLs : https://repo.anaconda.com/pkgs/main/osx-arm64
                                     https://repo.anaconda.com/pkgs/main/noarch
                                     https://repo.anaconda.com/pkgs/r/osx-arm64
                                     https://repo.anaconda.com/pkgs/r/noarch
          package cache : /Users/mac/anaconda3/pkgs
                                      /Users/mac/.conda/pkgs
       envs directories : /Users/mac/anaconda3/envs
                                     /Users/mac/.conda/envs
                   platform : osx-arm64
              user-agent : conda/24.1.0 requests/2.31.0 CPython/3.11.7 Darwin/23.0.0 OSX/14.0 solver/libmamba conda-libmamba-solver/23.11.0 libmambapy/1.5.3 aau/0.4.3 c/SebItLHGqIAPQnu5l4EPeg s/piIqTHBeLi6aFzSCHbfipg e/gZnUH7usRhoCmJYo2DgJMA
                  UID:GID : 501:20
               netrc file : None
           offline mode : False

### Description of expected behavior and the observed behavior
As part of the NumFOCUS SDG project where we are revitalizing and modernizing the example website, I was working on the [nyc_taxi example](https://holoviz-dev.github.io/examples/gallery/nyc_taxi/nyc_taxi.html). This involves updating all the necessary dependencies to the most recent working version.

However, after updating the dependencies to the latest versions and running the notebook, I observed that there was a Traceback error in one of the last plots where it previously used to work just fine. 

### Complete, minimal, self-contained example code that reproduces the issue

```
import holoviews as hv
import holoviews.operation.datashader as hd
...
pickups = hv.Points(df, ['pickup_x', 'pickup_y'])
[...]

PR #6155, opened 2024-03-19
Force vdims to strings for ImageStack

Hoxbro:
Fixes https://github.com/holoviz/holoviews/issues/6154

image

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -406,7 +406,7 @@ class aggregate(LineAggregationOperation):
             eldata = agg if ds_version > Version('0.5.0') else (xs, ys, agg.data)
             return self.p.element_type(eldata, **params)
         else:
-            params['vdims'] = list(agg.coords[agg_fn.column].data)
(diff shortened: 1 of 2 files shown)


Hoxbro:
Basic example of test which needs to be added:

``` python
import pandas as pd
import holoviews as hv
import datashader as ds

hv.extension("bokeh")
df = pd.DataFrame({"x": range(3), "y": range(3), "c": range(3)})
shade = datashade(hv.Points(df), aggregator=ds.count_cat("c"))
```

jbednar:
Thanks! `ds.count_cat()` is deprecated (or at least outdated); please use the more general `ds.by("c")`. Can we get a dev release with this fix included?

Hoxbro:
> ds.count_cat() is deprecated (or at least outdated); please use the more general ds.by("c")

As this is in the test suite, I see no problems using it. Also, from a user's point of view, it is not deprecated, as no warning was emitted when running the code. 

> Can we get a dev release with this fix included?

This was already released last week.

PR #6157, opened 2024-03-20
Update server tests to simulate loaded event

philippjfr:
Since Panel no longer processes events unless the server has been loaded we need to simulate loading in the tests.

### DIFF
--- a/holoviews/tests/plotting/bokeh/test_server.py
+++ b/holoviews/tests/plotting/bokeh/test_server.py
@@ -163,6 +163,9 @@ class TestBokehServer(ComparisonTestCase):
 
         cds = session.document.roots[0].select_one({'type': ColumnDataSource})
         self.assertEqual(cds.data['y'][2], 2)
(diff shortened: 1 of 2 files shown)


holoviews/tests/plotting/bokeh/test_server.py
```diff
@@ -163,6 +163,9 @@ def test_server_dynamicmap_with_stream(self):
 
         cds = session.document.roots[0].select_one({'type': ColumnDataSource})
         self.assertEqual(cds.data['y'][2], 2)
+        def loaded():
+            state._schedule_on_load(doc, None)
+        doc.add_next_tick_callback(loaded)
```
Hoxbro, line 168: Should we only add this if we are on Panel 1.4? Will this test now fail if we are using Panel 1.3?

philippjfr: Nope, it's safe either way.

PR #6163, opened 2024-03-27
Update holoviz_tasks version

Hoxbro:


Hoxbro:
I'm getting error 3221225781 on Windows. I looked here: https://stackoverflow.com/questions/49413443/trouble-debugging-error-with-exit-code-3221225781-missing-libraries-in-windows. 

I can see it is related to a missing library. I don't want to spend much more time on it, so I'm skipping it on Windows CI.

PR #6168, opened 2024-04-03
Add support for popups on selection streams

philippjfr:
This PR exposes the ability to pop up floating panes added in Bokeh 3.4 after a selection is made. The contents of the popup can be added to the `Stream` and will be populated when the user makes the first selection. The contents can be any Panel object or any component that can be rendered with Panel. To control the visibility of the popup this PR currently links the `visible` parameter of the provided component to the visibility of the popup itself.

Here's a simple example adding three distinct forms to each type of supported stream:

```python
points = hv.Points(np.random.randn(1000, 2))

def form(name):
    text_input = pn.widgets.TextInput(name='Description')
    button = pn.widgets.Button(name='Save', on_click=lambda _: layout.param.update(visible=False))
    layout = pn.Column(f'# {name}', text_input, button)
    return layout

hv.streams.BoundsXY(source=points, popup=form('Bounds'))
hv.streams.Lasso(source=points, popup=form('Lasso'))
hv.streams.Tap(source=points, popup=form('Tap'))

points.opts(tools=['box_select', 'lasso_select', 'tap'], width=600, height=600, size=6, color='black', fill_color=None)
```

floating_panel

- [x] Decide whether the `visible` approach is sensible
- [x] Ensure this works on the server
- [x] Add documentation
- [x] Add tests


jlstevens:
I know this is a draft but this looks like a decent API to me!

Initially I thought you were supplying a callback instead of a concrete layout to `popup`. While this is definitely a great start and it is good to support the basic case, do you think a callback could be supported as well? e.g. a callback that is supplied the stream instance so that you could make a pop-up that reflects the stream data (e.g. the position where you tapped)?

philippjfr:
Here's an example with the callback support:

```python
points = hv.Points(np.random.randn(1000, 2))

hv.streams.Selection1D(source=points, popup=lambda index: points.iloc[index].dframe().describe() if index else None)

points.opts(tools=['box_select', 'tap', 'lasso_select'], height=500, width=500, size=6, color='black', fill_color=None)
```

sel_popup

jlstevens:
How is the exact location of the popup decided? I assume there are multiple different policies that might be sensible depending on the stream? I suppose the default I would expect is the position of the cursor after the interaction generating the popup...

philippjfr:
>How is the exact location of the popup decided? I assume there are multiple different policies that might be sensible depending on the stream? I suppose the default I would expect is the position of the cursor after the interaction generating the popup...

Currently it's:

Tap: x/y location
Box-select: Top-right corner
Lasso: Top-right coordinate

But yes, clearly for actual selections like Selection1D it'd be better if it was based on the overall selection rather than the geometry of the last selection.

jbednar:
Beautiful! I'd expect this example to show Tap selecting (and highlighting) a point in the same way that the lasso and box select do; is there something more that needs to be added to provide selection of a data point and not simply tapping a location?

philippjfr:
>I'd expect this example to show Tap selecting (and highlighting) a point in the same way that the lasso and box select do; is there something more that needs to be added to provide selection of a data point and not simply tapping a location?

If you look closely it does show that, the default tap selection mode is to combine multiple selections though so you end up returning selections even when you click a random location (if a previous selection already exists).

philippjfr:
>Oh, so if Tap were the first thing you had used (and/or if you had been closer to a single point) it would have highlighted that point? If so, sounds good.

Yes, and if I hadn't hit a point it wouldn't have popped up anything.

jbednar:
Great. I guess there are two very different modes to think about how to use this functionality: (1) Annotating or calculating info or otherwise working with existing data points, where the tools are for data-point selection, and (2) Adding new entities, typically a range on an axis, or a bounding box or bounding region, then annotating that thing (e.g. "data drops out here", in a region where there are no data points). Presumably the same pop-up approach works for both, but figured I would bring it up just to be sure.

jlstevens:
In holonote, a `DynamicMap` creates an 'indicator' element to show where you want to add an annotation (which could well be somewhere where there is no existing data). As long as this approach could attach a popup to *that* element, then I think the second case mentioned can be handled as a consequence of the first. I think having the popup be associated with an existing element (even if it is a transient indicator) is a sensible approach.

Also, I imagine you could use your stream to create whatever element in a `DynamicMap` and still have the popup appear (this is probably how you would do it, I imagine).

ahuang11:
Just tried it on browser and it encountered the pending write issue. I think it has to due with the async task

```python
import panel as pn
import numpy as np
import holoviews as hv
hv.extension("bokeh")


points = hv.Points(np.random.randn(1000, 2))

def form(name):
    text_input = pn.widgets.TextInput(name='Description')
    button = pn.widgets.Button(name='Save', on_click=lambda _: layout.param.update(visible=False))
    layout = pn.Column(f'# {name}', text_input, button)
    return layout

hv.streams.BoundsXY(source=points, popup=form('Bounds'))
hv.streams.Lasso(source=points, popup=form('Lasso'))
hv.streams.Tap(source=points, popup=form('Tap'))

pn.panel(points.opts(tools=['box_select', 'lasso_select', 'tap'], width=600, height=600, size=6, color='black', fill_color=None)).show()

```

```python
Task exception was never retrieved
future: <Task finished name='Task-95' coro=<PopupMixin._populate() done, defined at [/Users/ahuang/repos/holoviews/holoviews/plotting/bokeh/callbacks.py:631](https://file+.vscode-resource.vscode-cdn.net/Users/ahuang/repos/holoviews/holoviews/plotting/bokeh/callbacks.py:631)> exception=RuntimeError('_pending_writes should be non-None when we have a document lock, and we should have the lock when the document changes')>
  File "/Users/ahuang/repos/holoviews/holoviews/plotting/bokeh/callbacks.py", line 664, in _populate
    self._panel.elements = [model]
  File "/Users/ahuang/miniconda3/envs/holoviews/lib/python3.10/site-packages/bokeh/core/has_props.py", line 338, in __setattr__
    return super().__setattr__(name, value)
  File "/Users/ahuang/miniconda3/envs/holoviews/lib/python3.10/site-packages/bokeh/core/property/descriptors.py", line 333, in __set__
    self._set(obj, old, value, setter=setter)
  File "/Users/ahuang/miniconda3/envs/holoviews/lib/python3.10/site-packages/bokeh/core/property/descriptors.py", line 621, in _set
    self._trigger(obj, old, value, hint=hint, setter=setter)
  File "/Users/ahuang/miniconda3/envs/holoviews/lib/python3.10/site-packages/bokeh/core/property/descriptors.py", line 699, in _trigger
    obj.trigger(self.name, old, value, hint, setter)
  File "/Users/ahuang/miniconda3/envs/holoviews/lib/python3.10/site-packages/bokeh/model/model.py", line 571, in trigger
    super().trigger(descriptor.name, old, new, hint=hint, setter=setter)
  File "/Users/ahuang/miniconda3/envs/holoviews/lib/python3.10/site-packages/bokeh/util/callback_manager.py", line 188, in trigger
    self.document.callbacks.notify_change(cast(Model, self), attr, old, new, hint, setter, invoke)
  File "/Users/ahuang/miniconda3/envs/holoviews/lib/python3.10/site-packages/bokeh/document/callbacks.py", line 249, in notify_change
    self.trigger_on_change(event)
  File "/Users/ahuang/miniconda3/envs/holoviews/lib/python3.10/site-packages/bokeh/document/callbacks.py", line 413, in trigger_on_change
    invoke_with_curdoc(doc, invoke_callbacks)
  File "/Users/ahuang/miniconda3/envs/holoviews/lib/python3.10/site-packages/bokeh/document/callbacks.py", line 443, in invoke_with_curdoc
    return f()
  File "/Users/ahuang/miniconda3/envs/holoviews/lib/python3.10/site-packages/bokeh/document/callbacks.py", line 412, in invoke_callbacks
    cb(event)
  File "/Users/ahuang/miniconda3/envs/holoviews/lib/python3.10/site-packages/bokeh/document/callbacks.py", line 276, in <lambda>
    self._change_callbacks[receiver] = lambda event: event.dispatch(receiver)
  File "/Users/ahuang/miniconda3/envs/holoviews/lib/python3.10/site-packages/bokeh/document/events.py", line 353, in dispatch
    super().dispatch(receiver)
  File "/Users/ahuang/miniconda3/envs/holoviews/lib/python3.10/site-packages/bokeh/document/events.py", line 219, in dispatch
    cast(DocumentPatchedMixin, receiver)._document_patched(self)
  File "/Users/ahuang/miniconda3/envs/holoviews/lib/python3.10/site-packages/bokeh/server/session.py", line 244, in _document_patched
    raise RuntimeError("_pending_writes should be non-None when we have a document lock, and we should have the lock when the document changes")
RuntimeError: _pending_writes should be non-None when we have a document lock, and we should have the lock when the document changes
```

ahuang11:
Multiple taps result in 
```python

UnknownReferenceError: can't resolve reference 'p1128'
```

ahuang11:
Okay refactored with comments; the two issues above are fixed. 

Base notebook; no more `UnknownReferenceError: can't resolve reference 'p1128'`
https://github.com/holoviz/holoviews/assets/15331990/c5cdd84c-2de5-4e74-9bd4-525a380dce46

Server; to fix `RuntimeError: _pending_writes should be non-None`, I dropped async
https://github.com/holoviz/holoviews/assets/15331990/896a1887-0e9e-4c58-97bf-0136fb4d1be7

Callbacks
https://github.com/holoviz/holoviews/assets/15331990/bcc025bd-ceb5-4ad8-9b12-935a497a98f3

ahuang11:
Okay, I think this is ready for review now.

MarcSkovMadsen:
A cool use case for communication would be selecting a set of points, tap to pop up a chat interface, then start talking about the selected points 😄

philippjfr (commented):
Since it's my PR I can't approve so I'll leave one comment (that doesn't have to be addressed here). The remaining thing that would be nice here if the position of the popup on a `Selection1D` stream was relative to the overall selection rather than the last selected geometry. Don't think that's straightforward though.

ahuang11:
> if the position of the popup on a Selection1D stream was relative to the overall selection

Relative as in the center of the entire selection?

But yes, maybe separately in a future feature request?

philippjfr:
Relative to the top-right most coordinate in the selection I would say.

Hoxbro:
I will review it more thoroughly tomorrow but I have two things right now:

1) I think we should make a new page for this pop-up feature. I'm not exactly sure where it should be, but I think it is important to clarify that this is only a bokeh feature. Another thing is we should inform people what Panel is. I think it is the first time it is used directly for a HoloViews feature (I haven't checked, but that is what my gut says :upside_down_face:)
2) We should have a try/except based if the callable fails.

ahuang11:
I think that whole notebook is only possible in Bokeh right?

> We should have a try/except based if the callable fails.

What happens on except? Would it show the traceback in the panel or generic say something failed? Maybe this should be on the user instead?

Hoxbro:
> I think that whole notebook is only possible in Bokeh right?

We use `v.extension('bokeh', 'matplotlib')` at the top, so I don't think it is. 

> Would it show the traceback in the panel or generic say something failed?

Not sure. I haven't played around with this. The problem is that a callable takes data from the stream, and data can be weird. This will likely break things, and I feel this will be somewhat hidden in a notebook environment.

ahuang11:
I don't think matplotlib supports Taps, or any interactivity though (at least through HoloViews or I'm unaware)
---
Try/except would make it more hidden I would imagine?

Hoxbro:
> I don't think matplotlib supports Taps, or any interactivity though (at least through HoloViews or I'm unaware)

It does not. I still think this is a new approach that needs a page for itself. 

> Try/except would make it more hidden I would imagine?

I'm not saying you should ignore it. We should have a good way to signal that something errored. What that should be I'm not sure about.

ahuang11:
I don't think it needs a new page; there's enough pages I think, but I'll defer to others' judgement

image

--- 

> We should have a good way to signal that something errored.

To the user or developer?

If for the developer:
```python
import numpy as np
import panel as pn
import holoviews as hv

hv.extension("bokeh")


def popup(index):
    raise
    if index:
        return pn.pane.Markdown(f"Selected index: {index[0]}")
    return pn.pane.Markdown("No selection")


points = hv.Points(np.random.randn(1000, 2))

hv.streams.Selection1D(
    source=points,
    popup=popup,
)

points.opts(
    tools=["box_select", "tap", "lasso_select"],
    height=500,
    width=500,
    size=6,
    color="black",
    fill_color=None,
)
```

image

Hoxbro:
> I don't think it needs a new page; there's enough pages I think.

Ok. I still think there is a gap of information by importing Panel and starts using it without any explanation. I would maybe import it when needed with a small explanation of what it is and then reference to the panel website. 

> If for the developer:

That's more than fine (I haven't run the code.) I was afraid it was hidden from the developer.

Another thing. How does this work with DynamicMaps?

philippjfr:
We should start with an example popping up another plot rather than a panel object.

ahuang11:
I think it's the same if it's a DynamicMap (or am I misunderstanding)?

image


```python
import xarray as xr
import holoviews as hv

hv.extension("bokeh")



def popup(x, y):
    print(y)
    return f"Y={y}"


ds = xr.tutorial.open_dataset("air_temperature").isel(time=slice(0, 3))
coord_stream = hv.streams.Tap(x=-88 + 360, y=40)
time_stream = hv.streams.Tap(x=ds["time"].values[0], popup=popup)

def create_geomap(x, y):
    ds_sel = ds.sel(time=x, method="nearest")
    return hv.Image(ds_sel, ["lon", "lat"], ["air"]).opts(
        "Image", title=str(x)[:16], tools=["tap", "hover"])

def create_timeseries(x, y):
    ds_sel = ds.sel(lon=x, lat=y, method="nearest")
    return hv.Curve(ds_sel, ["time"], ["air"]).opts(
        "Curve", title=f"{x:.1f}, {y:.1f}", framewise=True, tools=["tap", "hover"])

def create_point(x, y):
    return hv.Points((x, y)).opts("Points", color="red", marker="x", size=25)

def create_vline(x, y):
    return hv.VLine(x).opts("VLine", color="red")

# create base plots
geomap = hv.DynamicMap(create_geomap, streams=[time_stream])
timeseries = hv.DynamicMap(create_timeseries, streams=[coord_stream])

# link source
coord_stream.source = geomap
time_stream.source = timeseries

# create annotations
coord_point = hv.DynamicMap(create_point, streams=[coord_stream])
time_vline = hv.DynamicMap(create_vline, streams=[time_stream])

# overlay + layout
layout = geomap * coord_point + timeseries * time_vline
layout
```

examples/user_guide/13-Custom_Interactivity.ipynb
```diff
@@ -18,7 +18,8 @@
     "import holoviews as hv\n",
     "from holoviews import opts\n",
     "\n",
-    "hv.extension('bokeh', 'matplotlib')"
+    "hv.extension('bokeh', 'matplotlib')\n",
+    "pn.extension()"
```
Hoxbro, line 22: Is matplotlib necessary here?

And `pn.extension` should also be moved down.

philippjfr:
>I think it's the same if it's a DynamicMap (or am I misunderstanding)?

It's a problem, because right now it's not straightforward to figure out which frame of the DynamicMap is currently displayed. I could also imagine that we provide the current element to the callback.

Hoxbro (commented):
I'm really excited about this functionality! I have left some comments.  

Is there a better way to close the popup after it has been popup? It is too hard of a setup right now to be able to close it. I will make a quick attempt at it. 

An example of what I was talking about with DynamicMap is having it outside the function. The example below will always return no selection.
``` python
import numpy as np
import panel as pn
import holoviews as hv

hv.extension("bokeh")


def popup(index):
    if index:
        return pn.pane.Markdown(f"Selected index: {index[0]}")
    return pn.pane.Markdown("No selection")


streams = hv.streams.Selection1D(popup=popup)
points = hv.DynamicMap(
    lambda *args, **kwargs: hv.Points(np.random.randn(1000, 2)), streams=[streams]
)

points.opts(
    tools=["box_select", "tap", "lasso_select"],
    height=500,
    width=500,
    size=6,
    color="black",
    fill_color=None,
)
```

examples/user_guide/13-Custom_Interactivity.ipynb
Hoxbro: I have moved `pn.extension` down.

Hoxbro:
I have added a close button to the pop-up, making it easy to close a pop-up. There seem to be some problems when a pop-up has been closed, and another tool is used. It shows both of the popups. It does not seem to be a problem with callbacks. 

https://github.com/holoviz/holoviews/assets/19758978/7bf529a1-41ee-4155-84a3-0f010b315df9

There is a small regression in the UI test suite because now there are two buttons. It should be easy to fix.

examples/user_guide/13-Custom_Interactivity.ipynb
```diff
@@ -410,6 +411,169 @@
...
+    "    width=500,\n",
+    "    height=500\n",
+    ")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "It can also be a [Panel](https://panel.holoviz.org/index.html) object or any component that can be rendered with Panel, which is an open-source Python library designed to streamline the development of robust tools, dashboards, and complex applications entirely within Python.\n",
```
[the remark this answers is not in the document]
ahuang11: No, I don't think so. I took it from the front page of Panel. I would assume Marc wrote it: https://panel.holoviz.org/index.html

> what panel components can be used here
any component that can be rendered with Panel

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -547,14 +560,140 @@ def _process_msg(self, msg):
...
+        stream = self.streams[0]
+        if not getattr(stream, 'popup', None):
+            return
+        elif Panel is None:
+            warn("Popup requires Bokeh >= 3.4")
+            return
+        self._panel = Panel(
+            position=XY(x=math.nan, y=math.nan),
+            anchor="top_left",
+            stylesheets=[
```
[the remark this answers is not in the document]
ahuang11: Not sure where to expose it on.

examples/user_guide/13-Custom_Interactivity.ipynb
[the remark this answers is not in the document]
Hoxbro: Try to see this as a user's first introduction to the Panel and what they likely will need from Panel inside a pop-up. I'm pretty sure they won't need a dashboard, so no need to mention it.

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -547,14 +560,140 @@ def _process_msg(self, msg):
...
+            return dict(x=event.geometry['x1'], y=event.geometry['y1'])
+        elif event.geometry['type'] == 'poly':
+            return dict(x=np.max(event.geometry['x']), y=np.max(event.geometry['y']))
+
+    def _populate(self, event):
+        # ^ can't be async or else server doesn't work
+
+        if not event.final:
+            return
+        popup = self.streams[0].popup
```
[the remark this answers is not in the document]
ahuang11: Philipp and I decided that it wouldn't make much sense to have two popups simultaneously (at least for this PR). For now we will loop through and find the first one with popup.

holoviews/plotting/bokeh/callbacks.py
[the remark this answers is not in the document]
ahuang11: If the returned popup has no stylesheets, it will not default to this; otherwise, use the stylesheets from the user defined popup.

examples/user_guide/13-Custom_Interactivity.ipynb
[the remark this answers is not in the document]
ahuang11: Okay fair; I updated to mention panes and widgets and some notable ones.

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -547,14 +567,176 @@ def _process_msg(self, msg):
...
+            return dict(x=np.max(event.geometry['x']), y=np.max(event.geometry['y']))
+
+    def _schedule_populate(self, event):
+        # IMPORTANT: NEED FIX BY PHILIPP
+        # - doesn't work on server; tap x/y updates late; so results in None, None
+        # - if pan is active, and tap, it gets dragged
+        state.execute(partial(self._populate, event), schedule=True)
+
+    def _populate(self, event):
+        if not event.final:
```
Hoxbro, line 658:
```suggestion
        if not event.final or event.geometry["type"] not in self.geom_type:
```

This is to fix the problem I observed in https://github.com/holoviz/holoviews/pull/6168#issuecomment-2044671358 (Maybe need extra safeguards).

ahuang11: Thanks, that works.

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -547,14 +567,177 @@ def _process_msg(self, msg):
...
+            width: 100%;
+            height: 100%;
+            top: -1em;
+        }
+        .bk-btn, .bk-btn:hover, .bk-btn:active, .bk-btn:focus {
+            background: none;
+            border: none;
+            color: inherit;
+            cursor: pointer;
+            padding: 0.5em;  # increase hit box or else playwright can't click
```
Hoxbro, line 597: Can you try [evaluate](https://stackoverflow.com/questions/70361865/how-to-get-the-locator-for-an-css-after-selector) in the test (if you haven't already)?

Adding a Python comment in CSS also seems a bit weird.

holoviews/tests/ui/bokeh/test_callback.py
```diff
@@ -191,3 +191,120 @@ def range_function(x_range, y_range):
...
+    locator = page.locator(".bk-btn")
+    expect(locator).to_have_count(0)
+
+
+
+@pytest.mark.skipif(not bokeh34, reason="< Bokeh 3.4 does not support popup")
+@pytest.mark.usefixtures("bokeh_backend")
+def test_stream_popup_close_button(serve_hv):
+    def popup_form(x, y):
+        return "Hello"
```
Hoxbro, line 294: I don't think I had this problem with callbacks only the static popup

holoviews/tests/ui/bokeh/test_callback.py
```diff
@@ -191,3 +191,120 @@ def range_function(x_range, y_range):
...
+    points = hv.Points(np.random.randn(10, 2)).opts(tools=["tap", "box_select"])
+    hv.streams.Tap(source=points, popup=popup_form)
+    hv.streams.BoundsXY(source=points, popup=popup_form)
+
+    page = serve_hv(points)
+    hv_plot = page.locator('.bk-events')
+    expect(hv_plot).to_have_count(1)
+    hv_plot.click()
+
+    locator = page.locator(".bk-btn.bk-btn-default")
```
Hoxbro, line 305:
```suggestion
    locator = page.locator(".bk-btn .bk-btn-default")
```

examples/user_guide/13-Custom_Interactivity.ipynb
```diff
@@ -410,6 +410,170 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "def popup_tool_name(name):\n",
+    "    return f\"You used {name}!\"\n",
+    "\n",
+    "points = hv.Points(np.random.randn(1000, 2))\n",
+    "\n",
+    "hv.streams.BoundsXY(source=points, popup=popup_tool_name(\"Box Select\"))\n",
```
Hoxbro, line 436: Why don't we just use a string here?

holoviews/tests/ui/bokeh/test_callback.py
ahuang11: What problem?

Hoxbro: Sorry, I thought it was a test for my problem. But I think it is just a test for the close button.

holoviews/tests/ui/bokeh/test_callback.py
ahuang11:
```
E         - waiting for locator(".bk-btn .bk-btn-default")
E         -   locator resolved to 0 elements
```

ahuang11:
@philippjfr I think this is in a good place for you to take back over to handle async / delayed stream results

https://github.com/holoviz/holoviews/pull/6168/files#diff-c0b814bedcee02b97de63fa13fc6a00858b3def2afc67eef7cdfaf04524be863R654-R656

jbednar:
> In holonote, a DynamicMap creates an 'indicator' element to show where you want to add an annotation (which could well be somewhere where there is no existing data). As long as this approach could attach a popup to that element, then I think the second case mentioned can be handled as a consequence of the first. 

This approach may also be how to support datashaded or rasterized plots, creating a bounding box around the region selected, then running a query against the original dataset.

> I don't think matplotlib supports Taps, or any interactivity though

More precisely, _our_ matplotlib backend doesn't support that. Matplotlib itself offers Jupyter-based interactivity via `ipympl`.

> We should have a try/except based if the callable fails.

An app definitely needs to be robust against a failure in the callback or in any other computation resulting from a UI action. Maybe a JS console warning could be raised, at best, since everyone ignores those anyway. I think it's more likely to be noise than signal to print something on the Python terminal, because users can end up selecting things that just don't work with the callback, and the app shouldn't be affected by that.

philippjfr:
>I think it's more likely to be noise than signal to print something on the Python terminal, because users can end up selecting things that just don't work with the callback, and the app shouldn't be affected by that.

The developer should be made aware of this so they can add the appropriate error handling themselves.

ahuang11:
Thanks for looking into this. I couldn't get the actual contents to show up, but it might still be in progress.

https://github.com/holoviz/holoviews/assets/15331990/c00f3bdc-0490-41c4-84fa-d4fe5fe43b87

Also, using `pn.serve` results in
`exception=RuntimeError('_pending_writes should be non-None when we have a document lock, and we should have the lock when the document changes')>`

```python
import panel as pn
import holoviews as hv
import numpy as np

hv.extension("bokeh")

points = hv.Points(np.random.randn(1000, 2))

hv.streams.Selection1D(source=points, popup=lambda index: points.iloc[index].dframe().describe() if index else None)

points.opts(tools=['box_select', 'tap', 'lasso_select'], height=500, width=500, size=6, color='black', fill_color=None)
```

philippjfr:
Resolved the remaining issues, would appreciate more testing.

ahuang11:
Thanks! I think most of them are resolved, but I couldn't get lasso to work on server / vscode:

```python
import holoviews as hv
import panel as pn
import numpy as np
pn.extension()
hv.extension("bokeh")

points = hv.Points(np.random.randn(1000, 2))

def form(name):
    text_input = pn.widgets.TextInput(name='Description')
    button = pn.widgets.Button(name='Save', on_click=lambda _: layout.param.update(visible=False))
    layout = pn.Column(f'# {name}', text_input, button)
    return layout

hv.streams.BoundsXY(source=points, popup=form('Bounds'))
hv.streams.Lasso(source=points, popup=form('Lasso'))
hv.streams.Tap(source=points, popup=form('Tap'))

points.opts(tools=['box_select', 'lasso_select', 'tap'], width=600, height=600, size=6, color='black', fill_color=None)
```

https://github.com/holoviz/holoviews/assets/15331990/0473b95a-d6b8-4a12-bcdb-c0adcf466767

ahuang11:
Okay, lasso select for Selection1D does show final occasionally, but only on init, when index = [].

```python
import holoviews as hv
import panel as pn
import numpy as np
pn.extension()
hv.extension("bokeh")

def popup_distribution(index):
    print(index)
    x, y = points.iloc[index].data.T
    return hv.Distribution((x, y)).opts(
        width=100,
        height=100,
        toolbar=None,
        yaxis="bare",
        xlabel="",
        xticks=[-1, 0, 1],
        xlim=(-2, 2),
    )


points = hv.Points(np.random.randn(1000, 2))

hv.streams.Selection1D(
    source=points,
    popup=popup_distribution,
)

pn.serve(points.opts(
    tools=["box_select", "lasso_select"],
    active_tools=["lasso_select"],
    size=6,
    color="black",
    fill_color=None,
    width=500,
    height=500
))
```

https://github.com/holoviz/holoviews/assets/15331990/d5d93981-3e76-48b5-8cd6-86be872acca6

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -547,14 +566,177 @@ def _process_msg(self, msg):
...
+            popup = stream.popup
+            if popup is not None:
+                break
+
+        if callable(popup):
+            popup = popup(**stream.contents)
+
+        # If no popup is defined, hide the panel
+        if popup is None:
+            if self._panel.visible:
```
Hoxbro, line 676: Just saw this error in an app:

``` python-traceback
2024-04-17 09:58:13,275 Exception in callback functools.partial(<bound method IOLoop._discard_future_result of <tornado.platform.asyncio.AsyncIOMainLoop object at 0x76878123fc50>>, <Task finished name='Task-155' coro=<ServerSession.with_document_locked() done, defined at /home/shh/miniconda3/envs/holoviz/lib/python3.12/site-packages/bokeh/server/session.py:77> exception=AttributeError("'BoundsCallback' object has no attribute '_panel'")>)
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.12/site-packages/tornado/ioloop.py", line 750, in _run_callback
    ret = callback()
          ^^^^^^^^^^
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.12/site-packages/tornado/ioloop.py", line 774, in _discard_future_result
    future.result()
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.12/site-packages/bokeh/server/session.py", line 98, in _needs_document_lock_wrapper
    result = await result
             ^^^^^^^^^^^^
  File "/home/shh/projects/holoviz/repos/holoviews/holoviews/plotting/bokeh/callbacks.py", line 338, in on_event
    await self.process_on_event()
  File "/home/shh/projects/holoviz/repos/holoviews/holoviews/plotting/bokeh/callbacks.py", line 363, in process_on_event
    self.on_msg(msg)
  File "/home/shh/projects/holoviz/repos/holoviews/holoviews/plotting/bokeh/callbacks.py", line 677, in on_msg
    if self._panel.visible:
       ^^^^^^^^^^^
AttributeError: 'BoundsCallback' object has no attribute '_panel'

```

philippjfr: Indeed, that needs a guard.

philippjfr:
Please test again, I think I've now resolved the lasso select issues. Code isn't the prettiest tbh.

ahuang11:
Thanks! It works great. I added a few tests for it.

ahuang11 (approved):

holoviews/plotting/bokeh/callbacks.py
ahuang11: Is this it?
https://github.com/holoviz/holoviews/pull/6168/files#diff-c0b814bedcee02b97de63fa13fc6a00858b3def2afc67eef7cdfaf04524be863R670

Issue #6033, opened 2023-11-04
Autoranging on multi-line plots limits view to a single line


droumis:
#### ALL software version info
Python              :  3.10.13 | packaged by conda-forge | (main, Oct 26 2023, 18:09:17) [Clang 16.0.6 ]
Operating system    :  macOS-13.5.2-arm64-arm-64bit
Panel comms         :  default

holoviews           :  1.18.0
bokeh               :  3.3.0
hvplot              :  0.9.0
jupyterlab          :  4.0.7

#### Description of expected behavior and the observed behavior
Interacting, via zoom or pan, on a plot with multiple lines limits the viewport to a single line if `autorange='y'` is set.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import pandas as pd
import numpy as np
import hvplot.pandas

length = 100
parameters = [
    (0, 1),      # Mean 0, SD 1
    (100, 20),   # Mean 100, SD 20
    (1000, 200), # Mean 1000, SD 200
    (-500, 50)   # Mean -500, SD 50
]

df_ = pd.DataFrame({f"{i}": np.random.normal(mean, sd, length) for i, (mean, sd) in enumerate(parameters)})
df_ = df_.reset_index().melt(id_vars=['index'], var_name='source', value_name='value')
df_['source'] = df_['source'].astype(int)

# Single Line Plot
s = df_.hvplot(x="index", y="value", downsample=True, autorange='y').opts(width=450, shared_axes=False)

# Multi Line Plot
m= df_.hvplot(x="index", y="value", by='source', downsample=True, autorange='y').opts(show_legend=False, width=450, shared_axes=False)

s + m

```

#### Screenshots or screencasts of the bug in action

https://github.com/holoviz/hvplot/assets/6613202/f8a62e34-b017-4f02-a88d-4a8e440c9b24

- [ ] I may be interested in making a pull request to address this

droumis:
Update from @jlstevens:

Together with @maximlt, they found out there is a more general problem with autoranging and overlays that is probably the cause; probably a regression since the multi-axis work. As such it will need to be fixed for the next holoviews release.

JLS will have another look this week

droumis:
@jlstevens, any update from last week?

jlstevens:
I'm currently working on it - unfortunately, this is a really tricky bit of code!

droumis:
Hi @jlstevens, could you provide a status update on this?

jlstevens:
[...]

PR #6173, opened 2024-04-04
Ensure plot ranges for all renderers are combined in auto-ranging

philippjfr:
Fixes a small oversight in the auto-ranging implementation by combining the ranges for all renderers instead of overwriting them. Also simplifies the code by using the new `Bokeh.index.find_one` API instead of rolling our own implementation.

Fixes https://github.com/holoviz/holoviews/issues/6033

- [x] Add UI test

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -2968,6 +2968,9 @@ class OverlayPlot(GenericOverlayPlot, LegendPlot):
         if self.top_level:
             self.init_links()
 
+        if self.autorange:
+            self._setup_autorange()
+
         self._execute_hooks(element)
 
         return self.handles['plot']
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1349,30 +1349,10 @@ class ElementPlot(BokehPlot, GenericElementPlot):
             return invert ? [upper, lower] : [lower, upper]
           }}
 
-          const ref = plot.id
-
-          const find = (view) => {{
-            let iterable = view.child_views === undefined ? []  : view.child_views
-            for (const sv of iterable) {{
-              if (sv.model.id == ref)
-                return sv
-              const obj = find(sv)
-              if (obj !== null)
-                return obj
-            }}
-            return null
-          }}
-          let plot_view = null;
-          for (const root of plot.document.roots()) {{
-            const root_view = window.Bokeh.index[root.id]
-            if (root_view === undefined)
-              return
-            plot_view = find(root_view)
-            if (plot_view != null)
-              break
-          }}
-          if (plot_view == null)
+          let plot_view = Bokeh.index.find_one(plot)
+          if (plot_view == null) {{
             return
+          }}
 
           let range_limits = {{}}
           for (const dr of plot.data_renderers) {{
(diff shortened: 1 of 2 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -1393,20 +1372,23 @@ def _setup_autorange(self):
               }}
             }}
 
-            if (y_range_name) {{
+            if (y_range_name in range_limits) {{
+              const [vmin_old, vmax_old] = range_limits[y_range_name]
+              range_limits[y_range_name] = [Math.min(vmin, vmin_old), Math.max(vmax, vmax_old)]
```
philippjfr, line 1377: This is the core of the change, the rest of the PR is just cleanup.

jlstevens: So `range_limits` was accumulating a bunch of mins and maxes that were not being used!

One thing though, shouldn't it segregate aggregations separately on the left/right axis if you have twin axes enabled?

holoviews/plotting/bokeh/element.py
```diff
@@ -1393,20 +1372,23 @@ def _setup_autorange(self):
...
             }}
           }}
 
-           let range_tags_extras = plot.{dim}_range.tags[1]
-           if (range_tags_extras['autorange']) {{
-             let lowerlim = range_tags_extras['y-lowerlim'] ?? null
-             let upperlim = range_tags_extras['y-upperlim'] ?? null
-             let [start, end] = get_padded_range('default', lowerlim, upperlim, range_tags_extras['invert_yaxis'])
-             if ((start != end) && window.Number.isFinite(start) && window.Number.isFinite(end)) {{
-               plot.{dim}_range.setv({{start, end}})
```
jlstevens, line 1407: If I remember right, `dim` is what picks the axis for the twin axis support.

holoviews/plotting/bokeh/element.py
philippjfr: That's what the y_range_name is for.

jlstevens: Ok. I see that `y_range_name = renderer.model.y_range_name` and now I remember that this is how Bokeh defined multiple axes. Thanks for explaining!

holoviews/plotting/bokeh/element.py
```diff
@@ -2665,7 +2647,7 @@ class OverlayPlot(GenericOverlayPlot, LegendPlot):
                           'min_height', 'max_height', 'min_width', 'min_height',
                           'margin', 'aspect', 'data_aspect', 'frame_width',
                           'frame_height', 'responsive', 'fontscale', 'subcoordinate_y',
-                          'subcoordinate_scale']
+                          'subcoordinate_scale', 'autorange']
```
Hoxbro, line 2650: This is needed to get this to work, have updated the tests to account for this. 

``` python
import holoviews as hv; hv.extension('bokeh')
import numpy as np

np.random.seed(42)
ys = np.random.randn(101).cumsum(axis=0)

hv.Curve(ys) * hv.Curve(ys[::-1]).opts(autorange='y')
```

holoviews/plotting/bokeh/element.py
```diff
@@ -1393,20 +1372,23 @@ def _setup_autorange(self):
               }}
             }}
 
-            if (y_range_name) {{
```
Hoxbro, line 1396: Is `y_range_name` now always defined?

Hoxbro (approved):
Updated the code to work with overlay. Can you please review those changes?

Other than that, it looks good to me; you can merge after looking at my changes.

holoviews/plotting/bokeh/element.py
philippjfr: It always was, it's 'default' by default. I think the old PR was simply merged before this code was fully updated.

holoviews/plotting/bokeh/element.py
philippjfr: Yes, this looks good, did wonder why I had to set it on the Element and the overlay. Not sure why we hadn't set it to propagate before.

Hoxbro: Updated again to also work with this one `(hv.Curve(ys) * hv.Curve(ys[::-1])).opts(autorange='y')`. It also makes the range in the multiple-unit test reasonable.

PR #5780, opened 2023-06-22
Add an extra try when resolving attributes lookup

Hoxbro:
Fixes #5746

Giving 5 times to try to resolve the attribute with a delay of 0.01 seconds.

`asyncio.sleep(0.01)` is already used once before in the method, and 5 is an arbitrary choice that worked on my computer.

I also made minor changes by using `isequal` for comparison and setting an `Exception` to `AttributeError`, which are not related to the original issues.

PR #6189, opened 2024-04-12
Fix data aspect issues

### LINKED ISSUE #462: hvplot.image() zoom causes axes limits to bounce around
Thanks for contacting us! Please read and follow these instructions carefully, then delete this introductory text to keep your issue easy to read. Note that the issue tracker is NOT the place for usage questions and technical assistance; post those at [Discourse](https://discourse.holoviz.org) instead. Issues without the required information below may be closed immediately.

hvplot 0.5.2
bokeh 2.0.1

Plotting image results in axes and colorbar limits jumping around. A zoom box draw partially outside valid pixels of image causes further jumping around of axes limits.
 
https://gist.github.com/scottyhq/0e245fcd62d8a46b82a4186fcd8945e6#file-hvplot-zoom-ipynb
```
# Lots of window and colorbar rescaling happening when first plot, and upon zooming
band1.hvplot.image(x='x',y='y',
                   crs='epsg:32645', 
                   tiles='OSM', 
                   rasterize=True,
                   width=700, height=500, 
                   cmap='plasma', 
                   logz=True,
                   alpha=0.7)
```

 No error or traceback, but the behavior is buggy. See gif:

hvplotzoom

ahuang11:
The main issue here is that the [`current_l, current_r = plot.x_range.start, plot.x_range.end`](https://github.com/holoviz/holoviews/blame/main/holoviews/plotting/bokeh/element.py#L1163-L1165) **arrives late (or isn't up to date)** so the data aspect calculation is messed up. There is, however, `l, b, r, t = self.get_extents(element, ranges, dimension=range_dim)` which is up to date.

Here, "current bounds" (inaccurate ylim) is 0.39 instead of 0.67 as seen on the plot axes. I could remove those lines, and the data aspect no longer bounces.

image

It was introduced in this PR https://github.com/holoviz/holoviews/pull/4569 and fixes https://github.com/holoviz/hvplot/issues/462 so if I remove it, it reintroduces that error.

The only thing I'm pretty certain of is that if the plot isn't responsive, there's no need to re-calculate and ensure the data_aspect matches every time, which removes the bounciness:

https://github.com/holoviz/holoviews/assets/15331990/c6ec09ad-62d4-4549-979b-d2ead8b68fe4

I think if there's a way to introduce value throttling, it could help.

Ultimately, I don't fully understand the difference between ` l, b, r, t = self.get_extents(element, ranges, dimension=range_dim)` and `plot.x_range` & `plot.y_range`

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1141,6 +1141,8 @@ class ElementPlot(BokehPlot, GenericElementPlot):
                                'axes are %s, the option will '
                                'be ignored.' % ax_type)
         elif data_aspect:
+            if not options.get("responsive", False):
+                return
             plot = self.handles['plot']
             xspan = r-l if util.is_number(l) and util.is_number(r) else None
             yspan = t-b if util.is_number(b) and util.is_number(t) else None
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1141,13 +1141,13 @@ class ElementPlot(BokehPlot, GenericElementPlot):
                                'axes are %s, the option will '
                                'be ignored.' % ax_type)
         elif data_aspect:
-            if not options.get("responsive", False):
-                return
             plot = self.handles['plot']
             xspan = r-l if util.is_number(l) and util.is_number(r) else None
             yspan = t-b if util.is_number(b) and util.is_number(t) else None
 
             if self.drawn or (fixed_width and fixed_height) or (constrained_width or constrained_height):
+                if not options.get("responsive", False):
+                    return
                 # After initial draw or if aspect is explicit
                 # adjust range to match the plot dimension aspect
                 ratio = self.data_aspect or 1


philippjfr:
@ahuang11 Please make sure you always paste the code for your reproducers.

ahuang11:
```python
import panel as pn
import holoviews as hv

hv.Curve([0, 1, 2]).opts(data_aspect=1)
```

ahuang11:
Were you using this branch or main?

philippjfr:
main

ahuang11:
To reproduce, I need to zoom a decent amount at once.

ahuang11:
https://github.com/holoviz/holoviews/assets/15331990/3ab480eb-9797-4f39-b24a-72c77be74c20

philippjfr:
Okay, you should have specified that this is specific to VSCode (or rather to `pn.config.comms = 'ipywidgets'`).

Hoxbro:
This could be related to the change I made in https://github.com/holoviz/holoviews/pull/5780

philippjfr:
Neither notebooks nor the server cases seem to be affected here.

Hoxbro:
For versions, you could run `hv.show_versions()` next time.  This also outputs the comms.

ahuang11:
Are you able to reproduce with this?
https://github.com/holoviz/hvplot/issues/1105

>  VSCode (or rather to pn.config.comms = 'ipywidgets').

Okay I wasn't aware that VSCode notebook was different

philippjfr:
>Okay I wasn't aware that VSCode notebook was different

The difference is that:

- In classic notebook and JupyterLab we leverage regular notebook Comms, this allows for more direct communication
- In VSCode and some other non-standard notebook environments comms are not available therefore we have to wrap our components in an IPyWidget wrapper (provided by `jupyter-bokeh`), this means we can shuttle the comms via the ipywidget comm.
- On the server comms are handled via the dedicated websocket handler

In future just make sure you note which environment(s) you tested in.

philippjfr:
I suspect there's a deeper issue with syncing bokeh properties in the ipywidgets based scenario. Will keep digging.

ahuang11:
Okay, thanks for trying it out.

Anyhow, I think this PR might still be useful because I don't think the data aspect calculations (after the first) need to happen if the plot isn't `responsive=True`.

> In future just make sure you note which environment(s) you tested in.
In the future, if I happen to forget, I default to VSCode

philippjfr:
>Anyhow, I think this PR might still be useful because I don't think the data aspect calculations (after the first) need to happen if the plot isn't responsive=True.

This isn't true, axis labels can significantly impact the aspect.

philippjfr:
Really appreciate you chasing this btw. I just dug into the jupyter_bokeh IPyWidget wrapper and the issue seems to be to do with the serialization, it seems like the `rangesupdate` event that sends the updated ranges is not being serialized correctly which ends up serializing the entire plot (!!!) each time. This means the range update arrives late because there is such a big message being sent from frontend to the server.

philippjfr:
> I'm having a hard time wrapping my head around this

Me too, me too!

>After initialization, the plot's frame_width + frame_height is set

Will have to very carefully review that bit of code to make sure that always holds true.

Also my apologies, I meant the tick labels, not the axis labels.

philippjfr:
Turns out this was caused by a big oversight in `jupyter_bokeh` fixed here: https://github.com/bokeh/jupyter_bokeh/pull/204

The problem was that random events could end up being dropped and this reliably affected the end of the x_range, so the aspect calculation was thrown completely off by the fact that the `x_range.end` was incorrect.

ahuang11:
I'm wondering if there's an issue in Jupyter notebook too because in Jason's Gerrymandering recording, we see data aspect issue at around 5s.

https://github.com/holoviz-topics/examples/pull/369#issuecomment-2061262001

Issue #5952, opened 2023-10-24
`ylim` does not work for a `subcoordinates_y` overlay


droumis:
#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc)
<details>
<summary> show versions </summary>

Python              :  3.11.6 | packaged by conda-forge | (main, Oct  3 2023, 10:37:07) [Clang 15.0.7 ]
Operating system    :  macOS-13.5.2-arm64-arm-64bit
Panel comms         :  default

holoviews           :  1.18.0

bokeh               :  3.3.0
colorcet            :  3.0.1
cudf                :  -
dask                :  2023.10.0
datashader          :  -
geoviews            :  -
hvplot              :  0.9.0
ibis                :  -
IPython             :  8.16.1
jupyter_bokeh       :  -
jupyterlab          :  4.0.7
matplotlib          :  3.8.0
networkx            :  3.2
notebook            :  7.0.6
numba               :  0.57.1
numpy               :  1.24.4
pandas              :  2.1.1
panel               :  1.3.0
param               :  2.0.0
PIL                 :  10.1.0
plotly              :  -
pyarrow             :  13.0.0
scipy               :  1.11.3
skimage             :  0.22.0
spatialpandas       :  -
streamz             :  -
xarray              :  2023.10.1

</details>

#### Description of expected behavior and the observed behavior

With an overlay of curves with `subcoordinate_y=True`, each curve's subcoord y-axis is centered on integers in the underlying canvas axis (0,1,2..). My expected behavior was that setting ylim=(0,1) would display the center of the first curve axis until the center of the second. The observed behavior is that no y-limiting happens.

#### Complete, minimal, self-contained example code that reproduces the issue

```
import numpy as np
import holoviews as hv; hv.extension('bokeh')

# xlim works
xcurves = hv.Overlay([hv.Curve(np.random.rand(100), label=str(i)).opts(
    subcoordinate_y=True) for i in np.arange(10)]).opts(
    show_legend=False, xlim=(0,1))

# ylim does not work
ycurves = hv.Overlay([hv.Curve(np.random.rand(100), label=str(i)).opts(
    subcoordinate_y=True) for i in np.arange(10)]).opts(
    show_legend=False, ylim=(0,1))
```

#### Screenshots or screencasts of the bug in action
#####Setting xlim
[...]

PR #6190, opened 2024-04-15
subcoordinate_y: respect `ylim`

maximlt:
Fixes https://github.com/holoviz/holoviews/issues/5952

image

I want to wait for https://github.com/holoviz/holoviews/pull/6122 to be merged to fully set this PR. At the moment I think you can't zoom-out from the `ylim` which isn't good.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -508,11 +508,14 @@ class ElementPlot(BokehPlot, GenericElementPlot):
                 l, b, r, t = b, l, t, r
             if pos == 1 and self._subcoord_overlaid:
                 if isinstance(self.subcoordinate_y, bool):
-                    offset = self.subcoordinate_scale / 2.
-                    # This sum() is equal to n+1, n being the number of elements contained
-                    # in the overlay with subcoordinate_y=True, as the traversal goes through
-                    # the root overlay that has subcoordinate_y=True too since it's propagated.
-                    v0, v1 = 0-offset, sum(self.traverse(lambda p: p.subcoordinate_y))-2+offset
+                    if self.ylim and all(np.isfinite(val) for val in self.ylim):
+                        v0, v1 = self.ylim
+                    else:
+                        offset = self.subcoordinate_scale / 2.
+                        # This sum() is equal to n+1, n being the number of elements contained
+                        # in the overlay with subcoordinate_y=True, as the traversal goes through
+                        # the root overlay that has subcoordinate_y=True too since it's propagated.
+                        v0, v1 = 0-offset, sum(self.traverse(lambda p: p.subcoordinate_y))-2+offset
                 else:
                     v0, v1 = 0, 1
             else:
@@ -1043,7 +1046,6 @@ class ElementPlot(BokehPlot, GenericElementPlot):
             for m in models:
                 setattr(m, attr_accessor, val)
 
-
     def _update_grid(self, plot):
         if not self.show_grid:
             plot.xgrid.grid_line_color = None
@@ -1447,7 +1449,6 @@ class ElementPlot(BokehPlot, GenericElementPlot):
                 (isinstance(column, list) or column.dtype.kind not in 'SU')):
                 data[col] = [dims[i].pprint_value(v) for v in column]
 
-
     def get_aspect(self, xspan, yspan):
         """
         Computes the aspect ratio of the plot
@@ -1682,7 +1683,6 @@ class ElementPlot(BokehPlot, GenericElementPlot):
 
         return new_style
 
-
     def _glyph_properties(self, plot, element, source, ranges, style, group=None):
         properties = dict(style, source=source)
         if self.show_legend:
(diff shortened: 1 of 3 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -508,11 +508,14 @@ def _axis_props(self, plots, subplots, element, ranges, pos, *, dim=None,
...
-                    # in the overlay with subcoordinate_y=True, as the traversal goes through
-                    # the root overlay that has subcoordinate_y=True too since it's propagated.
-                    v0, v1 = 0-offset, sum(self.traverse(lambda p: p.subcoordinate_y))-2+offset
+                    if self.ylim and all(np.isfinite(val) for val in self.ylim):
+                        v0, v1 = self.ylim
+                    else:
+                        offset = self.subcoordinate_scale / 2.
+                        # This sum() is equal to n+1, n being the number of elements contained
+                        # in the overlay with subcoordinate_y=True, as the traversal goes through
+                        # the root overlay that has subcoordinate_y=True too since it's propagated.
```
philippjfr, line 517: Took me some time to parse this originally

```suggestion
                        # This sum() is equal to n+1, where n is the number of elements contained
                        # in the overlay with subcoordinate_y=True (including the the root overlay,
                        # which has subcoordinate_y=True due to option propagation)
```

philippjfr (approved):
Looks good.

maximlt:
`ylim` (`xlim` too) isn't a hard-bound in HoloViews unless `apply_hard_bounds` is set to `True` (https://github.com/holoviz/holoviews/pull/6056), instead, it sets the *initial* y-bounds. So users can freely zoom and pan outside of the initial view when `ylim` is set.

Let's dive more into the zoom UX with subcoord-y overlays!

By default, the user gets global `Box Zoom` and `Pan` zoom tools controlling the whole viewport (level 0) and a `Wheel Zoom` controlling the subcoord-y subplots (y-axis at level 1) => Users can zoom in globally with the `Box Zoom` but **can't zoom out at level 0** except with the `Reset` tool. This is a problem that was already there and that is made more obvious by this PR.

```python
import holoviews as hv
import numpy as np
hv.extension('bokeh')

x = np.linspace(0, 10*np.pi)
curves = [
    hv.Curve((x + i*np.pi/2, np.sin(x)), label=f'Line {i}').opts(
        subcoordinate_y=True,
    )
    for i in range(3)
]
hv.Overlay(curves).opts(show_legend=False)
```

holoviews_subcoor_ylim4

Users can add `Zoom In` and `Zoom Out` tools that control the subcoord-y subplots (y-axis at level 1).

```python
curves = [
    hv.Curve((x + i*np.pi/2, np.sin(x)), label=f'Line {i}').opts(
        subcoordinate_y=True, tools=['zoom_in', 'zoom_out'],
    )
    for i in range(3)
]
hv.Overlay(curves).opts(show_legend=False)
```

image

A little detour on a normal overlay behavior: it is not possible to directly add global tools to an overlay. This seems to be unrelated to subcoord-y.

```python
overlay = hv.Curve([-1, -2, -3]) * hv.Curve([1, 2, 3])
overlay.opts(tools=['zoom_in'])
# No zoom_in tool!
```

image

For normal overlays, the usual workaround, I think, is to directly add a tool to a least one element of the overlay.

```python
overlay = hv.Curve([-1, -2, -3]) * hv.Curve([1, 2, 3]).opts(tools=['zoom_in'])
overlay
```

image

This workaround doesn't work with subcoord-y overlays, it will not add a global `Zoom In` tool, instead, it'll just allow to zoom in on one subplot. That's because `wheel_zoom`, `zoom_in` and `zoom_out` tools are special-cased in the scope of a subcoord-y overlay, they're turned into subcoord-y zoom tools (y-axis at level 1).  In this case, the workaround is to **add a custom zoom instance to at least one element of the overlay**.

```python
from bokeh.models import WheelZoomTool

curves = [
    hv.Curve((x + i*np.pi/2, np.sin(x)), label=f'Line {i}').opts(
        subcoordinate_y=True, tools=[WheelZoomTool()],
    )
    for i in range(3)
]
hv.Overlay(curves).opts(show_legend=False, ylim=(0, 1))
```

holoviews_subcoor_ylim2

With **grouped** subcoord-y overlays, the same issues and workaround apply. Except that, at least at the moment, there are more zoom tools in the toolbar.

```python
curves = [
    hv.Curve((x + i*np.pi/2, np.sin(x)), group=group, label=f'{group} Line {i}').opts(
        subcoordinate_y=True, tools=[WheelZoomTool()],
    )
    for group in 'ABC'
    for i in range(3)
]
hv.Overlay(curves).opts(show_legend=False, ylim=(0, 1))
``` 

holoviews_subcoor_ylim3

In conclusion:

1. This PR exposes more the fact that setting a tool directly on an overlay is a no-op. I don't know if it's by design, a bug, or that it was just never implemented.
2. The workaround for subcoord-y overlays is to pass a custom wheel zoom / zoom in / zoom out instance.
3. This PR also exposes more of the issue users will have with differentiating a normal zoom tool with a subcoord-y zoom tool (y-axis at level 1). With @Hoxbro we discussed creating a custom icon, derived from the existing icons.

My take on these:

1. Looks like this deserves to be its own issue (I haven't found one)
2. I'd be okay documenting this workaround, as it's similar to the usual workaround with setting a tool on an overlay, at least until 1. is resolved.
3. I also think it deserves its own issue, the "hard" work being to create custom wheel zoom / zoom in / zoom out icons (no designer here 🙃 )

cc @droumis some food for thought

Issue #6027, opened 2023-12-11
By default `subcoordinate_y=True` should reverse element order


droumis:
A [ridgeplot](https://r-graph-gallery.com/294-basic-ridgeline-plot.html) (aka joyplot, aka ridgelineplot) is possible using the new `subcoordinate_y` feature. This plot usually makes use of an `Area` per curve. With `subcoordinate_y`, the order of elements should be reversed so that the bottom element is in the foreground. The issue was first identified [here](https://github.com/holoviz/hvplot/issues/1160).

<details><summary>Code</summary>

```python

import numpy as np
import holoviews as hv; hv.extension('bokeh')
from scipy.stats import gaussian_kde

categories = ['A', 'B', 'C', 'D', 'E']
data = {cat: np.random.normal(loc=i-2, scale=1.0, size=100) for i, cat in enumerate(categories)}
x = np.linspace(-5, 5, 100)

areas = []
for i, (cat, values) in enumerate(data.items()):
    pdf = gaussian_kde(values)(x)
        
    area = hv.Area((x, pdf), label=cat).opts(
        subcoordinate_y=True, 
        subcoordinate_scale=1.5,
    )
    areas.append(area)

ridge_plot_areas = hv.Overlay(areas).opts(
    width=900,
    height=400,
)

ridge_plot_areas.opts(show_legend=False)

```

</details>

<img width="721" alt="image" src="https://github.com/holoviz/hvplot/assets/6613202/dee05632-6a8b-4e5e-adeb-77d503cab99c">

Compare this to the bottom being in foreground (from [here](https://r-graph-gallery.com/294-basic-ridgeline-plot.html)):

<img width="389" alt="image" src="https://github.com/holoviz/holoviews/assets/6613202/7f62bdb7-f111-49d0-bb8f-1b26a4ee3b94">

PR #6194, opened 2024-04-15
subcoordinate_y: reverse the renderers by default

maximlt:
Addresses https://github.com/holoviz/holoviews/issues/6027

By reverting the orders of the subcoordinate-y glyph renderers.

```python
import numpy as np
import holoviews as hv; hv.extension('bokeh')
from scipy.stats import gaussian_kde

categories = ['A', 'B', 'C', 'D', 'E']
data = {cat: np.random.normal(loc=i-2, scale=1.0, size=100) for i, cat in enumerate(categories)}
x = np.linspace(-5, 5, 100)

areas = []
for i, (cat, values) in enumerate(data.items()):
    pdf = gaussian_kde(values)(x)
        
    area = hv.Area((x, pdf), label=cat).opts(
        subcoordinate_y=True, 
        subcoordinate_scale=1.5,
    )
    areas.append(area)

ridge_plot_areas = (hv.VSpan(-4, -3) * hv.Overlay(areas) * hv.VSpan(3, 4)).opts(
    width=900,
    height=400,
)

ridge_plot_areas.opts(show_legend=False)
```

- The blue VSpan is rendered first, that's the regular behavior as it's the first element in the overlay
- The subcoord-y plots are rendered in the reversed order
- The red VSpan is rendered last, that's the regular behavior

image

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -2941,6 +2941,7 @@ class OverlayPlot(GenericOverlayPlot, LegendPlot):
             self._update_ranges(element, ranges)
 
         panels = []
+        subcoord_y_glyph_renderers = []
         for key, subplot in self.subplots.items():
             frame = None
             if self.tabs:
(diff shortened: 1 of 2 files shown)


philippjfr (approved):
Looks great, thanks!

Issue #6192, opened 2024-04-15
Intermittent issue with downsample1d 'm4' algorithm, depending on number of data points


droumis:
#### ALL software version info

<details> <summary> versions </summary>

```bash
Python              :  3.11.8 | packaged by conda-forge | (main, Feb 16 2024, 20:49:36) [Clang 16.0.6 ]
Operating system    :  macOS-13.5.2-arm64-arm-64bit
Panel comms         :  vscode

holoviews           :  1.19.0a0.post28+g4056f6e10.dirty

bokeh               :  3.4.1
colorcet            :  3.1.0
cudf                :  -
dask                :  2024.4.1
datashader          :  0.16.1rc1
geoviews            :  1.12.0
hvplot              :  0.9.3.dev15+g01f6d84
ibis-framework      :  8.0.0
IPython             :  8.23.0
jupyter_bokeh       :  4.0.1
jupyterlab          :  4.1.6
matplotlib          :  3.8.4
networkx            :  3.3
notebook            :  7.1.2
numba               :  0.59.1
numpy               :  1.26.4
pandas              :  2.2.2
panel               :  1.4.1
param               :  2.1.0
...
spatialpandas       :  0.4.10
streamz             :  0.6.4
tsdownsample        :  0.1.3
xarray              :  2023.12.0
```

</details>

#### Description of expected behavior and the observed behavior

Intermittent error with downsample1d 'm4' algorithm, depending on number of data points

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import numpy as np
import pandas as pd
from holoviews.operation.downsample import downsample1d
import holoviews as hv; hv.extension('bokeh')

df = pd.DataFrame(np.random.random(1000), columns=['y'])

plot1 = downsample1d(hv.Curve(df), algorithm='minmax')
plot2 = downsample1d(hv.Curve(df), algorithm='lttb')
plot3 = downsample1d(hv.Curve(df), algorithm='minmax-lttb')
plot4 = downsample1d(hv.Curve(df), algorithm='m4')

plot1 + plot2 + plot3 + plot4
```

#### Stack traceback and/or browser JavaScript console output
```bash
  File "/Users/droumis/src/holoviews/holoviews/plotting/util.py", line 293, in get_plot_frame
    return map_obj[key]
           ~~~~~~~^^^^^
  File "/Users/droumis/src/holoviews/holoviews/core/spaces.py", line 1217, in __getitem__
    val = self._execute_callback(*tuple_key)
[...]

PR #6195, opened 2024-04-16
 Ensure that the downsample algorithm m4 n_out is always a multiple of 4 

Hoxbro:
Fixes https://github.com/holoviz/holoviews/issues/6192

### DIFF
--- a/holoviews/operation/downsample.py
+++ b/holoviews/operation/downsample.py
@@ -188,6 +188,7 @@ def _m4(x, y, n_out, **kwargs):
             'The m4 downsampling algorithm requires the tsdownsampler '
             'library to be installed.'
         ) from None
+    n_out = n_out - (n_out % 4)  # n_out must be a multiple of 4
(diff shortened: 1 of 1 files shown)


holoviews/operation/downsample.py
```diff
@@ -188,6 +188,7 @@ def _m4(x, y, n_out, **kwargs):
             'The m4 downsampling algorithm requires the tsdownsampler '
             'library to be installed.'
         ) from None
+    n_out = n_out - (n_out % 4)  # n_out must be a multiple of 4
```
jbednar, line 191: Wouldn't it be safer to _increase_ the number of samples, to handle degenerate cases where otherwise there might be only a tiny number of samples? Please ignore if I'm misunderstanding this.

Hoxbro: We never downsample something where the number of points is smaller than the width of the plot, so there will never be a tiny number of samples. 

The case reported in the issue `n_out` is around 210 to 217 samples, so adding or subtracting 3 at most for it seemed negligible.

PR #6206, opened 2024-04-23
Optimize rasterization of categorical NdOverlays

philippjfr:
Currently when we rasterize an `NdOverlay` we always concatenate all the data (and insert NaNs between each layer if we are rasterizing lines). This process of concatenation is very expensive and largely pointless if we are generating a categorical aggregate that is coincident with the dimension of the NdOverlay, i.e. each layer in the overlay is a distinct category. 

Therefore this PR adds an optimization for this specific codepath where instead of concatenating and rasterizing everything in a single datashader call, we keep each categories data as a distinct dataframe and make multiple calls to datashader, concatenating all the resulting arrays along the categorical dimension after the fact. This produces the same result without paying the cost.


jbednar:
Nice! Could consider applying the same optimization in the non-categorical case, but I guess that would need to be done by having Datashader accept a list or dict of lines, so that two-stage aggregators like `ds.mean` can be handled.

droumis:
I think I might be confused about the intention of this, but assuming the below is the before/after proposed API, so far the 'after' saves a bit of time by not having to do a groupby, but the rasterizing appears to be slower compared to main.


## Before
```python
curve_dict = {key: hv.Curve(value, kdims=['time'], vdims=['value', 'parameter'], label=key)
              for key, value in df.groupby('parameter')}

overlay = hv.NdOverlay(curve_dict, kdims='parameter')

rasterize(overlay, line_width=1, aggregator=ds.by('parameter', ds.count())).opts(
    responsive=True, min_height=400, show_legend=True, cmap='glasbey', cnorm='eq_hist', colorbar=False, tools=['hover'])

```

## After
```python
curve_dict = {key: hv.Curve(value, kdims=['time'], vdims=['value', 'parameter'], label=key)
              for key, value in df_dict.items()}

overlay = hv.NdOverlay(curve_dict, kdims='parameter')

rasterize(overlay, line_width=1, aggregator=ds.by('parameter', ds.count())).opts(
    responsive=True, min_height=400, show_legend=True, cmap='glasbey', cnorm='eq_hist', colorbar=False, tools=['hover'])
```

droumis:
Got it. It would be nice if there was a convenient way to report timing of the initial and subsequent paint. Hopefully soon! 

While a dict of df's is so much cleaner.. I wonder if the cost of manually inserting nan-separators into a df will be fully amortized over just a couple interaction updates, when using `rasterize(hv.Curve(df))`.

droumis:
This approach may not be feasible for dozens of lines. At some point the resulting imagestack (1 image per line) grows beyond what is reasonably performant

philippjfr:
Going to close, don't think this provides any clear benefits.

PR #6212, opened 2024-04-28
Bug fix for player example pause button

mirage007:
when the pause button is pressed, an exception is thrown that the callback is already removed. This is because it cannot find the the task object. the current code is giving it the wrong object for it to remove. 

To fix, the animate task is stored in a dictionary so that it can be used later and removed as necessary.


Hoxbro:
I forgot to thank you for this PR, sorry about that. 

Thank you!

mirage007:
No problem! Glad to contribute!

PR #6216, opened 2024-05-04
Use the new conda package format

hoxbro:
It seems logical to use the new format for conda packaging; it should be faster, which can't hurt.

https://conda.io/projects/conda-build/en/latest/resources/package-spec.html

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -51,7 +51,7 @@ jobs:
         if: always()
         with:
           name: conda
-          path: dist/*tar.bz2
+          path: dist/*.conda
           if-no-files-found: error
 
   conda_publish:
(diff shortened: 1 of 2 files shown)


maximlt:
Didn't get a chance to post my review but this was briefly discussed live yesterday. It's fine it got merged but I wouldn't push for migrating all the HoloViz pyviz packages to this format until `.tar.bz2` is deprecated, as inspecting a `.conda` file is cumbersome (un-archive the whole thing and then un-archive the two `.tar.zst` files it contains).

hoxbro:
It isn't that cumbersome. It _only_ takes a couple of seconds to do the extra step. And (good) archive managers can look inside nested archives. 

But I won't push it for other packages. I will let that be up to the lead maintainers of the respective project.

maximlt:
> It isn't that cumbersome. It only takes a couple of seconds to do the extra step. And (good) archive managers can look inside nested archives.

For you maybe. I tried on Windows and the default archive manager wouldn't work. Tried on Mac too and after extracting the first archive, somehow extracting `.tar.zst` created a `.tar.zst.cpgz`.

hoxbro:
That's what I meant by _good_. 

I don't think any people will unpack the conda package by themselves, and if they do, they will know how to do it.

philippjfr:
Inspecting a package seems like a pretty niche use case, reserved primarily to developers, i.e. us, so I'd also argue it's not a big deal.

PR #6223, opened 2024-05-08
Trying out pypy

hoxbro:
Just for fun

Also split `recommend` out to its own feature, as IPython does not work in PyPy


hoxbro:
This was just to check if PyPy could work which it could.

PR #6243, opened 2024-05-21
Document subcoordinate_group_ranges by extending the EEG Viewer demo

maximlt:
This makes the example a bit more complicated as it adds another data dimension but I think overall it is still fine, and from what I understand having multiple groups when dealing with EEG data is not unexpected.

image


droumis:
@maximlt please check my changes

maximlt:
@droumis very nice updates thanks! I simply added a redirect link in https://github.com/holoviz/holoviews/pull/6243/commits/3e8629d56f4455aa3f89fa77e0507a8b0d0fe6d7 since the EEG Viewer Bokeh demo was already released (I'm running a doc build on the CI now to check if it builds correctly). I'll merge when everything passes.

PR #6251, opened 2024-05-27
Move environment variables into Pixi.toml

hoxbro:
It seems better to set the environment variables in `pixi.toml` so that the local run mimics CI more. 

Also _sneaking_ in some other changes:
- Remove commented out dash
- Move custom unfixable to unsafe fixes. The main point still stands: I don't want formatters to automatically remove unused variables and imports. Could set pre-commit to unsafe fixes at some point.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -160,7 +160,6 @@ ignore = [
     "RUF005", # Consider {expr} instead of concatenation
     "RUF012", # Mutable class attributes should use `typing.ClassVar`
 ]
-
 extend-unsafe-fixes = [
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -160,7 +160,8 @@ ignore = [
     "RUF005", # Consider {expr} instead of concatenation
     "RUF012", # Mutable class attributes should use `typing.ClassVar`
 ]
-unfixable = [
+
```
hoxbro, line 163:
```suggestion
```

philippjfr (approved):

PR #6253, opened 2024-05-30
Update dependencies

hoxbro:
Updates required dependencies:
Numpy 1.0 --> 1.21
Pandas 0.20.0 --> 1.3
Bokeh (moved up from recommended)

Updates recommended dependencies:
Removed notebook and IPython. 
Moved Bokeh up to required.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,12 +31,13 @@ classifiers = [
 ]
 dependencies = [
     "param >=2.0,<3.0",
-    "numpy >=1.0",
+    "numpy >=1.21",
     "pyviz_comms >=2.1",
     "panel >=1.0",
     "colorcet",
     "packaging",
-    "pandas >=0.20.0",
+    "pandas >=1.3",
+    "bokeh >=3.1",
 ]
 
 [project.urls]
@@ -45,7 +46,7 @@ Source = "https://github.com/holoviz/holoviews"
 HoloViz = "https://holoviz.org/"
 
 [project.optional-dependencies]
-recommended = ["ipython >=5.4.0", "notebook", "matplotlib >=3", "bokeh >=3.1", "plotly >=4.0"]
+recommended = ["matplotlib >=3", "plotly >=4.0"]
 tests = ["pytest", "pytest-rerunfailures"]
 
 [project.scripts]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -30,14 +30,14 @@ classifiers = [
     "Topic :: Software Development :: Libraries",
 ]
 dependencies = [
-    "param >=2.0,<3.0",
-    "numpy >=1.21",
-    "pyviz_comms >=2.1",
-    "panel >=1.0",
+    "bokeh >=3.1",
     "colorcet",
+    "numpy >=1.21",
     "packaging",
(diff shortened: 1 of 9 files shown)


maximlt (commented):
To make it easier to understand the consequences of this change:

```
dependencies = [
    "param >=2.0,<3.0",
    "numpy >=1.0",
    "pyviz_comms >=2.1",
    "panel >=1.0",
    "colorcet",
    "packaging",
    "pandas >=0.20.0",
]
recommended = ["ipython >=5.4.0", "notebook", "matplotlib >=3", "bokeh >=3.1", "plotly >=4.0"]
```

Compared with what's on *conda-forge*:

```
  run:
    - python >=3.9
    - param >=1.12.0,<3.0
    - panel >=1.0
    - numpy >=1.0
    - pandas >=0.20.0
    - matplotlib-base >=3.0
    - bokeh >=3.1
    - colorcet
    - pyviz_comms >=0.7.4
    - packaging
```

And *defaults*:

```
  run:
    - python
    - colorcet
    - numpy >=1.0
    - packaging
    - panel >=1.0
    - pandas >=0.20.0
    - param >=1.12.0,<3.0
    - pyviz_comms >=0.7.4
    - matplotlib-base >=3
```

- I'd vote for aligning the runtime dependencies defined on these three channels.
- I care less about the `pyviz` channel, in fact I hope we could retire it in some way, only `pyviz/label/dev` seems useful these days. We should certainly remove references to the `-c pyviz` channel in the docs. Yet, removing/updating the `recommended` packages is a breaking change for users of this channel, so I'd motivate a change there by the goal of aligning the deps across the channels, rather than the intent to gain a few seconds on the CI.

hoxbro:
> goal of aligning the deps across the channels, rather than the intent to gain a few seconds on the CI.

Users are also affected by this. We pull in a big dependency that we don't use. I just noticed it is because of our CI. 

For alignment, I would propose these as required dependencies.

``` python
bokeh >=3.1
colorcet
numpy >=1.0
packaging
pandas >=0.20.0
panel >=1.0
param >=2.0,<3.0
pyviz_comms >=2.1
```

Maybe even bump the lower pin of pandas and numpy to something released in the last 3-5 years, or even remove them as they are so low it doesn't matter. 

(numpy 1.0 was released 25th October 2006, pandas 0.20.0 was released 5th May 2017)

pyproject.toml
```diff
@@ -30,13 +30,14 @@ classifiers = [
...
-    "param >=2.0,<3.0",
-    "numpy >=1.0",
-    "pyviz_comms >=2.1",
-    "panel >=1.0",
+    "bokeh >=3.1",
     "colorcet",
+    "numpy >=1.21",
     "packaging",
-    "pandas >=0.20.0",
+    "pandas >=1.3",
```
hoxbro, line 37: Pandas and numpy versions were changed to the first minor version with a micro release after Python 3.10 was released.

maximlt (approved):
LGTM. You may want to list the versions bumped and packages removed from `recommended` in the original post, to make it easier to write the release notes.

philippjfr (approved):
I'm fully on board with this.

Issue #5723, opened 2023-05-23
Add cticks (colorbar ticks) to opts


ahuang11:
There's support for `xticks` and `yticks`; would be nice to have `cticks` supported too, without having to use colorbar_opts.

Also, a parallel is there's xlabel, ylabel, clabel.

For example:

```
df.hvplot.paths(
    "RMM1",
    "RMM2",
    color="month",
).opts(
    clabel="Month",
    cmap="RdYlBu_r",
    colorbar=True,
    color_levels=4,
    colorbar_opts={
        "ticker": FixedTicker(ticks=sorted(df["month"].unique())),
    }
)
```

Would ideally be (supports lists instead of using FixedTicker)
```
df.hvplot.paths(
    "RMM1",
    "RMM2",
    color="month",
).opts(
    clabel="Month",
    cmap="RdYlBu_r",
    colorbar=True,
    color_levels=4,
    cticks=sorted(df["month"].unique())
)
```

PR #6257, opened 2024-06-04
Add support for cticks

ahuang11:
Closes https://github.com/holoviz/holoviews/issues/5723
```python
import numpy as np
import holoviews as hv

hv.extension("bokeh")

img = hv.Image(np.array([[0, 1], [2, 3]])).opts(
    cticks=[(1, "A"), (2, "B")], colorbar=True
)
img
```

image

Didn't implement for matplotlib because it already has `cbar_ticks`


holoviews/plotting/bokeh/element.py
```diff
@@ -1051,28 +1051,8 @@ def _axis_properties(self, axis, key, plot, dimension=None,
...
-                else:
-                    ticks, labels = ticker, None
-                if ticks and util.isdatetime(ticks[0]):
-                    ticks = [util.dt_to_int(tick, 'ms') for tick in ticks]
-                axis_props['ticker'] = FixedTicker(ticks=ticks)
-                if labels is not None:
-                    axis_props['major_label_overrides'] = dict(zip(ticks, labels))
+            if not (self._subcoord_overlaid and axis == 'y'):
+                axis_props.update(get_ticker_axis_props(ticker))
             elif self._subcoord_overlaid and axis == 'y':
```
hoxbro, line 1056:
```suggestion
            else:
```

Or move the if-part down to the else.

holoviews/plotting/bokeh/element.py
```diff
@@ -2470,6 +2450,11 @@ class ColorbarPlot(ElementPlot):
         #FFFFFFFF or a length 3 or length 4 tuple specifying values in
         the range 0-1 or a named HTML color.""")
 
+    cticks = param.Parameter(default=None, doc="""
```
hoxbro, line 2453: Can we use `param.ClassSelector` here?

Even if it is verbose, it makes it easier to understand what is permitted.  You are also welcome to update `xticks` and `yticks` if they are also parameters.

holoviews/plotting/bokeh/util.py
```diff
@@ -1167,3 +1173,30 @@ def dtype_fix_hook(plot, element):
                 for k, v in data.items():
                     if hasattr(v, "dtype") and v.dtype.kind == "U":
                         data[k] = v.tolist()
+
+
+def get_ticker_axis_props(ticker):
```
hoxbro, line 1178: I haven't checked, but this is just an extraction, correct?

holoviews/plotting/bokeh/element.py
ahuang11: :| Very verbose indeed.

holoviews/plotting/plot.py
```diff
@@ -1132,15 +1132,13 @@ class GenericElementPlot(DimensionedPlot):
     yrotation = param.Integer(default=None, bounds=(0, 360), doc="""
         Rotation angle of the yticks.""")
 
-    xticks = param.Parameter(default=None, doc="""
+    xticks = param.ClassSelector(class_=(int, list, tuple, np.ndarray), default=None, doc="""
```
ahuang11, line 1135: Not specific to Bokeh.

Issue #3785, opened 2019-06-19
Vectorfield broken hover tooltips


ahuang11:
`hv.VectorField(([-1, 0, 1, 2, 3], [-1, 0, 1, 2, 3], np.deg2rad([0, 45, 90, 180, 360]))).opts(tools=['hover'])`

eldad-a:
> `hv.VectorField(([-1, 0, 1, 2, 3], [-1, 0, 1, 2, 3], np.deg2rad([0, 45, 90, 180, 360]))).opts(tools=['hover'])`

@ahuang11 
Were you able to resolve this issue?
If you have found a solution, it would be great if you could share.

In the meantime, here's a work-around, in case it is useful for anyone.
It may require more thought in different cases though...

```python
import numpy as np
import holoviews as hv

hv.extension('bokeh')

# data
x, y = [-1, 0, 1, 2, 3], [-1, 0, 1, 2, 3]
angle = np.deg2rad([0, 45, 90, 180, 360])

# holoviews
field = hv.VectorField((x,y, angle))
heatmap = hv.HeatMap((x,y, angle), vdims=['angle']) # a temporary work-around for VectorField hover 

# viz
field * heatmap.opts(alpha=0, tools=['hover']) # alpha set to zero to hide the heatmap
```

TheoMathurin:
I resorted to a similar workaround as I was unable to make the tooltips work, except for x and y with a custom Bokeh HoverTool.

basnijholt:
@philippjfr or @jlstevens, do you guys know where to look if I would like to fix this properly? The proposed workaround doesn't work well either.

dwr-psandhu:
Still an issue. Any progress?

PR #6258, opened 2024-06-04
fix vectorfield hover

ahuang11:
```
import numpy as np
import holoviews as hv

hv.extension("bokeh")

hv.VectorField(
    ([-1, 0, 1, 2, 3], [-1, 0, 1, 2, 3], np.deg2rad([0, 45, 90, 180, 360]))
).opts(tools=["hover"])
```
Not sure there's a better solution, but I think it's better than no solution because hover is useful:

image

Closes https://github.com/holoviz/holoviews/issues/3785

### DIFF
--- a/holoviews/tests/plotting/bokeh/test_vectorfieldplot.py
+++ b/holoviews/tests/plotting/bokeh/test_vectorfieldplot.py
@@ -95,6 +95,6 @@ class TestVectorFieldPlot(TestBokehPlot):
         vectorfield = VectorField([(0, 0, 0), (0, 1, 1), (0, 2, 2)],
                         vdims='color').opts(tools=["hover"])
         plot = bokeh_renderer.get_plot(vectorfield)
-        keys = plot.handles["cds"].data.keys()
+        keys = sorted(plot.handles["cds"].data)
         assert len(keys) == 8
         assert keys == ["Angle", "Magnitude", "x", "x0", "x1", "y", "y0", "y1"]
--- a/holoviews/tests/plotting/bokeh/test_vectorfieldplot.py
+++ b/holoviews/tests/plotting/bokeh/test_vectorfieldplot.py
@@ -83,3 +83,29 @@ class TestVectorFieldPlot(TestBokehPlot):
             "for 'line_color' option and declare a color_index; ignoring the color_index.\n"
         )
         self.assertEqual(log_msg, warning)
+
+    def test_vectorfield_no_hover_columns(self):
+        vectorfield = VectorField([(0, 0, 0), (0, 1, 1), (0, 2, 2)],
+                        vdims='color').opts(tools=[])
+        plot = bokeh_renderer.get_plot(vectorfield)
+        keys = plot.handles["cds"].data.keys()
+        assert len(keys) == 4
+        assert "x0" in keys
+        assert "y0" in keys
+        assert "x1" in keys
+        assert "y1" in keys
+
+    def test_vectorfield_hover_columns(self):
+        vectorfield = VectorField([(0, 0, 0), (0, 1, 1), (0, 2, 2)],
+                        vdims='color').opts(tools=["hover"])
+        plot = bokeh_renderer.get_plot(vectorfield)
+        keys = plot.handles["cds"].data.keys()
+        assert len(keys) == 8
+        assert "x0" in keys
+        assert "y0" in keys
+        assert "x1" in keys
+        assert "y1" in keys
+        assert "x" in keys
+        assert "y" in keys
+        assert "Angle" in keys
+        assert "Magnitude" in keys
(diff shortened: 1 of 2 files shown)


hoxbro:
I'm not a fan of this solution. The main reason is you increase the data by a factor of 4.  

If we choose to go this way, we should at least limit this only to be done when we have an actual hover tool. By the looks of it, you send all the data to the front end no matter what.

holoviews/tests/plotting/bokeh/test_vectorfieldplot.py
```diff
@@ -83,3 +83,29 @@ def test_vectorfield_color_index_color_clash(self):
...
+    def test_vectorfield_no_hover_columns(self):
+        vectorfield = VectorField([(0, 0, 0), (0, 1, 1), (0, 2, 2)],
+                        vdims='color').opts(tools=[])
+        plot = bokeh_renderer.get_plot(vectorfield)
+        keys = plot.handles["cds"].data.keys()
+        assert len(keys) == 4
+        assert "x0" in keys
+        assert "y0" in keys
+        assert "x1" in keys
+        assert "y1" in keys
```
hoxbro, line 96:
```suggestion
    def test_vectorfield_no_hover_columns(self):
        vectorfield = VectorField([(0, 0, 0), (0, 1, 1), (0, 2, 2)],
                        vdims='color').opts(tools=[])
        plot = bokeh_renderer.get_plot(vectorfield)
        keys = sorted(plot.handles["cds"].data)
        assert keys == ["x0", "x1", "y0", "y1"]
```

It could also make these two functions a single parameterized test, but I don't have a strong opinion about that.

ahuang11: I'd prefer not.

hoxbro: Both?

ahuang11: What's both?

hoxbro: 1) My code suggestion, with compacts the code
2) parameterize the two functions.

holoviews/tests/plotting/bokeh/test_vectorfieldplot.py
```diff
@@ -83,3 +83,25 @@ def test_vectorfield_color_index_color_clash(self):
...
+        keys = plot.handles["cds"].data.keys()
+        assert len(keys) == 8
+        assert "x0" in keys
+        assert "y0" in keys
+        assert "x1" in keys
+        assert "y1" in keys
+        assert "x" in keys
+        assert "y" in keys
+        assert "Angle" in keys
+        assert "Magnitude" in keys
```
ahuang11, line 107:
```suggestion
        assert keys == ["Angle", "Magnitude", "x", "x0", "x1", "y", "y0", "y1"]
```

holoviews/tests/plotting/bokeh/test_vectorfieldplot.py
```diff
@@ -83,3 +83,18 @@ def test_vectorfield_color_index_color_clash(self):
...
+                        vdims='color').opts(tools=[])
+        plot = bokeh_renderer.get_plot(vectorfield)
+        keys = sorted(plot.handles["cds"].data)
+        assert keys == ["x0", "x1", "y0", "y1"]
+
+    def test_vectorfield_hover_columns(self):
+        vectorfield = VectorField([(0, 0, 0), (0, 1, 1), (0, 2, 2)],
+                        vdims='color').opts(tools=["hover"])
+        plot = bokeh_renderer.get_plot(vectorfield)
+        keys = plot.handles["cds"].data.keys()
```
ahuang11, line 98:
```suggestion
        keys = sorted(plot.handles["cds"].data.keys())
```

holoviews/tests/plotting/bokeh/test_vectorfieldplot.py
```diff
@@ -83,3 +83,18 @@ def test_vectorfield_color_index_color_clash(self):
...
+        plot = bokeh_renderer.get_plot(vectorfield)
+        keys = sorted(plot.handles["cds"].data)
+        assert keys == ["x0", "x1", "y0", "y1"]
+
+    def test_vectorfield_hover_columns(self):
+        vectorfield = VectorField([(0, 0, 0), (0, 1, 1), (0, 2, 2)],
+                        vdims='color').opts(tools=["hover"])
+        plot = bokeh_renderer.get_plot(vectorfield)
+        keys = sorted(plot.handles["cds"].data)
+        assert len(keys) == 8
```
ahuang11, line 99:
```suggestion
```

holoviews/tests/plotting/bokeh/test_vectorfieldplot.py
ahuang11: Oh :) I applied 1

philippjfr (approved):

Issue #6260, opened 2024-06-05
Proposal for improving support for wide data


philippjfr:
From the beginning HoloViews was designed primarily around [tidy data](https://cran.r-project.org/web/packages/tidyr/vignettes/tidy-data.html). This has the major benefit that data can clearly be delineated into key dimensions (or independent values / coordinates) and value dimensions, which represent a dependent variable, i.e. some kind of measurement. Additionally it makes it possible to easily perform the groupby operations that allow HoloViews to easily facet data in a grid (GridSpace), layout (NdLayout), using widgets (HoloMap/DynamicMap) and as a set of trace in a plot (NdOverlay). However in many common scenarios data will not be tidy, the most common of which is when you are storing a bunch of timeseries indexed by the date(time) and then store multiple measurements all representing the kind of value, e.g. the most common example is stock prices where the index is the date and each column records the stock price for a different ticker.

The problem with reshaping this data is that it's tremendously inefficient. Where before you could have one DataFrame you now have to create `N` DataFrames, one for each stock ticker. So here I will lay out my proposal for formally supporting wide data in HoloViews.

### The Problem

While today you can already organize data in such a way that you create an NdOverlay where each Element provides a view into one column in the wide DataFrame, it breaks HoloViews' internal model of the world. E.g. let's look at what the structure of the ticker data looks like if you do this:

```python
NdOverlay [ticker]
    Curve [datetime] (AAPL)
    Curve [datetime] (MSFT)
    Curve [datetime] (IBM)
```

Here the ticker names now become the values of the NdOverlay key dimension AND they are the value dimension names of each `Curve` elements. This is clearly inelegant and also conceptually not correct, i.e. AAPL is not a dimension, it does not represent some actual measurable quantity with some associated unit. The actual measurable quantity is "Stock Price". The reason this is necessary is because the element equates the value dimension with the name of the variable in the underlying data, i.e. the string 'AAPL' will be used to look up the column in the underlying DataFrame. Downstream this causes issues for the sharing of dimension ranges in plots and other features that rely on the identity of Dimensions.

### The proposal

[...]

PR #6262, opened 2024-06-05
Make Dimension.label source of truth for Dimension identity

philippjfr:
Implements https://github.com/holoviz/holoviews/issues/6260

### DIFF
--- a/holoviews/plotting/renderer.py
+++ b/holoviews/plotting/renderer.py
@@ -433,8 +433,6 @@ class Renderer(Exporter):
             from IPython.display import display
             display(data, raw=True)
             return {'text/html': '<div style="display: none"></div>'}, {}
-        elif config.comms == 'colab':
-            load_notebook(config.inline)
         return data, {}
 
     def static_html(self, obj, fmt=None, template=None):
--- a/holoviews/plotting/renderer.py
+++ b/holoviews/plotting/renderer.py
@@ -433,6 +433,8 @@ class Renderer(Exporter):
             from IPython.display import display
             display(data, raw=True)
             return {'text/html': '<div style="display: none"></div>'}, {}
+        elif config.comms == 'colab':
+            load_notebook(config.inline)
         return data, {}
 
     def static_html(self, obj, fmt=None, template=None):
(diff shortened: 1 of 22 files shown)


holoviews/plotting/renderer.py
```diff
@@ -433,6 +433,8 @@ def _render_ipywidget(self, plot):
             from IPython.display import display
             display(data, raw=True)
             return {'text/html': '<div style="display: none"></div>'}, {}
+        elif config.comms == 'colab':
+            load_notebook(config.inline)
```
hoxbro, line 437: This seems to be unrelated to this PR :upside_down_face:

philippjfr: Indeed :)

philippjfr, line 437:
```suggestion
```

hoxbro (commented):
My rough understanding is that `dim.name` will be used for data access, e.g., looking up a column in a DataFrame and `dim.label` will be used by HoloViews for ranges etc. Is this an ok (rough) understanding?

How thoroughly have you checked the codebase for this find and replace? The code still seems to use `name` for range lookup. 

https://github.com/holoviz/holoviews/blob/c8e7f5f887f05e74e6a8999c990dbd41265a086b/holoviews/plotting/util.py#L411-L411
https://github.com/holoviz/holoviews/blob/c8e7f5f887f05e74e6a8999c990dbd41265a086b/holoviews/operation/normalization.py#L109-L109

Have you checked if `hvplot` / `GeoViews` need updating too?

philippjfr:
>Is this an ok (rough) understanding?

Yes.

>How thoroughly have you checked the codebase for this find and replace? The code still seems to use name for range lookup.

Clearly not quite thoroughly enough, will do another more exhaustive check now.

>Have you checked if hvplot / GeoViews need updating too?

Yes.

philippjfr:
Did miss quite a few. Will scan more thoroughly one more time. Not that this only affects plotting code, so e.g. this isn't affected:

```
holoviews/holoviews/operation/normalization.py

Line 109 in c8e7f5f

 return {d.name: element.range(d.name, self.data_range) 
```

philippjfr:
Okay, quite confident now, thoroughly reviewed the plotting code and replaced all remaining cases.

philippjfr:
I think the only way we'll catch any more cases will be if we merge and include the change in the RC release.

hoxbro (approved):

Issue #6252, opened 2024-05-29
Group or hide hover tools in Bokeh toolbar


droumis:
#### Is your feature request related to a problem? Please describe.
Too many hover tools when overlaying many elements with custom hover.

#### Describe the solution you'd like
Group or hide the hover tooltips

#### Additional context
<details> <summary> Code </summary>

```python
import numpy as np
import pandas as pd
from datetime import datetime, timedelta

import holoviews as hv; hv.extension('bokeh')
import panel as pn; pn.extension()

amp_dim = hv.Dimension("amplitude", unit="µV")
time_dim = hv.Dimension("time", unit="ms")

n_channels = 10
n_seconds = 5
total_samples = 256*n_seconds
start_datetime = datetime(2024, 1, 1)
time = np.array([start_datetime + timedelta(seconds=t) for t in np.linspace(0, n_seconds, total_samples)])

data = np.random.randn(n_channels, total_samples).cumsum(axis=1)
channels = [f"EEG {i}" for i in range(n_channels)]

df = pd.DataFrame(data.T, index=time, columns=channels)

hover_tooltips=[
    ("type", "$group"),
    ("channel", "$label"),
    ("time", '@time{%H:%M:%S.%3N}'),
    ("amplitude"),
]

curves = {}
for channel_name, channel_data in df.items():
    ds = hv.Dataset((time, channel_data, channel), [time_dim, amp_dim, "channel"])
    curve = hv.Curve(ds, time_dim, [amp_dim, "channel"], label=channel_name, group='EEG')
    curve.opts(color="black", line_width=1, subcoordinate_y=True, subcoordinate_scale=3,
              hover_tooltips = hover_tooltips)
    curves[channel_name] = curve

curves_overlay = hv.Overlay(curves, kdims="channel").opts(padding=0, aspect=2, responsive=True,show_legend=False)

curves_overlay
```

</details>

![image](https://github.com/holoviz/holoviews/assets/6613202/c27bdc2c-a345-476d-bb9b-a21ec1b53716)

droumis:
@mattpap , this could be solved with the ability to group tools which is sounds like you are working on.

mattpap:
Yes. The relevant bokeh issue is https://github.com/bokeh/bokeh/issues/5497 (a very old one).

droumis:
observation: if I create a custom bokeh HoverTool, and pass that in like .opts(tools=[HoverTool()] then the hover tools are combined in the toolbar.

[...]

PR #6264, opened 2024-06-06
Fix imports for the multichan ts viewer demo

droumis:


maximlt (commented):
Could you separate the imports from the code and put them above `Generating Data`, together with `hv.extension('bokeh')`. This seems to be like the other demos are structured, e.g.:

image

---

I re-ran the code for good measure and noticed there's a hover tool icon per curve in the toolbar. I don't remember seeing that before. I assume this is due to using the new `hover_tooltips` option? I guess it's not exactly the expected outcome.

image

droumis:
> I re-ran the code for good measure and noticed there's a hover tool icon per curve in the toolbar. I don't remember seeing that before. I assume this is due to using the new hover_tooltips option? I guess it's not exactly the expected outcome.

Correct, I had opened an issue [here](https://github.com/holoviz/holoviews/issues/6252). At documented in that thread, the hover tools do get combined when it becomes a layout rather than an overlay.

droumis:
using hv.Layout((curves_overlay, hv.Empty())) actually combines the hover tools, but I don't think we should be advertising this workaround

droumis:
should be ready to merge. I'm guessing that the merging of the "subplot" tools within an overlay will be addressed another time, and otherwise should probably be default behavior for overlays, so there would be nothing to change further in this example.

droumis:
In future versions, we should update to utilize the wide df mapper approach and NdOverlay instead of hv.Dataset and Overlay. But some issues need to be addressed when combining data from multiple groups with the wide df approach, so let's merge this for now.

PR #6247, opened 2024-05-24
Hold streams together for faster plot callbacks

ahuang11:
Philipp suggested this; not sure if the following is necessary:
```
                if self.plot.comm:
                    push_on_root(self.plot.root.ref['id'])
```

PR #6265, opened 2024-06-06
Hold rendering until a plot has been fully updated

philippjfr:
Supersedes https://github.com/holoviz/holoviews/pull/6247 by always holding rendering until a plot has been fully updated.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -81,6 +81,7 @@ from .util import (
     get_ticker_axis_props,
     glyph_order,
     hold_policy,
+    hold_render,
     match_ax_type,
     match_dim_specs,
     match_yaxis_type_to_range,
@@ -2228,6 +2229,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
                         if isinstance(s, RangeXY) and not s._triggering:
                             s.reset()
 
+    @hold_render()
     def update_frame(self, key, ranges=None, plot=None, element=None):
         """
         Updates an existing plot with data corresponding
@@ -3379,6 +3381,7 @@ class OverlayPlot(GenericOverlayPlot, LegendPlot):
 
         return self.handles['plot']
 
+    @hold_render()
     def update_frame(self, key, ranges=None, element=None):
         """
         Update the internal state of the Plot to represent the given
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -2229,7 +2229,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
                         if isinstance(s, RangeXY) and not s._triggering:
                             s.reset()
 
-    @hold_render()
+    @hold_render
     def update_frame(self, key, ranges=None, plot=None, element=None):
         """
         Updates an existing plot with data corresponding
@@ -3381,7 +3381,7 @@ class OverlayPlot(GenericOverlayPlot, LegendPlot):
 
         return self.handles['plot']
 
-    @hold_render()
+    @hold_render
     def update_frame(self, key, ranges=None, element=None):
         """
         Update the internal state of the Plot to represent the given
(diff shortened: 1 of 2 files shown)


holoviews/plotting/bokeh/element.py
```diff
@@ -2228,6 +2229,7 @@ def _reset_ranges(self):
                         if isinstance(s, RangeXY) and not s._triggering:
                             s.reset()
 
+    @hold_render()
```
philippjfr, line 2232:
```suggestion
    @hold_render
```

holoviews/plotting/bokeh/element.py
```diff
@@ -3379,6 +3381,7 @@ def initialize_plot(self, ranges=None, plot=None, plots=None):
 
         return self.handles['plot']
 
+    @hold_render()
```
philippjfr, line 3384:
```suggestion
    @hold_render
```

holoviews/plotting/bokeh/util.py
```diff
@@ -856,6 +856,20 @@ def wrapper(self, *args, **kwargs):
...
+    Decorator that will hold render on a Bokeh ElementPlot until after
+    the method has been called.
+    """
+    def wrapper(self, *args, **kwargs):
+        hold = self.state.hold_render
+        self.state.hold_render = True
+        try:
+            f(self, *args, **kwargs)
+        finally:
+            self.state.hold_render = hold
```
philippjfr, line 870:
```suggestion
            self.state.hold_render = hold
        return wrapper
```

holoviews/plotting/bokeh/util.py
```diff
@@ -856,6 +856,21 @@ def wrapper(self, *args, **kwargs):
...
+        return wrapper
```
hoxbro, line 871:
```suggestion
    return wrapper
```

holoviews/plotting/bokeh/util.py
```diff
@@ -856,6 +856,21 @@ def wrapper(self, *args, **kwargs):
...
+def hold_render(f):
+    """
+    Decorator that will hold render on a Bokeh ElementPlot until after
+    the method has been called.
+    """
+    def wrapper(self, *args, **kwargs):
+        hold = self.state.hold_render
+        self.state.hold_render = True
+        try:
+            f(self, *args, **kwargs)
```
hoxbro, line 868:
```suggestion
            return f(self, *args, **kwargs)
```

holoviews/plotting/bokeh/util.py
```diff
@@ -856,6 +856,21 @@ def wrapper(self, *args, **kwargs):
     return wrapper
 
 
+def hold_render(f):
+    """
+    Decorator that will hold render on a Bokeh ElementPlot until after
+    the method has been called.
+    """
+    def wrapper(self, *args, **kwargs):
+        hold = self.state.hold_render
```
hoxbro, line 865: This could lead to a race condition. If a trigger happens when another trigger is running. 

A()  -> starting -> `hold = False` changes `self.state.hold_render = True`
B() -> starting  -> `hold = True` changes `self.state.hold_render = True`
A() -> Finishes -> changes `self.state.hold_render = False`
A() -> Finishes -> changes `self.state.hold_render = True`

Could we use an integer and add and subtract here?

hoxbro: > will therefore always restore `hold_render=False`

It will restore to what `hold` was when the function started. In my example, `B` would have `hold = True` because `A` had set it so. 

> It's not like there's anything async happening

We are getting closer and closer to a no-gil Python world. 

Maybe I'm just too cautious here or not understanding the logic well enough :upside_down_face:

philippjfr: Yeah, I don't think there's an issue here. Certainly none of our code would be ready to perform plot updates on different threads.

hoxbro (approved):

PR #6271, opened 2024-06-10
Ensure downsample works with non-contiguous arrays

philippjfr:
Many of the downsampling algorithms do not work with non-contiguous arrays. Therefore we ensure they are contiguous before passing them to the downsampling operation.

### DIFF
--- a/holoviews/operation/downsample.py
+++ b/holoviews/operation/downsample.py
@@ -100,6 +100,13 @@ def _lttb_inner(x, y, n_out, sampled_x, offset):
     )
 
 
+def _ensure_contiguous(x, y):
+    """
+    Ensures the arrays are contiguous in memory (required by tsdownsample).
+    """
+    return np.ascontiguousarray(x), np.ascontiguousarray(y)
+
+
 def _lttb(x, y, n_out, **kwargs):
     """
     Downsample the data using the LTTB algorithm.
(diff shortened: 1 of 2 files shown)


holoviews/operation/downsample.py
```diff
@@ -256,7 +256,7 @@ def _process(self, element, key=None, shared_data=None):
 
         if len(element) <= self.p.width:
             return element
-        xs, ys = (element.dimension_values(i) for i in range(2))
+        xs, ys = (np.ascontiguousarray(element.dimension_values(i)) for i in range(2))
```
hoxbro, line 259: this could potentially be a heavy computation, if this is only needed for the `tsdownsample` algorithms I would move this up into those functions.

holoviews/tests/operation/test_downsample.py
```diff
@@ -20,16 +21,26 @@ def test_downsample1d_multi(plottype):
     assert N > downsample1d.width
 
     if plottype == "overlay":
-        figure = hv.Overlay([hv.Curve(range(N)), hv.Curve(range(N))])
+        figure = Overlay([Curve(range(N)), Curve(range(N))])
     elif plottype == "ndoverlay":
-        figure = hv.NdOverlay({"A": hv.Curve(range(N)), "B": hv.Curve(range(N))})
+        figure = NdOverlay({"A": Curve(range(N)), "B": Curve(range(N))})
```
hoxbro, line 26: I have never been the fan of the way we write our test and the way we write our code differs. Makes it harder to copy/paste either from or to a notebook. 

Though, I'm curious is there any other reason for this change other than to align them with the rest?

philippjfr: Nope, just aligning with the style elsewhere.

PR #6272, opened 2024-06-10
Disable batching when subcoordinate_y is enabled

philippjfr:
Batching kicks in when you are plotting more than `legend_limit` elements inside an `NdOverlay`. However for a `subcoordinate_y` plot batching is not and cannot be supported since each element has to be plotted using a different renderer instance. Therefore we check if `subcoordinate_y` is enabled before enabling batching.

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -2113,7 +2113,6 @@ class ElementPlot(BokehPlot, GenericElementPlot):
             self.handles['xaxis'], self.handles['yaxis'] = axes
             self.handles['x_range'], self.handles['y_range'] = plot_ranges
             if self._subcoord_overlaid:
-
                 if style_element.label in plot.extra_y_ranges:
                     self.handles['subcoordinate_y_range'] = plot.y_range
                     self.handles['y_range'] = plot.extra_y_ranges.pop(style_element.label)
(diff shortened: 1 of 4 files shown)


holoviews/plotting/plot.py
```diff
@@ -1797,10 +1797,14 @@ def _apply_compositor(self, holomap, ranges=None, keys=None, dimensions=None):
...
 
+    @property
+    def _is_batched(self):
+        return self.batched and type(self.hmap.last) is NdOverlay
+
     def _create_subplots(self, ranges):
         # Check if plot should be batched
         ordering = util.layer_sort(self.hmap)
-        batched = self.batched and type(self.hmap.last) is NdOverlay
+        batched = self._is_batched
```
hoxbro, line 1807: Nit, but maybe just combine this with the next line.

hoxbro (commented):
Any way to test this?

holoviews/plotting/bokeh/element.py
```diff
@@ -2113,6 +2113,7 @@ def initialize_plot(self, ranges=None, plot=None, plots=None, source=None):
             self.handles['xaxis'], self.handles['yaxis'] = axes
             self.handles['x_range'], self.handles['y_range'] = plot_ranges
             if self._subcoord_overlaid:
+
```
philippjfr, line 2116:
```suggestion
```

PR #6274, opened 2024-06-10
Fix issue with range lookups in (y)-axis creation

philippjfr:
An issue introduced by https://github.com/holoviz/holoviews/pull/6262 where the dimension was not correctly looked up.

Also makes a change to more robustly look up overlay dimensions on subplots (sorry for sneaking that in here).


philippjfr:
Test would be nice. Will take a look shortly.

Issue #5658, opened 2023-03-12
`color` option not working for Bars plot with Plotly backend


MarcSkovMadsen:
First reported for hvPlot here https://github.com/holoviz/hvplot/issues/1037.

I would expect the bars in the below plot to have the `gold` color. It has if I select the `bokeh` extension.

![image](https://user-images.githubusercontent.com/42288570/224564448-849acfb8-8f49-44f2-bc53-a4f954721d92.png)

```python
import pandas as pd
import holoviews as hv
hv.extension("plotly")

data = pd.DataFrame({
    "framework": ["hvPlot", "HoloViews", "Panel"], 
    "stars": [700, 2400, 2600]
})
plot = hv.Bars(data, kdims="framework", vdims="stars").opts(color="gold", ylabel="⭐", width=1000)

import panel as pn
pn.extension()
pn.panel(plot).servable()
```

## Additional Context

The hvPlot issue also https://github.com/holoviz/hvplot/issues/1037 shows that `color` does not work if set to a list of colors or the name of a color of values. Both work for the Bokeh backend.

## Versions

holoviews=1.15.4, plotly=5.11.0

Hoxbro:
This seems to be more related to `plotly` not directly supporting `color=red`. E.g. this won't work:
``` python
import plotly.graph_objects as go

fig = go.Figure()

x = [1, 2, 3]
y = [4, 5, 6]
z = [12, 24, 48]

fig.add_trace(go.Bar(x=x, y=y, color="red"))
```

And something like this is needed: `fig.add_trace(go.Bar(x=x, y=y, marker=dict(color=x, colorscale=["red"] * 3)))`

![image](https://user-images.githubusercontent.com/19758978/231851538-02f245aa-1617-4b2c-9f68-b188c4e0944b.png)

adivekar-utexas:
As it is the 1-year anniversary of this issue, is there any plan to resolve it?

PR #6294, opened 2024-06-24
Add color to bars for Plotly

hoxbro:
Fixes https://github.com/holoviz/holoviews/issues/5658

image

### DIFF
--- a/holoviews/plotting/plotly/chart.py
+++ b/holoviews/plotting/plotly/chart.py
@@ -198,9 +198,7 @@ class BarPlot(BarsMixin, ElementPlot):
     show_legend = param.Boolean(default=True, doc="""
         Whether to show legend for the plot.""")
 
-    color = param.String(default=None, doc="The color of the bars.")
-
-    style_opts = ['visible']
+    style_opts = ['visible', 'color']
 
     selection_display = PlotlyOverlaySelectionDisplay()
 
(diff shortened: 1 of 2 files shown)


holoviews/plotting/plotly/chart.py
```diff
@@ -280,6 +281,10 @@ def get_data(self, element, ranges, style, **kwargs):
                     for d in (xdim, group_dim)],
                 y: np.nan_to_num(values)})
 
+        if self.color:
```
hoxbro, line 284: Is there a better method to put this in?

maximlt (commented):
I don't think making `color` a plot option is the right approach here. I've tried something else that seemed to work, it'd need some more testing and unit tests.

I'm also not sure what convention HoloViews follows on naming style options, does it attempt to standardize them across backends or does it instead try to stay closer to the backend naming scheme?

```diff
diff --git a/holoviews/plotting/plotly/chart.py b/holoviews/plotting/plotly/chart.py
index a7a831bbd..bcb7c82ee 100644
--- a/holoviews/plotting/plotly/chart.py
+++ b/holoviews/plotting/plotly/chart.py
@@ -198,9 +198,7 @@ class BarPlot(BarsMixin, ElementPlot):
     show_legend = param.Boolean(default=True, doc="""
         Whether to show legend for the plot.""")
 
-    color = param.String(default=None, doc="The color of the bars.")
-
-    style_opts = ['visible']
+    style_opts = ['visible', 'color']
 
     selection_display = PlotlyOverlaySelectionDisplay()
 
@@ -281,12 +279,14 @@ class BarPlot(BarsMixin, ElementPlot):
                     for d in (xdim, group_dim)],
                 y: np.nan_to_num(values)})
 
-        if self.color:
-            for bar in bars:
-                bar['marker_color'] = self.color
-
         return bars
 
+    def graph_options(self, element, ranges, style, **kwargs):
+        if 'color' in style:
+            style['marker_color'] = style.pop('color')
+        opts = super().graph_options(element, ranges, style, **kwargs)
+        return opts
+
     def init_layout(self, key, element, ranges, **kwargs):
         layout = super().init_layout(key, element, ranges)
         stack_dim = None
```

image

maximlt:
> it'd need some more testing and unit tests.

@hoxbro in fact you already added a test which I didn't need to touch, so we're good.

Issue #6298, opened 2024-06-28
Pandas interface doesn't handle kdims part of a Pandas MultiIndex


maximlt:
```python
import numpy as np
import pandas as pd
import holoviews as hv

hv.extension('bokeh')

dti = pd.date_range(start='1/1/2018', end='1/08/2018')

other = list(range(len(dti)))

t = pd.DataFrame(
    dict(
        n=range(5, len(dti) + 5),
        m=range(10, len(dti) + 10),
    )
)
dfm = pd.MultiIndex.from_frame(t)
df = pd.DataFrame(np.random.randn(len(dfm)), index=dfm, columns=['test'])
p = hv.Curve(df, 'm', 'test')
p
```

The plot correctly has the `m` x-axis but the values are off.

![image](https://github.com/holoviz/holoviews/assets/35924738/e0a6aebd-415f-4d3b-800b-215c7c496872)

And indeed they have been ignored in favor of a default "range-index" column that was created.

![image](https://github.com/holoviz/holoviews/assets/35924738/f209ac64-f334-4bd7-8fc2-6cc09f55e36b)

PR #6299, opened 2024-06-28
Ensure no auto index is created when a kdim is part of a multi index

maximlt:
Fixes https://github.com/holoviz/holoviews/issues/6298

<details><summary>Code</summary>
<p>

```python
import numpy as np
import pandas as pd
import holoviews as hv

hv.extension('bokeh')

dti = pd.date_range(start='1/1/2018', end='1/08/2018')

other = list(range(len(dti)))

t = pd.DataFrame(
    dict(
        n=range(5, len(dti) + 5),
        m=range(10, len(dti) + 10),
    )
)
dfm = pd.MultiIndex.from_frame(t)
df = pd.DataFrame(np.random.randn(len(dfm)), index=dfm, columns=['test'])
df.head()

p = hv.Curve(df, 'm', 'test')
p
```

</p>
</details> 

image

### DIFF
--- a/holoviews/core/data/pandas.py
+++ b/holoviews/core/data/pandas.py
@@ -1,3 +1,5 @@
+from itertools import chain
+
 import numpy as np
 import pandas as pd
 from packaging.version import Version
@@ -72,7 +74,7 @@ class PandasInterface(Interface, PandasAPI):
 
             if kdims and not (len(kdims) == len(index_names) and {dimension_name(kd) for kd in kdims} == set(index_names)):
                 kdim = dimension_name(kdims[0])
-                if eltype._auto_indexable_1d and ncols == 1 and kdim not in data.columns:
+                if eltype._auto_indexable_1d and ncols == 1 and kdim not in chain(data.columns, cls.indexes(data)):
                     data = data.copy()
                     data.insert(0, kdim, np.arange(len(data)))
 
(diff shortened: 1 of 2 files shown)


holoviews/core/data/pandas.py
```diff
@@ -72,7 +74,7 @@ def init(cls, eltype, data, kdims, vdims):
 
             if kdims and not (len(kdims) == len(index_names) and {dimension_name(kd) for kd in kdims} == set(index_names)):
                 kdim = dimension_name(kdims[0])
-                if eltype._auto_indexable_1d and ncols == 1 and kdim not in data.columns:
+                if eltype._auto_indexable_1d and ncols == 1 and kdim not in chain(data.columns, cls.indexes(data)):
```
hoxbro, line 77:
```suggestion
                if eltype._auto_indexable_1d and ncols == 1 and kdim not in [*data.columns, *cls.indexes(data)]:
```

Do you think this will work?

maximlt: I think so. Which approach do you prefer?

hoxbro: I think `chain` is a big tool for the job and adds an import. But I will let you choose, as I don't have that strong of an opinion.

hoxbro (approved):

PR #6303, opened 2024-06-29
Fix datashader example in 16-Streaming_Data.ipynb

Zentrik:
This change stops `from holoviews.operation.datashader import datashade` from erroring.

Also, it seems quite odd to me specify streams without `RangeXY` as that prevents datashader from rerendering the plot when zooming in. If that can't be made to work, I think this example should be removed.
https://github.com/holoviz/holoviews/blob/e5f7aede7a58902677eb995b8fd67c54ae9ae3ab/examples/user_guide/16-Streaming_Data.ipynb#L541

### DIFF
--- a/examples/user_guide/16-Streaming_Data.ipynb
+++ b/examples/user_guide/16-Streaming_Data.ipynb
@@ -532,6 +532,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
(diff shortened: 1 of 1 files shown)


Zentrik:
I think I was just mistaken, it seems to work fine.

PR #6314, opened 2024-07-08
De-parameterize the options Keywords objects

philippjfr:
The `Keywords` object is used to store the allowable keywords for an options object. It was `Parameterized` but is such a simple object that there is really no need and since it is created very, very frequently this turned out to be a very, very expensive thing to create. In many situations the creation of these objects could account for anywhere between 10-30% of the time to display anything.


philippjfr:
ffmpeg issue unrelated and fixed separately.

PR #6315, opened 2024-07-08
Freeze models while updating plot(s)

philippjfr:
Uses the ability to freeze the Bokeh `Document` model graph while a plot is being updated to avoid needlessly recomputing the graph over and over. Results in about a 10% speed improvement during plot updates.


philippjfr:
ffmpeg issue fixed by pinning in this PR but will update the lock separately.

PR #6318, opened 2024-07-09
Improve following behavior when streaming

philippjfr:
- Automatically stops following when the user zooms in and starts following when the user resets
- Ensures that if the stream is out-of-sync with the plot the full data is synced

~~There still appears to be a race condition in bokeh where if a stream event arrives while the frontend is rendering it can cause "backtracking" glitches.~~

TODO:
- [x] Recompute range based on every new data point
- [x] Fix backtracking (fixed in Panel)

<details> <summary> Code </summary>

```python
import asyncio
import pandas as pd
import numpy as np

import holoviews as hv
hv.extension('bokeh')

nchannels = 25
buffer = hv.streams.Buffer(data=pd.DataFrame(np.random.randn(10, nchannels).cumsum(axis=1), columns=[chr(65+i) for i in range(nchannels)]))

def out(data):
    return hv.NdOverlay(
        {chr(65+i): hv.Curve(data, 'index', (chr(65+i), 'Amplitude')).opts(subcoordinate_y=True) for i in range(nchannels)}
    )

dmap = hv.DynamicMap(out, streams=[buffer]).opts(legend_position='right', legend_cols=2, responsive=True, min_height=600)

pane = pn.pane.HoloViews(dmap)

async def stream(event):
    for i in range(1, 1000):
        buffer.event(data=pd.DataFrame(np.random.randn(10, nchannels).cumsum(axis=1), columns=[chr(65+i) for i in range(nchannels)], index=np.arange(i*10, i*10+10)))
        await asyncio.sleep(0.01)

pn.Column(pn.widgets.Button(on_click=stream, name='Stream'), pane).servable()
```

</details>

ezgif-3-bde2bf8b98

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -1018,16 +1018,18 @@ class ElementPlot(BokehPlot, GenericElementPlot):
         return fig
 
     def _disable_follow(self, event):
-        self._events.append(event)
-        self.state.hold_render = True
+        hold = self.state.hold_render
         for stream in self.streaming:
             stream.following = False
+        if not hold:
+            stream.trigger(self.streaming)
+            self.state.hold_render = True
 
     def _reset_follow(self, event):
-        self._events.append(event)
         self.state.hold_render = False
         for stream in self.streaming:
             stream.following = True
+        stream.trigger(self.streaming)
 
     def _plot_properties(self, key, element):
         """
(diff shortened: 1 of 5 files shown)


philippjfr:
The backtracking behavior was fixed in https://github.com/holoviz/panel/pull/7128

philippjfr:
With range handling:

ezgif-6-90d086a92b

hoxbro (approved):

Issue #6327, opened 2024-07-16
(Possibly) wrong way for determination about X- and Y- axis in hv.Image


arafune:
Thanks for contacting us! Please read and follow these instructions carefully, then delete this introductory text to keep your issue easy to read. Note that the issue tracker is NOT the place for usage questions and technical assistance; post those at [Discourse](https://discourse.holoviz.org) instead. Issues without the required information below may be closed immediately.

#### ALL software version info
Python 3.11
holoviews 1.19.1
Bokeh 3.5.0

#### Description of expected behavior and the observed behavior
Expect the same output, but actually not.

I had DataArrays that worked well in my script and others that did not, and for a long time, I couldn't understand why. Finally, I have figured it out. I consider this a rather serious issue and would like to hear your opinion.

Suppose the following two (essentially) identical xarray.DataArrays:

```python
import numpy as np
import holoviews as hv
import xarray as xr

hv.extension("bokeh")
matrixdata = np.arange(24.0).reshape((4,6))
x_axis = np.linspace(0,5, 4) # [0,, 1.6666, 3.3333, 5]
y_axis = np.linspace(1, 8, 6) #[1. , 2.4, 3.8, 5.2, 6.6, 8. ]


dataarray_a = xr.DataArray(matrixdata, dims=["X", "Y"], coords={"X": x_axis, "Y": y_axis})
dataarray_b = xr.DataArray(matrixdata, dims=["X", "Y"], coords={"Y": y_axis, "X": x_axis})
```

I believe that hv.Image(dataarray_a) and hv.Image(dataarray_b) should be identical because both have the same "dims", which is a tuple. But they actually are not.

The horizontal axis of hv.Image(dataarray_a) is "X", while that of hv.Image(dataarray_b) is "Y".

As the items in a dict object may be (essentially) ordered. I know the current Python dict is, but I believe HoloViews should not rely on it...

Is this the holoviews specifiacation? If yes, why was this choice made?  And let me know to avoid the problem (In other wards,  please let me know how to achieve the same output).

- [ ] I may be interested in making a pull request to address this

jbednar:
Thanks for pointing out this behavior. In your opinion, what should be the determining factor about what goes on the horizontal axis and what goes on the vertical? Should it be the order of the dims, the order of the coords, or some other convention?

jbednar:
> As the items in a dict object may be (essentially) ordered. I know the current Python dict is, but I believe HoloViews should not rely on it...

[...]

PR #6333, opened 2024-07-19
To work around #6327

arafune:
While this PR is not still sufficient to work around #6327, such revision must be required at least.

The essential point of the change is the following part.

-            kdim_len = len(kdim_param.default) if kdims is None else len(kdims)
+            if kdims is None:
+                kdims = list(data.dims)
+            kdim_len = len(kdims)

While reading the code, I found the following part difficult to understand:
```python
if not isinstance(data, (xr.Dataset, xr.DataArray)):
```
To improve readability, I changed it to a more straightforward approach.  I hope you will accept this change as well.


ahuang11:
Thanks for creating this PR.

> While this PR is not still sufficient to work around https://github.com/holoviz/holoviews/issues/6327, such revision must be required at least.

Can you elaborate a bit on this? What else is needed?

hoxbro:
Currently, tests are failing because of your changes. Some seem trivial fixes (moving order around), while others do not. 

Making it so the tests pass will be the next step.

hoxbro:
You are only running the `core-test`; try running the full test suite with `pixi run -e test-312 test-unit`.

hoxbro:
Yes, they should be able to run on MacOS, but it seems like a bad file. Try `rm -rf .pixi pixi.lock` and then `pixi run` again. 

Are you running on an ARM Mac? Or is it an Intel?

PR #6338, opened 2024-07-23
Ensure CI version check only tests compatible versions

philippjfr:
Previously it was considering versions even if the Python version was not compatible.


maximlt:
Pushed https://github.com/holoviz/holoviews/pull/6338/commits/53784aa8c778c71c335dfd4fbf1c9629342b7dac, without any HoloViews code change `check_latest_packages.py`doesn't run.

maximlt (approved):

Issue #6342, opened 2024-07-26
Cannot reopen popup when using reference examples


MarcSkovMadsen:
holoviews==1.19.1

I would like to try out the new pop up form with HoloViews and Panel.

Unfortunately the user experience of the Panel reference example in https://holoviews.org/user_guide/Custom_Interactivity.html is pretty bad if you choose to `close` the form by clicking the `close` button.

![image](https://github.com/user-attachments/assets/4454e5ec-b226-4336-887c-d37f6504fb8d)

When you select some other datapoints the description will be hidden.

![image](https://github.com/user-attachments/assets/a88a8da8-05cf-4b39-ac47-555f51339cd5)

```python
import panel as pn
import numpy as np
import holoviews as hv

pn.extension()

def popup_form(index):
    def hide_popup(_):
        layout.visible = False

    if not index:
        return
    df = points.iloc[index].dframe().describe()
    button = pn.widgets.Button(name="Close", sizing_mode="stretch_width")
    layout = pn.Column(button, df)
    button.on_click(hide_popup)
    return layout


points = hv.Points(np.random.randn(1000, 2))
hv.streams.Selection1D(source=points, popup=popup_form)

points.opts(
    tools=["box_select", "lasso_select", "tap"],
    active_tools=["lasso_select"],
    size=6,
    color="black",
    fill_color=None,
    width=500,
    height=500
)

pn.panel(points).servable()
```

MarcSkovMadsen:
I experience the same behavior with the example from the PR

```python
import panel as pn
import numpy as np
import holoviews as hv
hv.extension("bokeh")


points = hv.Points(np.random.randn(1000, 2))

def form(name):
    text_input = pn.widgets.TextInput(name='Description')
    button = pn.widgets.Button(name='Save', on_click=lambda _: layout.param.update(visible=False))
    layout = pn.Column(f'# {name}', text_input, button)
    return layout

hv.streams.BoundsXY(source=points, popup=form('Bounds'))
hv.streams.Lasso(source=points, popup=form('Lasso'))
hv.streams.Tap(source=points, popup=form('Tap'))

pn.panel(points.opts(tools=['box_select', 'lasso_select', 'tap'], width=600, height=600, size=6, color='black', fill_color=None)).servable()
```

Once you've clicked save once you cannot open/ see the pop up again.

MarcSkovMadsen:
The cause is around here

[...]

PR #6343, opened 2024-07-27
Fix popup not visible

MarcSkovMadsen:
Closing #6342


MarcSkovMadsen:
Will not solve the issue. I don't understand the implementation. And I cannot find any tests that can help me understand all the `if` cases. I give up.

PR #6351, opened 2024-08-01
🎨  Just improving logic structure.

### LINKED ISSUE #6327: (Possibly) wrong way for determination about X- and Y- axis in hv.Image

arafune:
While I'm not so sure, whether you can accept such PR or not.

In the meanwhile, I'm trying to fix #6327.  I sent the PR #6333.
However, at that moment, I didn't figure out the correct way of testing. I realized the previous PR is far from robustness.  Thus I feel I need analyze the code firstly and then think to improve the code to workaround it.

During the analysis of the code, I have felt the original code is not somehow elegant.  

In the original

if not isinstance(....):
    A
else:
   B

But I believe 

if isinstance(...):
   B
else:
   A

is much simpler and easy to read.

This PR is just change to the above.  No essential change.
I'm afraid the original authors' feel such change is needless.  But I believe for further analysis by many contributors such revision would be works good.

Could you please think about acceptance?

### DIFF
--- a/holoviews/core/data/xarray.py
+++ b/holoviews/core/data/xarray.py
@@ -125,7 +125,33 @@ class XArrayInterface(GridInterface):
                     vdims = [vdim]
                     data = data.to_dataset(name=vdim.name)
 
-        if not isinstance(data, (xr.Dataset, xr.DataArray)):
+        if isinstance(data, (xr.Dataset, xr.DataArray)):
+            # Started to warn in xarray 2023.12.0:
+            # The return type of `Dataset.dims` will be changed to return a
+            # set of dimension names in future, in order to be more consistent
+            # with `DataArray.dims`. To access a mapping from dimension names to
+            # lengths, please use `Dataset.sizes`.
+            data_info =  data.sizes if hasattr(data, "sizes") else data.dims
+            if not data.coords:
+                data = data.assign_coords(**{k: range(v) for k, v in data_info.items()})
+            if vdims is None:
+                vdims = list(data.data_vars)
+            if kdims is None:
+                xrdims = list(data_info)
+                xrcoords = list(data.coords)
+                kdims = [name for name in data.indexes.keys()
+                         if isinstance(data[name].data, np.ndarray)]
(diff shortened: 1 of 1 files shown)


hoxbro:
Thanks for the PR.

PR #6366, opened 2024-08-29
enh: Update norm function in transform.py

NolanBrb:
New feature / Bug Fix

Added default behaviour when values to normalize are all equal; previously holoviews would attempt division by zero and return a runtime warning. 
Proposed change : if all values are equal to zero, set all values to 0; otherwise, set all values to 1.


hoxbro:
Can you explain why this is needed?

NolanBrb:
Sure ! 

I ran into this error because I'm working on xarray datasets with multiple variables and procedurally creating vectorfields for each variable. I'm controlling the scale of the vectors using .opts(magnitude=hv.dim("Magnitude").norm())
Some of the variables I use have fixed magnitude, and when displaying the plot the line above returned a warning because since min==max==1, norm() was attempting to divide by zero.

hoxbro:
I'm not opposed to this PR, but currently, it breaks CI. 

Did you try to pass in limits to `hv.dim("Magnitude").norm(limits=(0, None))`? Or would you like it to be dynamically updated?

Would you happen to have a small example I could play around with?

holoviews/util/transform.py
```diff
@@ -47,6 +47,10 @@ def norm(values, min=None, max=None):
     """
     min = np.min(values) if min is None else min
     max = np.max(values) if max is None else max
+    if min==max:
+        if max==0:
+            return 0*values
+        return values/max
```
philippjfr: If the values are already zero you don't need to multiply them by zero. Otherwise this looks good. Thank you!

```suggestion
    if min == max:
        return values if max == 0 else (values / max)
```

hoxbro:
Thank you for the PR, this will be part of the next release 1.21.0

PR #6369, opened 2024-09-10
General maintenance 

hoxbro:


holoviews/__init__.py
```diff
@@ -114,12 +113,7 @@
     HoloviewsUserWarning,
 )
 
-# Suppress warnings generated by NumPy in matplotlib
-# Expected to be fixed in next matplotlib release
```
hoxbro, line 118: I don't know what the next matplotlib version was, but this was added nine years ago...

philippjfr (approved):

PR #6370, opened 2024-09-11
Implement show method

### LINKED ISSUE #6368: Add .show method on HoloViews elements
Since Panel is already a dependency, I think it's just wrapping the element in `pn.panel(el).show()`

ahuang11:
Closes https://github.com/holoviz/holoviews/issues/6368


philippjfr:
Apart from the API discussion this just doesn't seem like the right approach. Why not put it on some shared baseclass?

Issue #6372, opened 2024-09-16
resampling_when breaks when zooming in


ahuang11:
```python
import hvplot.pandas
import panel as pn
import pandas as pd

df = pd.DataFrame({
    "lon": [45.5531, 46.4731, 47.4731, 48.41234],
    "lat": [34.5531, 35.4731, 36.4731, 37.41234],
})

pn.serve((df.hvplot.points("lon", "lat", rasterize=True, resample_when=1000)))
```

<img width="464" alt="image" src="https://github.com/user-attachments/assets/9ad144de-e5fb-4d20-a3cd-d417065f53aa">

```
WARNING:bokeh.core.validation.check:W-1005 (FIXED_SIZING_MODE): 'fixed' sizing mode requires width and height to be set: figure(id='p1110', ...)
WARNING:param.dynamic_operation: Callable raised "ValueError("conflicting sizes for dimension 'lon': length 730 on the data but length 365 on coordinate 'lon'")".
Invoked as dynamic_operation(height=300, scale=1.0, width=365, x_range=(32.55793907008026, 49.477519405740225), y_range=(17.986703041165143, 30.624333539380807))
Traceback (most recent call last):
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/plotting/util.py", line 293, in get_plot_frame
    return map_obj[key]
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/core/spaces.py", line 1216, in __getitem__
    val = self._execute_callback(*tuple_key)
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/core/spaces.py", line 983, in _execute_callback
    retval = self.callback(*args, **kwargs)
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/core/spaces.py", line 552, in __call__
    return self.callable()
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/core/overlay.py", line 34, in dynamic_mul
    element = other[args]
  File "/Users/ahuang/miniconda3/envs/hvplot/lib/python3.10/site-packages/holoviews/core/spaces.py", line 1216, in __getitem__
    val = self._execute_callback(*tuple_key)
[...]

PR #6373, opened 2024-09-16
Ensure pixel_ratio is applied when computing resampling coordinates

ahuang11:
Closes https://github.com/holoviz/holoviews/issues/6372

Before:

https://github.com/user-attachments/assets/d5dc6738-d6f8-4b88-9355-fb073359b6b7

After:

https://github.com/user-attachments/assets/639a6b70-ed25-41f5-8642-3a4bcd29db31

```python
import pandas as pd
import holoviews as hv
from holoviews.operation import apply_when
from functools import partial
import panel as pn

hv.extension("bokeh")

df = pd.DataFrame(
    {
        "lon": [45.5531, 46.4731, 47.4731, 48.41234],
        "lat": [34.5531, 35.4731, 36.4731, 37.41234],
    }
)

def _resample_obj(operation, obj, opts):
    def exceeds_resample_when(plot):
        return len(plot) > 1000

    processed = apply_when(
        obj, operation=partial(operation, **opts), predicate=exceeds_resample_when
    )
    return processed


points = hv.Points(df, ["lon", "lat"])
resampled_points = _resample_obj(rasterize, points, {})
pn.serve(resampled_points)
```


holoviews/operation/datashader.py
```diff
@@ -362,7 +362,7 @@ def _process(self, element, key=None):
 
         if self.p.precompute:
             self._precomputed[element._plot_id] = x, y, data, glyph
-        (x_range, y_range), (xs, ys), (width, height), (xtype, ytype) = self._get_sampling(element, x, y)
+        (x_range, y_range), (xs, ys), (width, height), (xtype, ytype) = self._get_sampling(element, x, y, apply_pixel_ratio=False)
```
ahuang11, line 365: Without this, it'll error below on line 374 because width/height mismatch len(xs) and len(ys)

philippjfr: This defeats the point of the pixel ratio entirely as far as I can tell.

holoviews/tests/ui/bokeh/test_datashader.py
```diff
@@ -0,0 +1,54 @@
...
+except ImportError:
+    raise SkipTest("Datashader not available")
+from ...plotting.utils import ParamLogStream
+from .. import expect, wait_until
+
+pytestmark = pytest.mark.ui
+
+
+@pytest.mark.usefixtures("bokeh_backend")
+def test_apply_when_zoom(serve_hv):
```
philippjfr, line 21: I don't know why this would ever have to be a UI test.

philippjfr: UI tests should only be implemented for functionality that actually happens on the frontend. In this case you can simulate all the conditions entirely in Python.

holoviews/operation/datashader.py
philippjfr: Yeah, the correct fix here is to multiply the width/height before creating the coordinates, not to simply disable pixel ratio entirely.

philippjfr (approved):

PR #6378, opened 2024-09-19
Drop z dimension from ImageStack DataArray before shading

philippjfr:
When performing categorical rasterization with only one category the `XArrayInterface` would convert the 3D `DataArray` into a Dataset. When this 3D Dataset was then shaded we would convert the singular DataArray variable into a z-dimension, which meant we'd have two categorical variables (with the same value) and datashader would error out because it expects a 2D or 3D array.

Here we do a number of things:

- Allow the `XArrayInterface` to hold a DataArray with a singular value dimension
- Ensure that shade handles non-xarray datasets correctly (by promoting them to xarray)
- Ensure that it does not needlessly add a z-dimension if we have a singular value dimension
- Even if we have a 3D array and a multiple z-values we squeeze the DataArray to drop superfluous constant dimensions.
- Support dask array backed xarrays in shade (by materializing them)

### DIFF
--- a/holoviews/core/data/xarray.py
+++ b/holoviews/core/data/xarray.py
@@ -446,7 +446,11 @@ class XArrayInterface(GridInterface):
         Given a dataset object and data in the appropriate format for
         the interface, return a simple scalar.
         """
-        if not cls.packed(dataset) and len(data.data_vars) == 1:
+        if cls.packed(dataset):
+            array = data.squeeze()
+            if len(array.shape) == 0:
+                return array.item()
+        elif len(data.data_vars) == 1:
             array = data[dataset.vdims[0].name].squeeze()
             if len(array.shape) == 0:
                 return array.item()
--- a/holoviews/core/data/xarray.py
+++ b/holoviews/core/data/xarray.py
@@ -93,7 +93,7 @@ class XArrayInterface(GridInterface):
         if isinstance(data, xr.DataArray):
             kdim_len = len(kdim_param.default) if kdims is None else len(kdims)
             vdim_len = len(vdim_param.default) if vdims is None else len(vdims)
-            if vdim_len > 1 and kdim_len == len(data.dims)-1 and data.shape[-1] == vdim_len:
+            if kdim_len == len(data.dims)-1 and data.shape[-1] == vdim_len:
                 packed = True
             elif vdims:
                 vdim = vdims[0]
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1348,7 +1348,7 @@ class shade(LinkableOperation):
         if self.p.clims:
             shade_opts['span'] = self.p.clims
         elif ds_version > Version('0.5.0') and self.p.cnorm != 'eq_hist':
-            shade_opts['span'] = (array.min(), array.max())
+            shade_opts['span'] = (array.min().item(), array.max().item())
 
         params = dict(get_param_values(element), kdims=kdims,
                       bounds=bounds, vdims=RGB.vdims[:],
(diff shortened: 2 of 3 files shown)


holoviews/operation/datashader.py
```diff
@@ -1335,7 +1348,7 @@ def _process(self, element, key=None):
         if self.p.clims:
             shade_opts['span'] = self.p.clims
         elif ds_version > Version('0.5.0') and self.p.cnorm != 'eq_hist':
-            shade_opts['span'] = element.range(vdim)
+            shade_opts['span'] = (array.min(), array.max())
```
philippjfr, line 1351: This was not correctly fetching the span for a 3D array.

holoviews/core/data/xarray.py
```diff
@@ -93,7 +93,7 @@ def retrieve_unit_and_label(dim):
         if isinstance(data, xr.DataArray):
             kdim_len = len(kdim_param.default) if kdims is None else len(kdims)
             vdim_len = len(vdim_param.default) if vdims is None else len(vdims)
-            if vdim_len > 1 and kdim_len == len(data.dims)-1 and data.shape[-1] == vdim_len:
+            if kdim_len == len(data.dims)-1 and data.shape[-1] == vdim_len:
```
hoxbro, line 96: Why is `vdim_len` not needed anymore?

holoviews/operation/datashader.py
```diff
@@ -1286,17 +1286,35 @@ def _process(self, element, key=None):
...
+            element = element.clone(datatype=['xarray'])
+
         kdims = element.kdims
         if isinstance(element, ImageStack):
             vdim = element.vdims
             array = element.data
-            if hasattr(array, "to_array"):
-                array = array.to_array("z")
-            array = array.transpose(*[kdim.name for kdim in kdims], ...)
+            # If data is a Dataset it has to be converted to a
```
hoxbro, line 1297:
```suggestion
            # If data is a xarray Dataset it has to be converted to a
```

holoviews/operation/datashader.py
```diff
@@ -1286,17 +1286,35 @@ def _process(self, element, key=None):
...
+            kdims = [kdim.name for kdim in kdims]
+            if not element.interface.packed(element):
+                if len(vdim) == 1:
+                    array = array[vdim[0].name]
+                else:
+                    array = array.to_array("z")
+                    # If data is 3D then we have one extra constant dimension
+                    if array.ndim > 3:
+                        drop = [d for d in array.dims if d not in kdims+["z"]]
+                        array = array.squeeze(drop)
```
hoxbro, line 1309:
```suggestion
                        array = array.squeeze(dim=drop)
```

https://docs.xarray.dev/en/stable/generated/xarray.DataArray.squeeze.html#xarray.DataArray.squeeze

holoviews/operation/datashader.py
```diff
@@ -1286,17 +1286,35 @@ def _process(self, element, key=None):
...
+            array = array.transpose(*kdims, ...)
         else:
             vdim = element.vdims[0].name
             array = element.data[vdim]
 
+        # Dask is not supported by shade so materialize it
+        array = array.compute()
```
hoxbro, line 1316: This is a no-op if the underlying data is numpy and not dask, correct?

hoxbro (commented):
If you haven't already maybe also run the hvplot test suite (locally) to verify that this change does not break anything because of the changes in `XarrayInterface`.

holoviews/core/data/xarray.py
philippjfr: This was the proximate cause of the problem in the first place, as I described above categorical arrays with a single category were treated differently from categorical arrays with multiple dimensions. We now handle them consistently.

holoviews/operation/datashader.py
philippjfr: Thanks!

philippjfr:
>If you haven't already maybe also run the hvplot test suite (locally) to verify that this change does not break anything because of the changes in XarrayInterface.

Done, it's passing.

hoxbro (approved):

Issue #6386, opened 2024-09-24
Broken shared linked crosshair tool if set on one dimension only


TheoMathurin:
#### ALL software version info

<details>
    <summary>Software Version Info</summary>

```plaintext
holoviews 1.19.1
bokeh 3.5.2
panel 1.15.0
```    
</details>

#### Description of expected behavior and the observed behavior
The update to Bokeh 3.5 breaks shared Crosshair tools seemingly if they're not enabled for both width and height. Separate tools appear separately on each plot.

#### Complete, minimal, self-contained example code that reproduces the issue

NB: Pure bokeh example from their docs works as expected.

```python
import holoviews as hv
import numpy as np
from bokeh.models import CrosshairTool, Span
import panel as pn

hv.extension("bokeh")

s1 = hv.Scatter(np.random.randn(10, 10))
s2 = hv.Scatter(np.random.randn(10, 10))
s3 = hv.Scatter(np.random.randn(10, 10))
s4 = hv.Scatter(np.random.randn(10, 10))

height = Span(dimension="height")
width = Span(dimension="width")
cht1 = CrosshairTool(overlay=[height, width])
cht2 = CrosshairTool(overlay=height)
s1 = s1.opts(tools=[cht1])
s2 = s2.opts(tools=[cht1])
s3 = s3.opts(tools=[cht2])
s4 = s4.opts(tools=[cht2])

pn.Row(pn.Column(s1, s2), pn.Column(s3, s4)).servable()
```

#### Stack traceback and/or browser JavaScript console output
Nothing

#### Screenshots or screencasts of the bug in action
![crosshair](https://github.com/user-attachments/assets/794a8b70-3692-49ca-8f10-691a6195d8e2)

hoxbro:
Introduced by https://github.com/holoviz/holoviews/pull/6220

PR #6387, opened 2024-09-24
 Fix: Don't clone Span in `_init_tools`

hoxbro:
Fixes https://github.com/holoviz/holoviews/issues/6386

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -21,6 +21,7 @@ from bokeh.models import (
     GlyphRenderer,
     Legend,
     Renderer,
+    Span,
     Title,
     tools,
 )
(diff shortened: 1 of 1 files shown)


TheoMathurin:
Thanks for the lightning fast fix!

philippjfr (commented):
This is not safe. We clone models because a user provided `Model` can only be used within a single plot. What likely needs to happen is that instead of cloning and replacing the models on a per-plot basis we have to search the entire model graph and replace shared instances all at once.

hoxbro:
Just to be sure, we still clone the `CrosshairTool` with `tool = type(tool)(**properties)`. This is to avoid cloning `overlay=height` in the following. The span in `overlay=[height, width]` is not cloned. 

This is an regression and was only reasonly introduced in https://github.com/holoviz/holoviews/pull/6220.

``` python
height = Span(dimension="height")
width = Span(dimension="width")
cht1 = CrosshairTool(overlay=[height, width])
cht2 = CrosshairTool(overlay=height)
```

philippjfr:
Sorry, misread it, this does look correct.

PR #6411, opened 2024-10-14
Fix popup positioning on polygons

philippjfr:
The coordinate calculation for popups did not handle nested multi-array data as occurs when plotting polygons.

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -1191,13 +1191,24 @@ class Selection1DCallback(PopupMixin, Callback):
               }
               if (!xs || !ys) { return }
               for (const i of indices) {
-                const tx = xs[i]
-                if (!x || (tx > x)) {
-                  x = xs[i]
+                let ix = xs[i]
+                let iy = ys[i]
+                if (typeof ix === 'number') {
+                  ix = [ix]
+                  iy = [iy]
+                } else {
+                  while (ix.length && (typeof ix[0] !== 'number')) {
+                    ix = ix[0]
+                    iy = iy[0]
+                  }
                 }
-                const ty = ys[i]
+                const tx = Math.max(...ix)
+                if (!x || (tx > x)) {
+                  x = tx
+                }
+                const ty = Math.max(...iy)
                 if (!y || (ty > y)) {
-                  y = ys[i]
+                  y = ty
                 }
               }
               if (x && y) {


holoviews/tests/ui/bokeh/test_callback.py
```diff
@@ -266,6 +266,24 @@ def popup_form(name):
     expect(locator).to_have_count(1)
 
 
+@skip_popup
+@pytest.mark.usefixtures("bokeh_backend")
+def test_stream_popup_polygons(serve_hv):
```
hoxbro, line 271: This also passes on main

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -1191,13 +1191,24 @@ def _watch_position(self):
...
               if (!xs || !ys) { return }
               for (const i of indices) {
-                const tx = xs[i]
+                let ix = xs[i]
+                let iy = ys[i]
+                if (typeof ix === 'number') {
+                  ix = [ix]
+                  iy = [iy]
+                } else {
+                  while (ix.length && (typeof ix[0] !== 'number')) {
```
hoxbro, line 1200:
```suggestion
                  while (Array.isArray(ix) && (typeof ix[0] !== 'number')) {
```

`length` seems ambiguous

holoviews/plotting/bokeh/callbacks.py
```diff
@@ -1191,13 +1191,24 @@ def _watch_position(self):
...
+                    iy = iy[0]
+                  }
+                }
+                const tx = Math.max(...ix)
                 if (!x || (tx > x)) {
-                  x = xs[i]
+                  x = tx
                 }
-                const ty = ys[i]
+                const ty = Math.max(...iy)
```
hoxbro, line 1209: Can't we do the `Math.max` after the while loop and avoid the weird workaround if `ix` is "just" a number?

holoviews/plotting/bokeh/callbacks.py
philippjfr: No, the length check is intentional.

holoviews/plotting/bokeh/callbacks.py
philippjfr: Why "weird" and no I don't see how that fixes anything?

philippjfr: To elaborate, `Math.max(...ix)` will not work on a number because it's not iterable.

holoviews/plotting/bokeh/callbacks.py
philippjfr: If it's not an array or a number we have other problems I suppose.

hoxbro (approved):

PR #6412, opened 2024-10-15
enh: Rename `index` column to `__index__` and remove old version check

hoxbro:
The rename to `__index__` is needed as currently passing in the following `ds.summary(val=ds.min("s"), index=ds.where(ds.min("s")))` as an aggregator would always place `index` at the end. This can also happen with `__index__`, but the chance of this collision seems much lower. This was flagged in the [original pr](https://github.com/holoviz/holoviews/pull/5805#discussion_r1269628614) but was never done.

The version rename is to align the naming of the versions check across holoviews and will done for all files in https://github.com/holoviz/holoviews/pull/6413. The main point is also to make it easier to understand what the version constant means. 

The dropping of compatibility with Datashader 0.5.0, which was released somewhere between 2016 and 2018... 
image

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -69,9 +69,10 @@ from ..element.util import connect_tri_edges_pd
 from ..streams import PointerXY
 from .resample import LinkableOperation, ResampleOperation2D
 
-ds_version = Version(ds.__version__)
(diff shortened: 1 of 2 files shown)


philippjfr:
Thanks for the description, makes sense. Happy to see this merged.

Issue #6326, opened 2024-07-15
Categorical datashaded paths can raise an error


maximlt:
The code below:

- Creates a DataFrame that holds categorical paths data. The paths are separated with NaN values for x and y but empty strings for the categorical column.
- Datashade the paths using a count_cat aggregator.

With some pretty old versions of HoloViews, Datashader and Pandas, it works.

<img width="411" alt="image" src="https://github.com/user-attachments/assets/11fc8980-b256-4532-93f0-c98b870b1cfc">

But it breaks with newer versions with this error `ValueError: Dimension name cannot be empty`.

- `git bisect` points to https://github.com/holoviz/holoviews/pull/5751.
- Replacing `CAT_SEP = ''` with `CAT_SEP = np.nan` makes the code work now. So I wonder, should all the lines contain `NaN`, i.e. is the original example just bad data? Or is it just a regression?

---

```python
import datashader as ds
import numpy as np
import pandas as pd
import holoviews as hv

from holoviews.operation import datashader as hd

hv.extension('bokeh')

spaths = [
    np.array([[1, 2], [3, 4], [5, 6]]),
    np.array([[7, 8], [9, 10]]),
    np.array([[11, 12], [13, 14], [15, 16], [17, 18]]),
    np.array([[19, 20], [21, 22]]),
    np.array([[23, 24], [25, 26]]),
]

nan_row = np.array([[np.nan, np.nan]])
paths = np.concatenate([x for spath in spaths for x in [spath, nan_row]])

df = pd.DataFrame(paths, columns=['x', 'y'])

# CAT_SEP = np.nan
CAT_SEP = ''

origin = []
for s, spath in zip('abcdefgh', spaths):
    for _ in range(len(spath)):
        origin.append(s)
    origin.append(CAT_SEP)
df['origin'] = origin

df['origin'] = df['origin'].astype('category')

path = hv.Path(df, ['x', 'y'])

aggregator = ds.count_cat('origin')
datashaded = hd.datashade(
    path,
    aggregator=aggregator,
)
datashaded.opts(bgcolor='black')
```

Before:
<img width="305" alt="image" src="https://github.com/user-attachments/assets/0d6d60ac-616e-40b1-b03f-076b2ccdb06f">

<details><summary>Traceback now</summary>
<p>

```python
>
/
Name
Modified
File Size

spaths = [
    np.array([[1, 2], [3, 4], [5, 6]]),
    np.array([[7, 8], [9, 10]]),
[...]

PR #6429, opened 2024-10-30
Fix rasterizing a Path with categorical values including empty strings

maximlt:
Fixes https://github.com/holoviz/holoviews/issues/6326

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -449,7 +449,7 @@ class aggregate(LineAggregationOperation):
                     val[neg1] = np.nan
                 agg[col] = ((y.name, x.name), val)
 
-        if isinstance(agg_fn, ds.count_cat):
+        if isinstance(agg_fn, ds.by):
             col = agg_fn.column
             if '' in agg.coords[col]:
                 agg = agg.drop_sel(**{col: ''})
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -448,6 +448,12 @@ class aggregate(LineAggregationOperation):
                     val = val.astype(np.float64)
                     val[neg1] = np.nan
                 agg[col] = ((y.name, x.name), val)
+
+        if isinstance(agg_fn, ds.count_cat):
+            col = agg_fn.column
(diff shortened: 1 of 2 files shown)


holoviews/operation/datashader.py
```diff
@@ -448,6 +448,12 @@ def _apply_datashader(self, dfdata, cvs_fn, agg_fn, agg_kwargs, x, y):
                     val = val.astype(np.float64)
                     val[neg1] = np.nan
                 agg[col] = ((y.name, x.name), val)
+
+        if isinstance(agg_fn, ds.count_cat):
```
maximlt, line 452: Hopefully a good place to add this.

holoviews/tests/operation/test_datashader.py
```diff
@@ -1296,6 +1296,42 @@ def test_rasterize_apply_when_instance_with_line_width(self):
...
+        #         [0, 0],
+        #         [0, 0],
+        #         [0, 1]],
+        #     [[1, 0],
+        #         [0, 0],
+        #         [0, 0],
+        #         [0, 1]],
+        # ])
+        # expected = ImageStack((x, y, array), vdims=['a', 'b'])
+        # self.assertEqual(rasterized, expected)
```
maximlt, line 1333: @hoxbro I commented out this code for now as I realized that `compare.py` has not been updated to compare `ImageStack` objects part of the test suite. Is this intentional? Not that this part of the test is not strictly required, I worked on this fix together with @azaya89 and ended up showing him that `self.assertEqual` is extended by HoloViews. Ultimately, the real check is that there's no error raised and no empty string in the vdims.

hoxbro: It was not intentional. At some point, I want to transition away from using the stdlib unittest framework.  

Can you uncomment this and do some simple `assert` instead?

holoviews/operation/datashader.py
hoxbro: Placement looks fine to me. We also need this for `ds.by`:
```suggestion
        if isinstance(agg_fn, (ds.count_cat, ds.by)):
```

We could parameterize the test.

PR #6434, opened 2024-11-01
refactor: enable RUF005

hoxbro:
Ref: https://docs.astral.sh/ruff/rules/collection-literal-concatenation/

Will first merge it after 1.20 release.


holoviews/core/data/grid.py
```diff
@@ -122,7 +122,7 @@ def init(cls, eltype, data, kdims, vdims):
                 if shape[-1] != len(vdims):
                     raise error('The shape of the value array does not match the number of value dimensions.')
                 shape = shape[:-1]
-            if (not expected and shape == (1,)) or (len(set((shape,)+shapes)) == 1 and len(shape) > 1):
+            if (not expected and shape == (1,)) or (len({shape, *shapes}) == 1 and len(shape) > 1):
```
hoxbro, line 125:
```suggestion
            if (not expected and shape == (1,)) or (len(shape) > 1 and len({shape, *shapes}) == 1):
```

small perf

PR #6442, opened 2024-11-07
docs: Fix wrong variable name

ahuang11:
Without the fix:

`holoviews.core.data.interface.DataError: Following columns specified as dimensions but not found in data: ['detailed name']`

### DIFF
--- a/examples/user_guide/Linked_Brushing.ipynb
+++ b/examples/user_guide/Linked_Brushing.ipynb
@@ -284,7 +284,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "To link the two we will specify the `'detailed name'` column as the `index_cols`."
+    "To link the two we will specify the `detailed_name` column as the `index_cols`."
    ]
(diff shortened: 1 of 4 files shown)


examples/user_guide/Linked_Brushing.ipynb
```diff
@@ -293,7 +293,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "linked_choropleth = link_selections(choropleth + hist, index_cols=['detailed name'])"
+    "linked_choropleth = link_selections(choropleth + hist, index_cols=[detailed_name])"
```
hoxbro, line 296: Looking at the definition of `detailed_name`, it was a workaround Bokeh 3.0 change.

`"detailed_name = 'detailed_name' if counties[0].get('detailed_name') else 'detailed name' # detailed name was changed in Bokeh 3.0\n",`

Let us remove `detailed_name` and use `'detailed_name'` wherever needed.

PR #6443, opened 2024-11-07
ci: Move test-matrix into a separate file

hoxbro:
This in my opinion makes it a bit easier to read.


hoxbro:
Nah, doesn't work without cloning the repo. So will close this.

PR #6456, opened 2024-11-21
enh(dev): Make default environment the developer environment

hoxbro:
This PR does two main things:

1) Merge the features of `test` and `example` into `optional` as the overlap in dependencies is large.
2) Make the default environment the recommended developer environment for developers. 

@maximlt, I will update the docs with the dev recommendation if you agree with the approach. 

---
Image:


maximlt (commented):
I think that's a good approach. Maybe look for a better name instead of `optional`? Is the CI going to use the `dev` environment or it's just for contributors?

hoxbro:
> Maybe look for a better name instead of optional?

Open for suggestion only reason I took it was it looked nice in the toml file:
image

> Is the CI going to use the dev environment or it's just for contributors?

As is, it would not be run in CI. Have updated the title with `enh(dev)` to reflect this better.

maximlt:
Was it required to gather the `test` and `example` dependencies in one feature? Genuine question, I don't know much about pixi yet. I see `optional` is part of the `docs` environment so it sounds possible it'll install dependencies not required to build the docs.

hoxbro:
> Was it required to gather the `test` and `example` dependencies in one feature? [...] I see `optional` is part of the `docs` environment so it sounds possible it'll install dependencies not required to build the docs.

No. I think the overlap is significant enough that there is no need to have them in two different features. Other than the `docs` environment, every other group had both of them in it. And for the docs environment, the extra dependencies are `dask-expr`, `nbconvert`, `spatialpandas`, and `tsdownsample`. The others either are implicit dependencies or are also required by the docs feature. 

---

Maybe we should add an extra feature called `development`, which has dev-only features/commands in it. At the top of my head, I would add `Jupyter Lab` and maybe a task called `pixi run lab` (or similar).

maximlt:
> No. I think the overlap is significant enough that there is no need to have them in two different features. Other than the `docs` environment, every other group had both of them in it. And for the docs environment, the extra dependencies are `dask-expr`, `nbconvert`, `spatialpandas`, and `tsdownsample`. The others either are implicit dependencies or are also required by the docs feature.

I don't understand the need to merge the tests and examples environment? It'd make sense to me if they would 100% overlap but it's not the case. They quite overlap today but this might change. You're also going to loose flexibility. I'm also concerned with this practice spreading to all the repos.

> Maybe we should add an extra feature called `development`, which has dev-only features/commands in it. At the top of my head, I would add `Jupyter Lab` and maybe a task called `pixi run lab` (or similar).

Ah so ok now we just have a new `dev` environment, so people are supposed to run `pixi run test-unit -e dev`?

.github/workflows/test.yaml
```diff
@@ -107,7 +107,7 @@ jobs:
     name: Pixi lock
     runs-on: ubuntu-latest
     steps:
-      - uses: holoviz-dev/holoviz_tasks/pixi_lock@v0
+      - uses: holoviz-dev/holoviz_tasks/pixi_lock@ci_dont_install_lock
```
hoxbro, line 110: It needs to be changed back before merging.

hoxbro:
I have fixed your main observation from the last comment, getting to the real deal for developers. The default environment will be the "dev" environment, meaning there will be no selection menu and no adding `-e dev` to each command. This comes at the cost of a little more complicated  `pixi.toml` file.

I have made a dev doc [build](https://dev.holoviews.org/developer_guide/index.html).

doc/developer_guide/index.md
```diff
@@ -85,24 +105,38 @@ Syncing the git tags can be done with:
...
-### Editable install
+## Developer Environment
 
-It can be advantageous to install the HoloViews in [editable mode](https://pip.pypa.io/en/stable/topics/local-project-installs/#editable-installs):
+The `default` environment is meant to provide all the tools needed to develop HoloViews.
 
-```bash
-pixi run install
-```
+This environment can be created by running `pixi run install`, which will set up the environment and make an editable install of HoloViews.
```
maximlt, line 112: The docs already mention `pixi run install`, that feels a little redundant. I'd just remove it.

doc/developer_guide/index.md
```diff
@@ -85,24 +105,38 @@ Syncing the git tags can be done with:
...
 
-```bash
-pixi run -e test-ui install
-```
+This environment can also be selected in your IDE. In VS Code, this can be done by running the command `Python: Select Interpreter` and choosing `{'default': Pixi}`.
 
-:::
+<p style="text-align: center">
+  <img
+    src="https://github.com/user-attachments/assets/6a8f7f60-870d-4497-8e23-52b336cf7682"
```
maximlt, line 122: Are you going to upload these files somewhere else or leave them on Github?

doc/developer_guide/index.md
```diff
@@ -85,24 +105,38 @@ Syncing the git tags can be done with:
...
+  />
+</p>
+
+To confirm you are using this dev environment, check the bottom right corner:
+
+![003](https://github.com/user-attachments/assets/66b405e2-e6d0-4678-bd18-96b2a9e573ef)
+
+### Jupyter Lab
+
+You can launch Jupyter lab with the `default` environment with `pixi run lab`.
```
maximlt, line 139: I feel like we should let contributors know that they should run this command when:

- they need to edit the docs (as building the docs is usually overkill)
- they need to debug example tests

Either here, or perhaps preferably in the relevant sections.

doc/developer_guide/index.md
hoxbro: Uploaded on S3

hoxbro:
> should we ask people to run pixi run lint? Seems more like something you'd do in case you don't manage to install pre-commit. Maybe put it in an admonition?

I have kept it but added an admonition that you can use `pre-commit` directly.

pyproject.toml
```diff
@@ -128,6 +128,7 @@ filterwarnings = [
     "ignore::ResourceWarning",
     # 2024-11
     "ignore:The legacy Dask DataFrame implementation is deprecated:FutureWarning", # https://github.com/holoviz/spatialpandas/issues/146
+    "ignore:The cuda.cuda(rt)? module is deprecated:DeprecationWarning", # https://github.com/rapidsai/rmm/issues/1730
```
hoxbro, line 131:
```suggestion
```

No longer needed.

maximlt:
I'm working on migrating hvPlot to pixi and would like to implement this, is there an issue with the approach here? Asking since it has not yet been merged.

hoxbro:
Nothing other than you approving :+1:

hoxbro:
Failing tests are unrelated, I will look into it in another PR after the next dev release of Datashader.

PR #6466, opened 2024-12-05
fix: Don't have linked axis between source and target for RangeToolLink

hoxbro:


philippjfr:
Only worry I'd have is that this breaks streams attached to the source plot.

hoxbro:
> Only worry I'd have is that this breaks streams attached to the source plot.

I tried the following, but it does not seem to give any problem. I think `downsample1d` uses streams, but I could be wrong; if so, do you have something in mind?

``` python
import numpy as np

import holoviews as hv

hv.extension("bokeh")

x = np.linspace(0, 10, 10_000)
y = np.random.normal(size=10_000)

plot0 = downsample1d(hv.Curve((x, y))).opts(title="Source")
plot1 = hv.Curve((x, y)).opts(title="Target 1")

RangeToolLink(plot0, plot1, axes=["x", "y"])

plot0 + plot1
```

philippjfr:
Would try both in that case:

```python
import numpy as np

import holoviews as hv

hv.extension("bokeh")

x = np.linspace(0, 10, 10_000)
y = np.random.normal(size=10_000)

plot0 = downsample1d(hv.Curve((x, y))).opts(title="Source")
plot1 = downsample1d(hv.Curve((x, y))).opts(title="Target 1")

RangeToolLink(plot0, plot1, axes=["x", "y"])

plot0 + plot1
```

If that works I'm happy.

hoxbro:
Does not seem to give me any problems: 


https://github.com/user-attachments/assets/271345ea-4e50-49b0-9260-641c4acb7d73

philippjfr:
Neat!

philippjfr (approved):

PR #6476, opened 2024-12-13
refactor: Remove direct pandas import

hoxbro:
This PR introduces the utilities `_lazy_module` and `gen_types`. Both are meant to depend on 3rd party libraries without using direct imports. 

These have been used throughout the code, so `pandas` is no longer imported directly; this has significantly improved import time and, at the same time, set up the codebase to handle a world where `pandas` is not the only DataFrame library. Some functionality can likely be rewritten to not depend on `pandas`, but I will do that in future PRs.  

The `_lazy_module` has also "sneaked" into the Ibis data interface, which, in my opinion, makes it cleaner by not having all methods start with `import ibis`. 

Lastly, I have also removed the use of `lazy_isinstance` and deprecated it as the functionality can be done with `_lazy_module` and a bool check. This is done in `holoviews/operation/datashader.py`. 

### Hyperfine

Main (0ba7398ae61f15fb27e5066514d2bb04badcb0df)

``` bash
❯ hyperfine "python -c ''" "python -c 'import holoviews'" --warmup 5
Benchmark 1: python -c ''
  Time (mean ± σ):      16.0 ms ±   0.6 ms    [User: 12.0 ms, System: 3.9 ms]
  Range (min … max):    14.9 ms …  17.6 ms    171 runs

Benchmark 2: python -c 'import holoviews'
  Time (mean ± σ):     367.3 ms ±  11.7 ms    [User: 1938.8 ms, System: 57.8 ms]
  Range (min … max):   356.4 ms … 391.3 ms    10 runs

Summary
  python -c '' ran
   22.90 ± 1.10 times faster than python -c 'import holoviews'
```

This PR (https://github.com/holoviz/holoviews/pull/6476/commits/18f4db439ede2490b27f4eff3bbb602a0a79dcf8)
``` bash
❯ hyperfine "python -c ''" "python -c 'import holoviews'" --warmup 5
Benchmark 1: python -c ''
  Time (mean ± σ):      16.2 ms ±   0.6 ms    [User: 12.0 ms, System: 4.0 ms]
  Range (min … max):    14.9 ms …  19.1 ms    179 runs

Benchmark 2: python -c 'import holoviews'
  Time (mean ± σ):     189.7 ms ±   9.2 ms    [User: 1781.3 ms, System: 29.2 ms]
  Range (min … max):   182.5 ms … 216.8 ms    15 runs

Summary
  python -c '' ran
   11.74 ± 0.71 times faster than python -c 'import holoviews'
```

### Tuna

Running `python -X importtime -c 'import holoviews' 2> tuna.log && tuna tuna.log`

Main (0ba7398ae61f15fb27e5066514d2bb04badcb0df)

image

This PR (https://github.com/holoviz/holoviews/pull/6476/commits/18f4db439ede2490b27f4eff3bbb602a0a79dcf8)
image

<details><summary>Output of <code>hv.show_versions</code></summary>

``` yaml
Python              :  3.13.2 | packaged by conda-forge | (main, Feb 17 2025, 14:10:22) [GCC 13.3.0]
Operating system    :  Linux-6.13.5-arch1-1-x86_64-with-glibc2.41

holoviews           :  1.20.1.post1.dev7+g95d3af8b2

bokeh               :  3.6.3
colorcet            :  3.1.0
cudf                :  -
cupy                :  13.4.0
dask                :  2025.2.0
dask-expr           :  -
datashader          :  0.17.0.post1.dev5+gc3e4b58
geoviews            :  1.14.0.post1.dev2+gd89e2f7
hvplot              :  0.11.3.dev9+g9700bec.d20250301
ibis-framework      :  10.1.0
IPython             :  8.32.0
ipywidgets-bokeh    :  1.6.0
jupyter-bokeh       :  4.0.5
jupyterlab          :  4.3.5
matplotlib          :  3.10.1
networkx            :  3.4.2
notebook            :  -
numba               :  0.61.0
numpy               :  2.1.3
pandas              :  2.2.3
panel               :  1.6.1.post1.dev4+gf5170c481.d20250301
param               :  2.2.0.post1.dev22+gee19961
pillow              :  11.1.0
plotly              :  6.0.0
pyarrow             :  19.0.1
pyviz-comms         :  3.0.4
scikit-image        :  0.25.2
scipy               :  1.15.2
spatialpandas       :  0.5.0
streamz             :  -
tsdownsample        :  0.1.4.1
xarray              :  2025.1.2


```

</details>


holoviews/core/data/pandas.py
```diff
@@ -23,13 +23,26 @@ class PandasAPI:
             ...
     """
 
+pd = None  # Set when PandasInterface.applies is successful
```
hoxbro, line 26: Does not seem to work with test-gpu :man_shrugging:

holoviews/core/util/dependencies.py
```diff
@@ -0,0 +1,118 @@
...
+        self.__module_name = module_name
+        self.__package_name = package_name or module_name
+        self.__bool_use_sys_modules = bool_use_sys_modules
+
+    @property
+    def _module(self):
+        if self.__module is None:
+            try:
+                self.__module = import_module(self.__module_name)
+            except PackageNotFoundError:
```
maximlt, line 73: `importlib.import_module` raises `PackageNotFoundError`?

holoviews/core/util/dependencies.py
```diff
@@ -0,0 +1,118 @@
...
+
+        Parameters
+        ----------
+        module_name: str
+            The import name of the module, e.g. `import PIL`
+        package_name: str, optional
+            Name of the package, this is the named used for installing the package, e.g. `pip install pillow`.
+            Used for the __version__ if the module is not imported.
+            If not set uses the module_name.
+        bool_use_sys_modules: bool, optional, default False
```
maximlt, line 60: I get what that means but maybe add somewhere it means that bool can only be true if the module is imported.

holoviews/core/util/dependencies.py
```diff
@@ -0,0 +1,118 @@
...
+    return tuple(map(int, _re_no.findall(version_str)[:3]))
+
+
+_MIN_SUPPORTED_VERSION = {
+    "pandas": (1, 3, 0),
+}
+
+
+class _lazy_module:
+    __slots__ = ("__bool_use_sys_modules", "__module", "__module_name", "__package_name")
```
maximlt, line 44: Seems unnecessary.

holoviews/element/comparison.py
```diff
@@ -131,7 +131,9 @@ def register(cls):
         cls.equality_type_funcs[np.ma.masked_array]  = cls.compare_arrays
 
         # Pandas dataframe comparison
-        cls.equality_type_funcs[pd.DataFrame] = cls.compare_dataframe
+        if _is_installed("pandas"):
```
maximlt, line 134: Just noting it means that if one installs pandas after importing holoviews this compare_dataframe function won't get registered.

holoviews/operation/datashader.py
```diff
@@ -322,24 +328,24 @@ def get_agg_data(cls, obj, category=None):
         else:
             x, y = dims
 
+        bool_dd = bool(dd)  # Lazy module
         if len(paths) > 1:
             if glyph == 'line':
                 path = paths[0][:1]
-                if lazy_isinstance(path, "dask.dataframe:DataFrame"):
+                if bool_dd and isinstance(path, dd.DataFrame):
```
maximlt, line 335: Cannot this be just `if dd and instance(path, dd.DataFrame)`?

holoviews/core/util/dependencies.py
```diff
@@ -0,0 +1,118 @@
...
+    version_str = _get_version(package_name)
+    return tuple(map(int, _re_no.findall(version_str)[:3]))
+
+
+_MIN_SUPPORTED_VERSION = {
+    "pandas": (1, 3, 0),
+}
+
+
+class _lazy_module:
```
maximlt, line 43: Nit but could be callled `_Lazy_Module`.

holoviews/core/util/types.py
```diff
@@ -0,0 +1,116 @@
...
+    import numpy as np
+    import pandas as pd
+else:
+    cftime = _lazy_module("cftime", bool_use_sys_modules=True)
+    np = _lazy_module("numpy", bool_use_sys_modules=True)
+    pd = _lazy_module("pandas", bool_use_sys_modules=True)
+
+
+# gen_types is copied from param, can be removed when
+# we support 2.2 or greater
```
maximlt, line 19: Before doing that, I think `gen_types` should be documented/officially supported in Param. As far as I can see this wasn't done in https://github.com/holoviz/param/pull/966/files.

maximlt (approved):
I have a few comments but overall I'm good with these changes. Introducing this lazy import style doesn't guarantee code will run exactly as it used to (for example, in a notebook import holoviews, realize you miss some dependency, install it, keep running the code without restarting the kernel; or some patches). The pros outweigh the cons in this case imo.

holoviews/core/data/pandas.py
```diff
@@ -26,10 +32,23 @@ class PandasAPI:
...
+        # return 'pandas' in sys.modules
+        # 2025-02: As long as it is a required dependency and to not break
+        # existing behavior we will for now always return True
+        return bool(pd)
+
+    @classmethod
+    def applies(cls, obj):
+        if not cls.loaded():
+            return False
+        return isinstance(obj, (pd.DataFrame, pd.Series))
```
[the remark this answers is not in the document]
hoxbro: Yes. Though looking through the code again, the parent class has this in `applies`, so will remove the types. 

https://github.com/holoviz/holoviews/blob/5f4925cc0d8339d7fcc1045aa79c0dacdabac1c6/holoviews/core/data/interface.py#L151-L160

holoviews/core/util/__init__.py
```diff
@@ -2392,6 +2370,10 @@ def lazy_isinstance(obj, class_or_tuple):
         3) Do an isinstance check for dask.dataframe.DataFrame
 
     """
+    from ...util.warnings import deprecated
+
+    deprecated("1.22.0", "lazy_isinstance") # Not used in HoloViews anymore
```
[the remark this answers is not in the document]
hoxbro: Not sure. Have bumped to 1.23 for now. 

(Mostly added this here as a placeholder so I didn't forget that I wanted to deprecate it)

holoviews/core/util/dependencies.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __init__(self, msg, version=None, min_version=None, **kwargs):
+        self.version = version
+        self.min_version = min_version
+        super().__init__(msg, **kwargs)
+
+
+@lru_cache
+def _is_installed(module_name):
+    return find_spec(module_name) is not None
```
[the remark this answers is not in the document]
hoxbro: Yes, but as you say, the pros outweigh the cons.

holoviews/core/util/dependencies.py
hoxbro: Added more text below.

holoviews/core/util/types.py
hoxbro: I think this is a very library-related need; if you have an app, you are likely to do an import of the module and use `isinstance` or `issubclass` as you always have. 

I think we should document it, but I want to be very careful about it, as it could be more confusing than helpful.

holoviews/element/comparison.py
hoxbro: Yes. 

I have never been a massive fan of this file, or more specifically, its location. I would have liked it to be in a testing module.

holoviews/operation/datashader.py
hoxbro: It is done this way so we don't check `sys.modules` multiple times, if it has been installed but not imported.

holoviews/core/util/dependencies.py
hoxbro: Have renamed to `_LazyModule`

holoviews/core/util/dependencies.py
hoxbro: It has been removed. 

image

holoviews/operation/datashader.py
maximlt: If it's done this way for performance reasons, maybe write it down, as someone refactoring the code would be likely to clean this up.

maximlt (approved):

holoviews/operation/datashader.py
```diff
@@ -322,24 +328,24 @@ def get_agg_data(cls, obj, category=None):
         else:
             x, y = dims
 
+        bool_dd = bool(dd)  # Lazy module
```
hoxbro, line 331:
```suggestion
        bool_dd = bool(dd)  # NOTE: A lazy module and this avoids repeated lookups in `sys.modules`. 
```

holoviews/operation/datashader.py
hoxbro: I updated the text above to clarify this.

holoviews/operation/datashader.py
```diff
@@ -322,24 +328,24 @@ def get_agg_data(cls, obj, category=None):
         else:
             x, y = dims
 
+        bool_dd = bool(dd)  # NOTE: A lazy module and this avoids repeated lookups in `sys.modules`.
```
hoxbro, line 331:
```suggestion
        bool_dd = bool(dd)  # NOTE: A lazy module and this avoids repeated lookups in `sys.modules`.
```

Issue #6482, opened 2025-01-07
The "shellcheck" pre-commit hook fails with git configured with "autocrlf = true"


stanwest:
### Software version info

I observed this with the current "main" branch but expect that the issue pertains to v. 1.18.2a1 and later from #5991 and commit ba29c24.

### Description of expected behavior and the observed behavior

With the repository checked out with *git*'s `core.autocrlf` setting configured to `auto` (as is often the case on a Windows platform), linting fails. Specifically, the `shellcheck` pre-commit hook executed by `pixi run lint` rejects the carriage returns in the file "scripts/conda/build.sh".

### Complete, minimal, self-contained example code that reproduces the issue

Output of `pixi run lint` with "[...]" standing for unimportant lines:

```plaintext
Pixi task (lint in lint): pre-commit run --all-files
[...]
shellcheck...............................................................Failed
- hook id: shellcheck
- exit code: 1

In scripts/conda/build.sh line 1:
#!/usr/bin/env bash
                   ^-- SC1017 (error): Literal carriage return. Run script through tr -d '\r' .


In scripts/conda/build.sh line 2:
[...]
```

- [x] I may be interested in making a pull request to address this

[Edit: Added reference to PR 5991.]

PR #6484, opened 2025-01-07
fix(dev): Avoid checking out CRLF in shell scripts

stanwest:
This PR resolves issue #6482 by instructing *git* to check out "*.sh" files with LF line endings.

A shell script previously checked out with CRLF endings will need its endings changed by other means for linting to pass.


hoxbro:
Thank you.

PR #6492, opened 2025-01-16
fix: support the case of multiple line segments each with one vdim

grapesmoker:
Previously `Path` would accept data that was structured as a list of dictionaries, each dictionary representing a single line segment. Such lists would plot correctly but if you had any `vdims` defined on the dictionary, a warning would be raised when trying to create a `ColumnDataSource` because the `_element_transform` function would make it so that the list of vdims and the list of coordinates had different lengths. This PR fixes this issue, so a list of dictionaries, each with its own coordinates and vdims, is now allowed. The below screenshot illustrates the desired plot.

Screenshot 2025-01-16 at 4 42 01 PM

### DIFF
--- a/holoviews/plotting/bokeh/path.py
+++ b/holoviews/plotting/bokeh/path.py
@@ -57,7 +57,7 @@ class PathPlot(LegendPlot, ColorbarPlot):
         for el in element.split():
             new_el = transform.apply(el, ranges=ranges, flat=True)
             if len(new_el) == 1:
-                kdim_length = max([len(el[kdim]) for kdim in el.kdims])
+                kdim_length = len(el[el.kdims[0]])
                 transformed.append(np.tile(new_el, kdim_length - 1))
             else:
                 transformed.append(new_el)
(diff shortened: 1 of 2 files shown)


holoviews/plotting/bokeh/path.py
```diff
@@ -52,8 +52,17 @@ def _element_transform(self, transform, element, ranges):
...
                     new_data.append(d)
             return np.array(new_data)
-        return np.concatenate([transform.apply(el, ranges=ranges, flat=True)
-                               for el in element.split()])
+
+        transformed = []
+        for el in element.split():
+            new_el = transform.apply(el, ranges=ranges, flat=True)
+            if len(new_el) == 1:
+                kdim_length = max([len(el[kdim]) for kdim in el.kdims])
```
hoxbro, line 60:
```suggestion
                kdim_length = len(el[el.kdims[0]])
```

Can the first dim be used for the length? If it needs to check all kdims, it would be nice to have a test for that.

grapesmoker: Yes, I think we can use the first kdim. Is it safe to assume that the kdims values at this point are equal length?

hoxbro:
Thank you for the PR :+1:

PR #6497, opened 2025-01-27
enh(dev): Don't force holoviews to be installed for test-example and docs-build

hoxbro:
Right now, if we want to `pixi run test-example` and `pixi run docs-build`, it will fail because 'HoloViews' is not installed, and for new developers give a weird error message like `holoviews is not installed`. It is stated in the docs that you need to run `pixi run install` before doing anything, but people forget or don't read the docs. 

I can't currently think of a reason why this should be done, as holoviews is doing anything weird as installing/downloading doing the build step.

### DIFF
--- a/examples/conftest.py
+++ b/examples/conftest.py
@@ -7,14 +7,12 @@ import bokeh
 import pandas as pd
 from packaging.version import Version
 
+# Setting this to not error out if no install is done.
+os.environ["PYTHONPATH"] = os.path.dirname(os.path.dirname(__file__))
+
 system = platform.system()
 py_version = sys.version_info[:2]
 PANDAS_GE_2_0_0 = Version(pd.__version__).release >= (2, 0, 0)
-
-# Having "OMP_NUM_THREADS"=1, set as an environment variable, can be needed
-# to avoid crashing when running tests with pytest-xdist on Windows.
-# This is set in the .github/workflows/test.yaml file.
-# https://github.com/holoviz/holoviews/pull/5720
 
 collect_ignore_glob = [
     # Needs selenium, phantomjs, firefox, and geckodriver to save a png picture
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -13,6 +13,9 @@ authors = 'HoloViz developers'
 copyright = '2005 ' + authors
 description = 'Stop plotting your data - annotate your data and let it visualize itself.'
 
+# Setting this to not error out if no install is done
+sys.path.insert(0, os.path.dirname(os.path.dirname(__file__)))
+
 import holoviews
 version = release = base_version(holoviews.__version__)
 
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -14,7 +14,9 @@ copyright = '2005 ' + authors
 description = 'Stop plotting your data - annotate your data and let it visualize itself.'
 
 # Setting this to not error out if no install is done
-sys.path.insert(0, os.path.dirname(os.path.dirname(__file__)))
+root_path = os.path.dirname(os.path.dirname(__file__))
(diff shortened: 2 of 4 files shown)


examples/conftest.py
```diff
@@ -7,15 +7,13 @@
...
 from packaging.version import Version
 
+# Setting this to not error out if no install is done.
+os.environ["PYTHONPATH"] = os.path.dirname(os.path.dirname(__file__))
+
 system = platform.system()
 py_version = sys.version_info[:2]
 PANDAS_GE_2_0_0 = Version(pd.__version__).release >= (2, 0, 0)
 
-# Having "OMP_NUM_THREADS"=1, set as an environment variable, can be needed
```
hoxbro, line 14: This is currently set in the pixi_install task and no longer in the test.yaml itself.

maximlt (commented):
Wondering if prefixing the commands (`pytest ...`, `nbsite ...`) with `python -m` wouldn't do the trick?

FWIW I've done it a bit differently in hvPlot with the following as I prefer to make sure the package is installed.

```toml
[feature.required.tasks]
download-data = 'python scripts/download_data.py'
install = 'python -m pip install --no-deps --disable-pip-version-check -e .'
install-dev = { cmd = "pre-commit install", depends-on = ["install", "download-data"] }
```

and

```toml
[feature.doc.tasks]
docs-build = 'sphinx-build -b html doc builtdocs'
# docs-build-dev = { cmd = "pre-commit install", depends-on = ["install", "docs-build"] }
_docs-install = 'python -m pip install --no-deps --disable-pip-version-check -e .'
docs-build-dev = { depends-on = ["_docs-install", "docs-build"] }
```

hoxbro:
> Wondering if prefixing the commands (pytest ..., nbsite ...) with python -m wouldn't do the trick?

I tested it locally, and it did not do the trick. 

> FWIW I've done it a bit differently in hvPlot with the following as I prefer to make sure the package is installed.

I like the idea behind the `install-dev` command; I would maybe call it something like `setup-dev`, as I don't think it only needs to install stuff, and some would argue that downloading data is already more than what the name suggests.  

I don't like `docs-build-dev` mainly because it can be confusing to have two so similar tasks. For example, should I use `docs-build` or `docs-build-dev`? Oh, `docs-build` did not work at first, and `docs-build-dev` did. I would rather see 100% commitment to ensure the package is installed, so there only is `docs-build`, which does what `docs-build-dev` currently does.  Though, this discussion is out of scope for this PR :upside_down_face:

maximlt:
> I tested it locally, and it did not do the trick.

Ah ok that's a little disappointing 🙃 

> I like the idea behind the `install-dev` command; I would maybe call it something like `setup-dev`, as I don't think it only needs to install stuff, and some would argue that downloading data is already more than what the name suggests.

Yes `setup-dev` is better.

> I don't like `docs-build-dev` mainly because it can be confusing to have two so similar tasks. For example, should I use `docs-build` or `docs-build-dev`? Oh, `docs-build` did not work at first, and `docs-build-dev` did. I would rather see 100% commitment to ensure the package is installed, so there only is `docs-build`, which does what `docs-build-dev` currently does. Though, this discussion is out of scope for this PR 🙃

Yes also fine with having `docs-build` always install the package before building the docs. In my opinion this is just a workaround for now until hopefully `pixi` gains better support for developing a Python package.

hoxbro:
> Is it? You could already make these changes in this PR if you prefer, to avoid the workaround it implements.

Both methods should be seen as workarounds. I prefer my method. I still sometimes run `pixi run -e test-39 test-example` (or something similar). Installing is a different way to add the package to the path, but it has much more overhead. 

I will add a `setup-dev` task.

PR #6498, opened 2025-01-27
refactor: Use ruff on all python files

hoxbro:


hoxbro:
Failing tests are because of https://github.com/holoviz/holoviews/pull/6509 and https://github.com/holoviz/panel/pull/7715

Issue #6516, opened 2025-02-20
opts keyword regression between 1.20.0 and 1.20.1 release


dwr-psandhu:
### Software version info
Change in behavior between 1.20.0 and 1.20.1 releases. All other packages are the latest on conda-forge (python 3.13, etc)

### Description of expected behavior and the observed behavior

Code works fine at 1.20.0 but fails in 1.20.1 version

### Minimal example not easy 
I tried but could not get a minimal example to fail. Hope to construct one as I get more clues.

I am getting this error stack trace for this piece of code
```
... # some class defining the function 
    def get_name_to_color(self):
        """return a dictionary mapping column names to color names"""
        return hv.Cycle("Category10").values
...
            self.map_features = self.map_features.opts(
                color=dim(self.map_color_category).categorize(
                    self.dataui_manager.get_name_to_color(), default="blue"
                )
            )

```

The error that appears only in 1.20.1 is

```
  File "xxxx\Lib\site-packages\holoviews\plotting\bokeh\element.py", line 2110, in _init_glyphs
    renderer, glyph = self._init_glyph(plot, mapping, properties)
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "xxxx\Lib\site-packages\holoviews\plotting\bokeh\element.py", line 2963, in _init_glyph
    ret = super()._init_glyph(plot, mapping, properties)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "xxxx\Lib\site-packages\holoviews\plotting\bokeh\element.py", line 1800, in _init_glyph
    renderer = getattr(plot, plot_method)(**dict(properties, **mapping))
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "xxxx\Lib\site-packages\bokeh\plotting\glyph_api.py", line 1151, in scatter
    return self._scatter(*args, marker=marker_type, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
<<<<<< long stack trace removed >>>>>
 File "xxxx\Lib\site-packages\bokeh\core\property\bases.py", line 363, in prepare_value
    raise ValueError(f"failed to validate {obj_repr}.{name}: {error}")
[...]

PR #6519, opened 2025-02-21
fix: regression in get_value_array

hoxbro:
Fixes https://github.com/holoviz/holoviews/issues/6516

The performance improvement done in https://github.com/holoviz/holoviews/pull/6470 introduced a bug. The bug came when trying to bypass the for loop, which was done with a `.unique` call of the dataset. This bug caused the following regression. In short, when using dim-expression, the sensor's output would now be two-sized lengths instead of three-sized lengths like the rest of the data, resulting in an error when converting to Bokeh valid data.  

This PR replaces the unique call with a conversion to a numpy array, keeping the data size. 

``` python
import geopandas as gpd
import geoviews as gv
import pandas as pd

import holoviews as hv

hv.extension("bokeh")

df = pd.DataFrame(
    {"Longitude": [0, 1, 2], "Latitude": [0, 1, 2], "Sensor": ["S1", "S2", "S1"]}
)
gdf = gpd.GeoDataFrame(df, geometry=gpd.points_from_xy(df["Longitude"], df["Latitude"]))

gv.Points(gdf).opts(line_color=hv.dim("Sensor").categorize(["red", "blue", "yellow"]))

```

### DIFF
--- a/holoviews/core/data/spatialpandas.py
+++ b/holoviews/core/data/spatialpandas.py
@@ -643,9 +643,7 @@ def get_value_array(data, dimension, expanded, keep_index, geom_col,
     is_scalars = [isscalar(value) for value in column]
     all_scalar = all(is_scalars)
     if all_scalar and not expanded:
-        scal_unique = column.unique()
-        # .unique() on categorical dtype doesn't return a np array
-        return scal_unique if isinstance(scal_unique, np.ndarray) else scal_unique.to_numpy()
+        return np.asarray(column)
     arrays, scalars = [], []
     for i, geom in enumerate(data[geom_col]):
         val = column.iloc[i]
--- a/holoviews/tests/core/data/test_spatialpandas.py
+++ b/holoviews/tests/core/data/test_spatialpandas.py
@@ -4,6 +4,7 @@ Tests for the spatialpandas interface.
 from unittest import SkipTest
 
 import numpy as np
+import pandas as pd
 
 try:
     import spatialpandas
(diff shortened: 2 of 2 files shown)


holoviews/tests/core/data/test_spatialpandas.py
```diff
@@ -301,3 +304,14 @@ def test_dict_dataset_add_dimension_values(self):
 
     def test_sort_by_value(self):
         raise SkipTest("Not supported")
+
+
+
+def test_regression_get_value_array():
+    # See: https://github.com/holoviz/holoviews/pull/6519
```
maximlt, line 311: Do you prefer to link to the PR or to the issue?

holoviews/tests/core/data/test_spatialpandas.py
```diff
@@ -301,3 +304,14 @@ def test_dict_dataset_add_dimension_values(self):
 
     def test_sort_by_value(self):
         raise SkipTest("Not supported")
+
+
+
+def test_regression_get_value_array():
```
maximlt, line 310: Should this test run if spatialpandas is not installed?

holoviews/core/data/spatialpandas.py
```diff
@@ -643,9 +643,7 @@ def get_value_array(data, dimension, expanded, keep_index, geom_col,
     is_scalars = [isscalar(value) for value in column]
     all_scalar = all(is_scalars)
     if all_scalar and not expanded:
-        scal_unique = column.unique()
-        # .unique() on categorical dtype doesn't return a np array
-        return scal_unique if isinstance(scal_unique, np.ndarray) else scal_unique.to_numpy()
+        return np.asarray(column)
```
maximlt, line 646: I guess I got somehow confused with the meaning of `expanded` but yes that looks correct.

maximlt (approved):
LGTM overall. I find it easier to review PRs that describe what they do/what the issue was. Philipp does that pretty well for example.

holoviews/tests/core/data/test_spatialpandas.py
hoxbro: I don't think it hurts, as the test contains no spatialpandas code.

holoviews/tests/core/data/test_spatialpandas.py
hoxbro: It's definitely a PR. A good PR should contain all the relevant information. I have updated the main post with more  information.

PR #6529, opened 2025-03-07
compat: Bokeh 3.7

hoxbro:
Two failing UI tests.

### DIFF
--- a/holoviews/tests/ui/bokeh/test_callback.py
+++ b/holoviews/tests/ui/bokeh/test_callback.py
@@ -5,7 +5,7 @@ import pytest
 
 import holoviews as hv
 from holoviews import Curve, DynamicMap, Scatter
-from holoviews.plotting.bokeh.util import BOKEH_GE_3_4_0
+from holoviews.plotting.bokeh.util import BOKEH_GE_3_4_0, BOKEH_GE_3_7_0
 from holoviews.streams import (
     BoundsX,
     BoundsXY,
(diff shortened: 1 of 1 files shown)


holoviews/tests/ui/bokeh/test_callback.py
```diff
@@ -220,8 +220,12 @@ def test():
         assert s.xs == {'x': np.datetime64('2024-01-12T13:26:44.819277')}
         assert s.xs == {'x': np.datetime64('2024-01-12T13:26:44.819277')}
         assert len(s.ys) == 2
-        assert np.isclose(s.ys["y1"], 18.130705394191)
-        assert np.isclose(s.ys["y2"], 76.551867219917)
+        if BOKEH_GE_3_7_0:
```
hoxbro, line 223: This is because the datetime x-axis has changed to add the year, squeezing the values a bit. 
| Before | After |
| --- | --- |
| Screenshot From 2025-03-07 11-38-23 | Screenshot From 2025-03-07 11-37-49 |

holoviews/tests/ui/bokeh/test_callback.py
```diff
@@ -414,6 +418,7 @@ async def popup_form(x, y):
         locator = page.locator(".bk-btn")
         expect(locator).to_have_count(2)
 
+    @pytest.mark.filterwarnings("reference already known")
```
hoxbro, line 421: I tried to see if I could figure out where this was coming from but without luck. It is the button ID that is flagged.

``` python
import numpy as np
import panel as pn

import holoviews as hv

hv.extension("bokeh")

rng = np.random.default_rng(10)
points =  hv.Points(rng.normal(size=(1000, 2)))

def popup_form(x, y):
    def hide(_):
        button.visible = False

    button = pn.widgets.Button(name=f"{x},{y}", on_click=hide)
    return button

points.opts(tools=["tap"])
hv.streams.Tap(source=points, popup=popup_form)

pn.panel(points).servable()
```

holoviews/tests/ui/bokeh/test_callback.py
```diff
@@ -414,6 +418,7 @@ async def popup_form(x, y):
         locator = page.locator(".bk-btn")
         expect(locator).to_have_count(2)
 
+    @pytest.mark.filterwarnings("ignore:reference already known")
```
hoxbro, line 421: I tried to see if I could figure out where this was coming from but without luck. It is the button ID that is flagged.

``` python
import numpy as np
import panel as pn

import holoviews as hv

hv.extension("bokeh")

rng = np.random.default_rng(10)
points =  hv.Points(rng.normal(size=(1000, 2)))

def popup_form(x, y):
    def hide(_):
        button.visible = False

    button = pn.widgets.Button(name=f"{x},{y}", on_click=hide)
    return button

points.opts(tools=["tap"])
hv.streams.Tap(source=points, popup=popup_form)

pn.panel(points).servable()
```

hoxbro: @philippjfr, you are welcome to give it a go.

PR #6531, opened 2025-03-10
enh: Add format to Server HoverTool and improve datetime support

hoxbro:
``` python
import datetime as dt

import datashader as ds
import numpy as np
import pandas as pd
import panel as pn

import holoviews as hv

hv.extension("bokeh")

num = 10000
seed = np.random.default_rng(1)


dists = {
    cat: pd.DataFrame(
        {
            "x": seed.normal(x, s, num),
            "y": seed.normal(y, s, num),
            "s": s,
            "val": val,
            "cat": cat,
            "dt": dt.datetime(2020, 1, val + 1),
        }
    )
    for x, y, s, val, cat in [
        (2, 2, 0.03, 0, "d1"),
        (2, -2, 0.10, 1, "d2"),
        (-2, -2, 0.50, 2, "d3"),
        (-2, 2, 1.00, 3, "d4"),
        (0, 0, 3.00, 4, "d5"),
    ]
}

df["y"] += 100  # Offset to distinguish x from y
points = hv.Points(df)

plot = rasterize(points, aggregator=ds.min("s"), selector=ds.min("s")).opts(tools=["hover"])
pn.panel(plot).servable()
```


hoxbro:
Opened an issue in Bokeh: https://github.com/bokeh/bokeh/issues/14413

hoxbro:
We should merge this PR as is, as the datetime fix will likely sort itself with an upstream fix.

holoviews/plotting/bokeh/raster.py
```diff
@@ -92,9 +103,10 @@ def on_change(attr, old, new):
             if np.isinf(new).all():
                 return
             data_sel = self._hover_data.sel(**dict(zip(self._hover_data.coords, new)), method="nearest").to_dict()
-            # TODO: When ValueOf support formatter remove the rounding
-            # https://github.com/bokeh/bokeh/issues/14123
-            data_coords = {dim: round(data_sel['coords'][dim]['data'], 3) for dim in coords}
+            if BOKEH_GE_3_7_0:
+                data_coords = {dim: data_sel['coords'][dim]['data'] for dim in coords}
+            else:
+                data_coords = {dim: round(data_sel['coords'][dim]['data'], 3) for dim in coords}
```
hoxbro, line 109: It would be OK with me to only have the `BOKEH_GE_3_7_0` code.

philippjfr: What's the behavior for older bokeh versions?

hoxbro: This is only for the x,y-coords, so it would show `1.999999999999` instead of `2.0`. 

Though, I'm unsure how it would work if x,y-coords are datetime.

hoxbro: It errored out when we had datetime as x,y coords. I have fixed this in https://github.com/holoviz/holoviews/pull/6531/commits/3ae5e869f34dc026dead22e661f463bb13dfd447 and also removed the round here.

philippjfr (approved):
Apart from the failing tests this looks good.

PR #6534, opened 2025-03-13
fix: PolyEdit stream not synchronizes with CDS

### LINKED ISSUE #6532: ValueError when updating polyline using PolyEdit tool
https://github.com/holoviz/holoviews/blob/53e904958d5c07b467b4d8ab3683c1c1df30263b/holoviews/plotting/bokeh/callbacks.py#L1410

When updating polyline using PolyEdit tool this line results in the following error: 

`ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`

The reason for it is in case of EditTool value[2] is a list of numpy arrays, not a str, so quick fix would be to add extra condition to this elif:

`
elif (
    isinstance(values, list)
    and len(values) == 4
    and isinstance(values[2], str )  # Without this line bokeh failes to process updated polylines
    and values[2] in ("big", "little")
    and isinstance(values[3], list)
):
`

MikalaiPrakapenka:
Fixes #6532

In case of using PolyEdit tool value[2] is a list of lists or numpy arrays, not a str.

### DIFF
--- a/holoviews/plotting/bokeh/callbacks.py
+++ b/holoviews/plotting/bokeh/callbacks.py
@@ -1407,6 +1407,7 @@ class CDSCallback(Callback):
             elif (
                 isinstance(values, list)
                 and len(values) == 4
+                and isinstance(values[2], str)
                 and values[2] in ("big", "little")
                 and isinstance(values[3], list)
             ):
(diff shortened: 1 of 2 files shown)


hoxbro:
Thank you for the PR. Really nice you added a test! :+1:

MikalaiPrakapenka:
> Thank you for the PR. Really nice you added a test! 👍

Thank you ☺️.

PR #6541, opened 2025-03-14
ci: Remove old pages in docs publish

### LINKED ISSUE #6487: Remove stale website pages
The HoloViews web server provides some confusingly out of date pages like https://holoviews.org/user_guide/Styling_Plots.html that have changed names or been reorganized into pages like https://holoviews.org/user_guide/Style_Mapping.html .  People clicking around at holoviews.org probably can't get to that material, but it still shows up in Google and appears to be curreyt, yet it has ancient advice like using `.options`. We might be keeping that old material around so that links stay working, but to me it seems like it should be a redirect to a current page to avoid giving people bad advice.

hoxbro:
Fixes #6487

Test: this should not exists: https://dev.holoviews.org/user_guide/Styling_Plots.html

From: https://docs.aws.amazon.com/cli/latest/reference/s3/sync.html
```
--delete (boolean) Files that exist in the destination but not in the source are deleted during sync. Note that files excluded by filters are excluded from deletion.
```

Needed to add: `"s3:DeleteObject"` on the dev.holoviews.org bucket for it to work (ref: https://stackoverflow.com/a/41983721)

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -78,10 +78,10 @@ jobs:
           (github.event_name == 'workflow_dispatch' && github.event.inputs.target == 'dev') ||
           (github.event_name == 'push' && (contains(steps.vars.outputs.tag, 'a') || contains(steps.vars.outputs.tag, 'b') || contains(steps.vars.outputs.tag, 'rc')))
         run: |
-          aws s3 sync --quiet --delete ./builtdocs s3://dev.holoviews.org/
+          aws s3 sync --delete ./builtdocs s3://dev.holoviews.org/
       - name: upload main
         if: |
           (github.event_name == 'workflow_dispatch' && github.event.inputs.target == 'main') ||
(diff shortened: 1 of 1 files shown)


hoxbro:
It took 40ish minutes to run the first time, and the second time, it went down to 30 seconds.

It looks like the bucket had a size of 46 GB and 250.000 files.

hoxbro:
Seems like the page has been removed, but is still cached somewhere:

``` bash
❯ curl https://dev.holoviews.org/user_guide/Styling_Plots.html
<html>
<head><title>404 Not Found</title></head>
<body>
<h1>404 Not Found</h1>
<ul>
<li>Code: NoSuchKey</li>
<li>Message: The specified key does not exist.</li>
<li>Key: user_guide/Styling_Plots.html</li>
<li>RequestId: HGYP3561DV962DP9</li>
<li>HostId: f7XVUbE63ZbGauIehU7kktTvhvDANAs56A/e6td1lxPhqMKbmFasQuW4iIjEeOvUCwwFfZUBNgk=</li>
</ul>
<hr/>
</body>
</html>
```

hoxbro:
The page is no longer cached. Some stats for the bucket, before and after the clean up:

Screenshot From 2025-03-17 08-46-44


Screenshot From 2025-03-17 08-46-55

Issue #5111, opened 2021-11-05
TypeError when rendering a Holomap with Matplotlib whose data is a Dask DataFrame


maximlt:
Rendering an histogram whose source is a grouped-by dask array raises an error (`TypeError: len() of unsized object`) with the **Matplotlib** backend (works fine with Bokeh and Plotly).

```python
import dask.dataframe as dd
import holoviews as hv
from holoviews.operation import histogram

hv.extension('matplotlib')

flights = dd.read_parquet("https://assets.holoviews.org/data/airline_flights.parq").persist()
flight_subset = flights[flights.carrier.isin(['AA', 'US', 'OH'])]
by = 'carrier'

ds = hv.Dataset(flight_subset, by)
print(ds)

ds_grouped = ds.groupby(by)
print(ds_grouped)

hists = histogram(ds_grouped, dimension='depdelay')
print(hists)

hists
```

```python

---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
~/miniconda3/envs/hvplot-dev37/lib/python3.7/site-packages/IPython/core/formatters.py in __call__(self, obj, include, exclude)
    968 
    969             if method is not None:
--> 970                 return method(include=include, exclude=exclude)
    971             return None
    972         else:

~/miniconda3/envs/hvplot-dev37/lib/python3.7/site-packages/holoviews/core/dimension.py in _repr_mimebundle_(self, include, exclude)
   1314         combined and returned.
   1315         """
-> 1316         return Store.render(self)
   1317 
   1318 

~/miniconda3/envs/hvplot-dev37/lib/python3.7/site-packages/holoviews/core/options.py in render(cls, obj)
   1403         data, metadata = {}, {}
   1404         for hook in hooks:
-> 1405             ret = hook(obj)
   1406             if ret is None:
   1407                 continue

~/miniconda3/envs/hvplot-dev37/lib/python3.7/site-packages/holoviews/ipython/display_hooks.py in pprint_display(obj)
    280     if not ip.display_formatter.formatters['text/plain'].pprint:
    281         return None
--> 282     return display(obj, raw_output=True)
    283 
    284 

~/miniconda3/envs/hvplot-dev37/lib/python3.7/site-packages/holoviews/ipython/display_hooks.py in display(obj, raw_output, **kwargs)
    256     elif isinstance(obj, (HoloMap, DynamicMap)):
    257         with option_state(obj):
--> 258             output = map_display(obj)
[...]

PR #6542, opened 2025-03-17
fix: Histogram error with datatype `dask.array` and backend Matplotlib

hoxbro:
Fixes https://github.com/holoviz/holoviews/issues/5111

We call compute on edges if it is a `dask.array`, so it doesn't error later down the pipeline. A similar thing is done in the Bokeh backend:

https://github.com/holoviz/holoviews/blob/ea8e8057d6213843fcd1d9a9a3fc39e3bafd841e/holoviews/plotting/bokeh/chart.py#L444-L445

### DIFF
--- a/holoviews/tests/operation/test_operation.py
+++ b/holoviews/tests/operation/test_operation.py
@@ -603,7 +603,7 @@ class OperationTests(ComparisonTestCase):
             "carrier": ["A", "A", "A", "A", "B", "B", "B", "B", "B"],
             "depdelay": [127.0, 3.0, -3.0, 19.0, 264.0, -6.0, 1.0, 2.0, 83.0],
         }
-        flights = dd.from_pandas(pd.DataFrame(data))
+        flights = dd.from_pandas(pd.DataFrame(data), npartitions=2)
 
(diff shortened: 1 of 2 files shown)


holoviews/tests/operation/test_operation.py
```diff
@@ -593,6 +594,25 @@ def test_points_histogram_mean_weighted(self):
...
+        }
+        flights = dd.from_pandas(pd.DataFrame(data), npartitions=2)
+
+        by = "carrier"
+        ds = Dataset(flights, by)
+        ds_grouped = ds.groupby(by)
+        hists = histogram(ds_grouped, dimension="depdelay")
+
+        # Should not error
+        pn.panel(hists).get_root()
```
maximlt, line 614: Just out of curiosity, why using Panel?

maximlt (approved):

holoviews/tests/operation/test_operation.py
hoxbro: I couldn't get it to work yesterday without it, but today I can... Pushed an updating using `hv.renderer`

Issue #6561, opened 2025-04-20
Streams don't work with the reactive API (`.rx`)


MatusGasparik:
#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc should be added within the dropdown below.)

<details>
    <summary>Software Version Info</summary>

```plaintext
holoviews: 1.20.2
bokeh: 3.7.2
panel: 1.6.2
param: 2.2.0
python: 3.13
```
</details>

#### Description of expected behavior and the observed behavior
I am trying to recreate [this example](https://holoviews.org/user_guide/Responding_to_Events.html#using-parameterized-classes-as-a-stream), but using `param.Parameterized` classes that make use of the `pn.rx` API (similar to [this example](https://panel.holoviz.org/tutorials/intermediate/interactivity.html#making-the-class-based-approach-efficient)), but I haven't been able to make it work.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
class P(param.Parameterized):
    n = param.Integer(10, bounds=(1, 1000))
    n_rx = param.Parameter()

    def __init__(self, **params):
        super().__init__(**params)
        self.n_rx = self.param.n.rx() * 10  # reactive expr.

p = P()
slider = pn.widgets.IntSlider.from_param(p.param.n)

def plot(n):
    return hv.Scatter(np.random.random((n, 2)))
```

This works as expected:

```python
dmap = hv.DynamicMap(plot, streams=dict(n=p.param.n))

pn.Column(slider, dmap)
```

But this fails:

```python
dmap = hv.DynamicMap(plot, streams=dict(n=p.param.n_rx))

pn.Column(slider, dmap)
```
eventhough `type(p.param.n_rx)` returns `param.parameterized.Parameter`.

#### Stack traceback and/or browser JavaScript console output

<details>
    <summary>Traceback</summary>

```plaintext
WARNING:param.plot: Callable raised "TypeError("'rx' object cannot be interpreted as an integer")".
Invoked as plot(n=<param.reactive.rx object at 0x32c53d220>)
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
[...]

PR #6563, opened 2025-04-22
enh: Add basic reactive expression support in a streams dictionary

hoxbro:
Resolves https://github.com/holoviz/holoviews/issues/6561

For simplicity, I allow the dict list to be either parameters or reactive expressions. There isn't any technical reason not to allow this, but the order should be correct if we allow both.

 ``` python
import param
import panel as pn
import numpy as np
import holoviews as hv

hv.extension("bokeh")

class P(param.Parameterized):
    n = param.Integer(10, bounds=(1, 1000))

    def __init__(self, **params):
        super().__init__(**params)
        self.n_rx = self.param.n.rx() * 10  # reactive expr.

p = P()
slider = pn.widgets.IntSlider.from_param(p.param.n)

def plot(n):
    return hv.Scatter(np.random.random((n, 2)))

dmap = hv.DynamicMap(plot, streams=dict(n=p.n_rx))
pn.Column(slider, dmap)
```

https://github.com/user-attachments/assets/0a5ccb73-6435-4086-aa51-05c1c69ac23e

# TODO: 
Add test + document this currently only work with one or the other

### DIFF
--- a/holoviews/tests/core/test_dynamic.py
+++ b/holoviews/tests/core/test_dynamic.py
@@ -78,11 +78,6 @@ class DynamicMapConstructor(ComparisonTestCase):
     def test_simple_constructor_streams_dict_class_parameter(self):
         DynamicMap(lambda x: x, streams=dict(x=ExampleParameterized.param.example))
 
-    def test_simple_constructor_streams_dict_invalid(self):
-        regexp = "Cannot handle value 3 in streams dictionary"
-        with self.assertRaisesRegex(TypeError, regexp):
-            DynamicMap(lambda x: x, streams=dict(x=3))
-
     def test_simple_constructor_streams_invalid_uninstantiated(self):
         regexp = ("The supplied streams list contains objects "
                   "that are not Stream instances:(.+?)")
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by philippjfr** [comment]
Problem is that this still doesn't work, i.e. the value that's being returned is the raw parameter value not the transformed value. You can see that when the widget is 10 there aren't 100 points.

File: holoviews/tests/core/test_dynamic.py [hunk 1]
```diff
@@ -78,11 +78,6 @@ def test_simple_constructor_streams_dict_parameter(self):
     def test_simple_constructor_streams_dict_class_parameter(self):
         DynamicMap(lambda x: x, streams=dict(x=ExampleParameterized.param.example))
 
-    def test_simple_constructor_streams_dict_invalid(self):
-        regexp = "Cannot handle value 3 in streams dictionary"
-        with self.assertRaisesRegex(TypeError, regexp):
-            DynamicMap(lambda x: x, streams=dict(x=3))
```
hoxbro, line 84: Why does this not work anymore?

philippjfr: I'm allowing constants in the refs dictionary. I think it's convenient in the same way a partial is convenient but I guess I'd be okay restoring an error if all values are constant.

hoxbro: I'm unsure whether the extra convenience is worth the potential foot gun.  

I will trust your judgment, so please do what you think is best.

philippjfr: Agreed, passing the parameter value rather than a parameter object is a frequent source of confusion so we shouldn't make that more likely.

philippjfr: Fixed.

Issue #6565, opened 2025-04-23
Tooltip with aggregator and selector does not bring enough context


maximlt:
While reviewing https://github.com/holoviz/holoviews/pull/6553, I noticed that the hover tooltip displayed by default when `aggregator` and `selector` are used doesn't display enough information about the nature of the aggregators used. For example, have a look at this screenshot of a rasterized plot made using both `aggregator` and `selector`, and try to guess which kind of aggregator and selector was used.

<img width="850" alt="Image" src="https://github.com/user-attachments/assets/3d99832b-c9b9-4d87-8581-f7fdd066e57a" />

It seems to me that guessing correctly is very hard, except for the person who has created the plot, or if the viewer also has access to the code:

<img width="1377" alt="Image" src="https://github.com/user-attachments/assets/d5481a15-206b-4d41-8768-f645f10e6f0c" />

In this case:
- `x_y s` is computed by the `aggregator` and represents the mean value of `s` in that pixel
- `s`, `val` and `cat` are computed by the `selector` and are the values in that pixel where `s` is minimum

None of these tooltip labels describes how the value was actually calculated. At the moment, it seems like a user can only hide some labels (setting `hover_tooltips`), but not rename them to make them clearer. I think we should fix/improve that.

droumis:
hmmm... yea, we can rename a standard dimension, but not a selector dimension..

<img width="747" alt="Image" src="https://github.com/user-attachments/assets/35e95128-7275-4079-b445-8f1953333e39" />

<img width="747" alt="Image" src="https://github.com/user-attachments/assets/98f933b2-8461-400c-9d9f-84b78f511278" />

droumis:
If HoloViews could auto rename and e.g. the `min` selector is applied to the `s` col, would something like `s (s.min)`, `cat (s.min)`, `var (s.min)` be ok for tooltip entries?

jlstevens:
> If HoloViews could auto rename and e.g. the min selector is applied to the s col, would something like s (s.min), cat (s.min), var (s.min) be ok for tooltip entries?

I think that it would be an improvement for now, but imho it remains a little confusing and redundant. What I really would like to see is a tooltip with sections and section headings e.g. 

```
Selected by minimum s value:
   x:  ...
   y: ...
   cat: ...
```

Of course, you would also like to see the aggregator mentioned on the appropriate value e.g. `height (mean): ...`

PR #6568, opened 2025-04-24
enh: Improve information in hover tool for selector

hoxbro:
fixes https://github.com/holoviz/holoviews/issues/6565

Currently, this does two things:

1) Add a horizontal ruler that separates the selector hover information from the rest.
2) Add a way to rename labels for a hover tool.

image

I don't know how to provide more information without making it distracting after you have read it. 

At the same time, having the information on what is used for the selector if a long column name is long can be somewhat intrusive.
image

I also don't want something showing that people find annoying because it is currently not easy to remove. 

Maybe having a label/item showing it could be a way to give the information without to much repeating:

image

### DIFF
--- a/examples/user_guide/Interactive_Hover_for_Big_Data.ipynb
+++ b/examples/user_guide/Interactive_Hover_for_Big_Data.ipynb
@@ -171,7 +171,7 @@
     "\n",
     "This design avoids sending all the raw data to the client and only transmits the necessary information on-demand. You can enable this by adding a `selector` to a `rasterize` or `datashade` operation. \n",
     "\n",
-    "By adding an `selector` we now get the information of all the dimensions in the DataFrame along with the original Hover Information, for this example the new information is `s`, `val`, and `cat`. "
+    "By adding an `selector` we now get the information of all the dimensions in the DataFrame along with the original Hover Information, for this example the new information is `s`, `val`, and `cat`. The new information is split with a horizontal rule."
    ]
   },
   {
--- a/examples/user_guide/Interactive_Hover_for_Big_Data.ipynb
+++ b/examples/user_guide/Interactive_Hover_for_Big_Data.ipynb
@@ -190,7 +190,7 @@
    "id": "90f093c4-3a3a-4c2f-9d5b-bbb6a1963d43",
    "metadata": {},
    "source": [
-    "You can specify which columns to show in the HoverTool with `hover_tooltips`. You can also rename the label by passing in a tuple with `(label, column)`."
+    "You can specify which columns to show in the HoverTool with `hover_tooltips`. You can also rename the label by passing in a tuple with `(label, column)`. The information about the selector itself can be disabled by setting `selector_in_hovertool=False`."
    ]
   },
   {
@@ -200,7 +200,8 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "rasterized_with_selector.opts(hover_tooltips=[\"x\", \"y\", (\"s [mean(s)]\", \"x_y s\"), (\"s [min(s)]\", \"s\"), (\"cat [min(s)]\", \"cat\")])"
+    "hover_tooltips=[\"x\", \"y\", (\"s [mean(s)]\", \"x_y s\"), (\"s [min(s)]\", \"s\"), (\"cat [min(s)]\", \"cat\")]\n",
+    "rasterized_with_selector.opts(hover_tooltips=hover_tooltips, selector_in_hovertool=False)"
    ]
   },
   {
--- a/examples/user_guide/Interactive_Hover_for_Big_Data.ipynb
+++ b/examples/user_guide/Interactive_Hover_for_Big_Data.ipynb
@@ -210,7 +210,7 @@
    "metadata": {},
    "source": [
     ":::{note}\n",
-    "The syntax for `hover_tooltips` when using `selector` is only contain a subset of functionality compared to when used on a non-`selector` plot.\n",
+    "When `selector` is enabled, hover tooltips use a predefined grid layout with fixed styling, limiting the customization options that would otherwise be available through `hover_tooltips` in standard plots that don't utilize `selector`.\n",
     ":::"
    ]
   },
--- a/examples/user_guide/Interactive_Hover_for_Big_Data.ipynb
+++ b/examples/user_guide/Interactive_Hover_for_Big_Data.ipynb
@@ -190,7 +190,7 @@
    "id": "90f093c4-3a3a-4c2f-9d5b-bbb6a1963d43",
    "metadata": {},
    "source": [
-    "You can specify which columns to show in the HoverTool with `hover_tooltips`. "
+    "You can specify which columns to show in the HoverTool with `hover_tooltips`. You can also rename the label by passing in a tuple with `(label, column)`."
    ]
   },
   {
@@ -200,7 +200,17 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "rasterized_with_selector.opts(hover_tooltips=[\"x\", \"y\", \"s\", \"cat\"])"
+    "rasterized_with_selector.opts(hover_tooltips=[\"x\", \"y\", (\"s [mean(s)]\", \"x_y s\"), (\"s [min(s)]\", \"s\"), (\"cat [min(s)]\", \"cat\")])"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "214f31a0-16fe-4fc7-acc1-34e0bbe4f5ef",
+   "metadata": {},
+   "source": [
+    ":::{note}\n",
+    "The syntax for `hover_tooltips` when using `selector` is only contain a subset of functionality compared to when used on a non-`selector` plot.\n",
+    ":::"
    ]
   },
   {
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -75,7 +75,7 @@ DATASHADER_VERSION = ds_version.release
 DATASHADER_GE_0_14_0 = DATASHADER_VERSION >= (0, 14, 0)
 DATASHADER_GE_0_15_1 = DATASHADER_VERSION >= (0, 15, 1)
 DATASHADER_GE_0_16_0 = DATASHADER_VERSION >= (0, 16, 0)
-DATASHADER_GE_0_17_1 = DATASHADER_VERSION >= (0, 17, 1)
+DATASHADER_GE_0_18_1 = DATASHADER_VERSION >= (0, 18, 1)
 
 
 class AggregationOperation(ResampleOperation2D):
@@ -420,7 +420,7 @@ class aggregate(LineAggregationOperation):
             agg = self._apply_datashader(dfdata, cvs_fn, sum_agg, agg_kwargs, x, y, agg_state)
             agg.attrs["selector"] = (
                 str(sel_fn)
-                if DATASHADER_GE_0_17_1
+                if DATASHADER_GE_0_18_1
                 else f"{type(sel_fn).__name__}({getattr(sel_fn, 'column', '...')!r})"
             )
         else:
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -418,7 +418,11 @@ class aggregate(LineAggregationOperation):
                 params["vdims"] = [params["vdims"]]
             sum_agg = ds.summary(**{str(params["vdims"][0]): agg_fn, "__index__": ds.where(sel_fn)})
             agg = self._apply_datashader(dfdata, cvs_fn, sum_agg, agg_kwargs, x, y, agg_state)
-            agg.attrs["selector"] = str(sel_fn) if DATASHADER_GE_0_17_1 else None
+            agg.attrs["selector"] = (
+                str(sel_fn)
+                if DATASHADER_GE_0_17_1
+                else f"{type(sel_fn).__name__}({getattr(sel_fn, 'column', '...')!r})"
+            )
         else:
             agg = self._apply_datashader(dfdata, cvs_fn, agg_fn, agg_kwargs, x, y, agg_state)
 
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -75,6 +75,7 @@ DATASHADER_VERSION = ds_version.release
 DATASHADER_GE_0_14_0 = DATASHADER_VERSION >= (0, 14, 0)
 DATASHADER_GE_0_15_1 = DATASHADER_VERSION >= (0, 15, 1)
 DATASHADER_GE_0_16_0 = DATASHADER_VERSION >= (0, 16, 0)
+DATASHADER_GE_0_17_1 = DATASHADER_VERSION >= (0, 17, 1)
 
 
 class AggregationOperation(ResampleOperation2D):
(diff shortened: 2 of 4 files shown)


maximlt:
I think my favorite UX is @jlstevens' suggestion in https://github.com/holoviz/holoviews/issues/6565#issuecomment-2827643039, somewhat similar to the last screenshot but with a "proper" heading.

I'd also be fine if HoloViews provided the options to customize things, but wouldn't do it by default, so that hvPlot could potentially leverage that to set its own defaults.

droumis:
I'm against trying to render a selector expression into a sentence, as well as any indentation formatting. 

I think Simon's last suggestion, with a horizontal bar and then adding the `Selector: <op>` is almost good enough, and maybe we can just make this line Bold to indicate the hierarchy of entries below the horizontal line.

maximlt:
> I'm against trying to render a selector expression into a sentence, as well as any indentation formatting.

Just to understand, why?

I'm not sure `Selector` is user-friendly enough to be exposed, compared to Jean-Luc's `Selected by`. At which point, we may as well remove the column.

droumis:
hmmm 

I agree that "Selector" isn't user-friendly but "Selected by minimum s value" feels too verbose and still doesn't clearly communicate what's happening in the visualization at the level that the user needs (they haven't selected anything). And it's not just for the selector, `x_y s` doesn't really convey aggregation to the user, who may not even know they are looking at a datashaded plot.

So, instead of trying to translate the technical implementation into natural language, maybe the real solution is to try to do a (IMO) simpler version of [JLS' proposal](https://github.com/bokeh/bokeh/issues/13354), and convey to the user how they are looking at an aggregation that is pixel-based, and if a selector is specified, that there are representative points that are sampled (maybe by default, the user doesn't even need to know the specific selector used for the sampling).

So e.g. a hover tooltip for a simple datashaded plot would be:

```markdown
x: 3
y: 5
---
**Pixel aggregation**
total count: 7
mean(s): 3.4
---
**Sample:**
s: 0.03
cat: d1
val: 0
```

For the more complex case of categorical data with multiple categories per pixel, I guess we can use indentation or ticks to break out the pixel aggregation and maybe (by default) it's enough to just display a single sample:

```markdown
x: 3
y: 5
---
**Pixel aggregation**
total count: 9
category [count]:  mean(s)
- catA [4]: 2.5
- catB [3]: 1.73
- catC [2]: 7.3
**Sample catA**
s: 0.03
cat: d1
val: 0
```

And make things configurable enough to allow a developer to include a sample for each category if they really want:

```markdown
x: 3
y: 5
---
**Pixel aggregation**
total count: 9
category [count]:  mean(s)
- catA [4]: 2.5
- catB [3]: 1.73
- catC [2]: 7.3
**Sample A**
s: 0.03
cat: d1
val: 0
**Sample B**
s: 0.03
cat: d1
val: 0
**Sample C**
s: 0.03
cat: d1
val: 0
```

jlstevens:
I do agree `"Selected by minimum s value"` is too verbose (but also maximally clear).

Maybe `Selected by: min('s')` in Simon's second screenshot (instead of `Selector: min('s')`) wouldn't be a bad compromise? Really it is the header for the section so ideally it would stand out somehow (e.g. shown in a different color).

hoxbro:
> I do agree `"Selected by minimum s value"` is too verbose (but also maximally clear).

Maximally clear is maybe too strong a phrase. You still need background knowledge about what selected means and the reference coordinate system (x,y in this case). Minimum is also easy to understand, but what about `first`, would a user who sees the tooltip understand that? Or more advanced cases that use `ds.by`. 

I don't have a solution and don't think there is a "perfect" way to present the information. More verbosity is great the first time you see it, but it becomes annoying when you _always_ have to see it. 

> Unfortunately, I don't believe that bokeh currently makes that easy.

We are using a custom hover tooltip for this, which is great for us as HoloViews developers. However, it is very rigid, making it hard to expose this to app developers and users.

But here is an example of using the complementary color (`#e15d26`) of the standard Bokeh label color.

image

maximlt:
Good discussion! 

> And it's not just for the selector, `x_y s` doesn't really convey aggregation to the user, who may not even know they are looking at a datashaded plot.

Yep that's also an issue.

> maybe by default, the user doesn't even need to know the specific selector used for the sampling

Hmm I'd prefer if we were more explicit, it's not quite the same to see the `min` or the `max` selected values.

> I don't have a solution and don't think there is a "perfect" way to present the information. More verbosity is great the first time you see it, but it becomes annoying when you _always_ have to see it.

Yep true. We start from a place where it was pretty much impossible for someone looking at the tooltip to understand what it displayed, you had to see the code *and* know what it does. Since the whole explanation is a bit complex, we can't stuff it in the tooltip. However, we can add hints that make it more likely for someone to understand what they see. They also serve as clearer pointers for someone explaining what the tooltip displays (e.g., in our own docs, we could write "The heading below the divider indicates which dimension and function was used as a selector. The values of the selected row are displayed below the heading.").

How hard is it to add a custom (❓) tool to Bokeh's toolbar that, when clicked, opens a pop-up with some text content? If that's straightforward, we could add that to the docs to see how to augment the plot with some explanatory doc.

I'm fine with the latest proposal but would like to understand why the heading is structured as if it were another field (i.e. `{label}: {value}`)? Is it because it's easier to implement? Just wondering if following the same structure as a normal field wouldn't cause some confusion.

hoxbro:
> How hard is it to add a custom (❓) tool to Bokeh's toolbar that, when clicked, opens a pop-up with some text content? If that's straightforward, we could add that to the docs to see how to augment the plot with some explanatory doc.

The best I can do is this. Updating the text based on the selector is not straightforward. (sorry for the bad crop)

https://github.com/user-attachments/assets/f818fd43-4a75-481c-9f7d-80fd83faae8f

<details><summary>Ad-hoc code to make tooltip icon</summary>
<p>

``` diff
diff --git a/holoviews/plotting/bokeh/raster.py b/holoviews/plotting/bokeh/raster.py
index 2aa008628..cc7b8065c 100644
--- a/holoviews/plotting/bokeh/raster.py
+++ b/holoviews/plotting/bokeh/raster.py
@@ -1,11 +1,13 @@
+import base64
 import sys
 import uuid
+from textwrap import dedent
 
 import bokeh.core.properties as bp
 import numpy as np
 import param
 from bokeh.model import DataModel
-from bokeh.models import CustomJS, CustomJSHover, DatetimeAxis, HoverTool
+from bokeh.models import CustomAction, CustomJS, CustomJSHover, DatetimeAxis, HoverTool
 from bokeh.models.dom import Div, Span, Styles, ValueOf
 from panel.io import hold
 
@@ -17,12 +19,57 @@ from .chart import PointPlot
 from .element import ColorbarPlot, LegendPlot
 from .selection import BokehOverlaySelectionDisplay
 from .styles import base_properties, fill_properties, line_properties, mpl_to_bokeh
-from .util import BOKEH_GE_3_3_0, BOKEH_GE_3_4_0, colormesh
+from .util import BOKEH_GE_3_3_0, BOKEH_GE_3_4_0, BOKEH_GE_3_7_0, colormesh
 
 
 class ServerHoverMixin:
     _model_cache = {}
 
+    def _init_glyph(self, plot, mapping, properties):
+        """Returns a Bokeh glyph object and optionally creates a colorbar.
+
+        """
+        ret = super()._init_glyph(plot, mapping, properties)
+        tool = self.handles.get('hover')
+        if tool and 'hv_created' in tool.tags and isinstance(tool.tooltips, Div):
+            svg_icon = """\
+                <svg fill="#a1a6a9" height="52px" width="52px" version="1.1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 297 297">
+                  <path d="M163.207 181.637l42.749-24.76c3.358-1.945
+                      5.309-5.635 5.029-9.504-.28-3.867-2.743-7.238-6.344-8.68L13.9
+                      62.416c-3.759-1.504-8.057-.625-10.924 2.238-2.867 2.861-3.755
+                      7.156-2.258 10.918l76.065 191.278c1.436 3.613 4.81 6.084
+                      8.688 6.365 3.896.272 7.571-1.685 9.512-5.051l24.658-42.816
+                      59.882 60.053c1.903 1.906 4.483 2.978 7.179 2.978 2.692 0
+                      5.275-1.072 7.176-2.98l29.257-29.342c3.945-3.957 3.945-10.359
+                      0-14.313l-60.948-60.968zM186.702
+                      263.889l-62.057-62.232c-1.916-1.918-4.504-2.979-7.177-2.979-.44
+                      0-.881.029-1.323.086-3.127.414-5.888 2.258-7.461 4.992l-20.81
+                      36.131L28.269 89.994l149.497 59.783-36.117 20.92c-2.722
+                      1.576-4.556 4.326-4.969 7.445-.409 3.119.651 6.254 2.872
+                      8.479l62.093 62.277 10.057-10.01z"/>
+                  <path d="M234.977 8.621c-34.2 0-62.025 27.83-62.025 62.035 0
+                      5.596 4.54 10.135 10.136 10.135 5.599 0 10.137-4.539
+                      10.137-10.135 0-23.031 18.731-41.764 41.752-41.764 23.022 0
+                      41.752 18.732 41.752 41.764 0 23.025-18.73 41.762-41.752
+                      41.762-5.597 0-10.136 4.535-10.136 10.135v51.895c0 5.598
+                      4.539 10.135 10.136 10.135 5.598 0 10.136-4.537
+                      10.136-10.135v-42.586C274.508 127.008 297 101.408 297
+                      70.656c0-34.205-27.824-62.035-62.023-62.035z"/>
+                  <path d="M234.974 198.904c-2.665 0-5.28 1.084-7.165
+                      2.971-1.885 1.895-2.971 4.5-2.971 7.166 0 2.676 1.086 5.291
+                      2.971 7.176 1.885 1.885 4.5 2.971 7.165 2.971 2.667 0
+                      5.281-1.086 7.167-2.971 1.895-1.885 2.969-4.5 2.969-7.176
+                      0-2.666-1.074-5.271-2.969-7.166-1.886-1.887-4.5-2.971-7.167-2.971z"/>
+                </svg>"""
+            encoded_icon = base64.b64encode(dedent(svg_icon).encode()).decode('ascii')
+            info_tool = CustomAction(
+                icon=f"data:image/svg+xml;base64,{encoded_icon}",
+                description="This is an example of description\n New line",
+            )
+            self.handles["plot"].toolbar.tools.append(info_tool)
+
+        return ret
+
     def _update_hover(self, element):
         tool = self.handles['hover']
         if 'hv_created' in tool.tags and isinstance(tool.tooltips, Div):

```
</p>
</details> 

> I'm fine with the latest proposal but would like to understand why the heading is structured as if it were another field (i.e. {label}: {value})? Is it because it's easier to implement? Just wondering if following the same structure as a normal field wouldn't cause some confusion.

I think diverging too much from the structure will confuse, it will likely grab the attention because it is different. 

> What's the wildest example you have in mind? :D

Nothing crazy in mind, but as soon as something is more than one level deep, it becomes hard to really grasp, in my opinion. FWIW, the most advanced `aggregator` I could make was `ds.by("cat", ds.where(ds.min("s"), "val"))`.

hoxbro:
> But, if it could be adapted to explain how users can create their own custom tool separately

I'm using raw Bokeh so it will likely need to be done with a hook, making it very unelegant.

> I'm not convinced by the red color

I just chose the complementary color of light blue. 


> Can the whole line be formatted homogeneously (e.g. bold or underline)?

Yes, we can do it as long as it can be done in HTML. I would suggest having an opts, which makes it possible to hide it if the user decides it,

> We could just crop too long expressions.

If there is an elegant way to do it, sure, but it kind of defeats the purpose of having it there in the first place.

maximlt:
> I'm using raw Bokeh so it will likely need to be done with a hook, making it very unelegant.

Yeah ok, scratch that then. Still useful to think about this kind of stuff. If I were to use a selector like `ds.by("cat", ds.where(ds.min("s"), "val"))` one day, I would have to explain somewhere in a notebook or on a website what the tooltip displays, as it's very likely I wouldn't remember what it does.

> I would suggest having an opts, which makes it possible to hide it if the user decides it,

Yep up to you, I'm in favor of having the heading displayed by default.

> If there is an elegant way to do it, sure, but it kind of defeats the purpose of having it there in the first place.

Not really, it just makes it less useful for complex reduction reprs (which we could also decide to simplify). It also depends on the number of characters selected to crop, if that's the approach adopted.

hoxbro:
Pushed some changes so it now looks like this:

image

holoviews/operation/datashader.py
```diff
@@ -417,6 +418,7 @@ def _process(self, element, key=None):
                 params["vdims"] = [params["vdims"]]
             sum_agg = ds.summary(**{str(params["vdims"][0]): agg_fn, "__index__": ds.where(sel_fn)})
             agg = self._apply_datashader(dfdata, cvs_fn, sum_agg, agg_kwargs, x, y, agg_state)
+            agg.attrs["selector"] = str(sel_fn) if DATASHADER_GE_0_17_1 else None
```
hoxbro, line 421: I _think_ it is ok to infer it if we don't have the latest version.

holoviews/plotting/bokeh/raster.py
```diff
@@ -20,9 +20,12 @@
 from .util import BOKEH_GE_3_3_0, BOKEH_GE_3_4_0, colormesh
 
 
-class ServerHoverMixin:
+class ServerHoverMixin(param.Parameterized):
     _model_cache = {}
 
+    selector_in_hovertool = param.Boolean(default=True, doc="""
```
hoxbro, line 26: I'm not a big fan of this name, but I'm open to suggestions.

hoxbro:
I need to add some UI tests and documentation to hide the selector in the hover tool 

I'll do that after we have agreed that the latest look is ok.

maximlt:
The last screenshot looks good to me 👍

holoviews/operation/datashader.py
hoxbro: Done in https://github.com/holoviz/holoviews/pull/6568/commits/ec4dda9a221ee20e99aa31260b1f63fa7f66f021

droumis:
Yea I can live with the last screenshot  👍

examples/user_guide/Interactive_Hover_for_Big_Data.ipynb
droumis, line 1: > The syntax for `hover_tooltips` when using `selector` is only contain a subset of functionality compared to when used on a non-`selector` plot.

Should be:
"When `selector` is enabled, hover tooltips use a predefined grid layout with fixed styling, limiting the customization options that would otherwise be available through `hover_tooltips` in standard plots that don't utilize `selector`."

holoviews/plotting/bokeh/raster.py
droumis: I think "show_selector" is a bit better

droumis (changes requested):

droumis:
datashader op rendering issue

image

<details> <summary> Versions </summary>

```
Python              :  3.11.12 | packaged by conda-forge | (main, Apr 10 2025, 22:18:52) [Clang 18.1.8 ]
Operating system    :  macOS-15.4.1-arm64-arm-64bit

holoviews           :  1.21.0a0.post1.dev15+g9c2062fef

bokeh               :  3.7.2
colorcet            :  3.1.0
cudf                :  -
cupy                :  -
dask                :  -
dask-expr           :  -
datashader          :  0.18.0
geoviews            :  -
hvplot              :  -
ibis-framework      :  -
IPython             :  9.1.0
ipywidgets-bokeh    :  -
jupyter-bokeh       :  -
jupyterlab          :  4.4.1
matplotlib          :  3.10.1
networkx            :  3.4.2
notebook            :  -
numba               :  0.61.2
numpy               :  2.2.5
pandas              :  2.2.3
panel               :  1.6.3
param               :  2.2.0
pillow              :  11.2.1
plotly              :  -
pyarrow             :  -
pyviz-comms         :  3.0.4
scikit-image        :  -
scipy               :  1.15.2
spatialpandas       :  -
streamz             :  -
tsdownsample        :  -
xarray              :  2025.3.1
```
</details>

droumis:
Use of default dynspread is not compelling enough to include here...


image

hoxbro:
> datashader 0.18.0 op rendering issue

Fixed in https://github.com/holoviz/holoviews/pull/6568/commits/d88995d83a845034a6dce75c6cfa8de689c1c062, used wrong version :upside_down_face: 

> Use of default dynspread is not compelling enough to include here...

It was mostly to show it also works with dynspread.

image

droumis (approved):

PR #6570, opened 2025-04-29
perf: Don't import Panel and Pandas directly in IPython

hoxbro:
Continuing from https://github.com/holoviz/holoviews/pull/6476

I noticed that HoloViews was _very_ slow when importing in IPython. This is mainly because of the import of Panel and Pandas. This PR defers these imports. 

# Hyperfine

## Before (based on #6476)
 `hyperfine "ipython -c ''" "ipython -c 'import holoviews'" --warmup 5`

``` bash
❯ hyperfine "ipython -c ''" "ipython -c 'import holoviews'" --warmup 5
Benchmark 1: ipython -c ''
  Time (mean ± σ):     256.3 ms ±   2.9 ms    [User: 191.0 ms, System: 36.7 ms]
  Range (min … max):   252.2 ms … 260.9 ms    11 runs

Benchmark 2: ipython -c 'import holoviews'
  Time (mean ± σ):      1.398 s ±  0.017 s    [User: 2.820 s, System: 0.150 s]
  Range (min … max):    1.369 s …  1.416 s    10 runs

Summary
  ipython -c '' ran
    5.45 ± 0.09 times faster than ipython -c 'import holoviews'
```

## After (this PR)
``` bash
❯ hyperfine "ipython -c ''" "ipython -c 'import holoviews'" --warmup 5
Benchmark 1: ipython -c ''
  Time (mean ± σ):     256.7 ms ±   2.8 ms    [User: 194.2 ms, System: 34.5 ms]
  Range (min … max):   253.8 ms … 262.2 ms    11 runs

Benchmark 2: ipython -c 'import holoviews'
  Time (mean ± σ):     484.8 ms ±   5.7 ms    [User: 2020.4 ms, System: 57.8 ms]
  Range (min … max):   477.9 ms … 496.4 ms    10 runs

Summary
  ipython -c '' ran
    1.89 ± 0.03 times faster than ipython -c 'import holoviews'
```

# Tuna
`python -X importtime -m IPython -c 'import holoviews' 2> tuna.log && tuna tuna.log` (and not having setuptools_scm installed)

## Before (based on #6476)
image

## After (this PR)
image

### DIFF
--- a/holoviews/element/comparison.py
+++ b/holoviews/element/comparison.py
@@ -780,6 +780,56 @@ class ComparisonTestCase(Comparison, TestCase):
             self.addTypeEqualityFunc(k, v)
 
 
+class IPTestCase(ComparisonTestCase):
+    """This class extends ComparisonTestCase to handle IPython specific
+    objects and support the execution of cells and magic.
+
+    """
+
+    def setUp(self):
+        super().setUp()
+        try:
+            import IPython
+            from IPython.display import HTML, SVG
+            self.ip = IPython.InteractiveShell()
+            if self.ip is None:
+                raise TypeError()
+        except Exception as e:
+            raise SkipTest("IPython could not be started") from e
+
+        self.ip.displayhook.flush = lambda: None  # To avoid gc.collect called in it
+        self.addTypeEqualityFunc(HTML, self.skip_comparison)
+        self.addTypeEqualityFunc(SVG,  self.skip_comparison)
+
+    def skip_comparison(self, obj1, obj2, msg): pass
+
+    def get_object(self, name):
+        obj = self.ip._object_find(name).obj
+        if obj is None:
+            raise self.failureException(f"Could not find object {name}")
+        return obj
+
+
+    def cell(self, line):
+        """Run an IPython cell
+
+        """
+        self.ip.run_cell(line, silent=True)
+
+    def cell_magic(self, *args, **kwargs):
+        """Run an IPython cell magic
+
+        """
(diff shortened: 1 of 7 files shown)


holoviews/element/comparison.py
```diff
@@ -780,6 +780,56 @@ def __init__(self, *args, **kwargs):
             self.addTypeEqualityFunc(k, v)
 
 
+class IPTestCase(ComparisonTestCase):
```
hoxbro, line 783: Moving this here is a breaking change. I still think it is OK.

hoxbro: I have added a deprecation warning with removal in 1.22.0.

holoviews/ipython/__init__.py
```diff
@@ -28,54 +25,13 @@ def show_traceback():
...
-    def line_magic(self, *args, **kwargs):
-        """Run an IPython line magic
-
-        """
-        self.ip.run_line_magic(*args, **kwargs)
+def __getattr__(attr):
+    if attr == "IPTestCase":
+        from ..element.comparison import IPTestCase
+        from ..util.warnings import deprecated
+        deprecated("1.22.0", old="holoviews.ipython.IPTestCase", new="holoviews.element.comparison.IPTestCase")
```
hoxbro, line 32:
```suggestion
        deprecated("1.23.0", old="holoviews.ipython.IPTestCase", new="holoviews.element.comparison.IPTestCase")
```

maximlt (approved):
Suggested one small improvement.

PR #6571, opened 2025-04-30
fix: Don't assume all extents are string in `max_extents`

hoxbro:
Fixes the following hvplot problem.

``` python
import hvplot.pandas
import pandas as pd

df = pd.DataFrame({
    "col1": ["apple", "banana", "cherry", "date", "elderberry"],
    'date': pd.to_datetime(['2023-01-01', '2023-02-14', '2023-03-21', '2023-04-30', '2023-05-15'])
})
df.hvplot.scatter(x="index")
```

Or translated to HoloViews:

``` python
import holoviews as hv
import pandas as pd

hv.extension("bokeh") 

df = pd.DataFrame({
    "col1": ["apple", "banana", "cherry", "date", "elderberry"],
    'date': pd.to_datetime(['2023-01-01', '2023-02-14', '2023-03-21', '2023-04-30', '2023-05-15'])
})
hv.NdOverlay({col: hv.Scatter(df, kdims="index", vdims=col).opts(tools=["hover"]) for col in data})
```

### DIFF
--- a/holoviews/core/util.py
+++ b/holoviews/core/util.py
@@ -1121,13 +1121,13 @@ def max_extents(extents, zrange=False):
             if lower and isinstance(lower[0], datetime_types):
                 extents[lidx] = np.min(lower)
             elif any(isinstance(l, str) for l in lower):
-                extents[lidx] = np.sort(lower)[0]
+                extents[lidx] = sorted(lower, key=str)[0]
             elif lower:
                 extents[lidx] = np.nanmin(lower)
             if upper and isinstance(upper[0], datetime_types):
                 extents[uidx] = np.max(upper)
             elif any(isinstance(u, str) for u in upper):
-                extents[uidx] = np.sort(upper)[-1]
+                extents[uidx] = sorted(upper, key=str)[-1]
             elif upper:
                 extents[uidx] = np.nanmax(upper)
     return tuple(extents)


maximlt (commented):
I don't think the plot attempted to be created (in hvPlot and HoloViews) should be valid. Already, making a scatter plot with y as a categorical axis is a bit strange:
image

With these changes, it's possible to overlay this plot with a plot whose y-axis consists of datetime values. The result (which would throw an error before) isn't very appealing:
image

---

The original error was reported by a user on Discord who attempted to create an instance of the hvPlot Explorer with a similar dataframe. I feel the Explorer should pick more wisely the default plot it attempts to create. And, not crash entirely on "start-up" when it fails at displaying the default plot.

```python
data = {
    'col1': ['apple', 'banana', 'cherry', 'date', 'elderberry'],
    'col2': ['red', 'yellow', 'red', 'brown', 'purple'],
    'date': pd.to_datetime(['2023-01-01', '2023-02-14', '2023-03-21', '2023-04-30', '2023-05-15'])
}
df = pd.DataFrame(data)
```

With this branch checked out, the Explorer no longer fails, but the plot doesn't really make sense to be honest:

image

maximlt:
> I disagree. Even though it is indeed strange, I don't think we should decide if it is valid. This should be up to the individual user.

I get this point. Though be aware that, with this change, code that used to break in an unhandled way will now generate a plot. I'd call this a "new unblocked feature" (well, unless it worked before?). But if this is really new and not a regression, it opens the doors to users coming up with new bug reports/feature requests on this "new unblocked feature". You have to be sure that is something you want to support.

holoviews/core/util.py
```diff
@@ -1118,18 +1118,20 @@ def max_extents(extents, zrange=False):
...
-                extents[lidx] = np.nanmin(lower)
-            if upper and isinstance(upper[0], datetime_types):
-                extents[uidx] = np.max(upper)
-            elif any(isinstance(u, str) for u in upper):
-                extents[uidx] = np.sort(upper)[-1]
-            elif upper:
-                extents[uidx] = np.nanmax(upper)
+            if lower:
+                if any(isinstance(l, str) for l in lower):
+                    extents[lidx] = sorted(lower, key=str)[0]
```
maximlt, line 1123: Somehow, this got lost in the review I made. I had the feeling `max_extents()` was designed to deal with homogeneous types. In this case, when it runs the new test, it sorts a datetime and a string. Is this a valid thing to do?

hoxbro: It is very speculative. But if the original code had been something like `if isinstance(lower[0], str)`, like the datetime, I would likely have agreed with you, but the current code looks for any instances of str.

hoxbro:
`np.sort` has become stricter over time, see:

``` bash
❯ python -c "import numpy as np; print(np.__version__); print(np.sort(['a', np.datetime64('2020')]))"
1.12.1
['2020' 'a']
```

``` bash
❯ python -c "import numpy as np; print(np.__version__); print(np.sort(['a', np.datetime64('2020')]))"
2.2.5
  File "<string>", line 1, in <module>
    import numpy as np; print(np.__version__); print(np.sort(['a', np.datetime64('2020')]))
                                                     ~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/shh/.local/conda/envs/holoviz/lib/python3.13/site-packages/numpy/_core/fromnumeric.py", line 1122, in sort
    a.sort(axis=axis, kind=kind, order=order, stable=stable)
    ~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
numpy._core._exceptions._UFuncNoLoopError: ufunc 'less' did not contain a loop with signature matching types (<class 'numpy.dtypes.DateTime64DType'>, <class 'numpy.dtypes.StrDType'>) -> None
```

maximlt (approved):
LGTM as it looks like it is a regression.

PR #6574, opened 2025-05-13
fix: Don't unpack clas in annotation

hoxbro:
Follow up to https://github.com/holoviz/holoviews/pull/6476

### DIFF
--- a/holoviews/core/util/types.py
+++ b/holoviews/core/util/types.py
@@ -74,8 +74,8 @@ def datetime_types():
     yield from (dt.datetime, dt.date, dt.time)
     if np:
         yield np.datetime64
-    yield from pandas_datetime_types()
-    yield from cftime_types()
+    yield from pandas_datetime_types
+    yield from cftime_types
 
 
 @gen_types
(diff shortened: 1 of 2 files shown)


holoviews/element/annotation.py
```diff
@@ -170,7 +170,7 @@ class HLine(Annotation):
 
     group = param.String(default='HLine', constant=True)
 
-    y = param.ClassSelector(default=0, class_=(Number, *datetime_types), doc="""
+    y = param.ClassSelector(default=0, class_=(Number, datetime_types), doc="""
```
hoxbro, line 173: This meant that it would static, defeating the purpose.

holoviews/core/util/types.py
```diff
@@ -74,16 +74,16 @@ def datetime_types():
     yield from (dt.datetime, dt.date, dt.time)
     if np:
         yield np.datetime64
-    yield from pandas_datetime_types()
-    yield from cftime_types()
+    yield from pandas_datetime_types
```
hoxbro, line 77: No need to call it.

Issue #6515, opened 2025-02-20
Server side HoverTool follow ups


hoxbro:
HoloViews:

1. See https://github.com/holoviz/holoviews/pull/6422#discussion_r1957970505
https://github.com/holoviz/holoviews/blob/48f91a59cb961092e9c60ef068a54600e213645b/holoviews/operation/datashader.py#L1763-L1766

2. Add documentation explaining the difference between selector and aggregator, e.g., that selector is a subset of aggregator, basically that you select a row (`min`, `max`, `first`, `last`, etc.) and not "combine" multiple rows (`mean`, `std`, etc.)

Bokeh (as of Bokeh 3.6, not yet supported in Bokeh)

3. https://github.com/bokeh/bokeh/issues/14123
    Currently, we do a simple conversion
4. https://github.com/bokeh/bokeh/issues/9087
    Currently, we show the Hovertool even when no data is available, resulting in a HoverTool filled with nans.

droumis:
@hoxbro , the first point doesn't seem like a blocker, correct?

PR #6422, opened 2024-10-25
feat(bokeh): Server side HoverTool for rasterized/datashaded plots with selector

hoxbro:
Previously, we sent all hover data to the front end, which could lead to significant overhead on larger plots. With this PR, the data is pushed server-side to the hover tool with a custom data model.  

<details>
<summary>script</summary>

``` python
import itertools

import datashader as ds
import numpy as np
import pandas as pd
import panel as pn

import holoviews as hv
from holoviews.operation.datashader import datashade, dynspread, rasterize

hv.extension("bokeh")

num = 10000
seed = np.random.default_rng(1)


dists = {
    cat: pd.DataFrame(
        {
            "x": seed.normal(x, s, num),
            "y": seed.normal(y, s, num),
            "s": s,
            "val": val,
            "cat": cat,
        }
    )
    for x, y, s, val, cat in [
        (2, 2, 0.03, 0, "d1"),
        (2, -2, 0.10, 1, "d2"),
        (-2, -2, 0.50, 2, "d3"),
        (-2, 2, 1.00, 3, "d4"),
        (0, 0, 3.00, 4, "d5"),
    ]
}

df = pd.concat(dists, ignore_index=True)
df["y"] += 100  # Offset to distinguish x from y
points = hv.Points(df)


def dynspread_datashade(*args, **kwargs):
    return dynspread(datashade(*args, **kwargs))


def dynspread_rasterize(*args, **kwargs):
    return dynspread(rasterize(*args, **kwargs))


ops = (rasterize, dynspread_rasterize, datashade, dynspread_datashade)
aggs = (ds.min("s"), ds.by("cat"))
sels = (None, ds.min("s"))
# ops = (datashade,)
# aggs = (ds.by("cat"),)
# sels = (None,)
combinations = itertools.product(ops, aggs, sels)
plots = []
for op, agg, sel in combinations:
    title = f"{op.__name__.replace('_','+')}(agg={type(agg).__name__}, sel={type(sel).__name__ if sel else None})"
    plot = op(points, aggregator=agg, selector=sel).opts(
        tools=["hover"], title=title, width=400, height=400
    )
    plots.append(plot)

pn.panel(hv.Layout(plots).cols(4).opts(shared_axes=False)).servable()

```

</details>

image

Fixes #6016 
Fixes #6018 
Fixes #6020

PR #6581, opened 2025-05-21
refactor: Simplify selector data clone

hoxbro:
Fixes point one in https://github.com/holoviz/holoviews/issues/6515

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -1789,12 +1789,7 @@ class SpreadingOperation(LinkableOperation):
 
             for k, v in sel_data.items():
                 new_data[k].data = v
-
-            # TODO: Investigate why this does not work
-            # element = element.clone(data=new_data, kdims=element.vdims.copy(), vdims=element.vdims.copy())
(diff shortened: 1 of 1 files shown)


holoviews/operation/datashader.py
```diff
@@ -1789,12 +1789,7 @@ def _process(self, element, key=None):
 
             for k, v in sel_data.items():
                 new_data[k].data = v
-
-            # TODO: Investigate why this does not work
-            # element = element.clone(data=new_data, kdims=element.vdims.copy(), vdims=element.vdims.copy())
```
hoxbro, line 1794: Fresh eyes showed that `kdims=element.vdims.copy()` was the culprit; it should have been kdims.

maximlt: So there's no need to pass kdims and vdims to `clone()` (I assume so, just checking).

hoxbro: There shouldn't be. Have tested with the example used in  #6422 and saw no error.

PR #6584, opened 2025-05-22
fix: hover selector in overlay

maximlt:
This line was raising an error `tooltips = tuple(tool.tooltips) if tool.tooltips else ()` as `tool.tooltips` is a `Div`. With these changes this example works fine.

```python
import datashader as ds
import holoviews as hv
import numpy as np
import pandas as pd

hv.extension('bokeh')
num = 100
np.random.seed(1)

dists = {
    cat: pd.DataFrame(
        {
            "x": np.random.normal(x, s, num),
            "y": np.random.normal(y, s, num),
            "s": s,
            "val": val,
            "cat": cat,
        }
    )
    for x, y, s, val, cat in [
        (2, 2, 0.03, 0, "d1"),
        (2, -2, 0.10, 1, "d2"),
        (-2, -2, 0.50, 2, "d3"),
        (-2, 2, 1.00, 3, "d4"),
        (0, 0, 3.00, 4, "d5"),
    ]
}

layout = rasterize(hv.Points(df), selector=ds.first('val')).opts(tools=['hover']) * hv.Points([])
layout
```

image

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -3237,7 +3237,10 @@ class OverlayPlot(GenericOverlayPlot, LegendPlot):
                     else:
                         tool_type = type(tool)
                     if isinstance(tool, tools.HoverTool):
-                        tooltips = tuple(tool.tooltips) if tool.tooltips else ()
+                        if isinstance(tool.tooltips, bokeh.models.dom.Div):
+                            tooltips = tool.tooltips
+                        else:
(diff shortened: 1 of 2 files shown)


hoxbro:
How does the `mergetool` function work if the other plot has a hover tool too?

maximlt:
You get a sort of overlay of hover tooltips.

image

hoxbro:
I think that is OK

PR #6586, opened 2025-05-23
enh: Improve the content displayed when loading the extension in a notebook

maximlt:
Goals:
- Exploit the fact that HoloViews displays its logo and the backends logo to provide more information
  - Make the icons clickable, linking to the documentation
  - On hover, the icons show `{package} {version}`
- Let the user know they should probably not delete this cell. In the future I'd love to be able to link to a page that gives more explanation, it doesn't exist yet.
- Allow hvPlot to customize the icon and content that gets displayed

image

Hover on the HoloViews icon:
image

Hover on the Bokeh icon:
image

Hover on the question mark:
image

This enables hvPlot to also include its icon.

image


maximlt:
Draft for now but very interested in getting some feedback already to know if this is going the right way and could make it in the next release.

hoxbro:
I don't like the text; maybe add only the I icon and place it next to the logos. This is, in my opinion, less intrusive.

maximlt:
> I don't like the text; maybe add only the I icon and place it next to the logos. This is, in my opinion, less intrusive.

I've removed the text, though I would prefer to keep it as I don't find it more intrusive than the logos.

I've extended the help tooltip.

image

holoviews/ipython/__init__.py
```diff
@@ -33,7 +33,9 @@ def __getattr__(attr):
         return IPTestCase
     raise AttributeError(f"module {__name__!r} has no attribute {attr!r}")
 
-
+HOLOVIEWS_B64_LOGO = (
```
hoxbro: Can we keep this in the HTML, and have an option to overwrite it?

hoxbro (commented):
Can we align the (i) with the other icons?

image

holoviews/ipython/__init__.py
maximlt: Done in https://github.com/holoviz/holoviews/pull/6586/commits/b6693008d9e3d1f70ba54a4f1e4e207f9ca805f7

maximlt:
> Can we align the (i) with the other icons?

I don't think we should align it like you did. You've made it aligned with the plotting extension logos (there can be up to three), which is confusing imo.

hoxbro:
> I don't think we should align it like you did. You've made it aligned with the plotting extension logos (there can be up to three), which is confusing imo.

I disagree, the (i) is similar to a footnote and should not steal focus. The current implementation also looks like it could be caused by something else, whereas if it is next to the other icons, it is clear to me that it is related to HoloViews. 

Also, if people are confused, they can always hover over it.

maximlt:
> the (i) is similar to a footnote and should not steal focus.

Not in my opinion. The logos are there for aesthetic reasons only I think as the line above (`hv.extension('bokeh')` is already pretty clear about what is going on. The fact that the cell output should not be removed is purely implicit and users learn it the hard way.

> The current implementation also looks like it could be caused by something else, whereas if it is next to the other icons, it is clear to me that it is related to HoloViews.

I'm not quite sure how one could think it's not related to HoloViews as it would only appear after calling `hv.extension`.

I can agree with you that your suggestion is a bit more aesthetically pleasing in terms of how things end up being aligned. But the help icon is not about the plotting backends, which this alignment implies. It looked better when the "Extension loaded." text bit was there ;)

hoxbro:
> I'm not quite sure how one could think it's not related to HoloViews as it would only appear after calling hv.extension.

Your example only includes the import and extension, but there could easily be 5-10 other imports and settings in the first cell. 

> I can agree with you that your suggestion is a bit more aesthetically pleasing in terms of how things end up being aligned. But the help icon is not about the plotting backends, which this alignment implies.

I see it all as part of HoloViews' information, which, until now, has only shown which plotting backend has been activated. This is another information for a user.  

It is also crucial to have something aesthetically pleasing, especially since it will be shown in every notebook by default.

maximlt:
One day some day in another galaxy in another universe, the HoloViz group will have a UI/UX person who actually has a clue about these things. Until then, we'll align things.

image

philippjfr:
Thanks for doing this! I also prefer the tooltip only over the "Extension loaded" text.

Issue #6595, opened 2025-05-27
Error when applying dynspread to an imagestack with only one category


maximlt:
Quick bug report to make sure I don't forget it.

```python
import datashader as ds
import holoviews as hv
import holoviews.operation.datashader as hd
import numpy as np

hv.extension('bokeh')

N_OBS = 1000
x_data = np.random.random((N_OBS, N_OBS))

color_by = 'cell_type'

# Only one category
color_data = np.random.choice(['A'], size=N_OBS)

dataset = hv.Dataset(
    (x_data[:, 0], x_data[:, 1], color_data),
    ['X', 'Y'],
    color_by,
)
plot = dataset.to(hv.Points)
plot = hd.rasterize(
    plot,
    aggregator=ds.count_cat(color_by),
    selector=ds.first('X'),
)
plot = hd.dynspread(plot, threshold=0.5)
plot = plot.opts(tools=["hover"], title=color_by)
hv.render(plot)
```

```
  File "/Users/mliquet/dev/holoviews/holoviews/operation/datashader.py", line 1783, in _process
    new_data[k].data = array.sel(z=k)
                       ^^^^^^^^^^^^^^
  File "/Users/mliquet/dev/hv-anndata/.venv/lib/python3.12/site-packages/xarray/core/dataarray.py", line 1682, in sel
    ds = self._to_temp_dataset().sel(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mliquet/dev/hv-anndata/.venv/lib/python3.12/site-packages/xarray/core/dataset.py", line 2890, in sel
    query_results = map_index_queries(
                    ^^^^^^^^^^^^^^^^^^
  File "/Users/mliquet/dev/hv-anndata/.venv/lib/python3.12/site-packages/xarray/core/indexing.py", line 189, in map_index_queries
    grouped_indexers = group_indexers_by_index(obj, indexers, options)
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mliquet/dev/hv-anndata/.venv/lib/python3.12/site-packages/xarray/core/indexing.py", line 150, in group_indexers_by_index
    raise KeyError(
KeyError: "'z' is not a valid dimension or coordinate for Dataset with dimensions FrozenMappingWarningOnValuesAccess({'X': 400, 'Y': 400})"
```

PR #6596, opened 2025-05-27
fix: Dynspread and Selector with single categorical

hoxbro:
fixes https://github.com/holoviz/holoviews/issues/6595


maximlt (changes requested):
LGTM but it needs a test at least.

maximlt (approved):

droumis:
Noice

image

PR #6599, opened 2025-05-28
enh(bokeh): Add `radius` and `radius_dimension` to PointPlot

hoxbro:
Adds `radius` and `radius_dimension` to the Bokeh PointPlot. Because radius works with Circle, we update the `_plot_methods` before calling `_init_glyph`.


holoviews/plotting/bokeh/chart.py
```diff
@@ -63,12 +63,21 @@ class PointPlot(LegendPlot, ColorbarPlot):
...
         *base_properties, *line_properties, *fill_properties
     ]
 
     _plot_methods = dict(single='scatter', batched='scatter')
     _batched_style_opts = line_properties + fill_properties + ['size', 'marker', 'angle']
 
+    def _init_glyph(self, plot, mapping, properties):
+        if "radius" in properties:
+            self._plot_methods = dict(single='circle', batched='circle')
+            properties.pop("size", None)
```
hoxbro, line 76: This will make it so radius always has priority over size.

philippjfr:
Looks good!

PR #6600, opened 2025-05-28
fix: Wrong ordered dendrogram and add more linkage options

hoxbro:
Fixes a bug in the current implementation that reversed the order. The bokeh documentation has been updated to use the following example as it clearly (at least to me) shows how the dendrogram should behave.

Adding two more linkage options, `linkage_method` and `linkage_metric`. The default value for both is being set to mimic the scanpy implementation. 

I don't think a test is absolutely necessary, because, as far as I can see, we would only check for repeatability, not correctness, but I will, of course, add one if requested. 

``` python
import holoviews as hv
import pandas as pd

hv.extension("bokeh")

data = [
    [7, 8, 0, 0, 0, 0],
    [8, 6, 0, 0, 0, 0],
    [0, 0, 9, 8, 2, 7],
    [0, 0, 0, 1, 8, 2],
    [0, 7, 0, 7, 0, 0],
]
df = pd.DataFrame([
    {"cluster": f"clust {i}", "gene": f"gene {j}", "value": data[i][j]}
    for j in range(6) 
    for i in range(5)
])

opts = dict(
    color=hv.dim("value").norm(), size=(hv.dim("value") != 0) * 30, cmap="Reds", tools=["hover"]
)
plot = hv.Points(df, kdims=["gene", "cluster"]).opts(**opts)

hv.operation.dendrogram(plot, adjoint_dims=["cluster"], main_dim="value")

```
image

Refs:
- https://gist.github.com/droumis/985e7ccefcaf25d2b73b7b7fac6d79c1
- https://github.com/scverse/scanpy/blob/89780ec9cbc7ca8cfacc39a0f060602639ef3b25/src/scanpy/tools/_dendrogram.py#L166-L174

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -1249,6 +1249,23 @@ class dendrogram(Operation):
          https://docs.scipy.org/doc/scipy/reference/generated/scipy.cluster.hierarchy.linkage.html#scipy.cluster.hierarchy.linkage
          """)
 
+    linkage_method = param.Selector(
+        default="single",
+        objects=["single", "complete", "average", "centroid", "median", "ward", "weighted"]
+    )
+    linkage_metric=param.Selector(
+        default='euclidean',
+        objects=[
+            'braycurtis', 'canberra', 'chebyshev', 'cityblock',
+            'correlation', 'cosine', 'dice', 'euclidean', 'hamming',
+            'jaccard', 'jensenshannon', 'kulczynski1',
+            'mahalanobis', 'matching', 'minkowski', 'rogerstanimoto',
+            'russellrao', 'seuclidean', 'sokalmichener', 'sokalsneath',
+            'sqeuclidean', 'yule']
+    )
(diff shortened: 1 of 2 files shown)


holoviews/operation/element.py
```diff
@@ -1270,7 +1307,7 @@ def _process(self, element, key=None):
         sort_dims, dendros = [], {}
         for d in self.p.adjoint_dims:
             ddata = self._compute_linkage(dataset, d, self.p.main_dim)
-            order = [ddata["ivl"].index(v) for v in dataset.dimension_values(d)][::-1]
+            order = [ddata["ivl"].index(v) for v in dataset.dimension_values(d)]
```
hoxbro, line 1310: Removing this makes the output aligned with how scanpy looks (other than a reversed ordering of the axis):

image

image

hoxbro:
The push afterwards was just to rebase onto main.

Issue #6601, opened 2025-05-29
Support for Bokeh toolbar autohide


jbednar:
I [requested](https://github.com/bokeh/bokeh/issues/8284) that Bokeh add `autohide`  support for the toolbar in 2018, and eventually worked with @jsignell to get that [implemented](https://github.com/bokeh/bokeh/pull/8318), with the intention of making it the default in HoloViews. Making it the default never actually happened, due to some issues I vaguely remember being reported as "to do with layouts" and "in notebooks".  

Testing it today with Bokeh 3.6.2 and HoloViews 1.20.0 using code from [@SandervandenOord](https://stackoverflow.com/questions/59107482/how-to-autohide-the-bokeh-toolbar-when-using-holoviews), it seems like it does work on a single plot in a notebook:

```python
import numpy as np, pandas as pd, holoviews as hv, hvplot.pandas
hv.extension('bokeh', logo=False)

df = pd.DataFrame({'x': np.random.normal(size=30),
                   'y': np.random.normal(size=30)})

def enable_autohide(plot, element):
    bokeh_plot = plot.state
    bokeh_plot.toolbar.autohide = True

df.hvplot.scatter(height=200, width=400).opts(hooks=[enable_autohide], backend='bokeh')
```
<img width="439" alt="Image" src="https://github.com/user-attachments/assets/c2bc92c5-683c-4111-9e2f-f4417ea7a3f4" />

And as recalled, I can't get it to work using HoloViews notebook cell display with a layout. E.g. using a hook on each plot is accepted but has no effect:

```python
(df.hvplot.scatter(height=200, width=400).opts(hooks=[enable_autohide], backend='bokeh') + 
 df.hvplot.scatter(height=200, width=500).opts(hooks=[enable_autohide], backend='bokeh')).cols(1)
```

<img width="510" alt="Image" src="https://github.com/user-attachments/assets/d1620feb-b481-4c37-a857-53e298aece68" />

while trying to use a hook on the overall Layout (which is where I'd expect to set such an option since toolbars are being shared here) doesn't seem to work:

```python
(df.hvplot.scatter(height=200, width=400) + 
 df.hvplot.scatter(width=500)).cols(1).opts(hooks=[enable_autohide], backend='bokeh')
```

```
ValueError: Unexpected option 'hooks' for Layout type when using the 'bokeh' extension. No similar options found.
```

[...]

PR #6602, opened 2025-05-29
enh(bokeh): Add `hide_toolbar` options

hoxbro:
resolves https://github.com/holoviz/holoviews/issues/6601

TODO:

- [x] Rename to `autohide_toolbar`
- [x] Don't enable it if the toolbar is disabled.
- [x] See if we can move the parameter up a parent class to avoid repeating (likely `BokehPlot`)
- [x] Add a test for each of the basic layouts (if we can't move it up into the parent class)
- [x] Add docstring to the parameter. 
- [x] Document this feature _somewhere_

Still draft! Primarily to illustrate the basic requirements for it to function. Kept the status quo for now, and the name is also up for debate. 

``` python
import holoviews as hv
import hvplot.pandas
import numpy as np
import pandas as pd

hv.extension("bokeh", logo=False)

df = pd.DataFrame({"x": np.random.normal(size=30), "y": np.random.normal(size=30)})
df.hvplot.scatter(height=200, width=400).opts(hide_toolbar=True)
```
image
``` python
p = df.hvplot.scatter(height=200, width=400) + df.hvplot.scatter(height=200, width=500)
p.opts(hide_toolbar=True)
```

image

https://github.com/user-attachments/assets/dc7d1b4d-12c8-4dec-ae55-5a4193e58b2b

### DIFF
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -820,7 +820,8 @@ def save(obj, filename, fmt='auto', backend=None, resources='cdn', toolbar=None,
     if backend == "bokeh":
         if toolbar is not None:
             if toolbar:
-                obj = obj.opts(toolbar="above", autohide_toolbar=False, backend="bokeh", clone=True)
+                toolbar_location = obj.opts.get().kwargs.get('toolbar', 'right')
+                obj = obj.opts(toolbar=toolbar_location, autohide_toolbar=False, backend="bokeh", clone=True)
             else:
                 obj = obj.opts(toolbar=None, backend="bokeh", clone=True)
         elif (
--- a/holoviews/util/__init__.py
+++ b/holoviews/util/__init__.py
@@ -817,14 +817,17 @@ def save(obj, filename, fmt='auto', backend=None, resources='cdn', toolbar=None,
     """
     backend = backend or Store.current_backend
     renderer_obj = renderer(backend)
-    if (
-        not toolbar
-        and backend == "bokeh"
-        and (fmt == "png" or (isinstance(filename, str) and filename.endswith("png")))
-    ):
-        obj = obj.opts(toolbar=None, backend="bokeh", clone=True)
-    elif toolbar is not None and not toolbar:
-        obj = obj.opts(toolbar=None)
+    if backend == "bokeh":
+        if toolbar is not None:
+            if toolbar:
+                obj = obj.opts(toolbar="above", autohide_toolbar=False, backend="bokeh", clone=True)
+            else:
+                obj = obj.opts(toolbar=None, backend="bokeh", clone=True)
+        elif (
+            not toolbar
+            and (fmt == "png" or (isinstance(filename, str) and filename.endswith("png")))
+        ):
+            obj = obj.opts(toolbar=None, backend="bokeh", clone=True)
     if kwargs:
         renderer_obj = renderer_obj.instance(**kwargs)
     if isinstance(filename, Path):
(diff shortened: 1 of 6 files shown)


Azaya89:
https://github.com/user-attachments/assets/7e8f7a9c-6049-4312-a35e-6d45d1351e39

Azaya89:
> Thanks. Can you check what happens for `autohide=True` when doing `hv.save(toolbar=True)`? I'd hope that the toolbar is saved in that case, despite autohide being true.

Unfortunately I can't save to a png format with `hv.save` as I don't have geckodriver installed on my computer. I've been having issues with that for a while.

hoxbro:
> don't have geckodriver installed

You can install it with conda along with Firefox.

Azaya89:
> Thanks. Can you check what happens for `autohide=True` when doing `hv.save(toolbar=True)`? I'd hope that the toolbar is saved in that case, despite autohide being true.

The toolbar isn't saved
image

hoxbro:
> The toolbar isn't saved

You can likely disable autohide if show_toolbar is True inside `save()`

jbednar:
> The toolbar isn't saved

That confirms my hunch. :-/

Azaya89:
I've added the option to save the toolbar by setting `toolbar=True` to `hv.save()`

```python
import holoviews as hv
import numpy as np
import pandas as pd

hv.extension("bokeh", logo=False)

df = pd.DataFrame({"x": np.random.normal(size=30), "y": np.random.normal(size=30)})

plot = hv.Scatter(df)
plot.opts(autohide_toolbar=True)

hv.save(plot, 'autohide_plot.png') # default

hv.save(plot, 'autohide_plot2.png', toolbar=True) # toolbar=True
```
**Default**
image

**toolbar=True**
image

holoviews/util/__init__.py
```diff
@@ -817,14 +817,17 @@ def save(obj, filename, fmt='auto', backend=None, resources='cdn', toolbar=None,
...
-        and backend == "bokeh"
-        and (fmt == "png" or (isinstance(filename, str) and filename.endswith("png")))
-    ):
-        obj = obj.opts(toolbar=None, backend="bokeh", clone=True)
-    elif toolbar is not None and not toolbar:
-        obj = obj.opts(toolbar=None)
+    if backend == "bokeh":
+        if toolbar is not None:
+            if toolbar:
+                obj = obj.opts(toolbar="above", autohide_toolbar=False, backend="bokeh", clone=True)
```
hoxbro: Why is above needed?

Azaya89: The idea is to make sure the toolbar becomes visible when `toolbar=True` for bokeh plots even if `autohide_toolbar=True` in the original plot object. This implementation doesn't alter the original object but instead creates a clone for save purposes.

I can add a code comment above that line if it helps.

hoxbro: I understand that, but why did you change to the `above`? If a user has set something else, we should not overwrite it.

Azaya89: OK! The reason is because it is not possible to simply set `toolbar=True` there as the parameter only accepts valid options from this list: [above, below, left, right, disable, None]

See https://github.com/holoviz/holoviews/blob/main/holoviews/plotting/bokeh/element.py#L328

So, I just chose one of the options that are not "None" or "disable".

hoxbro: You should perform a check before overwriting. This can be done with something like this:

``` python
import holoviews as hv

hv.extension("bokeh")

p = hv.Curve([]).opts(toolbar="right")
p.opts.get().kwargs
```

Azaya89: Thanks.

philippjfr (approved):
Looks great!

PR #6604, opened 2025-05-30
docs: Update Customizing_Plots.ipynb

michaelaye:
fixing broken anchor link

### DIFF
--- a/examples/user_guide/Customizing_Plots.ipynb
+++ b/examples/user_guide/Customizing_Plots.ipynb
@@ -41,7 +41,7 @@
     "        - [Logarithmic axes](#log-axes)\n",
     "        - [Datetime axes](#datetime-axes)\n",
     "        - [Categorical axes](#categorical-axes)\n",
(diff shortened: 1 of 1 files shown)


michaelaye:
cannot fix that fail: `Error: No release type found in pull request title "docs : Update Customizing_Plots.ipynb". Add a prefix to indicate what kind of release this pull request corresponds to. For reference, see https://www.conventionalcommits.org/` added `docs` to the PR title as advised, still doesn't run through.

michaelaye:
> Thank you for the PR 👍
> 
> > added docs to the PR title as advised, still doesn't run through.
> 
> You had an extra space, have updated it

yeah but the instructions don't say it has to end with colon directly, right?

hoxbro:
> yeah but the instructions don't say it has to end with colon directly, right?

Correct. Would you be willing to update the error message in a new PR? 

Should be adding a `subjectPatternError` to https://github.com/holoviz/holoviews/blob/main/.github/workflows/conventional_commits.yaml, see https://github.com/amannn/action-semantic-pull-request.

PR #6608, opened 2025-06-05
build: Move to dependency groups

hoxbro:
Applies [PEP-734](https://peps.python.org/pep-0735/), so we don't expose tests to the end user, but still make them available as an option when working with the repository.

Command to run is `pip install --group tests .`


maximlt (approved):
LGTM (hopefully one day we will no longer have to duplicate stuff between pyproject.toml and pixi.toml!)

PR #6609, opened 2025-06-09
enh: Allow to not have to specify column in selector

hoxbro:
One of the main "complaints" about `selector` is that you have to specify the column for it. This uses the `rd.SpecialColumn.RowIndex` which was introduced in https://github.com/holoviz/datashader/pull/1234 and released in 1.15.1. I'm not entirely sure if I'm using RowIndex correctly, so I need someone to take a look at the code. 

``` python
import itertools

import datashader as ds
import numpy as np
import pandas as pd
import panel as pn

import holoviews as hv
from holoviews.operation.datashader import datashade, dynspread, rasterize

hv.extension("bokeh")

num = 10000
seed = np.random.default_rng(1)


dists = {
    cat: pd.DataFrame(
        {
            "x": seed.normal(x, s, num),
            "y": seed.normal(y, s, num),
            "cat": cat,
            "s": s,
            "val": val,
        }
    )
    for x, y, s, val, cat in [
        (2, 2, 0.03, 0, "d1"),
        (2, -2, 0.10, 1, "d2"),
        (-2, -2, 0.50, 2, "d3"),
        (-2, 2, 1.00, 3, "d4"),
        (0, 0, 3.00, 4, "d5"),
    ]
}

df["y"] += 100  # Offset to distinguish x from y
points = hv.Points(df)

plot = rasterize(points, aggregator=ds.first("s"), selector=ds.last()).opts(tools=["hover"])
plot
```

image

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -380,6 +380,8 @@ class aggregate(LineAggregationOperation):
             category = agg_fn.cat_column
         else:
             category = agg_fn.column if isinstance(agg_fn, ds.count_cat) else None
+        if sel_fn and sel_fn.column is None:
+            sel_fn.column = rd.SpecialColumn.RowIndex
 
         if overlay_aggregate.applies(element, agg_fn, line_width=self.p.line_width, sel_fn=sel_fn):
             params = dict(
@@ -428,7 +430,7 @@ class aggregate(LineAggregationOperation):
                 str(sel_fn)
                 if DATASHADER_GE_0_18_1
                 else f"{type(sel_fn).__name__}({getattr(sel_fn, 'column', '...')!r})"
-            )
+            ).replace(repr(rd.SpecialColumn.RowIndex), "")
         else:
             agg = self._apply_datashader(dfdata, cvs_fn, agg_fn, agg_kwargs, x, y, agg_state)
 
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -380,8 +380,8 @@ class aggregate(LineAggregationOperation):
             category = agg_fn.cat_column
         else:
             category = agg_fn.column if isinstance(agg_fn, ds.count_cat) else None
-        if sel_fn and sel_fn.column is None:
-            sel_fn.column = rd.SpecialColumn.RowIndex
+        if DATASHADER_GE_0_15_1 and sel_fn and sel_fn.column is None:
+            sel_fn = type(sel_fn)(column=rd.SpecialColumn.RowIndex)
 
         if overlay_aggregate.applies(element, agg_fn, line_width=self.p.line_width, sel_fn=sel_fn):
             params = dict(
--- a/examples/user_guide/Interactive_Hover_for_Big_Data.ipynb
+++ b/examples/user_guide/Interactive_Hover_for_Big_Data.ipynb
@@ -151,7 +151,7 @@
     "- `ds.first(<column>)`: Select the first value in the column\n",
     "- `ds.last(<column>)`: Select the last value in the column\n",
     "\n",
-    "Under the hood, the selected value has a corresponding row-index, which allows the collection and presentation of the entire row in the hover tooltip.\n",
+    "Under the hood, the selected value has a corresponding row-index, which allows the collection and presentation of the entire row in the hover tooltip. If no column is set, the selector will use the index to determine the sample.\n",
     "\n",
     "![selector ](../assets/selector.png)"
    ]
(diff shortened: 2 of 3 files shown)


hoxbro:
@maximlt, I think this would make it possible to do something like this in hvplot `df.hvplot(..., selector="last")`

jbednar:
It's still on my list to try to figure out the recent improvements to hover information such as https://github.com/holoviz/holoviews/pull/6568 . Is the first section showing an aggregation, while the second shows a sample?  If so, what does "x_y s" mean here? I can't tell what aggregation that would be. In the second section, is it really showing a Python concept like "last()" (i.e. a repr of a Python function name)? If so that doesn't seem right; Python code shouldn't be leaking into the hover text. Maybe it could say "last sample"? Whatever it says should somehow indicate that it's a sample, without showing Python code.

maximlt:
> Is the first section showing an aggregation, while the second shows a smaple? If so, what does "x_y s" mean here?

I think that UX deficiency has been mentioned in the long discussion in https://github.com/holoviz/holoviews/pull/6568. IIRC only a `count` aggregator ends up being displayed nicely. I'm not sure this is tracked in an issue.

> In the second section, is it really showing a Python concept like "last()" (i.e. a repr of a Python function name)? If so that doesn't seem right; Python code shouldn't be leaking into the hover text. Maybe it could say "last sample"? Whatever it says should somehow indicate that it's a sample, without showing Python code.

There's been a pretty long discussion in https://github.com/holoviz/holoviews/pull/6568 about what content should be displayed and how to display it. I suggest opening a new issue and linking to https://github.com/holoviz/holoviews/pull/6568 if you think this needs to be improved somehow.

holoviews/operation/datashader.py
```diff
@@ -380,6 +380,8 @@ def _process(self, element, key=None):
             category = agg_fn.cat_column
         else:
             category = agg_fn.column if isinstance(agg_fn, ds.count_cat) else None
+        if sel_fn and sel_fn.column is None:
```
maximlt: Should this be limited to `first` and `last` reductions only? I've tried `selector=ds.max()` and it did show something in the tooltip, but I don't understand how it was computed (which column was used to compute the max() row?)

image

holoviews/operation/datashader.py
```diff
@@ -380,6 +380,8 @@ def _process(self, element, key=None):
...
+            sel_fn.column = rd.SpecialColumn.RowIndex
```
maximlt: Would it make sense to create a new selector internally (with something like type(sel_fn)(column=rd.SpecialColumn.RowIndex)) instead of modifying the one passed by the user?

holoviews/operation/datashader.py
maximlt: Maybe also add a `DATASHADER_GE_0_15_1` condition?

hoxbro: So. I was not 100%, but I just checked `RowIndex` uses the index of the DataFrame, meaning max and min will equal first and last, respectively. 

``` python
import datashader as ds
import datashader.reductions as rd
import numpy as np
import pandas as pd
import xarray as xr

num = 10000
seed = np.random.default_rng(1)

dists = {
    cat: pd.DataFrame(
        {
            "x": seed.normal(x, s, num),
            "y": seed.normal(y, s, num),
            "cat": cat,
            "s": s,
            "val": val,
        }
    )
    for x, y, s, val, cat in [
        (2, 2, 0.03, 0, "d1"),
        (2, -2, 0.10, 1, "d2"),
        (-2, -2, 0.50, 2, "d3"),
        (-2, 2, 1.00, 3, "d4"),
        (0, 0, 3.00, 4, "d5"),
    ]
}

df["index_col"] = df.index
canvas = ds.Canvas(plot_width=400, plot_height=400)

agg_first = canvas.points(df, "x", "y", agg=ds.first(rd.SpecialColumn.RowIndex))
agg_min = canvas.points(df, "x", "y", agg=ds.min(rd.SpecialColumn.RowIndex))
agg_max = canvas.points(df, "x", "y", agg=ds.max(rd.SpecialColumn.RowIndex))
agg_last = canvas.points(df, "x", "y", agg=ds.last(rd.SpecialColumn.RowIndex))
agg_first_x = canvas.points(df, "x", "y", agg=ds.first("x"))
agg_first_index = canvas.points(df, "x", "y", agg=ds.first("index_col"))

assert agg_first.equals(agg_min)
assert agg_last.equals(agg_last)
assert not agg_first.equals(agg_last)
assert not agg_first.equals(agg_first_x)
assert agg_first.equals(agg_first_index)
```

hoxbro: I'm open to only allowing this with first and last.

hoxbro:
I will merge this. The design of HoverTool is out of scope for this PR and will be tackled in other PR(s).

Issue #6445, opened 2024-11-08
Central Deprecation Proposals Issue


hoxbro:
As we now have access to sub-issues, this is a good way to have a central issue for discussion. The name next to a proposal will be assigned to create a sub-issue, and when that is created, I will replace it with the issue number.

# Effort level: Big

1. Remove IPython magic (@hoxbro)
2. `.options` method (duplicative of `.opts`) (@jlstevens)
3. Limit `.to` API to only be available for `Dataset`  (@hoxbro)

# Effort level: Medium

1. Scrubber (@jlstevens)
2. `color_index` and family (@hoxbro)
    a. Check if it warns (Edit: It does)
    b. If it warns, we could likely remove it in the next minor; otherwise, we should make it warn in the next patch version.
3. Remove automatic `rc-file`  (@hoxbro)
    a. Replace it with `hv.extensions(rc_file="~/.holoviewsrc")`

# Effort level: Small

1. Remove unused configs `future_deprecations` and `warn_options_call` 
2. Remove dash support

---

These deprecations points were made in collaboration with @jlstevens and @azaya89

hoxbro:
Another one for me is to deprecate the default plotting backend when running `hv.extension`.

hoxbro:
### Reason for the deprecation of IPython “magic” commands (`%opts`, `%%opts`, `%output`) in HoloViews
* In version **1.21.0** (not yet released), the notebook magic commands `%%opts`, `%opts`, and `%output` were officially **deprecated**, with plans to remove them entirely in version **1.23.0**.
* These magics were once useful for concise styling in notebooks—e.g., applying options via `%opts Image [cmap='viridis']`—but they came with significant drawbacks:
  1. **Non‑standard syntax**: They rely on IPython-specific magic syntax, making code incompatible with other environments (e.g., Panel / Bokeh server and pure Python scripts).
  2. **Obscure parsing**: Their terse, non-Pythonic syntax can confuse even experienced users unfamiliar with HoloViews, hindering readability and maintainability.
* As of around **5 years ago**, the maintainers transitioned to a more robust and explicit **Pythonic API**:
  * Use of `.opts(...)` methods for per-object styling.
  * Use of `opts.defaults(...)` for global configuration.
  * Use of the `hv.output(...)` function to specify plotting backends and rendering options.
* :compass: **Why this matters**:
  | Issue       | Advantage of Pythonic API                |
  | ----------- | ---------------------------------------- |
[...]

PR #6610, opened 2025-06-09
enh!: Deprecate old features

hoxbro:
Part of the deprecation sprint in the fall, see https://github.com/holoviz/holoviews/issues/6445

1. Deprecate auto RC file loading (removed in 1.23.0)
This will now require a user to specify the HOLOVIEWSRC environment variable. Have refactored the code a bit. 

2. Remove unused `hv.configs`: `future_deprecations` and `warn_options_call`  (no timeline)
I'm also removing the unused parameters `future_deprecations` and `warn_options_call`, and adding a warning if a user has specified them for some reason. 

~~3. Deprecate dash  (removed in 1.22.0)
We don't test Dash at all, so I have moved this to the next minor version after this deprecation has been included.~~

4. Deprecate IPython magic (removed in 1.23.0)
This is done with a warning when it is called and a note in the notebook. 
	``` python
	import holoviews as hv
	
	hv.extension("bokeh")
	
	%%opts Curve (color="red" line_width=1.5)
	%%opts Curve  [height=200 width=900 xaxis=None tools=['hover']]
	
	hv.Curve([1, 2,3])
	```
5. Deprecate bare `hv.extension()`  (removed in 1.23.0)

--- 
image

### DIFF
--- a/doc/user_guide/Configuring.md
+++ b/doc/user_guide/Configuring.md
@@ -34,28 +34,15 @@ import holoviews as hv
 hv.extension(case_sensitive_completion=True)
 ```
 
-## The holoviews.rc file
+## The HoloViews RC file
 
-HoloViews searches for the first rc file it finds in the following places (in order):
-
-1. `HOLOVIEWSRC` environment variable if it is a valid path
-1. `holoviews.rc` in the parent directory of the top-level `__init__.py` file (useful for developers working out of the HoloViews git repo)
-1. `~/.holoviews.rc`
-1. `~/.config/holoviews/holoviews.rc`
-
-The rc file is a Python script and can be loaded with:
+If the `HOLOVIEWSRC` environment variable is a valid path, HoloViews will load the configuration from that file.
+This allows users to set their preferred options globally without needing to modify their scripts each time.
+An example of an RC file to include the various options discussed above might look like this:
 
 ```python
 import holoviews as hv
 
-hv.extensions(load_rc=True)
-```
-
-An example rc file to include various options discussed above might look like this:
-
-```python
-import holoviews as hv
-
-hv.config(warn_options_call=True)
+hv.config(no_padding=True)
 hv.extension.case_sensitive_completion = True
 ```
--- a/doc/user_guide/Configuring.md
+++ b/doc/user_guide/Configuring.md
@@ -3,25 +3,19 @@
 ## `hv.config` settings
 
 The default HoloViews installation will use the latest defaults and options available, which is appropriate for new users.
-If you want to work with code written for older HoloViews versions, you can use the top-level `hv.config` object to control various backwards-compatibility options:
-
-- `future_deprecations`: Enables warnings about future deprecations.
-- `warn_options_call`: Warn when using the to-be-deprecated `__call__` syntax for specifying options, instead of the recommended `.opts` method.
-
-It is recommended you set `warn_options_call` to `True` in your holoviews.rc file (see section below).
 
 It is possible to set the configuration using `hv.config` directly:
 
 ```python
 import holoviews as hv
 
-hv.config(future_deprecations=True)
+hv.config(no_padding=True)
 ```
 
 However, because in some cases this configuration needs to be declared before the plotting extensions are imported, the recommended way of setting configuration options is:
 
 ```python
-hv.extension("bokeh", config=dict(future_deprecations=True))
+hv.extension("bokeh", config=dict(no_padding=True))
 ```
 
 In addition to backwards-compatibility options, `hv.config` holds some global options:
@@ -44,13 +38,20 @@ hv.extension(case_sensitive_completion=True)
 
 HoloViews searches for the first rc file it finds in the following places (in order):
 
+1. `HOLOVIEWSRC` environment variable if it is a valid path
 1. `holoviews.rc` in the parent directory of the top-level `__init__.py` file (useful for developers working out of the HoloViews git repo)
-2. `~/.holoviews.rc`
-3. `~/.config/holoviews/holoviews.rc`
+1. `~/.holoviews.rc`
+1. `~/.config/holoviews/holoviews.rc`
 
-The rc file location can be overridden via the `HOLOVIEWSRC` environment variable.
+The rc file is a Python script and can be loaded with:
 
-The rc file is a Python script, executed as HoloViews is imported. An example rc file to include various options discussed above might look like this:
+```python
+import holoviews as hv
+
+hv.extensions(load_rc=True)
+```
+
+An example rc file to include various options discussed above might look like this:
 
 ```python
 import holoviews as hv
--- a/examples/user_guide/Notebook_Magics.ipynb
+++ b/examples/user_guide/Notebook_Magics.ipynb
@@ -8,6 +8,15 @@
    ]
   },
   {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    ":::{error} Deprecation\n",
+    "Notebook Magic has been deprecated in version 1.21.0 and will be removed in version 1.23.0.\n",
+    ":::"
+   ]
+  },
+  {
    "cell_type": "code",
    "execution_count": null,
    "metadata": {},
--- a/examples/user_guide/Notebook_Magics.ipynb
+++ b/examples/user_guide/Notebook_Magics.ipynb
@@ -12,7 +12,7 @@
    "metadata": {},
    "source": [
     ":::{error} Deprecation\n",
-    "Notebook Magic has been deprecated in version 1.21.0 and will be removed in version 1.23.0.\n",
+    "Notebook magic commands `%%opts`, `%opts`, and `%output` have been deprecated in version 1.21.0 and will be removed in version 1.23.0.\n",
     ":::"
    ]
   },
--- a/holoviews/core/util/__init__.py
+++ b/holoviews/core/util/__init__.py
@@ -111,9 +111,6 @@ class Config(param.ParameterizedFunction):
 
     """
 
-    future_deprecations = param.Boolean(default=False, doc="""
-       Whether to warn about future deprecations""")
-
     image_rtol = param.Number(default=10e-4, doc="""
       The tolerance used to enforce regular sampling for regular,
       gridded data where regular sampling is expected. Expressed as the
@@ -122,12 +119,6 @@ class Config(param.ParameterizedFunction):
 
     no_padding = param.Boolean(default=False, doc="""
        Disable default padding (introduced in 1.13.0).""")
-
-    warn_options_call = param.Boolean(default=True, doc="""
-       Whether to warn when the deprecated __call__ options syntax is
-       used (the opts method should now be used instead). It is
-       recommended that users switch this on to update any uses of
-       __call__ as it will be deprecated in future.""")
 
     default_cmap = param.String(default='kbc_r', doc="""
        Global default colormap. Prior to HoloViews 1.14.0, the default
@@ -144,6 +135,10 @@ class Config(param.ParameterizedFunction):
        1.14.0, the default value was the 'RdYlBu_r' colormap.""")
 
     def __call__(self, **params):
+        # Old parameters, which does nothing
+        params.pop("future_deprecations", None)
+        params.pop("warn_options_call", None)
+
         self.param.update(**params)
         return self
 
--- a/holoviews/core/util/__init__.py
+++ b/holoviews/core/util/__init__.py
@@ -23,6 +23,7 @@ from types import FunctionType, GeneratorType
 import numpy as np
 import param
 
+from ...util.warnings import warn
 from .dependencies import (  # noqa: F401
     NUMPY_GE_2_0_0,
     NUMPY_VERSION,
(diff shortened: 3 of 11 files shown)


holoviews/plotting/plotly/dash.py
```diff
@@ -38,6 +39,8 @@
 from dash import callback_context
 from dash.dependencies import Input, Output, State
 
+deprecated("1.22.0", "dash")
```
hoxbro: We don't test Dash at all, so I have moved this to the next minor version after this deprecation has been included.

holoviews/util/warnings.py
```diff
@@ -49,7 +49,7 @@ def find_stack_level():
     return stacklevel
 
 
-def deprecated(remove_version, old, new=None, extra=None):
+def deprecated(remove_version, old, new=None, extra=None, repr_old=True, repr_new=True):
```
hoxbro:
```suggestion
def deprecated(remove_version, old, new=None, extra=None, *, repr_old=True, repr_new=True):
```

holoviews/ipython/magics.py
```diff
@@ -320,6 +322,9 @@ def opts(self, line='', cell=None):
         util.parser.OptsSpec.
 
         """
+        with warnings.catch_warnings():
+            warnings.simplefilter("always", HoloviewsDeprecationWarning)
```
hoxbro: Always because it happens nested in IPython magic

holoviews/plotting/plotly/dash.py
philippjfr: Can we try confirming if this works at all? And then come up with a plan on potentially putting it somewhere.

hoxbro: Just tried the examples [here](https://dash.plotly.com/holoviews).

The simple examples worked, but the more advanced did not:
- Combining Datashader and Linked Selections
- Mapbox datashader and linked selections
- All of the streams examples

holoviews/util/__init__.py
```diff
@@ -710,6 +710,7 @@ def __call__(self, *args, **params):
             if p in self._backends:
                 imports.append((p, self._backends[p]))
         if not imports:
+            deprecated("1.23.0", "hv.extension()", 'hv.extension("matplotlib")')
```
maximlt: I found the message a little confusing starting with `hv.extension()` is deprecated. Suggesting `"Calling `hv.extension()` without argument"`.

hoxbro: image

holoviews/plotting/plotly/dash.py
philippjfr: Okay, let's deprecate and if someone complains we can decide to provide it in some other form.

philippjfr (approved):

doc/user_guide/Configuring.md
maximlt, line 1: Unrelated, but just realized this file is not part of the TOC but is listed on `/user_guide/index.html` as the first supplementary guide:
image

examples/user_guide/Notebook_Magics.ipynb
```diff
@@ -7,6 +7,15 @@
     "# Notebook Magics"
    ]
   },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    ":::{error} Deprecation\n",
+    "Notebook Magic has been deprecated in version 1.21.0 and will be removed in version 1.23.0.\n",
```
maximlt:
```suggestion
    "Notebook magic commands `%%opts`, `%opts`, and `%output` have been deprecated in version 1.21.0 and will be removed in version 1.23.0.\n",
```

holoviews/core/util/__init__.py
```diff
@@ -144,6 +136,12 @@ class Config(param.ParameterizedFunction):
        1.14.0, the default value was the 'RdYlBu_r' colormap.""")
 
     def __call__(self, **params):
+        # Old parameters, removed in HoloViews 1.21.0
+        if params.pop("future_deprecations", None):
```
maximlt, line 140: Was this added because these two parameters were not deprecated?

holoviews/plotting/plotly/dash.py
maximlt: > We don't test Dash at all

In general, I don't think the fact that a feature is untested means that it makes it easier to deprecate it. So how do we motivate the deprecation of this feature?

I didn't know about https://dash.plotly.com/holoviews, it's some nice docs. Should we let the Dash team that we plan to deprecate this?

Out of curiosity, I did a quick Github search for `holoviews.plotting.plotly.dash` (https://github.com/search?q=holoviews.plotting.plotly.dash&type=code) and it returned some recent hits like https://github.com/manishvenu/Global-Lake-Hydrology-Explorer/blob/21d7113cbe7153958b6dfe9d9d1365cae29ae638/GLHE/LIME/LIME_sample_dashboard.py#L8. Some maybe that code kind of still works, I don't know.

holoviews/core/util/__init__.py
hoxbro: Yes. I don't think anybody is using it, and they are not used anywhere in the code.

holoviews/plotting/plotly/dash.py
hoxbro: > In general, I don't think the fact that a feature is untested means that it makes it easier to deprecate it. So how do we motivate the deprecation of this feature?

For me, it is an indication of how well we support something. If something is undocumented and untested, we are at best naive and at worst negligent in assuming it will continue to work. 

It isn't IMO ok to keep this functionality in if it "only" works in some cases. But never really have any plans to fix it when it breaks.   

As this is OSS, I think it is acceptable to reduce our maintenance burden by being upfront about what we want to support.  

> I didn't know about [dash.plotly.com/holoviews](https://dash.plotly.com/holoviews), it's some nice docs. Should we let the Dash team that we plan to deprecate this?

Yes, but I will wait a bit to do so after the release, to see if there is a lot of pushback. 

> Some maybe that code kind of still works, I don't know.

The way the dash module is built is similar to a script, allowing people to copy it over to their project if needed.

hoxbro:
> Who knows, there might be notebook magic commands fan out there and it'd be nice for them to be able to track down why it's been deprecated.

They can always open an issue, and we can answer it. 

> As a note, since this is just one PR for multiple deprecations, I think the release notes should include one item per feature deprecated.

It was always the plan :+1:

maximlt:
> They can always open an issue, and we can answer it.

Sure, but please document the motivations for removing features. From HEP2:

> Before implementing a deprecation, maintainers:
> [ ] **Must** motivate the best they can why the feature is deprecated (e.g. old and unused API) in an issue or in the Pull Request (PR) deprecating the feature.

holoviews/plotting/plotly/dash.py
maximlt: > If something is undocumented and untested

I didn't know it was undocumented until now, and untested until you wrote it in a comment in this PR. In fact, it seems to be documented by Plotly pretty well.

> we are at best naive and at worst negligent in assuming it will continue to work

Yeah I think you're exaggerating here, saying we're in between naive and negligent... There's a lot of code out there in the wild that is in this very state, and luckily, people don't remove it, because that state doesn't prevent it from running fine :)

Overall I'm not against deprecating this feature. It's just that I found the argument "we remove it because we don't test it" a bit weak.

hoxbro: > In fact, it seems to be documented by Plotly pretty well.

But many of the plotly examples do not work... The document also uses an old API of their own as Mapbox has been [deprecated](https://plotly.com/python/mapbox-to-maplibre/) in recent Plotly versions.

> Yeah I think you're exaggerating here, saying we're in between naive and negligent... There's a lot of code out there in the wild that is in this very state, and luckily, people don't remove it, because that state doesn't prevent it from running fine :)

I'm definitely exaggerating. 

We don't remove the code from the surface of the earth; it remains in old versions of HoloViews and in the git history.  Keeping the code in gives the user the expectation that we support dash, which is not the case (we barely mention it at all in our code base). 

For the "a lot of the code in the wild," I assume you do not expect that it will be supported if it does not work, that is, at least what I do.

> Overall I'm not against deprecating this feature. It's just that I found the argument "we remove it because we don't test it" a bit weak.

This was only a reason for doing it in the next version (1.22.0) and not in (1.23.0). Can move it to 1.23.0 if you want.

hoxbro:
> Sure, but please document the motivations for removing features. From HEP2:

I tried to follow it, but it must have slipped as the hep is very dense. I have added a comment [here](https://github.com/holoviz/holoviews/issues/6445#issuecomment-2970525270).

holoviews/plotting/plotly/dash.py
philippjfr: Let me spend 15 minutes, trying it.

philippjfr: I just ran through all the examples in the Plotly docs and they all worked without issue for me (apart from removing the mapbox token).

Given that, I do not want to deprecate without having a clear plan on what happens with the code.

philippjfr: Wonder what the difference between our testing setups was. Just to prove I'm not making things up:

Screenshot 2025-06-16 at 14 57 37

hoxbro: It's not entirely broken, but when I zoom, it doesn't work correctly for me. I also remember it looking worse last time I checked. 

<details><summary>Details</summary>
<p>

Ran with `python example.py`


``` python
import numpy as np
import pandas as pd
from dash import Dash, html
from plotly.colors import sequential
from plotly.data import carshare

import holoviews as hv
from holoviews.plotting.plotly.dash import to_dash

# Mapbox token (replace with your own token string)
# mapbox_token = open(".mapbox_token").read()

# Convert from lon/lat to web-mercator easting/northing coordinates
df_original = carshare()
df_original["easting"], df_original["northing"] = hv.Tiles.lon_lat_to_easting_northing(
    df_original["centroid_lon"], df_original["centroid_lat"]
)

# Duplicate carshare dataframe with noise to produce a larger dataframe
df = pd.concat([df_original] * 5000)
df["easting"] = df["easting"] + np.random.randn(len(df)) * 400
df["northing"] = df["northing"] + np.random.randn(len(df)) * 400

# Build HoloViews Dataset and visual elements
dataset = hv.Dataset(df).redim.label(car_hours="Car Hours")
points = hv.Points(
    df, ["easting", "northing"]
).opts(color="crimson")
tiles = hv.Tiles().opts(mapstyle="light")
overlay = tiles * datashade(points, cmap=sequential.Plasma).opts(width=800)

# Build histogram of car_hours column
hist = hv.operation.histogram(dataset, dimension="car_hours", normed=False)

# Use plot hook to set the default drag mode to horizontal box selection
def set_hist_dragmode(plot, element):
    fig = plot.state
    fig['layout']['dragmode'] = "select"
    fig['layout']['selectdirection'] = "h"

hist.opts(hooks=[set_hist_dragmode], color="rebeccapurple", margins=(50, 50, 50, 30))

# Link selections across datashaded points and histogram
selection_linker = hv.selection.link_selections.instance()
linked_points = selection_linker(
    tiles * datashade(points, cmap=sequential.Plasma)
).opts(title="Datashader with %d points" % len(dataset), margins=(50, 0, 50, 30))
linked_hist = selection_linker(hist)

# Build app
app = Dash()
components = to_dash(app, [linked_points, linked_hist], reset_button=True)

app.layout = html.Div(
    components.children
)

if __name__ == '__main__':
    app.run(debug=True)
```

</p>
</details> 

https://github.com/user-attachments/assets/c104f1b6-315e-48ff-bbbb-ab48e1d3c65f

--- 

Another example is here:

<details><summary>Details</summary>
<p>

Ran with `python example.py`

```python
from dash import Dash, html

import numpy as np
import holoviews as hv
from holoviews import streams
from holoviews.plotting.plotly.dash import to_dash

# Declare distribution of Points
points = hv.Points(
    np.random.multivariate_normal((0, 0), [[1, 0.1], [0.1, 1]], (1000,))
)

# Declare points selection
sel = streams.Selection1D(source=points)

# Declare DynamicMap computing mean y-value of selection
mean_sel = hv.DynamicMap(
    lambda index: hv.HLine(points['y'][index].mean() if index else -10),
    kdims=[], streams=[sel]
)

# Declare a Bounds stream and DynamicMap to get box_select geometry and draw it
box = streams.BoundsXY(source=points, bounds=(0,0,0,0))
bounds = hv.DynamicMap(lambda bounds: hv.Bounds(bounds), streams=[box])

# Declare DynamicMap to apply bounds selection
dmap = hv.DynamicMap(lambda bounds: points.select(x=(bounds[0], bounds[2]),
                                                  y=(bounds[1], bounds[3])),
                     streams=[box])

# Compute histograms of selection along x-axis and y-axis
yhist = hv.operation.histogram(
    dmap, bin_range=points.range('y'), dimension='y', dynamic=True, normed=False
)
xhist = hv.operation.histogram(
    dmap, bin_range=points.range('x'), dimension='x', dynamic=True, normed=False
)

# Combine components and display
layout = points * mean_sel * bounds << yhist << xhist

# Create App
app = Dash()
components = to_dash(
    app, [layout], reset_button=True, use_ranges=False,
)

app.layout = html.Div(components.children)

if __name__ == "__main__":
    app.run(debug=True)

``` 

</p>
</details> 


https://github.com/user-attachments/assets/f321b996-542a-469e-8d5b-78887a9e1dfe


--- 

Zooming with responsive also looks wrong, but this could be a HoloViews issue:

image

hoxbro: Have reverted the deprecation for now.

philippjfr: Thanks, spent a few minutes looking at the implementation and have a fix for all the issues except for the responsive problem.

PR #6613, opened 2025-06-11
fix(bokeh): Add option to convert back to `size` with Points

hoxbro:
This would previously fail:

``` python
import holoviews as hv

hv.extension("bokeh")

plot = hv.Points([1, 2, 3]).opts(radius=0.1)
plot + plot.clone().opts(radius=None, size=2)
```

image

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -72,11 +72,13 @@ class PointPlot(LegendPlot, ColorbarPlot):
     _batched_style_opts = line_properties + fill_properties + ['size', 'marker', 'angle']
 
     def _init_glyph(self, plot, mapping, properties):
-        if "radius" in properties:
+        if properties.get("radius") is not None:
             self._plot_methods = dict(single='circle', batched='circle')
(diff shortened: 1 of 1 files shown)


holoviews/plotting/bokeh/chart.py
```diff
@@ -72,11 +72,13 @@ class PointPlot(LegendPlot, ColorbarPlot):
     _batched_style_opts = line_properties + fill_properties + ['size', 'marker', 'angle']
 
     def _init_glyph(self, plot, mapping, properties):
-        if "radius" in properties:
+        if properties.get("radius") is not None:
```
hoxbro, line 75:
```suggestion
        if properties.get("radius"):
```

Will also disable it for `0` or `False`, not sure about this.

philippjfr: I'd leave it as is.

philippjfr, line 75: There's almost an argument to make it a property, though that would require a separate options lookup.

philippjfr (approved):

PR #6614, opened 2025-06-11
enh: Allow passing dimension as style mapping option

philippjfr:
You should be able to provide dimensions (instead of just strings) when performing style mapping, e.g.:

```python
import holoviews as hv
hv.extension('bokeh')

x = hv.Dimension('x')
y = hv.Dimension('y')

p = hv.Points([1, 2, 3], kdims=[x, y])
p.opts(color=x)
```


maximlt:
That should probably be documented somewhere (at the very least in a docstring).

Issue #6615, opened 2025-06-12
Executing hv.extension("bokeh") takes a lot of time after importing geoviews


TheoMathurin:
<details>
    <summary>Software Version Info</summary>

```plaintext
holoviews 1.20.2
bokeh 3.7.3
```
</details>

First reported [here](https://github.com/holoviz/holoviews/issues/4741) (now locked), I'd like to bring up this issue again as there is a significant discrepancy between the loading time reported at the time in the aforementioned issue and what I'm experiencing. A plain `hv.extension("bokeh")` in a fresh environment takes a minimum of about 5 seconds to load on my laptop (CPU is i5-10310U, with Ubuntu's performance mode enabled, may be considerably more with less favourable settings).

Execution in production servers may be somewhat faster but nonetheless such loading time for the user that happens to open the first session is unacceptable. If you load it beforehand in the setup script (`--setup` argument to `panel serve`), it will still negatively impact development.

In any case I'd like to discuss this in case there is real performance degradation with recent versions, or if there are other factors pertaining to execution context that I don't know of but might impact loading time.

EDIT: it seems like what causes the long loading time is importing geoviews before loading the bokeh extension.

MRE:
```python
import holoviews as hv
import geoviews as gv
from time import monotonic

start = monotonic()
hv.extension("bokeh")
print(monotonic() - start)
```

hoxbro:
I don't get these numbers on the main branch; I get around 0.7 seconds. If I do a panel import before the time, it goes down to 0.12 seconds.

Can you show the output of `hv.show_versions()`?

How did you run the script just with `python example.py`, and did you run it more than once?

hoxbro:
I have tried some lazy importing in https://github.com/holoviz/holoviews/pull/6570 and https://github.com/holoviz/holoviews/pull/6476. However, `hv.extension()` imports the panel and calls `pn.extension` behind the scenes.

TheoMathurin:
Ok I think I found the culprit.

Sorry actually my MRE was not what I ran when testing this. I accidentally included a geoviews import, and this seems to considerably lengthen the loading time of the bokeh extension. I will edit my original post in that sense.

hoxbro:
I tracked down the biggest offender in #6616. This does not fix everything, but most of the other problems likely need to be fixed in datashader or geoviews itself.

maximlt:
> Execution in production servers may be somewhat faster but nonetheless such loading time for the user that happens to open the first session is unacceptable

[...]

PR #6616, opened 2025-06-12
fix: Lazy load `datashader.bundling`

hoxbro:
Fixes https://github.com/holoviz/holoviews/issues/6615 (or at least the most significant time robber )

Running the following script `python -X importtime example.py 2>/tmp/tuna.log && tuna /tmp/tuna.log`, showed most of the time was in `datashader.bundling`. This was jit-ified in https://github.com/holoviz/datashader/pull/1383, and therefore now takes three seconds to import. 

The geoviews imports setup hooks, which import `holoviews.operation.datashader` when switching to the Bokeh backend (by calling `hv.extension("bokeh")`). I didn't see any obvious way to remove this, and it will improve if something like https://github.com/holoviz/datashader/pull/1419 is merged. 

image

<details><summary>Script</summary>
<p>

``` python
from time import monotonic

import geoviews as gv

import holoviews as hv

start = monotonic()
hv.extension("bokeh")
print(monotonic() - start)
```

</p>
</details> 
 
FYI @amaloney

### DIFF
--- a/holoviews/operation/_datashader_bundling.py
+++ b/holoviews/operation/_datashader_bundling.py
@@ -1,0 +1,52 @@
+import param
+from datashader.bundling import (
+    directly_connect_edges as connect_edges,
+    hammer_bundle,
+)
+
+from ..core import Operation
+from .datashader import split_dataframe
+
+
+class _connect_edges(Operation):
+
+    split = param.Boolean(default=False, doc="""
+        Determines whether bundled edges will be split into individual edges
+        or concatenated with NaN separators.""")
+
+    def _bundle(self, position_df, edges_df):
+        raise NotImplementedError('_connect_edges is an abstract baseclass '
+                                  'and does not implement any actual bundling.')
+
+    def _process(self, element, key=None):
+        index = element.nodes.kdims[2].name
+        rename_edges = {d.name: v for d, v in zip(element.kdims[:2], ['source', 'target'], strict=None)}
+        rename_nodes = {d.name: v for d, v in zip(element.nodes.kdims[:2], ['x', 'y'], strict=None)}
+        position_df = element.nodes.redim(**rename_nodes).dframe([0, 1, 2]).set_index(index)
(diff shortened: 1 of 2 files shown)


holoviews/operation/_datashader_bundling.py
```diff
@@ -0,0 +1,52 @@
...
+        or concatenated with NaN separators.""")
+
+    def _bundle(self, position_df, edges_df):
+        raise NotImplementedError('_connect_edges is an abstract baseclass '
+                                  'and does not implement any actual bundling.')
+
+    def _process(self, element, key=None):
+        index = element.nodes.kdims[2].name
+        rename_edges = {d.name: v for d, v in zip(element.kdims[:2], ['source', 'target'], strict=None)}
+        rename_nodes = {d.name: v for d, v in zip(element.nodes.kdims[:2], ['x', 'y'], strict=None)}
```
amaloney: `strict=False` is default for `zip`
```suggestion
        rename_edges = {d.name: v for d, v in zip(element.kdims[:2], ['source', 'target'])}
        rename_nodes = {d.name: v for d, v in zip(element.nodes.kdims[:2], ['x', 'y'])}
```

amaloney (approved):
One suggested change

holoviews/operation/_datashader_bundling.py
hoxbro: This was just copied over. However, we have lint rules that enforce the use of strict, which always prompts us to consider whether we want to be strict or not. I have used `None` when I enabled the rule to show that no decision has been made for this.  By the look of it, the last one should be true, and the first one is likely also to be true.  

```suggestion
        rename_edges = {d.name: v for d, v in zip(element.kdims[:2], ['source', 'target'], strict=True)}
        rename_nodes = {d.name: v for d, v in zip(element.nodes.kdims[:2], ['x', 'y'], strict=True)}
```

TheoMathurin:
Many thanks for this improvement.

PR #6617, opened 2025-06-12
enh: Support passing selection dictionary to Dataset.select

philippjfr:
This PR proposes an API change (or rather an additional signature) for `Dataset.select`.

Currently we support passing per-dimensions select specifications as keyword arguments. This is generally quite convenient because in most cases dimensions are valid identifiers so the keyword syntax, e.g. `.select(x=(0, 10))` provides the shortest and most convenient syntax. However when the dimension name is not a valid identifier, e.g. it's a string digit or contains other non-valid identifiers you have to write it out using dictionary unpacking:

```python
x = hv.Dimension('0')

curve = hv.Curve([], kdims=[x])

curve.select(**{'0': (1, 3)})
```

This is not even particularly contrived because when you construct an element from a pandas DataFrame with default column names this is what happens. While a little cumbersome this use case at least works.

However, we are currently in the process of creating a new data interface for `anndata` and because of the complex data model we have to create special dimension objects, which are not easily mapped onto simple string names. This is where the current select approach completely breaks down since keyword arguments must be string based we cannot perform a select operation on elements backed by an `anndata` dataset, e.g.:

```python
x = hv.Dimension('0')

curve = hv.Curve([], kdims=[x])

curve.select(**{x: 1})
```

Will error because x is not a string. Therefore I propose we overload the `selection_expr` argument for `.select` making it possible to instead write select operations as:

```python
x = hv.Dimension('0')

curve = hv.Curve([], kdims=[x])

curve.select({x: 1})
```

This is fully backward compatible since it would previously just error. Before I do any more work on this I'd love to hear feedback.

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -611,10 +611,16 @@ class Dataset(Element, metaclass=PipelineMeta):
             from holoviews import dim
             ds.select(selection_expr=dim('x') % 2 == 0)
 
+        * selections dictionary: A dictionary of selections per dimension
+
+            ds.select({x: 3})
+
         Parameters
         ----------
-        selection_expr : holoviews.dim predicate expression
-            specifying selection.
+        selection_expr : A dim expression or dictionary of selections.
+            holoviews.dim predicate expression specifying selection or
+            a dictionary of selections (as an alternative to selecting
+            via keyword arguments).
         selection_specs : List of specs to match on
             A list of types, functions, or type[.group][.label]
             strings specifying which objects to apply the
@@ -629,6 +635,13 @@ class Dataset(Element, metaclass=PipelineMeta):
         or a scalar if a single value was selected
         """
         from ...util.transform import dim
+        if isinstance(selection_expr, dict):
+            if selection:
+                raise ValueError("""\
+                Selections may be supplied as keyword arguments or as a positional
+                argument, never both.""")
+            selection = selection_expr
+            selection_expr = None
         if selection_expr is not None and not isinstance(selection_expr, dim):
             raise ValueError("""\
             The first positional argument to the Dataset.select method is expected to be a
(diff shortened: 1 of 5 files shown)


hoxbro:
I don't see any problem with this approach.

philippjfr:
Found out that `Graph.select` and `Image.select` were not aligned with the `Dataset.select` signature.

hoxbro (approved):

holoviews/element/raster.py
```diff
@@ -415,6 +415,13 @@ def select(self, selection_specs=None, **selection):
         specs match the selected object.
 
         """
+        if isinstance(selection_expr, dict):
```
flying-sheep, line 418: This should be `Mapping`. Never check `isinstance(..., list)` or `isinstance(..., dict)` unless you know exactly why you want to treat these different from other `Sequence`s and `Mapping`s.

holoviews/core/data/__init__.py
```diff
@@ -629,6 +635,13 @@ def select(self, selection_expr=None, selection_specs=None, **selection):
         or a scalar if a single value was selected
         """
         from ...util.transform import dim
+        if isinstance(selection_expr, dict):
```
flying-sheep, line 638: ditto

flying-sheep:
This doesn't work at all when actually passing dimensions, see https://github.com/holoviz-topics/hv-anndata/blob/586e34beabeed853be172b9982e3efa7737834cd/tests/test_interface.py#L97-L121

1. when passing a `dict[Dimension, Any]`, like `select({A.obs["type"]: 0})`, it breaks here https://github.com/holoviz/holoviews/blob/9dc1d8b3a6812b9797f4d66bcf36e8616bee32a6/holoviews/core/data/__init__.py#L666

   ```
   TypeError: keywords must be strings
   ```

2. when passing strings that can be parsed as Dimension, it returns early here https://github.com/holoviz/holoviews/blob/9dc1d8b3a6812b9797f4d66bcf36e8616bee32a6/holoviews/core/data/__init__.py#L654-L657

philippjfr:
Thanks, rather than changing all the `Interface.select` implementations I'll probably just map the dimension to it's index inside `Dateset.select`.

Issue #6598, opened 2025-05-28
count_cat and by reductions not equivalent when applied to an NdOverlay


maximlt:
I'd expect a DynamicMap of ImageStack in both cases.

```python
import numpy as np
import pandas as pd
import holoviews as hv
import datashader as ds
from holoviews.operation.datashader import rasterize

hv.extension('bokeh')

def create_synthetic_dataset(x, y, s, val, cat):
    seed = np.random.default_rng(1)
    num = 10_000
    return pd.DataFrame(
        {"x": seed.normal(x, s, num), "y": seed.normal(y, s, num), "s": s, "val": val, "cat": cat}
    )

df = pd.concat(
    {
        cat: create_synthetic_dataset(x, y, s, val, cat)
        for x, y, s, val, cat in [
            (2, 2, 0.03, 0, "d1"),
            (2, -2, 0.10, 1, "d2"),
            (-2, -2, 0.50, 2, "d3"),
            (-2, 2, 1.00, 3, "d4"),
            (0, 0, 3.00, 4, "d5"),
        ]
    },
    ignore_index=True,
)

points = hv.Points(df, kdims=['x', 'y'], vdims=['cat']).groupby('cat').overlay()
print(points)
"""
:NdOverlay   [cat]
   :Points   [x,y]   (cat)
"""

r_by = rasterize(points, aggregator=ds.by('cat'))
r_count_cat = rasterize(points, aggregator=ds.count_cat('cat'))
r_by + r_count_cat
```

```python
print(r_by)
print(r_count_cat)

"""
:DynamicMap   []
   :ImageStack   [x,y]   (d1,d2,d3,d4,d5)
:DynamicMap   []
   :NdOverlay   [cat]
      :Image   [x,y]   (Count)
"""
```

<img width="624" alt="Image" src="https://github.com/user-attachments/assets/b1326e8c-7fcc-4fc1-94da-23e9f2f4e7c8" />

PR #6619, opened 2025-06-13
Skip overlay aggregate for NdOverlays with a count aggregator

maximlt:
Resolves https://github.com/holoviz/holoviews/issues/6598

By skipping `overlay_aggregate` when processing an NdOverlay and the aggregator is count_cat/by with count.

### DIFF
--- a/holoviews/tests/operation/test_datashader.py
+++ b/holoviews/tests/operation/test_datashader.py
@@ -274,18 +274,37 @@ class DatashaderAggregateTests(ComparisonTestCase):
         curve = Curve((dates, [1, 2, 3]))
         curve2 = Curve((dates, [3, 2, 1]))
         ndoverlay = NdOverlay({0: curve, 1: curve2}, 'Cat')
-        imgs = aggregate(ndoverlay, aggregator=ds.count_cat('Cat'), width=2, height=2,
+        img = aggregate(ndoverlay, aggregator=ds.count_cat('Cat'), width=2, height=2,
                          x_range=(xstart, xend), dynamic=False)
-        bounds = (np.datetime64('2015-12-31T23:59:59.723518'), 1.0,
-                  np.datetime64('2016-01-03T00:00:00.276482'), 3.0)
-        dates = [np.datetime64('2016-01-01T11:59:59.861759000',),
-                 np.datetime64('2016-01-02T12:00:00.138241000')]
-        expected = Image((dates, [1.5, 2.5], [[1, 0], [0, 2]]),
-                         datatype=['xarray'], bounds=bounds, vdims=Dimension('Count', nodata=0))
-        expected2 = Image((dates, [1.5, 2.5], [[0, 1], [1, 1]]),
-                         datatype=['xarray'], bounds=bounds, vdims=Dimension('Count', nodata=0))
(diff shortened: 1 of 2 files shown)


holoviews/tests/operation/test_datashader.py
```diff
@@ -274,18 +274,37 @@ def test_aggregate_ndoverlay_count_cat_datetimes_microsecond_timebase(self):
...
+            [0, 0],
+            [0, 0],
+            [1, 0]]
+        ], dtype=np.uint32)
+        xrds = xr.DataArray(
+            a,
+            dims=('y', 'x', 'Cat'),
+            coords={"y": y, "Cat": cat, "x": x}
+        )
+        expected = ImageStack(xrds, kdims=["x", "y"], vdims=["0.0", "1.0"])
```
maximlt, line 294: I tried to instantiate the ImageStack with `ImageStack(a, bounds=(xstart, 1.25, xend, 2.75), vdims=["Cat"])` but got this error, so I kept the test simpler. Not sure what I did wrong as the [docs](https://holoviews.org/reference/elements/bokeh/ImageStack.html) imply it should be possible.

```
ValueError: Supplied Image bounds do not match the coordinates defined in the data. Bounds only have to be declared if no coordinates are supplied, otherwise they must match the data. To change the displayed extents set the range on the x- and y-dimensions.
```

holoviews/tests/operation/test_datashader.py
```diff
@@ -274,18 +274,37 @@ def test_aggregate_ndoverlay_count_cat_datetimes_microsecond_timebase(self):
...
+        assert (expected.data == a).all()
```
maximlt, line 295: The fact that there's no registered `compare_x` testing function for ImageStack kept me busy for a little while, until I saw other ImageStack tests don't use `self.assertEqual()`.

philippjfr (commented):
I'm fine with this. If we're doing this we should also remove the if branch in `_process` that handles `count_cat`, since it's a) no longer used and b) was causing the different behaviors.

holoviews/tests/operation/test_datashader.py
hoxbro: Yes. At some point, I plan to switch from unittest to pytest. To avoid adding extra work in the future, I try to avoid using `self.assertEqual` for new tests.

holoviews/tests/operation/test_datashader.py
hoxbro: A bug. Will open an issue.

holoviews/tests/operation/test_datashader.py
maximlt: Ok get that. It's just it'd be nice to have a `compare_imagestack` function somewhere that can be used in these cases.

hoxbro: Just was hit by this myself... I wanted to look up this issue to write that we should add it. So thank you for the notification :100:

hoxbro: I also think I didn't completely understand your original point; that is on me.

Done in https://github.com/holoviz/holoviews/pull/6628/commits/09d4ece4e4ee4e11840760f726b35960bc7c03ed

Issue #6580, opened 2025-05-20
Miscategorizing the Bar data


droumis:
#### ALL software version info
<details>
    <summary>Software Version Info</summary>

```plaintext
Python              :  3.11.0 | packaged by conda-forge | (main, Jan 14 2023, 12:26:40) [Clang 14.0.6 ]
Operating system    :  macOS-15.4.1-arm64-arm-64bit
Panel comms         :  default

holoviews           :  1.20.2

bokeh               :  3.7.3
colorcet            :  3.1.0
cudf                :  -
cupy                :  -
dask                :  -
dask-expr           :  -
datashader          :  0.18.1
geoviews            :  -
hvplot              :  0.11.3
ibis-framework      :  -
IPython             :  9.2.0
ipywidgets-bokeh    :  -
jupyter-bokeh       :  -
jupyterlab          :  4.4.2
matplotlib          :  3.10.3
networkx            :  3.4.2
notebook            :  -
numba               :  0.61.2
numpy               :  2.2.6
pandas              :  2.2.3
panel               :  1.7.0
param               :  2.2.0
pillow              :  11.2.1
plotly              :  -
pyarrow             :  -
pyviz-comms         :  3.0.4
scikit-image        :  -
scipy               :  1.15.2
spatialpandas       :  -
streamz             :  -
tsdownsample        :  -
xarray              :  2025.4.0
```
</details>

#### Description of expected behavior and the observed behavior
Specifying a `color` dim when `cmap` is set results in miscategorizing the Bar data.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import pandas as pd
import holoviews as hv
from holoviews import opts
hv.extension('bokeh')

data = {
    'colA': ['A1', 'A1', 'A1', 'A1', 
             'A2', 'A2', 'A2', 'A2',
             'A3', 'A3', 'A3', 'A3'],
    'colB': ['cat1', 'cat2', 'cat3', 'cat4',
             'cat1', 'cat2', 'cat3', 'cat4',
             'cat1', 'cat2', 'cat3', 'cat4'],
    'count': [100, 50, 25, 10,
              80, 60, 30, 15,
              90, 45, 20, 5]
}

df = pd.DataFrame(data)

bar1 = hv.Bars(df, kdims=['colA', 'colB'], vdims=['count']).opts(
    height=400, 
    stacked=True, 
    xrotation=45, 
    legend_position='right', 
    cmap="Category10",
    title="Default behavior",
[...]

PR #6623, opened 2025-06-16
fix: Correctly apply Bars style transforms

philippjfr:
Fixes https://github.com/holoviz/holoviews/issues/6580

### DIFF
--- a/holoviews/tests/plotting/bokeh/test_barplot.py
+++ b/holoviews/tests/plotting/bokeh/test_barplot.py
@@ -1,5 +1,6 @@
 import numpy as np
 import pandas as pd
+import pytest
 from bokeh.models import (
     CategoricalColorMapper,
     DatetimeAxis,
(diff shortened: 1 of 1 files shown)


holoviews/plotting/bokeh/chart.py
```diff
@@ -866,6 +866,11 @@ def _axis_properties(self, axis, key, plot, dimension=None,
                 props['group_text_baseline'] = 'middle'
         return props
 
+    def _element_transform(self, transform, element, ranges):
+        if self.multi_level or self.stacked and len(element.kdims) > 1:
```
philippjfr:
```suggestion
        if (self.multi_level or self.stacked) and len(element.kdims) > 1:
```

holoviews/plotting/bokeh/chart.py
```diff
@@ -866,6 +866,11 @@ def _axis_properties(self, axis, key, plot, dimension=None,
                 props['group_text_baseline'] = 'middle'
         return props
 
+    def _element_transform(self, transform, element, ranges):
```
hoxbro, line 869: TIL about `_element_transform` :+1:

PR #6625, opened 2025-06-16
enh: Add `invert_dendrogram` option to dendrogram operation

hoxbro:
Setting `invert_dendrogram=True` mimics how ScanPy does it, but I don't think it should be the default. @droumis, can you play around with this? 

image

``` python
import holoviews as hv
import pandas as pd

hv.extension("bokeh")

data = [
    [7, 8, 0, 0, 0, 0],
    [8, 6, 0, 0, 0, 0],
    [0, 0, 9, 8, 2, 7],
    [0, 0, 0, 1, 8, 2],
    [0, 7, 0, 7, 0, 0],
]
df = pd.DataFrame([
    {"cluster": f"clust {i}", "gene": f"gene {j}", "value": data[i][j]}
    for j in range(6) 
    for i in range(5)
])

opts = dict(
    color=hv.dim("value").norm(), size=(hv.dim("value") != 0) * 30, cmap="Reds", tools=["hover"]
)
plot = hv.Points(df, kdims=["gene", "cluster"]).opts(**opts)

hv.operation.dendrogram(plot, adjoint_dims=["cluster"], main_dim="value")
hv.operation.dendrogram(plot, adjoint_dims=["cluster"], main_dim="value", invert_dendrogram=True)
```

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -1311,9 +1311,10 @@ class dendrogram(Operation):
             ic = ddata["icoord"]
             if self.p.invert_dendrogram:
                 ic = np.asarray(ic)
-                # Convert the smallest to the largest, while still being
-                # positive, offset (5) so we don't divide by zero
+                # Convert the smallest value to the largest value, while still
+                # being positive, offset (5) so we don't divide by zero
                 ic = ic.max() - ic + 5
+            # Important the kdims are unique
             dendros[d] = Dendrogram(ic, ddata["dcoord"], kdims=[f"__dendrogram_x_{d}", f"__dendrogram_y_{d}"])
 
         if not self.p.adjoined:
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -1274,7 +1274,8 @@ class dendrogram(Operation):
         https://docs.scipy.org/doc/scipy/reference/generated/scipy.cluster.hierarchy.linkage.html#scipy.cluster.hierarchy.linkage
         """
     )
-    invert_dendrogram = param.Boolean(default=False)
+    invert_dendrogram = param.Boolean(default=False, doc="""
+        Whether to invert the dendrogram axis.""")
 
     def _compute_linkage(self, dataset, dim, vdim):
         try:
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -1274,6 +1274,7 @@ class dendrogram(Operation):
         https://docs.scipy.org/doc/scipy/reference/generated/scipy.cluster.hierarchy.linkage.html#scipy.cluster.hierarchy.linkage
         """
     )
+    invert_dendrogram = param.Boolean(default=False)
 
     def _compute_linkage(self, dataset, dim, vdim):
         try:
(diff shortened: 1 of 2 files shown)


droumis:
I'm seeing an inversion of both axes:

image

<details> <summary> Versions </summary>

```bash
Python              :  3.11.0 | packaged by conda-forge | (main, Jan 14 2023, 12:26:40) [Clang 14.0.6 ]
Operating system    :  macOS-15.4.1-arm64-arm-64bit

holoviews           :  1.21.0b4.post1.dev1+g42f331095

bokeh               :  3.7.3
colorcet            :  3.1.0
cudf                :  -
cupy                :  -
dask                :  2025.5.1
dask-expr           :  -
datashader          :  0.18.1
geoviews            :  -
hvplot              :  0.11.3
ibis-framework      :  -
IPython             :  9.2.0
ipywidgets-bokeh    :  -
jupyter-bokeh       :  -
jupyterlab          :  4.4.2
matplotlib          :  3.10.3
networkx            :  3.4.2
notebook            :  -
numba               :  0.61.2
numpy               :  2.2.6
pandas              :  2.3.0
panel               :  1.7.1
param               :  2.2.1
pillow              :  11.2.1
plotly              :  -
pyarrow             :  20.0.0
pyviz-comms         :  3.0.5
scikit-image        :  -
scipy               :  1.15.3
spatialpandas       :  -
tsdownsample        :  -
xarray              :  2025.4.0
```

</details>

hoxbro:
I don't see this. Can you try from main and show me the main plot without any dendrogram on it for both this branch and main?

droumis:
## main branch
image

##  this branch
image

droumis:
Updating to latest python 3.11 patch doesn't resolve:

image

droumis:
Using python 3.13 

image

droumis:
might be `.reindex(element_kdims)` … can we just reindex the adjoint dim explicitly instead of all the dims?

hoxbro:
I can recreate the issue on a Mac and reduce it down to this MRE, showing it is a numpy sorting algorithm issue, as the default algorithm is quicksort, which is not stable [[ref](https://numpy.org/doc/stable/reference/generated/numpy.argsort.html#numpy.argsort)]  

``` python
import numpy as np

arr = np.tile(np.arange(5, 0, -1), 6)
np.argsort(arr)
```

image

image

droumis:
Nice! Looks good!

image

maximlt (commented):
Adding a `invert_dendogram` parameter to the `dendrogram` operation sounds redundant. Is this simply `invert_yaxis`?

hoxbro:
It is currently happening both for the top and bottom. It is possible that both dendrograms are inverted, and the main figures axes are following along accordingly. 

But I'm open to suggestions for better names.

hoxbro:
Do you prefer to have `invert_{x,y}axis` instead of? I'm also OK with that.

holoviews/operation/element.py
```diff
@@ -1298,16 +1300,23 @@ def _compute_linkage(self, dataset, dim, vdim):
...
             sort_dim = f"sort_{d}"
             dataset = dataset.add_dimension(sort_dim, 0, order)
             sort_dims.append(sort_dim)
 
+            ic = ddata["icoord"]
+            if self.p.invert_dendrogram:
+                ic = np.asarray(ic)
+                # Convert the smallest value to the largest value, while still
+                # being positive, offset (5) so we don't divide by zero
+                ic = ic.max() - ic + 5
```
droumis, line 1319:
```suggestion
                # Convert the smallest value to the largest value, while still
                # being positive
                INVERSION_OFFSET = 5  # Offset to ensure positive values
                ic = ic.max() - ic + INVERSION_OFFSET
```

droumis (changes requested):
Yea, I guess the current implementation applies the same inversion to all dendrograms, which is limiting.

Either having `invert_dendrogram` accept dims or going with `invert_xaxis`, `invert_yaxis` seems sufficient.

Otherwise, everything else looks good to me.

holoviews/operation/element.py
hoxbro: I think this makes the code harder to read, and even though 5 is a magic constant, it is only used in this one place, with a comment right above it.

hoxbro:
I have renamed the parameter to `invert`, which removes some redundancy.

> Yea, I guess the current implementation applies the same inversion to all dendrograms, which is limiting.

After giving some more thought, do we need more flexibility?  The only reason for adding it in the first place is to mimic the appearance of Scanpy's output; if we had matched up with Scanpy from the start, we would never have included the parameter. 

If there is demand for this flexibility in the future, we can expand the `invert` parameter as needed.

maximlt:
> I have renamed the parameter to invert, which removes some redundancy.

Thanks, I was mostly concerned about the redundancy, I leave it up to you both to decide the most appropriate behavior.

hoxbro:
I will merge this as I want to make an RC release.

PR #6627, opened 2025-06-17
chore: Update ruff to 0.12.0

hoxbro:
Most of them are pretty straightforward except PLW1641, which I did not know about.

### DIFF
--- a/holoviews/core/boundingregion.py
+++ b/holoviews/core/boundingregion.py
@@ -220,6 +220,9 @@ class BoundingBox(BoundingRegion):
         else:
             return False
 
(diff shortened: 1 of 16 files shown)


holoviews/core/boundingregion.py
```diff
@@ -220,6 +220,9 @@ def __eq__(self, other):
         else:
             return False
 
+    def __hash__(self):
+        return hash(self.aarect)
```
hoxbro, line 224: See https://docs.astral.sh/ruff/rules/eq-without-hash/#eq-without-hash-plw1641

Issue #6620, opened 2025-06-13
Errors with apply_when, rasterize, and ImageStack


maximlt:
I got two errors when running the script below (notebook and served) and zooming in.

The first one was raised in the web console and happens shortly after starting to zoom in. It seems like it happens when data from one category is no longer present in the viewport:
```
Error: Expected 5 not 4 colors in palette
    a https://cdn.bokeh.org/bokeh/release/bokeh-3.7.3.min.js:178
    c https://cdn.bokeh.org/bokeh/release/bokeh-3.7.3.min.js:178
    _v_compute_uint32 https://cdn.bokeh.org/bokeh/release/bokeh-3.7.3.min.js:455
    v_compute https://cdn.bokeh.org/bokeh/release/bokeh-3.7.3.min.js:383
    _flat_img_to_buf8 https://cdn.bokeh.org/bokeh/release/bokeh-3.7.3.min.js:559
...
```

The second one happens in Python, and later, I think at the moment when the `rasterize` operation is no longer applied by `apply_when`:
```
Traceback (most recent call last):
  File "/Users/mliquet/dev/holoviews/holoviews/plotting/util.py", line 305, in get_plot_frame
    return map_obj[key]
           ~~~~~~~^^^^^
  File "/Users/mliquet/dev/holoviews/holoviews/core/spaces.py", line 1302, in __getitem__
    val = self._execute_callback(*tuple_key)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mliquet/dev/holoviews/holoviews/core/spaces.py", line 1050, in _execute_callback
    retval = self.callback(*args, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/mliquet/dev/holoviews/holoviews/core/spaces.py", line 597, in __call__
    return self.callable()
           ^^^^^^^^^^^^^^^
  File "/Users/mliquet/dev/holoviews/holoviews/util/__init__.py", line 1071, in dynamic_operation
    key, obj = resolve(key, kwargs)
               ^^^^^^^^^^^^^^^^^^^^
  File "/Users/mliquet/dev/holoviews/holoviews/util/__init__.py", line 1060, in resolve
    return key, map_obj[key]
                ~~~~~~~^^^^^
  File "/Users/mliquet/dev/holoviews/holoviews/core/spaces.py", line 1302, in __getitem__
    val = self._execute_callback(*tuple_key)
[...]

PR #6628, opened 2025-06-18
fix: Wrong empty element when using apply_when

hoxbro:
Fixes some of the problems given in https://github.com/holoviz/holoviews/issues/6620

### DIFF
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -173,12 +173,9 @@ class AggregationOperation(ResampleOperation2D):
             params['xdensity'] = 1
         if height == 0:
             params['ydensity'] = 1
-        el = self.p.element_type(xarray, **params)
-        if isinstance(agg_fn, ds.count_cat):
-            vals = element.dimension_values(agg_fn.column, expanded=False)
-            dim = element.get_dimension(agg_fn.column)
-            return NdOverlay({v: el for v in vals}, dim)
-        return el
+        if isinstance(agg_fn, ds.by):
+            return ImageStack(xarray, **params)
+        return self.p.element_type(xarray, **params)
 
     def _get_agg_params(self, element, x, y, agg_fn, bounds):
         params = dict(get_param_values(element), kdims=[x, y],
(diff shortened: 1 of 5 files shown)


holoviews/operation/datashader.py
```diff
@@ -173,12 +173,9 @@ def _empty_agg(self, element, x, y, width, height, xs, ys, agg_fn, **params):
...
             params['ydensity'] = 1
-        el = self.p.element_type(xarray, **params)
-        if isinstance(agg_fn, ds.count_cat):
-            vals = element.dimension_values(agg_fn.column, expanded=False)
-            dim = element.get_dimension(agg_fn.column)
-            return NdOverlay({v: el for v in vals}, dim)
-        return el
+        if isinstance(agg_fn, ds.by):
+            return ImageStack(xarray, **params)
+        return self.p.element_type(xarray, **params)
```
hoxbro, line 178: Ideally, we should have a way to determine whether to use `ImageStack` or `Image` (the default) based on `self. p.element_type`. However, as we are nearing release, I would rather postpone it.

holoviews/operation/element.py
```diff
@@ -181,7 +181,7 @@ def _apply(self, element, x_range, y_range, invert=False):
         if x_range is not None and y_range is not None:
             selected = element[x_range, y_range]
         condition = self.predicate(selected)
-        if (not invert and condition) or (invert and not condition):
+        if invert ^ condition:
```
hoxbro, line 184: Not related, but just lovely to be able to use an XOR.

holoviews/plotting/bokeh/raster.py
```diff
@@ -466,6 +466,8 @@ def get_data(self, element, ranges, style):
             else element.dimension_values(vd, flat=False).transpose()
             for vd in element.vdims
         ])
+        if 0 in img.shape[:2]:  # Means we don't have any data
+            img = np.array([[[np.nan]]])
```
hoxbro, line 470: If we don't do this, the figure will not appear.

holoviews/operation/datashader.py
hoxbro: I will opened a PR with this in it: https://github.com/holoviz/holoviews/pull/6631

PR #6629, opened 2025-06-18
build: Add more project.urls

hoxbro:
Inspired by what [Litestar](https://pypi.org/project/litestar/) did

image
image


hoxbro:
Looks like this now:

image

Issue #6649, opened 2025-07-03
Plotly backend doesn't update downsample1d on zoom


pepijndevos:
#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc should be added within the dropdown below.)

<details>
    <summary>Software Version Info</summary>

```plaintext
annotated-types==0.7.0
anyio==4.9.0
anywidget==0.9.18
argon2-cffi==25.1.0
argon2-cffi-bindings==21.2.0
arrow==1.3.0
asttokens==3.0.0
async-lru==2.0.5
attrs==25.3.0
babel==2.17.0
beautifulsoup4==4.13.4
bleach==6.2.0
bokeh==3.7.3
bokeh_sampledata==2024.2
build==1.2.2.post1
certifi==2025.6.15
cffi==1.17.1
charset-normalizer==3.4.2
click==8.2.1
colorcet==3.1.0
comm==0.2.2
contourpy==1.3.2
cycler==0.12.1
datashader==0.18.1
debugpy==1.8.14
decorator==5.2.1
defusedxml==0.7.1
diskcache==5.6.3
docutils==0.21.2
executing==2.2.0
fastapi==0.115.12
fastjsonschema==2.21.1
fonttools==4.58.4
fqdn==1.5.1
h11==0.16.0
h5py==3.14.0
holoviews==1.21.0
httpcore==1.0.9
httptools==0.6.4
httpx==0.28.1
icalendar==6.3.1
idna==3.10
-e git+ssh://git@github.com/Innovoltive/InSpice.git@020f731cba1550419e825d0033351b32058bd01f#egg=InSpice
ipykernel==6.29.5
ipython==9.3.0
ipython_pygments_lexers==1.1.1
ipywidgets==8.1.7
isoduration==20.11.0
itsdangerous==2.2.0
jedi==0.19.2
Jinja2==3.1.6
json5==0.12.0
jsonpointer==3.0.0
jsonschema==4.24.0
jsonschema-specifications==2025.4.1
jupyter==1.1.1
jupyter-console==6.6.3
jupyter-events==0.12.0
jupyter-lsp==2.2.5
jupyter_bokeh==4.0.5
jupyter_client==8.6.3
jupyter_core==5.8.1
jupyter_server==2.16.0
jupyter_server_terminals==0.5.3
jupyterlab==4.4.4
jupyterlab_pygments==0.3.0
jupyterlab_server==2.27.3
[...]

PR #6650, opened 2025-07-03
enh: Sync tools with Bokeh and use auto_box_zoom as default

hoxbro:
Having auto_box_zoom as default is better, in my opinion, but if people disagree, they are welcome to revert. 

Code from: https://github.com/holoviz/holoviews/issues/6649

https://github.com/user-attachments/assets/4f1c1baa-56fd-4bd6-b227-73a57e1fffb9

Backward compatibility from 3.1.0 to the latest Bokeh release:

image

Checking if we can have auto box zoom default:

image

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
       - id: check-json
       - id: detect-private-key
   - repo: https://github.com/astral-sh/ruff-pre-commit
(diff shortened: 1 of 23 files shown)


philippjfr:
While this is a change in behavior I think it's completely uncontroversial. So I'm fine releasing this even in a patch release.

Issue #6660, opened 2025-08-10
Empty plot and numpy._core._exceptions._UFuncNoLoopError when plotting decimate with constant values (zero standard deviation)


PrateekArya:
#### ALL software version info
(this library, plus any other relevant software, e.g. bokeh, python, notebook, OS, browser, etc should be added within the dropdown below.)

<details>
    <summary>Software Version Info</summary>

```plaintext
Include version information here
python = 3.10.16
bokeh = 3.7.3
holoviews = 1.21.0
```
</details>

#### Description of expected behavior and the observed behavior

Empty plot and numpy._core._exceptions._UFuncNoLoopError (if 'x' is datetime) when plotting decimate with constant values (zero standard deviation)

#### Complete, minimal, self-contained example code that reproduces the issue

```python

import pandas as pd
import holoviews as hv
from holoviews.operation import decimate

data = pd.DataFrame({'x': pd.date_range(start='2025-08-10', periods=10, freq='D'), 'y': np.ones(10)}) # if 'x'  is defined as np.arange(10), numpy error will vanish, but the plot is still empty!

points = hv.Points(data, kdims=['x', 'y'])
decimate(points)

```

#### Screenshots or screencasts of the bug in action

- [ ] I may be interested in making a pull request to address this

PR #6661, opened 2025-08-11
fix: Don't select in decimate if start and end is the same

hoxbro:
Fixes https://github.com/holoviz/holoviews/issues/6660

When the start and end are the same, no data is selected, which causes the empty plot.

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -1010,8 +1010,12 @@ class decimate(Operation):
 
         # Slice element to current ranges
         xdim, ydim = element.dimensions(label=True)[0:2]
-        sliced = element.select(**{xdim: (xstart, xend),
-                                   ydim: (ystart, yend)})
+        select_dict = {}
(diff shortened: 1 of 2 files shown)


holoviews/operation/element.py
```diff
@@ -1010,8 +1010,12 @@ def _process_layer(self, element, key=None):
...
         # Slice element to current ranges
         xdim, ydim = element.dimensions(label=True)[0:2]
-        sliced = element.select(**{xdim: (xstart, xend),
-                                   ydim: (ystart, yend)})
+        select_dict = {}
+        if xstart != xend:
+            select_dict[xdim] = (xstart, xend)
+        if ystart != yend:
+            select_dict[ydim] = (ystart, yend)
+        sliced = element.select(**select_dict)
```
philippjfr:
```suggestion
        sliced = element.select(**select_dict) if select_dict else element
```

PR #6670, opened 2025-08-26
Allow discovering .apply operation on pipeline

philippjfr:
Otherwise operations applied with `.apply` aren't found.

### DIFF
--- a/holoviews/operation/element.py
+++ b/holoviews/operation/element.py
@@ -24,6 +24,7 @@ from ..core import (
     Operation,
     Overlay,
 )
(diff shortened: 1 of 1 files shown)


holoviews/operation/element.py
```diff
@@ -254,11 +255,20 @@ def find(self, operation, skip_nonlinked=True):
         """
         found = None
         for op in self.operations[::-1]:
+            if not op.link_inputs and skip_nonlinked:
```
maximlt, line 258: Made this change as the logic before didn't make sense to me.

PR #6673, opened 2025-08-27
enh: add simple from_sparse method for Graph

### LINKED ISSUE #6667: Add `hv.Graph.from_sparse`
#### Is your feature request related to a problem? Please describe.
We use `scipy.sparse.csr_array` (and `*_matrix`) for graph data storage. There is no fast way to use these data from holoviews.

#### Describe the solution you'd like
Ideally we’d be able to use a scipy sparse matrix’s nonzero elements as edges, and would be able to use the data in the matrix to map to some aesthetic (e.g. edge weight displayed as color)

#### Describe alternatives you've considered
- `hv.Graph.from_networkx` is really slow, as even if `g.nodes()` is just `range(len(g.nodes())` you need to pass a dict mapping node names to positions `dict[int, tuple[float, float]]` which will then be iterated over in Python
- converting the `csr_array` into a `coo_array` isn’t that much overhead, but also maybe inefficient. In any case it would be a nice enhancement.

#### Additional context
Would be great if `hv.Nodes` didn’t need a mandatory 3rd `kdims` dimension for this implicit case.

hoxbro:
resolves #6667

### DIFF
--- a/holoviews/element/graphs.py
+++ b/holoviews/element/graphs.py
@@ -480,14 +480,15 @@ class Graph(Dataset, Element2D):
 
     @classmethod
     def from_sparse(cls, edges, nodes, **params):
-        """Create a Graph element from a sparse adjacency array.
+        """Create a Graph element from a sparse adjacency array/matrix.
 
         Parameters
         ----------
-        edges : scipy.sparse array
-            A sparse array representing the graph adjacency array. The array
-            should have shape (n_nodes, n_nodes) where non-zero entries indicate
-            edges between nodes. The array will be converted to COO format.
+        edges : scipy.sparse array/matrix
+            A sparse array/matrix representing the graph adjacency
+            array/matrix. The array/matrix should have shape (n_nodes, n_nodes)
+            where non-zero entries indicate edges between nodes.
+            The array/matrix will be converted to COO format.
         nodes : array-like or Dataset
             Node positions or node information. Can be a 2D array of (x, y)
             coordinates, a Dataset with node information, or a Nodes element.
@@ -497,13 +498,13 @@ class Graph(Dataset, Element2D):
         Returns
         -------
         Graph
-            A Graph element with edges defined by the sparse array and the
+            A Graph element with edges defined by the sparse array/matrix and the
             provided node information.
 
         Raises
         ------
         TypeError
-            If edges is not a scipy sparse array.
+            If edges is not a scipy sparse array/matrix.
 
         Examples
         --------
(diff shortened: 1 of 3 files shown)


holoviews/tests/element/test_graphelement.py
```diff
@@ -128,6 +129,100 @@ def test_graph_redim_nodes(self):
         self.assertEqual(redimmed.nodes, graph.nodes.redim(x='x2', y='y2'))
         self.assertEqual(redimmed.edgepaths, graph.edgepaths.redim(x='x2', y='y2'))
 
+
+class TestFromSparse:
```
hoxbro, line 133: Tests are AI generated.

hoxbro:
@flying-sheep, This works for the "pure" sparse example - but did you also want it to support `adata` inputs?

holoviews/tests/element/test_graphelement.py
flying-sheep: My eyes start crossing when I try to read all that duplicated code. I wonder if things can be parametrized so the actual differences between the code blocks are easier to spot (by being defined in parameters instead of a mental code diff)

holoviews/element/graphs.py
```diff
@@ -478,6 +478,66 @@ def from_networkx(cls, G, positions, nodes=None, **kwargs):
         # Construct graph
         return cls((edge_data, nodes), vdims=edge_vdims)
 
+    @classmethod
+    def from_sparse(cls, edges, nodes, **params):
+        """Create a Graph element from a sparse adjacency array.
+
+        Parameters
+        ----------
+        edges : scipy.sparse array
```
flying-sheep: or matrix I assume?

flying-sheep:
> did you also want it to support `adata` inputs?

I think AnnData stuff should be confined to `hv-anndata` for now, this is great as it is!

holoviews/tests/element/test_graphelement.py
hoxbro: Thank you. Will clean this up.

holoviews/element/graphs.py
hoxbro: Yes. I wasn't sure about the convention in the sparse world, except that matrix is the "old" way and array is the "new" way.

flying-sheep: I think there is no universally agreed metaterm except for the clunky “sparse container”

holoviews/tests/element/test_graphelement.py
flying-sheep: I wouldn’t have said anything if I hadn’t been reminded recently how important it is to have terse tests: https://github.com/biomejs/biome/pull/6989#discussion_r2284632695

that test was broken, didn’t actually test what it was supposed to, but nobody ever spotted it since it was just a copy&paste job from another test.

PR #6674, opened 2025-08-27
fix: Silence Bokeh FIXED_SIZING_MODE warning on hv.save

### LINKED ISSUE #6668: Bokeh validation warning when saving a plot using hv.save with frame_width/frame_height options set

<details>

```plaintext
HoloViews = 1.21.1a0
Bokeh = 3.7.3
```
</details>

**Expected behavior**  
Saving a plot with `hv.save` should not trigger any warnings when `frame_height` or `frame_width` is specified.

**Actual behavior**  
When saving a plot with `hv.save`, specifying either `frame_height` or `frame_width` causes a Bokeh validation warning to appear, even though the plot is saved successfully.

```python
import holoviews as hv
hv.extension('bokeh')

plot = hv.Curve(([1,2,3,4,5], [2,4,6,8,10])).opts(frame_height=200) # frame_width also shows the warning

hv.save(plot, 'plot.png')

```

```
WARNING:bokeh.core.validation.check:W-1005 (FIXED_SIZING_MODE): 'fixed' sizing mode requires width and height to be set: figure(id='01850cd7-1c50-4293-9da6-f8c92809c23a', ...)
```

Azaya89:
closes #6668 

This PR ensures that saving plots with `frame_width` or `frame_height` set no longer triggers the Bokeh validation warning (FIXED_SIZING_MODE). Also:
- [x] Improved the context manager `silence_warnings` function that silences warnings only if they are not already silenced.
- [x] Added a test

### DIFF
--- a/holoviews/plotting/bokeh/styles.py
+++ b/holoviews/plotting/bokeh/styles.py
@@ -56,7 +56,6 @@ legend_dimensions = ['label_standoff', 'label_width', 'label_height', 'glyph_wid
 
 markers = {
     '+': {'marker': 'cross'},
-    'x': {'marker': 'cross', 'angle': np.pi/4},
     's': {'marker': 'square'},
     'd': {'marker': 'diamond'},
     '^': {'marker': 'triangle', 'angle': 0},
--- a/holoviews/tests/plotting/bokeh/test_pointplot.py
+++ b/holoviews/tests/plotting/bokeh/test_pointplot.py
@@ -3,6 +3,7 @@ import datetime as dt
 import numpy as np
 import pandas as pd
 import pytest
+from bokeh.core.enums import MarkerType
 from bokeh.models import (
     CategoricalColorMapper,
     Circle,
@@ -165,6 +166,17 @@ class TestPointPlot(TestBokehPlot):
         plot.initialize_plot()
         fig = plot.state
         self.assertEqual(len(fig.legend), 0)
+
+    def test_native_marker_legend(self):
+        """When one plots with a native bokeh marker, the legend uses that marker."""
+        for marker in MarkerType:
+            with self.subTest(marker=marker):
+                # Plot with a categorical color mapper solely to obtain a legend.
+                points = Points([(0, 0, "A"), (0, 1, "B")], vdims="color").opts(
+                    color="color", marker=marker
+                )
+                plot = bokeh_renderer.get_plot(points)
+                assert plot.state.legend[0].items[0].renderers[0].glyph.marker == marker
 
     def test_points_non_numeric_size_warning(self):
         data = (np.arange(10), np.arange(10), list(map(chr, range(94,104))))
--- a/holoviews/tests/plotting/bokeh/test_server.py
+++ b/holoviews/tests/plotting/bokeh/test_server.py
@@ -18,6 +18,7 @@ from holoviews.element.comparison import ComparisonTestCase
 from holoviews.plotting import Renderer
 from holoviews.plotting.bokeh.callbacks import Callback, RangeXYCallback, ResetCallback
 from holoviews.plotting.bokeh.renderer import BokehRenderer
+from holoviews.plotting.bokeh.util import BOKEH_GE_3_8_0
 from holoviews.streams import PlotReset, RangeXY, Stream
 
 bokeh_renderer = BokehRenderer.instance(mode='server')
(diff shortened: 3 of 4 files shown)


holoviews/util/__init__.py
```diff
@@ -841,8 +841,18 @@ def save(obj, filename, fmt='auto', backend=None, resources='cdn', toolbar=None,
...
+        try:
+            # Suppress only the specific Bokeh validator that warns when sizing_mode='fixed'
+            # but width/height are not both set.
+            from bokeh.core.validation.warnings import FIXED_SIZING_MODE
+
+            from ..plotting.bokeh.util import silence_warnings
+            with silence_warnings(FIXED_SIZING_MODE):
+                return renderer_obj.save(obj, filename, fmt=fmt, resources=resources, title=title)
+        except Exception as e:
+            param.main.param.warning(f"Could not silence Bokeh FIXED_SIZING_MODE: {e}")
```
hoxbro:
```suggestion
        # Suppress only the specific Bokeh validator that warns when sizing_mode='fixed'
        # but width/height are not both set.
        from bokeh.core.validation.warnings import FIXED_SIZING_MODE

        from ..plotting.bokeh.util import silence_warnings
        with silence_warnings(FIXED_SIZING_MODE):
            return renderer_obj.save(obj, filename, fmt=fmt, resources=resources, title=title)
```

This try/except is over-cautious.

holoviews/tests/plotting/bokeh/test_save.py
```diff
@@ -0,0 +1,28 @@
...
+    "opts",
+    [
+        {"frame_width": 200},
+        {"frame_height": 200},
+    ],
+)
+def test_save_suppresses_bokeh_fixed_sizing_mode(tmp_path, caplog, opts):
+    curve = hv.Curve([1, 2, 3]).opts(**opts, backend="bokeh")
+
+    out = tmp_path / "curve.html"
```
hoxbro, line 21: Can you double-check this actual fails without your code changes?

holoviews/tests/plotting/bokeh/test_save.py
```diff
@@ -0,0 +1,28 @@
+import pytest
+
+import holoviews as hv
+
+hv.extension("bokeh")
```
hoxbro:
```suggestion
```

You should not expose this at a top level in the tests as it can have side-effects. 

I think the reason you added it is that you had problems with the usefixtures. I will look into that.

holoviews/tests/plotting/bokeh/test_save.py
hoxbro: On my machine, it doesn't fail... I spent a significant amount of time trying to get it to fail on main and ended up with this.

``` python
def test_save_suppresses_bokeh_fixed_sizing_mode(tmp_path, caplog, opts):
    curve = hv.Curve([1, 2, 3]).opts(**opts)

    file = tmp_path / "curve.html"
    logging.getLogger("bokeh").propagate = True
    with caplog.at_level(logging.WARNING):
        hv.save(curve, file)

    assert file.exists()
    assert all(["FIXED_SIZING_MODE" not in e.msg for e in caplog.records])
```

hoxbro (changes requested):

holoviews/tests/plotting/bokeh/test_save.py
hoxbro: > I think the reason you added it is that you had problems with the usefixtures. I will look into that.

Should be solved with https://github.com/holoviz/holoviews/pull/6675

Azaya89: Thanks

holoviews/tests/plotting/bokeh/test_save.py
Azaya89: Thanks. I double checked and updated the file.

hoxbro: Did you have the same problem?

Azaya89: Yes. It wasn't failing on `main`, but the updated one fails on `main` now.

holoviews/util/__init__.py
```diff
@@ -841,8 +841,15 @@ def save(obj, filename, fmt='auto', backend=None, resources='cdn', toolbar=None,
             fmt = formats[-1]
         if formats[-1] in supported:
             filename = '.'.join(formats[:-1])
-    return renderer_obj.save(obj, filename, fmt=fmt, resources=resources,
-                             title=title)
+    if backend == "bokeh":
+        # Suppress only the specific Bokeh validator that warns when sizing_mode='fixed'
+        # but width/height are not both set.
```
hoxbro:
```suggestion
        # Suppress only the specific validator that warns when `sizing_mode='fixed'`
        # but width/height are not both set on a Bokeh Plot, this happens in HoloViews 
        # when `.opts(fixed_{width,height}=...)` are set,  which sets width/height to `None`. 
```

holoviews/tests/plotting/bokeh/test_save.py
```diff
@@ -0,0 +1,25 @@
...
+
+import holoviews as hv
+
+
+@pytest.mark.usefixtures("bokeh_backend")
+@pytest.mark.parametrize(
+    "opts",
+    [
+        {"frame_width": 200},
+        {"frame_height": 200},
```
hoxbro, line 13:
```suggestion
        {"frame_height": 200},
        {"frame_height": 200, "frame_width": 200},
        {},
```

Let's test every case.

PR #6676, opened 2025-08-27
fix: Use native bokeh "x" marker in plots and particularly in legends

### LINKED ISSUE #6345: x marker rendered as + in legend

```
holoviews==1.19.1
bokeh==3.5.1
```

When overlaying scatter plots with `x` markers, the marker is rendered as a `+` in the legend instead of a `x`.

```
import holoviews as hv

hv.extension('bokeh')

hv.Scatter([(1, 1)]).opts(marker='cross',size=10).relabel('cross') * hv.Scatter([(1,2)]).opts(marker='x', size=10).relabel('x')
```
 
image

stanwest:
Resolves #6345.

Previously, *Holoviews* converted the "x" into a rotated "+" as if it were a *matplotlib* marker that needed translation. However, because *Bokeh* draws legend markers with an angle of its choosing, the marker instead appeared as "+" within legends. Moreover, the translation appears to have been unnecessary, because *Bokeh* has had a native "x" marker for as far back as I could readily find.

With this change, *Holoviews* uses Bokeh's native "x" marker in plots and thereby obtains the expected "x" in legends also.

### DIFF
--- a/holoviews/tests/plotting/bokeh/test_pointplot.py
+++ b/holoviews/tests/plotting/bokeh/test_pointplot.py
@@ -184,7 +184,6 @@ class TestPointPlot(TestBokehPlot):
                 with self.subTest(marker=marker):
                     self._test_native_marker_legend(marker)
 
-    @pytest.mark.xfail(reason='Holoviews converts "x" marker to angled "cross".')
     def test_x_native_marker_legend(self):
         """When one plots with a native bokeh marker, the legend uses that marker."""
         self._test_native_marker_legend("x")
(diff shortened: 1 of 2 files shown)


holoviews/tests/plotting/bokeh/test_pointplot.py
```diff
@@ -166,6 +167,19 @@ def test_points_no_single_item_legend(self):
...
+        for marker in MarkerType:
+            with self.subTest(marker=marker):
+                # Plot with a categorical color mapper solely to obtain a legend.
+                points = Points([(0, 0, "A"), (0, 1, "B")], vdims="color").opts(
+                    color="color", marker=marker
+                )
+                plot = bokeh_renderer.get_plot(points)
+                self.assertEqual(
+                    plot.state.legend[0].items[0].renderers[0].glyph.marker, marker
+                )
```
hoxbro:
```suggestion
                assert plot.state.legend[0].items[0].renderers[0].glyph.marker == marker
```

stanwest:
You're welcome, and thanks for accepting it.

Issue #3835, opened 2019-07-19
Sankey node_color and node_fill_color options not applied


poplarShift:
```
import holoviews as hv
from holoviews import opts, dim
hv.extension('bokeh')
sankey = hv.Sankey([
       ['A', 'X', 5],
       ['A', 'Y', 7],
       ['A', 'Z', 6],
       ['B', 'X', 2],
       ['B', 'Y', 9],
       ['B', 'Z', 4]],
)
sankey.opts(
    node_color='red',
    node_fill_color='red',
)
```

Neither of those two options is applied.

mgsnuno:
@philippjfr  I was about to open this issue: if we use `color_index` explicitly then we can have some change in the colors, still we cannot set `node_color` based on a column of the data.

`color_index = 2` (default) in Sankey (compared to `None` in Graph). 
If `color_index = None` then we can use `node_color="red"`
but `node_color="certain_df_column"` it fails in bokeh with:

```
bokeh.core.validation.check - ERROR - E-1001 (BAD_COLUMN_NAME): Glyph refers to nonexistent column name. This could either be due to a misspelling or typo, or due to an expected column being missing. : key "fill_color" value "color" [renderer: GlyphRenderer(id='3238', ...)]
```

keepexploring:
I'm having the same issue, is there a solution to this?

philippjfr:
This issue has been mentioned on **HoloViz Discourse**. There might be relevant details there:

https://discourse.holoviz.org/t/node-color-parameter-does-not-work-with-sankey/8245/2

ahuang11:
For anyone interested in addressing this, I think it's around here:
https://github.com/holoviz/holoviews/blob/d776362f3b32054b8b1b447dd000a312141a8477/holoviews/plotting/bokeh/sankey.py#L95-L106

For reference:
https://github.com/holoviz/holoviews/blob/main/holoviews/plotting/bokeh/graphs.py#L209-L210

epaaso:
I just posted a PR with minimal internal code changes to fix this https://github.com/holoviz/holoviews/pull/6678.
With this fix, if you define a `Dataset` instead of a dataframe for the `nodes` and use `node_fill_color` instead of `node_color` in `bokeh`, you can get a correct color assignment of the nodes.

Here is a minimal script that correctly assigns color to the nodes:

```python
import pandas as pd
import holoviews as hv

# --- Choose backend ('matplotlib' or 'bokeh') ---
[...]

PR #6678, opened 2025-09-01
fix: handle 'node_color' in element transformations for Bokeh and MPL

epaaso:
Fixes Issue #3835.

To avoid messing with internal as much as possible, I changed the code such that to be able to use this fix with `mpl` you have to define the `nodes` object as a `Dataset` with the column for color as a `vdim`. Like in this snippet:

```python
nodes_ds = Dataset(nodes_df, kdims=['label'], vdims=['color'])
sk = hv.Sankey((links_df, nodes_ds), kdims=['source', 'target'], vdims=['value', 'color'])

sk = sk.opts(
        title='Minimal Sankey (matplotlib)',
        labels='label',
        edge_color='color',      # constant or map; here field -> edges (length = edges)
        node_color='color',      # taken from node vdims (length = nodes)
        fig_inches=500/150,
        aspect=350/500,
        bgcolor='white'
    )
```

The `bokeh` parameter `node_color` is for a fixed color, so one has to use `node_fill_color` instead:

```python
nodes_ds = Dataset(nodes_df, kdims=['label'], vdims=['color'])
sk = hv.Sankey((links_df, nodes_ds), kdims=['source', 'target'], vdims=['value', 'color'])
sk = sk.opts(
        title='Minimal Sankey (bokeh)',
        labels='label',
        edge_color='color',
        node_fill_color='color',
        width=500, height=350,
        bgcolor='white'
    )
```

There is an automatic assigning of `cdim` in both the pipeline for these backends and I do not prevent that. I just changed the code so that if `node_color` is defined, that parameter is given more importance. Maybe a more complete version of this PR would avoid auto-assigning `cdim` if `node_color` is given. But as `cdim` is deprecated, I decided to not waste much time looking for the auto-assign.

### DIFF
--- a/holoviews/tests/plotting/bokeh/test_graphplot.py
+++ b/holoviews/tests/plotting/bokeh/test_graphplot.py
@@ -3,7 +3,15 @@ from bokeh.models import EdgesAndLinkedNodes, NodesAndLinkedEdges, NodesOnly, Pa
 from bokeh.models.mappers import CategoricalColorMapper, LinearColorMapper
 
 from holoviews.core.data import Dataset
-from holoviews.element import Chord, Graph, Nodes, TriMesh, VLine, circular_layout, Sankey
+from holoviews.element import (
+    Chord,
+    Graph,
+    Nodes,
+    Sankey,
+    TriMesh,
+    VLine,
+    circular_layout,
+)
 from holoviews.plotting.bokeh.util import property_to_dict
 from holoviews.util.transform import dim
 


holoviews/plotting/mpl/element.py
```diff
@@ -653,6 +653,8 @@ def _apply_transforms(self, element, ranges, style):
             elif type(element) is Path:
                 val = np.concatenate([v.apply(el, ranges=ranges, flat=True)
                                       for el in element.split()])
+            elif 'node' in k:
+                val = v.apply(element.nodes)
```
philippjfr: I think we still want to provide the ranges here (and in the bokeh case).

holoviews/plotting/bokeh/element.py
```diff
@@ -1838,6 +1838,8 @@ def _apply_transforms(self, element, data, ranges, style, group=None):
                 ds = Dataset({d.name: v for d, v in self.overlay_dims.items()},
                              list(self.overlay_dims))
                 val = v.apply(ds, ranges=ranges, flat=True)[0]
+            elif 'node' in k:
```
philippjfr, line 1845: Ideally this would be implemented by implementing an overridden `_element_transform` method on GraphPlot. Annoyingly though we don't 
pass `k` through to that method, so for now I guess this is the only thing we can do.

philippjfr:
Thanks @epaaso, really appreciate the contribution. A simple test in `holoviews/tests/plotting/bokeh/test_graph.py` would also be greatly appreciated.

epaaso:
Hey, of course give me 1-2 days!

holoviews/plotting/mpl/element.py
RodrigoInvertox: Done!

holoviews/plotting/bokeh/element.py
RodrigoInvertox: oh i see, yeah that would be more general

holoviews/plotting/mpl/element.py
epaaso: Done!

holoviews/plotting/bokeh/element.py
epaaso: Oh yeah ,that would be more general.

epaaso:
Sorry for the delay is the test alright?

epaaso:
Apparently not, it ran in my pc with pixi, let me see why it doesnt now.

epaaso:
Thanks hoxbro

hoxbro:
Thank you for the fix :+1:

PR #6680, opened 2025-09-04
enh(bokeh): Optimize HeatMap rendering path if gridded and contiguous

philippjfr:
HeatMap is currently rendered using `rect` glyphs. This approach is flexible, since it allows the heatmap to be overlaid on arbitrary categorical axes, and each rectangle is drawn independently of the axis ordering. However, for large heatmaps with many categories, it becomes **extremely inefficient**: each rectangle (pixel) must be sent and rendered individually, which both increases data transfer and slows down rendering. To address this, we’ve added a new code path that is used when the HeatMap is already gridded and the data is contiguous with respect to the x- and y-axis factors.

### Example

```python
import numpy as np

arr = np.ones((10, 12)) * np.arange(12) * np.atleast_2d(np.arange(10)).T
hm_data = (list(map(str, range(12))), [chr(65+i) for i in range(10)], arr)

hv.HeatMap(hm_data) * hv.Points([('0', 'A'), ('11', 'E'), ('12', 'F')]).opts(color='red')
```

bokeh_plot - 2025-09-04T124540 963

### ToDo

- [x] Fix hover support
- [ ] Ensure invert_axes is handled correctly
- [ ] Ensure invert_xaxis and invert_yaxis are handled correctly
- [x] Add tests

### Observations

In implementing this I found a fairly undesirable behavior of our axis range/factor calculations. Specifically because we accumulate the ranges by element type the ordering of factors is not preserved correctly if there is a layer above and below that are of the same type.

As an example here we have `Points * HeatMap * Points`:

```python
hv.Points([('-1', '@')]) * hv.HeatMap(data) * hv.Points([('0', 'A'), ('11', 'E'), ('12', 'F')])
```

bokeh_plot - 2025-09-04T125023 924

because the factors of the Points are jointly calculated we end up with the x-axis order starting as `[-1, 0, 11, 12, ...]` rather than the `[-1, 0, 1, 2, ...]` ordering you'd expect if you calculated the element factors in sequence. In this case, this causes the heatmap to be non-contiguous, disabling the optimized codepath.

### DIFF
--- a/holoviews/plotting/bokeh/heatmap.py
+++ b/holoviews/plotting/bokeh/heatmap.py
@@ -56,6 +56,12 @@ class HeatMapPlot(ColorbarPlot):
         function, draw separation lines where function returns True for passed
         heatmap category.""")
 
+    optimize_gridded = param.Boolean(default=True, doc="""
+        Whether to optimize the heatmap rendering when the data is
+        gridded and contiguous. If enabled, a single image glyph will be
+        used to render the heatmap instead of individual quad glyphs,
+        significantly improving performance for large heatmaps.""")
+
 
     style_opts = ['cmap', 'color', 'dilate', *base_properties, *line_properties, *fill_properties]
 
@@ -66,8 +72,12 @@ class HeatMapPlot(ColorbarPlot):
         self._is_contiguous_gridded = False
 
     @property
+    def is_contiguous_gridded(self):
+        return self.optimize_gridded and self._is_contiguous_gridded
+
+    @property
     def _plot_methods(self):
-        return dict(single='image') if self._is_contiguous_gridded else dict(single='rect')
+        return dict(single='image') if self.is_contiguous_gridded else dict(single='rect')
 
     @classmethod
     def is_radial(cls, heatmap):
@@ -84,7 +94,7 @@ class HeatMapPlot(ColorbarPlot):
 
     def _update_hover(self, element):
         hover = self.handles["hover"]
-        if not self._is_contiguous_gridded or 'hv_created' not in hover.tags:
+        if not self.is_contiguous_gridded or 'hv_created' not in hover.tags:
             return super()._update_hover(element)
 
         source = self.handles["cds"]
@@ -136,7 +146,7 @@ class HeatMapPlot(ColorbarPlot):
 
         is_gridded = element.interface.gridded
         x_index = y_index = None
-        if is_gridded:
+        if is_gridded and self.optimize_gridded:
             x_range, y_range = self.handles['x_range'], self.handles['y_range']
             x_cat, y_cat = isinstance(x_range, FactorRange), isinstance(y_range, FactorRange)
             if x_cat:
@@ -155,7 +165,7 @@ class HeatMapPlot(ColorbarPlot):
                 y_index = ranges[y]['data'][0] if (not ydiff.size or np.allclose(ydiff, ydiff[0])) else None
 
         self._is_contiguous_gridded = is_gridded and x_index is not None and y_index is not None
-        if self._is_contiguous_gridded:
+        if self.is_contiguous_gridded:
             style = {k: v for k, v in style.items() if not k.startswith(('annular_', 'xmarks_', 'ymarks_'))}
             style['color_mapper'] = cmapper
             mapping = dict(image='image', x='x', y='y', dw='dw', dh='dh')


holoviews/core/util/__init__.py
```diff
@@ -1123,6 +1124,31 @@ def max_extents(extents, zrange=False):
...
+    b = np.asarray(b)
+
+    if len(a) == 0:
+        return 0
+    if len(a) > len(b):
+        return -1
+    W = sliding_window_view(b, len(a))
+
+    if nan_equal and a.dtype.kind == 'f':
+        eq = (W == a) | (np.isnan(W) & np.isnan(a))
```
hoxbro:
```suggestion
        eq = (W == a) |((W != W) & (a != a))
```

Alternative way. And seems to be a bit faster.

holoviews/core/util/__init__.py
```diff
@@ -22,6 +22,7 @@
 
 import numpy as np
 import param
+from numpy.lib.stride_tricks import sliding_window_view
```
hoxbro: The output of this can be quite large, but I'm not entirely sure what the expected input is.  

image

Just want to paste the note in here https://numpy.org/devdocs/reference/generated/numpy.lib.stride_tricks.sliding_window_view.html#numpy.lib.stride_tricks.sliding_window_view

image

holoviews/plotting/bokeh/heatmap.py
```diff
@@ -71,60 +79,105 @@ def get_data(self, element, ranges, style):
...
 
-        if self.static_source:
+        is_gridded = element.interface.gridded
+        x_index = y_index = None
+        if is_gridded:
+            x_range, y_range = self.handles['x_range'], self.handles['y_range']
+            x_cat, y_cat = isinstance(x_range, FactorRange), isinstance(y_range, FactorRange)
+            if x_cat:
+                xs = self._get_dimension_factors(element, ranges, element.get_dimension(x))
+                x_index = find_contiguous_subarray(xs, x_range.factors)
```
hoxbro: I think we can take advantage of the fact that the categories do not have any duplicates (or am I missing something?) Something like this (missing some of the checks)

``` python 
x = np.asarray([np.str_('A'), np.str_('B'), np.str_('C'), np.str_('D'), np.str_('E'), np.str_('F'), np.str_('G'), np.str_('H'), np.str_('I'), np.str_('J')])
y = np.asarray([ np.str_('E'), np.str_('F'), np.str_('G')])


def example(x, y):
    cmp = np.asarray(x) == y[0]
    
    if not cmp.any():
        return None
    
    idx = np.argmax(cmp)
    
    if not (x[idx:idx+len(y)] == y).all():
        return None
    
    return idx

example(x, y)

```

holoviews/core/util/__init__.py
```diff
@@ -1123,6 +1124,31 @@ def max_extents(extents, zrange=False):
...
+    Return the start index of `a` in `b` if `a` is a contiguous subarray of `b`.
+    Return -1 if not found.
+    """
+    a = np.asarray(a)
+    b = np.asarray(b)
+
+    if len(a) == 0:
+        return 0
+    if len(a) > len(b):
+        return -1
```
hoxbro: This offset the categorial axis for me.


image

``` python
from string import ascii_uppercase

import numpy as np

import holoviews as hv

hv.extension("bokeh")
arr = np.ones((10, 12)) * np.arange(12) * np.atleast_2d(np.arange(10)).T
arr = np.arange(120).reshape(-1, 12)
hm_data = (list(map(str, range(12))), [chr(65 + i) for i in range(10)], arr)
hm_data = (list(map(str, range(12))), [*ascii_uppercase[:10]], arr)
hm_data = (range(12), [*ascii_uppercase[:10]], arr)

p = hv.HeatMap(hm_data)
p.opts(tools=["hover"])
```

philippjfr: Yes, that should have been updated to None.

holoviews/plotting/bokeh/heatmap.py
philippjfr: True, that should work.

philippjfr:
>I can't think of a better way to get HoverTool to work than using CustomJS.

I can, will give that a shot.

holoviews/plotting/bokeh/heatmap.py
philippjfr: Okay, swapped it out.

philippjfr:
This ready to merge now?

hoxbro:
> > I can't think of a better way to get HoverTool to work than using CustomJS.
> 
> I can, will give that a shot.

The hover is not giving the same values as main:

``` python
import numpy as np
import holoviews as hv

hv.extension("bokeh")
arr = np.ones((10, 12)) * np.arange(12) * np.atleast_2d(np.arange(10)).T
hm_data = (list(map(str, range(12))), [chr(65+i) for i in range(10)], arr)

hv.HeatMap(hm_data).opts(tools=["hover"])
```

Main: 
Screenshot From 2025-10-20 13-36-32


This branch:
image

---

This example shows the code going down the wrong (new) codepath.

``` python
import numpy as np
import holoviews as hv

hv.extension("bokeh")
arr = np.ones((10, 12)) * np.arange(12) * np.atleast_2d(np.arange(10)).T

d = list(range(12))
d[-1] += 5
hm_data = (d, [chr(65+i) for i in range(10)], arr)
hv.HeatMap(hm_data).opts(tools=["hover"]) 
```

Main:
image

This branch:

image

---
When the above is fixed we should also verify that something like this works

``` python
import numpy as np
import panel as pn

import holoviews as hv

hv.extension("bokeh")
arr = np.ones((10, 12)) * np.arange(12) * np.atleast_2d(np.arange(10)).T


def func(value):
    d = np.asarray(list(range(12)))
    d[4:] += value
    hm_data = (d, [chr(65 + i) for i in range(10)], arr)
    return hv.HeatMap(hm_data).opts(tools=["hover"])


slider = pn.widgets.IntSlider(start=0, end=2, value=1)
pn.Row(slider, hv.DynamicMap(func, streams={"value": slider}))
```

hoxbro:
I think this should ready now @philippjfr 

> The hover is not giving the same values

This has now been solved by CustomJSHover in https://github.com/holoviz/holoviews/pull/6680/commits/c79679c6649b1ca3c0f74803218a792dd2b886f3

https://github.com/user-attachments/assets/b72fa6db-8588-4741-a8d8-b439790ad97f

> This example shows the code going down the wrong (new) codepath.

Solved by https://github.com/holoviz/holoviews/pull/6680/commits/aa9619c94fb2edda6cb2001bf810f938c9da41f1 (small fix in https://github.com/holoviz/holoviews/pull/6680/commits/f4d0993913433df40afbf748db846b5466578093)

> When the above is fixed we should also verify that something like this works

This likely requires a larger code change, so I think it is out of scope for this PR.

hoxbro:
The hover tool does not work great with a number axis, as it changes each integer and not per border. Can't really think of a way to fix this without being very hacky, will think a bit about it.

``` python
import numpy as np
from bokeh.io import show
from bokeh.models import ColumnDataSource, CustomJSHover, HoverTool

import holoviews as hv

hv.extension("bokeh")
arr = np.ones((10, 12)) * np.arange(12) * np.atleast_2d(np.arange(10)).T
hm_data = (list(map(int, range(0, 24, 2))), [chr(65 + i) for i in range(10)], arr)

hv.HeatMap(hm_data).opts(tools=["hover"])
```




https://github.com/user-attachments/assets/44d966f4-fc20-4bd7-b94e-ba0f2ecc3cee

EDIT: Wrong video

hoxbro:
Should be improved in https://github.com/holoviz/holoviews/pull/6680/commits/8450c8764f0a926ecd8a223b04b6aee9acfe5a2f, likely not ideal with floating numbers. 


https://github.com/user-attachments/assets/5000791c-f9cc-4a2e-9270-18ad435450d8

philippjfr:
Thanks for taking this over and addressing the remaining problems. Could you just quickly outline to me what the remaining issues are? What's the problem with floating numbers (and can't you just default to the default coordinate formatting in those cases)? What happens when switching between contiguous to non-contiguous data in a DynamicMap?

hoxbro:
> What's the problem with floating numbers (and can't you just default to the default coordinate formatting in those cases)?

Pushed a change to fix this (3d8c31c15). This is how it behaves now.


https://github.com/user-attachments/assets/122262b9-7296-4609-af71-b673a3186159



> What happens when switching between contiguous to non-contiguous data in a DynamicMap?

It goes blank when shifting between the different `_plot_methods`, [code as above](https://github.com/holoviz/holoviews/pull/6680#issuecomment-3421771801). 



https://github.com/user-attachments/assets/cd93e621-b779-4150-b41a-8d1be6d48578

hoxbro:
> we should add an option to disable the optimized path

Done in [1ddee1d](https://github.com/holoviz/holoviews/pull/6680/commits/1ddee1df5cd961cfcedbeb2da0f365b2519f106c)

PR #6697, opened 2025-09-18
Don't show tooltip if CI is detected

hoxbro:
Same as what is done in  https://github.com/holoviz/holoviews/pull/6630, but also looking at the CI environment variable, still to avoid it showing up when building docs.

https://www.scivision.dev/ci-detect-environment-variable/

We could also remove HV_DOCS_BUILD since it's not documented anywhere.

### DIFF
--- a/holoviews/ipython/__init__.py
+++ b/holoviews/ipython/__init__.py
@@ -260,7 +260,7 @@ class notebook_extension(extension):
             'bokeh_version':  bokeh_version,
             'mpl_version':    mpl_version,
             'plotly_version': plotly_version,
-            'show_tooltip': not os.getenv("HV_DOCS_BUILD")
+            'show_tooltip': not (os.getenv("CI") or os.getenv("HV_DOCS_BUILD"))
         })
         publish_display_data(data={'text/html': html})
 


maximlt (commented):
-1 on this one, some people may want to have the tooltip in their docs. Instead, I'd suggest:
- Adding a config parameter `show_tooltip_logo` (so one can do `hv.show_tooltip_logo = False`)
- This parameter can be controlled with the env var `HV_SHOW_TOOLTIP_LOGO`
- Documenting this.

hoxbro:
I want it disabled in the documentation builds by _default_. I don't want to see this in docs; the help message is for notebook users. I don't want anybody to have to look up a config or environment variable to hide this.

What about something like `hv.extension(show_tooltip_logo: bool | None=None)`? None will be based on the CI environment variable, while True or False will take priority if set.

maximlt:
> I don't want to see this in docs; the help message is for notebook users.

I think that's true for all the logos, not just the tooltip.

> What about something like `hv.extension(show_tooltip_logo: bool | None=None)`? None will be based on the CI environment variable, while True or False will take priority if set.

You still have to allow a user to display the tooltip if they want on their docs (why would HoloViews forbid that?). Would they have to set `CI=0`?

hoxbro:
> I think that's true for all the logos, not just the tooltip.

I think it is valuable to show the logos. After the first sentence (Extension loaded), I don't see any value in the tooltip; it only adds confusion.  

image

> You still have to allow a user to display the tooltip if they want on their docs

There would be `hv.extension(show_tooltip_logo=True)`. I don't think the tooltip asis is relevant for documentation, and shouldn't be enabled by default (for documentation).  

An alternative could be to only have the first sentence if we detect that we are running in Sphinx.

holoviews/ipython/__init__.py
```diff
@@ -260,7 +260,7 @@ def load_logo(cls, logo: dict | bool = False, bokeh_logo=False, mpl_logo=False,
             'bokeh_version':  bokeh_version,
             'mpl_version':    mpl_version,
             'plotly_version': plotly_version,
-            'show_tooltip': not os.getenv("HV_DOCS_BUILD")
+            'show_tooltip': not (os.getenv("CI") or os.getenv("HV_DOCS_BUILD"))
```
maximlt: How about something like that? I.e. `HV_HIDE_TOOLTIP_LOGO` takes precedence if set to true or false, then `CI`, and finally it's true by default. The idea being that if someone wants to show the tooltip in their docs they should be able to do it without having to touch every `hv.extension()` call.

```python
if os.getenv("HV_HIDE_TOOLTIP_LOGO") is not None:
    show_tooltip = os.getenv("HV_HIDE_TOOLTIP_LOGO", "").lower() in ("0", "false")
elif os.getenv("CI"):
    show_tooltip = False
else:
    show_tooltip=True
```

hoxbro: Should have applied your suggestion in [7c0032c](https://github.com/holoviz/holoviews/pull/6697/commits/7c0032cf1564feda5d90596e5232bdb1ca4ac95d)

PR #6700, opened 2025-09-29
enh: Add gridstyle to matplotlib

ahuang11:
In bokeh, it's possible to adjust the grid lines in a plot with `gridstyle`, but there was no equivalent for matplotlib.

```python
grid_style = {'grid_color': 'black', 'grid_linewidth': 1.5, 'xgrid_linestyle': 'dashed'}

hv.Points(np.random.rand(10, 2)).opts(gridstyle=grid_style, show_grid=True)
```

image

### DIFF
--- a/examples/user_guide/Plotting_with_Matplotlib.ipynb
+++ b/examples/user_guide/Plotting_with_Matplotlib.ipynb
@@ -144,14 +144,14 @@
    "source": [
     "## Grid lines\n",
     "\n",
-    "Grid lines can be controlled through the combination of ``show_grid`` and ``gridstyle`` parameters. The ``gridstyle`` allows specifying a number of options including:\n",
+    "Grid lines can be controlled through the combination of ``show_grid`` and ``gridstyle`` parameters. The ``gridstyle`` allows specifying a number of options including the following:\n",
     "\n",
-    "* ``grid_color``\n",
-    "* ``grid_alpha``\n",
-    "* ``grid_linestyle``\n",
-    "* ``grid_linewidth``\n",
+    "* ``color`` (or ``grid_color``)\n",
+    "* ``alpha`` (or ``grid_alpha``)\n",
+    "* ``linestyle`` (or ``grid_linestyle``)\n",
+    "* ``linewidth`` (or ``grid_linewidth``)\n",
     "\n",
-    "These options may also be applied to minor grid lines by prepending the ``'minor_'`` prefix and may be applied to a specific axis by replacing ``'grid_`` with ``'xgrid_'`` or ``'ygrid_'``. Here we combine some of these options to generate a complex grid pattern:"
+    "Here we combine some of these options to generate a complex grid pattern:"
    ]
   },
   {
@@ -161,6 +161,24 @@
    "outputs": [],
    "source": [
     "grid_style = {'grid_color': 'black', 'grid_linewidth': 1.5, 'xgrid_linestyle': 'dashed'}\n",
+    "\n",
+    "hv.Points(np.random.rand(10, 2)).opts(gridstyle=grid_style, show_grid=True)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "You can disable x or y grid lines individually by setting their alpha to 0."
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "grid_style = {'ygrid_alpha': 0}\n",
     "\n",
     "hv.Points(np.random.rand(10, 2)).opts(gridstyle=grid_style, show_grid=True)"
    ]
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -295,44 +295,55 @@ class ElementPlot(GenericElementPlot, MPLPlot):
         """Updates grid settings on the matplotlib axis based on gridstyle options.
 
         """
-        if not self.show_grid and not self.gridstyle:
+        if self.show_grid is False:
             axis.grid(False)
             return
 
-        style_items = list(self.gridstyle.items())
+        style_items = self.gridstyle.items()
         # Options that apply to both x and y grids, x, and then y
-        both = {k: v for k, v in style_items if k.startswith(('grid_', 'minor_grid'))}
+        both = {
+            k.removeprefix('grid_'): v
+            for k, v in style_items
+            if not k.startswith(('x', 'y'))
+        }
 
         # Merge options - both options first, then axis-specific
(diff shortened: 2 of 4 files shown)


holoviews/plotting/mpl/element.py
```diff
@@ -291,6 +291,49 @@ def _update_backend_opts(self):
...
+        # Apply grid visibility first
+        if xopts or yopts or self.show_grid:
+            axis.grid(True)
+
+        # Apply x-axis grid styling
+        if xopts:
+            xgridlines = axis.get_xgridlines()
+            for line in xgridlines:
+                for prop, val in xopts.items():
+                    if hasattr(line, f'set_{prop}'):
```
maximlt: Shouldn't this at least warn if the property is not found? Same comment for y handling.

holoviews/plotting/mpl/plot.py
```diff
@@ -110,6 +110,11 @@ class MPLPlot(DimensionedPlot):
     fig_size = param.Number(default=100., bounds=(1, None), doc="""
         Size relative to the supplied overall fig_inches in percent.""")
 
+    gridstyle = param.Dict(default={}, doc="""
+        Allows customizing the grid style, e.g. grid_line_color defines
```
maximlt: I don't think `set_line_color` is an existing Matplotlib method?

examples/user_guide/Plotting_with_Matplotlib.ipynb
```diff
@@ -138,6 +138,33 @@
...
+    "## Grid lines\n",
+    "\n",
+    "Grid lines can be controlled through the combination of ``show_grid`` and ``gridstyle`` parameters. The ``gridstyle`` allows specifying a number of options including:\n",
+    "\n",
+    "* ``grid_color``\n",
+    "* ``grid_alpha``\n",
+    "* ``grid_linestyle``\n",
+    "* ``grid_linewidth``\n",
+    "\n",
+    "These options may also be applied to minor grid lines by prepending the ``'minor_'`` prefix and may be applied to a specific axis by replacing ``'grid_`` with ``'xgrid_'`` or ``'ygrid_'``. Here we combine some of these options to generate a complex grid pattern:"
```
maximlt: Could you extend the example a little to also set minor grid lines? I feel it's a pretty common case for Matplotlib plots.

holoviews/plotting/mpl/element.py
```diff
@@ -291,6 +291,49 @@ def _update_backend_opts(self):
...
+                        getattr(line, f'set_{prop}')(val)
+
+        # Apply y-axis grid styling
+        if yopts:
+            ygridlines = axis.get_ygridlines()
+            for line in ygridlines:
+                for prop, val in yopts.items():
+                    if hasattr(line, f'set_{prop}'):
+                        getattr(line, f'set_{prop}')(val)
```
hoxbro:
```suggestion
                    if fn := getattr(line, f'set_{prop}', None):
                        fn(val)
```

Only a suggestion.

holoviews/plotting/mpl/element.py
```diff
@@ -291,6 +291,49 @@ def _update_backend_opts(self):
...
+        """Updates grid settings on the matplotlib axis based on gridstyle options.
+
+        """
+        if not self.show_grid and not self.gridstyle:
+            axis.grid(False)
+            return
+
+        style_items = list(self.gridstyle.items())
+        # Options that apply to both x and y grids, x, and then y
+        both = {k: v for k, v in style_items if k.startswith(('grid_', 'minor_grid'))}
```
hoxbro: Stupid question but why not `minor_grid_`?

holoviews/plotting/mpl/element.py
```diff
@@ -291,6 +291,49 @@ def _update_backend_opts(self):
...
+
+        # Merge options - both options first, then axis-specific
+        xopts = both.copy()
+        xopts.update({k.replace('xgrid_', ''): v for k, v in style_items if 'xgrid' in k})
+
+        yopts = both.copy()
+        yopts.update({k.replace('ygrid_', ''): v for k, v in style_items if 'ygrid' in k})
+
+        # Remove grid_ prefix from matplotlib property names
+        xopts = {k.replace('grid_', ''): v for k, v in xopts.items()}
```
hoxbro: This not only removes the prefix as the comment suggests. You can use `k.removeprefix("grid_")`. 

I also think this could be done before the `both.copy` to avoid doing it twice.

holoviews/plotting/mpl/element.py
```diff
@@ -291,6 +291,49 @@ def _update_backend_opts(self):
...
 
+    def _update_grid(self, axis):
+        """Updates grid settings on the matplotlib axis based on gridstyle options.
+
+        """
+        if not self.show_grid and not self.gridstyle:
+            axis.grid(False)
+            return
+
+        style_items = list(self.gridstyle.items())
```
hoxbro:
```suggestion
        style_items = self.gridstyle.items()
```

Not needed to convert to list as far as I can see.

holoviews/plotting/mpl/element.py
```diff
@@ -291,6 +291,49 @@ def _update_backend_opts(self):
...
+        yopts = both.copy()
+        yopts.update({k.replace('ygrid_', ''): v for k, v in style_items if 'ygrid' in k})
+
+        # Remove grid_ prefix from matplotlib property names
+        xopts = {k.replace('grid_', ''): v for k, v in xopts.items()}
+        yopts = {k.replace('grid_', ''): v for k, v in yopts.items()}
+
+        # Apply grid visibility first
+        if xopts or yopts or self.show_grid:
+            axis.grid(True)
```
hoxbro:
```suggestion
        axis.grid(True)
```

If I read the code correctly, I don't think this if statement is needed because of the first if statement in the function. It's possible that I don't fully understand the nuances of the code, though.

ahuang11:
Thanks both for the feedback.

> Is there a way around not having to prepend grid? gridstyle={"grid_color": "black", "grid_..."} feels redundant.

Okay, I now support prefixing (`gridstyle={"grid_color": "black"}`) and non-prefixing grid (`gridstyle={"color": "black"}`). I'm keeping both because the prefix is consistent with the bokeh backend.

image


> How can I only plot horizontal lines? I think it's because of axis.grid(True).

The way I do it is through `xgrid_alpha = 0`

> As a side note, I was very surprised to see this was not supported by HoloViews! It goes in my bucket of things that make HoloViews not a "real plotting library", making hvPlot's life harder.

To piggyback on this, I don't think there's a way to set minor ticks in HoloViews matplotlib backend AFAIK without hooks, thus:

> Could you extend the example a little to also set minor grid lines

I don't think this is doable yet; will save for another PR some day :)

> k.removeprefix

Cool, I wasn't aware of `removeprefix`.

PR #6704, opened 2025-10-08
compat: Python 3.14

hoxbro:
It appears that the only non-supported packages for HoloViews are numba and tsdownsample (at least for the lock). Let's see about the tests.

### DIFF
--- a/examples/conftest.py
+++ b/examples/conftest.py
@@ -62,6 +62,7 @@ if find_spec("datashader") is None:
         "reference/elements/plotly/ImageStack.ipynb",
         "user_guide/15-Large_Data.ipynb",
         "user_guide/16-Streaming_Data.ipynb",
+        "user_guide/Interactive_Hover_for_Big_Data.ipynb",
         "user_guide/Linked_Brushing.ipynb",
         "user_guide/Network_Graphs.ipynb",
     ]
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -84,23 +84,23 @@ jobs:
         run: |
           MATRIX=$(jq -nsc '{
               "os": ["ubuntu-latest", "macos-latest", "windows-latest"],
-              "environment": ["test-310", "test-313"]
+              "environment": ["test-310", "test-313", "test-314"]
```
hoxbro, line 87: When all packages can be installed by 3.14, we will remove test-313.

.github/workflows/test.yaml
```diff
@@ -84,23 +84,23 @@ jobs:
...
+              "environment": ["test-310", "test-311", "test-312", "test-313", "test-314"]
           }')
           echo "MATRIX=$MATRIX" >> $GITHUB_ENV
       - name: Set test matrix with 'downstream' option
         if: env.MATRIX_OPTION == 'downstream'
         run: |
           MATRIX=$(jq -nsc '{
               "os": ["ubuntu-latest"],
-              "environment": ["test-311"]
+              "environment": ["test-310"]
```
hoxbro, line 103: Changed this to one of the versions we also do by default.

pyproject.toml
```diff
@@ -139,6 +140,11 @@ filterwarnings = [
     "ignore:The legacy Dask DataFrame implementation is deprecated:FutureWarning", # https://github.com/holoviz/spatialpandas/issues/146
     # 2025-05
     "ignore:The 'shapely.geos' module is deprecated, and will be removed in a future version:DeprecationWarning", # https://github.com/geopandas/geopandas/pull/3453
+    # 2025-10
+    "ignore:`duckdb.typing` is deprecated and will be removed in a future version. Please use `duckdb.sqltypes` instead.", # https://github.com/narwhals-dev/narwhals/issues/3137
```
hoxbro, line 144: Not related to this PR

PR #6713, opened 2025-10-28
fix: add tools directly to Overlay

### LINKED ISSUE #6240: Adding tools to an Overlay has no effect
Adding tools to an Overlay via `opts` has no effect.

```python
import holoviews as hv
hv.extension('bokeh')

overlay = hv.Curve([-1, -2, -3]) * hv.Curve([1, 2, 3])
overlay.opts(tools=['zoom_in'])
```

image

To add tools to an Overlay, you need to add them to the elements it contains, one seems to be enough:

image

(Observed in https://github.com/holoviz/holoviews/pull/6190)

### LINKED ISSUE #6654: Overlay does not have separate zoom controls by axis

<details>

```
"panel>=1.7.2",
"holoviews>=1.21.0",
"bokeh>=3.7.3",
"pandas",
"numpy<2"

```
</details>

When plotting an overlay, the separate zoom controls (for X axis and Y axis) are not added to the toolbar, only the combined "Wheel Zoom" tool is added. The wheel zoom does still correctly zoom the X and Y axes separately if you hover over the axis while zooming.

The expected behaviour is for the X and Y zoom tools to be added separately to the toolbar if specified independently or together in `tools`. This is the behaviour seen when plotting an individual `Curve` or `Points` plot (for example).

```python
#!/usr/bin/env -S uv run
# /// script
# requires-python = ">=3.12"
# dependencies = [
#     "panel>=1.7.2",
#     "holoviews>=1.21.0",
#     "bokeh>=3.7.3",
#     "pandas",
#     "numpy<2"
# ]
# ///

import numpy as np
import pandas as pd
import holoviews as hv
import panel as pn

# Enable Panel/HoloViews
pn.extension()
hv.extension("bokeh")

# Generate random data
np.random.seed(42)
n_points = 100

# Data for curve plot
time_range = pd.date_range('2024-01-01', periods=n_points, freq='1min')
curve_data = pd.DataFrame({
    'time': time_range,
    'value': np.cumsum(np.random.randn(n_points)) + 100
})

# Data for points plot
points_data = pd.DataFrame({
    'time': np.random.choice(time_range, size=20),
    'value': np.random.uniform(95, 105, 20),
    'category': np.random.choice(['A', 'B', 'C'], 20)
})

# Data with different dimension names for decoupled plots
curve_data_separate = curve_data.rename(columns={'time': 'curve_time', 'value': 'curve_value'})
points_data_separate = points_data.rename(columns={'time': 'points_time', 'value': 'points_value'})

def create_plots():
    # Create curve plot for overlay
    curve = hv.Curve(
        curve_data,
        kdims=['time'],
        vdims=['value']
    ).opts(
        color='blue',
        line_width=2,
        tools=['xwheel_zoom', 'ywheel_zoom', 'hover'],
        title="Combined Curve + Points"
    )

    # Create points plot for overlay (using same dimensions for proper overlay)
    points = hv.Points(
        points_data,
        kdims=['time', 'value'],
        vdims=['category']
    ).opts(
        color='category',
        cmap={'A': 'red', 'B': 'green', 'C': 'orange'},
        size=10,
        tools=['xwheel_zoom', 'ywheel_zoom', 'hover']
    )

    # Create overlay with both plots
    overlay = hv.Overlay([curve, points])

    return overlay

def create_separate_curve():
    # Create standalone curve plot with different dimension names
    curve = hv.Curve(
        curve_data_separate,
        kdims=['curve_time'],
        vdims=['curve_value']
    ).opts(
        color='blue',
        line_width=2,
        tools=['xwheel_zoom', 'ywheel_zoom', 'hover'],
        title="Standalone Curve Plot"
    )
    return curve

def create_separate_points():
    # Create standalone points plot with different dimension names
    points = hv.Points(
        points_data_separate,
        kdims=['points_time', 'points_value'],
        vdims=['category']
    ).opts(
        color='category',
        cmap={'A': 'red', 'B': 'green', 'C': 'orange'},
        size=10,
        tools=['xwheel_zoom', 'ywheel_zoom', 'hover'],
        title="Standalone Points Plot"
    )
    return points

# Create the plots
overlay_pane = create_plots()
curve_pane = create_separate_curve()
points_pane = create_separate_points()

# Create Panel dashboard
dashboard = pn.Column(
    pn.pane.Markdown("## Minimal Reproduction: Curve + Points with Zoom Controls"),
    pn.pane.Markdown("### Combined Overlay (no separate zoom controls)"),
    pn.panel(overlay_pane, sizing_mode="stretch_both"),
    pn.pane.Markdown("### Separate Curve Plot (has separate zoom controls)"),
    pn.panel(curve_pane, sizing_mode="stretch_both"),
    pn.pane.Markdown("### Separate Points Plot (has separate zoom controls)"),
    pn.panel(points_pane, sizing_mode="stretch_both"),
    sizing_mode="stretch_both"
)

if __name__ == "__main__" or __name__.startswith("bokeh"):
    pn.serve(
        dashboard,
        title="Repro - Curve and Points",
        websocket_origin="*",
        show=True,
        port=5007
    )
```

N/A

Individual curve and points plots have both zoom tools independently on the toolbar (on the right hand side)
Image

An overlay of a curve and a points plot only has the combined wheel zoom tool on the right hand side:

Image

Azaya89:
fixes #6240 
fixes #6654

This PR fixes the problem where tools specified via `.opts(tools=[...])` on an Overlay were being ignored. The [OverlayPlot._init_tools()](https://github.com/holoviz/holoviews/blob/736589a78c058133fa9f3be1bbaded3f7de883af/holoviews/plotting/bokeh/element.py#L3232) method only collected tools from subplots but never processed tools from `self.default_tools` and `self.tools`.

Solution

- [x] Added logic to iterate through `self.default_tools + self.tools` and add them to the plot
- [x] Improved tool deduplication: Replaced simple type-based deduplication (`tool_types` list) with unique identifier-based deduplication (`tool_ids` set). This is better since most bokeh tools like [PanTool](https://github.com/bokeh/bokeh/blob/50c7e055c68794c935a8fc59b5a0567c5f11d3cf/src/bokeh/models/tools.py#L2134-L2136) have variants differentiated only by their dimensions (e.g., 'pan', 'xpan', 'ypan' are all of type PanTool)
- [x] Preserved special handling for hover tools and subcoordinate_y zoom tools which was causing the initial test failures
- [x] Added tests

```python
import holoviews as hv
hv.extension('bokeh')

overlay = (hv.Curve([1, 2, 3]).opts(tools=['xpan']) * hv.Curve([2, 3, 4]).opts(tools=['ypan', 'zoom_out'])).opts(tools=['zoom_in', 'xpan', 'ypan'])
overlay
```
Screenshot 2025-11-27 at 7 33 49 PM


Azaya89:
This works now but it seems a bit verbose to me. Will try to find a more elegant way to implement...

hoxbro:
Have pushed some changes. Left some todos and info with your name directly in the code. 

One thing I noticed is that the tool ordering seems weird; it would be nice to have more consistency. 

I also think we should check that this fixes  https://github.com/holoviz/holoviews/issues/6654.

holoviews/tests/plotting/bokeh/test_overlayplot.py
```diff
@@ -422,6 +424,138 @@ def test_ndoverlay_categorical_y_ranges(order):
...
+    assert tool_types == [*defaults, *new_tools]
+
+
+def test_overlay_opts_tools_with_element_tools():
+    overlay = Curve([1, 2, 3]).opts(tools=['zoom_out']) * Curve([2, 3, 4]).opts(tools=['zoom_in'])
+    plot = bokeh_renderer.get_plot(overlay)
+
+    tool_types = [type(tool).__name__ for tool in plot.state.tools]
+    defaults = ['WheelZoomTool', 'SaveTool', 'PanTool', 'BoxZoomTool', 'ResetTool']
+    # INFO(Azaya): Can this really be right?
```
Azaya89:
```suggestion
```
Hmm, I found this interesting...

The test passes because it validates the internal tool ordering (i.e, checking `plot.state.tools`) but visually when you actually plot the same code, the ordering is now different (`['PanTool', 'BoxZoomTool', 'WheelZoomTool', 'ZoomOutTool', 'ZoomInTool', 'SaveTool', 'ResetTool']`). Maybe this is a result of how Bokeh handles the visual arrangement of the tools?

Azaya89:
https://github.com/user-attachments/assets/2138d2c3-ecd3-473c-913b-952eebb848a7

holoviews/tests/plotting/bokeh/test_overlayplot.py
hoxbro: I don't know why this is the case at the top of my head. Try to see if you can understand by looking at the code and trying to use pure bokeh examples. Maybe also related to merge_tools.

holoviews/plotting/bokeh/util.py
```diff
@@ -1262,3 +1262,27 @@ def get_ticker_axis_props(ticker):
...
+        elif tool.startswith(('x', 'y')) and tool[1:] in directional_tools:
+            dimension = 'width' if tool.startswith('x') else 'height'
+            return tool_type, dimension
+        elif tool == 'auto_box_zoom':
+            return tool_type, 'auto'
+    else:
+        for name in ("dimensions", "tags", "name", "description", "icon"):
+            if identifier := getattr(tool, name, None):
+                # Convert lists/tuples to tuples (hashable)
+                if isinstance(identifier, list | tuple):
```
hoxbro:
```suggestion
                if isinstance(identifier, list):
```

Don't use pipe for isinstance and no need to check for tuple.

holoviews/tests/plotting/bokeh/test_overlayplot.py
Azaya89: I think it's definitely from bokeh
```python
from bokeh.plotting import figure, show
from bokeh.io import output_notebook

output_notebook()

x = [1, 2, 3, 4, 5]
y = [4, 5, 5, 7, 2]

default_tools = ['wheel_zoom', 'save', 'pan', 'box_zoom', 'reset']
extra_tools = ['zoom_out', 'zoom_in']

# create a plot
p = figure(
    title="Toolbars",
    tools=extra_tools+default_tools,
    width=400,
    height=300,
)

p.line(x, y)
show(p)
```
Screenshot 2025-12-02 at 9 43 14 AM

```python
p.toolbar.tools
```
```
[ZoomOutTool(id='p1475', ...),
 ZoomInTool(id='p1476', ...),
 WheelZoomTool(id='p1477', ...),
 SaveTool(id='p1478', ...),
 PanTool(id='p1479', ...),
 BoxZoomTool(id='p1480', ...),
 ResetTool(id='p1488', ...)]
```

So while `plot.toolbar.tools` respects the set order of the tools, visually it still follows the other way (i.e, `['PanTool', 'BoxZoomTool', 'WheelZoomTool', 'ZoomOutTool', 'ZoomInTool', 'SaveTool', 'ResetTool']`)

holoviews/plotting/bokeh/util.py
```diff
@@ -1262,3 +1262,27 @@ def get_ticker_axis_props(ticker):
...
+            return tool_type, dimension
+        elif tool == 'auto_box_zoom':
+            return tool_type, 'auto'
+    else:
+        for name in ("dimensions", "tags", "name", "description", "icon"):
+            if identifier := getattr(tool, name, None):
+                # Convert lists to tuples (hashable)
+                if isinstance(identifier, list):
+                    identifier = tuple(identifier)
+                return tool_type, identifier
```
hoxbro: What happens if the name is the same but the icon is not? They should likely not be identified as the same thing.

Azaya89:
> What happens if the name is the same but the icon is not? They should likely not be identified as the same thing.

I've added implementation to identify as different types if there are different disambiguation properties. This now means that something like this will produce multiple PanTool instances:
```python
import holoviews as hv
hv.extension('bokeh')

hv.Curve([1, 2, 3]).opts(tools=[PanTool(description='New Pan')])
```
Screenshot 2025-12-04 at 11 50 20 AM

holoviews/plotting/bokeh/util.py
```diff
@@ -1265,11 +1265,52 @@ def get_ticker_axis_props(ticker):
     return axis_props
 
 
+_DIRECTIONAL_TOOL_BASES = {
```
Azaya89, line 1268: Added more tools and tool aliases that were missed before.

holoviews/plotting/bokeh/util.py
```diff
@@ -1265,11 +1265,52 @@ def get_ticker_axis_props(ticker):
...
+
+_TOOL_ALIAS_IDS = {
+    'tap': 'tap',
+    'click': 'inspect',
+    'doubletap': 'doubletap',
+    'auto_box_zoom': 'auto',
+}
+
+
+def _get_tool_id_from_str(
```
Azaya89, line 1287: Moved the str check into a separate helper fxn

holoviews/plotting/bokeh/util.py
```diff
@@ -1265,11 +1265,52 @@ def get_ticker_axis_props(ticker):
...
+
+
 def get_tool_id(
     tool: str | tools.Tool,
     *,
-    properties: tuple[str, ...] = ("dimensions", "tags", "name", "description", "icon"),
-    skip_tags: set[str] | None = None
+    properties: tuple[str, ...] = (
+        'dimensions',
+        'dimension',
```
Azaya89, line 1307: Added 'dimension' for WheenPanTool specifically

Azaya89:
Also had GH Co-pilot help with adding more tests.

TheoMathurin:
I'm always struggling a bit, when thinking where I should declare my tools (and opts in general), if they're specific to an element of the overlay, or if they're common to elements of the same type, or of all elements of the overlay.

Issue #5526, opened 2022-11-28
ibis.common.exceptions.OperationNotDefinedError: No translation rule for <class 'ibis.expr.operations.generic.RowID'>


MarcSkovMadsen:
I'm trying to add a DuckDB user guide to hvPlot. When trying to create a `bar` plot using `hvplot` I get the error below. There is no issue with the matplotlib or bokeh backend.

```python
import duckdb
import ibis
import pandas as pd

from pathlib import Path
import hvplot.ibis

hvplot.extension("plotly")

DUCKDB_PATH = "DuckDB.db"

if not Path(DUCKDB_PATH).exists():
    pandas_df = pd.DataFrame(
        {
            "actual": [100, 150, 125, 140, 145, 135, 123],
            "forecast": [90, 160, 125, 150, 141, 141, 120],
            "numerical": [1.1, 1.9, 3.2, 3.8, 4.3, 5.0, 5.5],
            "date": pd.date_range("2022-01-03", "2022-01-09"),
            "string": ["Mon", "Tue", "Wed", "Thu", "Fri", "Sat", "Sun"],
        },
    )
    duckdb_con = duckdb.connect(DUCKDB_PATH)
    duckdb_con.execute("CREATE TABLE df AS SELECT * FROM pandas_df")

ibis.options.sql.default_limit = None

db = ibis.duckdb.connect(DUCKDB_PATH)

table = db.table("df")

plot = table.hvplot.bar(x="string", y="actual", line_width=2, alpha=0.7)

import panel as pn
pn.panel(plot).servable()
```

![image](https://user-images.githubusercontent.com/42288570/204207458-8577f3d4-096a-4081-b671-94143a99702f.png)

```OperationNotDefinedError: No translation rule for
Traceback (most recent call last):
  File "C:\repos\private\hvplot\.venv\lib\site-packages\bokeh\application\handlers\code_runner.py", line 231, in run
    exec(self._code, module.__dict__)
  File "C:\repos\private\hvplot\script2.py", line 34, in <module>
    pn.panel(plot).servable()
  File "C:\repos\private\hvplot\.venv\lib\site-packages\panel\viewable.py", line 379, in servable
    self.server_doc(title=title, location=location) # type: ignore
  File "C:\repos\private\hvplot\.venv\lib\site-packages\panel\viewable.py", line 871, in server_doc
    model = self.get_root(doc)
[...]

PR #6718, opened 2025-10-30
compat!: Deprecate IbisInterface

hoxbro:
This should now only come up when you explicitly set ibis as a datatype.

``` python
import ibis
import pandas as pd

import holoviews as hv

df = pd.DataFrame(
    {
        "name": ["Alice", "Bob", "Charlie"],
        "age": [25, 32, 37],
        "city": ["NYC", "LA", "Chicago"],
    }
)

t = ibis.memtable(df)
hv.Dataset(t)
```

image

### DIFF
--- a/holoviews/core/data/narwhals.py
+++ b/holoviews/core/data/narwhals.py
@@ -455,7 +455,7 @@ class NarwhalsInterface(Interface):
         is_lazy = isinstance(data, nw.LazyFrame)
         if not expanded:
             # LazyFrame does not support maintain_order, it can therefore for some
-            # backends return non-determistic results even for the same data.
+            # backends return non-deterministic results even for the same data.
             # It looks like polars.LazyFrame support it, but not duckdb
             data = data.unique(**({} if is_lazy else {"maintain_order": True}))
         if is_lazy:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,7 +17,7 @@ repos:
       - id: check-json
       - id: detect-private-key
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.0
+    rev: v0.14.3
     hooks:
       - id: ruff-check
   - repo: https://github.com/hoxbro/clean_notebook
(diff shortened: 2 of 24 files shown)


holoviews/tests/core/data/test_narwhalsinterface.py
```diff
@@ -146,6 +146,26 @@ def test_histogram(self):
         np.testing.assert_array_equal(edges, bins)
         assert np.all(hist >= 0)
 
+    def test_scalar_getitem(self):
```
hoxbro, line 149: Discovered this when trying to see if it fixed narwhals fixed: https://github.com/holoviz/holoviews/issues/5526

This tests what has been done in `NarwhalsInterface.select`

holoviews/core/data/narwhals.py
```diff
@@ -79,9 +78,10 @@ class NarwhalsInterface(Interface):
     def applies(cls, obj):
         return (
             isinstance(obj, (nw.Series, nw.DataFrame, nw.LazyFrame))
-            or is_into_dataframe(obj)
-            or is_into_series(obj)
-            or (cls.narwhals_backend and isinstance(obj, (dict, tuple, np.ndarray)))
+            or nw.dependencies.is_into_dataframe(obj)
+            or nw.dependencies.is_into_series(obj)
+            or nw.dependencies.is_ibis_table(obj)
```
hoxbro, line 83: Previously, this would not `apply` to Ibis tables.

holoviews/operation/element.py
```diff
@@ -865,7 +866,11 @@ def _process(self, element, key=None, groupby=False):
         elif isinstance(data, (nw.DataFrame, nw.LazyFrame, nw.Series)):
             if isinstance(data, nw.Series):
                 data = data.to_frame()
-            data = data.filter(nw.all().is_finite())
+            data = data.filter(~nw.all().is_null())
+            if not (data.implementation == nw.Implementation.IBIS and NARWHALS_VERSION <= (2, 10, 0)):
```
hoxbro: The fix has not yet been released, but is expected to be included in the next version.

hoxbro: Just been released. Not entirely  sure if we should keep this if statement.

hoxbro: I decided to remove it

PR #6732, opened 2025-11-13
enh: Add numpy masked array to  `masked_types`

hoxbro:
Surprised that `masked_types` didn't have numpy in it. And could see it was only used in one place in our codebase. 

Also adding a test for it.


hoxbro:
Have tested with the earliest supported versions:

image

Screenshot From 2025-11-13 15-16-47

PR #6744, opened 2025-11-25
fix: Suppress RunTime warnings for negative size values

### LINKED ISSUE #6742: RuntimeWarning with negative size values in matplotlib backend
There is a warning in the MPL backend which results from passing negative values to the `size` or `s` option. This warning does not arise in the bokeh backend (the warning is suppressed I guess).

_Originally posted by @Azaya89 in https://github.com/holoviz/hvplot/issues/1691#issuecomment-3571628713_

**MRE**
```python
import numpy as np
import pandas as pd
import holoviews as hv

hv.extension("bokeh", "matplotlib")

x = np.linspace(0.0, 10.0, 20)
y = list(range(-10, 10))
df = pd.DataFrame({'x': x, 'y': y})

# Bokeh plot - no warning
hv.Scatter(df, ['x'], ['y']).opts(size='y')
```

Image

```python
hv.output(backend="matplotlib")

# matplotlib plot - shows warning
hv.Scatter(df, ['x'], ['y']).opts(s='y')
```

Image

```
RuntimeWarning: invalid value encountered in sqrt
  scale = np.sqrt(self._sizes) * dpi / 72.0 * self._factor
```
This warning doesn't clearly indicate to users that the issue is with their negative size values, and for both backends, the negative values are filtered out when rendering the plot.

I'm proposing that we add a validation check that:
1. Detects when negative values are present in size-related options
2. Issues a clear, informative warning like: "Warning: Negative values detected in size dimension. Negative sizes may not render correctly."
3. Applies consistently across all backends before backend-specific rendering

I believe that this would provide better user experience and make the issue immediately clear to users

Azaya89:
closes #6742

Had to suppress the warning globally as this occurred in matplotlib's rendering code.

image

### DIFF
--- a/holoviews/plotting/mpl/renderer.py
+++ b/holoviews/plotting/mpl/renderer.py
@@ -164,9 +164,9 @@ class MPLRenderer(Renderer):
             )
             kw.update(kwargs)
 
-            # Attempts to precompute the tight bounding box
             with np.errstate(invalid="ignore"):
                 with suppress(Exception):
+                    # Attempts to precompute the tight bounding box
                     kw = self._compute_bbox(fig, kw)
                 bytes_io = BytesIO()
                 fig.canvas.print_figure(bytes_io, **kw)


hoxbro:
We should never globally suppress a warning! An example of how this can be done is shown in this diff:


``` diff
diff --git a/holoviews/plotting/mpl/element.py b/holoviews/plotting/mpl/element.py
index 1eee86fad..63262921b 100644
--- a/holoviews/plotting/mpl/element.py
+++ b/holoviews/plotting/mpl/element.py
@@ -663,6 +663,7 @@ class ElementPlot(GenericElementPlot, MPLPlot):
             # scatter have a default cmap and with an empty array will emit this warning
             warnings.filterwarnings('ignore', "No data for colormapping provided via 'c'")
+            warnings.filterwarnings('ignore', 'invalid value encountered in sqrt')
             artist = plot_fn(*plot_args, **plot_kwargs)
         return {'artist': artist[0] if isinstance(artist, list) and
                 len(artist) == 1 else artist}
diff --git a/holoviews/plotting/mpl/renderer.py b/holoviews/plotting/mpl/renderer.py
index fdef79d63..d7c72a073 100644
--- a/holoviews/plotting/mpl/renderer.py
+++ b/holoviews/plotting/mpl/renderer.py
@@ -169,7 +169,10 @@ class MPLRenderer(Renderer):
             except Exception:
                 pass
             bytes_io = BytesIO()
-            fig.canvas.print_figure(bytes_io, **kw)
+            import warnings
+            with warnings.catch_warnings():
+                warnings.filterwarnings('ignore', 'invalid value encountered in sqrt')
+                fig.canvas.print_figure(bytes_io, **kw)
             data = bytes_io.getvalue()
 
         if as_script:
```

Azaya89:
I already tried this approach but it didn't work. The warnings were still there.

hoxbro:
You should be able to suppress the warning only when we are calling it ourselves. We shouldn't manipulate global settings.

I seem to have gotten it to work with a little more work, around my original diff. 

image

holoviews/plotting/mpl/renderer.py
```diff
@@ -164,12 +165,11 @@ def _figure_data(self, plot, fmt, bbox_inches='tight', as_script=False, **kwargs
             kw.update(kwargs)
 
             # Attempts to precompute the tight bounding box
```
hoxbro: Let's move this down. to the relevant kw.

holoviews/plotting/mpl/element.py
```diff
@@ -660,7 +660,7 @@ def init_artists(self, ax, plot_args, plot_kwargs):
         if 'norm' in plot_kwargs: # vmin/vmax should now be exclusively in norm
             plot_kwargs.pop('vmin', None)
             plot_kwargs.pop('vmax', None)
-        with warnings.catch_warnings():
+        with (warnings.catch_warnings(), np.errstate(invalid="ignore")):
```
hoxbro, line 663: Check if this syntax also works on Python 3.10.

Azaya89: It works, thanks.

https://docs.python.org/3/whatsnew/3.10.html#parenthesized-context-managers

PR #6746, opened 2025-11-27
enh: Add support for `text_outline_width`

hoxbro:
Just merged in https://github.com/bokeh/bokeh/pull/14733 

Opening now, so we don't forget about it like we did with https://github.com/holoviz/holoviews/pull/6738


jbednar:
Nice! Was going here to open an issue about it, and I see you've already got the PR!

PR #6749, opened 2025-12-04
compat: Pandas 3.0.0

hoxbro:
Core test updates have been related to `StringDtype`. 

Some changes were also needed with respect to Copy-on-Write (CoW)

Also decided to remove the global configuration in our test dask setting `convert-string` to False.

### DIFF
--- a/holoviews/core/data/spatialpandas_dask.py
+++ b/holoviews/core/data/spatialpandas_dask.py
@@ -34,12 +34,17 @@ class DaskSpatialPandasInterface(SpatialPandasInterface):
 
     @classmethod
     def init(cls, eltype, data, kdims, vdims):
+        import dask
         import dask.dataframe as dd
         data, dims, params = super().init(
             eltype, data, kdims, vdims
         )
         if not isinstance(data, cls.frame_type()):
-            data = dd.from_pandas(data, npartitions=1)
+            # convert-string will convert the object dtype to string
+            # even if the object is a list
+            # https://github.com/dask/dask/issues/10631
+            with dask.config.set({"dataframe.convert-string": False}):
+                data = dd.from_pandas(data, npartitions=1)
         return data, dims, params
 
     @classmethod
--- a/holoviews/tests/conftest.py
+++ b/holoviews/tests/conftest.py
@@ -54,13 +54,6 @@ with contextlib.suppress(ImportError):
     mpl.use("agg")
 
 
-with contextlib.suppress(Exception):
-    # From Dask 2023.7.1 they now automatically convert strings
-    # https://docs.dask.org/en/stable/changelog.html#v2023-7-1
-    import dask
-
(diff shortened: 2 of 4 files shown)


holoviews/tests/core/data/test_narwhalsinterface.py
```diff
@@ -220,6 +221,9 @@ def frame(self, *args, **kwargs):
 
         return dd.from_pandas(pd.DataFrame(*args, **kwargs), npartitions=2)
 
+    @pytest.mark.xfail(PANDAS_GE_3_0_0, reason="pandas 3.0")
+    def test_dataset_nodata_range(self):
+        return super().test_dataset_nodata_range()
```
hoxbro: This can reproduce the problem; it seems like the `nw.when` expression converts the column to a string type. Not exactly sure where the problem lies. 

``` python
import dask
import dask.dataframe as dd
import narwhals as nw
import pandas as pd

print(pd.__version__)

# dask.config.set({"dataframe.convert-string": False})  # makes min fail

name = "Weight"
df_native = dd.from_pandas(pd.DataFrame({name: [15, 18, 10]}), npartitions=2)
df_column = nw.from_native(df_native).select(name)

expr = nw.when(nw.col(name) == 10).then(None).otherwise(nw.col(name)).alias(name)
df_column = df_column.select(expr)

df_column.select(cmin=nw.col(name).min())
print(df_column.collect_schema())
```

hoxbro: FYI: @MarcoGorelli and @FBruzzesi

MarcoGorelli: thanks for the ping - i'd suggest inverting the conditions:
```py
expr = nw.when(nw.col(name) != 10).then(nw.col(name))
```
then you don't even need the alias

hoxbro: Thank you - that did the trick.

holoviews/core/data/spatialpandas_dask.py
```diff
@@ -34,12 +34,17 @@ def frame_type(cls):
...
         data, dims, params = super().init(
             eltype, data, kdims, vdims
         )
         if not isinstance(data, cls.frame_type()):
-            data = dd.from_pandas(data, npartitions=1)
+            # convert-string will convert the object dtype to string
+            # even if the object is a list
+            # https://github.com/dask/dask/issues/10631
+            with dask.config.set({"dataframe.convert-string": False}):
+                data = dd.from_pandas(data, npartitions=1)
```
hoxbro, line 47: image

``` python
import pandas as pd
import dask
import dask.dataframe as dd

with dask.config.set({"dataframe.convert-string": False}):
    ddf = dd.from_pandas(pd.DataFrame([{'value': [5, 5, 5]}]))

```

PR #6756, opened 2025-12-11
docs: Update roadmap

hoxbro:


maximlt:
I'm not going to stamp this PR with a review as I believe this should preferably be done by the maintainers of a project.

I got great feedback from Philipp on the hvPlot roadmap PR update (https://github.com/holoviz/hvplot/pull/1694#pullrequestreview-3551463554), which I think would also apply to HoloViews roadmap as suggested.

philippjfr (commented):
Agree with @maximlt that it might be nice to align the HoloViews and hvPlot roadmaps. I think HoloViews could make a bigger deal about trimming the fat, i.e. deprecating some features, otherwise I think it makes sense to state some of the overall principles around focusing on improved testing, robustness and consistency over new features.

The only other thing I'm missing here, and though I'm unclear whether we actually have the time and energy to invest in this, the plan was also to move away from the user guides and align the documentation with the other libraries.

PR #6757, opened 2025-12-11
fix: Handle datetime with timezone in rasterize

### LINKED ISSUE #5127: Timezone-aware data is plotted incorrectly when using `rasterize` (plotted as though axis were UTC)
```python
import sys
import holoviews as hv
import pandas as pd
import datashader
import bokeh
print(f"{sys.version=}\n{hv.__version__=}\n{pd.__version__=}\n{datashader.__version__=}\n{bokeh.__version__=}")
```
```
sys.version='3.8.10 | packaged by conda-forge | (default, May 10 2021, 22:58:09) \n[Clang 11.1.0 ]'
hv.__version__='1.14.6'
pd.__version__='1.2.5'
datashader.__version__='0.13.0'
bokeh.__version__='2.4.0'
```

When plotting using `rasterize`, timezone-aware data is plotted incorrectly on the axes. (The plot incorrectly treats the x-axis as though it were UTC-local, and the plot appears blank when initially rendered.)

Somewhat related to this old bug https://github.com/holoviz/holoviews/issues/4526, but specific to `rasterize`.

When not using rasterize:
```python
import holoviews as hv
import pandas as pd
t = pd.date_range(start='2020-01-01 10:00', end='2020-01-01 12:00', freq='H', tz='Asia/Shanghai')
d = pd.DataFrame({'t': t, 'y': [1, 2, 3]})
curve = hv.Curve(d, kdims=['t'], vdims=['y'])
curve
```
it produces this correct graph as expected (local time on the x-axis):

When using `rasterize`, it still shows the x-axis as 10:00 to 12:00 (same as above). **However**, the data is incorrectly plotted as though the x-axis were UTC time (even though in this case it's actually Asia/Shanghai), and so the chart appears blank at first:

```python
rasterize(curve)
```

If we zoom out the x-axis, we find that the line was incorrectly drawn at 02:00-04:00 (rather than 10-12), which is the UTC-converted version of the timestamps (and not the correct Asia/Shanghai time):

Thank you! Really appreciate the great tools you've all built with holoviews, datashader, etc.

hoxbro:
Fixes #5127

``` python
import holoviews as hv
import pandas as pd
hv.extension("bokeh")
t = pd.date_range(start='2020-01-01 10:00', end='2020-01-01 12:00', freq='h', tz='Asia/Shanghai')
d = pd.DataFrame({'t': t, 'y': [1, 2, 3]})
curve = hv.Curve(d, kdims=['t'], vdims=['y'])
curve + rasterize(curve, line_width=1)
```

image

### DIFF
--- a/holoviews/tests/operation/test_datashader.py
+++ b/holoviews/tests/operation/test_datashader.py
@@ -1369,8 +1369,7 @@ class DatashaderRasterizeTests(ComparisonTestCase):
         t = pd.date_range(start='2020-01-01 10:00', end='2020-01-01 12:00', freq='h', tz='Asia/Shanghai')
         d = pd.DataFrame({'t': t, 'y': [1, 2, 3]})
         curve = Curve(d, kdims=['t'], vdims=['y'])
-
-        img = rasterize(curve, line_width=1, dynamic=False)
+        img = rasterize(curve, dynamic=False)
         assert isinstance(img, Image)
 
 


holoviews/tests/operation/test_datashader.py
```diff
@@ -1364,6 +1364,15 @@ def test_rasterize_path_empty_string_as_cat_sep(self):
         )
         xr.testing.assert_equal(rasterized.data, expected)
 
+    def test_rasterize_curve_with_timezone_aware_datetime(self):
+        # Test for https://github.com/holoviz/holoviews/issues/5127
+        t = pd.date_range(start='2020-01-01 10:00', end='2020-01-01 12:00', freq='h', tz='Asia/Shanghai')
+        d = pd.DataFrame({'t': t, 'y': [1, 2, 3]})
+        curve = Curve(d, kdims=['t'], vdims=['y'])
+
+        img = rasterize(curve, line_width=1, dynamic=False)
```
hoxbro:
```suggestion
        img = rasterize(curve, dynamic=False)
```

Issue #5222, opened 2022-02-23
Colorizing the points of a Scatter plot with the Plotly backend raises an error


maximlt:
Colorizing the points of a Scatter plot with the Plotly backend using `.opts(color=dim_name)` raises an error.

```python
import holoviews as hv
import pandas as pd

hv.extension('plotly')

df = pd.DataFrame([[1, 2, 'A', 0.1], [3, 4, 'B', 0.2], [5, 6, 'C', 0.3]],
                            columns=['x', 'y', 'category', 'number'])

hv.Scatter(df, kdims=['x'], vdims=['y', 'category']).opts(color='category')
```

Traceback:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
~/miniconda3/envs/hvplot-dev37/lib/python3.7/site-packages/IPython/core/formatters.py in __call__(self, obj, include, exclude)
    968 
    969             if method is not None:
--> 970                 return method(include=include, exclude=exclude)
    971             return None
    972         else:

~/work/dev/holoviews/holoviews/core/dimension.py in _repr_mimebundle_(self, include, exclude)
   1314         combined and returned.
   1315         """
-> 1316         return Store.render(self)
   1317 
   1318 

~/work/dev/holoviews/holoviews/core/options.py in render(cls, obj)
   1403         data, metadata = {}, {}
   1404         for hook in hooks:
-> 1405             ret = hook(obj)
   1406             if ret is None:
   1407                 continue

~/work/dev/holoviews/holoviews/ipython/display_hooks.py in pprint_display(obj)
    280     if not ip.display_formatter.formatters['text/plain'].pprint:
    281         return None
--> 282     return display(obj, raw_output=True)
    283 
    284 

~/work/dev/holoviews/holoviews/ipython/display_hooks.py in display(obj, raw_output, **kwargs)
    250     elif isinstance(obj, (CompositeOverlay, ViewableElement)):
    251         with option_state(obj):
--> 252             output = element_display(obj)
    253     elif isinstance(obj, (Layout, NdLayout, AdjointLayout)):
    254         with option_state(obj):

~/work/dev/holoviews/holoviews/ipython/display_hooks.py in wrapped(element)
    144         try:
    145             max_frames = OutputSettings.options['max_frames']
--> 146             mimebundle = fn(element, max_frames=max_frames)
    147             if mimebundle is None:
    148                 return {}, {}

[...]

PR #6735, opened 2025-11-17
test: Don't have test inheriting from `unittest.TestCase`

hoxbro:
This PR removes the test classes that inherit from `unittest.TestCase` and our own `ComparisonTestCase`, so we can no longer use `self.assertEqual`, etc. There is therefore added `holoviews.testing`, which contains `assert_data_equal`, `assert_element_equal`, and `assert_dict_equal`, and, combined with the normal assert, this should be sufficient; but we can always add more if we find the need. 

I have deprecated `ComparisonTestCase` along with `LoggingComparisonTestCase` in `holoviews/tests/utils.py` in version 1.25.0. So, a long time in the future.

Most of the code changes are simple replacements for `self.assertXXX`, though some small refactoring has been done, as we can now fully use pytest. 

I would like this merged after the release of 1.22.1. 

Checks:
- [x] verify no assert x,x is done
- [x] enable ruff pt
- [x] ~~check if `setup_method` -> `setup_class`~~. Will do this in a follow-up PR.
- [x] Force `zip(..., strict=True/False)`
- [x] `holoviews/tests/ipython/test_magics.py` + `utils/test_help_pattern` not able to run isolated. Couldn't figure out test_magics, and it's gonna be removed soon, as it is deprecated.
- [x] check GPU

image

Links:

- https://lorepirri.com/pytest-register-assert-rewrite.html
- https://sqlpey.com/python/top-5-methods-to-setup-and-teardown-for-your-pytest-class/

Script used:

<details><summary>patched ruff which don't update assertEqual </summary>

Run with `ruff-pt009 check holoviews --select=PT009 --unsafe-fixes`

``` diff
diff --git a/crates/ruff_linter/src/rules/flake8_pytest_style/rules/unittest_assert.rs b/crates/ruff_linter/src/rules/flake8_pytest_style/rules/unittest_assert.rs
index 6a5f89a6a..d02b5cedf 100644
--- a/crates/ruff_linter/src/rules/flake8_pytest_style/rules/unittest_assert.rs
+++ b/crates/ruff_linter/src/rules/flake8_pytest_style/rules/unittest_assert.rs
@@ -281,6 +281,11 @@ impl UnittestAssert {
     pub(crate) fn generate_assert(self, args: &[Expr], keywords: &[Keyword]) -> Result<Stmt> {
         let args = self.args_map(args, keywords)?;
         match self {
+            // Make Equal/Equals/NotEqual/NotEquals a no-op by returning an error
+            UnittestAssert::Equal
[...]

PR #6758, opened 2025-12-11
fix(plotly): categorical color error for Scatter plot

hoxbro:
Fixes: https://github.com/holoviz/holoviews/issues/5222

AI assisted by Claude Code.

Will merge it after https://github.com/holoviz/holoviews/pull/6735, as I want to update it to work with the

### DIFF
--- a/holoviews/tests/plotting/plotly/test_elementplot.py
+++ b/holoviews/tests/plotting/plotly/test_elementplot.py
@@ -290,7 +290,7 @@ class TestColorbarPlot(TestPlotlyPlot):
         ], vdims=['y', 'category']).opts(color='category', colorbar=True)
         state = self._get_plot_state(scatter)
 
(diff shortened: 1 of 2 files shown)


holoviews/plotting/util.py
```diff
@@ -1431,3 +1431,156 @@ def _process(self, element, key=None):
         except ImportError as exc:
             raise ImportError('Flattening ImageStacks requires datashader.') from exc
         return shade(element, **self.shade_params)
+
+
+# https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/named-color
+_NAMED_CSS_COLORS = (
```
philippjfr, line 1437: `param.Color._named_colors` has the list too.

hoxbro: Did not know that, but would rather have it directly in holoviews.

PR #6790, opened 2026-01-30
ci: Add Pull Request template

Azaya89:
This PR is mostly for me than anybody else but I think it will also be helpful to new contributors as well.


.github/pull_request_template.md
```diff
@@ -0,0 +1,60 @@
...
+- [ ] Bug fix
+- [ ] New feature
+- [ ] Breaking change
+- [ ] Documentation update
+- [ ] Other
+
+## Checklist
+
+- [ ] My PR title follows the conventional commit format shown above
+- [ ] I have added tests
```
hoxbro:
```suggestion
- [ ] I have added tests, and confirms that test suite is passing.
```

hoxbro (changes requested):
I'm not against this, but I think we should careful in adding stuff simply for oneself. 

I would like the information to be condenser, hidden by default (I don't want to see this when I open a PR every time), and link to developer guide for more information. 

Some other comments;
- Even if we are following conventional commit syntax, we only require this for the title, not every commit. This could likely be expressed clearer in the documentation.
- Type of Change section is redundant, that is why we use conventional commits syntax. Note that breaking changes is done with exclamation point `compat!:`

hoxbro (approved):
I have tighten up the template a bit. If you are happy with that we can merge this.

PR #6806, opened 2026-03-02
perf: Add numeric fast path to max_range and dimension_range

SimonHeybrock:
## Description

Profiling streaming dashboards points to frequent and suprisingly costly calls to `max_range` and `dimension_range`, accounting for nearly 10% of the server-side update time in our examples.

The vast majority of these calls receive a handful of numeric `(float, float)` tuples (often with `None` sentinels), but the implementation always allocates a NumPy array and runs `np.nanmin`/`np.nanmax`, which costs around 40–80 µs per call for this common case.

This PR adds a pure-Python fast path that handles numeric inputs using `math` operations, and an early return in `dimension_range` when both hard-range bounds are finite (since the rest of the computation would be overridden).

Non-numeric inputs (datetime, string, etc.) fall through to the existing implementation unchanged. Both paths share the same `_max_range_numeric` helper for numeric types, so there is only one implementation of the numeric min/max logic.

### Micro-benchmark

```python
import time, numpy as np
from holoviews.core.util import max_range

# Typical call pattern: 6 numeric range tuples from a 6-curve overlay
ranges = [(0.0, 100.0)] * 6

N = 100_000
t0 = time.perf_counter()
for _ in range(N):
    max_range(ranges)
us = (time.perf_counter() - t0) / N * 1e6
print(f"{us:.1f} µs/call")
# Before: ~78 µs   After: ~1.8 µs
```

A [streaming Panel dashboard](https://gist.github.com/SimonHeybrock/3e7109f958d839687303f769ae660ab0) was used to verify the improvement translates to reduced per-update latency in-browser (4 plots x 4 curves: ~120 ms → ~112 ms per update).

### Note on return types

The fast path returns Python `float` where the original `np.nanmin`/`np.nanmax` returned `np.float64`. I think it is unlikely to affect downstream code, but worth noting as a subtle behavioral change.

## Checklist

### DIFF
--- a/holoviews/core/util/__init__.py
+++ b/holoviews/core/util/__init__.py
@@ -981,18 +981,25 @@ _numeric_fast_types = (int, float, np.integer, np.floating)
 
 
 def _minmax_finite(values):
-    """Return (min, max) of finite floats, or (np.nan, np.nan) if none found."""
-    lo = math.inf
-    hi = -math.inf
-    found = False
+    """Return (min, max) of finite values, or (np.nan, np.nan) if none found.
+
+    Uses float conversion for NaN checks and comparison, but preserves the
+    original types in the returned values.
+    """
+    lo_f = math.inf
+    hi_f = -math.inf
+    lo_val = hi_val = None
     for v in values:
         if v is not None:
             fv = float(v)
             if not math.isnan(fv):
-                found = True
-                lo = min(lo, fv)
-                hi = max(hi, fv)
-    return (lo, hi) if found else (np.nan, np.nan)
+                if fv <= lo_f:
+                    lo_f = fv
+                    lo_val = v
+                if fv >= hi_f:
+                    hi_f = fv
+                    hi_val = v
+    return (lo_val, hi_val) if lo_val is not None else (np.nan, np.nan)
 
 
 def _max_range_numeric(ranges, combined):
--- a/holoviews/core/util/__init__.py
+++ b/holoviews/core/util/__init__.py
@@ -4,6 +4,7 @@ import hashlib
 import inspect
 import itertools
 import json
+import math
 import numbers
 import operator
 import pickle
@@ -976,6 +977,33 @@ def find_range(values, soft_range=None):
             return (None, None)
 
 
+_numeric_fast_types = (int, float, np.integer, np.floating)
+
+
+def _minmax_finite(values):
+    """Return (min, max) of finite floats, or (np.nan, np.nan) if none found."""
+    lo = math.inf
+    hi = -math.inf
+    found = False
+    for v in values:
+        if v is not None:
+            fv = float(v)
+            if not math.isnan(fv):
+                found = True
+                lo = min(lo, fv)
+                hi = max(hi, fv)
+    return (lo, hi) if found else (np.nan, np.nan)
+
+
+def _max_range_numeric(ranges, combined):
+    """Pure-Python fast path for max_range when all values are numeric or None."""
+    if combined:
+        return _minmax_finite(v for r in ranges for v in r)
+    lo, _ = _minmax_finite(r[0] for r in ranges)
+    _, hi = _minmax_finite(r[1] for r in ranges)
+    return lo, hi
+
+
 def max_range(ranges, combined=True):
     """Computes the maximal lower and upper bounds from a list bounds.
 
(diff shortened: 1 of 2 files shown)


holoviews/tests/core/test_utils.py
```diff
@@ -382,6 +420,102 @@ def test_max_range3(self):
         expected = (np.datetime64("1990", 'ns'), np.datetime64("1991", 'ns'))
         assert max_range(periods) == expected
 
+    # ── Numeric fast path ─────────────────────────────────────────
```
hoxbro: Maybe we should test if we actually call the fast path function?

hoxbro, line 472: Some small suggestions related to the tests:
- Maybe check that we actually call the fast path function.
- Try to see if we can `pytest.mark.parameterize` some of the tests, but not sure if this will make it harder to review the functionality.    
- Add a test mixing the different numeric types.

holoviews/core/util/__init__.py
```diff
@@ -976,6 +977,40 @@ def find_range(values, soft_range=None):
...
+        if v is not None:
+            fv = float(v)
+            if not math.isnan(fv):
+                if fv <= lo_f:
+                    lo_f = fv
+                    lo_val = v
+                if fv >= hi_f:
+                    hi_f = fv
+                    hi_val = v
+    return (lo_val, hi_val) if lo_val is not None else (np.nan, np.nan)
```
hoxbro, line 1004: Curious do we even need float conversion? Would something like this work:

``` python
def _minmax_finite(values):
    """Return (min, max) of finite values, or (np.nan, np.nan) if none found.
    """
    lo, hi, found = math.inf, -math.inf, False
    for v in values:
        if v is not None and not math.isnan(v):
            found = True
            lo = min(v, lo)
            hi = max(v, hi)
    return (lo, hi) if found else (np.nan, np.nan)
```

hoxbro (commented):
Thank you for the PR

holoviews/core/util/__init__.py
SimonHeybrock: There were some (seemingly) unrelated tests fails without the conversion - apparently some other code relied on the exact types, i.e., tests (of some other modules) failed if we naively convert `int->float`. I can have a look again whether we could/should fix those places instead, if you like?

SimonHeybrock: Here is the failed CI run related to this: https://github.com/holoviz/holoviews/actions/runs/22579349584/job/65407257546

holoviews/tests/core/test_utils.py
SimonHeybrock: I was in fact debating how one could check that the fast-path was called, as well as ensuring that behavior between the fast-path and regular path does not diverge on potential future changes, but I would not think of a good approach. Mocking to check the call count seems wrong here?

Will look into the other suggestions!

holoviews/core/util/__init__.py
hoxbro: As far as I can see that run did also do a float conversion. I'm asking if we can avoid the float conversion all together. It make sense tests could be failing if we do a float conversion.

holoviews/tests/core/test_utils.py
hoxbro: If it to hard to check, I don't think it is worth doing it. So if you don't find something which makes sense, we don't need to do it.

SimonHeybrock:
> As far as I can see that run did also do a float conversion. I'm asking if we can avoid the float conversion all together. It make sense tests could be failing if we do a float conversion.

Good question. We benchmarked dropping `float()` entirely -- the `min()`/`max()` vs `<=`/`>=` difference is negligible in the big picture (the real win is avoiding the numpy array path). But `float()` conversion does help noticeably for mixed numpy scalar types (e.g., `np.int64` vs `np.float64`), where without it the comparisons go through numpy's dispatch and are ~5x slower. Still fast compared to the old path, but a modest benefit for one line of code.

As a simplification based on your feedback, we dropped the explicit `math.isnan()` check -- NaN is already excluded by IEEE 754 comparison semantics (`NaN <= x` is always `False`), so it never updates the running min/max. One fewer branch in the loop.

On the test side: added type-preservation assertions (`type(lo) is int`, etc.) as a proxy for verifying the fast path is active -- the numpy slow path coerces Python `int` to `np.int64`, so if we get back a plain `int`, we know the fast path ran. Also added direct `_minmax_finite` tests, a mixed `int`/`float`/`np.int64`/`np.float64` test, and parametrized the combined/uncombined pairs.

hoxbro:
Did you forget to push?

> But float() conversion does help noticeably for mixed numpy scalar types (e.g., np.int64 vs np.float64),

I think we should write a small comment about this.

hoxbro (approved):

Issue #1947, opened 2017-10-01
Adding a colorbar changes plot dimensions (bokeh)


ceball:
https://github.com/ioam/holoviews/issues/167#issuecomment-129191860

<img width="511" alt="screen shot 2017-10-01 at 6 29 29 pm" src="https://user-images.githubusercontent.com/1929/31057194-89ed6c88-a6d6-11e7-839b-c698267dae55.png">

See https://github.com/bokeh/bokeh/issues/5186

Does not happen with mpl.

Hope you are ok having an issue that is only really a link to an issue on the underlying plotting backend, but initially I thought it was a problem with holoviews (until I did some digging).

philippjfr:
This will never get fixed because it's core to how bokeh's layouts work but you can now set the ``frame_width``/``frame_height`` to work around it.

PR #6815, opened 2026-03-09
fix(bokeh): Display colorbar label from value dimension

### LINKED ISSUE #1944: colorbar with bokeh is missing label
Bokeh:
screen shot 2017-10-01 at 10 31 27 am

Matplotlib:
screen shot 2017-10-01 at 10 32 12 am

I think this happens at least for bokeh 0.12.7 and 0.12.9 with hv 1.8.3, 1.8.4, and master (1.8.4-57-g96ecb1a).

### LINKED ISSUE #5982: A colorbar label doesn't automatically get created for labelled dataset for bokeh backend
I vaguely remember that the clabel automatically gets created a while back, so I think this might be a regression.

In matplotlib, it automatically shows "z"
```python
import numpy as np
import holoviews as hv
hv.extension('matplotlib')

ls  = np.linspace(0, 10, 400)
x,y = np.meshgrid(ls, ls)
img = hv.Image(np.sin(x)*np.cos(y)+0.1*np.random.rand(400,400), 
               bounds=(-20,-20,20,20)).opts(colorbar=True, xaxis=None, yaxis=None)

hv.Layout([img.relabel(c).opts(cmap=c) for c in ['gray','PiYG','flag','Set1']])
```
image

Changing it to bokeh, the clabel no longer exists.
image

Another example:
```python
import hvplot.xarray
import xarray as xr
import holoviews as hv
hv.extension("bokeh")

ds = xr.tutorial.open_dataset("air_temperature").isel(time=0)
hv.Image(ds, ["lon", "lat"], ["air"]).opts(colorbar=True).redim.label(lat="Latitude", lon="Longitude", air="Air Temperature")
```

SimonHeybrock:
## Description

The Bokeh backend's colorbar never displayed a title derived from the value dimension (name + unit via `pprint_label`), unlike the Matplotlib backend which does this correctly. For example, `hv.Dimension('Intensity', unit='counts')` would show `"Intensity (counts)"` as the colorbar label in Matplotlib but nothing in Bokeh.

Three issues contributed to this:

- `_draw_colorbar` only set the colorbar title when `clabel` was explicitly provided, never falling back to the dimension's `pprint_label`
- The `clabel` check used truthiness (`if self.clabel`) instead of `if self.clabel is not None`, so `clabel=""` could not be used to intentionally suppress the label
- `color_dim` was stored in handles without a prefix, while color mappers use prefixed keys (e.g. `fill_color_mapper`) — this meant multi-mapper scenarios would overwrite each other's stored dimension

I think this fixes #1944. There @philippjfr says *"did consider adding it at some point, but unfortunately I think adding it would make #1947 even worse, so at minimum we'd have to add some workaround to offset the size of the plot a bit."*, but in my experience that is still completely broken regardless.

Also fixes #5982, which looks like the same?

Fixes #1944
Fixes #5982

## How Has This Been Tested?

New unit tests for dimension label, clabel override, and clabel="" suppression. Verified tests fail when fix is reverted. Manual verification with Image, Points, HeatMap, QuadMesh, Scatter.

## AI Disclosure

Tools: Claude Code (Claude Opus)

## Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hoxbro:
Thank you for the PR.

Can you use the official PR template?

SimonHeybrock:
> Thank you for the PR.
> 
> Can you use the official PR template?

Sorry, updated now!

hoxbro (approved):

PR #6820, opened 2026-03-12
fix(plotting): Fix Bars crash with datetime x-axis for single or duplicate values

ghostiee-11:
## Description

Fixes `AttributeError: 'int' object has no attribute 'astype'` when rendering `hv.Bars` with a datetime x-axis that has either a single data point or all identical x-values.

In `BarPlot.get_data()`, when `np.diff()` returns an empty or all-zero array, `xdiff` gets reassigned to the plain Python int `1`. The subsequent `.astype('timedelta64[ms]')` call then fails because `int` doesn't have `.astype()`.

Fixed by wrapping the fallback as `np.array([1])` so `.astype()` and `np.min()` continue to work. Applied to both the Bokeh and Matplotlib backends.

Fixes #6813

## How Has This Been Tested?

Added two regression tests in `test_barplot.py`:
- `test_bars_continuous_datetime_single` - single datetime bar
- `test_bars_continuous_datetime_duplicates` - duplicate datetime x-values

Full barplot test suite passes (48/48).

## Checklist

### DIFF
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -1154,7 +1154,7 @@ class BarPlot(BarsMixin, ColorbarPlot, LegendPlot):
                 xdiff = np.abs(np.diff(xvals[xslice]))
                 diff_size = len(np.unique(xdiff))
                 if diff_size == 0 or (diff_size == 1 and xdiff[0] == 0):
-                    xdiff = 1
+                    xdiff = np.array([np.timedelta64(1, "D")]) if is_dt else np.array([1])
                 if is_dt:
                     width *= xdiff.astype("timedelta64[ms]").astype(np.int64)
                 else:
--- a/holoviews/plotting/bokeh/chart.py
+++ b/holoviews/plotting/bokeh/chart.py
@@ -1154,7 +1154,7 @@ class BarPlot(BarsMixin, ColorbarPlot, LegendPlot):
                 xdiff = np.abs(np.diff(xvals[xslice]))
                 diff_size = len(np.unique(xdiff))
                 if diff_size == 0 or (diff_size == 1 and xdiff[0] == 0):
-                    xdiff = np.array([np.timedelta64(1, "D")]) if is_dt else np.array([1])
+                    xdiff = np.array([np.timedelta64(1, "D")]) if is_dt else 1
                 if is_dt:
                     width *= xdiff.astype("timedelta64[ms]").astype(np.int64)
                 else:
--- a/holoviews/tests/plotting/bokeh/test_barplot.py
+++ b/holoviews/tests/plotting/bokeh/test_barplot.py
@@ -349,13 +349,11 @@ class TestBarPlot(TestBokehPlot):
         np.testing.assert_almost_equal(plot.handles["glyph"].width, 69120000.0)
 
     def test_bars_continuous_datetime_single(self):
-        # See: https://github.com/holoviz/holoviews/issues/6813
         bars = hv.Bars([(pd.Timestamp("2024-01-01"), 5)])
         plot = bokeh_renderer.get_plot(bars)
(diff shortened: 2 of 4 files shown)


hoxbro:
Can you disable your automatic code review?

holoviews/plotting/bokeh/chart.py
```diff
@@ -1154,7 +1154,7 @@ def get_data(self, element, ranges, style):
                 xdiff = np.abs(np.diff(xvals[xslice]))
                 diff_size = len(np.unique(xdiff))
                 if diff_size == 0 or (diff_size == 1 and xdiff[0] == 0):
-                    xdiff = 1
+                    xdiff = np.array([np.timedelta64(1, "D")]) if is_dt else np.array([1])
```
hoxbro:
```suggestion
                    xdiff = np.array([np.timedelta64(1, "D")]) if is_dt else 1
```

Curious why did you choose day?

holoviews/tests/plotting/bokeh/test_barplot.py
```diff
@@ -348,6 +348,18 @@ def test_bars_continuous_datetime(self):
         plot = bokeh_renderer.get_plot(bars)
         np.testing.assert_almost_equal(plot.handles["glyph"].width, 69120000.0)
 
+    def test_bars_continuous_datetime_single(self):
+        # See: https://github.com/holoviz/holoviews/issues/6813
```
hoxbro:
```suggestion
```

holoviews/tests/plotting/bokeh/test_barplot.py
```diff
@@ -348,6 +348,18 @@ def test_bars_continuous_datetime(self):
...
+        bars = hv.Bars([(pd.Timestamp("2024-01-01"), 5)])
+        plot = bokeh_renderer.get_plot(bars)
+        np.testing.assert_almost_equal(plot.handles["glyph"].width, 69120000.0)
+
+    def test_bars_continuous_datetime_duplicates(self):
+        # See: https://github.com/holoviz/holoviews/issues/6813
```
hoxbro:
```suggestion
```

The test speaks for itself. 

We should likely add similar test for the Matplotlib backend.

ghostiee-11: When i didnt added this ci failed, so I added this afterwards

hoxbro: What are you talking about? CI never ran for you here and this is about removing a comment.

holoviews/plotting/bokeh/chart.py
ghostiee-11: I chose day because when there's no interval to calculate from, we need a sensible fallback. Using something like 1ms would make the bar practically invisible. Day keeps it consistent with the existing daily datetime test too.

holoviews/tests/plotting/bokeh/test_barplot.py
ghostiee-11: Sorry, I confused it with something else. I saw similar # See: comments on other tests in the file (like test_bars_continuous_datetime_timezone_in_overlay) and followed that pattern, but you're right - the test name is clear enough. Removed the comments and added Matplotlib tests, pushing now.

holoviews/plotting/bokeh/chart.py
hoxbro: Can you show me a screenshot of the difference between day and ms?

ghostiee-11: Here's the comparison with the same axis range:

Left is 1 day fallback, right is 1 ms - practically invisible.
Screenshot 2026-03-13 at 1 47 31 PM

hoxbro:
Thank you for the PR. 

But can you please disable automatic PR review?

ghostiee-11:
Thanks for the review!!

I disabled the automatic PR review of Copilot.

PR #6831, opened 2026-03-18
perf: Cache `gen_types` type tuples

SimonHeybrock:
## Description

The `gen_types` metaclass (`_GeneratorIsMeta`) re-evaluates its generator function on every `isinstance()` and `issubclass()` call. Each evaluation checks `_LazyModule.__bool__` and `_LazyModule.__getattr__` for numpy, pandas, and cftime, then rebuilds the type tuple from scratch.

This PR caches the resulting type tuple after first evaluation via a `_get_types()` helper on the metaclass. Subsequent calls return the cached tuple directly, skipping the generator, lazy module checks, and nested `yield from` chains. Because the generators use `_LazyModule` with `bool_use_sys_modules=True`, the type tuple can grow as optional libraries get imported — the cache is invalidated whenever `len(sys.modules)` changes to handle this correctly.

As pointed out by @hoxbro in https://discourse.holoviz.org/t/optimizing-holoviews-and-param-for-faster-streaming-dynamicmap/9121/2, a more naive initial approach was insufficient as it failed to deal with import ordering issues. The cache invalidation mechanism added in this version did not measurably deteriorate the performance gains.

Headless benchmark (median ms/update, cProfile active, Bokeh validation disabled):

| Scenario | Before | After | Change |
|---|---|---|---|
| Curve x1 | 18.2 ms | 16.3 ms | -10% |
| Curve x6 | 54.9 ms | 45.2 ms | -18% |
| Scatter x6 | 72.2 ms | 61.3 ms | -15% |
| Points x6 | 72.6 ms | 62.6 ms | -14% |
| Image | 22.0 ms | 19.7 ms | -10% |
| Bars | 23.8 ms | 19.5 ms | -18% |
| HeatMap | 34.5 ms | 29.4 ms | -15% |

End-to-end with the [streaming app](https://gist.github.com/SimonHeybrock/880de22fd75dfa56d0299c9f4fe5773f) (4 plots x 6 curves, 100 ms interval, hold=True, freeze=True): 115 ms -> 107 ms median per update.

## How Has This Been Tested?

Existing core test suite (1963 tests) covers correctness of `isinstance`/`issubclass`/iteration for all gen_types. The caching is transparent.

Dedicated unit tests (`types_test.py`) verify cache invalidation: a `gen_types` using a `_LazyModule` with `bool_use_sys_modules=True` is evaluated before and after a late import. The tests confirm that the cached type tuple grows to include the newly available type, and that the cache is reused when `sys.modules` is unchanged.
However, these tests require some slightly hacky setup, so I am not certain they are truly useful to keep. They are isolated in their own module though. If undesired, we can just revert the second commit.

## AI Disclosure

Tools: Claude Code (Opus)

## Checklist


holoviews/core/util/types.py
```diff
@@ -20,21 +20,49 @@
 
 # gen_types is copied from param, can be removed when
 # we support 2.2 or greater
+
+_cache_state: list[int] = [0]  # mutable container for sys.modules length
```
hoxbro: Curious can we use `global` inside `_get_types` and have this as int? With performance in mind.

If not, comment is not needed. 

```suggestion
_cache_state: list[int] = [0]
```

holoviews/core/util/types.py
```diff
@@ -20,21 +20,49 @@
...
 class _GeneratorIsMeta(type):
+    def _get_types(cls) -> tuple[type, ...]:
+        # The gen_types generators use _LazyModule with
+        # bool_use_sys_modules=True, so `if np:` / `if pd:` / `if cftime:`
+        # return True only when the library is both installed AND already
+        # imported (in sys.modules). This means the type tuple a generator
+        # yields can grow over time as libraries get imported.
+        #
+        # We cache the evaluated tuple for performance, but invalidate all
+        # caches whenever sys.modules grows, so late imports are picked up.
```
hoxbro:
```suggestion
        # Cache types and invalidate on new imports
```

This is very verbose.

holoviews/core/util/types.py
```diff
@@ -20,21 +20,49 @@
...
+        n = len(sys.modules)
+        if n != _cache_state[0]:
+            _cache_state[0] = n
+            for sub in _GeneratorIs.__subclasses__():
+                try:
+                    del sub._cached_types
+                except AttributeError:
+                    pass
```
hoxbro: Can we have `_cached_types` as `None`, and remove the try/except handling for if-statement, again with performance in-mind.

holoviews/tests/core/util/types_test.py
```diff
@@ -0,0 +1,93 @@
...
+import types
+
+import pytest
+
+from holoviews.core.util.dependencies import _LazyModule
+from holoviews.core.util.types import gen_types
+
+
+class _FakeIsInstalled:
+    """Stand-in for the lru_cached _is_installed that treats a fake module as installed."""
```
hoxbro: This docstring seem out-of-date.

holoviews/tests/core/util/types_test.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def cache_clear(self):
+        pass
+
+
+FAKE_MODULE_NAME = "_hv_test_fake_lib"
+
+
+@pytest.fixture
+def _fake_optional_library():
```
hoxbro: Maybe this could be cleaner if we monkey patch `sys.modules`, and maybe with a real numpy?

hoxbro (commented):
Very happy with the invalidate cache approach here.

Have left some comments.

holoviews/core/util/types.py
SimonHeybrock: I did initially use `global` but there was a pre-commit check (ruff, iirc) that was not happy about it. Could revert to that and add an ignore?

holoviews/tests/core/util/types_test.py
SimonHeybrock: The problem with using a real `numpy` or similar is that they might have been imported already by the time these tests run, so we'd not be able to test what we want to test.

holoviews/core/util/types.py
hoxbro: Yes, you are welcome to ignore the rule on that line.

holoviews/tests/core/util/types_test.py
hoxbro: Can't we remove numpy for the monkey patched sys.modules?

SimonHeybrock: Regarding patching `sys.modules`, after thinking about this more (patch by removing numpy), I think it would work in principle, but it does not *grow* the length of `sys.modules` w.r.t. the length we may have stored earlier, i.e., the cache invalidation would not catch this?

hoxbro: I will play around with it a bit, to see what feels best, later this week.

hoxbro: Pushed some changes. Will merge later unless you object.

SimonHeybrock: Thanks, looks much cleaner, LGTM 👍

PR #6832, opened 2026-03-18
perf: Disable pipeline bookkeeping during plot refresh

SimonHeybrock:
## Description

Every public `Dataset` method call (`range`, `dimension_values`, `sort`, etc.) is wrapped by `PipelineMeta.pipelined`, which records operations into a pipeline object for reproducibility. This wrapper performs an import, `copy.copy` of the pipeline, `_in_method` flag tracking, and `method_op.instance(...)` creation on every call — even when pipeline recording is disabled via the existing `disable_pipeline()` context manager.

Two changes:

- Move the `PipelineMeta.disable` check to the top of `pipelined_fn`, before any bookkeeping. Previously the check happened after the method call, so all overhead was still incurred. Now disabled calls go straight to the wrapped method with zero overhead.
- Wrap `Plot.refresh()` → `_trigger_refresh()` in `disable_pipeline()`. During plot refresh, all Dataset method calls are read-only queries — no new Datasets are created that need pipeline tracking. The `disable_pipeline` context manager already existed and was used in two other places (renderer, compositor) but not in the update path that dominates streaming performance. **Note: This is a part I was not certain about, so in the end we added [f242880](https://github.com/holoviz/holoviews/pull/6832/commits/f242880a7a35acf7c41bd5a6f03ba4445df4652c) (second commit) - is this actually necessary, or is the first commit correct on its own?**

End-to-end with the [streaming app](https://gist.github.com/SimonHeybrock/880de22fd75dfa56d0299c9f4fe5773f) (4 plots x 6 curves, 20 ms interval, hold=True, freeze=True):
- 115 ms -> 101 ms median per update.
- 115 ms -> 96 ms median per update when combined with #6831.

Headless benchmark (Curve x6 overlay, 1000 pts/curve, 60 updates, median of 5 runs):

| Scenario | Before | After | Change |
|---|---|---|---|
| With Bokeh validation | 23.2 ms | 21.7 ms | -6.6% |
| Without Bokeh validation | 18.6 ms | 16.1 ms | -13.0% |

## How Has This Been Tested?

Existing Bokeh plotting tests (1180 passed) and core tests (719 passed) cover correctness. The pipeline mechanism is exercised throughout the test suite via normal Dataset operations.

## AI Disclosure

Tools: Claude Code (Opus)

## Checklist

### DIFF
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -188,7 +188,7 @@ def disable_pipeline():
 
 @contextmanager
 def enable_pipeline():
-    """Re-enable PipelineMeta class for storing pipelines."""
+    """Enable PipelineMeta class for storing pipelines."""
     prev = PipelineMeta.disable
     PipelineMeta.disable = False
     try:
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -178,11 +178,23 @@ class DataConversion:
 @contextmanager
 def disable_pipeline():
     """Disable PipelineMeta class from storing pipelines."""
+    prev = PipelineMeta.disable
     PipelineMeta.disable = True
     try:
         yield
     finally:
-        PipelineMeta.disable = False
+        PipelineMeta.disable = prev
+
+
+@contextmanager
+def enable_pipeline():
+    """Re-enable PipelineMeta class for storing pipelines."""
+    prev = PipelineMeta.disable
+    PipelineMeta.disable = False
+    try:
+        yield
+    finally:
+        PipelineMeta.disable = prev
 
 
 class PipelineMeta(ParameterizedMetaclass):
--- a/holoviews/core/data/__init__.py
+++ b/holoviews/core/data/__init__.py
@@ -207,6 +207,9 @@ class PipelineMeta(ParameterizedMetaclass):
     def pipelined(method_fn, method_name):
         @wraps(method_fn)
         def pipelined_fn(*args, **kwargs):
+            if PipelineMeta.disable:
+                return method_fn(*args, **kwargs)
+
             from ...operation.element import method as method_op
 
             inst = args[0]
@@ -217,8 +220,6 @@ class PipelineMeta(ParameterizedMetaclass):
 
             try:
                 result = method_fn(*args, **kwargs)
-                if PipelineMeta.disable:
-                    return result
 
                 op = method_op.instance(
                     input_type=type(inst),
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -243,7 +243,8 @@ class Plot(param.Parameterized):
             )
             stream_key = util.wrap_tuple_streams(key, self.dimensions, self.streams)
 
-            self._trigger_refresh(stream_key)
+            with disable_pipeline():
+                self._trigger_refresh(stream_key)
             if self.top_level:
                 self.push()
         except Exception as e:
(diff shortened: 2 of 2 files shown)


hoxbro:
Have you checked this does not break any operation/datashader workflow?

SimonHeybrock:
> Have you checked this does not break any operation/datashader workflow?

Have never used those features, so I am relying on Claude here. Built a demo app (see https://gist.github.com/SimonHeybrock/48a597fbc24a70f7577530efd968017d) to try it out. This all seems to work, and it actually proves that [f242880](https://github.com/holoviz/holoviews/pull/6832/commits/f242880a7a35acf7c41bd5a6f03ba4445df4652c) is necessary - if I revert that commit the `cat:` hover info is always just showing `cat: A`, for example).

With this branch (all commits):

image

Can you point me to further things to check, if any?

hoxbro:
> Can you point me to further things to check, if any?

Does it update the plot when you zoom in? If it stays pixelated that is a problem.

SimonHeybrock:
> > Can you point me to further things to check, if any?
> 
> Does it update the plot when you zoom in? If it stays pixelated that is a problem.

It does in the static case (top left) but not the DynamicMap (top right) - but that is the same behavior as in `main`. Can this work in principle with a dmap? I may be missing an insight on how to set it up.

hoxbro:
What does this do with this PR?

``` python
import holoviews as hv
import numpy as np

hv.extension("bokeh")

rasterize(hv.Scatter(np.random.randn(1_000_000, 2)))
````
This is how it looks for me:

https://github.com/user-attachments/assets/449ae475-0eb6-4d1c-9fa9-d20db86be335

**Comment by SimonHeybrock** [comment]
> What does this do with this PR?

Works (with and without [f242880](https://github.com/holoviz/holoviews/pull/6832/commits/f242880a7a35acf7c41bd5a6f03ba4445df4652c))!

**Comment by philippjfr** [comment]
Yeah, I don't think disabling the book-keeping while the plotting code executes is a problem. There's also perhaps a further optimization in tracking only certain operations. Any method that does not return a new HoloViews object does not need to be tracked (I think).

File: holoviews/plotting/plot.py [hunk 1]
```diff
@@ -1500,7 +1501,17 @@ def _get_frame(self, key):
...
+        # Plot.refresh() disables pipeline for the entire update path (see
+        # the disable_pipeline() call there), but user callbacks may call
+        # Dataset methods (e.g. select) whose results should retain pipeline
+        # provenance.
+        prev_disable = PipelineMeta.disable
+        PipelineMeta.disable = False
+        try:
+            frame = get_plot_frame(self.hmap, key_map, cached)
+        finally:
+            PipelineMeta.disable = prev_disable
```
hoxbro: I think we should update `disable_pipeline` with restoring to previous stage and use it here. 

Unless I'm misunderstanding something.

SimonHeybrock: Good idea! Latest commit does this and adds `enable_pipeline`.

holoviews/core/data/__init__.py
```diff
@@ -178,11 +178,23 @@ def __call__(self, new_type, kdims=None, vdims=None, groupby=None, sort=False, *
...
     try:
         yield
     finally:
-        PipelineMeta.disable = False
+        PipelineMeta.disable = prev
+
+
+@contextmanager
+def enable_pipeline():
+    """Re-enable PipelineMeta class for storing pipelines."""
```
hoxbro:
```suggestion
    """Enable PipelineMeta class for storing pipelines."""
```

hoxbro:
Let's merge this. Will cut a dev release after https://github.com/holoviz/holoviews/pull/6831 has been merged.

PR #6835, opened 2026-03-19
perf(bokeh): Skip Bokeh property validation in _update_datasource

SimonHeybrock:
## Description

Skip Bokeh's per-element property validation when updating `ColumnDataSource.data` in `_update_datasource`. **Users who already disable validation globally will see no change** - but this PR makes the optimization the default without requiring user-side configuration.

End-to-end improvement measured with the [streaming app](https://gist.github.com/SimonHeybrock/880de22fd75dfa56d0299c9f4fe5773f) (4 plots × 4 curves × 1000 points):

| | main | this PR |
|---|---|---|
| median | 97 ms | 86 ms |

**Mechanism:** Bokeh's `ColumnData` property is typed as `ColumnData(String, Seq(Any))`. On every data assignment, `Seq.validate()` iterates every element calling `Any.is_valid()` — which is a no-op since `Any` accepts everything. For a 6-curve overlay with 1000 points per curve, this amounts to ~12k no-op validation calls per update. Wrapping with Bokeh's `validate(False)` context manager skips these. The important safety checks — column-length equality assertions and `PropertyValueColumnData` wrapping — still run regardless of the validation flag.

See also https://discourse.holoviz.org/t/optimizing-holoviews-and-param-for-faster-streaming-dynamicmap/9121. See the [streaming app](https://gist.github.com/SimonHeybrock/880de22fd75dfa56d0299c9f4fe5773f), [streaming benchmark](https://gist.github.com/SimonHeybrock/5ffe34160f661f618f2d66e925e1ab1a), and [profiling suite](https://gist.github.com/SimonHeybrock/4d9654473549b22c138c09b9250ddd15) gists to reproduce.

## How Has This Been Tested?

Existing Bokeh plotting test suite passes. No dedicated test added — the change wraps existing code with a context manager and introduces no new behavior or code paths.

## AI Disclosure

Tools: Claude Code (Opus)

## Checklist


holoviews/plotting/bokeh/plot.py
```diff
@@ -214,10 +215,11 @@ def _update_datasource(self, source, data):
...
 
-        if cds_column_replace(source, data):
-            source.data = data
-        else:
-            source.data.update(data)
+        with validate(False):
+            if cds_column_replace(source, data):
+                source.data = data
+            else:
+                source.data.update(data)
```
hoxbro:
```suggestion
        if cds_column_replace(source, data):
            source.data = data
        else:
            with validate(False):
                source.data.update(data)
```

I think we should `validate` if we completely changes the data. Or am I missing something?

SimonHeybrock: I don't know enough to tell whether this data (constructed by the Holoviews plotting pipeline?) can ever be invalid. But Presumably this is not the hot branch, so not skipping validation would not hurt much.

hoxbro: Lets start with doing it only for `source.data.update`.

hoxbro (approved):

PR #6837, opened 2026-03-23
perf: Skip param.update when plot options unchanged

SimonHeybrock:
## Description

Cache the plot options dict returned by `lookup_options` and skip the `param.update()` call when the dict matches the previous frame. Applied in `ElementPlot.update_frame` and `OverlayPlot.update_frame` across all backends (bokeh, matplotlib, plotly) via `_prev_plot_opts` on`GenericElementPlot`.

`param.update` iterates all parameters to build a restore dict, sets attributes, and checks watchers. In streaming scenarios where plot options are static between frames, this is pure overhead.

End-to-end improvement measured with the Bokeh-backend
[streaming app](https://gist.github.com/SimonHeybrock/880de22fd75dfa56d0299c9f4fe5773f)
(4 plots × 4 curves × 1000 points, `panel serve`, Chrome):

| | main | this PR |
|---|---|---|
| median | 98 ms | 92 ms |

This is part of a series of independent performance improvements to the
streaming/DynamicMap update path.
See the [streaming app](https://gist.github.com/SimonHeybrock/880de22fd75dfa56d0299c9f4fe5773f),
[streaming benchmark](https://gist.github.com/SimonHeybrock/5ffe34160f661f618f2d66e925e1ab1a), and
[profiling suite](https://gist.github.com/SimonHeybrock/4d9654473549b22c138c09b9250ddd15)
gists to reproduce.

## How Has This Been Tested?

Existing plotting test suites pass across all backends (bokeh: 1193, matplotlib: 368). No dedicated test added — the change guards an existing call with a cache check and introduces no new behavior.

## AI Disclosure

Tools: Claude Code (Opus)

## Checklist

### DIFF
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -1384,7 +1384,7 @@ class GenericElementPlot(DimensionedPlot):
     def _apply_plot_opts(self, plot_opts: dict) -> None:
         """Apply plot options, skipping if unchanged from the previous frame."""
         try:
-            unchanged = plot_opts == self._prev_plot_opts
+            unchanged = bool(plot_opts == self._prev_plot_opts)
         except (ValueError, TypeError):
             # Options may contain numpy arrays where == is element-wise
             unchanged = False
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -1379,6 +1379,14 @@ class GenericElementPlot(DimensionedPlot):
 
     _multi_y_propagation = False
 
+    _prev_plot_opts = None
+
+    def _apply_plot_opts(self, plot_opts: dict) -> None:
+        """Apply plot options, skipping if unchanged from the previous frame."""
+        if plot_opts != self._prev_plot_opts:
+            self._prev_plot_opts = plot_opts
+            self.param.update(**plot_opts)
+
     def __init__(
         self,
         element,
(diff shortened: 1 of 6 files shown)


holoviews/plotting/plot.py
```diff
@@ -1379,6 +1379,14 @@ class GenericElementPlot(DimensionedPlot):
 
     _multi_y_propagation = False
 
+    _prev_plot_opts = None
+
+    def _apply_plot_opts(self, plot_opts: dict) -> None:
+        """Apply plot options, skipping if unchanged from the previous frame."""
+        if plot_opts != self._prev_plot_opts:
```
hoxbro: This could error out if something is a numpy array

Image

holoviews/plotting/plot.py
```diff
@@ -1379,6 +1379,19 @@ class GenericElementPlot(DimensionedPlot):
 
     _multi_y_propagation = False
 
+    _prev_plot_opts = None
+
+    def _apply_plot_opts(self, plot_opts: dict) -> None:
+        """Apply plot options, skipping if unchanged from the previous frame."""
+        try:
+            unchanged = plot_opts == self._prev_plot_opts
```
hoxbro:
```suggestion
            unchanged = bool(plot_opts == self._prev_plot_opts)
```

Just to be 100% sure it is a boolean.

An example for this is `bool(pd.NA == pd.NA)`, though I don't think this is a problem here.

PR #6839, opened 2026-03-24
perf(bokeh): Cache static layout properties in _update_plot

SimonHeybrock:
## Description

This is a PR similar to #6837. This adds caching for the computed layout properties (plot properties, axis label properties, title properties, and grid options) between frames in the Bokeh `ElementPlot`. When the values match the previous frame, skip the corresponding Bokeh model updates.

During streaming with a fixed layout (the common case, I presume) these properties are constant between frames, making the Bokeh model traversals and property-setting pure overhead.

End-to-end improvement measured with the Bokeh-backend [streaming app](https://gist.github.com/SimonHeybrock/880de22fd75dfa56d0299c9f4fe5773f) (4 plots × 4 curves × 1000 points, `panel serve`, Chrome):

| | main | this PR |
|---|---|---|
| median | 91 ms | 85 ms |

Headless [profiling suite](https://gist.github.com/SimonHeybrock/4d9654473549b22c138c09b9250ddd15) breakdown (cProfile-instrumented, median ms/update across 60 updates):

| scenario | main | this PR | change |
|---|---|---|---|
| Curve x1 | 14.2 | 8.4 | -41% |
| Curve x6 | 34.7 | 29.4 | -15% |
| Scatter x6 | 48.1 | 42.2 | -12% |
| Points x6 | 46.4 | 40.8 | -12% |
| Image | 18.0 | 12.4 | -31% |
| Bars | 16.4 | 10.3 | -37% |
| HeatMap | 25.0 | 19.9 | -20% |

cProfile inflates absolute times; relative comparisons are meaningful. Single-element scenarios show the largest relative improvement because layout overhead is a bigger fraction of total time.

## How Has This Been Tested?

Full Bokeh plotting test suite passes (1195 tests). Three dedicated tests cover the cache-hit path (static layout with streaming data), the cache-invalidation path (layout options changing between frames), and numpy array safety in grid style values.

## AI Disclosure

Tools: Claude Code (Opus)

## Checklist


holoviews/plotting/bokeh/element.py
```diff
@@ -1470,14 +1490,17 @@ def _update_labels(self, key, plot, element):
...
         recursive_model_update(plot.yaxis[0], props.get("y", {}))
 
     def _update_title(self, key, plot, element):
+        props = self._title_properties(key, plot, element)
+        if not self._update_cache("_prev_title_props", props):
+            return
         if plot.title:
-            plot.title.update(**self._title_properties(key, plot, element))
-        else:
-            plot.title = Title(**self._title_properties(key, plot, element))
```
hoxbro, line 1480: Not sure about this deletion, even if test doesn't hit this line.

hoxbro (commented):
Pushed some changes, as it was easier than explaining. Can you check if those changes still show good performance?

Otherwise, I only have one small comment.

holoviews/plotting/bokeh/element.py
SimonHeybrock: The coverage check failed on this because it was apparently unreachable (see description of https://github.com/holoviz/holoviews/pull/6839/changes/abf0582215656830026192fd2bed247231b785dd).

hoxbro: But could be a problem on older versions of Bokeh. 

Also if this line really is unreachable in all supported Bokeh versions, we should likely remove the if-statement completely `plot.title`.

SimonHeybrock: Good point, I only checked the current Bokeh version, did not consider old ones!

SimonHeybrock:
> Pushed some changes, as it was easier than explaining. Can you check if those changes still show good performance?

Yes, still good performance, no regression!

holoviews/plotting/bokeh/element.py
```diff
@@ -1470,14 +1487,19 @@ def _update_labels(self, key, plot, element):
...
     def _update_title(self, key, plot, element):
+        props = self._title_properties(key, plot, element)
+        if not self._has_changed("title", props):
+            return
         if plot.title:
-            plot.title.update(**self._title_properties(key, plot, element))
-        else:
-            plot.title = Title(**self._title_properties(key, plot, element))
+            plot.title.update(**props)
+        else:  # pragma: no cover
```
hoxbro:
```suggestion
        else:
```

I'm OK with having this being flagged as not run by coverage.

hoxbro (approved):

PR #6841, opened 2026-03-26
ci: Add retry option to download test data

hoxbro:
## Description

Github (used to download data) seems more and more flaky when downloading data. This adds a simple retry step, so we don't have to wait for CI to finish to rerun. 

## How Has This Been Tested?

CI


hoxbro:
This worked as intended:

image

PR #6851, opened 2026-04-08
type:  Add type checking on isolated and utility files

### LINKED ISSUE #6653: `stream.Pipe` memoize option unused?
Is was trying to understand what the `memoize` option to `holoviews.streams.Pipe` does, but contrary to what the docstring says this option seems unused (unless the is some metaclass magic that I cannot see):
https://github.com/holoviz/holoviews/blob/0541eae098bfd773c4cec419ee5d5e5ff35a95f1/holoviews/streams.py#L514-L540

Is this intentional or a bug?

hoxbro:
## Description

With param now having type hints. I want to start adding it to HoloViews. I have chosen to start soft by only using `TY` and starting with a small number files, this is done to avoid having one big PR and never having it merged.

The files I chosen for is defined in `pyproject.toml` and copied below for convenience: 

``` toml
[tool.ty.src]
include = [
    "holoviews/*.py",
    "holoviews/core/*.py",
    "holoviews/core/util/*.py",
    "holoviews/util/*.py",
    "scripts/*.py",
]
exclude = ["holoviews/*version.py"]
```

Most of the changes are related to tree things:

- Liskov substitution principle
- `zip(..., strict=None)` which I set to avoid enable linting rule on new code, but is flagged with type checking. Have tried my best to set `strict=True` but if in doubt I have used `False`
- Optional imports both in the codebase and in tests, where we now define them in one file to avoid a lot of the boilerplate needed for type checker. For codebase it is in `holoviews/core/util/dependencies.py` and for tests it is `holoviews/tests/utils.py`.

Resolves #6653

## How Has This Been Tested?

CI + Locally
## Checklist

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -216,7 +216,7 @@ jobs:
       - uses: holoviz-dev/holoviz_tasks/pixi_install@v0
         with:
           environments: ${{ matrix.environment }}
-      - name: Test Unit
+      - name: Type TY
         run: |
           pixi run -e ${{ matrix.environment }} type-ty
 
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -201,9 +201,28 @@ jobs:
         run: |
           pixi run -e ${{ matrix.environment }} test-unit
 
+  type_suite:
+    name: type:${{ matrix.environment }}:${{ matrix.os }}
+    needs: [pre_commit, setup, pixi_lock]
+    runs-on: ${{ matrix.os }}
+    if: needs.setup.outputs.code_change == 'true'
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest"]
+        environment: ["type"]
+    timeout-minutes: 5
+    steps:
+      - uses: holoviz-dev/holoviz_tasks/pixi_install@v0
+        with:
+          environments: ${{ matrix.environment }}
+      - name: Test Unit
+        run: |
+          pixi run -e ${{ matrix.environment }} type-ty
+
   result_test_suite:
     name: result:test
-    needs: [unit_test_suite, ui_test_suite, core_test_suite]
+    needs: [unit_test_suite, ui_test_suite, core_test_suite, type_suite]
     if: always()
     runs-on: ubuntu-latest
     steps:
--- a/holoviews/__init__.py
+++ b/holoviews/__init__.py
@@ -182,6 +182,8 @@ from .util.warnings import (
 )
 
 TYPE_CHECKING = False
+extension: type[extension]
+
 
 if hasattr(builtins, "__IPYTHON__"):
     from .ipython import notebook_extension
@@ -199,9 +201,7 @@ if "_pyodide" in sys.modules:
 
     # The notebook_extension is needed inside jupyterlite,
     # so the override is only done if we are not inside jupyterlite.
-    if in_jupyterlite():
-        extension.inline = False
-    else:
+    if not in_jupyterlite():
         extension = pyodide_extension
     del pyodide_extension, in_jupyterlite
 
--- a/holoviews/__init__.py
+++ b/holoviews/__init__.py
@@ -182,7 +182,8 @@ from .util.warnings import (
 )
 
 TYPE_CHECKING = False
-extension: type[extension]
+if TYPE_CHECKING:
+    extension: type[extension]
 
 
 if hasattr(builtins, "__IPYTHON__"):
@@ -207,7 +208,7 @@ if "_pyodide" in sys.modules:
 
 if TYPE_CHECKING:
     # Adding this here to have better docstring in LSP
-    from .util import extension  # noqa: TC004
+    from .util import extension
 
 _load_rc_file()
 
--- a/holoviews/annotators.py
+++ b/holoviews/annotators.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import sys
+import typing as t
 from inspect import getmro
 
 import param
@@ -8,7 +8,7 @@ from panel.layout import Row, Tabs
 from panel.pane import PaneBase
 from panel.util import param_name
 
-from .core import DynamicMap, Element, HoloMap, Layout, Overlay, Store, ViewableElement
+from .core import Dataset, DynamicMap, Element, HoloMap, Layout, Overlay, Store, ViewableElement
 from .core.util import isscalar
 from .element import Curve, Path, Points, Polygons, Rectangles, Table
 from .plotting.links import (
@@ -17,10 +17,15 @@ from .plotting.links import (
     SelectionLink,
     VertexTableLink,
 )
-from .streams import BoxEdit, CurveEdit, PointDraw, PolyDraw, PolyEdit, Selection1D
+from .streams import BoxEdit, CDSStream, CurveEdit, PointDraw, PolyDraw, PolyEdit, Selection1D
+
+if t.TYPE_CHECKING:
+    from collections.abc import Callable
+
+    _T = t.TypeVar("_T")
 
 
-def preprocess(function, current=None):
+def preprocess(function):
     """Turns a param.depends watch call into a preprocessor method, i.e.
     skips all downstream events triggered by it.
 
@@ -32,8 +37,6 @@ def preprocess(function, current=None):
     See https://github.com/holoviz/param/issues/332
 
     """
-    if current is None:
-        current = []
 
     def inner(*args, **kwargs):
         self = args[0]
@@ -225,7 +228,7 @@ class Annotator(PaneBase):
     )
 
     object = param.ClassSelector(
-        class_=Element,
+        class_=Dataset,
         doc="The Element to edit and annotate.",
     )
 
@@ -264,10 +267,14 @@ class Annotator(PaneBase):
 
     priority = 0.7
 
+    _init_stream: Callable[[], None]
+    _process_element: Callable[[t.Any], Dataset]
+    _stream: CDSStream
+    _selection: Selection1D
+    name: str
+
     @classmethod
     def applies(cls, obj):
-        if "holoviews" not in sys.modules:
-            return False
         return isinstance(obj, cls.param.object.class_)
 
     @property
@@ -363,7 +370,7 @@ class Annotator(PaneBase):
 
     @property
     def tables(self):
-        return list(zip(self.editor._names, self.editor, strict=None))
+        return list(zip(self.editor._names, self.editor, strict=True))
 
     @property
     def selected(self):
@@ -405,6 +412,8 @@ class PathAnnotator(Annotator):
     _vertex_table_link = VertexTableLink
 
     _triggers = ["annotations", "edit_vertices", "object", "table_opts", "vertex_annotations"]
+
+    _stream: PolyDraw
 
     def __init__(self, object=None, **params):
         self._vertex_table_row = Row()
@@ -533,9 +542,9 @@ class _GeomAnnotator(Annotator):
         doc="Opts to apply to the element.",
     )
 
-    _stream_type = None
-
     __abstract = True
+
+    _stream_type: t.ClassVar[type[CDSStream]]
 
     def _init_stream(self):
         name = param_name(self.name)
--- a/holoviews/annotators.py
+++ b/holoviews/annotators.py
@@ -19,8 +19,11 @@ from .plotting.links import (
 )
 from .streams import BoxEdit, CDSStream, CurveEdit, PointDraw, PolyDraw, PolyEdit, Selection1D
 
+NoNone = False
 if t.TYPE_CHECKING:
     from collections.abc import Callable
+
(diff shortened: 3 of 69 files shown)


.github/workflows/test.yaml
```diff
@@ -201,9 +201,28 @@ jobs:
...
+      fail-fast: false
+      matrix:
+        os: ["ubuntu-latest"]
+        environment: ["type"]
+    timeout-minutes: 5
+    steps:
+      - uses: holoviz-dev/holoviz_tasks/pixi_install@v0
+        with:
+          environments: ${{ matrix.environment }}
+      - name: Test Unit
```
hoxbro: Type TY

holoviews/streams.py
```diff
@@ -640,6 +643,7 @@ def _concat(self, data):
             elif data_length > self.length:
                 data = data.iloc[-self.length :]
         elif isinstance(data, dict) and data:
+            data = t.cast("dict[str, np.ndarray]", data)
```
hoxbro, line 644: Needed because of https://docs.astral.sh/ty/reference/typing-faq/#what-is-toplistunknown-and-why-does-it-appear and numpy type not narrowing down with isinstance.

holoviews/__init__.py
```diff
@@ -199,9 +201,7 @@ def __call__(self, *args, **kwargs):
 
     # The notebook_extension is needed inside jupyterlite,
     # so the override is only done if we are not inside jupyterlite.
-    if in_jupyterlite():
-        extension.inline = False
```
hoxbro, line 203: This is the default now so no need to do this.

holoviews/streams.py
```diff
@@ -509,7 +513,7 @@ class Pipe(Stream):
         doc="Arbitrary data being streamed to a DynamicMap callback.",
     )
 
-    def __init__(self, data=None, memoize=False, **params):
+    def __init__(self, data=None, **params):
```
hoxbro, line 513: Fixes https://github.com/holoviz/holoviews/issues/6653

holoviews/annotators.py
```diff
@@ -17,10 +17,15 @@
...
+from .streams import BoxEdit, CDSStream, CurveEdit, PointDraw, PolyDraw, PolyEdit, Selection1D
 
+if t.TYPE_CHECKING:
+    from collections.abc import Callable
 
-def preprocess(function, current=None):
+    _T = t.TypeVar("_T")
+
+
+def preprocess(function):
```
hoxbro, line 31: Current was never used. I think it is OK to remove it.

holoviews/core/util/__init__.py
```diff
@@ -2351,25 +2346,24 @@ def dt_to_int(value, time_unit="us"):
...
         try:
-            value = np.datetime64(value, "ns")
+            value = value.astype("datetime64[ns]")
             tscale = np.timedelta64(1, time_unit) / np.timedelta64(1, "ns")
             return int(value.tolist() / tscale)
-        except Exception:
+        except Exception as e:
             # If it can't handle ns precision fall back to datetime
             value = value.tolist()
+            if isinstance(value, int) or value is None:
```
hoxbro, line 2365: This would previously fail when running `.total_seconds()`

Image

See https://numpy.org/doc/stable/reference/arrays.datetime.html#converting-datetime-and-timedelta-to-python-object

holoviews/core/util/dependencies.py
```diff
@@ -60,17 +68,16 @@ def __init__(self, module_name, package_name=None, *, bool_use_sys_modules=False
             Name of the package, this is the named used for installing the package, e.g. `pip install pillow`.
             Used for the __version__ if the module is not imported.
             If not set uses the module_name.
-        bool_use_sys_modules: bool, optional, default False
+        use_sys_modules: bool, optional, default False
```
hoxbro: R

holoviews/core/options.py
```diff
@@ -683,21 +690,16 @@ def _merge_options(self, identifier, group_name, options):
...
+            return self.groups[key]
+        return super().__getitem__(key)
 
     def __getattr__(self, identifier):
         """Allows creating sub OptionTree instances using attribute
         access, inheriting the group options.
 
         """
-        try:
-            return super(AttrTree, self).__getattr__(identifier)
```
hoxbro, line 697: As far as I can see this would always raise `AttributeError` but correct me if I'm wrong.

holoviews/core/options.py
```diff
@@ -990,14 +994,8 @@ def collapse_element(cls, overlay, ranges=None, mode="data", backend=None):
             result = applicable_op.apply(sliced, ranges, backend)
             if applicable_op.group:
                 result = result.relabel(group=applicable_op.group)
-            if isinstance(overlay, Overlay):
-                result = [result]
-            else:
-                result = list(zip(sliced.keys(), [result], strict=None))
```
hoxbro, line 996: This branch was never hit. And we never account for the input to be a tuple in the list comprehension below, also makes the code a bit cleaner.

holoviews/core/options.py
```diff
@@ -486,8 +492,9 @@ def __init__(
             error = OptionError(invalid_kw, allowed_keywords, group_name=key)
             StoreOptions.record_skipped_option(error)
         if invalid_kws and self.warn_on_skip:
-            self.param.warning(
```
hoxbro, line 489: Not a `Parameterized` class so no `.param`

holoviews/core/util/__init__.py
```diff
@@ -1606,9 +1611,7 @@ def layer_sort(hmap):
         if len(okeys) == 1 and okeys[0] not in orderings:
             orderings[okeys[0]] = []
         else:
-            orderings.update(
-                {k: [] if k == v else [v] for k, v in zip(okeys[1:], okeys, strict=None)}
-            )
+            orderings.update({k: [] if k == v else [v] for v, k in itertools.pairwise(okeys)})
```
hoxbro, line 1621: swapped the order of `k, v` to make this work.

hoxbro (commented):
Most of the changes are related to tree things:

- Liskov substitution principle
- `zip(..., strict=None)` which I set to avoid enable linting rule on new code, but is flagged with type checking. Have tried my best to set `strict=True` but if in doubt have used False
- Optional imports both in the codebase and in tests, where we now define them in one file to avoid a lot of the boilerplate needed for type checker. For codebase it is in `holoviews/core/util/dependencies.py` and for tests it is `holoviews/tests/utils.py`.

holoviews/core/dimension.py
```diff
@@ -597,7 +603,29 @@ def id(self, opts_id):
             ref = weakref.ref(self, partial(cleanup_custom_options, opts_id))
             Store._weakrefs[opts_id].append(ref)
 
-    def clone(self, data=None, shared_data=True, new_type=None, link=True, *args, **overrides):
+    @t.overload
+    def clone(
+        self,
+        data=None,
+        shared_data=True,
```
hoxbro: False

holoviews/util/transform.py
```diff
@@ -1116,16 +1111,19 @@ def _compute_data(self, data, drop_index, compute):
...
-    def _coerce(self, dataset):
-        if self.interface_applies(dataset, coerce=False):
-            return dataset
-        pandas_interfaces = param.concrete_descendents(PandasInterface)
+    def _coerce(self, data):
+        if self.interface_applies(data, coerce=False):
+            return data
+        pandas_interfaces = t.cast(
+            "dict[str, type[PandasInterface]]",
+            param.concrete_descendents(parentclass=PandasInterface),
```
hoxbro, line 1119: Can be removed when this is merged: https://github.com/holoviz/param/pull/1137

holoviews/core/data/interface.py
```diff
@@ -130,7 +131,7 @@ def _perform_getitem(cls, dataset, indices):
 
 
 class Interface(param.Parameterized):
-    interfaces = {}
```
philippjfr, line 133: Technically `ClassVar` right? Not sure how much it matters.

holoviews/core/util/__init__.py
```diff
@@ -1686,28 +1702,16 @@ def is_series(data):
     return isinstance(data, tuple(types))
 
 
-def is_dask_array(data):
-    if "dask.array" in sys.modules:
-        import dask.array as da
-
-        return isinstance(data, da.Array)
-    return False
+def is_dask_array(data) -> bool:
```
philippjfr, line 1705: Move these to `utils/dependencies.py`? Not strong opinion though.

holoviews/core/util/dependencies.py
```diff
@@ -34,7 +38,7 @@ def _get_version(package_name):
         return "0.0.0"
 
 
-def _no_import_version(package_name) -> tuple[int, int, int]:
+def _no_import_version(package_name) -> tuple[int, ...]:
```
philippjfr, line 41: Wasn't that correct, tuple of 3 ints?

holoviews/core/options.py
philippjfr: Hmm, not sure about this but it would take a lot for me to understand the context again.

holoviews/core/options.py
philippjfr: Why never hit? I believe you but seems odd.

holoviews/util/transform.py
```diff
@@ -491,9 +498,6 @@ def __add__(self, other):
     def __and__(self, other):
         return type(self)(self, operator.and_, other)
 
-    def __div__(self, other):
```
philippjfr, line 494: What's up with the removals?

philippjfr (commented):
Left some comments, nothing major.

holoviews/core/data/interface.py
hoxbro: Correct. Have added it. But I will enable it for all places in the future when all files are typed.

holoviews/core/util/__init__.py
hoxbro: There are multiple of these checks. So will keep them for now.

holoviews/core/util/dependencies.py
hoxbro: The slice only does up to 3 items, and for most cases it should be so, but nothing preventing the version to only be two ints.

holoviews/core/options.py
hoxbro: `AttrTree` is a normal class, if that helps. And `OptionTree` only inherit from `AttrTree`. I think at some point this was a parameterized class where it make sense.


https://github.com/holoviz/holoviews/blob/32dc0207a434c6c553172851b2ad407cf7d39803/holoviews/core/tree.py#L7

https://github.com/holoviz/holoviews/blob/32dc0207a434c6c553172851b2ad407cf7d39803/holoviews/core/options.py#L600

holoviews/core/options.py
hoxbro: The code just below this line was:

``` python
processed[applicable_op] += [
    el for r in result for el in r.traverse(lambda x: x, [Element])
]
```

And the output of this line is a `list[tuple[Any, Any]]`. This means if the list comprehension was ever ran it would error on `r.traverse` because `tuple` does not have traverse in it.

holoviews/util/transform.py
hoxbro: The functions don't exists on operator. I think this is left over from Python 2.


Image

philippjfr: Ah right, replaced with truediv and floordiv.

holoviews/core/options.py
philippjfr: Hmm, but wouldn't methods on the class be looked up in this try/except?

hoxbro: Isn't `__getattr__` only used if something is missing? Defined methods are not missing. But the way this is set up with `super(AttrTree, self)`, always return `AttributeError`.  

Looking at this, it does also seem like we could remove the `try/except` from `AttrTree` or _at least_  need to update something.

``` python
class AttrTree:
    def __getattr__(self, attr):
        try:
            super().__getattr__(attr)
            print("success parent", attr)
        except AttributeError as e:
            print("error parent", e)


class OptionTree1(AttrTree):
    def find(self):
        print("FIND")

    def __getattr__(self, attr):
        print("get", attr)
        try:
            super().__getattr__(attr)
            print("success child 1", attr)
        except AttributeError as e:
            print("error child 1", e)

class OptionTree2(AttrTree):
    def find(self):
        print("FIND")

    def __getattr__(self, attr):
        print("get", attr)
        try:
            super(AttrTree, self).__getattr__(attr)
            print("success child 2", attr)
        except AttributeError as e:
            print("error child 2", e)


ot1 = OptionTree1()
ot1.find()
ot1.no_find

ot2 = OptionTree2()
ot2.find()
ot2.no_find

```
```
FIND
get no_find
error parent 'super' object has no attribute '__getattr__'
success child 1 no_find
FIND
get no_find
error child 2 'super' object has no attribute '__getattr__'
```

philippjfr (approved):

PR #6853, opened 2026-04-22
fix: Avoid trying to match dimension by name

philippjfr:
This is an attempted fix for https://github.com/holoviz/holoviews/issues/6799. The issue as described there (imo) misidentifies the actual problem. The issue is subtle and changing the source of truth for dimension identity (in https://github.com/holoviz/holoviews/issues/6260) made things even more subtle. In applying this change we did not carefully update all codepaths. Before describing the fix, just to recap the concepts (post #6260):

- `Dimension.name`: The name of the underlying data dimension, e.g. the column/index name in pandas
- `Dimension.label`: The user visible label and the identity of the dimension, e.g. to allow multiple columns to appear like one “dimension”

This does not mean that the label is now the "source of truth" for all purposes, it's simply the thing that uniquely identifies a dimension. In more complex usecases, such as accessors in (hv-)Anndata, where comparison is very subtle and no attempt should be made to try to match a dimension based on its name, our handling was not careful enough. Particularly inside the interfaces, where the name is used to match against the underlying data dimension we sometimes tried to compare dimension names directly with Dimension objects, causing problems. To resolve this, this PR instead ensures that we avoid name to dimension comparisons.

Additionally, in datashader operations, we accept aggregator objects that can reference a data column (e.g. `ds.mean('col')`). Whether these should match the dimension by `name` or `label` is ambiguous, since in datashader they refer to the underlying data column or coordinate name (which is the `Dimension.name`) but for the purposes of HoloViews should generally refer to the label (e.g. imagine datashading an `NdOverlay` containing Elements with wide data where each element refers to a different underlying column but we want to aggregate across columns, which are referred to by a shared label, e.g. stock ticker names as column names and "Price" as the label). To fix this particular issue we re-map the aggregator column (which might refer to the label) to the column name (unless the data is wide, in which case we remap the data from name -> label).


holoviews/core/data/grid.py
```diff
@@ -333,7 +333,11 @@ def canonicalize(cls, dataset, data, data_coords=None, virtual_coords=None):
             for name in data_coords
             if isinstance(cls.coords(dataset, name), get_array_types())
         ]
-        dropped = [dims.index(d) for d in dims if d not in dataset.kdims + virtual_coords]
+        dropped = [
+            dims.index(d)
+            for d in dims
+            if not any(d == dimension_name(dim) for dim in dataset.kdims + virtual_coords)
+        ]
```
hoxbro:
```suggestion
        dim_names = [dimension_name(dim) for dim in dataset.kdims + virtual_coords]
        dropped = [dims.index(d) for d in dims if not any(d == dim for dim in dim_names)]
```

For performance.

holoviews/operation/datashader.py
```diff
@@ -157,6 +157,71 @@ class AggregationOperation(ResampleOperation2D):
         "count_cat": rd.count_cat,
     }
 
+    @classmethod
+    def _overlay_wide_mapping(cls, element, elements):
```
hoxbro:
```suggestion
    @staticmethod
    def _overlay_wide_mapping(element, elements):
```

holoviews/operation/datashader.py
```diff
@@ -157,6 +157,71 @@ class AggregationOperation(ResampleOperation2D):
...
+        if not isinstance(element, NdOverlay):
+            return False, []
+        ydims = [el.dimensions()[1] for el in elements]
+        is_wide = len({yd.label for yd in ydims}) == 1 and len({yd.name for yd in ydims}) == len(
+            element
+        )
+        return is_wide, ydims
+
+    @classmethod
+    def _resolve_agg_column_name(cls, element, inner_element, column, is_wide, ydims):
```
hoxbro:
```suggestion
    @staticmethod
    def _resolve_agg_column_name(element, inner_element, column, is_wide, ydims):
```

holoviews/operation/datashader.py
```diff
@@ -419,19 +477,38 @@ def get_agg_data(cls, obj, category=None):
...
+                if is_ndoverlay:
                     df = df.assign(**dict(zip(obj.dimensions("key", True), key, strict=None)))
+                    y_dim = el.get_dimension(y)
+                    ydims.append(y_dim)
                 paths.append(df)
+
+            # If data is in wide format we align the column names on the Dimension.label
+            is_wide = len({yd.label for yd in ydims}) == 1 and len(
+                {yd.name for yd in ydims}
+            ) == len(obj)
```
hoxbro: Why are we not using `_overlay_wide_mapping` here?

holoviews/operation/datashader.py
```diff
@@ -419,19 +477,38 @@ def get_agg_data(cls, obj, category=None):
...
+            if is_ndoverlay and is_wide:
+                ydim = next(iter(ydims))
+                paths = [
+                    df.rename(columns={yd.name: yd.label})
+                    for yd, path in zip(ydims, paths, strict=True)
+                ]
+                is_wide = True
```
hoxbro:
```suggestion
```

holoviews/operation/datashader.py
```diff
@@ -419,19 +477,38 @@ def get_agg_data(cls, obj, category=None):
...
+                    y_dim = el.get_dimension(y)
+                    ydims.append(y_dim)
                 paths.append(df)
+
+            # If data is in wide format we align the column names on the Dimension.label
+            is_wide = len({yd.label for yd in ydims}) == 1 and len(
+                {yd.name for yd in ydims}
+            ) == len(obj)
+            if is_ndoverlay and is_wide:
+                ydim = next(iter(ydims))
```
hoxbro:
```suggestion
                ydim = ydims[0]
```

hoxbro (commented):
Running hv-anndata tests gives error on these tests:

``` 
tests/plotting/test_manifold.py::test_create_manifoldmap_plot_datashading[categorical]
tests/plotting/test_manifold.py::test_create_manifoldmap_plot_datashading[continuous]
tests/plotting/test_scanpy.py::test_basic[bokeh-violin-obsm-color]
tests/plotting/test_scanpy.py::test_basic[matplotlib-violin-obsm-color]
```

holoviews/operation/datashader.py
```diff
@@ -419,19 +477,38 @@ def get_agg_data(cls, obj, category=None):
         kdims = list(obj.kdims)
         vdims = list(obj.vdims)
         dims = obj.dimensions()[:2]
+        is_wide = False
```
philippjfr:
```suggestion
```

hoxbro:
Apply this patch in HoloViews and the following in hv-anndata, makes the tests in hv-anndata pass without warnings for me, but makes the following HoloViews test fail `pytest holoviews/tests/operation/test_datashader.py::test_wide_data_lines[aggregator1]`
``` patch
diff --git a/holoviews/operation/datashader.py b/holoviews/operation/datashader.py
index 7b483d700..0cedb481a 100644
--- a/holoviews/operation/datashader.py
+++ b/holoviews/operation/datashader.py
@@ -301,7 +301,7 @@ class AggregationOperation(ResampleOperation2D):
         elif agg_name == "Summary":
             vdims = list(agg_fn.keys)
         elif column:
-            dims = [d for d in element.dimensions("ranges") if d == column]
+            dims = [d for d in element.dimensions("ranges") if d.name == column]
             if not dims:
                 raise ValueError(
                     f"Aggregation column '{column}' not found on '{element}' element. "
```


``` patch
diff --git a/src/hv_anndata/interface.py b/src/hv_anndata/interface.py
index 9e755f4..4e59ac5 100644
--- a/src/hv_anndata/interface.py
+++ b/src/hv_anndata/interface.py
@@ -376,7 +376,7 @@ class AnnDataInterface(hv.core.Interface):
         for k in dimensions:
             dim = cls._dim(dataset, k)
             cls.validate_selection_dim(dim, "group")
-            values[k] = unique_iterator(adata[dim])
+            values[dim] = unique_iterator(adata[dim])
 
         # Get dimensions information
         dimensions = [dataset.get_dimension(d) for d in dimensions]
diff --git a/tests/plotting/test_manifold.py b/tests/plotting/test_manifold.py
index 1864f8d..a7f1b91 100644
--- a/tests/plotting/test_manifold.py
+++ b/tests/plotting/test_manifold.py
@@ -107,11 +107,11 @@ def test_create_manifoldmap_plot_datashading(
     dop = el.pipeline.find(dynspread, skip_nonlinked=False)
     if color_kind == "categorical":
         assert rop.name.startswith("datashade")
-        assert rop.aggregator.cat_column == by
+        assert rop.aggregator.cat_column == by.name
     elif color_kind == "continuous":
         assert rop.name.startswith("rasterize")
         assert type(rop.aggregator).__name__ == "mean"
-        assert rop.aggregator.column == by
+        assert rop.aggregator.column == by.name
     assert dop.name.startswith("dynspread")
     assert dop.threshold == pytest.approx(0.5)
 
```

image

PR #6875, opened 2026-05-07
fix: Memory leak in `Stream._subscribers`

### LINKED ISSUE #8580: Memory not freed on session closure in app with some holoviews Stream subscribers

<details>

```plaintext
panel 1.8.10
holoviews 1.22.1
bokeh 3.9.0
python 3.14.4
```

</details>

When a session is destroyed, I expect the RAM of (non-cached) objects tied to this session to be freed. However, there are instances where RAM mobilised by user sessions is not fully released after session closure.

After some digging I suspect cyclic references may be involved and prevent garbage collection.

Coming from my real-world app, the MRE I can provide involves holoviews streams, where a subscriber has an argument that is the a Viewer object creating the plot. This typically occurs when the subscriber is a method of said Viewer. There may be other, simpler ways to trigger this behaviour, probably without involving holoviews. It's possible that it can even be reproduced in pure bokeh, and I'll move the issue upstream if need be.

Now, if nothing can be done in panel or bokeh, maybe it should be considered a usage issue, and users should be warned against doing this in the docs.

```python
import gc
import sys

import holoviews as hv
import numpy as np
import pandas as pd
import panel as pn


class App(pn.viewable.Viewer):
    def __init__(self):
        super().__init__()
        self.load = pn.widgets.Button(name="Load plot", on_click=self.update_plot)
        self.log_memory_usage = pn.widgets.Button(
            name="Log large objects", on_click=log_large_objects
        )
        self.view = pn.Column(pn.Row(self.load, self.log_memory_usage))

    def __panel__(self):
        return self.view

    def update_plot(self, event):
        data = pd.DataFrame(
            {
                "x": np.random.uniform(-1, 1, 5000000),
                "y": np.random.uniform(-1, 1, 5000000),
                "var1": np.random.uniform(0, 10, 5000000),
                "var2": np.random.uniform(0, 10, 5000000),
            }
        )
        plot = hv.Points(data)
        xy_range = hv.streams.RangeXY()
        xy_range.source = plot
        xy_range.add_subscriber(self.dummy_subscriber)
        self.view.append(plot)

    def dummy_subscriber(self, x_range, y_range):
        pass


def log_large_objects(event):
    gc.collect()
    all_objects = gc.get_objects()
    size_threshold = 1e7  # 10 MB
    for obj in all_objects:
        size = sys.getsizeof(obj)
        if size < size_threshold:
            continue

        obj_info = {
            "size": size,
            "type": type(obj).__name__,
            "id": id(obj),
            "repr": repr(obj)[:100],
            "hex_id": hex(id(obj)),
        }

        print(f"  {obj_info['type']}: {obj_info['size'] / 1e6:.2f} MB")
        print(f"      ID: {obj_info['hex_id']}")
        print(f"      Preview: {obj_info['repr']}")


if pn.state.served:
    App().servable()
```

Run the server with memory logging (and reduced unused session lifetime to speed up the test):
```bash
python -m panel serve test_ram.py --mem-log-frequency 5000 --check-unused-sessions 1000 --unused-session-lifetime 1000
```

You'll notice that the memory usage does not drop as much as expected after closing the session, and the DataFrame object outlives the session. As a result, RAM is accumulating across successive sessions.

Removing the subscriber, or replacing the method subscriber with a standalone function (without referencing the _App_ instance), results in significantly better memory cleanup.

https://github.com/user-attachments/assets/4dd0795f-8de7-405c-ba4a-73ae0bcd9afd

hoxbro:
## Description
Fixes https://github.com/holoviz/panel/issues/8580

Bound-method subscribers in `Stream._subscribers` created a reference cycle (stream → subscriber → instance → stream) that prevented GC from collecting session objects after Panel session destruction. Wrapping bound methods in `_WeakMethodSubscriber` (a `weakref.WeakMethod` wrapper) breaks the cycle and dead refs are pruned during `cleanup`.

Also fixing a `unlink` exception when source had been GC.

## How Has This Been Tested?

CI

## AI Disclosure

Tools and Models: claude:claude-sonnet-4-6

Used it to find the problem.

## Checklist


holoviews/plotting/links.py
```diff
@@ -83,8 +83,11 @@ def link(self):
 
     def unlink(self):
         """Unregisters the Link"""
-        links = self.registry.get(self.source)
-        if self in links:
+        source = self.source
+        if source is None:
+            return
+        links = self.registry.get(source)
+        if links and self in links:
```
hoxbro, line 90: Unrelated to the original problem, just something AI flagged.

holoviews/streams.py
```diff
@@ -53,6 +54,35 @@ class _SkipTrigger:
...
+    def __bool__(self) -> bool:
+        method = self._ref()
+        return bool(method is not None and not util.is_param_method(method))
+
+    def __call__(self, *args, **kwargs):
+        method = self._ref()
+        if method is not None:
+            return method(*args, **kwargs)
+
+    @staticmethod
```
hoxbro: Move up

philippjfr (approved):

PR #6029, opened 2023-12-12
Add 'norm' in init_artists in holoviews/plotting/mpl/raster.py,

JulianGiles:
Added handling of 'norm' kwarg to init_artists in the same way is already present in /holoviews/plotting/mpl/element.py, in order to pass norm when plotting a quadmesh with matplotlib backend. This allows to pass a custom norm and colormap to be able to plot a colobar with discrete, uneven intervals correctly. Related to #5966

PR #6889, opened 2026-05-20
fix: norm and vmin/vmax in raster.py

### LINKED ISSUE #5966: Plotting with a discrete nonlinear colorbar assigns values to colors incorrectly
Name                    Version                   Build  Channel
holoviews                 1.17.1             pyhd8ed1ab_0    conda-forge
hvplot                    0.8.4              pyhd8ed1ab_1    conda-forge
bokeh                     3.2.2              pyhd8ed1ab_0    conda-forge
xarray                    2023.9.0           pyhd8ed1ab_0    conda-forge
python                    3.11.5          hab00c5b_0_cpython    conda-forge
spyder                    5.4.5           py311h38be061_0    conda-forge
spyder-kernels            2.4.4           unix_pyh707e725_0    conda-forge

I am trying to plot a QuadMesh with a discrete nonlinear colorbar. I have found help from several posts in the discourse and GitHub but mainly [this one](https://discourse.holoviz.org/t/custom-discrete-colormaps/2183/15). When plotting by passing a discrete list of colors (cmap) and levels (color_levels), a plot is produced which looks correct at first glance. But, on closer inspection or when the data is more challenging (more extreme values or very irregular color ranges) some values on the verge of a color level are incorrectly assign to a different color.

```
import xarray as xr
import hvplot.xarray
from bokeh.models import CategoricalColorMapper, ColorBar

COLORS = ["magenta", "green", "yellow", "orange", "red", "purple"]
BOUNDS = [0, 230, 250, 270, 300, 320, 400]

# define a function to plot a colorbar with equally-sized color bins
def cbar_hook(hv_plot, _):
    plot = hv_plot.handles["plot"]
    factors = [f"{BOUNDS[i]} - {BOUNDS[i + 1]}" for i in range(len(COLORS))]
    mapper = CategoricalColorMapper(
        palette=COLORS,
        factors=factors,
    )
    color_bar = ColorBar(color_mapper=mapper)
    plot.right[0] = color_bar


ds = xr.tutorial.open_dataset("air_temperature")
ds.hvplot.quadmesh("lon", "lat").opts(
    cmap=COLORS,
    color_levels=BOUNDS,
    clim=(BOUNDS[0], BOUNDS[-1]),
    hooks=[cbar_hook],
)

```
In this example, you can also change the extreme values in BOUNDS to be more extreme and the problem is more notorious.

image
See example of a value that should be colored yellow but it is green.

- [ x] I may be interested in making a pull request to address this

JulianGiles:
Fixing buggy code for when norm is passed to quadmesh using the matplotlib backend. Passing norm should make vmin and vmax equal to None BEFORE passing the plot_kwargs to pcolormesh. This is the proper fix that I tried to implement some years ago in #6029, apologies for not doing it properly in the first place. sigh.

Fixes #5966

### DIFF
--- a/holoviews/plotting/mpl/raster.py
+++ b/holoviews/plotting/mpl/raster.py
@@ -216,11 +216,11 @@ class QuadMeshPlot(ColorbarPlot):
 
     def init_artists(self, ax, plot_args, plot_kwargs):
         locs = plot_kwargs.pop("locs", None)
-        artist = ax.pcolormesh(*plot_args, **plot_kwargs)
-        colorbar = self.handles.get("cbar")
         if "norm" in plot_kwargs:  # vmin/vmax should now be exclusively in norm
             plot_kwargs.pop("vmin", None)
             plot_kwargs.pop("vmax", None)
+        artist = ax.pcolormesh(*plot_args, **plot_kwargs)
+        colorbar = self.handles.get("cbar")
         if colorbar and MPL_VERSION < (3, 1, 0):
             colorbar.set_norm(artist.norm)
             if hasattr(colorbar, "set_array"):
--- a/holoviews/tests/plotting/matplotlib/test_quadmeshplot.py
+++ b/holoviews/tests/plotting/matplotlib/test_quadmeshplot.py
@@ -62,7 +62,7 @@ class TestQuadMeshPlot(TestMPLPlot):
     def test_quadmesh_with_norm(self):
         arr = np.array([[0, 1, 2], [3, 4, 5]])
 
-        qmesh = hv.QuadMesh(hv.Image(arr)).opts(norm=Normalize())
+        qmesh = hv.QuadMesh(hv.Image(arr)).opts(norm=Normalize(), cmap="viridis")
 
         # Before PR 6889, getting the plot would raise a ValueError from
         # matplotlib because vmin and vmax were passed alongside the norm.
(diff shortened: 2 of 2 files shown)


hoxbro:
Failing test are unrelated and should be fixed by: https://github.com/holoviz/holoviews/pull/6890

Is it possible to add a simple test for this?

JulianGiles:
I've never written tests before, does something like this make sense:

```
def test_quadmesh_with_norm(self):
        from matplotlib.colors import Normalize
        
        arr = np.array([[0, 1, 2], [3, 4, 5]])
        
        qmesh = hv.QuadMesh(hv.Image(arr)).opts(norm=Normalize())
        
        # Before PR 6889, getting the plot would raise a ValueError from 
        # matplotlib because vmin and vmax were passed alongside the norm.
        plot = mpl_renderer.get_plot(qmesh)
        
        # Verify the plot handles were created and the norm is correctly applied
        artist = plot.handles["artist"]
        self.assertIsInstance(artist.norm, Normalize)
```

hoxbro:
> I've never written tests before, does something like this make sense:

That looks good to me. One thing is to replace `self.assertIsInstance(artist.norm, Normalize)` with `assert isinstance(artist.norm, Normalize)`. Also I'm pretty sure you don't need to inline the import. 

If you update the branch and add the test, the CI should not fail.

hoxbro:
LGTM, thank you :+1:

PR #6915, opened 2026-06-12
fix: Allow image coordinate irregularity equal to the tolerance

### LINKED ISSUE #6914: Image with a regularly sampled date–time coordinate warns of irregular sampling when the spacing is less than 1000 date–time units
<details>

```plaintext
  Name                      Version  Build         Channel
----------------------------------------------------------------
  bokeh                     3.9.1    pyhd8ed1ab_0  conda-forge
  holoviews                 1.22.1   pyhd8ed1ab_0  conda-forge
```
</details>

While creating images that have a date–time coordinate, I've noticed that *Holoviews* sometimes warns of uneven sampling even when the coordinate is perfectly regular (that is, it has exactly one step size). If the coordinate array is uniformly spaced, *Holoviews* should not issue the warning.

Below, we define data for a 2-D image with alternative date–time coordinates for the *x* dimension. One coordinate steps by 999 in its internal representation; the other, by 1000.

```python
import holoviews as hv
import numpy as np

hv.extension("bokeh")

x_0999 = np.array(
    ["2026-01-01 12:00:00.000", "2026-01-01 12:00:00.999"], dtype="datetime64[ms]"
)
x_1000 = np.array(
    ["2026-01-01 12:00:00.000", "2026-01-01 12:00:01.000"], dtype="datetime64[ms]"
)
y = np.arange(2)
z = np.arange(4).reshape((2, 2))
```

Constructing an `Image` with the 999-step coordinate produces the warning below:

```python
image_0999 = hv.Image((x_0999, y, z))
```

> WARNING:param.Image00855: Image dimension x is  not evenly sampled to relative tolerance of 0.001. Please use the QuadMesh element for irregularly sampled data or set a higher tolerance on hv.config.image_rtol or the rtol parameter in the Image constructor.

Because the *x* coordinate is regular, I expect *Holoviews* to construct the image without the warning.

The image appears as follows:

```python
image_0999
```

> Image

In contrast, using the 1000-step coordinate produces no warning:

```python
image_1000 = hv.Image((x_1000, y, z))
image_1000
```

> Image

- [X] I may be interested in making a pull request to address this

stanwest:
## Description

The change prevents an erroneous warning such as the following for regular date-time coordinates whose minimum step has coarse granularity in the internal date-time representation:

> WARNING:param.Image00855: Image dimension x is not evenly sampled to relative tolerance of 0.001. Please use the QuadMesh element for irregularly sampled data or set a higher tolerance on hv.config.image_rtol or the rtol parameter in the Image constructor.

For such a step, the right-hand side of the inequality in `holoviews.core.util.validate_regular_sampling` can evaluate to a time delta of zero (depending on how the time-delta object rounds when multiplied), and the former `<` inequality cannot be satisfied. Also, the change conforms the function to its docstring ("at most") and to the documentation of `Config.image_rtol` ("maximal allowable sampling difference").

Fixes #6914.

## Checklist

### DIFF
--- a/holoviews/tests/element/test_image.py
+++ b/holoviews/tests/element/test_image.py
@@ -118,8 +118,8 @@ class TestImage(LoggingComparison):
         Constructing an Image with an evenly spaced date-time coordinate does not warn
         about irregular sampling, even when the ratio of the date-time granularity to
         the coordinate step is greater than the default `image_rtol` (or twice
-        `image_rtol` for `datetime.datetime` types because of how they round when
-        multiplied by a `float`).
+        `image_rtol` for `dt.datetime` types because of how `dt.timedelta` instances
+        round when multiplied by a `float`).
         """
         y = np.arange(2)
         hv.Image((x, y, np.zeros((len(y), len(x)))))
--- a/holoviews/tests/element/test_image.py
+++ b/holoviews/tests/element/test_image.py
@@ -4,6 +4,8 @@ Unit tests of Image elements
 
 from __future__ import annotations
 
+import datetime as dt
+
 import numpy as np
 import pytest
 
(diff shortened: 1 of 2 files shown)


holoviews/tests/element/test_image.py
```diff
@@ -74,6 +76,55 @@ def test_image_rtol_config(self):
...
+                for unit in ["D", "h", "m", "s", "ms", "us"]
+            ),
+        ],
+    )
+    def test_image_datetime_rtol_failure(self, x):
+        """
+        Constructing an Image with a date-time coordinate warns about irregular sampling
+        if, approximately speaking, one less than the ratio of the max and min steps
+        exceeds the default `image_rtol`.
+        """
```
hoxbro, line 95:
```suggestion
```
Docstrings is not needed for tests.

holoviews/tests/element/test_image.py
```diff
@@ -74,6 +76,55 @@ def test_image_rtol_config(self):
...
+        ],
+    )
+    def test_image_datetime_rtol_granularity(self, x):
+        """
+        Constructing an Image with an evenly spaced date-time coordinate does not warn
+        about irregular sampling, even when the ratio of the date-time granularity to
+        the coordinate step is greater than the default `image_rtol` (or twice
+        `image_rtol` for `datetime.datetime` types because of how they round when
+        multiplied by a `float`).
+        """
```
hoxbro, line 123:
```suggestion
```

holoviews/tests/element/test_image.py
```diff
@@ -74,6 +76,55 @@ def test_image_rtol_config(self):
...
+            "set a higher tolerance on hv.config.image_rtol or the rtol parameter in "
+            "the Image constructor.",
+        )
+
+    @pytest.mark.parametrize(
+        "x",
+        [
+            # |granularity / step| = 1/400 = 0.0025
+            [dt.datetime(2017, 1, 1, 0, 0, 0, ms) for ms in [0, 400]],
+            # |granularity / step| = 1/800 = 0.00125
```
hoxbro, line 109: Why did you choose two different ratios?

hoxbro (commented):
Just want to verify with you that the tests fails without the suggested change.

Also have you been used AI for these changes?

stanwest:
> Just want to verify with you that the tests fails without the suggested change.

One of the added tests, namely `test_image_datetime_rtol_granularity`, covers the buggy behavior and does fail without the change. (The first of the two commits decorates the test with `pytest.mark.xfail` so that the test suite passes, then the second commit, which contains the fix, removes the decorator.) The other added test covers the desirable warning behavior and passes in both commits.

> Also have you been used AI for these changes?

I used no AI.

holoviews/tests/element/test_image.py
stanwest: I understand. I see the docstring as a natural place to explain the intent of a non-obvious test. Here, in particular, it relates the test to `image_rtol`, a relationship that isn't apparent from the test's parameters, which are declared independently of `image_rtol`. Would a comment be preferred over a docstring?

holoviews/tests/element/test_image.py
stanwest: Because `datetime.timedelta` and `np.timedelta64` instances round differently when multiplied by a `float` such as `image_rtol`, the range of coordinate steps that will cause the false warning differs for the two types. (A `datetime.timedelta` instance rounds to the nearest microseconds, whereas a `np.timedelta64` truncates toward zero.) Specifically, with `image_rtol` = 0.001, the false warning occurs for `datetime.datetime` steps of less than 500 microseconds and for `np.datetime64` steps of less than 1000 units. For each type, I chose a step that is a little smaller than the largest sufficient step for that type.

PR #6922, opened 2026-06-18
fix: Seed range streams on initial draw of overlay plots

### LINKED ISSUE #6921: Overlaid Plot Not Fully Rendered On Initial Load
<details>

```plaintext
holoviews==1.22.1
datashader==0.18.2
bokeh==3.8.2
jupyterlab==4.5.2
Python 3.14.2
```
</details>

#### Here is example code to demonstrate the issue

```python
import pandas as pd
import numpy as np
import holoviews as hv
from holoviews import opts
from holoviews.operation.datashader import datashade, dynspread, rasterize, shade, spread
hv.extension("bokeh")

num_rows = 200_000

timestamps = pd.date_range(
  start="2026-01-01",
  periods=num_rows,
  freq="us"
)

values = np.zeros(num_rows, dtype=np.int32)

bursts = [
  (20_000, 50_000),
  (80_000, 120_000),
  (150_000, 180_000)
]

for start, end in bursts:
  values[start:end] = np.random.randint(
    2000,
    2501,
    end - start
  )

df1 = pd.DataFrame({
  "Time": timestamps,
  "Value": values
})

for start, end in bursts:
  values[start:end] = np.random.randint(
    2200,
    2301,
    end - start
  )

df2 = pd.DataFrame({
  "Time": timestamps,
  "Value": values
})

s1 = hv.Scatter(df1, "Time", "Value", label="s1")
s2 = hv.Scatter(df2, "Time", "Value", label="s2")

overlay = s1 * s2
raster = rasterize(overlay)
spread = dynspread(raster, max_px=2, threshold=1)
spread.opts(width=1400, height=400, padding=(0.1, 0.1))
```

A fix to this issue would be greatly appreciated! Thank you!

#### Initial Render
Image

#### Reset Render (when clicking the 'Reset' button in the bokeh toolbar)
Image

hoxbro:
## Description

Fixes #6921

An overlay of rasterized elements rendered incompletely on initial load and only corrected itself after a range event (e.g. clicking Reset).

This now mirror the `ElementPlot` logic in `GenericOverlayPlot` to get padded extent.

https://github.com/holoviz/holoviews/blob/d2ae73a3e299669cf9856bb27b2b81fead658acc/holoviews/plotting/plot.py#L1707-L1720

## AI Disclosure

Assisted-by: Claude Code:claude-opus-4-8

## Checklist

### DIFF
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -2309,6 +2309,27 @@ class GenericOverlayPlot(GenericElementPlot):
         if not (("multi_y" in self.param) and self.multi_y):
             y0, y1 = util.dimension_range(y0, y1, self.ylim, (None, None))
 
+        # Seed any RangeX/RangeY/RangeXY streams (e.g. those driving a
+        # datashader operation) with the padded extents on the initial draw.
+        # Without this an overlay of rasterized elements would aggregate over
+        # the unpadded data range until a range event (e.g. clicking Reset)
+        # supplies the displayed range. See ElementPlot.get_extents for the
+        # equivalent logic used for non-overlay plots.
+        if not self.drawn:
+            x_range, y_range = ((y0, y1), (x0, x1)) if self.invert_axes else ((x0, x1), (y0, y1))
+            for stream in getattr(self, "source_streams", []):
+                if isinstance(stream, RangeX):
+                    params = {"x_range": x_range}
+                elif isinstance(stream, RangeY):
+                    params = {"y_range": y_range}
+                elif isinstance(stream, RangeXY):
+                    params = {"x_range": x_range, "y_range": y_range}
+                else:
+                    continue
+                stream.update(**params)
+                if stream not in self._trigger and (self.xlim or self.ylim):
+                    self._trigger.append(stream)
+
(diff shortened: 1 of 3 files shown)


holoviews/plotting/plot.py
```diff
@@ -2309,6 +2310,22 @@ def get_extents(self, overlay, ranges, range_type="combined", dimension=None, **
         if not (("multi_y" in self.param) and self.multi_y):
             y0, y1 = util.dimension_range(y0, y1, self.ylim, (None, None))
 
+        # Should match what is done in ElementPlot.get_extents
+        if not self.drawn:
```
maximlt, line 2314: If the code needs to be exactly the same in both places, shouldn't it be refactored into a method/function?

maximlt (commented):
One comment. I tried the fix with the code shared in the issue and it solved the problem.

holoviews/plotting/plot.py
hoxbro: I like the locality of having the code here, and I don't think making it a method makes sense in this case (it's only used in two places). It would also remove flexibility if we need the two methods to diverge at some point.

maximlt (approved):

PR #6969, opened 2026-07-24
docs: Hide notebook-specific resampling settings

### LINKED ISSUE #6044: Move ResampleOperation2D settings in docs notebooks to hidden cell
ResampleOperation2D settings for datashader are docs-website specific and could confuse users. Let's move them to hidden cells.

As was used [here](https://github.com/holoviz/hvplot/pull/1205#issuecomment-1866047361), MyST-NB supports by adding the `hide-cell` tag to the cell and the config on the clean_notebook hook now ignores the tags key in the cell metadata.

Kkkakania:
## Summary

- move the `ResampleOperation2D` width and height overrides into dedicated hidden cells
- keep the user-facing setup focused on settings that users may actually want to copy
- configure `clean-notebook` to preserve cell tags so `hide-cell` survives linting

Closes #6044.

## Validation

- `uvx pre-commit run --files .pre-commit-config.yaml examples/user_guide/15-Large_Data.ipynb`
- verified both resampling cells retain the `hide-cell` tag after the clean-notebook hook

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -25,7 +25,7 @@ repos:
     rev: 2e6f506fab3cdfbe16bc18fd0d24e828db773b74 # frozen: v0.2.1
     hooks:
       - id: clean-notebook
-        args: [--strip-trailing-newlines]
+        args: [--strip-trailing-newlines, --ignore, tags]
   - repo: https://github.com/crate-ci/typos
     rev: bee27e3a4fd1ea2111cf90ab89cd076c870fce14 # frozen: v1.48.0
     hooks:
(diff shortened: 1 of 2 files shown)


hoxbro:
We should also hide the import of `ResampleOperation2D`

Kkkakania:
Done in `880d5f39`. I moved the `ResampleOperation2D` import into the first hidden setup cell alongside the notebook-only width/height overrides, so neither the import nor the internal display tuning is shown to readers. The notebook remains valid JSON and all configured pre-commit hooks pass.

hoxbro:
Thank you

PR #6972, opened 2026-07-24
fix: set axis tick rotation mode

### LINKED ISSUE #6967: Matplotlib backend: rotated x-tick labels not anchored, drift off their tick

<details>

```plaintext
holoviews   1.23.0
matplotlib  3.10.9
bokeh       3.9.1
python      3.14.6
platform    Linux-7.1.4-arch1-1-x86_64-with-glibc2.43
```

</details>

When `xrotation` is set on the Matplotlib backend, rotated x-tick labels should pivot around the point where they meet the tick (i.e. `ha="right", rotation_mode="anchor"` for positive rotation), so the label still points back at its tick.

Instead, `ElementPlot._set_axis_ticks` (`holoviews/plotting/mpl/element.py`) only calls `tick.set_rotation(rotation)`, leaving `ha="center"` and `rotation_mode="default"`, rotating each tick text box around its center.

Other parts of HoloViews' own Matplotlib backend (e.g. `graphs.py`, `sankey.py`) already set `rotation_mode="anchor"` when rotating text, so this looks like an oversight specific to `_set_axis_ticks` rather than a deliberate choice.

```python
import matplotlib
matplotlib.use("Agg")
import holoviews as hv
hv.extension("matplotlib")

bars = hv.Bars([(f"category {i}", i) for i in range(5)]).opts(xrotation=30)
fig = hv.render(bars)
ax = fig.axes[0]
for label in ax.get_xticklabels():
    print(label.get_text(), label.get_rotation(), label.get_ha(), label.get_rotation_mode())
# actual:   category 0 30.0 center default
# expected: category 0 30.0 right  anchor
```

```python
bars  # from above
```

Image

flying-sheep:
## Description
Fixes #6967 (with matplotlib 3.11, lmk if we should expand the fix for older versions)

It’s enough to just `set_rotation_mode` as described in https://matplotlib.org/stable/gallery/ticks/ticklabels_rotation.html

----

changing the rotation in the [“bars economic” example](https://holoviews.org/gallery/demos/matplotlib/bars_economic.html) to `xrotation=30` now gives this:

grafik

it’s not perfect (I think the distance to the ticks is a bit large) but now it’s readable,

## Checklist

### DIFF
--- a/holoviews/core/options.py
+++ b/holoviews/core/options.py
@@ -52,6 +52,8 @@ from .tree import AttrTree
 from .util import group_sanitizer, label_sanitizer, sanitize_identifier
 
 if t.TYPE_CHECKING:
+    from holoviews.plotting import Renderer
+
     from ..util import _BackendT
     from ..util.settings import OutputSettings
 
@@ -1193,7 +1195,7 @@ class Store:
 
     """
 
-    renderers = {}  # The set of available Renderers across all backends.
+    renderers: dict[str, Renderer] = {}  # The set of available Renderers across all backends.
 
     # A mapping from ViewableElement types to their corresponding plot
     # types grouped by the backend. Set using the register method.
--- a/holoviews/plotting/mpl/chart.py
+++ b/holoviews/plotting/mpl/chart.py
@@ -1,5 +1,7 @@
 from __future__ import annotations
 
+from typing import TYPE_CHECKING
+
 import matplotlib as mpl
 import numpy as np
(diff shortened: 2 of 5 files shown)


hoxbro:
> with matplotlib 3.11, lmk if we should expand the fix for older versions

The failing test was caused by Matplotlib version 3.10.9, so I'm fine doing it for 3.11 and above. 

Thank you for the PR.

PR #6982, opened 2026-08-05
perf: Avoid importing colorcet to resolve matplotlib and bokeh colormaps

### LINKED ISSUE #6981: First trigger of colormap import causes multiple-second lag
The first `process_cmap()` call in a process takes nearly 3 seconds with matplotlib 3.11, because it imports `colorcet` unconditionally — even for a matplotlib colormap name — and `import colorcet` is quadratically slow against matplotlib 3.11's new "Did you mean …?" error machinery.

The user-visible consequence of this is a **multi-second delay** before, e.g. `Image` plots are shown for the first time (observed in our real dashboard, time of first switch to a `Tab` with `Image` plot goes from 4.5s to 1.5s if we do a warmup as a workaround, see scipp/esslivedata#1197).

This should likely be fixed in Matplotlib, but adding a mitigation in Holoviews seems relevant as well, see below.

```
holoviews  1.23.1
colorcet   3.2.1
matplotlib 3.11.1
bokeh      3.9.2
panel      1.9.3
param      2.4.1
numpy      2.5.1
python     3.12.13
Linux-6.8.0-110-generic-x86_64-with-glibc2.41 (24 cores, idle)
```

Resolving a matplotlib colormap by name should not cost seconds. Observed:

```
first  process_cmap('viridis'): 2.728 s
second process_cmap('viridis'): 1.04 ms
```

(two runs: 2.728 s / 2.916 s for the first call)

`cProfile` of the first call — 96% of it is `difflib` building error message strings that are then discarded:

```
   ncalls  tottime  cumtime  filename:lineno(function)
        1    0.000    8.833  holoviews/plotting/util.py:1043(process_cmap)
        3    0.000    8.826  holoviews/plotting/util.py:715(_list_cmaps)
        1    0.005    8.711  colorcet/__init__.py:1(<module>)
      420    0.001    8.599  colorcet/_dependencies.py:18(register_cmap)
      844    0.001    8.595  <frozen _collections_abc>:811(__contains__)
     1024    0.001    8.594  matplotlib/cm.py:94(__getitem__)
     1362    0.003    8.593  matplotlib/_api/__init__.py:265(getitem_checked)
      840    0.007    8.590  matplotlib/_api/__init__.py:158(list_suggestion_error_msg)
      840    0.259    8.478  difflib.py:666(get_close_matches)
   133534    0.180    5.810  difflib.py:597(ratio)
```

(absolute numbers inflated ~3x by the profiler; the shape is what matters)

The chain has four links:

1. **matplotlib 3.11** routed `ColormapRegistry.__getitem__` through `_api.getitem_checked`, which on `KeyError` calls `difflib.get_close_matches(name, all_registered_names, cutoff=0.5)` to suggest a correction ([matplotlib#30001](https://github.com/matplotlib/matplotlib/pull/30001), milestone v3.11.0). In 3.10 and earlier this was a plain `raise KeyError(...)`.

2. **`ColormapRegistry` does not define `__contains__`**, so it inherits `Mapping.__contains__`, i.e. `try: self[key] / except KeyError: return False`. A membership test that should be an O(1) dict lookup therefore pays full error-message construction. On this machine:

   | | registry size 182 | registry size 602 |
   |---|---|---|
   | `'viridis' in colormaps` (hit) | 0.0009 ms | — |
   | `'nope' in colormaps` (miss) | 0.23 ms | 1.88 ms |

3. **colorcet** registers 420 colormaps, and each registration performs two membership misses — its own idempotency guard plus the one inside `ColormapRegistry.register`:

   ```python
   # colorcet/_dependencies.py
   if name not in colormaps or colormaps[name] != cmap:   # miss #1
       colormaps.register(cmap, name=name)                # -> `if name in self:` = miss #2
   ```

   That is 840 `difflib` passes over a registry growing from 182 to 602 entries — quadratic, with long colorcet names (`linear_kryw_5_100_c64`) making each `SequenceMatcher` pass expensive.

4. **holoviews** builds all three provider lists before deciding which provider owns the name (`plotting/util.py:1054`):

   ```python
   mpl_cmaps = _list_cmaps("matplotlib")
   bk_cmaps  = _list_cmaps("bokeh")
   cet_cmaps = _list_cmaps("colorcet")   # <- `from colorcet import ...`
   ```

   so resolving `'viridis'`, a plain matplotlib name, imports colorcet.

Links 1–3 are each individually reasonable (typo suggestions are a genuine UX improvement; colorcet's guard exists because `register` raises `ValueError` on a duplicate name). The defect is link 2, and it is matplotlib's. But holoviews is where users hit it, and link 4 is what turns "you asked for a colorcet colormap" into "you asked for any colormap at all".

The practical impact for us: in a Panel/Bokeh server, this lands on a session's IOLoop during the first plot render and blocks every other request for ~3 s — including static asset requests, which then time out.

Holoviews can mitigate this cheaply, and also `process_cmap` importing colorcet to resolve a matplotlib name looks avoidable regardless of the matplotlib bug (I have not reported to matplotlib yet, but think it would be worth fixing there as well):

- resolve providers lazily / short-circuit — check matplotlib first, only fall through to bokeh and colorcet if the name is not found, and build the full three-provider list only for the "not found among …" error path;
- and/or cache `_list_cmaps` per provider.

Either would leave a fast path for the common case of a matplotlib or bokeh colormap name. Neither fixes it for genuine colorcet names, which still pay the import.

```python
import time
from holoviews.plotting.util import process_cmap

t = time.perf_counter(); process_cmap('viridis'); first = time.perf_counter() - t
t = time.perf_counter(); process_cmap('viridis'); second = time.perf_counter() - t
print(f"first {first:.3f} s   second {second * 1e3:.2f} ms")
# first 2.728 s   second 1.04 ms
```

<details>
<summary>Isolating the root cause — no holoviews involved</summary>

Monkeypatching an O(1) `__contains__` onto `ColormapRegistry` is the only difference between the two runs:

```python
import sys, time
import matplotlib, matplotlib.cm as cm

if "--patched" in sys.argv:
    cm.ColormapRegistry.__contains__ = lambda self, key: key in self._cmaps

before = len(matplotlib.colormaps)
t = time.perf_counter()
import colorcet          # what process_cmap() triggers
dt = time.perf_counter() - t
print(f"import colorcet: {dt:.3f} s  ({before} -> {len(matplotlib.colormaps)} colormaps)")
```

```
$ python repro.py             ->  import colorcet: 2.811 s  (182 -> 602 colormaps)
$ python repro.py --patched   ->  import colorcet: 0.068 s  (182 -> 602 colormaps)
```

41x, from one line. Confirming that the entire cost is discarded error messages:

```python
import time, difflib, matplotlib._api as api

orig, n, spent = difflib.get_close_matches, [0], [0.0]
def counted(*a, **k):
    n[0] += 1; t = time.perf_counter()
    try: return orig(*a, **k)
    finally: spent[0] += time.perf_counter() - t
api.difflib.get_close_matches = counted

t = time.perf_counter(); import colorcet; total = time.perf_counter() - t
print(f"import colorcet: {total:.3f} s, of which difflib {spent[0]:.3f} s in {n[0]} calls")
# import colorcet: 2.697 s, of which difflib 2.592 s in 840 calls
```

</details>

N/A — no error, purely a performance issue.

#### Screenshots or screencasts of the bug

N/A.

SimonHeybrock:
## Description

Fixes #6981.

I was chasing multi-second stalls in a Panel dashboard: switching to a tab containing `Image` plots for the first time took ~4.5 s, and every other request to the server was blocked for the duration. It turned out to be the first `process_cmap()` call in the process, which takes nearly 3 seconds with matplotlib 3.11:

```python
import time
from holoviews.plotting.util import process_cmap

t = time.perf_counter(); process_cmap('viridis', 3); first = time.perf_counter() - t
t = time.perf_counter(); process_cmap('viridis', 3); second = time.perf_counter() - t
print(f"first {first * 1e3:.0f} ms   second {second:.2f} ms")
```

| | before | after |
|---|---|---|
| `process_cmap('viridis', 3)` | 2798 ms | 69 ms |
| `process_cmap('Category20', 3)` | 3009 ms | 95 ms |
| `process_cmap('kbc', 3)` (a genuine colorcet name) | 2754 ms | 2727 ms |

The cost is `import colorcet`, which `process_cmap` triggered for *every* string colormap because it built all three provider lists before deciding which provider owns the name. Consulting the providers in the same order but only until one claims the name keeps colorcet out of the picture for matplotlib and bokeh names.

Importing colorcet is slow because [matplotlib 3.11](https://github.com/matplotlib/matplotlib/pull/30001) added `difflib`-based "Did you mean …?" suggestions to `ColormapRegistry.__getitem__`, and `ColormapRegistry` inherits `Mapping.__contains__`, which is implemented as `try: self[key] / except KeyError`. Colorcet's 420 registrations each perform two membership *misses*, so each one builds a suggestion string over a registry growing from 182 to 602 entries. That is matplotlib's bug and I've reported it there as well — the holoviews side is worth fixing regardless, since resolving `'viridis'` never needed colorcet.

There is one intentional behaviour change beyond the speedup. Colorcet registers its colormaps with matplotlib under a `cet_` prefix, so `process_cmap('cet_fire')` only ever worked if something had already imported colorcet — on a fresh interpreter it raised `ValueError`. Those names are now made visible explicitly, so they resolve on the first call too.

Issue #6981 also suggests caching `_list_cmaps`. I left that alone: `list_cmaps` is public API reading live registries, and a cache would go stale as soon as a user registers a colormap.

A note on the tests: I am not sure they are very useful, as they test an implementation detail that only matters due to the problematic combination of behaviors of the involved libraries. The subprocess is unavoidable though — both properties it asserts are about colorcet *not* having been imported yet, which no in-process test can observe.

## AI Disclosure

Tool & Model: Claude Code + Opus 5
Usage: Used to profile the stall down to `difflib.get_close_matches`, and to draft the patch and the two tests. The patch was validated by running `process_cmap` over a matrix of colormap names and providers under both the old and new implementation, in separate interpreters, and diffing the results — the first draft of the patch silently broke `cet_`-prefixed names, which is why that case now has a test. Both new tests were confirmed to fail against unpatched `util.py`.

## Checklist

### DIFF
--- a/holoviews/plotting/util.py
+++ b/holoviews/plotting/util.py
@@ -1057,6 +1057,10 @@ def process_cmap(cmap, ncolors=None, provider=None, categorical=False):
     elif isinstance(cmap, list):
         palette = cmap
     elif isinstance(cmap, str):
+        if cmap.startswith("cet_"):
+            # colorcet registers its colormaps with matplotlib under a 'cet_' prefix,
+            # so matplotlib only knows these names once colorcet has been imported.
+            import colorcet  # noqa: F401
         # Providers are consulted in order and only until one claims the name, so
         # that a matplotlib or bokeh colormap does not pay for importing colorcet.
         if provider == "matplotlib" or (
@@ -1069,11 +1073,6 @@ def process_cmap(cmap, ncolors=None, provider=None, categorical=False):
             palette = bokeh_palette_to_palette(cmap, ncolors, categorical)
         elif provider == "colorcet" or (provider is None and _provides_cmap("colorcet", cmap)):
             palette = colorcet_cmap_to_palette(cmap, ncolors, categorical)
-        elif provider is None and _provides_cmap("matplotlib", cmap, cmap.lower()):
-            # colorcet registers its colormaps with matplotlib under a 'cet_' prefix.
-            # Those names only become visible once the check above has imported it,
-            # so matplotlib is consulted a second time before giving up.
-            palette = mplcmap_to_palette(cmap, ncolors, categorical)
         else:
             raise ValueError(
                 f"Supplied cmap {cmap} not found among {providers_checked} colormaps."
--- a/holoviews/plotting/util.py
+++ b/holoviews/plotting/util.py
@@ -1040,6 +1040,12 @@ register_cmaps("Uniform Categorical", "colorcet", "cet", "dark", ["glasbey_light
 register_cmaps("Uniform Categorical", "colorcet", "cet", "light", ["glasbey_dark"])
 
 
+def _provides_cmap(provider, *names):
+    """Whether the provider offers a colormap under any of the given names."""
+    cmaps = _list_cmaps(provider)
+    return any(name in cmaps for name in names)
+
+
 def process_cmap(cmap, ncolors=None, provider=None, categorical=False):
     """Convert valid colormap specifications to a list of colors."""
     providers_checked = "matplotlib, bokeh, or colorcet" if provider is None else provider
@@ -1051,19 +1057,23 @@ def process_cmap(cmap, ncolors=None, provider=None, categorical=False):
     elif isinstance(cmap, list):
         palette = cmap
     elif isinstance(cmap, str):
-        mpl_cmaps = _list_cmaps("matplotlib")
-        bk_cmaps = _list_cmaps("bokeh")
-        cet_cmaps = _list_cmaps("colorcet")
+        # Providers are consulted in order and only until one claims the name, so
+        # that a matplotlib or bokeh colormap does not pay for importing colorcet.
         if provider == "matplotlib" or (
-            provider is None and (cmap in mpl_cmaps or cmap.lower() in mpl_cmaps)
+            provider is None and _provides_cmap("matplotlib", cmap, cmap.lower())
         ):
             palette = mplcmap_to_palette(cmap, ncolors, categorical)
         elif provider == "bokeh" or (
-            provider is None and (cmap in bk_cmaps or cmap.capitalize() in bk_cmaps)
+            provider is None and _provides_cmap("bokeh", cmap, cmap.capitalize())
         ):
             palette = bokeh_palette_to_palette(cmap, ncolors, categorical)
-        elif provider == "colorcet" or (provider is None and cmap in cet_cmaps):
+        elif provider == "colorcet" or (provider is None and _provides_cmap("colorcet", cmap)):
             palette = colorcet_cmap_to_palette(cmap, ncolors, categorical)
+        elif provider is None and _provides_cmap("matplotlib", cmap, cmap.lower()):
+            # colorcet registers its colormaps with matplotlib under a 'cet_' prefix.
+            # Those names only become visible once the check above has imported it,
+            # so matplotlib is consulted a second time before giving up.
+            palette = mplcmap_to_palette(cmap, ncolors, categorical)
         else:
             raise ValueError(
                 f"Supplied cmap {cmap} not found among {providers_checked} colormaps."
--- a/holoviews/tests/plotting/test_plotutils.py
+++ b/holoviews/tests/plotting/test_plotutils.py
@@ -1,5 +1,9 @@
 from __future__ import annotations
 
+import sys
+from subprocess import check_output
+from textwrap import dedent
+
 import numpy as np
 import pytest
 
@@ -19,6 +23,8 @@ from holoviews.plotting.util import (
     split_dmap_overlay,
 )
 from holoviews.streams import PointerX
+
+from .._deps import mpl_skip
 
 
 class TestOverlayableZorders:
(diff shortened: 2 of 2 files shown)


hoxbro:
I applied a fix upstream https://github.com/matplotlib/matplotlib/pull/32172, which will be released in Matplotlib 3.11.2.

I don't think it is worth adding this complexity. Could you until the next matplotlib release either use an older version of matplotlib or make an active import of colorcet?

SimonHeybrock:
Yes, that works for me.

Did you note the

> There is one intentional behaviour change beyond the speedup. Colorcet registers its colormaps with matplotlib under a cet_ prefix, so process_cmap('cet_fire') only ever worked if something had already imported colorcet — on a fresh interpreter it raised ValueError. Those names are now made visible explicitly, so they resolve on the first call too.

Not relevant for me, just wanted to bring it up before closing in case it matters for someone.

hoxbro:
I did not, I have opened https://github.com/holoviz/holoviews/pull/6983 with a fix. 

Thank you for investigating and figuring out this regression.

PR #6997, opened 2026-08-17
ci: fix failing CI tests

Azaya89:
Fix failing tests in the CI. I think this is due to the recent dev releases of bokeh that has an incompatibility with the latest Panel version. 

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -70,7 +70,7 @@ nomkl = "*"
 pip = "*"
 python-gil = "*"
 # Required
-bokeh = ">=3.1"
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -70,7 +70,7 @@ nomkl = "*"
 pip = "*"
 python-gil = "*"
 # Required
-bokeh = ">=3.1"
+bokeh = ">=3.1,<3.10"
```
hoxbro:
```suggestion
bokeh = ">=3.1,<3.10.0a0" # upper pin until support in panel
```

PR #6998, opened 2026-08-20
compat: Bokeh 3.10.0

hoxbro:
## Description

Needs a new panel dev release, fixed in https://github.com/holoviz/panel/pull/8716/. 

Bokeh no longer require contourpy.

### DIFF
--- a/holoviews/tests/operation/test_operation.py
+++ b/holoviews/tests/operation/test_operation.py
@@ -28,6 +28,7 @@ from holoviews.operation.element import (
 from holoviews.testing import assert_element_equal
 
 from .._deps import (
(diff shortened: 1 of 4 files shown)


holoviews/tests/operation/test_operation.py
```diff
@@ -139,559 +140,208 @@ def test_image_gradient(self):
             img.clone(np.array([[3.162278, 3.162278], [3.162278, 3.162278]]), group="Gradient"),
         )
 
-    def test_image_contours(self):
```
hoxbro, line 142: The big diff here is just moving contour related test into its own class.

PR #6999, opened 2026-08-21
docs: add markdown and llms.txt docs

Azaya89:
## Description

This PR documents the addition of the markdown and `llms.txt` build process to the docs following 85b69cebe. See https://github.com/holoviz/hvplot/pull/1751

- [x] Added documentation

### DIFF
--- a/doc/developer_guide/index.md
+++ b/doc/developer_guide/index.md
@@ -249,6 +249,17 @@ You can also disable the reference gallery by setting the environment variable `
 
 A development version of HoloViews can be found [here](https://dev.holoviews.org/). You can ask a maintainer if they want to make a dev release for your PR, but there is no guarantee they will say yes.
 
+### Markdown docs and `llms.txt`
+
+As part of `docs-build`, HoloViews also generates clean Markdown versions of its documentation together with an `llms.txt` index file that points AI assistants at the most relevant pages.
+
+The build produces:
+
+- `builtdocs/markdown/` - a Markdown rendering of the documentation
+- `builtdocs/llms.txt` - an index of the most useful documentation pages for LLM-assisted development
+
+See [Using HoloViews with an AI assistant](../user_guide/use_llms.md) for how to consume these docs with an AI assistant.
+
 ## Build
 
 HoloViews have two build tasks. One is for building packages for Pip, and the other is for building packages for Conda.
--- a/doc/developer_guide/index.md
+++ b/doc/developer_guide/index.md
@@ -249,17 +249,6 @@ You can also disable the reference gallery by setting the environment variable `
 
 A development version of HoloViews can be found [here](https://dev.holoviews.org/). You can ask a maintainer if they want to make a dev release for your PR, but there is no guarantee they will say yes.
 
-### Markdown docs and `llms.txt`
-
-As part of `docs-build`, HoloViews also generates clean Markdown versions of its documentation together with an `llms.txt` index file that points AI assistants at the most relevant pages.
-
-The build produces:
-
-- `builtdocs/markdown/` - a Markdown rendering of the documentation
-- `builtdocs/llms.txt` - an index of the most useful documentation pages for LLM-assisted development
-
-See [Using HoloViews with an AI assistant](../user_guide/use_llms.md) for how to consume these docs with an AI assistant.
-
 ## Build
 
 HoloViews have two build tasks. One is for building packages for Pip, and the other is for building packages for Conda.
--- a/doc/user_guide/use_llms.md
+++ b/doc/user_guide/use_llms.md
@@ -1,0 +1,70 @@
+# Using HoloViews with an AI assistant
+
+AI assistants (like Claude, GitHub Copilot, or OpenAI Codex) can write HoloViews code much more reliably when they have access to up-to-date, structured documentation instead of guessing from stale training data. This guide shows you how to give your assistant that context, either by pointing it at HoloViews' generated `llms.txt` and Markdown docs, or by installing the HoloViz Agent Skills.
+
+## What you'll accomplish
+
+By the end of this guide you'll be able to:
+
+- Install the HoloViz skills so your assistant follows HoloViews best practices automatically
+- Give your AI assistant a curated index of the most useful HoloViews documentation pages
+
+## Prerequisites
+
+- A python package installer like `pip` or `uvx` if you want to install the skills
+- An AI assistant that can read web content or local files (for example Claude, GitHub Copilot, or OpenAI Codex)
+
+## Install the HoloViz skills
+
+The [holoviz-skills](https://holoviz-dev.github.io/holoviz-skills/) repository provides a curated collection of [Agent Skills](https://www.anthropic.com/news/skills) for the HoloViz ecosystem, including HoloViews. Agent Skills are folders of instructions and references that teach an assistant how to work with a library correctly.
+
+::::{tab-set}
+
+:::{tab-item} uv
+
+```bash
+uvx holoviz-skills@latest install
+```
+
+:::
+
+:::{tab-item} pip
+
+```bash
+pip install holoviz-skills
+holoviz-skills install
+```
+
+:::
+
+::::
+
+The `install` command auto-detects the AI tools on your machine. You can also target a specific tool, for example:
+
+```bash
+holoviz-skills install --claude-code
+holoviz-skills install --copilot
+```
+
+You may need to restart your assistant so it picks up the newly installed skills.
+
+## Point your assistant at the docs
+
+HoloViews publishes a machine-friendly index of its documentation that your assistant can fetch directly.
+
+1. Tell your assistant to fetch the index:
+
+   `https://holoviews.org/llms.txt`
+
+2. The index lists the most relevant pages and links to their Markdown versions under:
+
+   `https://holoviews.org/markdown/`, for example `https://holoviews.org/markdown/getting_started/1-Introduction.md`
+
+3. Ask your assistant to read the pages relevant to your task before writing code. For example, to create a plot you might have it read the Getting Started tutorial and the API reference page for the method you need.
+
+The Markdown pages are intentionally free of the navigation chrome, CSS, and JavaScript of the full HTML site, so they consume far fewer tokens and are easier for an assistant to parse.
+
+## Which approach should I use?
+
+- **Point your assistant at the docs** is a good fit for a one-off task or when you can't install packages, and it always reflects the latest published docs.
+- **Install the HoloViz skills** gives your assistant packaged, opinionated guidance that works offline, but you need to update it as HoloViews evolves. Once the skills are installed, pointing to the Markdown docs should no longer be needed if you want your agent to be able to fetch the latest docs, as the skills already teach the agent how to do that.
--- a/doc/user_guide/use_llms.md
+++ b/doc/user_guide/use_llms.md
@@ -58,7 +58,7 @@ HoloViews publishes a machine-friendly index of its documentation that your assi
 
 2. The index lists the most relevant pages and links to their Markdown versions under:
 
-   `https://holoviews.org/markdown/`, for example `https://holoviews.org/markdown/getting_started/1-Introduction.md`
+   `https://holoviews.org/markdown/`, for example `https://holoviews.org/markdown/getting_started/Introduction.md`
 
 3. Ask your assistant to read the pages relevant to your task before writing code. For example, to create a plot you might have it read the Getting Started tutorial and the API reference page for the method you need.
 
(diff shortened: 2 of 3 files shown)


doc/developer_guide/index.md
```diff
@@ -249,6 +249,17 @@ You can also disable the reference gallery by setting the environment variable `
 
 A development version of HoloViews can be found [here](https://dev.holoviews.org/). You can ask a maintainer if they want to make a dev release for your PR, but there is no guarantee they will say yes.
 
+### Markdown docs and `llms.txt`
```
hoxbro: I don't think this is relevant here.

doc/user_guide/use_llms.md
```diff
@@ -0,0 +1,70 @@
...
+
+HoloViews publishes a machine-friendly index of its documentation that your assistant can fetch directly.
+
+1. Tell your assistant to fetch the index:
+
+   `https://holoviews.org/llms.txt`
+
+2. The index lists the most relevant pages and links to their Markdown versions under:
+
+   `https://holoviews.org/markdown/`, for example `https://holoviews.org/markdown/getting_started/1-Introduction.md`
```
hoxbro: This is not taken into account the removal of the `1-` prefix like done in the official page: https://holoviews.org/getting_started/Introduction.html

Azaya89: In the markdown version, the `1-` prefix is not removed. It's still named `1-Introduction.md`. This is because the conversion is done from the notebooks in the `examples/` directory.

doc/developer_guide/index.md
Azaya89: Do you suggest I remove it entirely or move it somewhere else?

hoxbro: Remove it entirely.

doc/user_guide/use_llms.md
hoxbro: I think that is a mistake. It should remove the prefix to match up as much with the real url.

Azaya89: This will prove a bit more difficult because of the way HoloViews docs are built. As I understand it, the source for the `getting_started` section is the `examples/` folder which contains the notebooks. The notebooks are then run/evaluated and copied over to the `docs/` directory (still maintaining the original file name). Then a `.rst` file is created which now has the file name prefix removed. This `.rst` file is what is now served in the website. However, the content of the `.rst` file is just a redirect to the evaluated notebook in the `docs/` directory. 

The issue I had when making the [conversion script on nbsite](https://github.com/holoviz-dev/nbsite/blob/main/nbsite/scripts/_build_llms_txt.py) is that when converting these `.rst` files to markdown, what appears is just the almost empty redirect link that is not useful. So, a workaround was just to convert directly from the source notebook preferably.

hoxbro: I think you are overanalyzing it. See this patch by AI:


<details><summary>Transcript</summary>
<p>

# Locate nbsite number prefix removal logic

---

## User

Can you find the place where nbsite strip out the prefix of numbers for files like 1-?

---

## Assistant

**Tool: grep**

---

## Assistant

cmd.py:205 and 242 look most relevant.

**Tool: read**

---

## Assistant

`_path_and_order` is the function that strips the numeric prefix.

**Tool: grep**

---

## Assistant

Found it: `nbsite/cmd.py:365-371`, the `_path_and_order` function.

```python
def _path_and_order(filepath, keep_numbers):
    num_name = os.path.basename(filepath)
    leading_num = re.match(r"^\d+", num_name)
    if not keep_numbers:
        name = re.split(r"^\d+( |-|_)", num_name)[-1]
        filepath = filepath.replace(num_name, name)
    return filepath, int(leading_num.group(0)) if leading_num else None
```

It matches a leading `\d+` followed by a space, `-`, or `_` (via `re.split(r"^\d+( |-|_)", num_name)`), and if found and `keep_numbers` is not set, strips that prefix from the filename (used to build the rst path). The leading number is also captured separately as the `order` value used for toctree sorting. This is called from `generate_rst` (line 266) and `_toctree` (line 377).

---

## User

How hard would it be to recreate this for the llms.txt stuff? So something like https://holoviews.org/markdown/getting_started/1-Introduction.md would be https://holoviews.org/markdown/getting_started/Introduction.md`

---

## Assistant

**Tool: read**

---

## Assistant

Not hard — you'd apply the same regex-based stripping nbsite already does in `cmd.py`'s `_path_and_order`, but here at the point where `_build_llms_txt.py` computes `rel`/`destination` filenames.

Currently the destination path is built verbatim from the source:

```python
rel = path.relative_to(source.source_dir)
destination = source.output_dir / rel
```

So `1-Introduction.ipynb` becomes `1-Introduction.md`, and the numeric prefix propagates through `generated_paths` into all the labels, links, and llms.txt sections downstream (since everything else works off the already-stripped `generated_paths`/`destination`).

To fix it, add a small helper mirroring nbsite's logic and apply it per-path-component (since prefixes can appear on directories too, e.g. `getting_started` isn't prefixed but a subfolder could be):

```python
_NUM_PREFIX_RE = re.compile(r"^\d+[-_ ]")

def _strip_numeric_prefix(rel: Path) -> Path:
    return Path(*(_NUM_PREFIX_RE.sub("", part) for part in rel.parts))
```

Then in `build_markdown_docs`, change:

```python
rel = path.relative_to(source.source_dir)
destination = source.output_dir / rel
```

to:

```python
rel = path.relative_to(source.source_dir)
destination = source.output_dir / _strip_numeric_prefix(rel)
```

That's essentially the whole change, since:

- `generated.append(destination.relative_to(markdown_root))` and all label builders (`default_label`, `index_label`) already just use `path.stem`/`path.parent.name`, so once the destination filename is fixed, labels, links, and llms.txt URLs are automatically correct.
- `generate_index_pages` globs `category_dir.glob("*.md")`, which will just see the stripped filenames.
- Ordering (nbsite's toctree sort-by-leading-number) isn't used here — `_iter_source_files`/`generated_paths` are just sorted alphabetically/by glob order, so if you want the number preserved for sort order (like nbsite does for the toctree) but stripped from the visible filename, you'd want to capture the leading number similarly to `_path_and_order` and use it as a sort key when building `LlmsSection` bodies/index pages, rather than dropping it entirely. That's a bit more work but optional depending on whether ordering currently matters.

Caveats worth checking:
1. **Collisions**: stripping prefixes could cause two files to map to the same stripped name (e.g. `1-intro.md` and `intro.md` in the same directory). Worth adding a check/warning, since nbsite doesn't need to worry about this as much (it operates in the rst-generation step before merge, per source dir).
2. **Internal links**: `_strip_markdown_noise`'s `_to_markdown_link` rewrites `.html`/`.ipynb` links to `.md` but keeps the original filename (with number prefix) in the link target. If notebooks cross-link to each other by filename (e.g. `[..](1-Introduction.ipynb)`), those links need the same prefix-stripping applied, or they'll point to now-nonexistent filenames. This is the main non-trivial part — you'd want a link-rewriting pass that applies `_strip_numeric_prefix` to the path portion of markdown links after conversion.

Overall: straightforward for the primary case (top-level file renaming), but you should also handle #1 (collision safety) and #2 (rewriting cross-links) to fully mirror nbsite's behavior rather than just the filename stripping.

---

## User

collisions is not a problem as this will also happen with the real examples.

---

## Assistant

**Tool: edit**

---

## Assistant

Now update `build_markdown_docs` to strip the prefix when computing the destination:

**Tool: edit**

---

## Assistant

Now fix cross-links so links between pages point at the stripped filenames too:

**Tool: read**

---

## Assistant

**Tool: edit**

---

## Assistant

Add the `PurePosixPath` import and make `_strip_numeric_prefix` accept both `Path` and `PurePosixPath`:

**Tool: edit**

---

## Assistant

**Tool: bash**

---

## Assistant

Failures are pre-existing (pandoc not installed), unrelated to my change. Let me write a quick targeted test to confirm the prefix-stripping and link-rewriting work.

**Tool: bash**

---

## Assistant

**Tool: write**

---

## Assistant

**Tool: bash**

---

## Assistant

**Tool: bash**

---

## Assistant

Filenames are stripped correctly and the cross-links now correctly resolve to the renamed file (`2-Installation.html` → `Installation.md`, fragment preserved). Cleaning up the temp test:

**Tool: bash**

---

## Assistant

Summary of changes in `nbsite/scripts/_build_llms_txt.py`:

1. Added `_strip_numeric_prefix(rel: Path)` — strips a leading `\d+[-_ ]` from every path component, mirroring nbsite's `cmd._path_and_order`.
2. `build_markdown_docs` now computes `destination = source.output_dir / _strip_numeric_prefix(rel)` instead of using the raw `rel`, so `1-Introduction.ipynb`/`.md`/`.rst` outputs become `Introduction.md`.
3. `_to_markdown_link` (used by `_strip_markdown_noise` to rewrite `.html`/`.ipynb` cross-references to `.md`) now also strips the numeric prefix from the link target path before appending `.md`, so internal links between pages still resolve correctly after renaming (fragments preserved).

Verified with an ad-hoc script (not committed) that:
- `getting_started/1-Introduction.md` → `getting_started/Introduction.md`
- A link `[Installation](2-Installation.html)` inside it correctly rewrites to `[Installation](Installation.md)`, including fragment handling for `2-Installation.ipynb#setup` → `Installation.md#setup`.

Existing test suite (`nbsite/tests/test_build_llms_txt.py`) passes except two pre-existing failures unrelated to this change (missing `pandoc` binary in this environment).

---

## User

write a short summary.

---

## Assistant

Added numeric-prefix stripping to the llms.txt markdown builder, matching nbsite's existing rst-generation behavior.

**Changes in `nbsite/scripts/_build_llms_txt.py`:**
- New `_strip_numeric_prefix()` helper strips a leading `\d+[-_ ]` from each path component (e.g. `1-Introduction` → `Introduction`).
- `build_markdown_docs` applies it to output filenames, so `1-Introduction.ipynb` now generates `Introduction.md` instead of keeping the ordering prefix.
- `_to_markdown_link` (cross-reference rewriting) applies the same stripping to link targets, so internal links between pages (e.g. `2-Installation.html#setup`) still resolve correctly to the renamed file (`Installation.md#setup`).

Verified manually that stripping and link rewriting work together correctly; existing test suite passes (two unrelated pre-existing failures due to missing `pandoc` in this environment). Collisions from stripping (e.g. two differently-numbered files reducing to the same name) are accepted as expected, matching real-world example behavior.


</p>
</details> 


```patch
diff --git a/nbsite/scripts/_build_llms_txt.py b/nbsite/scripts/_build_llms_txt.py
index c74760b..1ce0bf4 100644
--- a/nbsite/scripts/_build_llms_txt.py
+++ b/nbsite/scripts/_build_llms_txt.py
@@ -13,7 +13,7 @@ import tempfile
 
 from dataclasses import dataclass, field
 from itertools import groupby
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 from typing import Callable, Iterable, Sequence
 
 from bs4 import BeautifulSoup
@@ -82,6 +82,19 @@ _META_REFRESH_RE = re.compile(
 )
 
 
+_NUM_PREFIX_RE = re.compile(r"^\d+[-_ ]")
+
+
+def _strip_numeric_prefix(rel: Path) -> Path:
+    """Strip a leading ``<digits>-``/``_``/`` `` prefix from each path part.
+
+    Mirrors nbsite's ``cmd._path_and_order`` behavior for rst generation, so
+    e.g. ``1-Introduction.ipynb`` becomes ``Introduction.md`` in the markdown
+    output tree instead of keeping the ordering prefix in the visible name.
+    """
+    return Path(*(_NUM_PREFIX_RE.sub("", part) for part in rel.parts))
+
+
 def default_label(path: Path) -> str:
     return path.stem.replace("_", " ")
 
@@ -484,7 +497,9 @@ def _strip_markdown_noise(text: str) -> str:
         url, fragment = match.group(1), match.group(2) or ""
         if url.startswith(("http:", "https:", "#", "mailto:")):
             return f"({url}{fragment})"
-        return f"({url.removesuffix('.html').removesuffix('.ipynb')}.md{fragment})"
+        stem = url.removesuffix('.html').removesuffix('.ipynb')
+        stem = _strip_numeric_prefix(PurePosixPath(stem)).as_posix()
+        return f"({stem}.md{fragment})"
 
     text = re.sub(
         r"\((?!http|https|#|mailto)([^()]*\.(?:html|ipynb))(#[^)]*)?\)",
@@ -570,7 +585,7 @@ def build_markdown_docs(
     for source in sources:
         for path in _iter_source_files(source):
             rel = path.relative_to(source.source_dir)
-            destination = source.output_dir / rel
+            destination = source.output_dir / _strip_numeric_prefix(rel)
 
             if path.suffix == ".md" and source.copy_markdown:
                 destination.parent.mkdir(parents=True, exist_ok=True)
```

Azaya89: OK, thanks.

hoxbro (approved):

PR #7014, opened 2026-08-28
ci: Avoid latest matplotlib versions

hoxbro:
## Description

Because of the regression in Matplotlib our CI is significant slower when running example notebook 5-10 second overhead per notebook where colorcet is imported in some way.

## Checklist


hoxbro:
It aint worth it for the not updated warnings.

PR #7025, opened 2026-08-31
compat!: Drop Python 3.10 and 3.11 support

hoxbro:
## Description

Drop support for Python 3.10 and 3.11. The reasoning for this is:
- Latest Bokeh (3.10) and next panel (1.10) will both drop support for them.
- Some good typing stuff which will help my work on adding basic typing to HoloViews.
- Other libraries like numpy have also dropped it. 
- conda-forge dropped support for Python 3.10 by default, meaning new releases would not be built for Python 3.10.

Also bumping a lot of the python versions in environment to 3.14.

### DIFF
--- a/holoviews/core/util/types.py
+++ b/holoviews/core/util/types.py
@@ -16,22 +16,18 @@ if t.TYPE_CHECKING:
     from pandas.core.dtypes.dtypes import DatetimeTZDtype
     from pandas.core.dtypes.generic import ABCExtensionArray, ABCIndex, ABCSeries
 
-    _PdDatetimeT: t.TypeAlias = pd.Timestamp | pd.Period | DatetimeTZDtype
-    _PdTimedeltaT: t.TypeAlias = pd.Timedelta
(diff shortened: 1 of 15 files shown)


holoviews/core/util/types.py
```diff
@@ -16,22 +16,18 @@
...
+    type _DatetimeT = dt.datetime | dt.date | dt.time | np.datetime64 | _PdDatetimeT | _CftimeT
+    type _TimedeltaT = dt.timedelta | np.timedelta64 | _PdTimedeltaT
+    type _ArraylikeT = np.ndarray | nw.Series | ABCIndex | ABCSeries | ABCExtensionArray
+    type _MaskedT = np.ma.core.MaskedArray | BaseMaskedArray
+
+    class _GenFunc[YieldT](t.Protocol):
         __name__: str
 
-        def __call__(self) -> t.Iterator[_YieldT]: ...
+        def __call__(self) -> t.Iterator[YieldT]: ...
```
hoxbro: T

philippjfr (approved):

PR #7034, opened 2026-09-01
fix: Apply xlabel/ylabel to the visual axes under invert_axes

### LINKED ISSUE #7033: Cannot set xlabel, ylabel on horizontal bar
### ALL software version info
hvplot-0.8.1

### Description of expected behavior and the observed behavior
Cannot set xlabel, ylabel on horizontal bar  

### Complete, minimal, self-contained example code that reproduces the issue
mpg = pd.read_csv('https://raw.githubusercontent.com/tidyverse/ggplot2/master/data-raw/mpg.csv')

#### Without specifying an axis label
image

#### Specifying xlabel uses a wrong label
image

#### Specifying ylabel uses a different, wrong label
image

Azaya89:
Fixes #7033

## Description

This PR fixes the label swap on inverted plots (e.g. `hvplot.barh`, which is `Bars + invert_axes=True`). Custom `xlabel/ylabel` previously labeled the element's data dimensions and were swapped by each backend under `invert_axes`.

This PR makes `xlabel/ylabel/zlabel` refer to the visual axes, never swapped, while auto-derived dimension labels still follow their dimension (swapped under `invert_axes`). 

### HoloViews
```python
import pandas as pd
import holoviews as hv
hv.extension('bokeh', 'matplotlib')

mpg = pd.read_csv('https://raw.githubusercontent.com/tidyverse/ggplot2/master/data-raw/mpg.csv')
mpg = mpg['manufacturer'].value_counts(sort=False)

# bokeh
hv.Bars(mpg).opts(
    invert_axes=True,
    xlabel='Car count',
    ylabel='Car Make'
)
```
bokeh_axes

```python
# mpl
hv.output(backend='matplotlib')

hv.Bars(mpg).opts(
    invert_axes=True,
    xlabel='Car count',
    ylabel='Car Make'
)
```
Screenshot 2026-09-01 at 4 49 37 PM

### hvPlot

```python
import hvplot.pandas  # noqa

df = hvplot.sampledata.penguins('pandas')
grouped = df.groupby('species', observed=True)['body_mass_g'].mean()

grouped.hvplot.barh(
    x='species',
    y='body_mass_g',
    title='Horizontal Bar Chart (Bokeh)',
    xlabel='Body Mass (g)',
    ylabel='Species',
)
```
hvplot_axes

## AI Disclosure

Tool & Model: Kilo (kilo-auto/balanced)
Usage: Drafted the implementation and tests; reviewed and edited by me.

Tool & Model:
Usage:

## Checklist

### DIFF
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -972,10 +972,11 @@ class ElementPlot(BokehPlot, GenericElementPlot):
             if dim:
                 axis_label = str(dim)
             else:
-                xlabel, ylabel, _zlabel = self._get_axis_labels(dims if dims else (None, None))
-                if self.invert_axes:
-                    xlabel, ylabel = ylabel, xlabel
-                axis_label = ylabel if pos else xlabel
+                axis_label = dim_axis_label(dims[0]) if dims and dims[0] else ""
+                if pos == 0 and self.xlabel is not None:
+                    axis_label = self.xlabel
+                elif pos == 1 and self.ylabel is not None:
+                    axis_label = self.ylabel
             if dims:
                 dims = dims[:2][::-1]
 
@@ -1514,8 +1515,6 @@ class ElementPlot(BokehPlot, GenericElementPlot):
             for axis, dim in zip(["x", "y"], dimensions, strict=False)
         }
         xlabel, ylabel, _zlabel = self._get_axis_labels(dimensions)
-        if self.invert_axes:
-            xlabel, ylabel = ylabel, xlabel
         props["x"]["axis_label"] = xlabel if "x" in self.labelled or self.xlabel else ""
         props["y"]["axis_label"] = ylabel if "y" in self.labelled or self.ylabel else ""
         if not self._has_changed("label", props):
--- a/holoviews/plotting/bokeh/element.py
+++ b/holoviews/plotting/bokeh/element.py
@@ -972,7 +972,7 @@ class ElementPlot(BokehPlot, GenericElementPlot):
             if dim:
                 axis_label = str(dim)
             else:
-                axis_label = dim_axis_label(dims[0]) if dims and dims[0] else ""
+                axis_label = dim_axis_label(dims) if dims else ""
                 if pos == 0 and self.xlabel is not None:
                     axis_label = self.xlabel
                 elif pos == 1 and self.ylabel is not None:
--- a/holoviews/plotting/plot.py
+++ b/holoviews/plotting/plot.py
@@ -1745,27 +1745,39 @@ class GenericElementPlot(DimensionedPlot):
         return (x0, y0, x1, y1)
 
     def _get_axis_labels(self, dimensions, xlabel=None, ylabel=None, zlabel=None):
-        if self.xlabel is not None:
-            xlabel = self.xlabel
-        elif dimensions and xlabel is None:
+        """Returns axis labels derived from the supplied dimensions.
+
+        Derived labels follow their data dimension and are swapped when
+        invert_axes is set. The xlabel/ylabel/zlabel plot options refer
+        to the visual axes and override the derived labels after that swap.
+        Explicit xlabel/ylabel/zlabel arguments still override the derived
+        labels before the invert_axes swap.
+
+        """
+        if dimensions and xlabel is None:
             xdims = dimensions[0]
             xlabel = dim_axis_label(xdims) if xdims else ""
 
-        if self.ylabel is not None:
-            ylabel = self.ylabel
-        elif len(dimensions) >= 2 and ylabel is None:
+        if len(dimensions) >= 2 and ylabel is None:
             ydims = dimensions[1]
             ylabel = dim_axis_label(ydims) if ydims else ""
 
-        if getattr(self, "zlabel", None) is not None:
-            zlabel = self.zlabel
-        elif (
+        if (
             isinstance(self.projection, str)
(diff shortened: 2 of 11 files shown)


holoviews/tests/plotting/bokeh/test_barplot.py
```diff
@@ -612,6 +612,36 @@ def test_bar_narrow_non_monotonous_xvals(self):
...
+        assert plot.xaxis[0].axis_label == "Species"
+        assert plot.yaxis[0].axis_label == "species"
+
+    def test_bars_invert_axes_ylabel_only(self):
+        bars = hv.Bars([("A", 1), ("B", 2)], "species", "body_mass_g").opts(
+            invert_axes=True, ylabel="Body Mass (g)"
+        )
+        plot = bokeh_renderer.get_plot(bars).state
+        assert plot.xaxis[0].axis_label == "body_mass_g"
+        assert plot.yaxis[0].axis_label == "Body Mass (g)"
```
hoxbro: I don't think this is the right way to test this. Shouldn't `ylabel` have been `"Species"` to make sense?

This is also the problem with the other tests.

holoviews/plotting/bokeh/element.py
```diff
@@ -972,10 +972,11 @@ def _axis_props(
...
             else:
-                xlabel, ylabel, _zlabel = self._get_axis_labels(dims if dims else (None, None))
-                if self.invert_axes:
-                    xlabel, ylabel = ylabel, xlabel
-                axis_label = ylabel if pos else xlabel
+                axis_label = dim_axis_label(dims) if dims else ""
+                if pos == 0 and self.xlabel is not None:
+                    axis_label = self.xlabel
+                elif pos == 1 and self.ylabel is not None:
+                    axis_label = self.ylabel
```
hoxbro, line 979:
```suggestion
                xlabel, ylabel, _zlabel = self._get_axis_labels(dims if dims else (None, None))
                axis_label = ylabel if pos else xlabel
```

I think this should be OK.

holoviews/plotting/plot.py
```diff
@@ -1745,27 +1745,39 @@ def get_extents(
...
-        if self.xlabel is not None:
-            xlabel = self.xlabel
-        elif dimensions and xlabel is None:
+        """Returns axis labels derived from the supplied dimensions.
+
+        Derived labels follow their data dimension and are swapped when
+        invert_axes is set. The xlabel/ylabel/zlabel plot options refer
+        to the visual axes and override the derived labels after that swap.
+        Explicit xlabel/ylabel/zlabel arguments still override the derived
+        labels before the invert_axes swap.
```
hoxbro: This is AI language, especially `still`.

hoxbro (commented):
Some small comments.

I think it makes sense to also test the behavior for plotly.

holoviews/tests/plotting/bokeh/test_barplot.py
Azaya89: The idea here is to test that the axis labels now follow the visual pattern rather than the data dimensions by intentionally using a mismatched label. So for tests that only use custom `xlabel` or `ylabel` (like this line), we want to see that the derived labels still work as intended and the custom labels also work too.

Azaya89:
> I think it makes sense to also test the behavior for plotly.

OK. Although I deliberately left them out 'cos we were moving away from supporting plotly, I suppose.

holoviews/plotting/bokeh/element.py
Azaya89: This wouldn't work for grouped `x-axis`:
```python
import holoviews as hv
hv.extension('bokeh')

bars = hv.Bars([
    ('A', 'X', 1), ('A', 'Y', 2),
    ('B', 'X', 3), ('B', 'Y', 4),
], kdims=['Category', 'Subcategory'], vdims=['Value'])

bars # xlabel shows 'Category' only instead of 'Category, Subcategory'
```

Azaya89: I added tests for that now.

holoviews/plotting/plot.py
Azaya89: Yeah, but it is precise and accurate. I couldn't think of a better way to phrase that.

hoxbro: There is no need to say `still`. We don't care about the past.

hoxbro:
> OK. Although I deliberately left them out 'cos we were moving away from supporting plotly, I suppose.

Nothing has been decided on that matter.

holoviews/tests/plotting/bokeh/test_barplot.py
hoxbro: I still don't think this is the right way to test it, it doesn't make any sense to me, why are we overriding species here?

image

Azaya89: We're not really overriding 'species' particularly. We're just checking whether if the user enters their own custom label, it will work as intended while still leaving the derived label (`xlabel` in this case) as it is. It could have been any string but this was chosen to also match closely the `xlabel`. Under the old code, this would fail and produce this output:

image

hoxbro: I'm talking about the plot itself. Nobody would write the code in the unit test.

Azaya89: OK. Will make it more realistic then.

hoxbro (approved):

PR #7042, opened 2026-09-06
ci: Small updates to pixi-diff.py

hoxbro:
## Description

Adds cache and extract full sha for main to get real cache.


hoxbro:
## Pixi.lock changes

No changes to pixi.lock.

PR #7070, opened 2026-09-14
fix: keep the category axis on single-category datashade aggregates

ahuang11:
## Description

`datashade` renders a fully transparent image when the aggregated categorical column has exactly one category. `tf.shade` receives a 2D array while `is_categorical` is still derived from the element, so a `color_key` is passed with it, datashader reads the x axis as the category axis, and `color_data` comes out all-NaN. `tf.shade` on the same aggregate renders fine, so this sits in the `shade` operation.

Two routes reach that 2D array, split by `element.interface.packed(element)`:

- `_extract_data` selected the lone value dimension with `array[vdim]` for an unpacked `ImageStack`, returning 2D; it now builds the stack axis. Reachable with `selector=ds.first()`.
- `_process` squeezed a packed `(h, w, 1)` aggregate on `array.shape[-1] == 1`; it now computes `is_categorical` before the squeeze and skips it for categorical aggregates.

```python
import datashader as ds, holoviews as hv, numpy as np, pandas as pd, panel as pn
hv.extension('bokeh')
print(hv.__version__, ds.__version__)   # 1.23.2, 0.19.1

rng = np.random.default_rng(0)
df = pd.DataFrame({
    'x': rng.normal(size=5000),
    'y': rng.normal(size=5000),
    'c': pd.Categorical(['a'] * 5000),  # exactly one category
})
pn.serve(datashade(
    hv.Points(df, ['x', 'y'], ['c']), aggregator=ds.count_cat('c'), color_key={'a': '#ff7f0e'}
))
# Before: blank plot, "RuntimeWarning: All-NaN slice encountered" in the terminal.
# After:  an orange point cloud.
```

Before:
image

After:
image

Two or more categories render either way. Adding `selector=ds.first()` gives the same blank output via the `_extract_data` route.

## AI Disclosure

Tool & Model: Claude Opus 5 (claude.ai)
Usage: Diagnosed the failure, wrote the patch, drafted this description. I ran the experiments and confirmed the behaviour.

## Checklist


holoviews/operation/datashader.py
```diff
@@ -1498,14 +1498,14 @@ def _extract_data(self, element):
...
-                    array = array[vdim]
-                else:
-                    array = array.to_array("z")
-                    # If data is 3D then we have one extra constant dimension
-                    if array.ndim > 3:
-                        drop = set(array.dims) - {*main_dims, "z"}
-                        array = array.squeeze(dim=drop)
+                # Keep the stack axis even for a single level: selecting the
+                # lone value dimension returns 2D, and _process still passes a
+                # color_key, so datashader would read x as the category axis.
```
hoxbro, line 1503: I think we should avoid comments which tell about prior knowledge. There is no need to mention what used to happen.

holoviews/operation/datashader.py
```diff
@@ -1536,7 +1536,12 @@ def _process(self, element, key=None):
         # Dask is not supported by shade so materialize it
         array = array.compute()
 
-        if array.shape[-1] == 1:
+        # Set before the squeeze below, which must not collapse a categorical
+        # aggregate: a single-category (h, w, 1) array would become 2D while
+        # still being shaded with a color_key.
```
hoxbro, line 1541:
```suggestion
```

The code is enough.

holoviews/operation/datashader.py
```diff
@@ -1970,11 +1974,10 @@ def _process(self, element, key=None):
                 ) in enumerate("RGBA"):
                     new_data[k].data = img[:, :, idx]
             elif isinstance(element, ImageStack):
-                if len(element.vdims) == 1:
-                    new_data[element.vdims[0].name].data = array
-                else:
-                    for k in map(str, element.vdims):
-                        new_data[k].data = array.sel(z=k)
+                # _extract_data keeps the z axis even for one level, so a
+                # single-vdim stack selects the same way as a multi-vdim one.
```
hoxbro, line 1978: A comment about what used to happen again.

holoviews/tests/operation/test_datashader.py
```diff
@@ -1710,6 +1710,41 @@ def test_selector_single_categorical():
...
+@pytest.mark.parametrize("use_selector", [False, True], ids=["no_selector", "selector"])
+@pytest.mark.parametrize("ncat", [1, 2], ids=["one_category", "two_categories"])
+def test_datashade_count_cat_single_category_not_transparent(ncat, use_selector):
+    """Shading a count_cat aggregate with one category must leave visible pixels.
+
+    A one-category aggregate lost its trailing axis while is_categorical stayed
+    True, so a color_key reached tf.shade with a 2D array and the image came out
+    fully transparent. test_selector_single_categorical covers the selector path
+    for #6595, but asserts only that rendering raises nothing.
+    """
```
hoxbro, line 1722:
```suggestion
```

hoxbro (commented):
Code changes looks great. I think the comments/docstring are way to verbose and maybe even redundant.

hoxbro (approved):

PR #7071, opened 2026-09-16
perf: Speed up `rolling`, `Dataset` methods and `MultiInterface.dtype`

hoxbro:
## Description

Speeds up the slowest notebooks of the docs build, found by profiling them. The docs build itself is sped up in holoviz-dev/nbsite#367.

1. **Use pandas rolling methods in `rolling`** (240fc211c, 9ff2b2f3b)
   `rolling` applied the function to every window in Python, e.g. 3.9 million `np.mean` calls in the profile of a cell of the Dashboards user guide. Without `window_type` and `min_periods`, a window with a NaN is never passed to the function, so the pandas rolling methods give the same result, up to floating-point rounding. They are now used for every function in `_PANDAS_FUNC_LOOKUP` with a rolling method: `sum`, `mean`, `median`, `min`, `max`, `std` and `var` (with `ddof=0`), including the NaN and builtin variants. Other functions, like `np.prod`, and `min_periods` still apply the function to every window, and `window_type` still only supports `np.mean` and `np.sum`.

2. **Only create the pipeline operation when a `Dataset` method records it** (0404d23a6)
   Every public `Dataset` method created a `method` operation, a `Parameterized` instance, even when it was thrown away because the call was nested in another method or did not return a `Dataset` or a mapping, as for `dimension_values` and `range`. This is a large part of computing ranges when plotting many elements.

3. **Don't create a `Dataset` for every `MultiInterface.dtype` call** (5247c8516, 6faa84cf3)
   `MultiInterface.dtype` created a template `Dataset` for every call to ask its subpath interface for the dtype, and it is called for every element and dimension when computing ranges. The subpath interface is now the first subtype whose `applies` accepts the data, as in `Interface.initialize`, and its `dtype` gets a lightweight stand-in for the subpath with the data and dimensions. This also covers the `geom_dictionary` subtype of GeoViews. Datasets with a `level` use the template as before.

4. **Reduce the number of items in the Dashboards user guide** (554821260)
   The static docs embed every combination of the widget values, 5 symbols, 6 variables, 3 rolling windows and 2 datashade values, which is 180 renders of the plot for one cell. With 3 symbols and 4 variables it is 72.

  For the `AdvancedStockExplorer` cell of `examples/user_guide/17-Dashboards.ipynb`, profiled in a kernel as it runs in the docs build, steps 1-3 take it from 81.6s to 55.8s, and step 4 to 26.7s.

### Benchmarks

Best of 5 repeats, on `main` and on this branch:

| Benchmark | `main` | This PR |
|---|---|---|
| `rolling` mean | 11940 us | 1694 us |
| `rolling` sum | 7938 us | 1690 us |
| `rolling` max | 7724 us | 1719 us |
| `rolling` median | 43542 us | 2627 us |
| `rolling` std | 37350 us | 1725 us |
| `Curve.dimension_values` | 101.2 us | 39.4 us |
| `Curve.range` | 161.9 us | 92.1 us |
| `Path.range` with 20 paths | 2217 us | 891 us |
| `MultiInterface.dtype` array paths | 250.5 us | 0.9 us |
| `MultiInterface.dtype` dict paths | 281.2 us | 8.9 us |
| `MultiInterface.dtype` GeoViews geometry paths | 283.0 us | 14.0 us |

<details>
<summary>Benchmark script</summary>

```python
import sys
import timeit

import numpy as np
import pandas as pd

import holoviews as hv
from holoviews.operation.timeseries import rolling


def bench(name, func, number):
    seconds = min(timeit.repeat(func, number=number, repeat=5)) / number
    print(f"{name:<32} {seconds * 1e6:10.1f} us")


def main():
    rng = np.random.default_rng(0)

    dates = pd.date_range("2020-01-01", periods=5000, freq="D")
    curve = hv.Curve((dates, rng.standard_normal(5000).cumsum()))
    for function in [np.mean, np.sum, np.max, np.median, np.std]:
        bench(
            f"rolling {function.__name__}",
            lambda function=function: rolling(curve, rolling_window=20, function=function),
            20,
        )

    curve = hv.Curve((np.arange(1000), rng.standard_normal(1000)))
    bench("Curve.dimension_values", lambda: curve.dimension_values(1), 5000)
    bench("Curve.range", lambda: curve.range(1), 5000)

    paths = [np.column_stack([np.linspace(0, 1, 200), rng.random(200)]) for _ in range(20)]
    path = hv.Path(paths, ["x", "y"])
    bench("Path.range", lambda: path.range("y"), 500)
    bench("MultiInterface.dtype array", lambda: path.interface.dtype(path, "y"), 2000)

    dicts = [{"x": np.linspace(0, 1, 200), "y": rng.random(200)} for _ in range(20)]
    path = hv.Path(dicts, ["x", "y"])
    bench("MultiInterface.dtype dict", lambda: path.interface.dtype(path, "y"), 2000)

    if "--geoviews" in sys.argv:
        import geoviews as gv
        from shapely.geometry import LineString

        lines = [{"geometry": LineString(rng.random((200, 2))), "value": i} for i in range(20)]
        path = gv.Path(lines, vdims=["value"])
        bench("MultiInterface.dtype geometry", lambda: path.interface.dtype(path, "Longitude"), 2000)


if __name__ == "__main__":
    main()
```

</details>

## AI Disclosure

Tool & Model: Claude Code + Claude Opus 5
Usage: Profiling the notebooks of the docs build, writing the changes and tests, and benchmarking them.

## Checklist


holoviews/operation/timeseries.py
```diff
@@ -74,7 +74,15 @@ def _process_layer(self, element, key=None):
         xdim = element.kdims[0].name
         df = PandasInterface.as_dframe(element)
         df = df.set_index(xdim).rolling(win_type=self.p.window_type, **self._roll_kwargs())
-        if self.p.window_type is None:
+        if (
+            self.p.window_type is None
+            and self.p.min_periods is None
+            and self.p.function in (np.mean, np.sum)
```
hoxbro: in _PANDAS_FUNC_LOOKUP?

holoviews/core/data/multipath.py
```diff
@@ -8,6 +8,26 @@
 from .dictionary import DictInterface
 from .interface import DataError, Interface
 
+# The subpath interface only depends on the type of the subpath data and the
+# subtypes, and finding it by creating a Dataset for every call is slow
+_subpath_interfaces = {}
+
+# Datatypes whose dtype method only uses the data and dimensions of the Dataset
+_DTYPE_FROM_DATA = {"dictionary", "dataframe", "array", "dask", "narwhals"}
```
hoxbro: GeoPandas

holoviews/core/data/__init__.py
```diff
@@ -222,44 +222,45 @@ def pipelined_fn(*args, **kwargs):
...
+                            getitem_op = method_op.instance(
+                                input_type=type(result), method_name="__getitem__", args=[key]
+                            )
+                            element._pipeline = inst_pipeline.instance(
+                                operations=[*inst_pipeline.operations, op, getitem_op],
+                                output_type=type(result),
+                            )
             finally:
-                if not in_method:
-                    inst._in_method = False
```
philippjfr, line 262: Jeez, that really wasn't well written.

philippjfr (approved):

PR #7080, opened 2026-09-21
fix: Sort full grid keys of sparse `GridSpace`

### LINKED ISSUE #7079: Internal layout issue for GridSpace with sparse dict keys
<details>

```plaintext
holoviews 1.23.2
```
</details>

Trying to circumvent #7055, I renamed the keys of the vertical kdim so that they're sorted in reverse, in the hope that I could achieve the inverted layout I need. Schematically, if I had ["a", "b", "c"], I apply the permutation "a" -> "c", "b" -> "b", "c" -> "a".

In my case the grid is not fully populated. It is indeed a legitimate use case to have plots defined _e.g._ only for a triangular matrix. And indeed Holoviews handles this well when the keys are defined in a certain order, with blank spaces put in places where there is no plot defined in the dictionary.

It works well if you have keys like `[(0, 0), (0, 1), (1, 1), (1, 2), (2, 2)]`, but with `[(0, 2), (0, 1), (1, 1), (1, 0), (2, 0)]`, the layout is clearly wrong. It looks like the position of elements in the grid is set in a definitive way based on the order they're found when sorted. For the first value of the first key (0), elements are sorted based on the second key values found in pairs like `(0, X)`, _i.e._ `(0, 2)` and `(0, 1)`, ignoring that there may be other values for the second key (namely 0, found in other pairs). Because it considered 1 to be the lowest value for the second key, it is located at the bottom once and for all, even if there is a lower value afterwards, and these will be located above, which makes no sense.

This will be much clearer with an example and the output:

```python
import numpy as np

import holoviews as hv
import numpy as np

hv.extension('bokeh')

def sine_curve(phase_index, freq_index):
    xvals = [0.1 * i for i in range(100)]
    phase = phases[phase_index]
    freq = frequencies[freq_index]
    return hv.Curve((xvals, [np.sin(phase + freq * x) for x in xvals]))

keys1 = [
    (0, 0), (0, 1), (1, 1), (1, 2), (2, 2)
]

keys2 = [
    (0, 2), (0, 1), (1, 1), (1, 0), (2, 0)
]
phases = [0, np.pi / 2, np.pi, 3*np.pi / 2]
dict1 = {keys: sine_curve(*keys) for keys in keys1}
dict2 = {keys: sine_curve(*keys) for keys in keys2}
gridspace1 = hv.GridSpace(dict1, kdims=["phase index", "frequency index"])
gridspace2 = hv.GridSpace(dict2, kdims=["phase index", "frequency index"])
gridspace1 + gridspace2
```

Image

hoxbro:
## Description

Fixes #7079

`GridSpace.keys(full_grid=True)` built the cross-product of the key values, but kept the second dimension's values in order of first appearance. For sparse grids such as `[(0, 2), (0, 1), (1, 1), (1, 0), (2, 0)]` the y-axis became `1, 2, 0` (bottom to top). All plotting backends use this method for the layout, so all were affected.

The full grid keys are now sorted with `util.dimension_sort` when `GridSpace.sort` is enabled (the default), matching how the regular keys are sorted.

```python
import holoviews as hv
import numpy as np

hv.extension("bokeh")

keys = [(0, 2), (0, 1), (1, 1), (1, 0), (2, 0)]
hv.GridSpace(
    {k: hv.Curve(np.sin(np.linspace(0, 10) * (k[1] + 1) + k[0])) for k in keys},
    kdims=["phase index", "frequency index"],
)
```

Added a test for the grid keys, which fails on `main`.

Example for original issue:

image

## AI Disclosure

Tool & Model: Claude Code + Opus 5
Usage: Reproduced the issue with a Playwright script, implemented the fix and test.

## Checklist


TheoMathurin:
Many thanks for the quick PR!

